PR #2, opened 2018-09-12
rewrite import paths and setup deps

BenTheElder:
- rewrite imports
- setup basic dep init including some dev-time tools for the next pr (setting up workspace & hack scripts)


BenTheElder:
/hold cancel
this unbreaks `go get sigs.k8s.io/kind` and is pruned down to just go code / licenses etc. will follow up with hack scripts to maintain it and a `WORKSPACE`

PR #3, opened 2018-09-13
add hack/update-deps.sh

BenTheElder:
.. and run it to fix `bazel build //:kind`

TODO: add verify scripts, presubmits, and guides for how to develop.


hack/.kazelcfg.json
```diff
@@ -0,0 +1,4 @@
+{
+	"GoPrefix": "k8s.io/test-infra",
```
BenTheElder: I wound up fixing this in the next commit, FYI

PR #15, opened 2018-09-20
fix symlink handling

BenTheElder:
we need to resolve symlinks when staging bits to the docker image builds, this updates the golang filesystem helpers to do that properly 

Fixes flakes when building the node image with bazel-built kubernetes bits
/kind bug
/priority critical-urgent

### DIFF
--- a/pkg/build/base/base.go
+++ b/pkg/build/base/base.go
@@ -93,7 +93,7 @@ func (c *BuildContext) Build() (err error) {
 		buildDir = filepath.Join(buildDir, "images", "base")
 
 	} else {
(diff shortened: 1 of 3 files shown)


BenTheElder:
example flake I hope this will resolve: https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/logs/ci-kubernetes-kind-conformance-parallel/531

BenTheElder:
whoo! go1.11 for formatting from now on FYI @munnerz

will add a commit to fix that now that the test images support it :-)

PR #16, opened 2018-09-20
eliminate test-infra from docker labels and hack scripts

BenTheElder:
we still had a few `test-infra`'s in here that aren't related to intentionally linking to test-infra (EG kubetest), this PR fixes those

### DIFF
--- a/pkg/build/node/node.go
+++ b/pkg/build/node/node.go
@@ -160,7 +160,7 @@ func (c *BuildContext) populateBits(buildDir string) error {
 }
 
 // BuildContainerLabelKey is applied to each build container
-const BuildContainerLabelKey = "io.k8s.test-infra.kind-build"
+const BuildContainerLabelKey = "io.k8s.sigs.kind.build"
 
 // private kube.InstallContext implementation, local to the image build
 type installContext struct {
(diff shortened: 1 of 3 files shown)


pkg/build/node/node.go
```diff
@@ -160,7 +160,7 @@ func (c *BuildContext) populateBits(buildDir string) error {
 }
 
 // BuildContainerLabelKey is applied to each build container
-const BuildContainerLabelKey = "io.k8s.test-infra.kind-build"
+const BuildContainerLabelKey = "io.k8s.sigs.kind.build"
```
BenTheElder: Per: https://docs.docker.com/config/labels-custom-metadata/#key-format-recommendations
These should be reverse DNS for a domain we "own", `io.k8s.sigs` seems natural but really we are `kind/ios.sigs.k8s` which is not a domain, so -> `io.k8s.sigs.kind.` is the new common prefix / namespace for our docker object labels

BenTheElder: @munnerz this is a breaking change of sorts FYI, but we don't exactly support long-lived clusters to begin with and this detail was not documented as a public API (yet?)

BenTheElder:
/tbr
this is very minor and I'm about to build on top of it to try to improve stability (need to use this labeling scheme for volumes too)

PR #20, opened 2018-09-21
explicitly use an infinite sleep to keep the node build container

BenTheElder:
/shrug
this one is not a big deal but it should clarify things a bit.


pkg/build/node/node.go
```diff
@@ -300,7 +300,10 @@ func (c *BuildContext) createBuildContainer(buildDir string) (id string, err err
 		// label the container to make them easier to track
 		"--label", fmt.Sprintf("%s=%s", BuildContainerLabelKey, time.Now().Format(time.RFC3339Nano)),
 		"-v", fmt.Sprintf("%s:/build", buildDir),
-		c.baseImage,
+		// the container should hang forever so we can exec in it
+		"--entrypoint=sleep",
```
BenTheElder: so this is terrible because it saves the container this way /shrug

probably we should not rely on the base image entrypoint anyhow, though, writing a fix

Issue #23, opened 2018-09-21
improve logging


BenTheElder:
- [x] make log level configurable
- [x] improve differentiation between print, info, debug etc.
  - we should use print & friends for normal user level info
  - use info for debug-ish details like "running this command"
  - use debug for trace-ish level details
  - use warn & error as appropriate
- [x] revisit structured logging
- [x] look at adding "quiet" modes to all commands

/kind cleanup
/assign

BenTheElder:
I've done most of this
/lifecycle active

BenTheElder:
xref: #109

BenTheElder:
still a major issue, probably will switch to klog / logr to match the ecosystem and rework what we log then...

BenTheElder:
this is more or less done now, we can make the logger a little nicer but we've improved a lot here

PR #24, opened 2018-09-21
image pull fixes

BenTheElder:
- try to explicitly pre-pull images as these are large
- make sure exec.CombinedOutputLines will return the lines even if the command errors so we can debug stdout / stderr
- make sure image creation failures include the lines
- also ensure the CI e2es are using the locally built image ...

/kind bug

### DIFF
--- a/pkg/exec/exec.go
+++ b/pkg/exec/exec.go
@@ -73,7 +73,7 @@ func (cmd *Cmd) runLoggingOutputOnFail() error {
 	cmd.Stdout = &buff
 	cmd.Stderr = &buff
 	err := cmd.Cmd.Run()
-	if cmd.LogOutputOnFail {
+	if err != nil && cmd.LogOutputOnFail {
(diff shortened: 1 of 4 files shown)


pkg/exec/exec.go
```diff
@@ -73,7 +73,7 @@ func (cmd *Cmd) runLoggingOutputOnFail() error {
 	cmd.Stdout = &buff
 	cmd.Stderr = &buff
 	err := cmd.Cmd.Run()
-	if cmd.LogOutputOnFail {
+	if err != nil && cmd.LogOutputOnFail {
```
BenTheElder: TODO: majorly clean up this package when we clean up logging xref #23

BenTheElder:
TBR: need to let this soak overnight to see if it sorts failures like https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/logs/ci-kubernetes-kind-conformance-parallel/683

Issue #26, opened 2018-09-22
install a default storage class


BenTheElder:
without this, many of the e2e tests will not work correctly.

/priority important-soon
/kind feature
/assign

PR #27, opened 2018-09-22
Add default storage class

BenTheElder:
This should fix https://github.com/kubernetes-sigs/kind/issues/26, and in turn improve the ability to run e2e tests.

TODO: it might be reasonable to allow overriding this. OTOH this can be overridden _after_ the cluster is up anyhow.

I also found and fixed the `The connection to the server localhost:8080 was refused - did you specify the right host or port?` error during bootup, we just needed to supply the kubeconfig to `kubectl version` when installing weave.


BenTheElder:
TBR @munnerz, this should clean up & speed up the conformance CI.

Issue #4, opened 2018-09-13
setup new repo


BenTheElder:
- [x] transfer history from [kubernetes/test-infra/kind](https://github.com/kubernetes/test-infra/tree/master/kind) - https://github.com/kubernetes-sigs/kind/pull/1
- [x] setup `dep` / vendor deps - https://github.com/kubernetes-sigs/kind/pull/2
- [x] setup scripts for managing deps and `BUILD` files -  https://github.com/kubernetes-sigs/kind/pull/3
- [x] setup `WORKSPACE` / fix `bazel build //:kind` - https://github.com/kubernetes-sigs/kind/pull/3
- [ ] ~setup guides for development and contributing~ moved to https://github.com/kubernetes-sigs/kind/issues/78, https://github.com/kubernetes-sigs/kind/issues/79
- [x] file issues with and without help-wanted [WIP], see #5, #6, ...
- [x] point old location to this repo - https://github.com/kubernetes/test-infra/pull/9382 
- [x] add presubmits [kubernetes/test-infra#9409](https://github.com/kubernetes/test-infra/pull/9409)  [kubernetes/test-infra#9410](https://github.com/kubernetes/test-infra/pull/9410) [kubernetes/test-infra#9591](https://github.com/kubernetes/test-infra/pull/9591)

/assign

BenTheElder:
Closing this out, moved usage / developer guides to their own tracking issues

PR #35, opened 2018-09-25
Reduce base image size.

tao12345666333:
Remove /tmp/cni.tgz file. 

```
# du -sh /tmp/cni.tgz
15M     /tmp/cni.tgz

```


munnerz:
You'll need to regenerate the go-bindata files to resolve the test failure. There should be a script for that in hack/

This seems like a sensible change, I'll defer to @bentheelder for lgtm however as I'm not in a position to run a full build & test right now!

tao12345666333:
@munnerz  Thanks.  Let me see.

BenTheElder:
Thanks! Looks like verify is failing for two reasons:

1) this is go 1.11 gofmt vs go < 1.11 gofmt, please use 1.11 (sorry! Will add this to development guides ref #4)

```
func (nh *nodeHandle) CopyTo(source, dest string) error {
 	cmd := exec.Command("docker", "cp")
 	cmd.Args = append(cmd.Args,
-		source, // from the source file
+		source,               // from the source file
 		nh.nameOrID+":"+dest, // to the node, at dest
 	)
```

2) I think you've got an `entrypoint` binary on disk somehow that is getting included in the generated sources:
```
-var _imagesBaseEntrypointEntrypoint = 
...
```
This shouldn't happen. I'd be curious if you have any idea how it did but in the meantime can you remove it then run hack/update-generated?

BenTheElder:
> And I have build the entrypoint binary for test build base image locally.

Did you use docker build manually? `kind build base` should automatically stage the build into a tempdir which should prevent this (and other things).

tao12345666333:
Yes, I run docker build manually. 😂😂😂
Thanks for your suggestion, I will use kind instead it.

BenTheElder:
Awesome, thanks. Wanted to make sure there wasn't a bug in that :-)

If you can fix the generated file this LGTM 👍

tao12345666333:
@BenTheElder Thanks, addressed.

BenTheElder:
hmmm did something change with approval?
/approve

Issue #21, opened 2018-09-21
Adopt kubernetes API tooling


BenTheElder:
`pkg/cluster/config/...` should adopt Kubernetes's API tooling to automatically generate deep copy and other API tooling.

To do this we need to vendor the generators and add them to `hack/update-generate.sh`

Marking this help wanted, if anyone wants to take this on feel free to contact me for help. I'll tackle this myself eventually.
/priority important-soon
/kind cleanup
/help

Lion-Wei:
Hi, Benjamin. I think I can help. :smile:

BenTheElder:
Awesome!

Yeah unfortunately you have to separately request membership in the other kubernetes orgs like this one, it's pretty easy though ref: https://github.com/kubernetes/org/issues/110

I'll consider this taken for now, thanks! :-)
/lifecycle active

BenTheElder:
@Lion-Wei has this, we still can't assign but I can remove-help (oops! done now) :-)

BenTheElder:
this is sorta done. cc @munnerz 
thanks again @Lion-Wei!

Lion-Wei:
@BenTheElder Be my guest, and I'd like shoulder more works in this repo. : )

BenTheElder:
/assign @Lion-Wei
😄

BenTheElder:
Thanks again for the work on this, opened a new issue to track any remaining cleanup work.

PR #36, opened 2018-09-26
adopt Kubernetes's API tooling to automatically generate

Lion-Wei:
xref: #21


Lion-Wei:
Hi, @BenTheElder , PTAL. And I have something not sure: 
1. Should we keep `Any` after we adopt API tool?
2. Should we move v1alpha1 Config to `pkg/cluster/config/v1alpha1`, and generate the conversion.

Not very familiar with those tools, if there are anything misconduct or mistake please let me know. : )

BenTheElder:
Thanks for working on this!

> Should we keep Any after we adopt API tool?

I think we may need some variant on Any to still mange the serialization

> Should we move v1alpha1 Config to pkg/cluster/config/v1alpha1, and generate the conversion.

Yes! That sounds great to me. Imho we should have a v1alpha1 api package, and we should always maintain a `pkg/cluster/config` with the current api version. (perhaps by just aliasing the types?)

BenTheElder:
Maybe we should break up the PRs, the first one can handle deepcopy with commits something like like:
- update gopkg.toml to vendor the tools
- run hack/update-deps.sh to actually vendor
- update `hack/update-generated.sh` to use the deepcopy generator
- actually replace the hand-rolled deepcopy
- update any tests as necessary

Then we can follow up with adopting the other portions of the API tooling? I think it will probably be easier to sort out the interfaces etc if we break it up.

BenTheElder:
actually re-reading this, its already pretty easy to review :-)

BenTheElder:
I'll work on a PR to make it possible to ignore golint for the generated API code

Lion-Wei:
Yeah, I was pretty concerned about the commit organize during this work. :relieved: 
But if we want add v1alpha1 package, then the code should be reorganized, do you mind I open another pr to add v1alpha1 package?

BenTheElder:
I'm super happy with this PR as a starting point, we just need to sort out the golint failure. I got pulled off looking into https://github.com/kubernetes-sigs/kind/issues/43 :/

munnerz:
> Should we keep Any

Based on how Kubernetes projects work, 'Any' doesn't really make sense long
term IMO.

The whole point of conversions is to allow users to write their manifests
in any API version, and the internal machinery will then convert into a
single API version so that 'kind' itself can work with those types without
having to maintain different code paths for different versions.

Wrt serialisation, when it comes to writing the manifest back out to
somewhere, we will write it in whichever version is appropriate (usually
named the 'preferred' version, however there is nothing requiring a
particular version).

It's worth noting as well, that a conversion must preserve *all* fields
when going A->B->A. This means that if a new field is added, or a type
changed, our conversion code *must* preserve this new data on the old type
*somehow* (usually achieved through annotations, adding the new field to
the old version, or some kind of casting/type conversion depending on the
change).

Once done properly, we should be able to encode, decode, convert and
default all via a runtime.Scheme instance, thus removing the need for an
interface like Any.

I'll not be able to take a look at this PR until Monday - but if this
blocks other things then I can always do a retroactive review after merge 😄

Thanks for picking this up!! I am especially looking forward to no longer
writing manually created deep copy functions 😅

BenTheElder:
Deferring to @munnerz on this who has vastly more experience with this tooling and building APIs like this :-)

BenTheElder:
the gopkg.lock / gopkg.toml conflict now. I'm working on the verify updates

BenTheElder:
oops, filing a fix for the 1-11 job

Gopkg.toml
```diff
@@ -42,7 +47,11 @@ required = [
   name = "github.com/spf13/cobra"
   version = "0.0.3"
 
-
 [[constraint]]
   branch = "master"
   name = "k8s.io/apimachinery"
```
munnerz: We should be using matching versions of code-generator and apimachinery... perhaps best to switch code-generator to master for now too, if we require master of apimachinery? /cc @BenTheElder

pkg/cluster/config/doc.go
```diff
@@ -16,4 +16,8 @@ limitations under the License.
 
 // Package config implements the current apiVersion of the `kind` Config
 // along with some common abstractions
+
+// +k8s:deepcopy-gen=package
+// +k8s:conversion-gen=sigs.k8s.io/kind/pkg/cluster/config
```
munnerz: afaict, we don't use conversion-gen right now, so this can be removed?

pkg/cluster/config/encoding/encoding.go
```diff
@@ -92,7 +92,7 @@ func Unmarshal(raw []byte) (config.Any, error) {
 	// load version
 	var cfg config.Any
 	switch version {
-	case config.APIVersion:
+	case config.SchemeGroupVersion.String():
```
munnerz: 👍 so for the time being, we use the current conversion interface/logic. SGTM.

pkg/cluster/config/register.go
```diff
@@ -0,0 +1,50 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// GroupName is the group name used in this package
+const GroupName = "kind.sigs.k8s.io"
+
+var (
+	// SchemeGroupVersion is group version used to register these objects
+	SchemeGroupVersion = schema.GroupVersion{Group: GroupName, Version: "v1alpha1"}
```
munnerz: Given we are using the kube-generators, I think it makes sense to move to a kube-style package layout too, i.e. `pkg/apis/kind/v1alpha1`.

Probably best to do that as a separate PR for now though, to reduce the blast radius for this PR! 😄

BenTheElder:
discussed with munnerz, let's get this in and follow up with improving our _full_ migration to the tooling next. I'll send some more sweeping changes today to fix up kind ref #59

PR #37, opened 2018-09-26
drop --tag flags

BenTheElder:
I think these are just more confusing than having `--image` and `--base-image` tags. normalizing all commands to take `--image` (and `--base-image` for building the node image).

### DIFF
--- a/pkg/build/base/base.go
+++ b/pkg/build/base/base.go
@@ -24,26 +24,20 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	"sigs.k8s.io/kind/pkg/build/base/sources"
-	"sigs.k8s.io/kind/pkg/docker"
 	"sigs.k8s.io/kind/pkg/exec"
 	"sigs.k8s.io/kind/pkg/fs"
 )
 
-// DefaultImageName is the default name of the built base image
-const DefaultImageName = "kindest/base"
-
-// DefaultImageTag is the default tag of the built base image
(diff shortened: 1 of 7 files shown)


pkg/build/base/base.go
```diff
@@ -24,26 +24,20 @@ import (
...
 	"sigs.k8s.io/kind/pkg/exec"
 	"sigs.k8s.io/kind/pkg/fs"
 )
 
-// DefaultImageName is the default name of the built base image
-const DefaultImageName = "kindest/base"
-
-// DefaultImageTag is the default tag of the built base image
-const DefaultImageTag = "latest"
+// DefaultImage is the default tag of the built base image
```
amwat: nit: default "name"

pkg/build/node/node.go
```diff
@@ -32,19 +32,11 @@ import (
 	"sigs.k8s.io/kind/pkg/fs"
 )
 
-// DefaultImageName is the default name for the built image
-const DefaultImageName = "kindest/node"
+// DefaultImage is the default tag for the built image
```
amwat: same

spiffxp:
thanks for the cc, fine by me

pkg/build/base/base.go
BenTheElder: er docker is super unclear on this, it's the `-t` (tag) argument when building. /shrug

Issue #7, opened 2018-09-13
support all the kubeadm config flavors in templates


neolit123:
the kubeadm configuration is moving fast and ever changing.
it was `v1alpha2` in 1.11 and in 1.12 it would be `v1alpha3`.
even more changes are planned for `v1beta1` in 1.13.

based on a parseable `kuberentes-version` flag of sorts, kind would have to pick the right template to construct a kubeadm config.

for ref:
https://github.com/kubernetes-sigs/kind/blob/130a7fe9212c3fba6fc84d785759785218ce5128/pkg/cluster/kubeadm/config.go#L55
kubeadm's api:
https://github.com/kubernetes/kubernetes/tree/master/cmd/kubeadm/app/apis/kubeadm

/assign
/kind feature
/priority important-longterm

BenTheElder:
Definitely, thanks for filing this.

We know the Kubernetes version from `/kind/version` (a regular text file containing `gitVersion` for the kubernetes installed at build time) [in the node image](https://github.com/kubernetes-sigs/kind/blob/master/docs/node-image.md#design), the value of which is supplied to the template as an arg. We probably want to supply a parsed version of this to the template so it can select over minor / patch etc.

neolit123:
understood, just had a look how the version is passed to the template.

BenTheElder:
Counting the PR above and discussion here as making it active for now :-)
Trying out the active label for tracking in this repo.

neolit123:
> Counting the PR above and discussion here as making it active for now

ok, makes sense :+1:

BenTheElder:
https://github.com/kubernetes/kubernetes/pull/68922 merged but kind is currently broken and we'll need to wait for it to be synced out by the publishing bot

neolit123:
i need to figure out what is the best way vendor `pkg/util/version` in the manifest list e2e that i'm doing too, because i homebrew a semVer in there.

the code has to live in `k/kubeadm` for this test.

neolit123:
looking into this right now with pseudo version checks.
do we have to vendor `util/version` in the kind tree at this point?

BenTheElder:
Thanks again! fixed in https://github.com/kubernetes-sigs/kind/pull/46

Will probably follow up with revisiting the templates a bit, but we do have support now :-)

PR #45, opened 2018-09-28
add nsswitch work around

BenTheElder:
This works around https://github.com/kubernetes/kubernetes/issues/69195 for the short term

This will almost unbreak CI, next we need to solve #7 as v1alpha2 is no longer supported in master.

### DIFF
--- a/pkg/cluster/kubeadm/config.go
+++ b/pkg/cluster/kubeadm/config.go
@@ -61,6 +61,15 @@ clusterName: {{.ClusterName}}
 # to the cluster after rewriting the kubeconfig to point to localhost
 apiServerCertSANs: [localhost]
 kubernetesVersion: {{.KubernetesVersion}}
+# TODO(bentheelder): fix this upstream!
+# we need nsswitch.conf so we use /etc/hosts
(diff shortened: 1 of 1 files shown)


BenTheElder:
fixes https://github.com/kubernetes-sigs/kind/issues/43

pkg/cluster/kubeadm/config.go
```diff
@@ -61,6 +61,15 @@ clusterName: {{.ClusterName}}
 # to the cluster after rewriting the kubeconfig to point to localhost
 apiServerCertSANs: [localhost]
 kubernetesVersion: {{.KubernetesVersion}}
+# TODO(bentheelder): fix this upstream!
+# we need nsswitch.conf so we use /etc/hosts
+# https://github.com/kubernetes/kubernetes/issues/69195
```
BenTheElder: once this is fixed upstream we should clean this up and try to only do it when necessary or just remove it. I'll work on an upstream fix tomorrow

BenTheElder:
TBR again as this is key to unbreaking CI until we can get something merged upstream ...

CC @munnerz

PR #59, opened 2018-10-08
switch cli to kind verb noun

### LINKED ISSUE #56: normalize CLI
IE make all commands fit the form: `kind [verb] [noun]`

We should do this to match [`clusterctl`](https://github.com/kubernetes-sigs/cluster-api/tree/master/clusterctl), `kubectl`, `kops`, etc...

/kind cleanup
we should prioritize this to avoid breaking users that start to depend on the CLI...
/priority important-soon
/lifecycle active
/assign

FYI @mitar @Lion-Wei @m1093782566 @munnerz @spiffxp 
I think we can temporarily support both `kind create` and `kind create cluster` and print a warning, but we should eliminate `kind create` before we cut any release.

This leaves us room to expand the CLI in a consistent way in the future.

BenTheElder:
fixes #56 
also starts to switch cobra commands to RunE instead of Run, which will let us test them with mocking better later, instead of calling fatal within the commands.

I'll follow up with volume fixes, the debug command, and logging cleanup soon!

### DIFF
--- a/cmd/kind/kind.go
+++ b/cmd/kind/kind.go
@@ -18,6 +18,8 @@ limitations under the License.
 package kind
 
 import (
+	"os"
+
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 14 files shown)


cmd/kind/kind.go
```diff
@@ -54,6 +56,7 @@ func Main() {
 		TimestampFormat: "0102-15:04:05",
 	})
 	if err := Run(); err != nil {
-		log.Fatal(err)
+		os.Stderr.WriteString(err.Error())
+		os.Exit(-1)
```
amwat: Why the switch from log.fatal?

BenTheElder: part of #23, cli tools should print an error to stdout and exit non-zero instead of `FATA ....` ... I'm going to reserve logging for actual debug info soon and use println and friends for actual user oriented output.

failing to run needs to be user oriented rather than logging. if this were eg a server I'd still do fatal

Issue #30, opened 2018-09-22
pre-pull images that are not part of the build


BenTheElder:
When we build Kubernetes we are able to preload built images into the `kind` node image, a typical kubernetes build does not actually build important relatively fixed images like `etcd` though, we should look at querying kubeadm for these at build time and pre-pulling them. 

Possibly we can use https://github.com/google/containerregistry to pull these directly into the built image without docker. These tools have worked pretty well for us underneath `rules_docker` in test-infra.

/priority important-soon
/kind enhancement
/assign

BenTheElder:
running in the default image:
```
root@b0816c5f4071:/# kubeadm config images list
k8s.gcr.io/kube-apiserver-amd64:v1.11.3
k8s.gcr.io/kube-controller-manager-amd64:v1.11.3
k8s.gcr.io/kube-scheduler-amd64:v1.11.3
k8s.gcr.io/kube-proxy-amd64:v1.11.3
k8s.gcr.io/pause:3.1
k8s.gcr.io/etcd-amd64:3.2.18
k8s.gcr.io/coredns:1.1.3
```
Checking upstream, these are ~ mapped by kubernetes version, not config, so we should be able to pre-pull them at build time easily, we probably want to pull `.*etcd.*` and `.*pause.*`

ref: https://github.com/kubernetes/kubernetes/blob/69da1f83995f0fa1e555dfacef9e97abc63b9333/cmd/kubeadm/app/constants/constants.go#L376

neolit123:
etcd, pause and coredns versions are hardcoded in a constants file. their version is bumped on each release. the control plane version in 1.11 was mapped to the `stable-1.11` endpoint, in 1.12 we moved to `stable-1` (this needs a cherry pick enhancement).

BenTheElder:
Excellent, thanks! For the control plane in `docker` and `bazel` builds we have locally built images, but not in `apt`. For all of them we don't have pause, DNS, etcd since those aren't built in k8s.

BenTheElder:
Fixed in https://github.com/kubernetes-sigs/kind/pull/89, will push new images.

PR #65, opened 2018-10-12
pre-pull images that are not part of the build

fabriziopandini:
fixes: https://github.com/kubernetes-sigs/kind/issues/30

Since docker in docker is not available at build time, I'm using docker of the host machine for pre-pulling images. This has the advantage that pre-pulled images are reused in subsequent builds...

@BenTheElder 
I saw only now this was already active 😭 feel free to close it

### DIFF
--- a/pkg/build/node/node.go
+++ b/pkg/build/node/node.go
@@ -21,6 +21,8 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"path/filepath"
+	"regexp"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


BenTheElder:
Thanks for the PR!

I like the pull to host approach, but I think maybe we need a slightly different approach regarding detecting which images. The presubmit is failing because we're trying to pull images that we actually built that don't exist upstream.

fabriziopandini:
@BenTheElder 
If we rely on kubeadm config images what I can do is to detect that kebernetes version is a build version and then prefix with ci/ the version number passed to kubeadm (this should shift from `k8s.gcr.io` to `gcr.io/kubernetes-ci-images`). And as additional measure I can make pre-pull errors not blocking.
Wdyt?

BenTheElder:
It shouldn't be pulling this image at all as it has a local copy. It shouldn't need to pull any images from ci/ as we build all of those (except apt "build", which also shouldn't have ci/ images...)

BenTheElder:
Thanks for working on this, I implemented something pretty similar in https://github.com/kubernetes-sigs/kind/pull/89, though it detects which images don't need pulling a bit differently.

fabriziopandini:
@BenTheElder sorry for not following up.
I plan to get back on helping on kind after release cut

PR #68, opened 2018-10-19
explicitly set the API server to a random port

BenTheElder:
/hold
cc @Katharine


pkg/cluster/node.go
```diff
@@ -65,9 +78,9 @@ func createNode(name, image, clusterLabel string) (handle *nodeHandle, err error
 			"--name", name, // ... and set the container name
 			// label the node with the cluster ID
 			"--label", clusterLabel,
-			"--expose", "6443", // expose API server port
-			// pick a random ephemeral port to forward to the API server
-			"--publish-all",
+			// publish selected port for the API server
+			"--expose", fmt.Sprintf("%d", port),
+			"-p", fmt.Sprintf("%d:%d", port, port),
```
munnerz: If we keep this as %d:6443, we can avoid the config template change I think? Not sure what the original motivations for the PR are

pkg/cluster/kubeadm/config.go
```diff
@@ -56,7 +58,10 @@ const DefaultConfigTemplateAlphaV1orV2 = `# config generated by kind
 apiVersion: kubeadm.k8s.io/v1alpha2
 kind: MasterConfiguration
 kubernetesVersion: {{.KubernetesVersion}}
-# TODO(bentheelder): fix this upstream!
+clusterName: {{.ClusterName}}
```
munnerz: Do we need to set a cluster name? Is this related to setting the port?

pkg/cluster/kubeadm/config.go
```diff
@@ -65,7 +70,6 @@ apiServerExtraVolumes:
   hostPath: /etc/nsswitch.conf
   writeable: false
   pathType: FileOrCreate
-clusterName: {{.ClusterName}}
```
BenTheElder: previously set here

pkg/cluster/node.go
BenTheElder: making sure things consistently hit the correct port by informing everything what the api server port is.

neolit123:
the InitConfiguration part LGTM

i will double check when 1.13 is near, but for v1beta1, and the current use case we can just do `DefaultConfigTemplateAlphaV3orBetaV1` it feels like.

pkg/cluster/node.go
neolit123: i assume that the random port is motivated by security?

BenTheElder: actually because we support multiple clusters on one machine and we set localhost:port to point to them, we already set a random port today, we just don't use the same port within the container which makes some inconsistency with the api server port

BenTheElder:
/hold cancel
I think we should actually do this. We're going to switch to kustomize to avoid issues with dowstream users setting the template, instead they will set a patch or patches to apply.

pkg/cluster/node.go
```diff
@@ -83,9 +96,9 @@ func createNode(name, image, clusterLabel string) (handle *nodeHandle, err error
 		handle = &nodeHandle{name}
 	}
 	if err != nil {
-		return handle, errors.Wrap(err, "docker run error")
+		return handle, 0, errors.Wrap(err, "docker run error")
 	}
-	return handle, nil
+	return handle, port, nil
```
amwat: Add port to the nodeHandle struct?

BenTheElder: the port is totally orthogonal to the node handle.

Issue #39, opened 2018-09-26
add environment diagnostics


BenTheElder:
we should have a number of checks when kind fails, to help diagnose issues in the environment, ideally this can also be invoked by another command, some rough ideas:
- [ ] check if there are existing containers tagged with the current cluster when creating
- [ ] more gracefully handle lack of cluster artifacts when deleting
- [ ] gracefully handle incompatible docker / lack of docker
- [ ] base image builds should check that go exists

/priority important-soon

BenTheElder:
/assign
The "cluster already exists" / "no cluster exists" for create and delete would be nice to get into our first release.

BenTheElder:
this work is now stale, will come back to it later

BenTheElder:
I beleive @neolit123 added the "existing cluster name" check

alejandrox1:
haven't been participating as much as I'd like to but i'd like to get back to a more regular workflow and start working on this. I'll get something for handling incompatible/lack of docker :slightly_smiling_face:

aojea:
@BenTheElder please add a check for IPv6 if IPfamily with IPv6 to the list😅 
 it took me 30 mins to realize it was failing because the docker daemon didn´t have IPv6 enabled

stmcginnis:
Another possible check - we've had some issues due to `DOCKER_DEFAULT_PLATFORM` being set. Could include a check for that as a data point when collecting troubleshooting data.

Another common issue could be helped with collecting any available Docker/Docker Desktop information.

PR #69, opened 2018-10-23
Add status tracking with loading spinners, clean up logging

BenTheElder:
- Adds a `--loglevel` flag for setting the logrus log level
- moves the default log level to warn
- adds a nice loading spinner system that tracks phases of creating the cluster for friendlier output while waiting for the cluster to boot

TODO (followup): 
- log kubeadm output at trace level (?) after upgrading logrus, or log it to a debug log file
- give helpful output when kubeadm init fails
- add preflight checks for docker, previous clusters, etc. #39 
- possibly just replace node.Run with RunQ


BenTheElder:
cc @Katharine @munnerz 

This gives shiny output like:

```console
$ kind create cluster --name=asdf
Creating cluster 'kind-asdf' ...
 ✓ Ensuring node image (kindest/node:v1.11.3@sha256:855562266d5f647b5770a91c76a0579ed6487eb5bf1dfe908705dad285525483) 🖼
 ✓ [kind-asdf-control-plane] Creating node container 📦
 ✓ [kind-asdf-control-plane] Fixing mounts 🗻
 ✓ [kind-asdf-control-plane] Starting systemd 🖥
 ✓ [kind-asdf-control-plane] Waiting for docker to be ready 🐋
 ✓ [kind-asdf-control-plane] Starting Kubernetes (this may take a minute) ☸

export KUBECONFIG="/Users/bentheelder/.kube/kind-config-asdf"
kubectl cluster-info
```

BenTheElder:
trying to record this with asciinema is amusing https://asciinema.org/a/kNsHMXcXpRAj4z0lUqM8l4CRV

BenTheElder:
2x speed gif: ezgif-1-a66c6f2cdbf8

neolit123:
nice :+1:

cmd/kind/kind.go
```diff
@@ -20,40 +20,76 @@ package kind
...
 		Short: "kind is a tool for managing local Kubernetes clusters",
 		Long:  "kind creates and manages local Kubernetes clusters using Docker container 'nodes'",
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			return runE(flags, cmd, args)
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return cmd.Help()
+		},
 	}
+	cmd.PersistentFlags().StringVar(&flags.LogLevel, "loglevel", defaultLevel.String(), "logrus log level")
```
munnerz: Can we add `e.g. info, error, warn` or something to this? It gets quite frustrating having to remember what strings to use for all different log levels 😅

cmd/kind/kind.go
```diff
@@ -20,40 +20,76 @@ package kind
...
 	cmd.AddCommand(create.NewCommand())
 	cmd.AddCommand(delete.NewCommand())
 	return cmd
 }
 
+func runE(flags *Flags, cmd *cobra.Command, args []string) error {
+	level := defaultLevel
+	parsed, err := log.ParseLevel(flags.LogLevel)
+	if err != nil {
+		log.Warnf("Invalid log level '%s', defaulting to '%s'", flags.LogLevel, level)
```
munnerz: nit: May be worth doing {something} to ensure this message is always visible? `defaultLevel` is currently `warn`, but I'm slightly concerned if this gets changed in future 🙄

pkg/cluster/node.go
```diff
@@ -209,7 +238,7 @@ func tryUntil(until time.Time, try func() bool) bool {
 // LoadImages loads image tarballs stored on the node into docker on the node
 func (nh *nodeHandle) LoadImages() {
 	// load images cached on the node into docker
-	if err := nh.Run(
+	if err := nh.RunQ(
```
munnerz: It'd be awesome if we can make `--loglevel=debug` also print all of these messages, for debug purposes.

From what I can tell, with this patch there will no longer be any way to debug failures.

Also, if a particular step fails can we make it print out the stdout/stderr of that failed command?

BenTheElder: I was thinking maybe `--loglevel=trace` should do this after we upgrade logrus, but debug actually sounds good.

Definitely we should print out failed command output. Especially for kubeadm.

cmd/kind/kind.go
BenTheElder: Hmm, we could just make this yet another thing that does not actually use logging, I almost wound up writing our own small logging package :-)

cmd/kind/kind.go
BenTheElder: Yeah, this flag info is not particularly helpful.

BenTheElder: added this

pkg/cluster/node.go
BenTheElder: I think what we really want is:
- when critical things fail, inform the user and tell them what they can do about it
- if the log level is cranked high enough, log even "silent" command output on exit

cmd/kind/kind.go
BenTheElder: we're discussing broader logging options out of band, but this seems to be fine for now.

BenTheElder:
Lots of log output will still mostly work with this, you can set `--loglevel` across all commands now.
FWIW, I've done quite a bit of `kind` debugging now, and the output from the command we call has not been helpful. I'll be PRing a kind command for obtaining useful debugging info in a follow up. https://github.com/kubernetes-sigs/kind/pull/69#discussion_r227423197

PR #74, opened 2018-10-24
Add a debug flag to retain the broken nodes for debugging.

tao12345666333:
xref: https://github.com/kubernetes-sigs/kind/issues/23

### DIFF
--- a/cmd/kind/create/cluster/createcluster.go
+++ b/cmd/kind/create/cluster/createcluster.go
@@ -30,6 +30,7 @@ type flagpole struct {
 	Name      string
 	Config    string
 	ImageName string
+	Retain    bool
 }
 
 // NewCommand returns a new cobra.Command for cluster creation
@@ -46,6 +47,7 @@ func NewCommand() *cobra.Command {
 	cmd.Flags().StringVar(&flags.Name, "name", "1", "cluster context name")
 	cmd.Flags().StringVar(&flags.Config, "config", "", "path to a kind config file")
 	cmd.Flags().StringVar(&flags.ImageName, "image", "", "node docker image to use for booting the cluster")
+	cmd.Flags().BoolVar(&flags.Retain, "retain", false, "whether retain the broken nodes for debugging")
 	return cmd
 }
 
@@ -67,7 +69,7 @@ func run(flags *flagpole, cmd *cobra.Command, args []string) {
 		log.Fatal("Aborting due to invalid configuration.")
 	}
 	// create a cluster context and create the cluster
-	ctx, err := cluster.NewContext(flags.Name)
+	ctx, err := cluster.NewContext(flags.Name, flags.Retain)
 	if err != nil {
 		log.Fatalf("Failed to create cluster context! %v", err)
 	}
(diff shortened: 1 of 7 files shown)


cmd/kind/delete/cluster/deletecluster.go
```diff
@@ -25,7 +25,8 @@ import (
 )
 
 type flags struct {
-	Name string
+	Name  string
+	Debug bool
```
BenTheElder: /hold
I don't think the flag should literally be `debug`, and I'm having second thoughts about even having this flag, it's sort of a weird mode to enable to begin with. cc @munnerz

BenTheElder:
Thanks for the PR btw, I think this might need more discussion.

cmd/kind/delete/cluster/deletecluster.go
tao12345666333: > /hold
> I don't think the flag should literally be `debug`, and I'm having second thoughts about even having this flag, it's sort of a weird mode to enable to begin with. cc @munnerz

Sure, we can have other names or modify this behavior. 
Or modify this `retention mode` to a global scope? Bind this behavior to log level? 

I want to know your thoughts. Thanks.

tao12345666333: ok, it can also clearly indicate the meaning.

tao12345666333: I will start to modify it later.

tao12345666333:
@BenTheElder  Addressed. Thanks.

cmd/kind/create/cluster/createcluster.go
```diff
@@ -47,6 +48,7 @@ func NewCommand() *cobra.Command {
 	cmd.Flags().StringVar(&flags.Name, "name", "1", "the cluster context name")
 	cmd.Flags().StringVar(&flags.Config, "config", "", "path to a kind config file")
 	cmd.Flags().StringVar(&flags.ImageName, "image", "", "node docker image to use for booting the cluster")
+	cmd.Flags().BoolVar(&flags.Retain, "retain", false, "whether retain the broken nodes for debugging")
```
BenTheElder: tiny nit, maybe description should be: "retain nodes for debugging when cluster creation fails"? even if a node is not broken, this flag will retain all of them in the near future.

tao12345666333: ok, I will update it.

cmd/kind/delete/cluster/deletecluster.go
```diff
@@ -41,12 +42,13 @@ func NewCommand() *cobra.Command {
 		},
 	}
 	cmd.Flags().StringVar(&flags.Name, "name", "1", "the cluster name")
+	cmd.Flags().BoolVar(&flags.Retain, "retain", false, "whether retain the broken nodes for debugging")
```
BenTheElder: delete probably shouldn't take a retain? since it's always going to delete

pkg/cluster/context.go
```diff
@@ -44,6 +44,7 @@ import (
 type Context struct {
 	name   string
 	status *logutil.Status
+	retain bool
```
BenTheElder: maybe just export this field for now, let create command set it, and I'll refactor in a follow-up. `NewContext` needs refactoring anyhow (going to delay validation..)

cmd/kind/delete/cluster/deletecluster.go
tao12345666333: sure

tao12345666333:
@BenTheElder  PTAL, thanks. :heart:

BenTheElder:
will follow up with a refactor to NewContext, thanks again for the PR!
/lgtm
/approve
/hold cancel

PR #75, opened 2018-10-24
add `kind get kubeconfig-path`

BenTheElder:
... so users can stop hard-coding the KUBECONFIG path. cc @munnerz @liztio @Katharine


liztio:
thanks for this!

PR #81, opened 2018-10-26
[WIP] add tooling for version management and building releases

BenTheElder:
This allows us to:
- build release binaries with `hack/build/release.sh`
- generate, commit & tag a release, then generate, commit & tag a new pre-release version (with the patch version bumped and `alpha` pre-release version set) with `VERSION=0.1.0 hack/release.sh`
- get the current version with `kind --version`

The version constant is in an autogenerated file, for which there is new go-generate tooling.
I opted for this route over using ldflags and x def to leave `go get` / `go build` etc. working fine for builds. 

This means we don't get fancy details like the git commit, but for published binaries we should use the release tags anyhow. 

TODO: vendor the semver library used to bump the release (github.com/coreos/go-semver/semver)

/assign @munnerz


BenTheElder:
I think this is mostly sane, it will also need tag pushing and some more discussion.

Holding off for now since we're not quite ready to release yet anyhow, we should fix eg #82

BenTheElder:
going to do something a lot simpler today. we'll come back to something more complex later.

PR #92, opened 2018-11-08
Remove existing packages

tao12345666333:
Because we have changed the base image to `Ubuntu:18.04`, the `bash` and `mount` packages are already existed.

xref:
https://github.com/kubernetes-sigs/kind/pull/89

Signed-off-by: Jintao Zhang <zhangjintao9020@gmail.com>

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -52,8 +52,8 @@ RUN chmod +x /usr/local/bin/clean-install
 RUN clean-install \
       apt-transport-https ca-certificates curl software-properties-common gnupg2 lsb-release \
       systemd systemd-sysv libsystemd0 \
-      iptables iproute2 ethtool socat util-linux mount ebtables udev kmod aufs-tools \
-      bash rsync \
+      iptables iproute2 ethtool socat util-linux ebtables udev kmod aufs-tools \
+      rsync \
     && find /lib/systemd/system/sysinit.target.wants/ -name "systemd-tmpfiles-setup.service" -delete \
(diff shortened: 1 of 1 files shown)


images/base/Dockerfile
```diff
@@ -52,8 +52,8 @@ RUN chmod +x /usr/local/bin/clean-install
 RUN clean-install \
       apt-transport-https ca-certificates curl software-properties-common gnupg2 lsb-release \
       systemd systemd-sysv libsystemd0 \
-      iptables iproute2 ethtool socat util-linux mount ebtables udev kmod aufs-tools \
```
BenTheElder: hmm we explicitly need these though, what's the downside to installing them?
we may change the base image again in the future. 🤔

tao12345666333: Duplicate installations only add time, but because they already exist, there is nothing else going on.

tao12345666333: For the current situation, the base images are built ahead of time and it seems that there will be no problem. It’s just my habit of building an image that will remove unwanted things.

BenTheElder: 👍 

It shouldn't add much time though? Ot just needs to look up that they are installed and then skip them. 

The base image is intended to only be built when we change it / upgrade a package.  Ideally not very often. I might be missing a use case though 🙃

tao12345666333: yes. I will close it.

BenTheElder: thanks again anyhow! always appreciate PRs :-)

PR #93, opened 2018-11-08
fix timeout bug

BenTheElder:
🤦‍♂️ 

/assign @krzyzacy @amwat


BenTheElder:
Probably. It's only like 5 lines though, refactoring mistakes aside. Not worth a dependency to understand / keep up to date / maintain correct usage of.

PR #96, opened 2018-11-09
switch to yaml.UnmarshallStrict for config

BenTheElder:
FYI @munnerz

I *think* this is the last breaking change to squeeze in before sorting out release. Other changes will be preflight checks and feature enhancements (like log dumping) which can come at any time.

FYI @liztio :grimacing: hopefully recent changes should just make kind faster without affecting y'all.

/kind cleanup

### DIFF
--- a/pkg/cluster/config/encoding/encoding.go
+++ b/pkg/cluster/config/encoding/encoding.go
@@ -23,7 +23,7 @@ import (
 	"fmt"
 	"io/ioutil"
 
-	"github.com/ghodss/yaml"
+	"sigs.k8s.io/yaml"
 
 	"sigs.k8s.io/kind/pkg/cluster/config"
 )
--- a/pkg/cluster/config/encoding/encoding.go
+++ b/pkg/cluster/config/encoding/encoding.go
@@ -79,7 +79,7 @@ func detectVersion(raw []byte) (version string, err error) {
 	return "", fmt.Errorf("invalid version: %v", c.APIVersion)
 }
 
-// Unmarshal is an apiVersion aware yaml.Unmarshall for config.Any
+// Unmarshal is an apiVersion aware yaml.UnmarshalStrict for config.Any
 func Unmarshal(raw []byte) (config.Any, error) {
 	if raw == nil {
 		return nil, fmt.Errorf("nil input")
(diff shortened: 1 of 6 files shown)


pkg/cluster/config/encoding/encoding.go
```diff
@@ -89,13 +89,20 @@ func Unmarshal(raw []byte) (config.Any, error) {
...
 	switch version {
 	case config.SchemeGroupVersion.String():
-		cfg = &config.Config{}
+		// this allows strict unmarshaling with the version header
+		layout := &struct {
+			configHeader
+			config.Config
+		}{}
+		cfg = &layout.Config
+		rawCfg = layout
```
BenTheElder: sliiightly gross, but it keeps the coupling pretty low, this function is the only thing that needs to know about actual config + the header, just a tiny bit of boilerplate here for each version, and we gain error output on unknown fields

neolit123:
i think i'm seeing a bug when parsing a strict KubeletConfiguration using this library.
https://github.com/kubernetes-sigs/yaml/issues/10

will have a look soon.

BenTheElder:
uh oh, skimmed that yaml bug, is there a workaround?

neolit123:
seems to not be a bug in the yaml library.
what are you seeing?

the problem i found is that the kubelet configuration has to be unmarshaled into the external configuraiton in https://github.com/kubernetes/kubelet/tree/master/config/v1beta1
and not in the internal kubelet configuration type https://github.com/kubernetes/kubernetes/tree/master/pkg/kubelet/apis/config

i have a PR in flight to enable strict unmarshal for kubeadm in 1.13:
https://github.com/kubernetes/kubernetes/pull/70901

BenTheElder:
kind is not yet using v1beta1, I haven't had any issues yet but I'm also not really using the patches terribly much yet. we should switch for >= 1.13 as well, will do that in a moment...

neolit123:
enabling strict unmarshal from the get go is a very wise thing to do!
api-machinery, kubectl, kubeadm etc are in a state where we can throw warnings but we cannot error for a variety of reasons.

in terms of kubeadm's v1beta1 the state is pretty much ready for release at this point.

PR #102, opened 2018-11-13
refactoring: convert exec.Cmd to an interface, break node logic into …

BenTheElder:
…another package

This is the first part of major code cleanup. This will make it possible to test with fake exec later (even if we say, switch to the docker API, we still need to execute some actual commands on the nodes etc.). 

The other plan is to eventually have nodes implement the exec.Cmder and minimize the exposed surface of that logic so we generally have less direct calls to `os/exec.Command` left and right, which should further help with future cleanup ...

### DIFF
--- a/pkg/exec/exec.go
+++ b/pkg/exec/exec.go
@@ -14,66 +14,76 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Package exec contains helpers for os/exec
+// Package exec contains an interface for executing commands, along with helpers
+// TODO(bentheelder): add standardized timeout functionality & a default timeout
+// so that commands cannot hang indefinitely (!)
 package exec
 
 import (
 	"bufio"
 	"bytes"
+	"io"
 	"os"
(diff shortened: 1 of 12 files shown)


pkg/exec/exec.go
```diff
@@ -14,66 +14,76 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Package exec contains helpers for os/exec
+// Package exec contains an interface for executing commands, along with helpers
+// TODO(bentheelder): add standardized timeout functionality & a default timeout
```
BenTheElder: this is a huge follow up I intend to do, reworking this so that Run() respects a standardized timeout whether running commands locally or on the nodes, commands hang sometimes, particularly when we have bugs, we should ensure that no command can hang indefinitely.

BenTheElder: I think handling this properly is pretty key to robustness

BenTheElder:
/hold cancel
will iterate on this further in a follow-up

PR #104, opened 2018-11-13
Fix build node-image cmd and add user docs

### LINKED ISSUE #103: Build node-image using apt build type fails
Trying to build the `node-image` using the `apt` type fails:
```
INFO[15:00:54] Pulling image: k8s.gcr.io/kube-apiserver:v1.12.2 ... 
ERRO[15:00:55] Image build Failed! exit status 1            
FATA[15:00:56] Error building node image: exit status 1
```

The problems seems to be that during the build process, `kind` tries to save the required images to a temporary directory of the type `/tmp/kind-node-imageXXXXXXXXX/bits/images`:
https://github.com/kubernetes-sigs/kind/blob/6177a98223feedc7b8ab2e860b2311cb1f739acf/pkg/build/node/node.go#L368-L369

Extracting the output from the `docker.Save` command:
```
INFO[15:00:55] exit status 1                                
INFO[15:00:55] failed to save image: unable to validate output path: directory "/tmp/kind-node-image241009176/bits/images" does not exist
```

## Possible solution
During a previous step in the build process, build artifacts are moved to the temporary build directory by `populateBits`. This call will create the sub directory `bits` within the build directory `/tmp/kind-node-imageXXXXXXXX`.

https://github.com/kubernetes-sigs/kind/blob/6177a98223feedc7b8ab2e860b2311cb1f739acf/pkg/build/node/node.go#L150-L167

We could add another `exec.Command` here to create the `images` sub directory within `bits`.

Another possibility, would be to do this when the images are being pulled in (here we could more easily track where directories are being created)
https://github.com/kubernetes-sigs/kind/blob/6177a98223feedc7b8ab2e860b2311cb1f739acf/pkg/build/node/node.go#L359-L375

/kind bug

alejandrox1:
Resolves #103 
/kind bug
/kind documentation
    
### DIFF
--- a/cmd/kind/build/build.go
+++ b/cmd/kind/build/build.go
@@ -30,7 +30,7 @@ func NewCommand() *cobra.Command {
 		// TODO(bentheelder): more detailed usage
 		Use:   "build",
 		Short: "Build one of [base-image, node-image]",
-		Long:  "Build one of [base-image, node-image]",
+		Long:  "Build the base node image (base-image) or the node image (node-image)",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return cmd.Help()
 		},
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,7 +1,9 @@
 # `kind` Documentation
 
-For the primary design start with [design.md]
+For the primary design start with [design][design].
 
-<!--TODO(bentheelder): add usage docs, link to some here and in the root README-->
+For usage documentation see the [user guide][user guide].
 
-[design.md]: ./design.md
+
+[design]: ./design.md
+[user guide]: ./user-guide.md
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -1,0 +1,53 @@
+# User Guide
+
+## Advanced
+
+### Building the base and node images
+#### Base image
+`kind` runs a local kubernetes cluster by using Docker containers as "nodes."
+`kind` uses the [`node-image`][node image] to run kubernetes artifacts, such 
+as `kubeadm` or `kubelet`.
+The `node-image` in turn is built off the [`base-image`][base image], which
+installs all the dependencies needed for Docker and Kubernetes to run in a
+container.
+
+To build the `base-image` we use the `build` command:
+```
+$ kind build base-image
+```
+
+If you want to specify the path to the base image source files you can use the
+`--source` flag.
+If `--source` is not specified, `kind` has enbedded the contents of the in
+default base image in [`pkg/build/base/sources`][pkg/build/base/sources] and
+will use this to build it.
+
+By default, the base image will be tagged as `kindest/base:latest`.
+If you want to change this, you can use the `--image` flag.
+
+```
+$ kind build base-image --image base:v0.1.0
+```
+
+#### Node image
+Currently, `kind` supports three different ways to build a `node-image`: via
+`apt`, or if you have the [kubernetes][kubernetes] source in your host machine
+(`$GOPATH/src/k8s.io/kubernetes`), by using `docker` or `bazel`.
+To specified the build type use the flag `--type`.
+`Kind` will default to using the build type `docker` if none is specified.
+
+```
+$ kind build node-image --type apt
+```
+
+Similarly as for the base-image command, you can specify the name and tag of
+the resulting node image using the flag `--image`.
+
+If you previously changed the name and tag of the base image, you can use here
+the flag `--base-image` to specify the name and tag you used.
+
+
+[node image]: ./base-image.md
+[base image]: ./node-image.md
+[pkg/build/base/sources]: ./../pkg/build/base/sources
+[kubernetes]: https://github.com/kubernetes/kubernetes
--- a/pkg/build/node/node.go
+++ b/pkg/build/node/node.go
@@ -188,6 +188,10 @@ func (c *BuildContext) getBuiltImages() (sets.String, error) {
 // BuildContainerLabelKey is applied to each build container
 const BuildContainerLabelKey = "io.k8s.sigs.kind.build"
 
+// DockerImageArchives is the path within the node image where image archives
+// will be stored.
+const DockerImageArchives = "/kind/images"
+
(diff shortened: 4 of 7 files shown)


cmd/kind/build/build.go
```diff
@@ -30,7 +30,7 @@ func NewCommand() *cobra.Command {
 		// TODO(bentheelder): more detailed usage
 		Use:   "build",
 		Short: "Build one of [base-image, node-image]",
-		Long:  "Build one of [base-image, node-image]",
+		Long:  "Build the base node image (base-image) or the node image (node-image)",
```
BenTheElder: what does this look like at the CLI when printed? 
the short usage listing the valid options seems really handy, though obviously it would be great to also explain those options :-)

pkg/build/node/node.go
```diff
@@ -355,6 +355,12 @@ func (c *BuildContext) prePullImages(dir, containerID string) error {
 		return err
 	}
 
+	// Create "images" subdir.
+	imagesDir := path.Join(dir, "bits", "images")
```
BenTheElder: at https://github.com/kubernetes-sigs/kind/pull/104/files#diff-7033e7fa5be88fc307b556fb3888ef42R374 we should use `pullTo := path.Join(imagesDir, image)` now that we have this value, so we can ensure they stay consistent.

pkg/build/node/node.go
```diff
@@ -373,6 +379,12 @@ func (c *BuildContext) prePullImages(dir, containerID string) error {
 			movePulled = append(movePulled, fmt.Sprintf("/build/bits/images/%s", pullName))
 		}
 	}
+
+	// Create the /kind/images directory inside the container.
+	if err = execInBuild("mkdir", "-p", "/kind/images/"); err != nil {
```
BenTheElder: would you mind adding a TODO about pulling "/kind/images/" out to be in a constant?

docs/user-guide.md
```diff
@@ -0,0 +1,51 @@
...
+## Building the base and node images
+### Base image
+`kind` runs a local kubernetes cluster by using Docker containers as "nodes."
+`kind` uses the [`node-image`][node image] to run kubernetes artifacts, such 
+as `kubeadm` or `kubelet`.
+The `node-image` in turn is built off the [`base-image`][base image], which
+installs all the dependencies needed for Docker and Kubernetes to run in a
+container.
+
+To build the `base-image` we use the `build` command:
```
BenTheElder: Users generally shouldn't need to build the base image, as we've published a pre-built one, we should probably detail this in a more advanced guide and focus on something like:

---
# User Guide
// what kind is
## Creating a Cluster
// how to create a generic cluster
// note about using a custom image
## Building Images
// how to build the node image, note about the base image existing

## Advanced
### Building The Base Image
// how to build the base image, how to use it with the node image
### Using Config
// how to use a config file and e.g. patch the kubeadm config

----

That way we can focus on the easy and most relevant parts first and progressively get into the weeds on implementation details and advanced usage. Most users don't need to worry too much about the base image, just know that it exists.

BenTheElder:
Thanks so much for working on this! A couple of comments / questions. :smile: 
/ok-to-test

cmd/kind/build/build.go
neolit123: usually OK to omit long if short fits on a single line.

pkg/build/node/node.go
```diff
@@ -355,6 +355,12 @@ func (c *BuildContext) prePullImages(dir, containerID string) error {
 		return err
 	}
 
+	// Create "images" subdir.
+	imagesDir := path.Join(dir, "bits", "images")
+	if err := os.Mkdir(imagesDir, 0777); err != nil {
```
BenTheElder: We should use `MkdirAll` here, which won't fail when the directory already exists (it will for the other build modes)
```suggestion
	if err := os.MkdirAll(imagesDir, 0777); err != nil {
```

cmd/kind/build/build.go
alejandrox1: kind-pr103
 What do you think?

pkg/build/node/node.go
alejandrox1: Done

cmd/kind/build/build.go
BenTheElder: Looks good to me, thanks!

docs/README.md
```diff
@@ -1,7 +1,9 @@
 # `kind` Documentation
 
-For the primary design start with [design.md]
+For the primary design start with [design][design].
 
-<!--TODO(bentheelder): add usage docs, link to some here and in the root README-->
+For usage documentation see the [user guide][user guide].
```
BenTheElder: nit: let's put the shiny new user guide first :-)

alejandrox1:
@BenTheElder made the changes. Tell me what you think?
Thank you very much for your comments :smile:

pkg/build/node/node.go
```diff
@@ -365,15 +375,21 @@ func (c *BuildContext) prePullImages(dir, containerID string) error {
 			}
 			// TODO(bentheelder): generate a friendlier name
 			pullName := fmt.Sprintf("%d.tar", i)
-			pullTo := fmt.Sprintf("%s/bits/images/%s", dir, pullName)
+			pullTo := fmt.Sprintf("%s/%s", imagesDir, pullName)
```
BenTheElder: we should use `path.Join` because this is on the host so the separator may not be `/`, whereas the paths "on the node" will be linux paths. (in practice this code should generally be fine, but we should be consistent anyhow)
```suggestion
			pullTo := path.Join(imagesDir, pullName)
```

alejandrox1:
So, I moved the link to the user guide to the top of the README in the docs. Is that good enoguh for now or are there any other changes you'd like to see?

alejandrox1:
@BenTheElder , placed the ref to the user guide befor the design docs and linked to it from the README.
Tell me if that looks alright

docs/user-guide.md
BenTheElder: actually I'll rearrange a bit in a follow up

PR #106, opened 2018-11-14
More code cleanup

BenTheElder:
- add godoc to the new nodes package 
- use the Node type in more places

/assign @amwat

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -170,7 +170,7 @@ func (c *Context) provisionControlPlane(
 	if err := node.FixMounts(); err != nil {
 		// TODO(bentheelder): logging here
 		// TODO(bentheelder): add a flag to retain the broken nodes for debugging
-		nodes.Delete(node.String())
+		nodes.Delete(node)
 		return "", err
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/cluster/cluster.go
```diff
@@ -367,6 +367,6 @@ func stringSliceToByteSliceSlice(ss []string) [][]byte {
 }
 
 // ListNodes returns the list of container IDs for the "nodes" in the cluster
-func (c *Context) ListNodes() (containerIDs []string, err error) {
+func (c *Context) ListNodes() ([]*nodes.Node, error) {
```
munnerz: (nit) is it useful/does it make sense for this to be a slice of pointers, as opposed to `[]nodes.Node`?

munnerz:
Looks good to me!

/lgtm
/hold

pkg/cluster/cluster.go
BenTheElder: This plays nicer with Delete(), so I'm on the fence about what to do here.

munnerz: SGTM. I don't think it's a big issue really 😄 

Feel free to `/hold cancel`!

BenTheElder: This actually annoys me but I haven't thought of a better solution yet so I'll just merge this for now 😅

PR #108, opened 2018-11-14
clean up pkg/cluster/config

alejandrox1:
/kind cleanup

### DIFF
--- a/pkg/cluster/config/default.go
+++ b/pkg/cluster/config/default.go
@@ -20,24 +20,27 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
-// DefaultImage is the default for the Config.Image field, aka the default node image
+// DefaultImage is the default for the Config.Image field, aka the default node
+// image.
 const DefaultImage = "kindest/node:v1.12.2@sha256:6ac1dc1750fc0efd13d4e294115f9012a21282957e4380a5535bd32154193d4d"
 
-// New returns a new default Config
 func addDefaultingFuncs(scheme *runtime.Scheme) error {
 	return RegisterDefaults(scheme)
 }
 
-func SetDefaults_Config(obj *Config) {
+// SetDefaultsConfig sets uninitialized fields to their default value.
+func SetDefaultsConfig(obj *Config) {
 	if obj.Image == "" {
 		obj.Image = DefaultImage
 	}
 }
 
+// ApplyDefaults sets the defaults to a Config.
 func (c *Config) ApplyDefaults() {
-	SetDefaults_Config(c)
+	SetDefaultsConfig(c)
 }
 
+// New returns a new default Config.
 func New() *Config {
 	cfg := &Config{}
 	cfg.ApplyDefaults()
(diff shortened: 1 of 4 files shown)


pkg/cluster/config/register.go
```diff
@@ -21,18 +21,21 @@ import (
...
 
 var (
-	// SchemeGroupVersion is group version used to register these objects
+	// SchemeGroupVersion is group version used to register these objects.
 	SchemeGroupVersion = schema.GroupVersion{Group: GroupName, Version: "v1alpha1"}
+	// SchemeBuilder is a type to collect functions that add things to a schema.
 	SchemeBuilder      runtime.SchemeBuilder
 	localSchemeBuilder = &SchemeBuilder
-	AddToScheme        = localSchemeBuilder.AddToScheme
+	// AddToScheme applies all the sotred functions in the localSchemeBuilder
```
munnerz: `s/sotred/sorted`

pkg/cluster/config/register.go
```diff
@@ -21,18 +21,21 @@ import (
...
 
-// GroupName is the group name used in this package
+// GroupName is the group name used in this package.
 const GroupName = "kind.sigs.k8s.io"
 
 var (
-	// SchemeGroupVersion is group version used to register these objects
+	// SchemeGroupVersion is group version used to register these objects.
 	SchemeGroupVersion = schema.GroupVersion{Group: GroupName, Version: "v1alpha1"}
+	// SchemeBuilder is a type to collect functions that add things to a schema.
```
munnerz: `s/things/API types`? 'things' feels a bit wishy-washy 😄

pkg/cluster/config/default.go
```diff
@@ -20,24 +20,27 @@ import (
...
+// image.
 const DefaultImage = "kindest/node:v1.12.2@sha256:6ac1dc1750fc0efd13d4e294115f9012a21282957e4380a5535bd32154193d4d"
 
-// New returns a new default Config
 func addDefaultingFuncs(scheme *runtime.Scheme) error {
 	return RegisterDefaults(scheme)
 }
 
-func SetDefaults_Config(obj *Config) {
+// SetDefaultsConfig sets uninitializex fields to their default value.
```
munnerz: `s/uninitializex/uninitialized`

pkg/cluster/config/types.go
```diff
@@ -42,6 +42,8 @@ type Config struct {
 	ControlPlane *ControlPlane `json:"ControlPlane,omitempty"`
 }
 
+// ControlPlane holds node life cycle hooks to make the actual state of all
```
BenTheElder: Hmm I'd say something like above instead, ControlPlane holds configuration specific to the "control plane" node(s) (currently the only node)

pkg/cluster/config/register.go
alejandrox1: haha it was actually supposed to be "stored."

pkg/cluster/config/types.go
alejandrox1: change made

PR #114, opened 2018-11-15
add `kind get clusters`

BenTheElder:
so you can list the clusters that appear to be running currently

next up: use this new underlying functionality to check if a given cluster exists before trying to create or delete one

/kind enhancement
/assign @amwat


pkg/cluster/context.go
```diff
@@ -63,9 +63,17 @@ func NewContext(name string) (ctx *Context, err error) {
 			name, validNameRE.String(),
 		)
 	}
+	return newContextNoValidation(name), nil
```
BenTheElder: actually New() probably should not validate or error and there should be a separate validation call when creating things, I'll do this in some follow-up

BenTheElder: added a TODO

PR #116, opened 2018-11-16
cleanup: make nodes implement exec.Cmder

BenTheElder:
... and move all execution on nodes to use this instead

this unifies all of the node exec code into one place behind a cleanish interface which should make logging and other refactors much easier, along with testing other packages eventually

/kind cleanup
/assign @amwat


BenTheElder:
updated to move a lot more docker related logic into the docker package, include the new exec implementation. nodes now forward to a `docker.ContainerCmder(node.nameOrID)` for `node.Command(...)` etc., so the bulk of the `docker exec` logic is now properly in the docker package and generically implemented.

This further minimizes the places we actually need to update how exec works, while letting lots of places call commands.

I've also made a quick update to doc/todo.md to capture some of the improvements and newly planned work

PR #125, opened 2018-11-21
WIP: Feature userns host

### LINKED ISSUE #124: Kind is not able to start containers when userns-remap is enabled on the Docker daemon
Kind is not able to start containers when `userns-remap` is enabled on the Docker daemon. The error that occurs is the following:
```
time="13:37:34" level=error msg="docker: Error response from daemon: privileged mode is incompatible with user namespaces.  You must run the container in the host namespace when running privileged mode."
```

[Here](https://docs.docker.com/engine/security/userns-remap/#user-namespace-known-limitations) it  says that this is the correct behaviour.  

A solution could be that we add the flag `--userns=host` in the `docker run` commands that start the containers and contain the flag `--privileged`.

vasrem:
Fixes #124 

TODO:
- Identify if `userns-remap` is enabled and change [this](https://github.com/kubernetes-sigs/kind/blob/9a6f91042e59104390dd50a942b6cd6ff5ad5ac7/pkg/cluster/nodes/node.go#L167) line accordingly 

If you find this feature good, I would like some guidance regarding the implementation of the userns-remap check.


BenTheElder:
/ok-to-test
Thanks for working on this!

vasrem:
The structure doesn't seem perfect. If you have any suggestions, I would be happy to hear them. 

The `FixMounts` function should be deleted.

pkg/build/node/node.go
```diff
@@ -415,6 +415,29 @@ func (c *BuildContext) prePullImages(dir, containerID string) error {
 		log.Errorf("Image build Failed! %v", err)
 		return err
 	}
+
+	// systemd-in-a-container should have read only /sys
+	// https://www.freedesktop.org/wiki/Software/systemd/ContainerInterface/
+	// however, we need other things from `docker run --privileged` ...
+	// and this flag also happens to make /sys rw, amongst other things
+	// Need --privileged flag in docker exec
+	if err = execInBuild("mount", "-o", "remount,ro", "/sys"); err != nil {
```
BenTheElder: why is this happening at build time?? mounts are created when the container is created.

pkg/build/node/node.go
```diff
@@ -415,6 +415,29 @@ func (c *BuildContext) prePullImages(dir, containerID string) error {
...
+		log.Warningf("Failed to mount /sys: %v", err)
+		return err
+	}
+	// kubernetes needs shared mount propagation
+	// Need --privileged flag in docker exec
+	if err = execInBuild("mount", "--make-shared", "/"); err != nil {
```
BenTheElder: again, this needs to happen on the actual node containers for mount propogation, which is something kubernetes uses. this can't be at build time unless docker added some new magic I don't know about :confused: 

same for the other mount below

pkg/build/node/node.go
```diff
@@ -427,6 +450,7 @@ func (c *BuildContext) createBuildContainer(buildDir string) (id string, err err
 		[]string{
 			"-d", // make the client exit while the container continues to run
 			// label the container to make them easier to track
+			"--privileged",
```
BenTheElder: the node build should not need to be priviliged?

BenTheElder:
I think instead we will probably need to detect userns remapping and attempt to deploy workarounds, including potentially skipping mount fixes, that may degrade functionality but allow the cluster to come up, perhaps printing a warning about this.

I don't have particularly convenient access to a setup with this enabled so it might be a bit before I can look at it myself, but I think it would be fairly easy for cluster creation to detect this and handle it.

pkg/docker/userns_remap.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"strings"
+
+	"sigs.k8s.io/kind/pkg/exec"
+)
+
+// UsernsRemap checks if userns-remap is enabled in dockerd
+func UsernsRemap() bool {
+	cmd := exec.Command("cat", "/etc/docker/daemon.json")
```
BenTheElder: I think we probably want to detect this from the daemon with something like `docker info`, I've not yet found the way to do this, but I suspect we can do something less brittle.

vasrem: I guess the only way to identify this via `docker info` is to check if the `Docker Root Dir` has in the end of the path something in the format of `UID.GID`. For example `/var/lib/docker/1000.1000`. 

Do you find this more safe?

vasrem: Aha! It seems that I found something in `docker info --format '{{json .SecurityOptions}}'`. Refactoring.

pkg/docker/userns_remap.go
```diff
@@ -0,0 +1,38 @@
...
+
+// UsernsRemap checks if userns-remap is enabled in dockerd
+func UsernsRemap() bool {
+	cmd := exec.Command("docker", "info", "--format", "'{{json .SecurityOptions}}'")
+	lines, err := exec.CombinedOutputLines(cmd)
+	if err != nil {
+		return false
+	}
+	if len(lines) > 0 {
+		if strings.Contains(lines[0], "name=userns") {
```
BenTheElder: Note to self: not worth it right now but we should eventually collect this structure early in the CLI and parse out the fields for other info as well.

BenTheElder:
(updated title to remove the merge blocking label)

Issue #130, opened 2018-11-23
Improve config for multi node scenario


fabriziopandini:
as per https://github.com/kubernetes-sigs/kind/issues/117#issuecomment-440531457, one of the step for implementing multi node support is to improve Kind config in order to support the concept of topology: how many control-plane nodes should be in the cluster? how many worker nodes? should I add an external-load balancer? add an external-etcd?

A first high level idea for the new config design is something in the middle between current Kind config and cluster api. This imply the config will be composed by a combination of following objects:
- a `Cluster` object for "cluster wide settings" 
- a `Node`object (or Container/Machine to avoid confusion with Kubernets `Node`object) for "node specific settings"
- eventually a `NodeSet` (or ContainerSet/MachineSet) object for defining groups of machines starting from a template

The new config api should be ready for supporting following use cases:
- different CRI/different node-image (and mixed clusters)
- version skew (between control-plane and kubelet or between nodes)
- volume mounts
- kubeadm config patching (should apply only to the bootstrap control-plane node?)
- LifecycleHooks (in which forms, considering different types of nodes?)
- Adding/Removing nodes to the cluster
- ...

Opinions about the proposed api design? 
Preferences on object names? 
Are there other use cases to be supported (in the first release)? Are there use cases not necessary (in the first release)?

/assign

PR #137, opened 2018-11-26
Split internal/current config from v1alpha1 and start using k8s.Scheme 

fabriziopandini:
This PR sets the ground for addressing https://github.com/kubernetes-sigs/kind/issues/130 

Preparation works include the refactor of the current kind config by splitting internal/current from v1alpha1 API versions and switching to k8s.Scheme for managing encoding, conversions, and defaulting.

Note for reviewers
1. Sorry for the size of the PR, but API changes are difficult to break down into separated PRs.
2. I'm still learning kind codebase, so might be I didn't catch some internals of the current config; please let me know :sweat_smile:


pkg/cluster/config/encoding/encoding.go
```diff
@@ -1,146 +0,0 @@
...
-	case config.SchemeGroupVersion.String():
-		// this allows strict unmarshaling with the version header
-		layout := &struct {
-			configHeader
-			config.Config
-		}{}
-		cfg = &layout.Config
-		rawCfg = layout
-	}
-	err = yaml.UnmarshalStrict(raw, rawCfg)
```
BenTheElder: this does unfortunately mean we lose strict unmarshalling (See the config issue for dicussion) but it is currently par for the course in kubernetes tooling.

pkg/cluster/config/types.go
```diff
@@ -17,63 +17,61 @@ limitations under the License.
...
 	// KubeadmConfigPatches are applied to the generated kubeadm config as
 	// strategic merge patches to `kustomize build` internally
 	// https://github.com/kubernetes/community/blob/master/contributors/devel/strategic-merge-patch.md
 	// This should be an inline yaml blob-string
-	KubeadmConfigPatches []string `json:"kubeadmConfigPatches,omitempty"`
+	KubeadmConfigPatches []string
 	// KubeadmConfigPatchesJSON6902 are applied to the generated kubeadm config
 	// as patchesJson6902 to `kustomize build`
-	KubeadmConfigPatchesJSON6902 []kustomize.PatchJSON6902 `json:"kubeadmConfigPatchesJson6902,omitempty"`
+	KubeadmConfigPatchesJSON6902 []kustomize.PatchJSON6902
```
BenTheElder: I'm not the most familiar with the k8s API machinery internals, but what's the intention with removing omitempty?

munnerz: This is on the internal API type, which never gets encoded/decoded from external (JSON) representations, so it's okay here 😄

pkg/cluster/config/v1alpha1/doc.go
```diff
@@ -14,20 +14,9 @@ See the License for the specific language governing permissions and
...
-}
-
-// APIVersion returns the `apiVersion:` for Config
-func (c *Config) APIVersion() string {
-	return SchemeGroupVersion.String()
-}
+// Package v1alpha1 implements the v1alpha1 apiVersion of the `kind` Config
+//
+// +k8s:deepcopy-gen=package
+// +k8s:conversion-gen=sigs.k8s.io/kind/pkg/cluster/config/
```
munnerz: (not necessarily for this PR): perhaps we should move this to `sigs.k8s.io/kind/pkg/apis/kind[/v1alpha1]` for consistency with other k8s api packages?

pkg/cluster/config/types.go
BenTheElder: it doesn't _currently_ but I don't see any external type? at some point I was going to save the config to nodes and include it in the log dumps for debugging.

pkg/cluster/config/default.go
```diff
@@ -28,21 +28,9 @@ func addDefaultingFuncs(scheme *runtime.Scheme) error {
 	return RegisterDefaults(scheme)
 }
 
-// SetDefaultsConfig sets uninitialized fields to their default value.
-func SetDefaultsConfig(obj *Config) {
+// SetDefaults_Config sets uninitialized fields to their default value.
+func SetDefaults_Config(obj *Config) {
```
munnerz: Typically defaults are applied to *external* API versions before converting to the internal representation. This allows for default values to be versioned as well as fields etc.

Applying defaults to the internal representation will mean all API versions will always have to have the same default values.

pkg/cluster/config/encoding/scheme.go
```diff
@@ -0,0 +1,93 @@
...
+	var config = &config.Config{}
+
+	// decode data into a config object
+	_, _, err := Codecs.UniversalDecoder().Decode(data, nil, config)
+	if err != nil {
+		return nil, errors.Wrap(err, "decoding failure")
+	}
+
+	// apply defaults
+	Scheme.Default(config)
```
munnerz: So in other Kubernetes projects, typically the application itself works with a particular API version.

For example, kube-controller-manager works with the versioned representation of types.

From my understanding, the universal decoder should be able to just directly decode into the versioned representation.

pkg/cluster/config/types.go
BenTheElder: :man_facepalming: I see those now. this is what I get for trying to rush through a first pass over my PR backlog

pkg/cluster/config/encoding/encoding.go
neolit123: strict unmarshal warnings (or errors?) would be pretty easy to add after this PR is merged as a pre-step before Decode.

pkg/cluster/config/v1alpha1/doc.go
neolit123: would `sigs.k8s.io/kind/pkg/apis/kind[/v1alpha1]` imply `conversion-gen-external-types`?

e.g. how it's done for the kubelet:
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/config/v1beta1/doc.go#L18-L19

pkg/cluster/config/encoding/scheme.go
neolit123: this is how it's done in kubeadm at least.

pkg/cluster/config/encoding/scheme_test.go
```diff
@@ -0,0 +1,83 @@
...
+		{
+			Name:        "invalid yaml",
+			Path:        "./testdata/invalid-yaml.yaml",
+			ExpectError: true,
+		},
+	}
+	for _, tc := range cases {
+		_, err := Load(tc.Path)
+		if err != nil && !tc.ExpectError {
+			t.Errorf("case: '%s' got error`Load`ing and expected none: %v", tc.Name, err)
```
neolit123: are the `Load` in quotes a search and replace typo?

pkg/cluster/config/encoding/scheme_test.go
```diff
@@ -0,0 +1,83 @@
...
+			t.Errorf("case: '%s' got no error `Load`ing but expected one", tc.Name)
+		}
+	}
+}
+
+func TestLoadDefault(t *testing.T) {
+	cfg, err := Load("")
+	if err != nil {
+		t.Errorf("got error `Load`ing default config but expected none: %v", err)
+		t.FailNow()
```
neolit123: nit: or just `t.Fatalf()`

pkg/cluster/config/encoding/encoding.go
neolit123: had a look at how to create a custom decoder too. seems doable in the near future.

pkg/cluster/config/v1alpha1/doc.go
fabriziopandini: @munnerz you are right, this is not necessary now, but this will be used as soon as v1alpha2 will be added, so if you don't mind I will leave it

pkg/cluster/config/encoding/scheme.go
fabriziopandini: @munnerz @neolit123 
I'm proposing to apply defaults to the internal types because this IMO provides consistent defaulting across different API versions and makes simpler ussge of both defaults (no need of conversion between public API/internal api) and decoding (yaml->internal->default Vs yaml->public api->default->internal).

Nevertheless, no problem for me to change to a more canonical approach. Just let me known

fabriziopandini: Now that I think to this again there is value in using the "canonical" approach (e.g. avoid confusion in people reading the code).
So I'm moving defaulting back to the versioned config

BenTheElder: I think the reason to do defaulting in the versioned config is also to opt into versioned behavior, which would be nice to preserve in the future. In the alpha state, we can and will break anything and everything between commits so I'm not too worried about this

pkg/cluster/config/default.go
fabriziopandini: Now defaults are on the external API

pkg/cluster/config/encoding/encoding.go
fabriziopandini: @BenTheElder, considering above comments it seems to me that this can be acceptable for now right?

pkg/cluster/config/encoding/scheme.go
fabriziopandini: default is in versioned api now so we are already on the right track

fabriziopandini:
Rebase + squash commits
Almost all the comments are addressed, few last comments waiting for feedback

pkg/cluster/config/encoding/encoding.go
neolit123: > this can be acceptable for now right

yes,
it's just that i recommend that we enable strict unmarshal with the first kind release.

since we cannot merge a strict decoded in api-machinery in this release, we should probably go for the way kubeadm is doing it. calling `yaml.UnmarshalStrict` before Decode.

BenTheElder: yes, sorry, I wanted to call out that we had this functionality change, but it makes sense.

we can sort out strict unmarshal again later, I'm ok with that not being in the first "release" since the first release will be very alpha grade and not stable. I'm going to make this clear with the versioning of the APIs and kind itself. 

we should get strict unmarshal before we make compatibility guarantees between versions so it isn't a breaking change later.

pkg/cluster/config/v1alpha1/doc.go
BenTheElder: +1, let's do this in a follow up.

I put it under pkg/cluster/config in an attempt to signal the scope of the config object, let's consider that as we improve the config / API structure, I'd like to leave room to grow in new and unexpected ways ... other container runtimes, possibly CLI tools ... :-)

PR #138, opened 2018-11-27
update kustomize, drop lots of transitive dependencies

BenTheElder:
updates to kustomize to [ef51cceff55b17542fc91d6c4e30ea0bbb9641f7](https://github.com/kubernetes-sigs/kustomize/commit/ef51cceff55b17542fc91d6c4e30ea0bbb9641f7), which drops lots of dependencies (see: https://github.com/kubernetes-sigs/kustomize/issues/559), including ~63k loc / ~ 3MB of `vendor/`

thanks @monopole!

FYI @neolit123


BenTheElder:
/hold
```
I1127 00:07:19.914] time="00:07:19" level=fatal msg="Failed to create cluster: failed to create kubeadm config: loadResMapFromBasesAndResources: rawResources failed to read Resources: Missing metadata.name in object {map[apiServerExtraVolumes:[map[writeable:false hostPath:/etc/nsswitch.conf mountPath:/etc/nsswitch.conf name:nsswitch pathType:FileOrCreate]] apiVersion:kubeadm.k8s.io/v1alpha2 clusterName:kind-1 kind:MasterConfiguration kubernetesVersion:v1.11.6-beta.0.1+5933b9771b71c2 api:map[bindPort:36555] apiServerCertSANs:[localhost]]}\n\n"
```

oof, this may be a problem. cc @munnerz - we have a single config object so of course it does not have a name meta. might be OK after #137

neolit123:
>  might be OK after #137

yes, possibly.

BenTheElder:
rethinking, this is not related to #137, possibly we need to add resource names to the kubeadm config and then make the patches implicitly target these.

that's sorta hacky, @neolit123 do you know if kubeadm supports specifying a `name` field in the config just for this sort of purpose?

neolit123:
> that's sorta hacky, @neolit123 do you know if kubeadm supports specifying a name field in the config just for this sort of purpose?

yes, this seems like something kustomize is complaining about.
so the kubeadm v1alpha2 config should support something like:
```
kind:MasterConfiguration
metadata:
  name: some-name
```

btw, i do have some scepticismthat kustomize will be able to handle full blown -> v1beta1 patching.
did you test that on a full v1beta1 config?
possibly some unit tests will be needed too.

BenTheElder:
> btw, i do have some scepticismthat kustomize will be able to handle full blown -> v1beta1 patching.
did you test that on a full v1beta1 config?

hmm? It can patch all of the objects. do you mean patching across versions?

> possibly some unit tests will be needed too.

yes, but those will be more maintainable once we have a stable config upstream, and in the meantime most of it is tested end-to-end, which is not really surprising given what kind's use-case is :^)

neolit123:
> do you mean patching across versions?

yes.
say if the the user wants to use apply settings that make sense only for the kubeadm v1beta1 (e.g. new ones). how does that work?

BenTheElder:
they always have to patch the right GVK. since it's only intended for advanced users, and kubeadm will eventually get GA config, this seems OK for now.

BenTheElder:
see EG here:
https://github.com/kubernetes-sigs/kind/blob/0908a44305862982013a6484c544784adefbb7bb/docs/user/kind-example-config.yaml#L6-L21
GVK are specified

neolit123:
did you try the metadata "fix"?
```
metadata:
  name: some-name
```

i don't understand why customize is complaining about that, but we sure don't require the field in regular parsing.

BenTheElder:
I haven't yet, lots going on with kubecon coming etc., but it needs the name because it internally matches over which objects the patches should apply to, so it wants to target GVK + name from the patch to GVK + name to the object to be patched.

Possibly we should magically specify {G,V,K,name} for the user based on the one we generated, so they only need to write the actual patch... :thinking: 

All of this will be cleaner when we target a single config version of course :-)

neolit123:
my guess is that we want this PR for 1.0?

BenTheElder:
yes, definitely. I will update this to the latest, probably we will need to
name the generated objects and ensure patches target the name :thinking:

BenTheElder:
this is outdated due to dep -> go modules

I'm exploring options, either we need to require users set a `metadata:\nname: ` entry matching our generated objects, or we need to figure out how to auto add this to the yaml patches.

For JSON patches the target metadata is totally out of band to the patch, we can already auto inject the target name.

BenTheElder:
xref: https://github.com/kubernetes-sigs/kustomize/issues/627

PR #139, opened 2018-11-28
clean up context, flag structs, and runE

BenTheElder:
some general code base cleanup:
- all cobra commands should use runE (and return an error) instead of just `run`. this will be helpful later for testing etc.
- all flags structs are `flagpole` (where else would you put flags :stuck_out_tongue:), so as not to have `flags := &flags{}` and other slightly confusing code, this was previously changed in a few places but not all
- cluster.Context validation now only occurs during Create, this allows eg `kind delete`ing a cluster with a previously valid but now invalid context name, or at least attempting to, should we change the validation. it will not allow creating new things with currently invalid names
- the helper functions for `cluster.Context.Create` now use a wrapper type with some extra fields, including the recently added `retain` field, which was moved to there. Context no longer takes a `retain`

### DIFF
--- a/cmd/kind/build/baseimage/baseimage.go
+++ b/cmd/kind/build/baseimage/baseimage.go
@@ -17,7 +17,8 @@ limitations under the License.
 package baseimage
 
 import (
-	log "github.com/sirupsen/logrus"
+	"fmt"
+
 	"github.com/spf13/cobra"
 
 	"sigs.k8s.io/kind/pkg/build/base"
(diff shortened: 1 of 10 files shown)


cmd/kind/build/baseimage/baseimage.go
```diff
@@ -23,14 +23,14 @@ import (
 	"sigs.k8s.io/kind/pkg/build/base"
 )
 
-type flags struct {
+type flagpole struct {
```
amwat: /shrug

BenTheElder:
rebased

PR #141, opened 2018-11-28
Add flag to Wait for node to reach ready state

### LINKED ISSUE #42: wait for CNI to be ready
or more generally, the e2e tests need the node to be in the ready state, we should either block `kind create` on this, or give a separate command to block on readiness.

I think probably making `kind create` block on this with some flag to toggle it is the way to go.

/kind bug
/priority important-soon
/assign

alejandrox1:
Add --wait flag which will make kind block until the master node of the
cluster reaches a ready state.

Resolves #42

/kind feature

### DIFF
--- a/cmd/kind/create/cluster/createcluster.go
+++ b/cmd/kind/create/cluster/createcluster.go
@@ -19,6 +19,7 @@ package cluster
 
 import (
 	"fmt"
+	"time"
 
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -33,6 +34,7 @@ type flagpole struct {
 	Config    string
 	ImageName string
 	Retain    bool
+	Wait      time.Duration
 }
 
 // NewCommand returns a new cobra.Command for cluster creation
@@ -50,6 +52,7 @@ func NewCommand() *cobra.Command {
 	cmd.Flags().StringVar(&flags.Config, "config", "", "path to a kind config file")
 	cmd.Flags().StringVar(&flags.ImageName, "image", "", "node docker image to use for booting the cluster")
 	cmd.Flags().BoolVar(&flags.Retain, "retain", false, "retain nodes for debugging when cluster creation fails")
+	cmd.Flags().DurationVar(&flags.Wait, "wait", time.Duration(0), "Wait for control plane node to be ready (default 0s)")
 	return cmd
 }
 
@@ -81,7 +84,7 @@ func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
 			return fmt.Errorf("aborting due to invalid configuration")
 		}
 	}
-	if err = ctx.Create(cfg, flags.Retain); err != nil {
+	if err = ctx.Create(cfg, flags.Retain, flags.Wait); err != nil {
 		return fmt.Errorf("failed to create cluster: %v", err)
 	}
 
--- a/pkg/cluster/context.go
+++ b/pkg/cluster/context.go
@@ -48,9 +48,10 @@ type Context struct {
 // createContext is a superset of Context used by helpers for Context.Create()
 type createContext struct {
 	*Context
-	status *logutil.Status
-	config *config.Config
-	retain bool // if we should retain nodes after failing to create
+	status       *logutil.Status
+	config       *config.Config
+	retain       bool          // if we should retain nodes after failing to create.
+	waitForReady time.Duration // Wait for the control plane node to be ready.
 }
 
 // similar to valid docker container names, but since we will prefix
@@ -121,16 +122,17 @@ func (c *Context) KubeConfigPath() string {
 }
 
 // Create provisions and starts a kubernetes-in-docker cluster
-func (c *Context) Create(cfg *config.Config, retain bool) error {
+func (c *Context) Create(cfg *config.Config, retain bool, wait time.Duration) error {
 	// validate config first
 	if err := cfg.Validate(); err != nil {
 		return err
 	}
 
 	cc := &createContext{
-		Context: c,
-		config:  cfg,
-		retain:  retain,
+		Context:      c,
+		config:       cfg,
+		retain:       retain,
+		waitForReady: wait,
 	}
 
 	fmt.Printf("Creating cluster '%s' ...\n", c.ClusterName())
@@ -346,6 +348,14 @@ func (cc *createContext) provisionControlPlane(
 			if err := runHook(node, &hook, "postSetup"); err != nil {
 				return kubeadmConfig, err
 			}
+		}
+	}
+
+	// Wait for the control plane node to reach Ready status.
+	isReady := nodes.WaitForReady(node, time.Now().Add(cc.waitForReady))
+	if cc.waitForReady > 0 {
+		if !isReady {
+			log.Warn("timed out waiting for control plane to be ready")
 		}
 	}
 
--- a/pkg/cluster/nodes/nodes.go
+++ b/pkg/cluster/nodes/nodes.go
@@ -19,6 +19,7 @@ package nodes
 import (
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	"sigs.k8s.io/kind/pkg/cluster/consts"
@@ -98,4 +99,38 @@ func list(visit func(string, *Node), filters ...string) error {
 		visit(cluster, FromID(names[0]))
 	}
 	return nil
+}
+
+// WaitForReady uses kubectl inside the "node" container to check if the
+// control plane nodes are "Ready".
+func WaitForReady(node *Node, until time.Time) bool {
+	return tryUntil(until, func() bool {
+		cmd := node.Command(
+			"kubectl",
+			"--kubeconfig=/etc/kubernetes/admin.conf",
+			"get",
+			"nodes",
+			"--selector=node-role.kubernetes.io/master",
+			// When the node reaches status ready, the status field will be set
+			// to true.
+			"-o=jsonpath='{.items..status.conditions[-1:].status}'",
+		)
+		lines, err := exec.CombinedOutputLines(cmd)
+		if err != nil {
+			return false
+		}
+
+		// 'lines' will return the status of all nodes labeled as master. For
+		// example, if we have three control plane nodes, and all are ready,
+		// then the status will have the following format: `True True True'.
+		status := strings.Fields(lines[0])
+		for _, s := range status {
+			// Check node status. If node is ready then this wil be 'True',
+			// 'False' or 'Unkown' otherwise.
+			if !strings.Contains(s, "True") {
+				return false
+			}
+		}
+		return true
+	})
 }
--- a/docs/user/user-guide.md
+++ b/docs/user/user-guide.md
@@ -35,6 +35,11 @@ For a sample kind configuration file see
 To use a kind configuration file use the `--config` flag and pass the path to
 the file.
 
+If you want the `create cluster` command to block until the control plane
+reaches a ready status, you can use the `--wait` flag and specify a timeout.
+To use `--wait` you must specify the units of the time to wait. For example, to
+wait for 30 seconds, do `--wait 30s`, for 5 minutes do `--wait 5m`, etc.
+
 
 ## Building Images
 


pkg/cluster/nodes/node.go
```diff
@@ -224,3 +225,33 @@ func (n *Node) WriteKubeConfig(dest string) error {
...
+			return false
+		}
+		if len(lines) <= 1 {
+			return false
+		}
+
+		for _, line := range lines {
+			// We assume the control plane node contains the "control-plane"
+			// substring.
+			if strings.Contains(line, "control-plane") {
```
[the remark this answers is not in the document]
alejandrox1: I like that idea. How about `isConntrolPlane bool` ?

pkg/cluster/nodes/node.go
```diff
@@ -224,3 +225,33 @@ func (n *Node) WriteKubeConfig(dest string) error {
...
+				// Each line will have the following fields:
+				// NAME STATUS ROLES AGE VERSION
+				fields := strings.Fields(line)
+				if fields[1] == "Ready" {
```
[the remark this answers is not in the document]
alejandrox1: Completely forgot that was a thing. Thanks

pkg/cluster/nodes/node.go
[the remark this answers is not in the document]
BenTheElder: so we're going to do this with the multi node work: https://github.com/kubernetes-sigs/kind/issues/130 regarding having different kinds, will think about the short term fix for a bit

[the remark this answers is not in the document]
BenTheElder: we still need a way to detect this from a running node, for consistency. everything else we use / know about a node is either:
- explicitly passed through in create
- derivable at runtime

I think instead Create() can track the name(s) or `*Node` handles of control-plane node(s) and wait for these at the end of create

pkg/cluster/nodes/node.go
```diff
@@ -224,3 +225,33 @@ func (n *Node) WriteKubeConfig(dest string) error {
 
 	return ioutil.WriteFile(dest, buff.Bytes(), 0600)
 }
+
+// WaitForControlPlane uses kubectl inside the "node" container to check if
+// the control plane node status is "Ready". The control plane node is found by
+// looking for the node containing the substring "control-plane" in its name.
+func (n *Node) WaitForControlPlane(until time.Time) bool {
```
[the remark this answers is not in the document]
BenTheElder: we can use a selector instead, kubeadm applies the label `node-role.kubernetes.io/master: ""`

pkg/cluster/nodes/node.go
```diff
@@ -224,3 +225,33 @@ func (n *Node) WriteKubeConfig(dest string) error {
...
+	return tryUntil(until, func() bool {
+		cmd := n.Command("kubectl", "--kubeconfig=/etc/kubernetes/admin.conf", "get", "nodes")
```
BenTheElder: adding `"--selector=node-role.kubernetes.io/master"` should eliminate the need to filter below

pkg/cluster/context.go
```diff
@@ -349,6 +351,17 @@ func (cc *createContext) provisionControlPlane(
 		}
 	}
 
+	// Wait for control plane node to reach Ready status.
+	if cc.waitForControlPlane {
+		if !node.WaitForControlPlane(time.Now().Add(time.Second * 30)) {
+			// TODO: logging here
+			if !cc.retain {
+				nodes.Delete(*node)
```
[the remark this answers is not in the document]
BenTheElder: I don't think we should delete here at all, your node not being ready doesn't mean the cluster is broken if you choose a short timeout. This wait is not critical.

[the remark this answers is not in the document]
BenTheElder: imo, this timeout failing should default to just a warning, it should not return failure or delete nodes. We're just waiting a bit to see if we can let the cluster be warmed up before beginning whatever usage of it we had planned.

Tests and users still need to handle a broken cluster even if this passes, but if it times out, it does not mean their tests will definitely fail.

pkg/cluster/context.go
```diff
@@ -349,6 +351,17 @@ func (cc *createContext) provisionControlPlane(
...
+			}
+			return "", fmt.Errorf("timed out waiting for control plane to be ready")
```
BenTheElder: it should not return "", because the kubeadmConfig still needs to be cleaned up afterwards.

but also, let's just log a warning here.

cmd/kind/create/cluster/createcluster.go
```diff
@@ -50,6 +51,7 @@ func NewCommand() *cobra.Command {
 	cmd.Flags().StringVar(&flags.Config, "config", "", "path to a kind config file")
 	cmd.Flags().StringVar(&flags.ImageName, "image", "", "node docker image to use for booting the cluster")
 	cmd.Flags().BoolVar(&flags.Retain, "retain", false, "retain nodes for debugging when cluster creation fails")
+	cmd.Flags().BoolVar(&flags.Wait, "wait", false, "Wait for master node to be ready")
```
BenTheElder: "wait for the control-plane nodes to be ready", since while we only ever have one node today, this will still be true and we will continue to use this flag for multi-node

pkg/cluster/context.go
```diff
@@ -349,6 +351,17 @@ func (cc *createContext) provisionControlPlane(
 		}
 	}
 
+	// Wait for control plane node to reach Ready status.
+	if cc.waitForControlPlane {
+		if !node.WaitForControlPlane(time.Now().Add(time.Second * 30)) {
```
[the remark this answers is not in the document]
BenTheElder: yes, let's make this configurable

cmd/kind/create/cluster/createcluster.go
BenTheElder: let's make this a time.Duration string as well, and then the default can be "0" for don't wait, that way the user can decide how long they want to wait before resuming whatever the rest of their test rig wants to do

pkg/cluster/context.go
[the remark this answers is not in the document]
neolit123: kubeadm has an initial timeout of 40sec for the kubelet to ready up post init and  in parallel it has a healthz check for the api server that runs for 5 minutes (config via TimeoutForControlPlane in v1beta1).
https://godoc.org/k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm#APIServer

pkg/cluster/context.go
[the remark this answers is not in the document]
neolit123: +1 the node can remain there even if no control plane is up on it.
also we might want to collect logs from pods before deleting.

pkg/cluster/nodes/node.go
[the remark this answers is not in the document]
neolit123: with regard to a kubectl based check, i think the ready state on the node object is a safer bet, because the taints and labels are a `kubeadm init phase` that can be omitted using `kubeadm init --skip-phases=mark-control-plane` in 1.13.

on the other hand, i'm pretty sure i've seen a case where a node can be ready while some pods are failing, so from my experience watching for an expected list of pods is the only truth of a healthy node.

alejandrox1:
Thank you for the comments. Changes have been made.
Tell me what yall think about it. Here I went for formatting kubectl's output since the call to the function is made inside `ProvisionControlPlane` which makes available the node's name.
I feel like using the selector `"node-role.kubernetes.io/master"` would be better for a method that waits for all master nodes (ha? cluster).

BenTheElder:
> Tell me what yall think about it. Here I went for formatting kubectl's output since the call to the function is made inside ProvisionControlPlane which makes available the node's name.

I don't think we should pattern match on node names.

> I feel like using the selector "node-role.kubernetes.io/master" would be better for a method that waits for all master nodes (ha? cluster).

Well that's roughly equivalent to listing all nodes then matching the names, just using an actual identifier instead. Waiting for all of them to be ready is the same as waiting on one control plane to be ready when we have a single control plane node.

It also means we can use the same flag and same code when we do get multi-node.

alejandrox1:
@BenTheElder that makes sense, thank you.
And change made

pkg/cluster/nodes/node.go
```diff
@@ -224,3 +224,35 @@ func (n *Node) WriteKubeConfig(dest string) error {
 
 	return ioutil.WriteFile(dest, buff.Bytes(), 0600)
 }
+
+// WaitForReady uses kubectl inside the "node" container to check if the
+// master nodes are "Ready".
+func (n *Node) WaitForReady(nodeName string, until time.Time) bool {
```
BenTheElder: let's make this a function on Context, and check for all nodes to be ready?

BenTheElder: all nodes matching the selector*

alejandrox1: @BenTheElder  The function relies on `tryUntil` from pkg/cluster/nodes, so not importable.
What if we added it as a utility function in pkg/cluster/nodes/nodes.go?
Change function sig to `func WaitForReady(node *nodes.Node, until time.Time) bool`?

BenTheElder: That makes sense to me. We can always move it later as well.

neolit123:
/retest
the -parellel job seems to be failing a lot.

pkg/cluster/context.go
```diff
@@ -349,6 +351,14 @@ func (cc *createContext) provisionControlPlane(
 		}
 	}
 
+	// Wait for control plane node to reach Ready status.
+	isReady := nodes.WaitForReady(node, time.Now().Add(cc.waitForReady))
+	if cc.waitForReady > 0 {
+		if !isReady {
+			log.Warn("timed out waiting for control plane to be ready")
```
neolit123: `for control` -> `for the control`

pkg/cluster/context.go
```diff
@@ -48,9 +48,10 @@ type Context struct {
 // createContext is a superset of Context used by helpers for Context.Create()
 type createContext struct {
 	*Context
-	status *logutil.Status
-	config *config.Config
-	retain bool // if we should retain nodes after failing to create
+	status       *logutil.Status
+	config       *config.Config
+	retain       bool          // if we should retain nodes after failing to create.
+	waitForReady time.Duration // Wait for master node to be ready.
```
neolit123: `master node` -> `a control plane node`

pkg/cluster/nodes/nodes.go
```diff
@@ -99,3 +100,35 @@ func list(visit func(string, *Node), filters ...string) error {
 	}
 	return nil
 }
+
+// WaitForReady uses kubectl inside the "node" container to check if the
+// master nodes are "Ready".
```
neolit123: `control plane nodes`

pkg/cluster/nodes/nodes.go
```diff
@@ -99,3 +100,35 @@ func list(visit func(string, *Node), filters ...string) error {
...
+	return tryUntil(until, func() bool {
+		cmd := node.Command(
+			"kubectl",
+			"--kubeconfig=/etc/kubernetes/admin.conf",
+			"get",
+			"nodes",
+			"--selector=node-role.kubernetes.io/master",
+			// When the node reaches status ready, the status field will be set
+			// to true.
+			"-o=jsonpath='{.items..status.conditions[-1:].status}'",
```
neolit123: like i mentioned in one of the comments the better approach is to watch an expected list of pods, but this is also OK in most cases.

cmd/kind/create/cluster/createcluster.go
```diff
@@ -50,6 +52,7 @@ func NewCommand() *cobra.Command {
 	cmd.Flags().StringVar(&flags.Config, "config", "", "path to a kind config file")
 	cmd.Flags().StringVar(&flags.ImageName, "image", "", "node docker image to use for booting the cluster")
 	cmd.Flags().BoolVar(&flags.Retain, "retain", false, "retain nodes for debugging when cluster creation fails")
+	cmd.Flags().DurationVar(&flags.Wait, "wait", time.Duration(0), "Wait for master node to be ready")
```
neolit123: `master` -> `control plane`
needs examples such `30s`, `10m`

pkg/cluster/nodes/nodes.go
```diff
@@ -99,3 +100,35 @@ func list(visit func(string, *Node), filters ...string) error {
 	}
 	return nil
 }
+
+// WaitForReady uses kubectl inside the "node" container to check if the
+// master nodes are "Ready".
+func WaitForReady(node *Node, until time.Time) bool {
+	return tryUntil(until, func() bool {
```
neolit123: `tryUntil` might need an extension to treat `0` as an infinite attempts.
to my understanding with the default value of `0` it will not execute the first kubectl call.

would it make sense to add a default value of say ~4 minutes like in kubeadm?

pkg/cluster/nodes/nodes.go
```diff
@@ -99,3 +100,35 @@ func list(visit func(string, *Node), filters ...string) error {
...
+		if len(lines) != 1 {
+			return false
+		}
+
+		// Check node status. The jsonpath output will come with quotes around
+		// the status.
+		if lines[0] == "'True'" {
+			return true
+		}
+
```
neolit123: it might be worth adding a small time.Sleep() duration in the function. 1 second seems OK to me.

cmd/kind/create/cluster/createcluster.go
alejandrox1: Added a default in the cmd line flag help message which indicates the use of units.
Other than that, to clarify this I added it in the user guide to avoid overcrowding the help message.

pkg/cluster/nodes/nodes.go
BenTheElder: I think we should actually default to 0, with no extra waiting.

Why?

Well waiting is mostly a mis-optimization to reduce flakiness of certain tests suites (_glares at k/k e2e_) that do not handle waiting for cluster readiness very well, there's still reasons the code doing tests should handle this. 

Most CI jobs will probably do something like (ideally - we should document this):

- bring up cluster
- do some other tasks (build thing to test etc)
- start testing
- dump logs
- bring down cluster

For that case, it's better to just exit as soon as possible and get on to the other work while the cluster finishes minor readying (CNI really) in the background.


More importantly, the CLI will hang for local users for up to 4 minutes, which is not great. They again will likely do:
- create cluster
- start building something
- make cup of coffee
- run some tests
- ...

We shouldn't hang any more than kubeadm will.
If for some reason this turns out to be problematic, we can add a small default wait and increase it as needed in future versions. 

I'd rather we avoid anything to slow down this part until we actually need it though. So far it's just a single k8s e2e job that has issues with this.

pkg/cluster/nodes/nodes.go
BenTheElder: the e2e tests already do a variant of this, and it doesn't work properly. the pods exist before CNI is ready.

pkg/cluster/nodes/node.go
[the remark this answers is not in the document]
BenTheElder: we're not waiting to see if it's healthy, we're waiting for scheduling to likely not fail for test pods in https://testgrid.k8s.io/conformance-kind#kind,%20master%20(dev)%20%5Bnon-serial%5D, 

I don't think kind should build it's own healthchecking yet. but logically workloads cannot schedule if the nodes are not in the ready state.

pkg/cluster/context.go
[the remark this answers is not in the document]
BenTheElder: so the control plane is up, but that doesn't mean non-control plane workloads are ready to schedule yet. I think default should be zero (cluster create is snappy locally), and configurable for CI where you can decide yourself how long is worth waiting before just trying to run your tests.

If your tests fail, you need to identify why already, and this way it's up to you how much you want to try to wait. 

The wait still won't be a strong guarantee, and I don't think we want to solve that yet, that seems like something that should be solved with a kubeadm command instead of being implemented just for `kind`.

On manual local usage, you won't be instantly trying to schedule workloads anyhow.

pkg/cluster/nodes/nodes.go
```diff
@@ -99,3 +100,35 @@ func list(visit func(string, *Node), filters ...string) error {
...
+			"--selector=node-role.kubernetes.io/master",
+			// When the node reaches status ready, the status field will be set
+			// to true.
+			"-o=jsonpath='{.items..status.conditions[-1:].status}'",
+		)
+		lines, err := exec.CombinedOutputLines(cmd)
+		if err != nil {
+			return false
+		}
+		if len(lines) != 1 {
```
BenTheElder: this parsing will not work for multi-node, we should add a TODO around this at least as we have ongoing work for multi-node now.

pkg/cluster/nodes/node.go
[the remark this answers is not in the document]
neolit123: for workloads this is OK.
for things like `join --experimental-control-plane` it can fail if the existing etcd cluster is not ready to join yet another etcd node. there are some tricky timeouts around HA, but we shall see.

pkg/cluster/nodes/nodes.go
neolit123: given we are hardcoding weave, we can potentially anticipate it (somehow).

pkg/cluster/nodes/nodes.go
alejandrox1: Completely missed this! Thank you. I think I fixed this

pkg/cluster/nodes/nodes.go
neolit123: > Well waiting is mostly a mis-optimization to reduce flakiness of certain tests suites (glares at k/k e2e) that do not handle waiting for cluster readiness very well, there's still reasons the code doing tests should handle this.

true, random waiting is super nondeterministic and asking for trouble.

so default of 0 means no kubectl calls, as far as i can see it.
sounds good, and hopefully it works. :)

PR #144, opened 2018-11-29
Fix diagram image in design doc

benmoss:
Could also use a relative link, but :man_shrugging:


BenTheElder:
/retest
something got flakier on kubernetes master :(

BenTheElder:
/retest
if this fails I will manually merge and sort this out later. also looking into this out of band, I don't think it's any change to `kind`.

PR #146, opened 2018-12-04
hotfix: add conntrack to base image

BenTheElder:
this is now a dependency of the kubelet debs: https://github.com/kubernetes/kubernetes/commit/2d7b386193d2fc1f4f4669ca0b237fd8474d63c3#diff-b0da275520918e23dd615e2a747528f1

/kind bug
/priority critical-urgent

this should bring master back to functioning again (after getting some bugs fixed upstream!) 😅


hack/build/push-base.sh
```diff
@@ -36,7 +36,7 @@ KIND="${KIND:-$(get_kind)}"
 
 # generate tag
 DATE="$(date +v%Y%m%d)"
-TAG="${DATE}-$(git describe --tags --always --dirty)"
+TAG="${DATE}-$(git describe --always --dirty)"
```
BenTheElder: we now have tags, but _only_ at releases, so using tags doesn't make sense

BenTheElder:
@munnerz 
TBR - this unbreaks CI at k8s master, which will allow other PRs here to merge ...

PR #143, opened 2018-11-29
add v1alpha2 config version

fabriziopandini:
This PR adds v1alpha2 config version.

v1alpha2 currently is copy/paste from v1alpha1, but in following PRs it will be modified for supporting multi node as defined https://github.com/kubernetes-sigs/kind/issues/130.
 
Fuzzer tests were added to make sure conversion works.
Ending test for the new version were added as well

**Note for reviewers** 
To make everyhing work it was necessary to update some dependencies, and this accounts the size of the PR 😭(all deps are in a separated commit)  :
- `k8s.io/apimachinery` and `k8s.io/code-generator` now are both from master branch (before `k8s.io/code-generator` was pinned to `release-v1.11`)
- vendored `k8s.io/apimachinery`, `k8s.io/code-generator` and `k8s.io/gengo` where updated to get a consistent set of dependencies after the switch to klog

Issue #133, opened 2018-11-23
Define how flags / cli / configuration should look for multi node


fabriziopandini:
Multi node is going to increase complexity of the config options available in kind (see https://github.com/kubernetes-sigs/kind/issues/130).

It should be defined how this will impact on  flags / cli, considering at least following categories of users/use cases:
- "beginner": just give me a working, single node cluster.
- "intermediate": give me a working cluster with more than one node with some Kubernetes feature enabled
- "advanced": give me a working cluster with (almost) everything customized
- "developer": let me spin a cluster - any of the above variants - and give me full control of the actions executed after creating the cluster (see https://github.com/kubernetes-sigs/kind/issues/131)

BenTheElder:
In no particular order CCing some others for their thoughts on this: @neolit123 @liztio @munnerz

I think roughly the plan outlined above makes a lot of sense, if we go this route we should group the flags by "usage level" in the CLI help text.

neolit123:
random thoughts:

> --worker-nodes number of worker nodes
--control-plane-nodes number of control-plane nodes
--external-etcd to create and extenal-etcd cluster

while we can create replicas based on a template, there has to be a way to configure parameters for them - i.e. passing a template, or a ConfigMap (like proposed in the multi-node PR).

while passing a number of certain node types may satisfy the basic use case, soon after the user will ask "but how do i configure these replicas?"

> --kube-dns to use kube-dns instead of core-dns
--dynamic-kubelet-config to configure

seems to me that these are fairly low priority, in fact i think we should _only_ expose them when we reach the point of user demand. DKC is problematic, while kube-dns will see less use in time.

> "intermediate": additional flags could be added to link to one or more patches file to be applied to the kubeconfig files
--patches

what if the user wants to patch something all the way down to the api-server manifest on all CP replicas?

> "advanced": use following kind create cluster flags to get full control
--config path to a config file

i can consider config file use as pretty standard for beginners too, so my design would have been to use a config as swiss army knife for everything but only expose CLI flags that make sense for the sake of not creating a config file.

> "developer": use following kind create cluster flags
--action to specify which action kind should perform after container creation ("none" means bare containers)

we need to provide a way to target a container or target a set of replicas.
i wonder if we should not expose actions from the CLI, but either accept an action file (config format) or support Actions in the same config that holds the Nodes objects.

fabriziopandini:
According to feedbacks/latest PR merged:
[...]

PR #147, opened 2018-12-04
Multi node config

fabriziopandini:
This PR completes the implementation of support for multi node configurations in Kind, leveraging on changes introduced by https://github.com/kubernetes-sigs/kind/pull/137 and https://github.com/kubernetes-sigs/kind/pull/143

fixes: https://github.com/kubernetes-sigs/kind/issues/130

**Note for reviewers** 
While implementing I tried to keep the kind Config surface as minimal as possible and now the `v1alpha2` version of kind configuration has only one object called `Node`, that can be repeated many times in a yaml document and/or set to automatically generate replicas. e.g.

```yaml
# ha cluster with 3 control-plane nodes, 2 workers and an external load balancer
kind: Node
role: control-plane
replicas: 3
---
kind: Node
role: worker
replicas: 2
---
kind: Node
role: external-load-balancer
```

Nodes are then grouped in a `Config` object, but this object exists only in the  internal version of kind configuration and so the user should not care about it.

Automatic conversion from `v1alpha1` configurations is supported.

Finally, please note that while the `v1alpha2` version of kind configuration supports multi-node, the rest of Kind still not, so the user will be temporarily blocked when trying to create clusters with n>1 nodes.


pkg/cluster/config/config_types_test.go
```diff
@@ -0,0 +1,189 @@
...
+			checkNodeList(t2, cfg.Workers(), c.ExpectWorkers)
+			checkNode(t2, cfg.ExternalEtcd(), c.ExpectEtcd)
+			checkNode(t2, cfg.ExternalLoadBalancer(), c.ExpectLoadBalancer)
+		})
+	}
+}
+
+func checkNode(t *testing.T, n *Node, name *string) {
+	if (n == nil) != (name == nil) {
+		t.Errorf("expected %v node, saw %v", name, n)
```
neolit123: isn't it the same as doing `n != name` ?

also not sure how to formulate this message.
ideally we should be comparing objects of the same type e.g. bool vs bool.

pkg/cluster/config/encoding/scheme.go
```diff
@@ -43,59 +49,172 @@ func init() {
...
-	utilruntime.Must(scheme.SetVersionPriority(v1alpha1.SchemeGroupVersion))
+	utilruntime.Must(v1alpha2.AddToScheme(scheme))
+	utilruntime.Must(scheme.SetVersionPriority(v1alpha2.SchemeGroupVersion))
+}
+
+// Load reads the file at path and attempts to convert into a `kind` Config; the file
+// can be one of the different API versions defined in scheme.
+// If path == "" then the default config is returned
+func Load(path string) (*config.Config, error) {
+	if path == "" {
```
neolit123: are there messages in the pipeline saying "using default config path" vs "using **your** config path"

pkg/cluster/config/encoding/scheme.go
```diff
@@ -43,59 +49,172 @@ func init() {
...
+	Contents         []byte
+}
 
-	// decode data into a config object
-	_, _, err := Codecs.UniversalDecoder().Decode(data, nil, cfg)
+// unmarshalConfig attempt to decode data into a `kind` Config; data can be
+// one of the different API versions defined in the Scheme; for v1alpha2
+// multiple yaml documents within the same yaml file are supported
+func unmarshalConfig(contents []byte) (*config.Config, error) {
+	// parses yamlDocuments separated by --- directives
```
neolit123: so basically multiple objects of kind Node at this point?

pkg/cluster/config/config_types.go
```diff
@@ -0,0 +1,220 @@
...
+			// assign selected name for worker node
+			replica.Name = "worker"
+			// stores the node in derivedConfigData
+			c.workers = append(c.workers, replica)
+		}
+
+		// node with external etcd role
+		if replica.IsExternalEtcd() {
+			if c.externalEtcd != nil {
+				return fmt.Errorf("invalid config. there are two nodes with role %q", ExternalEtcdRole)
```
neolit123: we should use the pkg/errors package everywhere.
https://godoc.org/github.com/pkg/errors

pkg/cluster/config/encoding/scheme.go
```diff
@@ -43,59 +49,172 @@ func init() {
...
-	// read in file
-	contents, err := ioutil.ReadFile(path)
-	if err != nil {
-		return nil, err
+	return cfg, nil
+}
+
+// splitYAMLDocuments divides a yaml file into yamlDocuments,
+// and identifies each document with its own GroupVersionKind
+func splitYAMLDocuments(contents []byte) (yamlDocuments []yamlDocument, err error) {
```
neolit123: we really need a unified location for this type of utility.
this is pretty much duplicating kubeadm at this point. :\

and once the ComponentConfig KEP stuff is in place we need to revise all locations that bake their own utility.

pkg/cluster/config/encoding/scheme_test.go
```diff
@@ -17,66 +17,213 @@ limitations under the License.
...
+	"io/ioutil"
 	"reflect"
 	"testing"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
-// TODO(fabrizio pandini): once we have multiple config API versions we
-// will need more tests
+func TestSplitYAMLDocuments(t *testing.T) {
```
neolit123: and if we have a unified location for the split utility we can avoid some of the tests too.

pkg/cluster/config/types.go
```diff
@@ -75,3 +108,52 @@ type LifecycleHook struct {
 	// the boot process will continue
 	MustSucceed bool
 }
+
+// IsControlPlane returns true if the node hosts a control plane instance
+func (n *Node) IsControlPlane() bool {
```
neolit123: this is something that i've mentioned earlier - are public methods for root internal config objects something we want?

pkg/cluster/config/v1alpha2/doc.go
```diff
@@ -0,0 +1,23 @@
...
+limitations under the License.
+*/
+
+// Package v1alpha2 implements the v1alpha2 apiVersion of the `kind` Config
+// that introduces multi node support
+//
+// +k8s:deepcopy-gen=package
+// +k8s:conversion-gen=sigs.k8s.io/kind/pkg/cluster/config
+// +k8s:defaulter-gen=TypeMeta
+package v1alpha2
```
neolit123: maybe it's time to start adding docs with Node config examples.

pkg/cluster/config/v1alpha2/types.go
```diff
@@ -0,0 +1,101 @@
...
+	// TODO(bentheelder): split this into image and tag?
+	Image string `json:"image,omitempty"`
+	// KubeadmConfigPatches are applied to the generated kubeadm config as
+	// strategic merge patches to `kustomize build` internally
+	// https://github.com/kubernetes/community/blob/master/contributors/devel/strategic-merge-patch.md
+	// This should be an inline yaml blob-string
+	KubeadmConfigPatches []string `json:"kubeadmConfigPatches,omitempty"`
+	// KubeadmConfigPatchesJSON6902 are applied to the generated kubeadm config
+	// as patchesJson6902 to `kustomize build`
+	KubeadmConfigPatchesJSON6902 []kustomize.PatchJSON6902 `json:"kubeadmConfigPatchesJson6902,omitempty"`
```
neolit123: side note: i'd be interested to hear user feedback on `KubeadmConfigPatches` and `KubeadmConfigPatchesJSON6902`

pkg/cluster/config/validate.go
```diff
@@ -26,12 +26,53 @@ import (
 // with the config, or nil if there are none
 func (c *Config) Validate() error {
 	errs := []error{}
-	if c.Image == "" {
+
+	// All nodes in the config should be valid
+	for _, n := range c.nodes {
+		if err := n.Validate(); err != nil {
+			errs = append(errs, fmt.Errorf("Please fix invalid configuration for node %q: \n%v", n.Name, err))
```
neolit123: errors in this file seem to start with uppercase.
also we should use pkg/errors.

pkg/cluster/config/validate_test.go
```diff
@@ -93,39 +94,137 @@ func TestConfigValidate(t *testing.T) {
...
-		// - nil, in which case we should expect no errors or fail
-		if err == nil {
-			if tc.ExpectedErrors != 0 {
-				t.Errorf("received no errors but expected errors for case %s", tc.TestName)
+		t.Run(tc.TestName, func(t2 *testing.T) {
+			var c = Config{}
+			// Adding nodes to the config
+			for _, n := range tc.Nodes {
+				if err := c.Add(n); err != nil {
+					t.Fatalf("Unexpected error while adding nodes: %v", err)
```
neolit123: lowercase error.

pkg/cluster/context.go
```diff
@@ -138,15 +138,19 @@ func (c *Context) Create(cfg *config.Config, retain bool) error {
 	cc.status.MaybeWrapLogrus(log.StandardLogger())
 
 	defer cc.status.End(false)
-	image := cfg.Image
+
+	// TODO(fabrizio pandini): usage of BootStrapControlPlane() is temporary / WIP
+	// kind v1alpha config fully supports multi nodes, but the cluster creation logic implemented in
```
neolit123: `v1alpha2 config`

pkg/cluster/config/config_types.go
```diff
@@ -0,0 +1,220 @@
+/*
```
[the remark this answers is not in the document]
BenTheElder: :+1: I'd kinda prefer that the config stay just data and we put logic elsewhere other than like, copy, convert. @munnerz voiced a similar opinion.

OTOH, a few of those things are public methods like DeepCopy

pkg/cluster/config/v1alpha2/types.go
BenTheElder: as possibly the sole user currently:
- it works ish
- sometimes for advanced field patching the 6902 format is more powerful, but the plain yaml patch style is much nicer to write and very simple to understand, it just merges the yaml kinda like apply.
- it's not the _best_ UX, but you also can ignore it and isntead run config patching in a lifecycle hook, though the UX will be worse for that.

pkg/cluster/config/validate.go
fabriziopandini: fixed

pkg/cluster/config/validate_test.go
fabriziopandini: fixed

pkg/cluster/context.go
fabriziopandini: fixed

pkg/cluster/config/encoding/scheme.go
fabriziopandini: The only use case where "your config path" will be used, is when the user pass `--config` explicitly, so he/she is already aware of what's going on...

pkg/cluster/config/config_types_test.go
fabriziopandini: n != name can't be used because n and name are different types

the test is validate node ordering, and checking only the node name is a way to do this and at the same time keep test case definition simple and readable

pkg/cluster/config/config_types.go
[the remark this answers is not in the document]
fabriziopandini: @BenTheElder @neolit123 @munnerz 
There are drawbacks in what you are asking

`Config` in v1alpha2 is an helper object that wraps many nodes. It is defined only in the internal version of the config, and it is not available in the public api.

By having this helper object nears few utilities functions I can keep `Config` fields private, and thus ensure that `Nodes` and all the info derived by Nodes respect a set of assumptions that  will greatly simplifies the rest of the code. e.g. Nodes are ordered by provisioning order, Node names are unique, info derived from node are properly set (e.g. bootstrap control plane identified in a consistent way etc)

A prior art for this approach is [`Scheme`](https://github.com/kubernetes/apimachinery/blob/master/pkg/runtime/scheme.go): it is used only internally, it acts as a container for other types, it has private fields and public func for handling internal state.

If we want to split Config by its helper functions probably I will be forced to make `Config` fields public, and then loose some/all the above guarantees. As a consequence all the rest of the code IMO will result much more fragile/error prone.

There are other notable prior arts for where helper function are kept near types e.g.:
https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/meta.go
https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/helpers.go

Said that I'm open to discuss any suggestion helps in cleaning config while preserving above guarantees.

In the meantime, I tried to make things more clear and now types and helper function are in separated files. Also comments were improved accordingly

fabriziopandini:
@BenTheElder 
Rebased after merge of #143, addressed minor comments, answered to other comments

@neolit123 
> i think we should also expose a way to have a common configuration... i wonder if we can have something like a ConfigMap

IMO providing a shortcut for configuration changes is part of #133.
However, it is possible to iterate on this also in future because, if possible, I prefer to keep this PR small and focused.

> how about multi-cluster?

As far as I understand multi cluster is supported in `kind` by decoupling the cluster name from the config. This is preserved by this PR, so you can have multi-cluster/multi-nodes

neolit123:
> As far as I understand multi cluster is supported in kind by decoupling the cluster name from the config. This is preserved by this PR, so you can have multi-cluster/multi-nodes

using the already existing decoupling, would it be possible to configure a multi-cluster topology where an external load balancer serves in front of both clusters?

thanks for the updates.
/lgtm 
/hold

fabriziopandini:
> would it be possible to configure a multi-cluster topology where an external load balancer serves in front of both clusters?

In my hacky prototype I'm creating an ha proxy for each cluster, and I think I will start from this assumption also in moving this feature upstream because implementation will be isimpler without cross cluster dependencies.

rdodev:
This looks really great @fabriziopandini looking forward to take it for a spin.

BenTheElder:
> using the already existing decoupling, would it be possible to configure a multi-cluster topology where an external load balancer serves in front of both clusters?

related: https://github.com/kubernetes-sigs/federation-v2 is doing some work with `kind`, we should speak to them.

https://github.com/kubernetes-sigs/federation-v2/blob/c8398aa4f6439e4e0603c75e513d394a9b4f2595/scripts/download-e2e-binaries.sh#L36-L45

pkg/cluster/config/v1alpha2/doc.go
BenTheElder: yes, let's document this once it's ready for use, and until then keep it to the config example file, for those that want to do some digging.

pkg/cluster/config/v1alpha2/zz_generated.conversion.go
```diff
@@ -149,6 +123,38 @@ func Convert_config_LifecycleHook_To_v1alpha2_LifecycleHook(in *config.Lifecycle
...
+// Convert_v1alpha2_Node_To_config_Node is an autogenerated conversion function.
+func Convert_v1alpha2_Node_To_config_Node(in *Node, out *config.Node, s conversion.Scope) error {
+	return autoConvert_v1alpha2_Node_To_config_Node(in, out, s)
+}
+
+func autoConvert_config_Node_To_v1alpha2_Node(in *config.Node, out *Node, s conversion.Scope) error {
+	out.Replicas = (*int32)(unsafe.Pointer(in.Replicas))
+	out.Role = NodeRole(in.Role)
+	out.Image = in.Image
+	out.KubeadmConfigPatches = *(*[]string)(unsafe.Pointer(&in.KubeadmConfigPatches))
```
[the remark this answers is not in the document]
BenTheElder: ok thanks, I wonder why they're doing this so much.

pkg/cluster/config/helpers.go
```diff
@@ -0,0 +1,213 @@
...
+	"github.com/pkg/errors"
+)
+
+// IsControlPlane returns true if the node hosts a control plane instance
+func (n *Node) IsControlPlane() bool {
+	return n.Role == ControlPlaneRole
+}
+
+// IsWorker returns true if the node hosts a worker instance
+func (n *Node) IsWorker() bool {
```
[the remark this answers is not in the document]
BenTheElder: thanks!

pkg/cluster/config/v1alpha2/zz_generated.conversion.go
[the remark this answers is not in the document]
fabriziopandini: I can only guess:
- because conversion-gen always assumes that we are converting from TypeA.vx to TypeA.vY and does not detect cases when the type is external from the API (the type is the same)
or
- because this is a convenient way for doing "deepcopy" with types not supporting it explicitly

fabriziopandini:
@neolit123 if I'm not wrong all the comment s are addressed.
Could you kindly leave hold?

pkg/cluster/config/config_types.go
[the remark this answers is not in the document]
neolit123: some random points:

> Config in v1alpha2 is an helper object that wraps many nodes. It is defined only in the internal version of the config, and it is not available in the public api.

i guess we are also talking about the Node object, that is public and has it's fields public.
so we don't have to have public methods atteched to the Node struct, but we can instead use helper functions that accept a *Node as parameter.

> https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/meta.go

the metav1 is a good comparison of an API that has public methods for the config struct, i guess.

i was thinking of a package that is importable and has types and interfaces but no definitions. but golang smears the line between declarations and definitions.

leaving it to @BenTheElder for the final call on the public methods.

neolit123:
thanks for the updates @fabriziopandini
/hold cancel

neolit123:
the more i think about it the more i prefer the node list under a single config parent object:
```yaml
kind: KindConfig
nodes:
- role: control-plane
  replicas: 3
- role: worker
  replicas: 2
- role: external-load-balancer
```

i see this method as easier to "UX" and maintain.
maybe i'm missing the benefits of doing the separate GVK for Node objects?

fabriziopandini:
@neolit123 
There are pro and cons. A list is not ideal when you have only one node, and a list somehow constraints you in having all the elements of a list equal, which I'm note sure is something we want in the long term (e.g. having hooks for external load balancer and etcd does not meek entirely sense to me)

Said that, I think that at this stage it is really important to go out as soon as possible in order to start getting using feedback and unblock the remaining part of multi-node activities (some new requirements for config can be identified along the way as well)

BenTheElder:
I finally got some time to just think about this some more and I think the nodes should be a list in a top level object. like @neolit123 pointed out

- This gives us room to add fields that are config wide, rather than on each node
- this ensures that nodes do not have distinct GVK (!), which would otherwise be problematic to support well.

pkg/cluster/config/config_types.go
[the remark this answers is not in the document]
BenTheElder: this seems fairly easy to re-organize before V1, but I would tend to agree that config should be plain old data and helpers can accept a config object.

BenTheElder:
Apologies for the delay on this PR, KubeCon and then post-con "plague" (a cold?) built up a pretty large backlog of reviews. Will be getting to these more regularly now.

fabriziopandini:
@BenTheElder I hope you are well now!

I'm going to address comments because I think it is important to get a first release merged ASAP to get moving also following parts of multi-node efforts, but IMO this leads to a more fragile design, but this is not blocking because we can always iterate in future.

Those are the point that IMO will be the candidate for future iterations:
- having "config wide" fields is an orthogonal problem to having a list/a sequence of objects (in any case "config wide" fields should be hosted on a different object than the list itself)
- having "config wide" fields opens up to precedence/override problems on corresponding node fields (if any)
- having a list assumes all the node are equal, which is a wrong assumption in multi-node/HA scenario
- both the kubeadm config (after several reworks) and the cluster API are implementing different approaches

BenTheElder:
Feeling a bit better, thanks :-)

Note: I do *not* want to rush multi-node _too_ much, we are already in need to removing some technical debt from kind. Previous iterations had multi-node and were not right.

These are interesting points but I'm not convinced.

> having "config wide" fields is an orthogonal problem to having a list/a sequence of objects (in any case "config wide" fields should be hosted on a different object than the list itself)

The list would become a subset. The reason for doing this is to always give the rest of the code a "Config" object to work with, which should be as much as possible simply something read from disk and desearialized with minimal mutation (just defaulting).

> having "config wide" fields opens up to precedence/override problems on corresponding node fields (if any)

The Node fields take precedence over the config wide if any of them somehow clashed. They generally shouldn't though. The same is said in k8s of say resource requests vs inline requests.

> having a list assumes all the node are equal, which is a wrong assumption in multi-node/HA scenario

It only assumes there is a list of "Node" type - why wouldn't there be? Nodes can have different roles.

> both the kubeadm config (after several reworks) and the cluster API are implementing different approaches

I don't think the cluster API in it's current state is desirable to emulate for local clusters. Kubeadm only has separate objects due to the phases. We're not exposing that. There's also no real reason these can't be combined into a single object, I don't think we want the burden of nodes with different GVK.

In general elevating individual node specs to have GVK seems very problematic.

BenTheElder:
Having nodes as a field also means we can default adding a single nodespec when no nodes are specified.

fabriziopandini:
@BenTheElder I would like to discuss with you how to best address the following point

> always give the rest of the code a "Config" object to work with, which should be as much as possible simply something read from disk and desearialized with minimal mutation (just defaulting).

Currently, we are using a replica field at node level. e.g. from @neolit123 example

```
kind: KindConfig
nodes:
- role: control-plane
  replicas: 3
- role: worker
  replicas: 2
- role: external-load-balancer
```

However this is something the rest of the code should not work with, because the rest of the code needs something where the expected "config nodes" are derived from to the replicas number; additionally this "derive" process should take care also of assigning unique names to nodes. e.g. the list of "config nodes" for the above example should be:

```
derivedNodes:
- role: control-plane
  name: control-plane1
- role: control-plane
  name: control-plane2
- role: control-plane
  name: control-plane3
- role: worker
  name: worker1
- role: worker
  name: worker2
- role: external-load-balancer
  name: lb
```

Now that we are exposing the Config object as a public object type, it becames harder to keep the derived part "hidden" from the api-machinery that handles type conversion and defaulting, so I think it will be useful to do a quick pair review before investing to much time into this refactoring

BenTheElder:
I don't think we need to derive them on the object all. Config is config,
and at runtime we can have a completely distinct structure with the node
handles we've produced. When creating any node we can note the replication
number in the config.

neolit123:
i think we should have a discussion about this eventually.

fabriziopandini:
@BenTheElder @neolit123 
Moving to Config objects with nodes list as requested.
Derived fields are now sort of materialized views exiting only in the internal API; this makes easy to refactor them as soon as a better collocation is defined

BenTheElder:
plan is to iterate on this, @fabriziopandini's code is quite excellent and will serve as a great MVP for multi-node, thank you so much for working on this.

As discussed, we're opting to go ahead and merge this now and handle bikeshedding via follow up PR instead of infinite back and forth on the existing PRs :wink: 

Sorry for the delay, I was maybe going to cut another smaller release before putting these in, but we're just going to forge ahead now.

/hold cancel

Issue #28, opened 2018-09-22
First class image side-loading support


BenTheElder:
This will be very useful for CI tasks. Currently this is achievable by adding image archives to a node image in `/kind/images/` but we should support this first-class.

cc @munnerz @sebastienvas

/kind feature
/priority important-longterm

/assign

neolit123:
the pause image has a bit of a special case in terms of the kubelet:
https://github.com/kubernetes/kubeadm/issues/1186#issuecomment-433383514

BenTheElder:
/lifecycle active
FYI @liztio

BenTheElder:
Rough design I didn't finish implementing:
```
kind load docker-image [--name=clustername] [--nodes=] image...
kind load image-archive [--name=clustername] [--nodes=] /path/to/image.tar
```
- implicitly load to all nodes if `--nodes` is unset
- load to specific nodes if `--nodes` is a comma separated list of node of names

`kind load docker-image` should pipe `docker save image/name` on the host to loading on each of the nodes, IE essentially be `kind load image-archive <(docker save image)`.

rohansingh:
@BenTheElder That seems very reasonable to me. I'm happy to do a first-pass on the implementation if nobody else is; being able to load images is a blocker for my use case.

stevesloka:
I should have posted here @rohansingh but I just about have a PR ready to solve this.

BenTheElder:
#309
https://kind.sigs.k8s.io/docs/user/quick-start/#loading-an-image-into-your-cluster

rohansingh:
Thanks again @BenTheElder and @stevesloka. This is invaluable for the "local clusters on developer machines" use case.

BenTheElder:
I hope it will make CI better as well!

kind uses a similar technique internally to load Kubernetes and avoid needing to push images when testing from Kubernetes source, so far that has been great 😄

PR #151, opened 2018-12-05
Attempting to make kind a little less Docker Centric

rhatdan:
I believe that kind should be able to run with Podman as
well as Docker.

Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>


BenTheElder:
Huge fan of this idea! I've been mulling changing it to something like:
> kind - friendly local clusters

(and dropping the acronym) with exactly this purpose in mind (also why the docker related code has been moving into a separate package slowly)

> I did not discover a way of building this, so I relied on validata-all to look for bugs.

`go build` / `go install` should cover the main binary, for the images `kind build base-image` and `kind build node-image`, the former is mostly a `docker build` of `images/base`, but the latter is a bit trickier right now so we can install debians without keeping a copy of them in the image.

I'm not very familiar with podman yet, but we should be able to come up with something, and there are pre-built images at least.

---

Will take a look through the code later today 😄

docs/devel/README.md
```diff
@@ -32,22 +32,31 @@ $ go version
...
+This documentation is written using Podman version 0.12.1.
+
 
 
 [github]: https://github.com/
 [community]: https://github.com/kubernetes/community
 [contributor]: https://github.com/kubernetes/community/blob/master/contributors/guide/README.md
 [golang]: https://golang.org/doc/install
 [docker]: https://docs.docker.com/install/#supported-platforms
+[podman]: https://github.com/containers/libpod/install
```
alejandrox1: This links doesn't exist.  May be https://github.com/containers/libpod/blob/master/install.md ?

rhatdan: Oops

rhatdan: FIxed

pkg/container/run.go
```diff
@@ -25,17 +25,17 @@ import (
 	"sigs.k8s.io/kind/pkg/exec"
 )
 
-// Docker container IDs are hex, more than one character, and on their own line
+// Container IDs are hex, more than one character, and on their own line
 var containerIDRegex = regexp.MustCompile("^[a-f0-9]+$")
 
-// Run creates a container with "docker run", with some error handling
+// Run creates a container with "ContainreEngine run", with some error handling
```
BenTheElder: nit: s/ContainreEngine/ContainerEngine/

pkg/container/init.go
```diff
@@ -0,0 +1,36 @@
...
+// Package container contains helpers for working with container engines
+// This package has no stability guarantees whatsoever!
+package container
+
+import (
+	"os/exec"
+)
+
+//Engine is the container engine to use to launch containers
+var Engine string
```
BenTheElder: can we document the initialization behavior IE something like "by default this will be the first of podman or docker to be found in $PATH, with a fallback to docker" ?

We should probably also note supported values.

pkg/cluster/nodes/node.go
```diff
@@ -108,16 +111,16 @@ func tryUntil(until time.Time, try func() bool) bool {
...
+// LoadImages loads image tarballs stored on the node into container storage on the node
 func (n *Node) LoadImages() {
-	// load images cached on the node into docker
+	// load images cached on the node into Container Storage
 	if err := n.Command(
 		"find",
 		"/kind/images",
 		"-name", "*.tar",
-		"-exec", "docker", "load", "-i", "{}", ";",
+		"-exec", container.Engine, "load", "-i", "{}", ";",
```
BenTheElder: this is calling docker inside the "node" container, so this should still be `"docker"` for now since that's still what runs inside.

pkg/cluster/nodes/node.go
```diff
@@ -141,17 +144,17 @@ func (n *Node) LoadImages() {
 	// TODO(bentheelder): this is a bit gross, move this logic out of bash
 	if err := n.Command(
 		"/bin/bash", "-c",
-		`docker images --format='{{.Repository}}:{{.Tag}}' | grep -v amd64 | xargs -L 1 -I '{}' /bin/bash -c 'docker tag "{}" "$(echo "{}" | sed s/:/-amd64:/)"'`,
+		fmt.Sprintf(`%s images --format='{{.Repository}}:{{.Tag}}' | grep -v amd64 | xargs -L 1 -I '{}' /bin/bash -c '%s tag "{}" "$(echo "{}" | sed s/:/-amd64:/)"'`, container.Engine, container.Engine),
```
BenTheElder: ditto here, this command runs inside the "node" container so it shouldn't use the host container runtime, just `"docker"` for now.

pkg/cluster/nodes/node.go
```diff
@@ -76,17 +76,20 @@ func (n *Node) String() string {
...
 // CopyTo copies the source file on the host to dest on the node
 func (n *Node) CopyTo(source, dest string) error {
-	return docker.CopyTo(source, n.nameOrID, dest)
+	return container.CopyTo(source, n.nameOrID, dest)
 }
 
-// WaitForDocker waits for Docker to be ready on the node
-// it returns true on success, and false on a timeout
+// WaitForDocker waits for Docker Daemon to be ready on the node
+// it returns true on success or if you are not using Docker, and false on a timeout
```
BenTheElder: This function also should not change, this is waiting for the docker service inside the node container to be ready. 

This is the change that is causing the CI to fail.

BenTheElder:
Found three confusing bits -- there's some `node.Command("docker" ...` calls that are running commands against the nested container runtime on the "node"s.

Those should stay the same until we're actually running different runtimes there, and even then should use that container engine, not the hosts's.

Once those are reverted, I think this should pass tests 👍

There's also a small merge conflict now in `pkg/cluster/context.go`, as another PR merged to that file.

runcom:
There would be also be a way to specify which runtime kube uses? I would love to run the cluster with CRI-O container runtime for instance

BenTheElder:
@runcom, yes that is on the TODO but would make a larger follow-up, we will need to figure out:
- the best way to ship the base image once we have variants, or if we should just install multiple runtimes
- what tweaks we need to run cri-o in a container, 
I haven't found anyone looking into this yet and haven't had a chance myself
- CI?

I've punted this for now in favor of getting the rest stable, docker-in-a-container is a path I have some existing expertise with from https://kubernetes/test-infra, but if nobody else gets to it I definitely will eventually :-)

runcom:
> * what tweaks we need to run cri-o in a container,
>   I haven't found anyone looking into this yet and haven't had a chance myself

I'm gonna look into this, I'll hack around with kind's base image itself and see if tweaking kubeadm plus having CRI-O in it is a  thing.

I think  a reasonable first step is the above: have a mean to support another container runtime, be it *in* docker or podman doesn't really matter for now. Afterwads we can chat and agree about the way we're going to ship the base/node image (multiple runtimes or different images or whatever). I'll take a stab at first running CRI-O to begin with.

For the CI, I'm not sure how things are setup right now in this repo, but once we agree on a delivery path for the images, it hopefully is going to be straightforward to figure out and integrate with the current CI, right?

pkg/container/run.go
rhatdan: Fixed

pkg/container/init.go
rhatdan: Improved the comment.

pkg/cluster/nodes/node.go
rhatdan: Ok I will change it back, but I actually think podman inside of that container would make more sense since you don't need ot setup a separate daemon, or is this container volume mounting in the docker.sock?

pkg/cluster/nodes/node.go
rhatdan: Ok switched back to Docker, although left them as being passed in, so we could move quickly to using alternatives inside of the container.

pkg/cluster/nodes/node.go
rhatdan: Ok, I guess if you were using podman on the host and docker within the container this would be an issue.  But this clearly shows for this use case why podman inside of the container would be better, no daemon to start and wait for.

BenTheElder: Sure, but podman does not implement a CRI for Kubernetes as far as I know. Docker in the container is purely for that use case.

We will be exploring cri-o as an option soon. @runcom started looking into this.

rhatdan:
@runcom lets work together to get CRI-O running with Docker.  But I think we will beed a decent amount of work in node directory to support two or more container engines.

BenTheElder:
> Right, although I think you are combining two different things.

Which two things?
Loading the images necessarily requires knowing about the runtime on the node.

> We can do podman load to load the images into CRI-O Storage, then if we have to actually talk to CRI-O we would need to do crictl.

We don't have cri-o storage to talk to yet. And we'll need to detect and handle the node CRI in general distinctly from the host runtime, for which we need another set of unrelated changes.

For now it doesn't make sense to touch the code running inside the node because that can only possibly be talking to docker. We can follow up with that.

Shipping multiple runtimes will have some tradeoffs no matter how we do it, and should come with some further discussion.

pkg/container/init.go
BenTheElder: thanks

pkg/cluster/nodes/node.go
BenTheElder: podman does look very cool for that, however:
- We're using docker for now because of existing expertise and accesibility to both our maintainers and the bulk of our known current and potential users. 
- I also need to maintain another more general docker-in-docker setup for [kubernetes/test-infra](https://github.com/kubernetes/test-infra) where it is used pervasively by many projects we merely support and that is less likely to change, as we'd need to convince them all to switch. @munnerz also maintains a parallel test-infra setup to this one.

It's certainly worth looking at, but I don't think the second point would happen and there is very large overlap in maintaining and debugging these setups. There's also a _lot_ more information out there from third parties related to dind. I think @runcom might be the first person to run cri-o in a container :upside_down_face: 

We're not mounting sockets.

Other than loading the images, only Kubernetes is supposed to be talking to this, however some advanced users are already talking to it, mostly to load their images until #28.

README.md
```diff
@@ -28,8 +28,8 @@ If you have [go] and [docker] installed `go get sigs.k8s.io/kind && kind create
 
 You can install `kind` with `go get sigs.k8s.io/kind`
 
-To use `kind`, you will need to [install docker].  
-Once you have docker running you can create a cluster with `kind create cluster`  
+To use `kind`, you will need to [install docker] or [install podman].
+Once you have installed podman or have docker service running you can create a cluster with `kind create cluster`
```
BenTheElder: nit: "the docker service"

docs/design/base-image.md
```diff
@@ -1,14 +1,14 @@
...
+The ["base" image][base image] is a small-ish container image for running
 nested containers, systemd, and kubernetes components.
 
 To do this we need to set up an environment that will meet the CRI 
 (currently just docker) and systemd's particular needs. Documentation for each
 step we take is inline to the image's [Dockerfile][dockerfile]),
 but essentially:
 
-- we preinstall tools / packages expected by systemd / Docker / Kubernetes other
+- we preinstall tools / packages expected by systemd / Docker | Podman / Kubernetes other
```
BenTheElder: this is not true at the moment. this image ships docker for now. it does not ship podman

docs/design/base-image.md
```diff
@@ -1,14 +1,14 @@
 # The Base Image
 
-The ["base" image][base image] is a small-ish Docker image for running
+The ["base" image][base image] is a small-ish container image for running
```
BenTheElder: This particular image is built from a `Dockerfile`, I'm not sure how helpful this change is.
I think podman / cri-o / containerd users know that a "docker image" is not limited to Docker. :^)

docs/design/design.md
```diff
@@ -63,17 +63,17 @@ For more see [node-image.md][node-image.md].
 
 ### Cluster Creation
 
-Each "node" runs as a docker container. Each container initially boots to a
+Each "node" runs as a container. Each container initially boots to a
 pseudo "paused" state, with [the entrypoint][entrypoint] waiting for `SIGUSR1`.
-This allows us to manipulate and inspect the container with `docker exec ...`
+This allows us to manipulate and inspect the container with `containerEngine exec ...`
```
BenTheElder: I think this sort of change is going to be more confusing than anything. podman is the only tool I know of besides docker that provides something like this, and does so by mimicing the docker CLI AFAICT... "docker exec" is searchable when looking for upstream doc / what it is "containerEngine exec" is not.

how about "with eg `docker exec`" - it seems like podman users will still understand what this maps to.

docs/design/design.md
```diff
@@ -63,17 +63,17 @@ For more see [node-image.md][node-image.md].
...
 and other tools prior to starting systemd and all of the components.
 
-This setup includes fixing mounts and pre-loading saved docker images.
+This setup includes fixing mounts and pre-loading saved container images.
 
 Once the nodes are sufficiently prepared, we signal the entrypoint to actually
 "boot" the node.
 
-We then wait for the Docker service to be ready on the node before running
+We then wait for the Docker service to be ready if we are using Docker on the node before running
```
BenTheElder: as it stands, we are only using docker for this case.

docs/design/design.md
```diff
@@ -84,7 +84,7 @@ At this point users can test Kubernetes by using the exported kubeconfig.
 
 ### Cluster Deletion
 
-All "node" containers in the cluster are tagged with docker labels identifying
+All "node" containers in the cluster are tagged with container labels identifying
```
BenTheElder: again, this is searchable, I'm not sure I want to sacrifice this. docker labels have a lot of restrictions and particular behavior, what those are is probably interesting to many of our users / developers who may not have used these yet.

docs/design/node-image.md
```diff
@@ -35,6 +35,6 @@ each "node" container with [kubeadm][kubeadm].
 [base image]: ./base-image.md
 [build package]: ./../../pkg/build
 [cluster package]: ./../../pkg/cluster
-[docker image archives]: https://docs.docker.com/engine/reference/commandline/save/
+[container image archives]: https://docs.docker.com/engine/reference/commandline/save/
```
BenTheElder: unless we have a generic reference for this instead, it seems like renaming this and then linking to docker is not particularly helpful.

docs/devel/README.md
```diff
@@ -6,7 +6,7 @@ Before you being you must have:
 * [GitHub account][github]
 * `git`
 * `go`
-* `Docker`
+* `Docker` | `Podman`
```
BenTheElder: `s/|/or/`

pkg/container/run.go
```diff
@@ -25,17 +25,17 @@ import (
 	"sigs.k8s.io/kind/pkg/exec"
 )
 
-// Docker container IDs are hex, more than one character, and on their own line
+// Container IDs are hex, more than one character, and on their own line
 var containerIDRegex = regexp.MustCompile("^[a-f0-9]+$")
 
-// Run creates a container with "docker run", with some error handling
+// Run creates a container with "ContainerEngine run", with some error handling
```
BenTheElder: let's just say Engine since at least that maps to the var we're using. the reason for mentioning "docker run" is to correlate with the documented behavior there without documenting it ourselves.

docs/design/design.md
neolit123: > how about "with eg docker exec" - 
or something like:
> with tools like `docker exec`

pkg/container/init.go
```diff
@@ -0,0 +1,38 @@
...
+	"os/exec"
+)
+
+// Engine is the container engine to use to launch containers
+// Currently we support podman and docker.  Kind will use podman if it is
+// found in $PATH else it fall back to docker.
+var Engine string
+
+func init() {
+	for _, engine := range []string{"podman", "docker"} {
```
neolit123: i think this needs to be exposed from a CLI flag or via an environment variable (preferring the former).
having this precedence can be problematic if someone has podman installed but wants to use docker for kind instead.

docs/devel/README.md
neolit123: IMO, better to have "a container runtime" here and have a separate section that lists the supported options.

BenTheElder:
I've been thinking about how to do better with container runtimes both on the nodes and on the host going forward ... I think we should get away from the `container.Engine` global and `init` handling this. Still mulling the best option.

randomvariable:
@rhatdan , is there something else in libpod that's needed to get this working locally?

```
ERRO[10:45:27] unable to pull kindest/node:v1.13.2@sha256:e14edfa4950e009fe560499c9db6e89daae8bd18bcb372caca6d321a86c52cda: error getting default registries to try: unable to parse 'kindest/node:v1.13.2@sha256:e14edfa4950e009fe560499c9db6e89daae8bd18bcb372caca6d321a86c52cda': Docker references with both a tag and digest are currently not supported
```

BenTheElder:
We're making a concerted effort to refactor and support this, I'd like to implement differently (no globals vas in the packages, no lookup in `init()`, an actual per-runtime implementation so we can handle differences ...) and I'd like to not drop all references to docker in the docs quite as agressively so as to not confuse our existing users.

See also: https://kind.sigs.k8s.io/docs/design/principles/#target-cri-functionality, making this possible is a documented requirement for a little while now.

This is tracked in #154.

rhatdan:
Awesome, thanks.  Please contact me if you need any changes to our container runtimes. (Engines)

PR #157, opened 2018-12-07
kubeadm: support v1beta1 and add all config objects

BenTheElder:
- use the v1beta1 api types internally :tada: 
- add all top level config objects, even if we don't set fields on them, so that users can apply patches to them

/cc @neolit123


pkg/cluster/kubeadm/config.go
```diff
@@ -110,18 +156,24 @@ func Config(data ConfigData) (config string, err error) {
...
 		return "", err
 	}
-	// The complexity of the config does not require special handling
-	// between v1alpha1 and v1alpha2 yet.
-	templateSource := DefaultConfigTemplateAlphaV3
-	if ver.LessThan(version.MustParseSemantic("v1.12.0")) {
-		templateSource = DefaultConfigTemplateAlphaV1orV2
+
+	templateSource := ConfigTemplateBetaV1
+	if ver.LessThan(version.MustParseSemantic("v1.13.0")) {
```
BenTheElder: This is wrong (this logic needs to be in the opposite order, lowest check first), will fix when I get back to a computer.

pkg/cluster/kubeadm/config.go
```diff
@@ -101,6 +105,48 @@ kind: InitConfiguration
...
+// ConfigTemplateBetaV1 is the kubadm config template for API version v1beta1
+const ConfigTemplateBetaV1 = `# config generated by kind
+apiVersion: kubeadm.k8s.io/v1beta1
+kind: ClusterConfiguration
+kubernetesVersion: {{.KubernetesVersion}}
+clusterName: {{.ClusterName}}
+# on docker for mac we have to expose the api server via port forward,
+# so we need to ensure the cert is valid for localhost so we can talk
+# to the cluster after rewriting the kubeconfig to point to localhost
+apiServerCertSANs: [localhost]
```
neolit123: it's has changed in v1beta1:
```
apiSever:
  certSANs: ....
```

kustomize is confusing me.

BenTheElder: thanks

which part is confusing? we aren't using it here

BenTheElder: fixed this

pkg/cluster/kubeadm/config.go
BenTheElder: updated

BenTheElder:
/retest
updated the scripts, should be able to see what's going on a bit better

pkg/cluster/kubeadm/config.go
```diff
@@ -101,6 +105,49 @@ kind: InitConfiguration
...
+// ConfigTemplateBetaV1 is the kubadm config template for API version v1beta1
+const ConfigTemplateBetaV1 = `# config generated by kind
+apiVersion: kubeadm.k8s.io/v1beta1
+kind: ClusterConfiguration
+kubernetesVersion: {{.KubernetesVersion}}
+clusterName: {{.ClusterName}}
+# on docker for mac we have to expose the api server via port forward,
+# so we need to ensure the cert is valid for localhost so we can talk
+# to the cluster after rewriting the kubeconfig to point to localhost
+apiSever:
```
BenTheElder: s/apiSever/apiServer/

BenTheElder:
bad copy paste on the alphav3 kubelet apiVersion

neolit123:
looks like i'm missing stuff today, sorry.
/lgtm

BenTheElder:
no worries! thanks for following up :-)

that was the last issue :sweat_smile:

PR #162, opened 2018-12-09
fixes command for building from source

k8k:
Building without an image flag defaults to a different image (not the one you just built with `kind build node-image`).

This PR adds an image flag so the image built from source is used.


k8k:
CLA signed

Issue #131, opened 2018-11-23
Introduce actions


fabriziopandini:
as per https://github.com/kubernetes-sigs/kind/issues/117#issuecomment-440531457, one of the required steps for implementing multi node support is to add better support for taking actions across a topology of nodes.

The rationale behind the need of "actions" dependent on "topology of nodes", is that different nodes in the cluster require different actions to be executed after creating the container, e.g. 
- the bootstrap control-plane must execute `kubeadm init` 
- the additional control-plane nodes (0..n) must execute `kubeadm join --control-plane`
- the worker nodes (0..n) must execute `kubeadm join`

By introducing actions as a separated step from container creation, a set of interesting possibilities open up for kind, like e.g.
- to automate new actions to be executed after cluster creation (e.g. `kubeadm upgrade`, `e2e` etc. etc.)
- to allow users to create "bare containers" and then execute actions step by step --> this use case will allow usage of kind for kubeadm - or kubernetes - dev/test iterations

A first proposal of the initial set of about actions to be implemented is:
1. kubeadm config
2. kubeadm init
3. kubeadm apply network
4. kubeadm join (both control-plane and worker nodes)

opinions?
/assign

Issue #161, opened 2018-12-08
write a kubetest deployer for kind


neolit123:
@BenTheElder not sure if a tracking issue already exists for this (tried searching for it).
i can try helping with the initial work for a deployer.

/assign
/kind feature
/priority important-soon

BenTheElder:
Thanks, one did not and that would be great!

I had something sketched out but never got it done. Two ideas from that:

- fetch a binary release for the platform and stick it in `$HOME/.kubetest/kind/$binary` so we can call it from kubetest.
  - checksum (sha256) of the binary there would let us avoid re-downloading when it is already downloaded. We could pre-download in the kubetest CI image too once this works.
- we'll probably need to integrate with --build too so it does a kind build node-image instead. We will probably have to make build depend on the deployer 🤷‍♂️

fabriziopandini:
@neolit123 @BenTheElder 
Might be a silly question, but are there some documents about how kubetest could be configured / how kubetest parameters will flow down to kind?

I ask this because IMO those info are the key requirements for designing kind UX for CI use case; more specifically I'm thinking at:
- how can I specify cluster Config/a desired cluster topology
- how can I specify a complex test plan like e.g init-join-e2e and then upgrade-join another node-e2e again etc

neolit123:
> Might be a silly question, but are there some documents about how kubetest could be configured / how kubetest parameters will flow down to kind?

i don't know if such a document exists.

looking at my experience with k-a, we have to store the config scenarios "somewhere" as we cannot expose the whole config or actions as CLI flags from jobs to be passed to the kind kubetest deployer.

in the case of k-a the configs were stored in the k-a repo and we were binding test-infra (kubeadm image) to a specific version of k-a. in my view this is problematic because any tiny change we make in the config scenarios has to be reflected in test-infra and images have to be rebuild.

that is the reason i pushed hard for the creation of the k-a `kubeadm-e2e` branch. this removed the need for us to hardcode a specific SHA and made it possible for use to use the tip of a branch.

so my proposal for kind is to use something similar. it's just more flexible.

neolit123:
@BenTheElder 
> For parameters flowing down, we can just plumb through a few like the kind config path

i started implementing the deployer interface yesterday and this is pretty much the only flag for now.

i have a number of questions, but the few major ones are:
1) where to store the configs?
2) would the config allow actions and test scenarios (cc @fabriziopandini ) my assumption is that this is our plan?
[...]

PR #164, opened 2018-12-10
Introduce Actions; fully support multi-node with n>0 workers

fabriziopandini:
This PR introduces actions in `kind`

Actions consist of set of repeatable tasks to be executed on a `kind` cluster.
Usage of actions allows defining composable, high-level workflows that can automatically adapt to different cluster topologies.

In order to make this more concrete, the first set of action supporting multi-node with n>0 workers was implemented in this PR.

Fixes: https://github.com/kubernetes-sigs/kind/issues/131

**Note for reviewers** 
This PR depends on https://github.com/kubernetes-sigs/kind/pull/147; please consider only last commit

/cc @BenTheElder @neolit123 
@rdodev


pkg/cluster/actions.go
```diff
@@ -0,0 +1,231 @@
...
+
+	// PlannedTask should respects the given order of actions and tasks
+	actionIndex int
+	taskIndex   int
+}
+
+// ExecutionPlan contain an ordered list of Planned Tasks
+// Please note that the planning order is critical for providing a
+// predictable, "kubeadm friendly" and consistent execution order.
+type ExecutionPlan []*PlannedTask
```
neolit123: questions:
- is the node blocking until a planning tasks finishes before executing the next planned task?
i'm yet to see the rest of the code but tasks should be added in a queue of sorts.
- what if a task blocks indefinitely, is there a way to get out of that?

pkg/cluster/actions.go
```diff
@@ -0,0 +1,231 @@
...
+// NodeSelector defines a function returning a subset of nodes where tasks
+// should be planned.
+type NodeSelector func(*config.Config) config.NodeList
+
+// PlannedTask defines a Task planned for execution on a given node.
+type PlannedTask struct {
+	// task to be executed
+	Task Task
+	// node where the task should be executed
+	Node *config.Node
```
neolit123: PlannedTask has a Node reference, but also a Task and this Task holds a list of Nodes.

ideally an Action should have a list of Tasks.
and a Task should execute on a list of Nodes.

we should think of how to omit the PlannedTasks object.

pkg/cluster/actions.go
```diff
@@ -0,0 +1,231 @@
...
+func (t ExecutionPlan) Swap(i, j int) {
+	t[i], t[j] = t[j], t[i]
+}
+
+// ExecutionOrder returns a string that can be used for sorting planned tasks
+// into a predictable, "kubeadm friendly" and consistent order.
+// NB. we are using a string to combine all the item considered into something
+// that can be easily sorted using a lexicographical order
+func (p *PlannedTask) ExecutionOrder() string {
+	return fmt.Sprintf("Node.ProvisioningOrder: %d - Node.Name: %s - actionIndex: %d - taskIndex: %d",
```
neolit123: i might need a diagram to better understand how this work j/k.
in all seriousness, i think the sorting introduces a non-deterministic factor that could be avoided.

some questions:
- why do we even need the sorting?
- what if we want to execute tasks with an order not based on the node names?

pkg/cluster/kubeadm/const.go
```diff
@@ -19,3 +19,6 @@ package kubeadm
 // APIServerPort is the expected default APIServerPort on the control plane node(s)
 // https://kubernetes.io/docs/reference/access-authn-authz/controlling-access/#api-server-ports-and-ips
 const APIServerPort = 6443
+
+// Token defines a dummy, well known token for automating TLS bootstrap process
+const Token = "abcdef.0123456789abcdef"
```
neolit123: `DefaultToken`?

pkg/cluster/kubeadm-join.go
```diff
@@ -0,0 +1,102 @@
...
+	if err := node.Command(
+		"kubeadm", "join",
+		// the control plane address uses the docker ip and a well know APIServerPort that
+		// are accessible only inside the docker network
+		fmt.Sprintf("%s:%d", controlPlaneIP, kubeadm.APIServerPort),
+		// uses a well known token and skipping ca certification for automating TLS bootstrap process
+		"--token", kubeadm.Token,
+		"--discovery-token-unsafe-skip-ca-verification",
+		// preflight errors are expected, in particular for swap being enabled
+		// TODO(bentheelder): limit the set of acceptable errors
```
neolit123: is this code (and code like this) copied from another location but is not removed from that original location in the diff?

pkg/cluster/nodes/create.go
```diff
@@ -97,7 +128,19 @@ func CreateControlPlaneNode(name, image, clusterLabel string) (handle *Node, por
 		}
 	}
 	if err != nil {
-		return handle, 0, errors.Wrap(err, "docker run error")
+		return handle, errors.Wrap(err, "docker run error")
+	}
+
+	// Deletes the machine-id embedded in the node imaga and regenerate a new one.
```
neolit123: `imaga` typo.

pkg/cluster/actions.go
```diff
@@ -0,0 +1,231 @@
...
+
+// Action define a set of tasks to be executed on a `kind` cluster.
+// Usage of actions allows to define repetitive, high level abstractions/workflows
+// by composing lower level tasks
+type Action interface {
+	// Tasks returns the list of task that are identified by this action
+	// Please note that the order of task is important, and it will be
+	// respected during execution
+	Tasks() []Task
+}
```
neolit123: i guess the list of actions is not exposed on the user side yet - i.e. we expose the Node list in the config separated by `---`, but i didn't see a way to also provide a list of actions?

neolit123:
might have been better if we have a discussion over Zoom about this first.

pkg/cluster/actions.go
```diff
@@ -0,0 +1,231 @@
...
+// It is required for making ExecutionPlan sortable.
+func (t ExecutionPlan) Len() int {
+	return len(t)
+}
+
+// Less return the lower between two elements of the ExecutionPlan, where the
+// lower element should be executed before the other.
+// It is required for making ExecutionPlan sortable.
+func (t ExecutionPlan) Less(i, j int) bool {
+	return t[i].ExecutionOrder() < t[j].ExecutionOrder()
```
ereslibre: Isn't it a bit strange to relay on string sorting here given that `ExecutionOrder()` returns a `string`? Could we have a utility function that takes into account the node `ProvisioningOrder()` and the `actionIndex` and `taskIndex`? https://play.golang.org/p/z8ZGpyl_EPt

Maybe it's minor as long as the order is always preserved.

cmd/kind/create/cluster/createcluster.go
```diff
@@ -74,10 +74,33 @@ func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
...
+		return fmt.Errorf("multi node support is still a work in progress, currently external etcd node is not supported")
+	}
+
 	// create a cluster context and create the cluster
 	ctx := cluster.NewContext(flags.Name)
 	if flags.ImageName != "" {
-		cfg.Image = flags.ImageName
+		// Apply image override to all the Nodes defined in Config
+		// TODO(fabrizio pandini): this should be reconsidered when implementing
+		//     https://github.com/kubernetes-sigs/kind/issues/133
```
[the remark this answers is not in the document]
fabriziopandini: Custom images per nodes can be achieved without using --image override.
I agree that upgrades should be kept simple.

pkg/cluster/actions.go
```diff
@@ -0,0 +1,231 @@
...
+// At exec time the logical step will then apply to the current cluster
+// topology, and be planned for execution zero, one or many times accordingly.
+type Task struct {
+	// Description of the task
+	Description string
+	// TargetNodes define a function that identifies the nodes where this
+	// task should be executed
+	TargetNodes NodeSelector
+	// Run the func that implements the task action
+	Run func(ec *execContext, configNode *config.Node) error
```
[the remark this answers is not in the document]
fabriziopandini: done

pkg/cluster/actions.go
fabriziopandini: ExecutionPlan is an ordered list, and the current execution model is sequential
If something blocks, the way to get out of this is ctrl+c (same as the current version of kind)

pkg/cluster/actions.go
fabriziopandini: @neolit123  The distinction between `Task` and `PlannedTask` is the key to manage variable list of actions in a way that can adapt to different cluster topologies

`Task` are defined at programming time - before knowing the cluster topology -, and they have a `NodeSelector` function, not a list of nodes.

`PlannedTask` is the "runtime version" of a `Task`, that is created when assigning a "logical task" to a Node in the actual cluster topology (a node that matches `NodeSelector` criteria).

pkg/cluster/actions.go
fabriziopandini: > why do we even need the sorting?
> what if we want to execute tasks with an order not based on the node names?

Sorting is used to provide a predictable, "kubeadm friendly" and consistent task exection order that can automatically adapt to any requested list of actions and to any cluster topology supported by kind. 

Such order is not based on the node name only, but on a combination of 4 attributes considered with a well known priority: node provisioning order, node name, action order, task order (I improved comments to make this more clear).

By combining above parameters you get a task execution order consistent with all the major kubeadm workflows (init, join control-plane, join) or (upgrade apply, upgrade node control-plane, upgrade node).

See discussion below with @ereslibre about implementation details, but in any case the ordering is 100% deterministic and it is already covered by unit tests signal.

As noted in `NewExecutionPlan` comments, I already know that this won't be enough for complex CI workflows, but this is a good step in that direction and this is why I'm putting order into this PR.

pkg/cluster/kubeadm/const.go
fabriziopandini: I used the same convention used for `APIServerPort` (without default).

pkg/cluster/kubeadm-join.go
fabriziopandini: This code is intentionally copied here because it is the same problem/TODO already defined for init and they should be addressed with the same solution

But I agree, the diff for context.go looks ugly and doesn't help...

pkg/cluster/nodes/create.go
fabriziopandini: Fixed

pkg/cluster/actions.go
fabriziopandini: You are right. UX around actions is not part of this PR.

With regards to exposing actions via Config/Api objects in a future, nothing in this PR prevent this, but IMO the decision mostly depends by the answer to this thread https://github.com/kubernetes-sigs/kind/issues/161#issuecomment-445802956.

pkg/cluster/actions.go
fabriziopandini: > the order is always preserved

The ordering is 100% deterministic, and this is covered by unit tests as well

Nevertheless, I'm ready to change implementation if this one seems counter-intuitive, but I'm not 100% sure the alternative reads better. see https://gist.github.com/fabriziopandini/e544842b8c38c90974b215167a77d934

fabriziopandini:
@neolit123 @ereslibre thanks for the round of feedbacks!
Some of them are now addressed, and answers are provided for the others.

@neolit123 of course ok for me to discuss this in slack; might be we want to wait for next week to get more people involved, but up to you

neolit123:
@fabriziopandini 
> might be we want to wait for next week to get more people involved, but up to you

SGTM. thanks for the replies.

pkg/cluster/actions.go
BenTheElder: I'm going to simplify actions quite a bit in a follow-up O(very soon)

BenTheElder:
xref: https://github.com/kubernetes-sigs/kind/pull/147#issuecomment-452444418
/hold cancel

BenTheElder:
I've rebased this to fix merge conflicts, in the process I did break controlplane meta, but will fix that after this merges in the general bikeshedding that will ensue :upside_down_face:

Issue #156, opened 2018-12-06
kubectl cannot communicate with new cluster


xstevens:
I was trying out KIND on macOS. The cluster seems to startup fine and there's no indication of errors in the logs. However, kubectl cannot communicate with the server using the generated kubeconfig. 

```
$ kubectl cluster-info

To further debug and diagnose cluster problems, use 'kubectl cluster-info dump'.
Unable to connect to the server: EOF
```

Here's a log of shell commands:

https://gist.github.com/xstevens/f43236505e7849d4079f8204c4a3f52b

BenTheElder:
`kind export logs` will give us logs to look at, it does look like the cluster came up fine initially at least from what you posted.

/assign

BenTheElder:
how much RAM does your docker daemon have configured?

docker for mac being low on RAM or disk space in the hyperkit vm is usually why something like this happens, that's something we need to cover more in the docs and maybe try to detect.

xstevens:
I killed the previous cluster and tried 4 CPU / 4GB RAM for Docker and ended with the same result.
```
$ docker ps
CONTAINER ID        IMAGE                  COMMAND                  CREATED              STATUS              PORTS                      NAMES
c0261f7512fd        kindest/node:v1.12.2   "/usr/local/bin/entr…"   About a minute ago   Up About a minute   0.0.0.0:64907->64907/tcp   kind-1-control-plane
$ docker exec -it c0261f7512fd /bin/sh
# docker ps -a
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
#
```

Attaching the export logs.
[kubelet.log](https://github.com/kubernetes-sigs/kind/files/2655457/kubelet.log)
[journal.log](https://github.com/kubernetes-sigs/kind/files/2655458/journal.log)
[kubernetes-version.txt](https://github.com/kubernetes-sigs/kind/files/2655459/kubernetes-version.txt)
[docker.log](https://github.com/kubernetes-sigs/kind/files/2655460/docker.log)
[docker-info.txt](https://github.com/kubernetes-sigs/kind/files/2655461/docker-info.txt)

BenTheElder:
Thanks! looks like disk space from the kubelet log:
```
Dec 07 00:37:53 kind-1-control-plane kubelet[688]: I1207 00:37:53.229561     688 eviction_manager.go:340] eviction manager: must evict pod(s) to reclaim ephemeral-storage
[...]

PR #163, opened 2018-12-10
remove duplicate word

tao12345666333:
Signed-off-by: Jintao Zhang <zhangjintao9020@gmail.com>

PR #165, opened 2018-12-10
Add system requirements for macos/windows

alejandrox1:
ref: #156

/kind documentation

### DIFF
--- a/docs/user/README.md
+++ b/docs/user/README.md
@@ -18,6 +18,18 @@ By default, the cluster will be given the name `kind-1`. "1," here, is the
 default context name. 
 Use the `--name` flag to assign the cluster a different context name.
 
+**Note**: If you are running `kind` on MacOS or Windows then it is recommended
+that you have at least 4GB of RAM and disk space (these are estimates for a
+single node `kind` cluster) dedicated to the virtual machine (VM) running the
+Docker engine otherwise the Kubernetes cluster might fail to start up.
+
+To change the resource limits for the Docker engine you'll need to open the
+**Preferences** menu, go to the **Advanced** settings page, and change the
+settings there, see [changing Docker's resource limits][Docker resource lims].
+
+You may also try removing any unused data left by the Docker engine - e.g.,
+`docker system prune`.
+
 
 ## Interacting With Your Cluster
 After [creating a cluster](#creating-a-cluster), you can use [kubectl][kubectl]
@@ -179,3 +191,4 @@ The logs contain information about the Docker host, the containers running
 [kubernetes]: https://github.com/kubernetes/kubernetes
 [kindest/node]: https://hub.docker.com/r/kindest/node/
 [kubectl]: https://kubernetes.io/docs/reference/kubectl/overview/
+[Docker resource lims]: https://docs.docker.com/docker-for-mac/#advanced


docs/user/README.md
```diff
@@ -18,6 +18,14 @@ By default, the cluster will be given the name `kind-1`. "1," here, is the
 default context name. 
 Use the `--name` flag to assign the cluster a different context name.
 
+**Note**: If you are running `kind` of MacOS or Windows then it is recommended
```
neolit123: `recommended that`

docs/user/README.md
```diff
@@ -18,6 +18,14 @@ By default, the cluster will be given the name `kind-1`. "1," here, is the
...
+you have at least 4GB of RAM and disk space dedicated to the virtual machine 
+(vm) running the Docker engine otherwise the Kubernetes cluster might fail to
```
neolit123: `vm` should be `VM`

docs/user/README.md
```diff
@@ -18,6 +18,14 @@ By default, the cluster will be given the name `kind-1`. "1," here, is the
 default context name. 
 Use the `--name` flag to assign the cluster a different context name.
 
+**Note**: If you are running `kind` of MacOS or Windows then it is recommended
+you have at least 4GB of RAM and disk space dedicated to the virtual machine
```
neolit123: what are we basing the estimate here - 4GB?
is this the minimum for a single node setup?

also how much disk space do we require?

docs/user/README.md
```diff
@@ -87,7 +95,7 @@ The `node-image` in turn is built off the [`base-image`][base image], which
 installs all the dependencies needed for Docker and Kubernetes to run in a
 container.
 
-See [base image](#base-image), for more advanced information information.
+See [base image](#base-image), for more advanced information.
```
neolit123: there is a separate PR that fixes this:
https://github.com/kubernetes-sigs/kind/pull/163

docs/user/README.md
```diff
@@ -18,6 +18,14 @@ By default, the cluster will be given the name `kind-1`. "1," here, is the
 default context name. 
 Use the `--name` flag to assign the cluster a different context name.
 
+**Note**: If you are running `kind` of MacOS or Windows then it is recommended
+you have at least 4GB of RAM and disk space dedicated to the virtual machine 
+(vm) running the Docker engine otherwise the Kubernetes cluster might fail to 
+start up.
+If you come across erros where `kubectl` cannot communicate with the cluster
+and you have increased the memory and size of the VM, then try cleaning up
+space by performing a `docker system prune`.
```
neolit123: i think this can be changed to:
> If you've experienced errors and had to increase the memory and disk space of the VM, don't forget to "prune" your container runtime - e.g. `docker system prune`.

docs/user/README.md
alejandrox1: 4 GB of ram seemed to work to resolve #165 and that's what the Docker installation instructions for macos mention.
I did add that the estimate assumes a single-node cluster.

docs/user/README.md
alejandrox1: I fixed the wording somewhat. Does one "prune a container runtime"? 
To prune a container runtime makes me understand that there is something to change from the way a container is run.

alejandrox1:
Thank you for all your comments @neolit123

docs/user/README.md
neolit123: 4GB RAM and another 4GB of disk space?
isn't 4GB of disk too low?

docs/user/README.md
```diff
@@ -18,6 +18,13 @@ By default, the cluster will be given the name `kind-1`. "1," here, is the
 default context name. 
 Use the `--name` flag to assign the cluster a different context name.
 
+**Note**: If you are running `kind` of MacOS or Windows then it is recommended
+that you have at least 4GB of RAM and disk space (these are estimates for a
```
BenTheElder: We should mention or link to how to do this. Many users have never needed to touch this.

alejandrox1: that's a good point. Added it just below. Tell me what you think

neolit123:
i don't think we allow merging merge commits as part of PRs. :)
i think you should rebase / squash them.

alejandrox1:
Thanks for catching that. Did the resolve through the ui and forgot that was the default

docs/user/README.md
```diff
@@ -18,6 +18,18 @@ By default, the cluster will be given the name `kind-1`. "1," here, is the
 default context name. 
 Use the `--name` flag to assign the cluster a different context name.
 
+**Note**: If you are running `kind` of MacOS or Windows then it is recommended
```
BenTheElder: s/of/on

PR #173, opened 2018-12-15
README.md: invite reaching out

BenTheElder:
... to the maintainers

I read back through the readme tonight, and this jumped out ... I think this is a bit more inviting, which is really what I wanted out of this line. not sure how it got left in the previous state

/cc @munnerz


BenTheElder:
??
/test pull-kind-verify

Issue #177, opened 2018-12-16
current kubernetes version CVE


traceypooh:
looks like v1.12.2 is being installed.  bump to v1.12.3 or v1.13?
```
root@kind-1-control-plane:/# kubectl --kubeconfig /etc/kubernetes/admin.conf  get pods,deployments,services,replicasets,nodes --all-namespaces
NAMESPACE     NAME                                               READY   STATUS    RESTARTS   AGE
kube-system   pod/coredns-576cbf47c7-2c48t                       1/1     Running   0          8m4s
kube-system   pod/coredns-576cbf47c7-krfsv                       1/1     Running   0          8m4s
kube-system   pod/etcd-kind-1-control-plane                      1/1     Running   0          8m58s
kube-system   pod/kube-apiserver-kind-1-control-plane            1/1     Running   0          8m34s
kube-system   pod/kube-controller-manager-kind-1-control-plane   1/1     Running   1          9m
kube-system   pod/kube-proxy-ksxf9                               1/1     Running   0          8m5s
kube-system   pod/kube-scheduler-kind-1-control-plane            1/1     Running   0          8m50s
kube-system   pod/weave-net-ncs9s                                2/2     Running   0          8m5s

NAMESPACE     NAME                            DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
kube-system   deployment.extensions/coredns   2         2         2            2           8m47s

NAMESPACE     NAME                 TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)         AGE
default       service/kubernetes   ClusterIP   10.96.0.1    <none>        443/TCP         9m41s
kube-system   service/kube-dns     ClusterIP   10.96.0.10   <none>        53/UDP,53/TCP   8m47s

NAMESPACE     NAME                                       DESIRED   CURRENT   READY   AGE
kube-system   replicaset.extensions/coredns-576cbf47c7   2         2         2       8m5s

NAMESPACE   NAME                        STATUS   ROLES    AGE     VERSION
            node/kind-1-control-plane   Ready    master   9m42s   v1.12.2
roo```

BenTheElder:
`kind` is also intended to be used with locally built images against arbitrary v1.11+ Kubernetes, and the default images provided were initially meant to be a convenience, but with emerging use cases besides the local development of Kubernetes mean we should be on top of this better!

working on an image update now.

/cc @munnerz - I think this is our first use case for release branches and a cherry pick - just the image bump

[...]

PR #179, opened 2018-12-16
upgrade to v1.12.3

BenTheElder:
addresses #177, I've just built and pushed this new default image for v1.12.3

/priority critical-urgent
/kind bug


BenTheElder:
ignoring e2e test results since the only change is the default image and these are against k8s releases. we will have to revisit fighting these upstream flakes, going to just merge this one myself for now

PR #183, opened 2018-12-17
add Dockerfile

zoobab:
add Dockerfile. Builds the binary in /go/bin/kind.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,4 +2,5 @@ FROM golang:1.11.3-stretch
 COPY . /go/src/kind
 WORKDIR /go/src/kind
 RUN go get .
-RUN CGO_ENABLED=0 GOOS=linux go install -a -ldflags '-extldflags "-static"' .
+RUN CGO_ENABLED=0 GOOS=linux go install -a -ldflags '-s -w -extldflags "-static"' .
+ENTRYPOINT ["/bin/cp", "-v", "/go/bin/kind", "/out"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,5 @@
+FROM golang:1.11.3-stretch
+COPY . /go/src/kind
+WORKDIR /go/src/kind
+RUN go get .
+RUN CGO_ENABLED=0 GOOS=linux go install -a -ldflags '-extldflags "-static"' .
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+set -e
+
+banner() {
+echo "================================================================"
+echo "$1"
+echo "================================================================"
+}
+
+PROJECT="kind"
+
+banner "Building docker image "$PROJECT" ..."
+docker build -t $PROJECT .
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM golang:1.11.3-stretch
+COPY . /go/src/kind
+WORKDIR /go/src/kind
+RUN go get .
+RUN CGO_ENABLED=0 GOOS=linux go install -a -ldflags '-extldflags "-static"' .
```
neolit123: if we are adding a Docker file, we might as well `go get sigs.k8s.io/kind` from inside the container and allow the user to specify a kind tag.
also add docs on how to `docker run` and mount a volume where the binary will be built.

with this current Docker file it is expected from the user to have already downloaded the project source code "somehow".

BenTheElder: Besides those suggestions, this assumes Linux, while the project supports Mac and is working on Windows. Ideally we would allow cross compiling.

I'd also rather not add more files to the root of the project, this should go in something like hack/build with instructions for using it.

alejandrox1: @zoobab one detail that's missing is how what you are suggesting will help make getting started with kind easier than simply downloading a binary for a given architecture?

zoobab: I will add a build.sh script and change the ENTRYPOINT to copy the binary. I don't know how those released binaries are produced/compiled, but my experience with Github releases is that it comes from the dev laptop, or travis-ci integration at best. But a simple Dockerfile outputting a binary, with multiple devs obtaining the same md5sum at the end should be a better release process IMHO.

zoobab: I will add the same entrypoint and build.sh as I did here: https://github.com/zoobab/terraform-provider-infoblox/blob/Dockerfile/Dockerfile and https://github.com/zoobab/terraform-provider-infoblox/blob/Dockerfile/build.sh .

Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM golang:1.11.3-stretch
+COPY . /go/src/kind
+WORKDIR /go/src/kind
```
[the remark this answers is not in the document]
zoobab: What is the right GOPATH then?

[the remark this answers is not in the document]
BenTheElder: we are `sigs.k8s.io/kind`

Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM golang:1.11.3-stretch
```
BenTheElder: This needs to be 1.11.4 at minimum (ideally newer)

I'd rather put these in a hack/ subdirectory than clutter the root directory further if we can avoid it, I don't expect most users will use this.
(kind / kubernetes developers will likely have go installed, end users will either `go get` install for convenience, or get a pre-built release binary).

BenTheElder:
you can now do a canonical build of `kind` with `make` (just requires docker), the build is cached etc. but technically doesn't require a dockerfile currently.

PR #184, opened 2018-12-17
Cleanup the generated kube config

blakestoddard:
This came out of a [discussion in Slack](https://kubernetes.slack.com/archives/CEKK1KTN2/p1545074247192800) when I had forgotten that I had set `$KUBECONFIG` to point to my `kind` config and couldn't figure out how to get back to the rest of my previously-defined contexts.

This PR simply removes the generated config on `kind delete`, checks if `$KUBECONFIG` equals the path of the (just removed) generated config, and if so, tells them to unset it to get back to normal.


blakestoddard:
I had my work email on my CLA profile and my personal email on my commit, let's see if it's appeased now.  ¯\_(ツ)_/¯

pkg/cluster/context.go
```diff
@@ -178,6 +178,15 @@ func (c *Context) Delete() error {
 	if err != nil {
 		return fmt.Errorf("error listing nodes: %v", err)
 	}
+
+	// clean up the kind kube config file that was created earlier
+	// also let the user know that they need to unset $KUBECONFIG to
+	// go back to their normal contexts
+	os.Remove(c.KubeConfigPath())
```
neolit123: i will leave it to @BenTheElder for the decision on cleanup here.
my vote leans towards not doing it by default, but having an explicit CLI flag for it.

pkg/cluster/context.go
```diff
@@ -178,6 +178,15 @@ func (c *Context) Delete() error {
...
+	if os.Getenv("KUBECONFIG") == c.KubeConfigPath() {
+		fmt.Printf("kind kubectl config file removed, but you will need to unset KUBECONFIG to return to your normal contexts:\n\nunset KUBECONFIG\n")
```
neolit123: `kind kubectl config file removed`
^ happens outside of the `if os.Getenv("KUBECONFIG") == c.KubeConfigPath() {` branch so ideally there should be two `fmt.Printf/ln()`.

- one about removing the file.
- one for letting the user they have to now unset `KUBECONFIG`.

pkg/cluster/context.go
BenTheElder: we should probably handle the error here with a warning at least

pkg/cluster/context.go
```diff
@@ -178,6 +178,15 @@ func (c *Context) Delete() error {
 	if err != nil {
 		return fmt.Errorf("error listing nodes: %v", err)
 	}
+
+	// clean up the kind kube config file that was created earlier
+	// also let the user know that they need to unset $KUBECONFIG to
+	// go back to their normal contexts
```
alejandrox1: I think this could be changed to something like:

> // Remove the kind kube config file created by "kind create cluster."
> // If $KUBECONFIG was set to the kubeconfig file created by kind, then remind user to unset its value.

and 
```go
fmt.Println("kind kubectl config file was removed.\nRemeber to unset $KUBECONFIG, if set")
```

pkg/cluster/context.go
BenTheElder: actually it should gracefully drop back to the default location if kubeconfig contains no files.

we also should really suggest the correct thing for windows as well, maybe just leave a todo here 🤔

neolit123: for win32:
`set KUBECONFIG=`

instead of `runtime.GOOS` abstraction we might as well just instruct:
>set the KUBECONFIG with value X`

>unset the KUBECONFIG`

pkg/cluster/context.go
alejandrox1: sounds interesting, why do you think it shouldn't be done by default?

pkg/cluster/context.go
BenTheElder: I'm not even sure we should mention this, does eg minikube? the workings of KUBECONFIG seem to be something orthogonal to learn vs kind 🤔

pkg/cluster/context.go
```diff
@@ -178,6 +178,18 @@ func (c *Context) Delete() error {
 	if err != nil {
 		return fmt.Errorf("error listing nodes: %v", err)
 	}
+
+	// try to remove the kind kube config file generated by "kind create cluster"
+	err = os.Remove(c.KubeConfigPath())
+	if err != nil {
+		fmt.Printf("Tried to remove %s but recieved error: %s\n", c.KubeConfigPath(), err)
```
alejandrox1: I think it this should be logged instead of just printed.

pkg/cluster/context.go
neolit123: minikube only has it in the docs, while kubeadm prints the stdout linux mini-guide on how to setup kubeconfig.

pkg/cluster/context.go
blakestoddard: Yeah, I think you're right. Will fix.

blakestoddard: What level tho, warning?

pkg/cluster/context.go
BenTheElder: maybe let's improve to a miniguide in a follow-up and handle windows too for set and unset

pkg/cluster/context.go
```diff
@@ -178,6 +178,18 @@ func (c *Context) Delete() error {
...
+	}
+
+	// check if $KUBECONFIG is set and let the user know to unset if so
+	if os.Getenv("KUBECONFIG") == c.KubeConfigPath() {
+		fmt.Printf("$KUBECONFIG is still set to use %s even though that file has been deleted, remember to unset it to get back to your original contexts\n", c.KubeConfigPath())
```
alejandrox1: Also, may be get rid of a the part mentioning contexts. 
To my understanding, a context is a combination of access parameters - it
Takes into account user accounts and namespaces as well, not just clusters.

pkg/cluster/context.go
BenTheElder: Warning 👍

alejandrox1: I would think it deserves an error

pkg/cluster/context.go
BenTheElder: yeah, contexts is orthogonal to entire config files, also if your original contexts were in the default file you don't even need to do this, and if they weren't, unsetting it won't be sufficient, you'll need to re-set it to wherever those were

pkg/cluster/context.go
alejandrox1: What Ben said lol

BenTheElder: I think cleanup failures are generally not treated as critical in kind, but the user / logging should be informed. we can revisit it later if usage proves that to be incorrect.

pkg/cluster/context.go
blakestoddard: Are you in favor of erring on the side of verbosity and saying "remember to unset it to use the default kube config, or set it back to your original value", or just saying "remember to unset it"?

BenTheElder: I think just "remember to unset it", we should probably discuss in a follow-up how to improve the message from there. 
EG handling windows is probably something we _should_ do but more complex than the first PR needs to be 😅 

We should look around at other tools for guidance 🙃

pkg/cluster/context.go
```diff
@@ -178,6 +178,18 @@ func (c *Context) Delete() error {
...
 	}
+
+	// try to remove the kind kube config file generated by "kind create cluster"
+	err = os.Remove(c.KubeConfigPath())
+	if err != nil {
+		log.Warningf("Tried to remove %s but recieved error: %s\n", c.KubeConfigPath(), err)
+	}
+
+	// check if $KUBECONFIG is set and let the user know to unset if so
+	if os.Getenv("KUBECONFIG") == c.KubeConfigPath() {
```
BenTheElder: note to self: a more complete version of this will need to consider how to handle multiple files, which KUBECONFIG / kubectl support.

PR #187, opened 2018-12-19
Adding some meta information to the 2 contexts for Create

### LINKED ISSUE #186: Return meta information about the cluster
Right now the call to `context.Create` only returns a possible `error`

```go
func (c *Context) Create(cfg *config.Config, retain bool, wait time.Duration) error {
     // ...
}
```

If you drill down in to the logic contained in the method you find that we eventually reach `provisionControlPlane`

```go
// provisionControlPlane provisions the control plane node
// and the cluster kubeadm config
func (cc *createContext) provisionControlPlane(
	nodeName string,
) (kubeadmConfigPath string, err error) {
     // ...
}
```

In the `provisionControlPlane` method there is valuable information being lost such as the ephemeral port that the container has listening for the API server.

If using KIND to write a unit test, being able to define the port to test on is critical. 

Can I please PR a feature that allows for this valuable information to be captured from a calling piece of code?

I think we can do it by adding a new return type to `context.Create` or caching the information at the package level and pulling from the cache. 

Would like to use something like this for TGIK on Friday.

Thanks for the great project!

kris-nova:
Closes #186

### DIFF
--- a/pkg/cluster/context.go
+++ b/pkg/cluster/context.go
@@ -43,6 +43,7 @@ import (
 // Context is used to create / manipulate kubernetes-in-docker clusters
 type Context struct {
 	name string
+	ControlPlaneMeta *ControlPlaneMeta
 }
 
 // createContext is a superset of Context used by helpers for Context.Create()
@@ -52,6 +53,7 @@ type createContext struct {
 	config       *config.Config
 	retain       bool          // if we should retain nodes after failing to create.
 	waitForReady time.Duration // Wait for the control plane node to be ready.
+	ControlPlaneMeta *ControlPlaneMeta
 }
 
 // similar to valid docker container names, but since we will prefix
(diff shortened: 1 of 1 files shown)


pkg/cluster/context.go
```diff
@@ -42,16 +42,18 @@ import (
 
 // Context is used to create / manipulate kubernetes-in-docker clusters
 type Context struct {
-	name string
+	name             string
+	ControlPlaneMeta *ControlPlaneMeta
```
neolit123: in the long run, we might want to treat the Context object as an API and have public getters and setters instead of public fields.

BenTheElder: I think I'd like to make this private with a cached "getter" later so that it works for contexts referring to an existing cluster. 

I'm happy to file a follow-up PR to work on this though. There's a ~similar pattern internally for `Node.KubernetesVersion()`

kris-nova: If you have strong opinions I would vote we follow up with a second PR?

BenTheElder: SGTM

BenTheElder:
Thanks again :smile:

Issue #176, opened 2018-12-16
Introduce a logging shim


BenTheElder:
related to #23, but distinct - we should introduce a minimal logging shim interface / package, with a default instance. This will let consumers of the libraries shim the logging to their preferred package or even just dev/null it. Same goes for the status output.

We don't need anything fancy here, probably just a couple of levels like Info, Debug, Warning, Error with printf ... we can shim it to logrus for now to minimize the changes and improve on it over time.

cc @vincepri @liztio I think this will be a big improvement for library consumption

Also I know _everyone_ has a logging package, but that's exactly why I want a tiny shim to whatever package we actually use underneath - until there's an agreed upon interface for this we should make the smallest one reasonable so library consumers can swap in their own log impl.

/priority important-soon
/kind feature

vincepri:
This use case might be covered if we decided to switch to [klog](https://github.com/kubernetes/klog/) which supports `logr` and others output.

BenTheElder:
We actually have an open PR / issue for switching to klog but ran into some issues with that. I think maybe kind can actually provide an even smaller and more easily replaced interface, and I'd also like to dissallow things like `klog.Fatal()` from entering the codebase. Our code should not cause applications to exit.

I'm also not super convinced that these frameworks are suited as well to CLI vs servers, but I'd happily switch to klog / logr or a small shim over them to get along with the rest of the ecosystem if we can get it working well. 👍

vincepri:
I think `klog` might be valuable especially when using `kind` as a library. I agree that Fatal() or Panic() shouldn't be allowed. That could be disallowed in CI or through PRs / contribution guidelines.

BenTheElder:
this is done!

PR #192, opened 2018-12-20
logging: use a custom logger interface

neolit123:
The pkg namespace should not use an external top level logger.
This can cause problems if multiple packages such as kind/pkg are vendored
in a third party project and each one of them uses a custom top level
logger.

Arguably top level loggers such as klog and logrus should only be exposed
on the cmd side.

Introduce a Logger interface in kind/pkg/log and implement an object
that can be used for logging in kind/cmd, but also in kind/pkg. Currently the object is of type
logrus.Logger.

/kind cleanup
/priority important-longterm
/assign @BenTheElder @munnerz 

@BenTheElder this is my view of the discussion yesterday.
random nodes:
- WithError() and WithField() etc, were replaced, because i didn't want to implement a generic interface that supports this functionality for all possible loggers.

xref: https://github.com/kubernetes-sigs/kind/issues/176


pkg/log/log.go
```diff
@@ -0,0 +1,107 @@
...
+package log
+
+import (
+	"io"
+)
+
+// Logger is a interface that should be used for all logging inside kind/pkg.
+// Consumers of kind/pkg such as kind/cmd should define an object that satisfies
+// this interface.
+type Logger interface {
```
BenTheElder: have we looked at how well we can shim this to klog? I think glog / klog is the biggest secondary target to look at. I _think_ this interface should shim well.

pkg/log/log.go
```diff
@@ -0,0 +1,107 @@
...
+	Infof(string, ...interface{})
+
+	Debug(...interface{})
+	Debugf(string, ...interface{})
+}
+
+var logger Logger
+
+// GetLogger returns the locally defined logger that was previously set using SetLogger().
+func GetLogger() Logger {
```
BenTheElder: unless we're wrapping this, maybe we should just make the variable public? /shrug

pkg/log/log.go
neolit123: the klog Verbose type would satisfy it partially, but would still need some wrapping for the Output() and SetOutput() methods.
https://github.com/kubernetes/klog/blob/master/klog.go#L711

also even if we have to wrap every single method for the kind Logger interface the compiler will pretty much inline the calls.

glog is a super weird implementation.

pkg/log/log.go
neolit123: public var is also OK.
the only benefits of these getter/setting here is that they leave the option to do extra work eventually without breaking the user.

pkg/log/log.go
BenTheElder: +1 to that, glog is super weird :-)

pkg/log/log.go
```diff
@@ -0,0 +1,107 @@
...
+	logger = l
+}
+
+// Output returns the defined io.Writer where the kind/pkg logger writes everything.
+func Output() io.Writer {
+	return logger.Output()
+}
+
+// SetOutput sets where the kind/pkg logger writes everything.
+// It makes a local copy but also assumes that the Logger based object has a the same method.
```
neolit123: need to fix this comment. there is no local copy anymore.

pkg/log/log.go
BenTheElder: SGTM

BenTheElder:
> WithError() and WithField() etc, were replaced, because i didn't want to implement a generic interface that supports this functionality for all possible loggers.

Yeah, we should just stop using those, I'm pretty convinced at this point that structured logs do not help for CLI.

pkg/log/log.go
```diff
@@ -0,0 +1,107 @@
...
+// Package log contains logging related functionality
+package log
+
+import (
+	"io"
+)
+
+// Logger is a interface that should be used for all logging inside kind/pkg.
+// Consumers of kind/pkg such as kind/cmd should define an object that satisfies
+// this interface.
```
BenTheElder: Let's also add a TODO to consider klog / `logr` when it has more traction. (EG logr support in klog... ) @munnerz was working on this, but I don't think it will be ready for a while.

cmd/kind/kind.go
```diff
@@ -74,13 +75,13 @@ func NewCommand() *cobra.Command {
 
 func runE(flags *Flags, cmd *cobra.Command, args []string) error {
 	level := defaultLevel
-	parsed, err := log.ParseLevel(flags.LogLevel)
+	parsed, err := logrus.ParseLevel(flags.LogLevel)
```
philoserf: This continues to expose logrus directly. Shouldn't that be an implementation detail of pkg/log and abstracted here?

philoserf: ack

BenTheElder:
thanks! I think this is good, i'm just prioritizing some other refactors
first to give it a little more thought & discussion :sweat_smile:

neolit123:
ping @munnerz @BenTheElder 
let me know if we can green light this idea or close at this point.

neolit123:
added xref to https://github.com/kubernetes-sigs/kind/issues/176
in the OP.

recent discussion at the k-dev mailing list started leaning towards a wider usage of the new logr implementation called klogr.

klogr will certainly replace this PR.

BenTheElder:
Let's come back to this but keep it in mind for reference 🙏 

thanks again, keeping an eye on klogr.

PR #193, opened 2018-12-25
refactor(trivial): use the logrus import that was first

philoserf:
In commit 33696bd22fe80982e38fb92c5f3ca7f0d5690ce6 by @BenTheElder;

> Add status tracking with loading spinners, clean up logging

we got a second import of logrus. We do'nt need two. Keep the one that
was already there.

trivial

Signed-off-by: Mark Ayers <mark@philoserf.com>


BenTheElder:
Thanks for the PR! Seems like #192 will indeed work as well, but I appreciate this PR :-)

PR #198, opened 2019-01-04
New argument for 'build node-image' for passing the kuberoot

inercia:
This allows users to build their images with something like `kind build node-image --kuberoot $GOPATH/src/github.com/user/kubernetes`.

### DIFF
--- a/cmd/kind/build/nodeimage/nodeimage.go
+++ b/cmd/kind/build/nodeimage/nodeimage.go
@@ -29,6 +29,7 @@ type flagpole struct {
 	BuildType string
 	Image     string
 	BaseImage string
+	KubeRoot  string
 }
 
 // NewCommand returns a new cobra.Command for building the node image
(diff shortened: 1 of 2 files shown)


cmd/kind/build/nodeimage/nodeimage.go
```diff
@@ -29,6 +29,7 @@ type flagpole struct {
 	BuildType string
 	Image     string
 	BaseImage string
+	Kuberoot  string
```
neolit123: `KubeRoot` seems better here.

cmd/kind/build/nodeimage/nodeimage.go
```diff
@@ -52,6 +53,11 @@ func NewCommand() *cobra.Command {
 		node.DefaultImage,
 		"name:tag of the resulting image to be built",
 	)
+	cmd.Flags().StringVar(
+		&flags.Kuberoot, "kuberoot",
```
neolit123: possibly `kube-root`

cmd/kind/build/nodeimage/nodeimage.go
```diff
@@ -52,6 +53,11 @@ func NewCommand() *cobra.Command {
...
+		"",
+		"kubernetes sources root directory (default is autodetected)",
```
neolit123: > Path to the Kubernetes source directory (if empty, the path is autodetected)

pkg/build/node/node.go
```diff
@@ -72,6 +72,13 @@ func WithMode(mode string) Option {
 	}
 }
 
+// WithKuberoot sets the kubernetes sources root directory
```
neolit123: possibly adapt the comment and function name as per my previous comments.

Issue #99, opened 2018-11-11
Support exposing additional ports to be used for NodePort services


nilebox:
`kind` currently only exposes a single Docker port to the host machine (Kubernetes API server).
As a result, It is impossible to access other services running inside the cluster.

In `minikube` NodePort is a solution for that.
`minikube` exposes all ports on a VM's IP address:
```
❯ minikube service list
|----------------|-----------------------|--------------------------------|
|   NAMESPACE    |         NAME          |              URL               |
|----------------|-----------------------|--------------------------------|
| example        | example               | http://192.168.99.100:30980
```
that just print a VM IP address + port that you can query via
```
❯ kubectl get -n example service example -o wide
NAME       TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)                         AGE
example    NodePort    10.110.138.8     <none>        8080:30980/TCP,9090:30990/TCP   22m
```
This is also useful for testing Ingress controllers like Contour, https://github.com/heptio/contour/blob/master/docs/minikube.md

The trivial change I made to make it working is added extra flags to `docker run` in https://github.com/nilebox/kind/commit/99c4505c04b8b4893e2540ed9cab7aaeed9d9bf8.
After that I just declared a service with a `NodePort` type and a port from the exposed range, and I was able to access it via `curl` from the host machine.

Would it make sense to support such feature in `kind` upstream?
e.g. a flag for exposing extra ports to be passed to `docker run`.

BenTheElder:
Supporting extra ports makes sense, note though that you can access (TCP) services via `kubectl port-forward`:
https://kubernetes.io/docs/tasks/access-application-cluster/port-forward-access-application-cluster/
UDP is not yet supported by kubernetes for this though.

I think we'd need to gate exposing extra ports on the node container behind some config, `kind` nominally supports running multiple clusters per machine (`--name`) so we couldn't just bind those ports on all clusters by default as far as I know.

/kind feature

BenTheElder:
Er, and I'm on the fence as to if this should be a flag or be in config. I'd like to keep the flag surface from exploding, but it seems like this might be commonly used enough to put in both with the flag overriding.

cc @munnerz @neolit123

neolit123:
+1 for config only.

`ExposedPortRanges []strings`
per cluster.


[...]

Issue #62, opened 2018-10-09
Allow mounting host host directories to kind nodes


mitar:
It seems there is no way to request some host volumes to be mounted inside the kind container. So then pods cannot access those host volumes through `hostPath`. This is useful for us to bring testing data in.

BenTheElder:
definitely, this relates to #28, mostly we need to think about how best to expose this. cc @munnerz

BenTheElder:
There is not, that sort of thing would create some interesting stability guarantees, eg at some point I think we may want to move to a docker client library instead of `exec`ing `docker` at which point any option to hook it with arbitrary args would no longer work. 

I'd like to avoid breaking changes going forward as much as possible, so I'd like to think about how exactly we expose options like this. I think we can provide config for host path mounts specifically, and think a bit about how that relates to multi-node (which is forthcoming...)

 The closest work around right now is patching kind itself...

BenTheElder:
/priority backlog
I definitely intend to support this

BenTheElder:
/assign
/lifecycle active
I'll file a PR after https://github.com/kubernetes-sigs/kind/pull/77, also going to clean up the node lifecycle stuff a bit along with it.

fabriziopandini:
+1 noted

BenTheElder:
[Using this to track volumes as well IE #246]

Multi-node is pretty solidly landed, experimenting with this again currently. This is actually active again, apologies for the delay.

BenTheElder:
[apologies, turns out the config format for this is tricky and very coupled to docker ... stay tuned ...]

The docker client library is imo, even worse for this. Evaluating options to make this better than literally plumbing through extra user provided flags to the container creation.

mitar:
Looking forward to this. For now we have to use a fork to pass those arguments through.

BenTheElder:
I did some more looking yesterday, what we can do to start is mimic the format used by Kubernetes CRI to describe container mounts, which is of course quite reasonable.
/priority important-soon

BenTheElder:
We discussed this in today's meeting, triaging to the next minor release ideally, and following Kubernetes CRI for how to describe this.

BenTheElder:
Fixed in https://github.com/kubernetes-sigs/kind/pull/313

Quick example usage here: https://github.com/kubernetes-sigs/kind/pull/313#issuecomment-464967105

Will document more as we finish fleshing it out.

mitar:
Awesome! Thanks!

[...]

PR #199, opened 2019-01-05
Add parameters for node container

pablochacin:
What this PR does
---------------------------
Add a parameter to create cluster command for specifying
arguments to be passed to the node container run command.

This allows for example, exposing additional ports, or attaching
additional volumes. This is necessary to complement the ability
to create custom node images.

The 'args' parameter receives a string with the parameters
which are appended to the docker run command used for
launching the node container.

For example the command bellow allows mapping port 8000 
exposed by a custom node image.

`kind create cluster --args "-p 8000:8000`

How to test feature
---------------------------

Create cluster exposing port 2379 (etcd)
`>kind create cluster --args "--expose 2379 -p 2379:2379`

After the cluster has been create, check the port is actually exposed:
`> docker inspect kind-1-control-plane | jq '.[].HostConfig.PortBindings[][].HostPort' | grep -o 2379`
2379

Signed-off-by: Pablo Chacin <pchacin@suse.com>

### DIFF
--- a/cmd/kind/create/cluster/createcluster.go
+++ b/cmd/kind/create/cluster/createcluster.go
@@ -35,6 +35,7 @@ type flagpole struct {
 	ImageName string
 	Retain    bool
 	Wait      time.Duration
+	Args      string
 }
 
 // NewCommand returns a new cobra.Command for cluster creation
@@ -51,6 +52,7 @@ func NewCommand() *cobra.Command {
 	cmd.Flags().StringVar(&flags.Name, "name", "1", "cluster context name")
 	cmd.Flags().StringVar(&flags.Config, "config", "", "path to a kind config file")
 	cmd.Flags().StringVar(&flags.ImageName, "image", "", "node docker image to use for booting the cluster")
+	cmd.Flags().StringVar(&flags.Args, "args", "", "additional parameters for launching node container")
 	cmd.Flags().BoolVar(&flags.Retain, "retain", false, "retain nodes for debugging when cluster creation fails")
 	cmd.Flags().DurationVar(&flags.Wait, "wait", time.Duration(0), "Wait for control plane node to be ready (default 0s)")
 	return cmd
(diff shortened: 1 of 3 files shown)


pablochacin:
@BenTheElder Could you please elaborate on how this PR could make sense. I agree that maybe using more specific arguments like "--port" or "--volume" may better. Do you think this may fit?

Regarding the potential conflict with multi-node, can you please point-me to some specific issues you foresee or the code base for such feature such that I can check it myself?

All that said, I presently needing this feature in order to use Kind as a development environment for the [gardener project](https://github.com/gardener/gardener/blob/master/docs/development/local_setup.md#check-local-setup) and I'm planning to add kind as a supported environment.

BenTheElder:
Sorry for the delay - I focused on getting the initial-multi node in recently as we need it for testing, catching back up on these now 😓 

For multi-node and extra flags to docker when creating the nodes you likely want these to be different depending on the node, probably we should add some extra fields to the node configs to allow exposing extra ports and mounting extra volumes.

BenTheElder:
> All that said, I presently needing this feature in order to use Kind as a development environment for the gardener project and I'm planning to add kind as a supported environment.

Also, that sounds _awesome_ :-) Would love to help make that happen. 

Can we discuss how extra ports / volumes should work with others in issues first?
It is a high priority that the exposed API / flags stay very stable by the end of the k8s 1.14 release so we can ship something everyone can build on. 

I'd like to add something a little less brittle than directly plumbing through flags to docker (which also leaves room for EG #151)

For ports we can use issue #99, and for volumes issue #62 should work I think.

pablochacin:
Thanks for the feedback @BenTheElder 

I was waiting to see how #147  evolved as t has a significant impact on how to expose the node's container engine parameters. I'm now also looking at #151 but apparently, besides changing `docker` for `container.Engine` the structure of code, parameters and main logic remains the same as before. 

I agree that addressing specific uses cases as you propose in #99 and #62 is a much better approach.  Should we consider closing this PR then and move the conversation to those more specific cases?

BenTheElder:
I think that's the way to go, thanks for understanding 😅

BenTheElder:
closing for now, we will definitely aim to prioritize #99 and #62 in this k8s release cycle.

PR #203, opened 2019-01-08
Print version of Kind on stdout instead of stderr

nlamirault:


cmd/kind/version/version.go
```diff
@@ -31,9 +31,8 @@ func NewCommand() *cobra.Command {
 		Use:   "version",
 		Short: "prints the kind CLI version",
 		Long:  "prints the kind CLI version",
-		RunE: func(cmd *cobra.Command, args []string) error {
+		Run: func(cmd *cobra.Command, args []string) {
 			println(Version)
```
neolit123: something else here.

this PR does not change the stderr -> stdout for version, because `println` remains in use.
it only changes `RunE` -> `Run`.

`println` always prints to stderr.

to achieve what you want you need to use:
`fmt.Println()`

or `fmt.Fprintf(os.Stdout....`

cmd/kind/kind.go
```diff
@@ -51,8 +51,8 @@ func NewCommand() *cobra.Command {
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
 			return runE(flags, cmd, args)
 		},
-		RunE: func(cmd *cobra.Command, args []string) error {
-			return cmd.Help()
+		Run: func(cmd *cobra.Command, args []string) {
+			cmd.Help()
```
neolit123: like mentioned in my previous comment this only changes `RunE` -> `Run`.
the help screen is printed to stdout for `RunE` already.

neolit123:
@nlamirault 

looks like the Help() change is not needed.
you need to simply replace `println()` with `fmt.Println`.

BenTheElder:
TIL println goes to stderr in go.

Yes, please just switch that line to fmt.Println for now.

akutz:
Hi all,

Help output typically goes to stderr. I would leave things the way they are.

akutz:
Have you read https://www.jstorimer.com/blogs/workingwithcode/7766119-when-to-use-stderr-instead-of-stdout? When one invokes a help flag directly the output goes to stdout, otherwise help output should not be passed to the next program.

neolit123:
> When one invokes a help flag directly the output goes to stdout, otherwise help output should not be passed to the next program.

akutz:
I used to work on Cobra — It can be configured. It was changed to default to stdout at https://github.com/spf13/cobra/pull/71, but when printing usage due to an error it should switch to stderr.

akutz:
Also, FWIW, I prefer the variants of fmt.Fprint and selecting the writer explicitly — os.Stderr or os.Stdout. You can also use this to pass the configured usage output writer for Cobra. Which itself is backwards compatible with the flags package, and thus has a global, configured output stream.

cmd/kind/version/version.go
```diff
@@ -18,21 +18,24 @@ limitations under the License.
...
+
 	"github.com/spf13/cobra"
 )
 
 // Version is the kind CLI version
 const Version = "0.1.0-alpha"
 
 // NewCommand returns a new cobra.Command for version
-func NewCommand() *cobra.Command {
+func NewCommand(w io.Writer) *cobra.Command {
```
BenTheElder: cobra commands already have an io.Writer for output IIRC (?)

neolit123: > if one wants to print something custom they need to use an explicit stream.

actually...there are these wrappers.
https://godoc.org/github.com/spf13/cobra#Command.Println

feels like there are too many ways to customize this.
my vote would be to still print help and version both to stdout (one way or another).

BenTheElder: Yeah, I'm not sure that we should be passing stdout explicitly to this one command, there's probably a better pattern 🤔

I haven't had a chance to dig around and fight if there is a standard for which goes to what. Normally I'd use stdout and depending on the program use stderr for ... Errors 🤔

neolit123: kubeadm uses the pattern of passing the io.Writer around: https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/cmd/version.go#L55

but frankly we can just `fmt.Println()` here for now.

neolit123:
i think we should have this in the next minor.
@BenTheElder please outline if you demand changes here?

BenTheElder:
> i think we should have this in the next minor.
> @BenTheElder please outline if you demand changes here?

apologies, this slipped through my inbox somehow -- I'm OK with this but I've been waiting for clear evidence one way or another.

currently I'm actually looking at refactoring the commands to take a struct with dependencies so we can easily do things like feature gates or letting library users wrap the commands and inject i/o...

BenTheElder:
Done in https://github.com/kubernetes-sigs/kind/issues/533

Thanks for the PR.

PR #206, opened 2019-01-09
Restore multinode config

fabriziopandini:
@BenTheElder 
This PR fixes and rebases #164 and #147 toghether, so now you get a single config object with a list of nodes and actions for joining worker nodes


pkg/cluster/config/encoding/scheme.go
```diff
@@ -59,164 +52,39 @@ func AddToScheme(scheme *runtime.Scheme) {
...
-		}
-
-		// Build a GroupVersionKind object from the TypeMeta object
-		documentGroupVersion, err := schema.ParseGroupVersion(typeMetaInfo.APIVersion)
-		if err != nil {
-			return nil, errors.Wrap(err, "unable to parse apiVersion")
-		}
-		documentGroupVersionKind := documentGroupVersion.WithKind(typeMetaInfo.Kind)
+	// apply defaults
+	Scheme.Default(latestPublicConfig)
```
BenTheElder: note to self: defaulting & loading should probably be distinct ... for combining with flags etc before defaulting ...

PR #212, opened 2019-01-11
refactor

BenTheElder:
refactoring pkg/cluster in particular to be a little easier to follow hopefully. trying to avoid massive files that do all the things :^)

- moved a lot of code into `internal` packages within cluster
- weakened some coupling

These internal interfaces are still not quite what i'd like, but they should help bring things back in that direction. cluster.Context code is mostly more high level again.


BenTheElder:
easier to follow: maybe
less monolithic: yes

pkg/cluster/internal/meta/meta.go
```diff
@@ -0,0 +1,62 @@
...
+*/
+
+package meta
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"sigs.k8s.io/kind/pkg/cluster/consts"
```
neolit123: no blocking nit: i would have renamed `consts` to `constants(.go)` to be less slangy.

pkg/cluster/internal/meta/meta.go
```diff
@@ -0,0 +1,62 @@
...
+
+// Name returns the cluster's name
+func (c *ClusterMeta) Name() string {
+	return c.name
+}
+
+// KubeConfigPath returns the path to where the Kubeconfig would be placed
+// by kind based on the configuration.
+func (c *ClusterMeta) KubeConfigPath() string {
+	// TODO(bentheelder): Windows?
```
neolit123: `%HOME%/.kube` is a valid path on win32.
this function is portable, so we can omit the TODO?

neolit123:
oops missed the WIP part.

pkg/cluster/internal/meta/meta.go
neolit123: ok, i guess that's true. it was set for me on Windows 7 (explicitly), but not on a Windows XP VM and also on Windows 10.

BenTheElder: this matches upstream now: https://github.com/kubernetes/client-go/blob/ee7a1ba5cdf1292b67a1fdf1fa28f90d2a7b0084/tools/clientcmd/loader.go#L51

pkg/cluster/internal/meta/meta.go
BenTheElder: done

BenTheElder:
/hold cancel
still not the end state I want, but I'd rather get what we have now in and iterate further later. it's time to cut a release after this I think :^)

PR #215, opened 2019-01-11
correctly check KUBECONFIG when deleting clusters

### LINKED ISSUE #211: correctly check KUBECONFIG when deleting clusters
This check is incomplete:
https://github.com/kubernetes-sigs/kind/blob/2b270a02556671be069c19c38a404a1eaf5951a9/pkg/cluster/context.go#L382-L385

Per [the official docs], `KUBECONFIG` may contain multiple paths. The current naive check is insufficient.

NOTE: this only affects wether we give help text or not, so it is not critical. But we should do the correct thing.

/priority backlog
/help

[the official docs]: https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/#set-the-kubeconfig-environment-variable

manparvesh:
Fixes #211

### DIFF
--- a/pkg/cluster/context.go
+++ b/pkg/cluster/context.go
@@ -380,9 +380,7 @@ func (c *Context) Delete() error {
 	}
 
 	// check if $KUBECONFIG is set and let the user know to unset if so
-	// TODO(bentheelder): this is technically not a sufficient check due to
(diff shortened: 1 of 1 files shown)


manparvesh:
@neolit123 thanks for the link, I wasn't aware that we can rebase in PRs too.

PR #222, opened 2019-01-16
overhaul image defaulting

BenTheElder:
.. and validation

also, update to the latest (Kubernetes 1.13.2) image default


pkg/cluster/context.go
```diff
@@ -101,7 +102,11 @@ func (c *Context) ClusterName() string {
 
 // Create provisions and starts a kubernetes-in-docker cluster
 func (c *Context) Create(cfg *config.Config, retain bool, wait time.Duration) error {
-	// validate config first
+	// default config fields (important for usage as a library, where the config
+	// may be constructed in memory rather than from disk)
+	encoding.Scheme.Default(cfg)
```
BenTheElder: I don't totally love what all is going on here, but to use kind as a library we need to apply defaulting etc on supplied config because we don't know that it went through load etc.

hack/update-generated.sh
```diff
@@ -32,7 +32,9 @@ GOBIN="${OUTPUT_GOBIN}" go install ./vendor/k8s.io/code-generator/cmd/conversion
 # NOTE: go will only take package paths, not absolute directories
 export PATH="${OUTPUT_GOBIN}:${PATH}"
 go generate ./...
+
 deepcopy-gen -i ./pkg/cluster/config/ -O zz_generated.deepcopy --go-header-file hack/boilerplate.go.txt
+defaulter-gen -i ./pkg/cluster/config/ -O zz_generated.default --go-header-file hack/boilerplate.go.txt
```
neolit123: might as well bump the boilerplate to 2019?

BenTheElder: SGTM :+1:

PR #223, opened 2019-01-16
adding "why kind?" and alternatives list

jimangel:
From slack conversation. Open to feedback. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,25 @@ See the [issue tracker] if you're unsure where to start, or feel free to reach o
 
 See also: the Kubernetes [community page].
 
+## Why kind?
+
+ - kind supports building official Kuberbetes release builds from source
+ - kind ships upstream Kubernetes release builds
+ - kind supports building Kubernetes with make/bash/docker, bazel, or installing from apt, in addition to pre-published builds.
+ - kind is written in go, and can be used as a library, has binary releases
+ - kind supports windows in addition to mac and linux
+ - for more details see [the design documentation][design doc]
+
+## Alternatives
+
+Some other open source projects with slightly different but very overlapping use cases, features etc.
+
+- https://github.com/kubernetes-sigs/kubeadm-dind-cluster
+- https://github.com/kinvolk/kube-spawn
+- https://github.com/kubernetes/minikube
+- https://github.com/danderson/virtuakube
+- https://github.com/ubuntu/microk8s
+
 ### Code of conduct
 
 Participation in the Kubernetes community is governed by the [Kubernetes Code of Conduct].
--- a/README.md
+++ b/README.md
@@ -62,8 +62,7 @@ See also: the Kubernetes [community page].
 
 ## Why kind?
 
- - kind supports building official Kuberbetes release builds from source
- - kind ships upstream Kubernetes release builds
+ - kind supports building Kubernetes release builds from source
  - kind supports building Kubernetes with make/bash/docker, bazel, or installing from apt, in addition to pre-published builds.
  - kind is written in go, and can be used as a library, has binary releases
  - kind supports windows in addition to mac and linux


README.md
```diff
@@ -60,6 +60,26 @@ See the [issue tracker] if you're unsure where to start, or feel free to reach o
 
 See also: the Kubernetes [community page].
 
+## Why kind?
+
+kind exists because we wanted something we were comfortable using for Kubernetes's own CI to test Kubernetes itself.
+
+ - kind doesn't use hyperkube or patch in binaries to existing images, we ~build a "real" kubernetes release, our images are more or less (soon to be fully) self-contained.
```
BenTheElder: this one is a rather specific callout, not sure if we should keep it as is :grimacing:

README.md
```diff
@@ -60,6 +60,26 @@ See the [issue tracker] if you're unsure where to start, or feel free to reach o
...
+ - for more details see [the design documentation][design doc]
+
+## Alternatives
+
+Other open source projects with slightly different but very overlapping use cases, features etc.
+
+- https://github.com/kinvolk/kube-spawn
+- https://github.com/kubernetes/minikube
+- https://github.com/danderson/virtuakube
+- https://github.com/ubuntu/microk8s
```
BenTheElder: we should include https://github.com/kubernetes-sigs/kubeadm-dind-cluster

README.md
```diff
@@ -60,6 +60,26 @@ See the [issue tracker] if you're unsure where to start, or feel free to reach o
...
+ - kind doesn't use hyperkube or patch in binaries to existing images, we ~build a "real" kubernetes release, our images are more or less (soon to be fully) self-contained.
+ - kind supports building kubernetes with make/bash/docker, bazel, or installing from apt, in addition to pre-published builds.
+ - kind is written in go, and can be used as a library, has binary releases
+ - kind is a clean rewrite vs previous attempts such as test-infra/dind
+ - kind supports windows in addition to mac and linux
+ - for more details see [the design documentation][design doc]
+
+## Alternatives
+
+Other open source projects with slightly different but very overlapping use cases, features etc.
```
BenTheElder: maybe "Some other ... include", we probably haven't captured all of them :-)

README.md
```diff
@@ -60,6 +60,26 @@ See the [issue tracker] if you're unsure where to start, or feel free to reach o
 
 See also: the Kubernetes [community page].
 
+## Why kind?
+
+kind exists because we wanted something we were comfortable using for Kubernetes's own CI to test Kubernetes itself.
```
BenTheElder: maybe s/were comfortable using/we could use/

README.md
```diff
@@ -60,6 +60,26 @@ See the [issue tracker] if you're unsure where to start, or feel free to reach o
...
+
+ - kind doesn't use hyperkube or patch in binaries to existing images, we ~build a "real" kubernetes release, our images are more or less (soon to be fully) self-contained.
+ - kind supports building kubernetes with make/bash/docker, bazel, or installing from apt, in addition to pre-published builds.
+ - kind is written in go, and can be used as a library, has binary releases
+ - kind is a clean rewrite vs previous attempts such as test-infra/dind
```
BenTheElder: I think we can probably drop this point too, it has plusses and minuses and suggests other projects are somehow unclean

BenTheElder:
thanks for this! bike-shedded a bit :upside_down_face:

README.md
jimangel: which point? The "clean rewrite"?

README.md
jimangel: What if we trimmed it down to "kind doesn't use hyperkube or patch in binaries to existing images"?

jimangel:
> thanks for this! bike-shedded a bit 🙃

No worries! :)

README.md
BenTheElder: thought: maybe the opposite direction? something along the lines of """"kind ships upstream release builds""" ? hyperkube is also in-and-of-itself fine (and pretty neat!), it's just less common I think, and we ship a very typical upstream build w/o any patching business.

jimangel: nit: s/ships/supports/ ?

BenTheElder: It does both, not sure which is clearer. 
We support building from apt, or building from k8s source in the same style as a release, out of the box though we also ship pre-built images that contain a release .. :man_shrugging:

jimangel:
added nits and squashed.

README.md
```diff
@@ -60,6 +60,25 @@ See the [issue tracker] if you're unsure where to start, or feel free to reach o
 
 See also: the Kubernetes [community page].
 
+## Why kind?
+
+ - kind supports building official Kuberbetes release builds from source
+ - kind ships upstream Kubernetes release builds
```
BenTheElder: now that I think about it, this particular point might also be misleading, the builds in our "official" kind images ~~ match how upstream builds but are actually happening on my devices currently. let's add it back when that changes. I think the above point covers it.

README.md
```diff
@@ -60,6 +60,25 @@ See the [issue tracker] if you're unsure where to start, or feel free to reach o
 
 See also: the Kubernetes [community page].
 
+## Why kind?
+
+ - kind supports building official Kuberbetes release builds from source
```
BenTheElder: "official" is probably actually misleading too, these are not blessed by the release team, they just mimic those as much as possible. :thinking: 

s/Kuberbetes/Kubernetes/

Issue #197, opened 2019-01-02
Build and release `kindest/node:VERSION` on kubernetes release


chuckha:
Right now the node image that `kind` uses is published by hand by @BenTheElder or @munnerz. It would be good to get this into the release pipeline somehow so that when a new version of kubernetes is published we also get a new node image for `kind` to use. The tooling exists, it's just a matter of getting the image pushed to the correct place. See the tooling at https://github.com/kubernetes-sigs/kind/blob/master/hack/build/push-node.sh

For complete context, please see [this slack thread](https://kubernetes.slack.com/archives/CEKK1KTN2/p1546469825101400).

There are a few approaches discussed in the thread:

1. Shove this into anago in a similar way that the conformance image works. This is easy but the downside is that anago is getting too big and we shouldn't be adding more stuff to anago. This may become a slippery slope and we don't want anago to grow.

2. periodically check if the release tags have changed. If they have changed, do a build, otherwise don't do anything.

3. @BenTheElder has been looking at extending prow to trigger off GCS/GCR pubsub so we can kick off a build after a normal release is finished.

We would like to move this off [dockerhub](https://hub.docker.com/r/kindest/node/) before 1.0, but there isn't really a great place to put it right now. Ideally we could have a CNCF sponsored gcr.io bucket so that google isn't responsible for the storage.

neolit123:
i guess, i'm only -1 only on the anago approach.

> We would like to move this off dockerhub before 1.0, but there isn't really a great place to put it right now. Ideally we could have a CNCF sponsored gcr.io bucket so that google isn't responsible for the storage.

this overlaps with the work by the k8s-infra-team, so if they manage to do the above soon the kind image can be on CNCF ground, otherwise dockerhub seems like a viable option for 1.0.

BenTheElder:
Ditto w/ @neolit123 on anago per our previous discussion, I don't think Kubernetes the core project should be in the business of releasing SIG subprojects just yet, we can ensure we publish images for each release by other means. 

For now if anyone really wants them today, they can check out a k8s release tag and build an "unnoficial image" with the same tools we use.
 
I'm also ambivalent on dockerhub, there haven't been major downsides so far but I'd be happy to move it to whatever hosting k8s-infra-team settles on going forward. The current dockerhub was indeed a stopgap to have joint access with @munnerz but has worked fine.

chuckha:
cross posting a few communication links

email to steering committee about this
[...]

PR #226, opened 2019-01-16
tag 0.1.0 release

### LINKED ISSUE #208: Binary release cadence
Hi folks,

I wrote a blog post on the heels of KubeCon Seattle explaining what KinD is and how it can be used for testing with Kubernetes. I wanted to write a follow-up but wasn't sure if `go get` is still the preferred installation method?

There is a binary 0.1 release but it looks out of date to me vs. master (I may be wrong)

What would you recommend here? It would be ideal if people didn't need go as a pre-req - especially if testing on a minimal VM.

Thanks 👍 

Alex

BenTheElder:
it's time :-)
fixes #208

### DIFF
--- a/cmd/kind/version/version.go
+++ b/cmd/kind/version/version.go
@@ -22,7 +22,7 @@ import (
 )
 
 // Version is the kind CLI version
-const Version = "0.1.0"
+const Version = "0.2.0-alpha"
 
 // NewCommand returns a new cobra.Command for version
 func NewCommand() *cobra.Command {
--- a/cmd/kind/version/version.go
+++ b/cmd/kind/version/version.go
@@ -22,7 +22,7 @@ import (
 )
 
 // Version is the kind CLI version
-const Version = "0.1.0-alpha"
+const Version = "0.1.0"
 
 // NewCommand returns a new cobra.Command for version
 func NewCommand() *cobra.Command {


tao12345666333:
@BenTheElder  can we add some ci jobs using GitHub API to do this? https://developer.github.com/v3/repos/releases/#create-a-release

we can add one `VERSION` file and make the ci job watch it for release.

BenTheElder:
@tao12345666333 creating a release like that requires referencing a tag (`tag_name`) so I think it won't buy us much. We also still have to push a tag, write the release notes, and post the binaries. I looked at automating more of that but the uploading and tagging part is quick & simple, and automated release notes will not be as good as human curated notes I think :-)

We script (`hack/release/create.sh`):
- updating the version file
- creating the tag
- building binaries for all platforms
- updating the version file again

Manually:
- PR the version file
- Push the tag
- Create a release entry / write release notes, upload the binaries to that

BenTheElder:
we are looking at #197 though for automation of the images, and if we do this more often it will take less time to write the notes hopefully :-)

tao12345666333:
Thanks for your explanation. :+1:

neolit123:
> The current kind Config is now at kind.sigs.k8s.io/v1alpha2, with support for multi-node clusters! (thanks @fabriziopandini!)

we should have outlined that multi-node is enabled on the config side, but not in the backend?

BenTheElder:
@neolit123 multi-node should work (?), but not EG HA, external etcd etc.

neolit123:
hmm, @fabriziopandini could you please comment.
was under the impression that there is still something in the works there.
maybe i should just test myself...

BenTheElder:
There is some in the works, particularly HA related.

fabriziopandini:
@BenTheElder @neolit123 we currently support n>1 workers only.
I will send asap a PR for supporting n>1 control-plane nodes (with external LB) which is P0 for kubeadm, while external-etcd IMO is a lower priority.
And after this IMO we can focus on UX

Issue #228, opened 2019-01-16
use of multi-node clusters is not documented


rimusz:

alejandrox1:
Thank you for filing this issue @rimusz . We fell behind in documentation during the holidays but this is a work in progress. We will get some updates to all the docs as very soon and any help is welcomed.

rimusz:
even some simple example for the start would be good :-)
if I understand correctly that will be done via config file?

neolit123:
> use of multi-node clusters is not documented

multi-node is WIP still and shouldn't be exposed to users yet.
once we have it working we can add some basic documentation.

/kind documentation
suggested rename for the ticket title "Document usage of multi-node clusters".

rimusz:
gotcha

BenTheElder:
Fixed in #232

PR #232, opened 2019-01-18
document multi-node cluster config

### LINKED ISSUE #217: Creating multi-node clusters
Documentation around config to create multi-node clusters is not very clear. I tried this
```
apiVersion: kind.sigs.k8s.io/v1alpha1
kind: Config
---
kind: Node
apiVersion: kind.sigs.k8s.io/v1alpha1
role: control-plane
replicas: 3
---
kind: Node
apiVersion: kind.sigs.k8s.io/v1alpha1
role: worker
replicas: 2
```
But that still created a single node cluster. What did I miss in the config?

BenTheElder:
fixes #217 #228

### DIFF
--- a/docs/user/README.md
+++ b/docs/user/README.md
@@ -29,11 +29,11 @@ single node `kind` cluster) dedicated to the virtual machine (VM) running the
 Docker engine otherwise the Kubernetes cluster might fail to start up.
 
 To change the resource limits for the Docker engine on Mac, you'll need to open the
-**Preferences** menu.
+**Preferences** menu.  
 <img src="./images/docker-pref-1.png"/>
 
 Now, go to the **Advanced** settings page, and change the
-settings there, see [changing Docker's resource limits][Docker resource lims].
+settings there, see [changing Docker's resource limits][Docker resource lims].  
 <img src="./images/docker-pref-2.png"/>
 
 You may also try removing any unused data left by the Docker engine - e.g.,
@@ -152,6 +152,17 @@ To specify a configuration file when creating a cluster, use the `--config`
 flag.
 For a sample kind configuration file see [kind-example-config][kind-example-config].
 
+In particular, many users may be interested in multi-node clusters. A simple
+configuration for this can be acheived with the following config file contents:
+```yaml
+# three node cluster config, with two worker nodes
+kind: Config
+apiVersion: kind.sigs.k8s.io/v1alpha2
+nodes:
+- role: control-plane
+- role: worker
+  replicas: 2
+```
 
 ### Exporting `kind`'s Logs
 `kind` has the ability to export all `kind` related logs for you to explore.
--- a/docs/user/kind-example-config.yaml
+++ b/docs/user/kind-example-config.yaml
@@ -1,23 +1,28 @@
 # this config file contains all config fields with comments
-apiVersion: kind.sigs.k8s.io/v1alpha1
 kind: Config
+apiVersion: kind.sigs.k8s.io/v1alpha2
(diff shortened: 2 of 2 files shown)


docs/user/kind-example-config.yaml
```diff
@@ -1,23 +1,28 @@
...
 kind: Config
+apiVersion: kind.sigs.k8s.io/v1alpha2
+# 1 control plane node and 3 workers
+nodes:
+# the control plane node config
+- role: control-plane
+  # patch the generated kubeadm config with some extra settings
+  kubeadmConfigPatches:
+  - |
+    apiVersion: kubeadm.k8s.io/v1alpha3
```
alejandrox1: Change to `apiVersion: kubeadm.k8s.io/v1beta1`

docs/user/kind-example-config.yaml
```diff
@@ -1,23 +1,28 @@
...
+    kind: ClusterConfiguration
+    networking:
+      serviceSubnet: 10.0.0.0/16
+  # patch it further using a JSON 6902 patch
+  kubeadmConfigPatchesJson6902:
+  - group: kubeadm.k8s.io
+    version: v1alpha3
```
alejandrox1: change to `version: v1beta1`

docs/user/kind-example-config.yaml
```diff
@@ -1,23 +1,28 @@
...
+    kind: ClusterConfiguration
+    patch: |
+      - op: add
+        path: /apiServerCertSANs/-
+        value: my-hostname
```
alejandrox1: Should we add some documentation for this? Explaining the syntax/use case?

docs/user/README.md
```diff
@@ -29,11 +29,11 @@ single node `kind` cluster) dedicated to the virtual machine (VM) running the
 Docker engine otherwise the Kubernetes cluster might fail to start up.
 
 To change the resource limits for the Docker engine on Mac, you'll need to open the
-**Preferences** menu.
+**Preferences** menu.
```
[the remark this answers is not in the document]
BenTheElder: I meant to do this and forgot to mention it, this is a new line to clear the image in some markdown renderers.

docs/user/kind-example-config.yaml
BenTheElder: Thanks. These are problematic because they depend on which k8s version is used (needs to match underlying APIs), will change this 👍

docs/user/kind-example-config.yaml
BenTheElder: JSON 6902 is an IETF RFC we should link to that and probably kustomize somewhere in the docs, the GVK bit comes from k8s to define which object, and we're shimming kustomize to make this all inline.

FWIW, I have mixed feelings on this field. It will be a really handy and stable tool once the kubeadm API version remains stable I think 🤔

docs/user/kind-example-config.yaml
BenTheElder: done

docs/user/kind-example-config.yaml
BenTheElder: done

alejandrox1:
This doesn;t have any effect since im not a reviewer but I always wanted to do it:

/lgtm

BenTheElder:
> This doesn't have any effect since im not a reviewer but I always wanted to do it:

not true! reviewer is about assignment and being in the heirarchy, but technically lgtm is restricted only to org members for each org :-) (same for /hold IIRC)

also, you're in the org now! :tada:

Issue #134, opened 2018-11-23
External load balancer


fabriziopandini:
Multi node scenario might require an external load balancer in front of control-plane instances (if more than one should be deployed in a cluster)

A possible solution is to use an HAProxy container to be created in addition to the cluster nodes, but other solutions could be considered as well.

The HAProxy solution, even if simple, introduces a "dependency" from a component not currently included in the node-image, and this should be considered as well.

opinions?
/assign
/

neolit123:
> The HAProxy solution, even if simple, introduces a "dependency" from a component not currently included in the node-image, and this should be considered as well.

so if HAProxy is installed, but disabled by default on the default node image and we have control of what we can execute on such a node we can set this node as an external LB.

one problem here is that the node-image is pretty big and i don't think it should use it for just this purpose, but for testing it seems like an option to consider as well.

i think we should start estimating RAM usage for common setups too.

i've commented here on the related topic:
https://github.com/kubernetes-sigs/kind/issues/135#issuecomment-442670776

BenTheElder:
how big is a haproxy install, exactly?

I've requested that we try to keep things versioned within that image. This means that eg our first release has two artifacts:
- `kind` CLI binaries
- a pinned image used by those binaries, which we can never change because we pin it by the sha.

for the default usage, absolutely everything is stable in this case (well except weave, but that's a bug not a pattern to follow)

If it's not too big, the image downloads are compressed anyhow and the containers should be COW, so it's probably ok to just add it to the base image.

neolit123:
this page tells me 1MB for the DEB of recent haproxy:
https://ubuntu.pkgs.org/18.10/ubuntu-main-amd64/haproxy_1.8.13-2build1_amd64.deb.html

> If it's not too big, the image downloads are compressed anyhow and the containers should be COW, so it's probably ok to just add it to the base image.

oh, copy on write, of course. i forgot about this magic.
i see the the CNI plugins are part of the base image so its seems like the right place for it.

fabriziopandini:
@BenTheElder it is time to start implementing external load balancer

If I got it right solutions discussed above are:
1. Use ha proxy image for external external load balancing nodes
2. Use kind default image, start docker in docker, start haproxy using docker in docker (eventually pre-loaded)

1 is simplest, and probably with less footprint, but 2 address the foundament requirement keeping things versioned within a unique node image.
So I'm taking two as working assumption.

Please let me know if there are concerns/other things to consider

[...]

PR #233, opened 2019-01-18
support n>1 control plane nodes and the external load balancer

fabriziopandini:
This PR adds support for n>1 control plane nodes and for the external load balancer.

Implementing this PR is necessary for starting using kind for improving kubeadm test grid for HA scenario, as required by https://github.com/kubernetes/enhancements/issues/357

Fixes: https://github.com/kubernetes-sigs/kind/issues/134

### DIFF
--- a/pkg/cluster/context.go
+++ b/pkg/cluster/context.go
@@ -158,7 +158,7 @@ func (c *Context) Create(cfg *config.Config, retain bool, wait time.Duration) er
 	// Kubernetes cluster; please note that the list of actions automatically
 	// adapt to the topology defined in config
 	// TODO(fabrizio pandini): make the list of executed actions configurable from CLI
-	err = cc.Exec(nodeList, []string{"config", "init", "join"}, wait)
+	err = cc.Exec(nodeList, []string{"haproxy", "config", "init", "join"}, wait)
 	if err != nil {
 		// In case of errors nodes are deleted (except if retain is explicitly set)
 		log.Error(err)
--- a/pkg/cluster/internal/create/haproxy.go
+++ b/pkg/cluster/internal/create/haproxy.go
@@ -1,0 +1,135 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package create
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+
+	"sigs.k8s.io/kind/pkg/cluster/internal/haproxy"
+	"sigs.k8s.io/kind/pkg/cluster/internal/kubeadm"
+)
+
+// HAProxyAction implements action for configuring and starting the
+// external load balancer in front of the control-plane nodes.
+type HAProxyAction struct{}
+
+func init() {
+	registerAction("haproxy", NewHAProxyAction)
+}
+
+// NewHAProxyAction returns a new HAProxyAction
+func NewHAProxyAction() Action {
+	return &HAProxyAction{}
+}
+
+// Tasks returns the list of action tasks
+func (b *HAProxyAction) Tasks() []Task {
+	return []Task{
+		{
+			Description: "Starting the external load balancer ⛵",
+			TargetNodes: selectExternalLoadBalancerNode,
+			Run:         runHAProxy,
+		},
+	}
+}
+
+// runKubeadmJoin executes haproxy
+func runHAProxy(ec *execContext, configNode *NodeReplica) error {
+	// collects info about the existing controlplane nodes
+	var backendServers = map[string]string{}
+	for _, n := range ec.ControlPlanes() {
+		// gets the handle for the control plane node
+		controlPlaneHandle, ok := ec.NodeFor(n)
+		if !ok {
+			return fmt.Errorf("unable to get the handle for operating on node: %s", n.Name)
+		}
+
+		// gets the IP of the control plane node
+		controlPlaneIP, err := controlPlaneHandle.IP()
+		if err != nil {
+			return errors.Wrapf(err, "failed to get IP for node %s", n.Name)
+		}
+
+		backendServers[n.Name] = fmt.Sprintf("%s:%d", controlPlaneIP, kubeadm.APIServerPort)
+	}
+
+	// create haproxy config file writing a local temp file on the host machine
+	haproxyConfig, err := createHAProxyConfig(
+		haproxy.ConfigData{
+			ControlPlanePort: haproxy.ControlPlanePort,
+			AdminPort:        haproxy.AdminPort,
+			BackendServers:   backendServers,
+		},
+	)
+	if err != nil {
+		// TODO(bentheelder): logging here
+		return fmt.Errorf("failed to create kubeadm config: %v", err)
+	}
+
+	// defer deletion of the local temp file
+	defer os.Remove(haproxyConfig)
+
+	// get the target node for this task (the load balancer node)
+	node, ok := ec.NodeFor(configNode)
+	if !ok {
+		return fmt.Errorf("unable to get the handle for operating on node: %s", configNode.Name)
+	}
+
+	// copy the haproxy config file from the host to the node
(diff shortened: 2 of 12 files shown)


pkg/cluster/nodes/node.go
```diff
@@ -88,6 +88,11 @@ func (n *Node) CopyTo(source, dest string) error {
 	return docker.CopyTo(source, n.nameOrID, dest)
 }
 
+// CopyFrom copies the source file on the node to dest on the host
```
BenTheElder: note that this does have limitations around symlinks, I initially implemented this method and then dropped it for log dumping because of this. (we should consider a similar implementation here in the future, but this seems reasonable for now)

pkg/cluster/internal/haproxy/config.go
```diff
@@ -0,0 +1,94 @@
...
+
+// ConfigData is supplied to the haproxy config template
+type ConfigData struct {
+	ControlPlanePort int
+	AdminPort        int
+	BackendServers   map[string]string
+}
+
+// DefaultConfigTemplate is the haproxy config template
+const DefaultConfigTemplate = `
```
BenTheElder: @munnerz @neolit123 are you familiar with haproxy config? I need to do some reading :-)

BenTheElder:
let's _NOT_ do this in the MVP PR, but for compatibility reasons I think we should handle the case that haproxy is not present in the node, and we should also probably think about detecting the haproxy version and handling that.

those should come later though, so as to keep the PR size down and give us some time to explore options. (EG maybe we can install haproxy at creation time if it isn't present)

pkg/cluster/context.go
```diff
@@ -158,7 +158,7 @@ func (c *Context) Create(cfg *config.Config, retain bool, wait time.Duration) er
 	// Kubernetes cluster; please note that the list of actions automatically
 	// adapt to the topology defined in config
 	// TODO(fabrizio pandini): make the list of executed actions configurable from CLI
-	err = cc.Exec(nodeList, []string{"config", "init", "join"}, wait)
+	err = cc.Exec(nodeList, []string{"haproxy", "config", "init", "join"}, wait)
```
BenTheElder: BTW I still intend to refactor this some soon, thinking about how we can make it harder to do wrong (EG strings that aren't valid), and more highly concurrent for those speedy test times :upside_down_face: 

Making it internal for now allows us to delay that in favor of landing awesome stuff like this PR :D

pkg/cluster/nodes/node.go
fabriziopandini: Added a note to keep track of this.
We should consider that this should probably go away when kubeadm automatic copy certs lands, otherwise it should be refactored in something more robust in the long term

pkg/cluster/internal/create/haproxy.go
```diff
@@ -0,0 +1,135 @@
...
+			BackendServers:   backendServers,
+		},
+	)
+	if err != nil {
+		// TODO(bentheelder): logging here
+		return fmt.Errorf("failed to create kubeadm config: %v", err)
+	}
+
+	// defer deletion of the local temp file
+	defer os.Remove(haproxyConfig)
```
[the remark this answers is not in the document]
fabriziopandini: Done also for the kubeadm-config action

pkg/cluster/context.go
fabriziopandini: Concurrency will give a boost on the multi-node use cases, while for single node I assume this will be not relevant. In turn, we will lose something on the UX (e.g the nice spinner or the readability of the command output)

I think that as of today it is already possible to introduce concurrency on the node provisioning part; actions instead have an intrinsic notion of order that should be considered, but it is possible to get some improvements there as well. Whenever you want to do this ping me and I will share a possible solution  😉.

I got your point about keeping things internal, but there are also some backside in this as well; this makes this point an interesting topic the upcoming kind meetings!

BenTheElder:
SGTM 👍

I've only had a chance to take a quick pass so far, had a lot come up today
🙄 Will look more this weekend or first thing in the new week.

First pass looked great 😄

pkg/cluster/nodes/create.go
```diff
@@ -67,6 +68,39 @@ func CreateControlPlaneNode(name, image, clusterLabel string) (node *Node, err e
 	return node, nil
 }
 
+// CreateExternalLoadBalancerNode creates an external loab balancer node
+// and gets ready for exposing the the API server and the load balancer admin console
+func CreateExternalLoadBalancerNode(name, image, clusterLabel string) (node *Node, err error) {
+	// gets a random host port for the API server
```
neolit123: "gets a random port for the control-plane load balancer"?

pkg/cluster/internal/create/haproxy.go
```diff
@@ -0,0 +1,134 @@
...
+// NewHAProxyAction returns a new HAProxyAction
+func NewHAProxyAction() Action {
+	return &HAProxyAction{}
+}
+
+// Tasks returns the list of action tasks
+func (b *HAProxyAction) Tasks() []Task {
+	return []Task{
+		{
+			Description: "Starting the external load balancer ⛵",
```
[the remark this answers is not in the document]
BenTheElder: It is normal to have a single system preferred font for these glyphs, EG on android it is "Noto Color Emoji", typically there's a font stack with one or more extended &/or emoji fonts handling these as one of the "fallback" fonts. These are all common ones that should render.

OTOH, the default cmd.exe shell on windows only has some pretty old raster font rendering, it is the only major shell I know if that is fully incapable of handling these out of the box :|

I don't think moving these to a constant gains us anything, they are in constant strings as is, and moving the whole string to a constants file would give less context.

pkg/cluster/internal/create/haproxy.go
```diff
@@ -0,0 +1,134 @@
...
+
+// runKubeadmJoin executes haproxy
+func runHAProxy(ec *execContext, configNode *NodeReplica) error {
+	// collects info about the existing controlplane nodes
+	var backendServers = map[string]string{}
+	for _, n := range ec.ControlPlanes() {
+		// gets the handle for the control plane node
+		controlPlaneHandle, ok := ec.NodeFor(n)
+		if !ok {
+			return fmt.Errorf("unable to get the handle for operating on node: %s", n.Name)
```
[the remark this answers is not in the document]
BenTheElder: FWIW these are the status quo, going through and migrating to a more consistent set of error utils would be great though. might be worth an issue to track and list options?

pkg/cluster/internal/haproxy/config.go
BenTheElder: Correct, not looking to block on that! If anyone else is already familiar a review would be appreciated, otherwise I intend to do some reading for the future :-)

Re: admin -- I think we should remove it initially and deal with the implications in a follow-up later. I've been thinking more about the config for these things - have a couple ideas I want to run by you all soon but I don't think they need to be in this PR.

pkg/cluster/context.go
BenTheElder: yeah totally agree with those, been thinking about this one too. the UX is tough but I do think it is important to keep the time minimal, fast iteration loops make developers (like me!) happy and I think we can find a good middle ground. 

again have a few half baked ideas to discuss soon, but I don't want to introduce the complexity right now either :-)

pkg/cluster/internal/create/kubeadm-join.go
```diff
@@ -40,49 +43,130 @@ func newKubeadmJoinAction() Action {
...
+		// are accessible only inside the docker network
+		joinAddress,
+		// set the node to join as control-plane
+		"--experimental-control-plane",
+		// uses a well known token and skips ca certification for automating TLS bootstrap process
+		"--token", kubeadm.Token,
+		"--discovery-token-unsafe-skip-ca-verification",
+		// preflight errors are expected, in particular for swap being enabled
+		// TODO(bentheelder): limit the set of acceptable errors
+		"--ignore-preflight-errors=all",
```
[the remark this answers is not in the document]
BenTheElder: possibly, OTOH I'd rather continue on for now and handle actual failures. I've been meaning to write some of these things down in the dev guide, graceful degradation is a core tenent so far.

environmental checks are also generally expected to fail given our environment not being technically supported for many reasons.

I consider this similar to how we do not fail on ensuring the images and instead fail if `docker run $img` fails - kind will continue on until it can't :^)

pkg/cluster/internal/haproxy/config.go
fabriziopandini: Ok, admin console is gone now!

neolit123:
thanks for the update @fabriziopandini 
this is LGTM on my side.

pkg/cluster/config/encoding/testdata/v1alpha2/valid-full-ha.yaml
```diff
@@ -2,9 +2,9 @@
 kind: Config
 apiVersion: kind.sigs.k8s.io/v1alpha2
 nodes:
-- role: control-plane
-  replicas: 3
-- role: worker
-  replicas: 2
+- role: control-plane
```
BenTheElder: baby nit for a follow-up: non-markdown EOL whitespace?

PR #236, opened 2019-01-21
Use templates for the main kubeadm commands

inercia:
`kubeadm X` command are currently hardcoded in the `pkg/cluster/internal/create/kubeadm-X.go` files. This limits the flexibility for invoking these commands, something that can be specially important for developers as they could need to experiment with different flags.

This PR adds the ability to specify these commands using Go templates, where some well-known variables are available (ie, the token, the config file path).

For example, the following config file makes use of the templates for increasing the verbosity of the `kubeadm init` and `kubeadm join` commands:.

```yaml
kind: Config
kubeadm:
  init: kubeadm init -v=4 --ignore-preflight-errors=all --config={{.Config}}
  join: kubeadm join -v=4 {{.IP}}:{{.Port}} --token {{.Token}} --discovery-token-unsafe-skip-ca-verification --ignore-preflight-errors=all
nodes:
- role: control-plane
- role: worker
```


inercia:
> > init: kubeadm init -v=4 --ignore-preflight-errors=all --config={{.Config}}
> > join: kubeadm join -v=4 {{.IP}}:{{.Port}} --token {{.Token}} --discovery-token-unsafe-skip-ca-verification --ignore-preflight-errors=all
> > nodes:
> 
> we'd need to think about making it possibly to expose HA and non-HA joining scenarios using the same template. currently `--experimental-control-plane` is used for HA, while in the future more HA specific flags will be added.

Yeah, I agree, we could add specific configuration for regular join vs extra control planes...

> also "actions" are supposed to provide a solution for custom node commands in kind.

I think that the `kubeadm-init` and `kubeadm-join` are already actions, right? But the problem is that they use hardcoded strings while in my view they should be configured from the config file...
 
> /assign @fabriziopandini @BenTheElder
> WDYT?

BenTheElder:
I'm leaning towards we should just increase the verbosity by default instead or maybe even have a configuration value for that.

Exposing templates externally for this sort of thing is a footgun because then we have to expose details of how we run things internally and make that stable. When we change it later we will break all template users, like how @neolit123 pointed out that this is already broken by HA join.

If we need to customize more flags then kubeadm's --config must be lacking and probably should be enhanced upstream.

inercia:
> I'm leaning towards we should just increase the verbosity by default instead or maybe even have a configuration value for that.

I used the `--v=4` as an example, but that is not the ultimate intention of this PR. The idea was to remove some of the main hardcoded strings and replace them by configurable options.

> Exposing templates externally for this sort of thing is a footgun because then we have to expose details of how we run things internally and make that stable. When we change it later we will break all template users, like how @neolit123 pointed out that this is already broken by HA join.

Then we could add more configurable templates instead of adding more hardcoded strings. For example, I wanted to try to replace the `--ignore-preflight-errors=all` by a more strict set of checks. That would be impossible to do without recompilation. I understand your concerns but I don't think mosot users will use this feature. So by keeping everything hardcoded you get more security, but I get less flexibility as a user.

BenTheElder:
> The idea was to remove some of the main hardcoded strings and replace them by configurable options.

If these things should be configured, we should actually configure them instead of overriding the entire command as a string (also commands as templated strings are generally a bad idea as this assumes properties of running them in a particular shell, we intentionally use argument vectors to avoid these pitfalls).

> That would be impossible to do without recompilation. I understand your concerns but I don't think mosot users will use this feature. So by keeping everything hardcoded you get more security, but I get less flexibility as a user.

Making things first-class configurable is more maintainable than adding direct templates, we tried templates in the past and quickly found that those templates became broken as kind itself relied on setting more things itself which was not possible with outdated templates.

On the other hand, the bulk of options are in kubeadm config, and we allow patching _on top_ of the configuration we internally generate to change additional options in a structured way. As kubeadm config becomes a stable format, this should remain stable and give us room to set new fields in the base config as necessary.

pablochacin:
@inercia If I understand well your main use case is for developers, not final users. Actions are a great move forward in that direction. What you propose is basically been able to tweak the options used in the actions, right?  Otherwise you could replace the action's implementation. 

Do you have any concrete scenario which is not covered by modifying the configuration used for a particular action? It is recompiling the only limitation you see?

inercia:
> @inercia If I understand well your main use case is for developers, not final users. Actions are a great move forward in that direction. What you propose is basically been able to tweak the options used in the actions, right? Otherwise you could replace the action's implementation.
> 
> Do you have any concrete scenario which is not covered by modifying the configuration used for a particular action? It is recompiling the only limitation you see?

Yes, @pablochacin, the idea is to provide some configurability for developers, so they could test `kubeadm` or tools that use `kubeadm` for creating the cluster.

As an example, we are currentlly not using `kubeadm` directly in [`kubic`](https://github.com/kubic-project/kubic-init/) but we start a container that runs `kubeadm` under the hood. If we could replace the `kubeadm init` command by a `docker run our-image` we could test our setup with `kind`...

BenTheElder:
So I see what you are trying to do but a couple of points on this ...

- To my knowledge no other cluster tool (minikube, kops, you name it) exposes this sort of detail. This is not usually an extension point for cluster provisioning.

- Actions, templated commands in the API, etc... these things all expose our internal bootstrap procedures and inputs as API details. Doing this will make it difficult to improve or refactor them in the future.
  - In particular doing this likely locks us out of things like using kubeadm phases in the future, because we've exposed a single nebulous kubeadm call step, where in the future maybe we run the phases and run additional tasks in between ...

- In order to work in dind, we very much must control and supply various options to the kubernetes components

I do not want to merge this until we have thought through how to support it long term. It is important to our other users that we stabilize cluster creation APIs, and continue to support and maintain them.

It is also not clear to me that this is a good option to test your custom kubeadm call, because of dind we must choose options that may not make sense in a "real" deployment (EG ignoring preflight errors, due to being in an unsupported environment that _will_ fail checks) - what exactly does kubic-init do specially?

pablochacin:
> As an example, we are currentlly not using kubeadm directly in kubic but we start a container that runs kubeadm under the hood. If we could replace the kubeadm init command by a docker run our-image we could test our setup with kind...

@inercia In that particular case, besides the fact that you want to avoid recompiling the code for testing different options, one solution could be an alternative implementation of the corresponding `kind' action (I'm not yet familiar with the actions, maybe @BenTheElder could clarify this further). But then you would be basically repeating all the logic for calling those commands just to change the way to invoke the external tool.

The only stable/maintainable solution I see would be to have an API between kind and the external bootstrap tool and make that external tool pluggable. But this looks like a lot of work. There's always the poor man's version: make the external tool configurable and assume it should understand kubeadm's parameter.

fabriziopandini:
@BenTheElder @inercia, my two cents

> Verbosity

Increasing verbosity could be useful, but I think this should be and explicit flag or implicitly linked the loglevel flag (I don't like the idea of users playing with kubeadm commands)

> Experimenting

I'm sponsor/contributor of a better support in kind for kubeadm developers, and I guess this overlaps with @inercia idea of experimenting.

However, I don't think this should be Config driven, because this approach can't cover all the possible scenarios I usually face in my dev/test iterations.

I will be happy to share an alternative proposal asap, and or demo it in one kind meeting, but IMO this discussion should be moved on https://github.com/kubernetes-sigs/kind/issues/133 or on a dedicated issue

neolit123:
@fabriziopandini 

> I will be happy to share an alternative proposal asap, and or demo it in one kind meeting, but IMO this discussion should be moved on #133 or on a dedicated issue

@inercia 

+1 for closing the PR and proceeding with alternative approaches.

inercia:
Ok, sure, I will try to find some other alternative approach.

MalloZup:
thx @inercia and @fabriziopandini @neolit123 for this :+1:

neolit123:
@inercia you can add this for the agenda on the meeting on monday if you can attend.
details can be found in the #kind channel.

PR #240, opened 2019-01-21
New 'zypper' build type

inercia:
New `zypper` build type, for installing packages in openSUSE-based base images with the [`zypper`](https://en.opensuse.org/SDB:Zypper_usage) installer.

This can be used in conjunction with [a Dockerfile for building a openSUSE base image](https://github.com/kubic-project/kubic-init/blob/master/build/kind/base/Dockerfile).

### DIFF
--- a/pkg/build/kube/zypperbits.go
+++ b/pkg/build/kube/zypperbits.go
@@ -1,0 +1,90 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kube
+
+import (
+	"fmt"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	// the packages repo
+	repo = "https://download.opensuse.org/tumbleweed/repo/oss/"
+
+	// packages necessary to install
+	packages = "kubernetes-kubelet kubernetes-kubeadm kubernetes-client"
+)
+
+// ZypperBits implements Bits for the official upstream openSUSE packages
+type ZypperBits struct {
+}
+
+var _ Bits = &ZypperBits{}
+
+func init() {
+	RegisterNamedBits("zypper", NewZypperBits)
+}
+
+// NewZypperBits returns a new Bits backed by the upstream debian packages
+func NewZypperBits(kubeRoot string) (bits Bits, err error) {
+	return &ZypperBits{}, nil
+}
+
+// Build implements Bits.Build
+// for ZypperBits this does nothing
+func (b *ZypperBits) Build() error {
+	return nil
+}
+
+// Paths implements Bits.Paths
+func (b *ZypperBits) Paths() map[string]string {
+	return map[string]string{}
+}
+
+// Install implements Bits.Install
+func (b *ZypperBits) Install(install InstallContext) error {
+	addRepo := fmt.Sprintf("zypper --non-interactive --gpg-auto-import-keys addrepo %s repo-opensuse", repo)
+	if err := install.Run("/bin/sh", "-c", addRepo); err != nil {
+		log.Errorf("Adding Kubernetes apt repository failed! %v", err)
+		return err
+	}
+	// install packages
+	if err := install.Run("/bin/sh", "-c", fmt.Sprintf("zypper in -y %s", packages)); err != nil {
+		log.Errorf("Installing Kubernetes packages failed! %v", err)
+		return err
+	}
+	// get version for version file
+	lines, err := install.CombinedOutputLines("/bin/sh", "-c", `kubelet --version`)
+	if err != nil {
+		log.Errorf("Failed to get Kubernetes version! %v", err)
+		return err
+	}
+	// the output should be one line of the form `Kubernetes ${VERSION}`
+	if len(lines) != 1 {
+		log.Errorf("Failed to parse Kubernetes version with unexpected output: %v", lines)
+		return fmt.Errorf("failed to parse Kubernetes version")
+	}
+	// write version file
+	version := strings.SplitN(lines[0], " ", 2)[1]
+	if err := install.Run("/bin/sh", "-c", fmt.Sprintf(`echo "%s" >> /kind/version`, version)); err != nil {
+		log.Errorf("Failed to get Kubernetes version! %v", err)
+		return err
+	}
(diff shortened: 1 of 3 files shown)


pkg/build/kube/zypperbits.go
```diff
@@ -0,0 +1,90 @@
...
+	// the packages repo
+	repo = "https://download.opensuse.org/tumbleweed/repo/oss/"
+
+	// packages necessary to install
+	packages = "kubernetes-kubelet kubernetes-kubeadm kubernetes-client"
+)
+
+// ZypperBits implements Bits for the official upstream openSUSE packages
+type ZypperBits struct {
+}
```
neolit123: please move `}` to the previous line: `{}`

pkg/build/kube/zypperbits.go
```diff
@@ -0,0 +1,90 @@
...
+// Install implements Bits.Install
+func (b *ZypperBits) Install(install InstallContext) error {
+	addRepo := fmt.Sprintf("zypper --non-interactive --gpg-auto-import-keys addrepo %s repo-opensuse", repo)
+	if err := install.Run("/bin/sh", "-c", addRepo); err != nil {
+		log.Errorf("Adding Kubernetes apt repository failed! %v", err)
+		return err
+	}
+	// install packages
+	if err := install.Run("/bin/sh", "-c", fmt.Sprintf("zypper in -y %s", packages)); err != nil {
+		log.Errorf("Installing Kubernetes packages failed! %v", err)
```
neolit123: `! ` -> `: `
[1]

pkg/build/kube/zypperbits.go
```diff
@@ -0,0 +1,90 @@
...
+		return err
+	}
+	// get version for version file
+	lines, err := install.CombinedOutputLines("/bin/sh", "-c", `kubelet --version`)
+	if err != nil {
+		log.Errorf("Failed to get Kubernetes version! %v", err)
```
neolit123: [1]

pkg/build/kube/zypperbits.go
```diff
@@ -0,0 +1,90 @@
...
+		return err
+	}
+	// the output should be one line of the form `Kubernetes ${VERSION}`
+	if len(lines) != 1 {
+		log.Errorf("Failed to parse Kubernetes version with unexpected output: %v", lines)
+		return fmt.Errorf("failed to parse Kubernetes version")
```
neolit123: please use `github.com/pkg/errors` `errors.New()`

pkg/build/kube/zypperbits.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	// write version file
+	version := strings.SplitN(lines[0], " ", 2)[1]
+	if err := install.Run("/bin/sh", "-c", fmt.Sprintf(`echo "%s" >> /kind/version`, version)); err != nil {
+		log.Errorf("Failed to get Kubernetes version! %v", err)
```
neolit123: [1]

pkg/build/kube/zypperbits.go
inercia: This is [direct translation](https://github.com/kubernetes-sigs/kind/blob/3a6fe2fee6256318e59b1edfe04264942c68894c/pkg/build/kube/aptbits.go#L27-L28) of the `apt` installer, but I will change it.

BenTheElder:
Hi - Thanks for the PRs but I have some concerns --

So this build mode is only going to work with a custom base image? 😬

This mode will not work generally then, if a user runs `kind build node --type=zypper` it will fail. Also if we need to install more things it will be out of sync.

Note that the only reason we have --type=apt is that those packages are published by the Kubernetes project, and that mode may be dropped in the future.

I think maybe instead you should create a node image dockerfile that does this installation, to go with the base image.

BenTheElder:
This is cool work FWIW, but I don't think we want to maintain multiple base images in kind, or ship build options that typically won't work (IE because the special base is required).

inercia:
> Hi - Thanks for the PRs but I have some concerns --
> 
> So this build mode is only going to work with a custom base image? 
> 
> This mode will not work generally then, if a user runs `kind build node --type=zypper` it will fail. Also if we need to install more things it will be out of sync.
> 
> Note that the only reason we have --type=apt is that those packages are published by the Kubernetes project, and that mode may be dropped in the future.
> 
> I think maybe instead you should create a node image dockerfile that does this installation, to go with the base image.

Yes, this mode is only going to work with custom-made images, but I understood that that is something encouraged by `kind` as the `kind build base-image` is there for doing that. I think `kind` should not leave users with the only option of a _DIY_ Dockerfile and then a `docker build` as the next step for creating node-images. I think we should make things easier on users on this...

However, I understand your concerns about having a `zypper` build that is only useful for creating a node-image from a very specific base-image. I was thinking that maybe we could address this concern by implementing the same functionality in some other way.

For example, we could replace the `apt` build by a general `binaries` or `packages` build that could detect the distro and then install the rright packages with `apt`, `zypper`, etc. Or maybe we could also allow to provide some flags for installing pre-downloaded packages, or maybe even the raw binaries...

Taking into account this step can be very simple when installing binaries from a repo fro creating the node-image, we could also create a more  generic build that could just run some command, so users could do:

```sh
$ kind build node-image --image kindest/node:inercia --base-image kindest/base:inercia \
  --type sh
  --cmd 'zypper in -y kubernetes-kubelet kubernetes-kubeadm kubernetes-client'
```

What do you guys think? I think any of these solution would definetively make things easier for users than creating their own Dockerfile...

BenTheElder:
> Yes, this mode is only going to work with custom-made images, but I understood that that is something encouraged by kind as the kind build base-image is there for doing that. 

Er no, it's so we don't keep rebuilding the base portion when building k8s from source and installing it to a node image for running. The focus was entirely on that workflow initially, but we've extended it to publishing the built node images as well for users that do not need to build k8s.

I believe we mention in the docs already that it is not expected that end users will build the base image themselves, however we publish our own tooling.

> What do you guys think? I think any of these solution would definetively make things easier for users than creating their own Dockerfile...

In your example you still have your own dockerfile (the base), I'm suggesting that you ship your own node file in addition to your base file, not that users should generally do this. 

> For example, we could replace the apt build by a general binaries or packages build that could detect the distro and then install the rright packages with apt, zypper, etc. Or maybe we could also allow to provide some flags for installing pre-downloaded packages, or maybe even the raw binaries..

That requires zypper etc to be in the base image to install with it ...?

An option to """build""" and simply copy in binaries might be interesting but... I'm hesitant to add this complexity when building a node image from the same k8s source tree with one of the existing options we have should give the same results.

neolit123:
@inercia i'm also not convinced that we should add more PMs in kind.
we should instead trim the list, possibly.

and as @BenTheElder mentioned kind's scope is for testing Kubernetes and not the base image.
do you have the limitation where all your testing **must** use Kubic as the base distro?

i'm mostly 60/40 for not applying this change, but at the same time it's not that big of a change (non-intrusive) and we can always remove it if you guys decide to drop maintenance.

inercia:
Ok, guys, I fully understand that this implements a feature that would be too specific for us. I think it would be better to close the PR and open some issues about some extension points for `kind`.

PR #242, opened 2019-01-21
Remove dependency constraints on master branches

ZachEddy:
This pull request removes the `master` branch constraints specified in the `Gopkg.toml` file. The goal is to simplify dependency management for projects that import `kind` libraries and the constrained dependencies (e.g. the `k8s.io/apimachinery` package). By removing the constraints, users don't have to update their dependencies to satisfy the constraints or provide package overrides.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -43,16 +43,3 @@ required = [
 [[constraint]]
   name = "github.com/spf13/cobra"
   version = "0.0.3"
-
-[[constraint]]
-  branch = "master"
-  name = "k8s.io/apimachinery"
-
-[[constraint]]
- name = "k8s.io/code-generator"
- branch = "master"
-
-
-[[constraint]]
-  branch = "master"
-  name = "k8s.io/utils"


BenTheElder:
/assign @neolit123 @munnerz 

I think y'all have more experience managing this 😅 
It made sense to me, but would like a confirmation :-)

neolit123:
@ZachEddy please squash the commits into one.
i **think** this is OK as much as my godep understanding goes.

ZachEddy:
@neolit123 Done 👍

munnerz:
So this will be okay now, but when it comes to updating the vendored dependency versions we will not be able to pin to compatible versions of the k/* staging dependencies.

I’d like to understand more about your issues vendoring ‘kind’ in a consumer package as I’m not too sure how this PR helps. ‘dep’ exposes knobs to allow you to force certain constraints in your own project - does that not suffice here?

BenTheElder:
yeah the dep changes are fine, mostly sanity checking the constraint itself.

we may also need to look at switching to go modules soon, as that seems to be where things are eventually headed. dep is serving us quite well for now though :-)

munnerz:
We originally pinned to master as we genuinely need some changes that only exist in master iirc?
With the intention to switch back to the latest stable release-* branch in future.

PR #247, opened 2019-01-24
Add cli flags for multinode

fabriziopandini:
This PR adds CLI flags to change the number of nodes without being forced to use the config files

Fixes: https://github.com/kubernetes-sigs/kind/issues/133

### DIFF
--- a/cmd/kind/create/cluster/createcluster.go
+++ b/cmd/kind/create/cluster/createcluster.go
@@ -31,12 +31,20 @@ import (
 	"sigs.k8s.io/kind/pkg/util"
 )
 
+const (
+	configFlagName            = "config"
+	controlPlaneNodesFlagName = "control-plane-nodes"
+	workerNodesFLagName       = "worker-nodes"
+)
+
 type flagpole struct {
-	Name      string
-	Config    string
-	ImageName string
-	Retain    bool
-	Wait      time.Duration
+	Name          string
+	Config        string
+	ImageName     string
+	Workers       int32
+	ControlPlanes int32
+	Retain        bool
+	Wait          time.Duration
 }
 
 // NewCommand returns a new cobra.Command for cluster creation
@@ -52,7 +60,9 @@ func NewCommand() *cobra.Command {
 		},
 	}
 	cmd.Flags().StringVar(&flags.Name, "name", cluster.DefaultName, "cluster context name")
-	cmd.Flags().StringVar(&flags.Config, "config", "", "path to a kind config file")
+	cmd.Flags().StringVar(&flags.Config, configFlagName, "", "path to a kind config file")
+	cmd.Flags().Int32Var(&flags.ControlPlanes, controlPlaneNodesFlagName, 1, "number of control-plane nodes in the cluster")
+	cmd.Flags().Int32Var(&flags.Workers, workerNodesFLagName, 0, "number of worker nodes in the cluster")
 	cmd.Flags().StringVar(&flags.ImageName, "image", "", "node docker image to use for booting the cluster")
 	cmd.Flags().BoolVar(&flags.Retain, "retain", false, "retain nodes for debugging when cluster creation fails")
 	cmd.Flags().DurationVar(&flags.Wait, "wait", time.Duration(0), "Wait for control plane node to be ready (default 0s)")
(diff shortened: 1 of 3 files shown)


cmd/kind/create/cluster/createcluster.go
```diff
@@ -50,15 +52,25 @@ func NewCommand() *cobra.Command {
 	}
 	cmd.Flags().StringVar(&flags.Name, "name", "1", "cluster context name")
 	cmd.Flags().StringVar(&flags.Config, "config", "", "path to a kind config file")
+	cmd.Flags().Int32Var(&flags.ControlPlanes, "control-planes", 1, "number of control-plane nodes in the cluster")
+	cmd.Flags().Int32Var(&flags.Workers, "workers", 0, "number of worker nodes in the cluster")
```
BenTheElder: Maybe somewhere in the usage we can mention the conflict between these flags and config?

pkg/cluster/config/encoding/scheme.go
```diff
@@ -50,12 +50,28 @@ func AddToScheme(scheme *runtime.Scheme) {
 
 // Load reads the file at path and attempts to convert into a `kind` Config; the file
 // can be one of the different API versions defined in scheme.
-// If path == "" then the default config is returned
-func Load(path string) (*config.Config, error) {
+// If path == "" then the default config is returned according to requested number of
+// control-plane and worker nodes
+func Load(path string, controlPlanes, workers int32) (*config.Config, error) {
```
BenTheElder: Combining this logic into load seems unusual and possibly hostile to library users.

BenTheElder: If we have those fields set, we should just create an internal config in the cmd, set these fields, then default.

BenTheElder:
/hold
Seeing as this doesn't seem to be blocking our current users, did you have need of this feature Fabrizio? If not maybe it will make more sense against the next config type anyhow.

I'm on the fence, I generally dislike flags that cannot be combined with other flags and think that itself makes poor UX, but flags are less effort than config 🤔

BenTheElder:
(also sorry, I believe I requested this initially, but the eagerness people have had to go ahead and use the feature anyhow has been surprising, perhaps we can avoid the messes of combining config and flags, and avoid disjoint flags)

neolit123:
> I'm on the fence, I generally dislike flags that cannot be combined with other flags and think that itself makes poor UX, but flags are less effort than config 

so in terms of the test jobs we want to add to test-infra these flags will definitely help.
i.e. i we don't need pull a repo that stores configs for the sake of creating a custom cluster with topology X.

yet we still have to pull a repo for the test-cmd script, so we might as well store the topology in a config, so that we don't have to touch test infra if a topology changes.

cmd/kind/create/cluster/createcluster.go
```diff
@@ -50,15 +52,25 @@ func NewCommand() *cobra.Command {
 	}
 	cmd.Flags().StringVar(&flags.Name, "name", "1", "cluster context name")
 	cmd.Flags().StringVar(&flags.Config, "config", "", "path to a kind config file")
+	cmd.Flags().Int32Var(&flags.ControlPlanes, "control-planes", 1, "number of control-plane nodes in the cluster")
```
munnerz: 'control-planes' sounds confusing to me. It kind of suggests to me that there could be multiple clusters.

Perhaps this could be something more like `--control-plane-nodes`? Or `--num-nodes control-plane=3 --num-nodes workers=3`? 🤔

cmd/kind/create/cluster/createcluster.go
```diff
@@ -50,15 +52,25 @@ func NewCommand() *cobra.Command {
...
 	return cmd
 }
 
 func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
+	if cmd.Flags().Changed("config") && (cmd.Flags().Changed("control-planes") || cmd.Flags().Changed("workers")) {
+		return fmt.Errorf("flag --config can't be used in combination with --control-plane-nodes or --worker-nodes flags")
+	}
+
+	if flags.ControlPlanes < 0 || flags.Workers < 0 {
+		return fmt.Errorf("flags --control-plane-nodes and --worker-nodes should not be a negative number")
```
munnerz: --control-plane or --control-plane-nodes? 🤔

cmd/kind/create/cluster/createcluster.go
```diff
@@ -50,15 +52,25 @@ func NewCommand() *cobra.Command {
...
+	cmd.Flags().Int32Var(&flags.ControlPlanes, "control-planes", 1, "number of control-plane nodes in the cluster")
+	cmd.Flags().Int32Var(&flags.Workers, "workers", 0, "number of worker nodes in the cluster")
 	cmd.Flags().StringVar(&flags.ImageName, "image", "", "node docker image to use for booting the cluster")
 	cmd.Flags().BoolVar(&flags.Retain, "retain", false, "retain nodes for debugging when cluster creation fails")
 	cmd.Flags().DurationVar(&flags.Wait, "wait", time.Duration(0), "Wait for control plane node to be ready (default 0s)")
 	return cmd
 }
 
 func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
+	if cmd.Flags().Changed("config") && (cmd.Flags().Changed("control-planes") || cmd.Flags().Changed("workers")) {
```
munnerz: `control-planes` here too. Could the flag name be moved into a const if we are referencing it so many times?

pkg/cluster/config/encoding/scheme.go
munnerz: 👍 it feels awkward to pass some random extra vars to this function

cmd/kind/create/cluster/createcluster.go
neolit123: the flags can be renamed `control-plane-nodes`, `worker-nodes`
or something like `--nodes=control-planes=3;workers=5`

but this reminds me of something else.
we also wanted to support multi-cluster with kind, so the flags have to target a cluster. o_o

BenTheElder: --name controls the target cluster for multi cluster. we already support that :-)

timothysc:
This seems generally useful for testing, what's the reservation?

BenTheElder:
> This seems generally useful for testing, what's the reservation?

We've still been discussing how not to dig a hole mixing config and flags. Testing w/ @neolit123 has done pretty well using config only so far so it doesn't seem urgent.

Something like https://github.com/kubernetes-sigs/kind/pull/274 in conjunction with this probably makes more sense. We've been discussing reshaping the config to make the single implicit control plane load balancer (for HA) more explicit.

timothysc:
fwiw as a user if I want to quick-turn a parameter, having a flag is always easier for override. 

config is great for default, but if you readily tweek parameters...  then not so much.  

I'm hoping @luxas library can help with that in the future.

BenTheElder:
Fair.

In some respects this will behave a little more like `clusterctl` as we add things like #62, with a definition for each node (though nodes can be repeated).

Mixing number of node flags with specific node config won't go well, unless component config pulls some real magic :^)

Just logically the behavior for merging these is confusing. I think initially we can't allow mixing these flags with config.

If I do config:
```yaml
nodes:
- role: worker
  replicas: 3
  image: foo
- role: control-plane
  image: bar
```

`kind create cluster --config=config.yaml --num-worker-nodes=5`

Then what happens?

(with this PR `--num-worker-nodes` will be silently ignored in favor of the config)

neolit123:
i'm near 50/50 on this feature.

NOTE: i'm not seeing users complain about the lack of these flags and the kind config is fairly simple to use, so slightly leaning towards not having these flags in the stock kind CLI for now.

BenTheElder:
I'm going to +1 on we probably don't need this, we can revisit in the future if it becomes an issue

PR #250, opened 2019-01-27
run e2e with multiple nodes

BenTheElder:
We should move to more robust ochestration (#161 ?) for running e2es. In the meantime we now _need_ multiple nodes on kubernetes master, this PR fixes that.

### DIFF
--- a/hack/ci/e2e.sh
+++ b/hack/ci/e2e.sh
@@ -93,16 +93,49 @@ build() {
 
 # up a cluster with kind
 create_cluster() {
+    # create the config file
+    cat <<EOF > "${ARTIFACTS}/kind-config.yaml"
+# config for 1 control plane node and 2 workers
+# necessary for conformance
+kind: Config
+apiVersion: kind.sigs.k8s.io/v1alpha2
+nodes:
+# the control plane node
+- role: control-plane
+- role: worker
+  replicas: 2
+EOF
     # mark the cluster as up for cleanup
     # even if kind create fails, kind delete can clean up after it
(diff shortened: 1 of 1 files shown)


BenTheElder:
to be reviewed -- this fixes CI

hack/ci/e2e.sh
```diff
@@ -93,17 +93,50 @@ build() {
...
+
     # base kubetest args
     KUBETEST_ARGS="--provider=skeleton --test --check-version-skew=false"
 
+    # get the number of worker nodes
+    # TODO(bentheelder): this is kinda gross
+    NUM_NODES="$(kubectl get nodes \
+        -o=jsonpath='{range .items[*]}{.metadata.name}{"\t"}{.spec.taints}{"\n"}{end}' \
+        | grep -cv "node-role.kubernetes.io/master" \
+    )"
```
neolit123: do we want to add a helper command for that in kind?
given kubectl already prints detailed information, we might simply add output for how many CP and worker nodes there are:

e.g
```
kind get num-nodes control-plane // 1
kind get num-nodes worker // 2
kind get num-nodes // 3
```

as per our "what is a node in kind" discussion this should exclude external LBs.

BenTheElder: Possibly. An alternative options is that e2e.test could have fallback detection of this (I poked around and I don't think it does).. if it doesn't have that, this will be a problem for sonobuoy I think.

This didnt take long and should work pretty fine so I didn't dig further yet.

Issue #259, opened 2019-02-01
figure out how to handle --name flag for clusters


neolit123:
currently for cluster `delete` and `create` the `--name` flag defaults to `1`.
while the cluster name ends up being as `1` in the cluster list, the nodes are prefixed with `kind-1-`.

figure out a plan how to handle the name on the user side. some options.
- default cluster names to `kind-1` for `delete` and `create`
for kind the UX must be as simple as possible so defaulting to a cluster name makes sense.
- allow custom cluster names without the implicit `kind` prefix which will allow users to prefix nodes in a custom way as well.

optionally / possibly separate PRs:
- add methods to list nodes per cluster on the CLI.
- `delete cluster` without y/n is a bit scary.
for kubeadm we have a stdin prompt unless `--force` is passed.

/assign 
/priority important-soon
/kind design

mauilion:
However this works out it would be awesome to expose it as an env var. I frequently use tooling like direnv.net to set the configuration of the kubeadm-dind stuff and would very much like to persist this to the kind project.

BenTheElder:
can you use `kind create cluster --name=$KIND_NAME` ? 

the "name" is just use to differentiate creating multiple instances from a config (or the default config) we've kept almost everything else in config files for now.

we're leaning towards not using many unversioned flags.

neolit123:
i will possibly send a PR for the initial changes tomorrow.

BenTheElder:
fixed in #260, thanks!

PR #260, opened 2019-02-03
CLI improvements to create, delete and get

neolit123:
xref https://github.com/kubernetes-sigs/kind/issues/259

better view this PR per commit.
can split in multiple PRs if needed:

/kind cleanup
/kind feature
/priority important-longterm
/assign @BenTheElder

### DIFF
--- a/cmd/kind/get/nodes/nodes.go
+++ b/cmd/kind/get/nodes/nodes.go
@@ -1,0 +1,69 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package nodes implements the `nodes` command
+package nodes
+
+import (
+	"fmt"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+
+	"sigs.k8s.io/kind/pkg/cluster"
+	clusternodes "sigs.k8s.io/kind/pkg/cluster/nodes"
+)
+
+type flagpole struct {
+	Name string
+}
+
+// NewCommand returns a new cobra.Command for getting the list of nodes for a given cluster
+func NewCommand() *cobra.Command {
+	flags := &flagpole{}
+	cmd := &cobra.Command{
+		Args:  cobra.NoArgs,
+		Use:   "nodes",
+		Short: "lists existing kind nodes by their name",
+		Long:  "lists existing kind nodes by their name",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runE(flags, cmd, args)
+		},
+	}
+	cmd.Flags().StringVar(
+		&flags.Name,
+		"name",
+		cluster.DefaultName,
+		"the cluster context name",
+	)
+	return cmd
+}
+
+func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
+	// List nodes by cluster context name
+	n, err := clusternodes.ListByCluster()
+	if err != nil {
+		return err
+	}
+	nodes, known := n[flags.Name]
+	if !known {
+		return errors.Errorf("unknown cluster %q", flags.Name)
+	}
+	for _, node := range nodes {
+		fmt.Println(node.String())
+	}
+	return nil
+}
--- a/cmd/kind/get/clusters/clusters.go
+++ b/cmd/kind/get/clusters/clusters.go
@@ -20,7 +20,6 @@ package clusters
 import (
 	"fmt"
 
-	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
 	"sigs.k8s.io/kind/pkg/cluster"
@@ -44,7 +43,7 @@ func NewCommand() *cobra.Command {
 func runE(cmd *cobra.Command, args []string) error {
 	clusters, err := cluster.List()
 	if err != nil {
-		return errors.Wrap(err, "error listing clusters")
+		return err
 	}
 	for _, cluster := range clusters {
 		fmt.Println(cluster.Name())
--- a/cmd/kind/get/get.go
+++ b/cmd/kind/get/get.go
@@ -22,6 +22,7 @@ import (
 
 	"sigs.k8s.io/kind/cmd/kind/get/clusters"
 	"sigs.k8s.io/kind/cmd/kind/get/kubeconfigpath"
+	"sigs.k8s.io/kind/cmd/kind/get/nodes"
 )
 
 // NewCommand returns a new cobra.Command for get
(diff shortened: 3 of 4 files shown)


cmd/kind/delete/cluster/deletecluster.go
```diff
@@ -42,11 +45,24 @@ func NewCommand() *cobra.Command {
...
 	}
-	cmd.Flags().StringVar(&flags.Name, "name", "1", "the cluster name")
+	cmd.Flags().StringVar(&flags.Name, "name", cluster.DefaultName, "the cluster name")
+	cmd.Flags().BoolVar(&flags.Force, "force", false, "discard the check if a cluster exists")
 	return cmd
 }
 
 func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
+	// Check if the cluster name exists
+	if !flags.Force {
```
neolit123: if `create` was called but `ctrl+c`-ed immediately it will create a context with that name, but `delete` will not work on it.
so the clusters is in a weird state where both `create` and `delete` don't work on the name.

adding this `force` makes it possible to skip the check and force the deletion.
opened to ideas, but might remove this check.

adding TODO for `[y/n]` prompt` might be good too.

pkg/cluster/context.go
```diff
@@ -50,10 +50,10 @@ var validNameRE = regexp.MustCompile(`^[a-zA-Z0-9_.-]+$`)
 // DefaultName is the default Context name
 // TODO(bentheelder): consider removing automatic prefixing in favor
 // of letting the user specify the full name..
-const DefaultName = "1"
+const DefaultName = "kind-1"
```
BenTheElder: nit: why not just kind? 🤔

cmd/kind/delete/cluster/deletecluster.go
BenTheElder: > but delete will not work on it.

hmm? how so? delete will delete any tagged containers, and we don't leave anything behind without containers ...?

cmd/kind/get/nodes/nodes.go
```diff
@@ -0,0 +1,68 @@
...
+		Use:   "nodes",
+		Short: "lists existing kind nodes by their name",
+		Long:  "lists existing kind nodes by their name",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runE(flags, cmd, args)
+		},
+	}
+	cmd.Flags().StringVar(
+		&flags.ClusterName,
+		"cluster-name",
```
BenTheElder: hmm, I think we should use the same `--name` flag consistently

BenTheElder:
Most of this looks good, thanks. I'm not sure I understand the desire for `--force` though. 
The presence of a cluster is purely defined by the presence of containers tagged as that cluster. It either exists or it doesn't, and `delete` should already handle that properly. Did I miss something?

neolit123:
> The presence of a cluster is purely defined by the presence of containers tagged as that cluster. It either exists or it doesn't, and delete should already handle that properly. Did I miss something?

i was able to get it into a weird state:
- create says cluster name already exists.
- delete says cluster does not exists.

cannot reproduce this anymore, thus removed the "--force" flag in that context, yet added it back again because i've added a [y/n] confirmation prompt.
local e2e bash update, but the kind deployer in kubetest needs a update too.

BenTheElder:
If we got in that state that's due to a bug we need to fix. Delete should work even when the cluster is in a broken state.

neolit123:
> cannot reproduce this anymore

cmd/kind/delete/cluster/deletecluster.go
```diff
@@ -45,6 +47,7 @@ func NewCommand() *cobra.Command {
 		},
 	}
 	cmd.Flags().StringVar(&flags.Name, "name", cluster.DefaultName, "the cluster name")
+	cmd.Flags().BoolVarP(&flags.Force, "force", "f", false, "skip the confirmation prompt")
```
BenTheElder: Do similar tools require this in non-interactive mode?

I also have some limited concern that this encourages the wrong way of thinking about kind clusters -- they're meant to be highly disposable and shouldn't be storing any critical state. Accidentally deleting one should be OK.

neolit123: kubeadm does it for `reset` nowadays.

i see the possibility of kind being used for local development where a accidental delete can remove one's progress with no undo.

BenTheElder: > kubeadm does it for reset nowadays.

fair, but `kubeadm reset` seems like a different class of command. these clusters may not be ephemeral. 
It doesn't look like `minikube delete` has any confirmation logic.

> i see the possibility of kind being used for local development where an accidental delete can remove one's progress with no undo.

Hmm, I understand the sentiment but... what progress? That still sounds like important state is being preserved inside the kind cluster, rather than just a copy being tested on it, which is a practice we should discourage.

When we have #62 (expect a form of it by ~monday) preserving data this way should be preferable to ensuring the cluster continues to exist.

This also means that even the simplest CI scripts (`kind create cluster && do_my_test || kind delete cluster`) now need a flag to work correctly and will otherwise hang indefinitely. 🤔

neolit123: we could do a quick poll in `#kind`

BenTheElder: WDYT about moving this commit to a second PR?

neolit123: sure thing.

neolit123: separate PR sent here: https://github.com/kubernetes-sigs/kind/pull/261

cmd/kind/export/logs/logs.go
```diff
@@ -48,6 +49,14 @@ func NewCommand() *cobra.Command {
 }
 
 func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
+	// Check if the cluster name exists
+	known, err := cluster.IsKnown(flags.Name)
```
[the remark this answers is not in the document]
neolit123: yes, sounds like something for a different command.

pkg/cluster/context.go
```diff
@@ -96,9 +96,8 @@ func (c *Context) GetControlPlaneMeta() (*ControlPlaneMeta, error) {
 }
 
 // ClusterName returns the Kubernetes cluster name based on the context name
-// currently this is .Name prefixed with "kind-"
 func (c *Context) ClusterName() string {
-	return fmt.Sprintf("kind-%s", c.Name())
+	return c.Name()
```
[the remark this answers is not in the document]
neolit123: ok, leaving it for now.

neolit123:
PR updated.

cmd/kind/get/nodes/nodes.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"sigs.k8s.io/kind/pkg/cluster"
+	clusternodes "sigs.k8s.io/kind/pkg/cluster/nodes"
+)
+
+type flagpole struct {
+	Name string
+}
+
+// NewCommand returns a new cobra.Command for getting the list of clusters
```
neolit123: copy paste typo `list of clusters` -> `list of nodes`

BenTheElder: ah yeah
this won't auto-merge because netlify is still on and the working config is not in yet.. I can merge it as is or wait for this fix 🙃

neolit123: found another typo so i will send update.

neolit123: updated.

PR #261, opened 2019-02-03
support a confirmation prompt for `delete`

neolit123:
use by default a confirmation [y/n] prompt for `kind delete cluster`.
make it possible to skip the prompt by passing `-f` / `--force`.

/priority awaiting-more-evidence
/kind feature
/assign @BenTheElder

### DIFF
--- a/cmd/kind/delete/cluster/deletecluster.go
+++ b/cmd/kind/delete/cluster/deletecluster.go
@@ -18,6 +18,9 @@ limitations under the License.
 package cluster
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


alejandrox1:
I vote :+1: on this
It would come in handy for people who may have multiple kind clusters - prevent accidental erasures.
Plus it doesn't seem to clutter the delete code.

tao12345666333:
IMO the cluster created by kind is cheap. Maybe we don't need to add this mechanism.

neolit123:
we got mixed opinions on this feature between slack and github and most importantly the maintainers do not agree on this feature, thus i'm going to close this PR for now.

we might bring this back again if there is demand in on a later stage.

PR #267, opened 2019-02-05
README: add notes about HA and other `Node` fields

neolit123:
/kind documentation
/priority important-soon
/cc @alejandrox1 
/assign @BenTheElder 

fixes #268

### DIFF
--- a/site/content/docs/user/quick-start.md
+++ b/site/content/docs/user/quick-start.md
@@ -157,6 +157,10 @@ To specify a configuration file when creating a cluster, use the `--config`
 flag.
 For a sample kind configuration file see [kind-example-config][kind-example-config].
 
+The kind configuration object is outlined [here][godoc-config].
+It contains a list of nodes (`nodes`) that is composed of `Node` configuration objects.
+The `Node` configuration object is outlined [here][godoc-node].
+
 In particular, many users may be interested in multi-node clusters. A simple
 configuration for this can be acheived with the following config file contents:
 ```yaml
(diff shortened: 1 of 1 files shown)


docs/user/README.md
```diff
@@ -159,6 +159,45 @@ nodes:
...
+apiVersion: kind.sigs.k8s.io/v1alpha2
+nodes:
+- role: control-plane
+  image: mycontrolplane-node-image:latest
+- role: worker
+  image: myworker-node-image:latest
+  replicas: 2
+```
+
+The `Node` object also allows you to patch the kubeadm config that the node uses by applying
```
alejandrox1: nit: I think there should be a more explicit mention of what the `Node` object is. This reference kind of appears out of nowhere.
Maybe right below the first config file example add something like "each entry in 'nodes' represents a [`Node` object](ref to godoc)."

docs/user/README.md
```diff
@@ -159,6 +159,45 @@ nodes:
...
+apiVersion: kind.sigs.k8s.io/v1alpha2
+nodes:
+- role: control-plane
+  replicas: 3
+- role: worker
+  replicas: 3
+- role: external-load-balancer
+```
+
+The `external-load-balancer` node is [HAProxy][haproxy] based and it acts like a load
```
alejandrox1: nit: "...node is HAProxy based" sounds a bit awkward. Does this mean that the node image is based on a HAProxy image?

docs/user/README.md
```diff
@@ -159,6 +159,45 @@ nodes:
...
+balancer in front of the control-plane nodes. This node has to be added explicitly.
```
alejandrox1: Could we add some clarification on what it means when it says a node has to be added explicitly?
It isn't clear from a first read.

docs/user/README.md
BenTheElder: Currently the fact that it is haproxy based is pretty internal, we don't make any guarantees about this yet. It could very well be something else in the future

neolit123:
thanks for the comments, will try to address them later today.

neolit123:
will address the comments once we have a non-explicit LB.

docs/user/README.md
alejandrox1: I recall at some point hearing from the sig-cluster-lifecycle people that kind having a HAproxy LB was a big deal. This detail may be important but Lubomir would know more about it.

neolit123: i don't think it's a big deal. it's just a decent LB solution.

neolit123:
updated.

BenTheElder:
needs rebase, the main user guide got moved to being the quick-start guide. we should also add more sections to the user guide, perhaps this goes in a new advanced usage guide :sweat_smile:

neolit123:
> needs rebase,

updated

> the main user guide got moved to being the quick-start guide. we should also add more sections to the user guide, perhaps this goes in a new advanced usage guide

we can restructure in another PR.

site/content/docs/user/quick-start.md
```diff
@@ -169,6 +173,37 @@ nodes:
...
+use a custom image per node:
+```yaml
+# a setup with custom control-plane and worker node images
+kind: Config
+apiVersion: kind.sigs.k8s.io/v1alpha2
+nodes:
+- role: control-plane
+  image: mycontrolplane-node-image:latest
+- role: worker
+  image: myworker-node-image:latest
```
BenTheElder: we should put a concrete upstream kind image for each of these (EG the current default for one of them)

neolit123: we can do that, except that someone has to update this document on each release.

BenTheElder: It's OK if these get old, but it sends a better signal than suggesting totally custom images currently.
We also need to update these to the current API version each release ideally anyhow until we get out of alpha APIs.

BenTheElder:
https://deploy-preview-267--kindocs.netlify.com/

neolit123:
/hold
but this change is invalid without this:
https://github.com/kubernetes-sigs/kind/pull/274

BenTheElder:
er right, thanks. have some more thoughts on that, writing some more for this before the next meeting..

neolit123:
outdated.

Issue #136, opened 2018-11-26
overlay network cannot be applied when host is behind a proxy


senthilrch:
### Environment
Host OS: RHEL 7.4
Host Docker version: 18.09.0
Host go version: go1.11.2
Node Image: kindest/node:v1.12.2

### kind create cluster
```
[root@localhost bin]# kind create cluster
Creating cluster 'kind-1' ...
 ✓ Ensuring node image (kindest/node:v1.12.2) 🖼
 ✓ [kind-1-control-plane] Creating node container 📦
 ✓ [kind-1-control-plane] Fixing mounts 🗻
 ✓ [kind-1-control-plane] Starting systemd 🖥
 ✓ [kind-1-control-plane] Waiting for docker to be ready 🐋
 ✗ [kind-1-control-plane] Starting Kubernetes (this may take a minute) ☸
FATA[07:20:43] Failed to create cluster: failed to apply overlay network: exit status 1
```

Code below in pkg/cluster/context.go is trying to extract k8s version using kubectl version command in order to download the version-specific weave net.yaml. The code is not ok:-
```go
        // TODO(bentheelder): support other overlay networks
        if err = node.Command(
                "/bin/sh", "-c",
                `kubectl apply --kubeconfig=/etc/kubernetes/admin.conf -f "https://cloud.weave.works/k8s/net?k8s-version=$(kubectl version --kubeconfig=/etc/kubernetes/admin.conf | base64 | tr -d '\n')"`,
        ).Run(); err != nil {
                return kubeadmConfig, errors.Wrap(err, "failed to apply overlay network")
        }
```
Why is the output of kubectl version command, base64 encoded?

alejandrox1:
Hi @senthilrch , thank you for filling this issue.
Could you run `kind create cluster --loglevel debug` and put the output of this command in a [gist](https://gist.github.com/) ?

As for your uestion about installing weave net, you can read more about it [here](https://www.weave.works/docs/net/latest/kubernetes/kube-addon/).

BenTheElder:
Yep, as @alejandrox1 noted, the base64 encoding is from their guide. The reason for this is to pass it as an HTTP query parameter to weave so that their site can serve the appropriate weave version based on your Kubernetes version.

In the future we might use fixed weave versions, but this is the correct and normal way to install it per their upstream documentation.

> kubectl apply -f "https://cloud.weave.works/k8s/net?k8s-version=$(kubectl version | base64 | tr -d '\n')"

[...]

PR #275, opened 2019-02-06
Pass PROXY environment variables to nodes

### LINKED ISSUE #270: Configure nodes to use a proxy server
In order to solve #136 the `HTTP_PROXY` and `HTTPS_PROXY` environment variables should be passed from the host machine to the node container as suggested [here](https://github.com/kubernetes-sigs/kind/issues/136#issuecomment-459248404).

pablochacin:
Fixes #270 

Pass the environment variables HTTP_PROXY/HTTPS_PROXY from the host environment to the nodes, so that the docker running on the node could download images when running behind a proxy (see issue #136)

Signed-off-by: Pablo Chacin <pchacin@suse.com>

### DIFF
--- a/pkg/cluster/nodes/create.go
+++ b/pkg/cluster/nodes/create.go
@@ -19,6 +19,7 @@ package nodes
 import (
 	"fmt"
 	"net"
+	"os"
 
 	"github.com/pkg/errors"
 	"sigs.k8s.io/kind/pkg/cluster/config"
@@ -128,6 +129,16 @@ func createNode(name, image, clusterLabel string, role config.NodeRole, extraArg
 		"--label", fmt.Sprintf("%s=%s", constants.ClusterRoleKey, role),
 		// explicitly set the entrypoint
 		"--entrypoint=/usr/local/bin/entrypoint",
+	}
+
+	// pass proxy environment variables to be used by node's docker deamon
+	httpProxy := os.Getenv("HTTP_PROXY")
+	if httpProxy != "" {
+		runArgs = append(runArgs, "-e", "HTTP_PROXY="+httpProxy)
+	}
+	httpsProxy := os.Getenv("HTTPS_PROXY")
+	if httpsProxy != "" {
+		runArgs = append(runArgs, "-e", "HTTPS_PROXY="+httpsProxy)
 	}
 
 	// adds node specific args


pkg/cluster/nodes/create.go
```diff
@@ -130,6 +131,15 @@ func createNode(name, image, clusterLabel string, role config.NodeRole, extraArg
...
 	}
 
+	// pass proxy environment variables to be used by node's docker deamon
+	if os.Getenv("HTTP_PROXY")!= "" {
+	  runArgs = append(runArgs,"-e HTTP_PROXY=",os.Getenv("HTTP_PROXY"))
+	}
+
+	if os.Getenv("HTTPS_PROXY") != "" {
+	  runArgs = append(runArgs,"-e HTTPS_PROXY=",os.Getenv("HTTPS_PROXY"))
+	}
```
neolit123: please run `./hack/update-gofmt.sh` and include the changes to be able to pass CI.
if you are using editors like VS Code it should format the code correctly for you.

non-blocking nit would be to store the ENV variables into intermediate variables - e.g.:
```
http_proxy := getenv(...)
if http_proxy != "" {
    ....
}
https_proxy := getenv(...)
if https_proxy != "" {
    ....
}
```

pablochacin:
> thanks for the update @pablochacin
> please squash the commits. this looks good, unless @BenTheElder has comments?

Done

pkg/cluster/nodes/create.go
```diff
@@ -130,6 +131,16 @@ func createNode(name, image, clusterLabel string, role config.NodeRole, extraArg
 		"--entrypoint=/usr/local/bin/entrypoint",
 	}
 
+	// pass proxy environment variables to be used by node's docker deamon
+	httpProxy := os.Getenv("HTTP_PROXY")
+	if httpProxy != "" {
+		runArgs = append(runArgs, "-e HTTP_PROXY=", httpProxy)
```
BenTheElder: shouldn't this be 
```suggestion
		runArgs = append(runArgs, "-e", "HTTP_PROXY="+httpProxy)
```
?

pablochacin: Yes. Much clearer. Thanks.

BenTheElder:
[I think netlify fails because it doesn't merge the branch into master before building, so this PR branch doesn't contain the netlify configs...]

BenTheElder:
[manual merge due to bogus netlify failures]

unautre:
Hello,
I am in a similar situation, with a corporate HTTP proxy for outside traffic. After re-installing kind to the latest version (after that merge), it works better.
There is still a problem though, in that kubernetes tries to access cluster-local URLs (172.17.0.2 in my case) through the proxy. Adding those URLs to the NO_PROXY helps kind create the cluster to completion.
Would it be feasible to automatically add local URLs to the NO_PROXY variable ?

Same problem a bit later, the cluster is started, but looking at the logs (kind export logs) it says repeatedly that it's trying to pull a docker image, and failing ; I guess the docker service (inside the container) is not configured to use the proxy.

BenTheElder:
> Would it be feasible to automatically add local URLs to the NO_PROXY variable ?

Yeah we could add the clusters IPs to this, it would take a bit of work to compute these.

It also looks like instead users can set https://docs.docker.com/network/proxy/#configure-the-docker-client on the host side before kind creates any clusters.

PR #277, opened 2019-02-06
add a flag to configure docker network

### LINKED ISSUE #273: Allow configurable docker network for kind cluster nodes
It would be nice to be able to specify the network that the cluster uses.

foolusion:
I've implemented a minimal solution for adding a docker network.
This adds a flag `--network` to the create cluster command.

It will error if the network doesn't exist.
If the network flag is not passed it uses the default bridge network.

We can work from here on design and implementation.

resolves #273

### DIFF
--- a/cmd/kind/create/cluster/createcluster.go
+++ b/cmd/kind/create/cluster/createcluster.go
@@ -32,6 +32,7 @@ import (
 
 type flagpole struct {
 	Name      string
+	Network   string
 	Config    string
 	ImageName string
 	Retain    bool
@@ -51,6 +52,7 @@ func NewCommand() *cobra.Command {
 		},
 	}
 	cmd.Flags().StringVar(&flags.Name, "name", cluster.DefaultName, "cluster context name")
+	cmd.Flags().StringVar(&flags.Network, "network", "", "network to use for cluster")
 	cmd.Flags().StringVar(&flags.Config, "config", "", "path to a kind config file")
 	cmd.Flags().StringVar(&flags.ImageName, "image", "", "node docker image to use for booting the cluster")
 	cmd.Flags().BoolVar(&flags.Retain, "retain", false, "retain nodes for debugging when cluster creation fails")
@@ -87,6 +89,9 @@ func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
 
 	// create a cluster context and create the cluster
 	ctx := cluster.NewContext(flags.Name)
+	if flags.Network != "" {
+		ctx.SetNetwork(flags.Network)
+	}
 	if flags.ImageName != "" {
 		// Apply image override to all the Nodes defined in Config
 		// TODO(fabrizio pandini): this should be reconsidered when implementing
--- a/cmd/kind/create/cluster/createcluster.go
+++ b/cmd/kind/create/cluster/createcluster.go
@@ -52,7 +52,7 @@ func NewCommand() *cobra.Command {
 		},
 	}
 	cmd.Flags().StringVar(&flags.Name, "name", cluster.DefaultName, "cluster context name")
-	cmd.Flags().StringVar(&flags.Network, "network", "", "network to use for cluster")
+	cmd.Flags().StringVar(&flags.Network, "network", "", "container network to use for cluster")
 	cmd.Flags().StringVar(&flags.Config, "config", "", "path to a kind config file")
 	cmd.Flags().StringVar(&flags.ImageName, "image", "", "node docker image to use for booting the cluster")
 	cmd.Flags().BoolVar(&flags.Retain, "retain", false, "retain nodes for debugging when cluster creation fails")
@@ -88,10 +88,7 @@ func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
 	}
 
 	// create a cluster context and create the cluster
-	ctx := cluster.NewContext(flags.Name)
-	if flags.Network != "" {
-		ctx.SetNetwork(flags.Network)
-	}
+	ctx := cluster.NewContext(flags.Name, cluster.WithNetwork(flags.Network))
 	if flags.ImageName != "" {
 		// Apply image override to all the Nodes defined in Config
 		// TODO(fabrizio pandini): this should be reconsidered when implementing
--- a/pkg/cluster/context.go
+++ b/pkg/cluster/context.go
@@ -54,12 +54,16 @@ const DefaultName = "kind"
 
 // NewContext returns a new cluster management context
 // if name is "" the default name will be used
-func NewContext(name string) *Context {
+func NewContext(name string, options ...Option) *Context {
 	if name == "" {
 		name = DefaultName
 	}
+	var c contextConfig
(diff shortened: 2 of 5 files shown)


pkg/cluster/nodes/create.go
```diff
@@ -48,18 +48,23 @@ func getPort() (int, error) {
...
 
-	node, err = createNode(name, image, clusterLabel, config.ControlPlaneRole,
+	extraArgs := []string{
 		// publish selected port for the API server
 		"--expose", fmt.Sprintf("%d", port),
 		"-p", fmt.Sprintf("%d:%d", port, kubeadm.APIServerPort),
-	)
+	}
+	if network != "" {
+		extraArgs = append(extraArgs, "--network", network)
```
neolit123: this binds a kind CLI flag to the underlying CR (docker).
i guess we need to evaluate if we want this coupling for the 1.0 release.

pkg/cluster/nodes/create.go
```diff
@@ -48,18 +48,23 @@ func getPort() (int, error) {
...
+	}
+
```
neolit123: optional nit: we can omit this empty line.
[1]

pkg/cluster/nodes/create.go
```diff
@@ -72,18 +77,23 @@ func CreateControlPlaneNode(name, image, clusterLabel string) (node *Node, err e
...
+	extraArgs := []string{
 		// publish selected port for the control plane
 		"--expose", fmt.Sprintf("%d", port),
 		"-p", fmt.Sprintf("%d:%d", port, haproxy.ControlPlanePort),
-	)
+	}
+	if network != "" {
+		extraArgs = append(extraArgs, "--network", network)
+	}
+
```
neolit123: [1]

pkg/cluster/internal/meta/meta.go
```diff
@@ -44,6 +45,14 @@ func (c *ClusterMeta) Name() string {
 	return c.name
 }
 
+func (c *ClusterMeta) Network() string {
+	return c.network
+}
+
+func (c *ClusterMeta) SetNetwork(network string) {
```
BenTheElder: cluster meta is not supposed to be mutable.

I should really clean up how create works :|

pkg/cluster/nodes/create.go
BenTheElder: yeah, i'm iffy on this. 🤔 
I this probably does make sense as a flag, I guess, but as a rule of thumb we should prefer config... 🤔

cmd/kind/create/cluster/createcluster.go
```diff
@@ -51,6 +52,7 @@ func NewCommand() *cobra.Command {
 		},
 	}
 	cmd.Flags().StringVar(&flags.Name, "name", cluster.DefaultName, "cluster context name")
+	cmd.Flags().StringVar(&flags.Network, "network", "", "network to use for cluster")
```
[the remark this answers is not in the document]
foolusion: I checked out a few of the other container runtimes.

rkt has a `--net` flag that you can pass one or more networks I believe it uses CNI.
docker seems to also support the shorter `--net` flag

CRI-O, containerd do not support networking directly, you must configure it through runc or CNI.

pkg/cluster/internal/meta/meta.go
foolusion: Ah, I didn't want to add anything to the `NewClusterMeta` func yet since it would require more changes. I can add it as a variadic param to NewClusterMeta since network is not required and existing code doesn't break.

cmd/kind/create/cluster/createcluster.go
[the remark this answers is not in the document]
neolit123: thanks for checking.
CRI-O and containerd are the 2 more active CR after docker in the k8s space, while rkt is not used a lot.

pkg/cluster/nodes/create.go
foolusion: If it was in the config would it be at the top level?
Is there any setup where your nodes would be in different networks?

BenTheElder: sorry, time flew by so fast :(

i'm not sure.. I could see this maybe, we could start top level and do an override.

Either way we're mostly decoupled from docker at the user interface right now, and I do intend to support podman etc., I meant to investigate further if there was a better way to specify this or any other work arounds.

I do see how this is a feature we need to enable.

pkg/cluster/context.go
```diff
@@ -207,3 +211,17 @@ func (c *Context) CollectLogs(dir string) error {
...
+
+type contextConfig struct {
+	network string
+}
+
+// Option allows optional configuration of the cluster context.
+type Option func(*contextConfig)
+
+// WithNetwork sets the container network to be used when creating a cluster.
+func WithNetwork(network string) func(*contextConfig) {
```
BenTheElder: is there any reason this should be an option on context instead of Create()? do you forsee other calls needing this?

:+1: for this pattern in general.

pkg/cluster/internal/meta/meta.go
BenTheElder: er what I really meant by this, is that clusterMeta is a terrible name, this is really just the cluster key (name) and some utils right now, I suspect Create(...) is the right place to plumb this through.

BenTheElder:
I think we should add this in the `network` struct proposed in #340 instead.

pkg/cluster/context.go
foolusion: I didn't know if anything else was calling Create so I didn't want to change the parameters. Either way works. I haven't dug into the code enough to see how/where it is used.

foolusion:
> I think we should add this in the `network` struct proposed in #340 instead.

Would the only change be removing the flag and setting the value from the config or do you have a different idea?

DazWorrall:
Out of scope for this PR but I wanted to mention it somewhere: for our use case being able to use `--net-alias` is also important, so somewhere to hang container-runtime specific configuration would be really useful.

foolusion:
I realized that I can just join the kind containers to a custom network after creation. For example `docker network connect my-custom-network kind-control-plane` Maybe that makes this PR less useful. Wondering what your thoughts are @BenTheElder @neolit123

BenTheElder:
Apologies -- I've been heads down on a big refactor (#360) of cluster creation to make extending this sort of thing easier hopefully... I hoped to get it in before too many PRs adding more to the existing cluster creation code piled up. (should be more or less going in now finally..)

I think a field in config for runtime specific options makes sense, ideally we should continue to do this one step above flags so we at least leave eg the docker client library open as an option in the future:

maybe:
```yaml
kind: Cluster
dockerOptions:
  net: "foo"
   ...
```

BenTheElder:
superseded by https://github.com/kubernetes-sigs/kind/pull/484, it turns out there are some other issues with using a non-default docker network (the behavior changes on docker's end a bit and we need to work around it)

thanks again for the PR, hopefully some variant of this feature will be in 0.4

PR #268, opened 2019-02-05
initial netlify + hugo setup

BenTheElder:
This is extremely minimal and rough -- I opted to start by borrowing a few ideas from my past experience with hugo just to get something rough up and running.

EDIT: once the netlify preview is working I will add some more CSS, meta-docs, etc.

PR #279, opened 2019-02-07
Adding one more related project

mitar:


BenTheElder:
will also need to capture in https://github.com/kubernetes-sigs/kind/pull/268.

BenTheElder:
[manual merge due to netlify failure blocking the robot, xref #268]

mitar:
If you are reordering things in #268 I would suggest this one in this PR is also put towards the top because it is very aligned/similar to this project as well.

BenTheElder:
ordering in #268 is based on visual length, otherwise kubeadm-dind-cluster
would be at the top as the closest, what I meant by capturing it there is
also in the site since that copies from the readme.

Issue #280, opened 2019-02-07
IPv6 Support


Arvinderpal:
Kubernetes currently lacks IPv6 CI support. Some work has been done to add such support via DinD here (https://github.com/kubernetes/test-infra/pull/7529). However, similar support in KinD is very desirable. 

Furthermore, there is an effort in the community to bring dual-stack support to kubernetes (https://github.com/kubernetes/enhancements/pull/648). Having an IPv6 CI in place will greatly aid that work. 

Full IPv6 support in KinD can progress in small steps, which I will try to outline below. 
	1. Add support for other CNIs (beyond weave); in particular the bridge and host-local IPAM plugins. Bridge and host-local IPAM are both simplistic which is desirable from a CI point of view as plugin specific problems can be minimized. Both plugins are under the CNCF umbrella (https://github.com/containernetworking) and have had IPv6 support for sometime. 
	2. Support bridge+host-local on a single node with IPv4 addressing. 
	3. Support bridge+host-local on a single node with IPv6 addressing. 
	4. Support bridge+host local on a multi-node cluster with IPv4 addressing.
	5. Suppor bridge+host-local on a multi-node cluster with IPv6 addressing.  

I would envision the above features coming in a as a series of small PRs. I have wip that brings both 1 and 2 above to KinD. I will share that shortly and would love to get feedback.

BenTheElder:
We are tracking / discussing 1.) in #205 / #278 FYI.

In general I would love to see this 🙃

neolit123:
kubeadm _should be_ ready for this, but don't quote me on that. we had some recent PRs that fixed some IPv4 assumptions last cycle.

but we don't have test signal for IPv6, so /shrug

Arvinderpal:
Hi @BenTheElder @neolit123 
Here is one approach https://github.com/kubernetes-sigs/kind/pull/281 
Please have a look. It's just a wip that I thought I share to get early feedback. I'm sure there are a multitude of ways we could go about doing this. :)

Arvinderpal:
Hmm. Looks like I need to sign the CLA before you guys can look at the PR above :(
I'll try to get this resolved asap with my company.

aojea:
Google doc with a proposal to support IPb6 and Dual stack clusters in kind https://docs.google.com/document/d/17e3TWWLfnIZrsVxpln9wNi4x0JVn2oHIHDYjaeENdVE/edit?usp=sharing

BenTheElder:
[...]

Issue #278, opened 2019-02-07
provide an option to bring up a cluster without a cni installed.


mauilion:
This option will enable the user to deploy their choice of cni in advance of kind supporting a variety of cni implementations.

neolit123:
we should bring the CNI topic on monday.

my proposal ATM is the following:
- add optional skip of the CNI phase in kind.
- let the user call `kubectl apply` their preferred plugin.

related topic is enabling offline support for a set of CNI plugins.
my proposal there is:
- we could bundle images and manifests for the top 3 - flannel, weave, callico (as per the kubeadm survey)
or decide to only bundle weave and force the rest of the users to do the alternative in the previous proposal (skip, kubeclt).

BenTheElder:
-1 on shipping all 3 unless they are small, our images are big.

I've been working on a rough design for alternate images, and looking at other options for pre-loading / caching (eg a local image mirror).

The additional thing to skipping the overlay apply step in kind is we also need to skip the node readiness check / `--wait` will be unsupported, but this is likely an acceptable tradeoff for applying arbitrary CNI.

The other reason to consider doing this is EG Cillium, which we simply can't ship by default currently because it relies on kernel modules and we need to support docker for mac / windows + docker-ce on random distros.

neolit123:
> -1 on shipping all 3 unless they are small, our images are big.

i think they are roughly 5-10 MB.

> The additional thing to skipping the overlay apply step in kind is we also need to skip the node readiness check / `--wait` will be unsupported, but this is likely an acceptable tradeoff for applying arbitrary CNI.

that is true.

> The other reason to consider doing this is EG Cillium, which we simply can't ship by default currently because it relies on kernel modules and we need to support docker for mac / windows + docker-ce on random distros.

flannel is a step ahead on the Windows front.
but the functionally is lacking documentation and windows images and node selectors are missing in their official manifests.

in fact i just had to patch their 0.11.0 manifest a couple of days ago.

BenTheElder:
er docker for windows != windows images, still linux images. (we only support linux containers for now.... windows nodes may be interesting someday)

neolit123:
ah, my confusion. windows containers are out of scope for now.
also quite crazy due to the image sizes.

Arvinderpal:
[...]

PR #281, opened 2019-02-08
Support for CNCF containernetworking plugins

Arvinderpal:
Add support for CNCF containernetworking plugins; in particular, bridge and host-local ipam.
This approach extends KinD's Config with 3 additional fields: CNI, Config File Name and Config File Body.
If these fields are specified then KinD will copy the config file to the /etc/cni/net.d directory under the desired name.
In theory, this approach should work with other plugins that only require a configuration file and their binaries have already been placed in the /opt/cni/bin directory.
** WIP **

Addresses the following issue: https://github.com/kubernetes-sigs/kind/issues/280

### DIFF
--- a/docs/user/kind-bridge-ipv4-config.yaml
+++ b/docs/user/kind-bridge-ipv4-config.yaml
@@ -1,0 +1,41 @@
+# this config file contains all config fields with comments
+kind: Config
+apiVersion: kind.sigs.k8s.io/v1alpha2
+# 1 control plane node and 3 workers
+nodes:
+# the control plane node config
+- role: control-plane
+  cni: cncf-containernetworking-plugin
+  cniConfigFileName: 100-bridge.conf
+  cniConfig:
+  - |
+    {
+        "cniVersion": "0.3.0",
+        "name": "mynet",
+        "type": "bridge",
+        "bridge": "cni0",
+        "isDefaultGateway": true,
+        "ipMasq": true,
+        "ipam": {
+            "type": "host-local",
+            "subnet": "10.0.0.0/16"
+        }
+    }
+  # patch the generated kubeadm config with some extra settings
+  kubeadmConfigPatches:
+  - |
+    apiVersion: kubeadm.k8s.io/v1beta1
+    kind: ClusterConfiguration
+    networking:
+      serviceSubnet: 192.168.0.0/16
+      podSubnet: 10.0.0.0/10
+  # patch it further using a JSON 6902 patch
+  kubeadmConfigPatchesJson6902:
+  - group: kubeadm.k8s.io
+    version: v1beta1
+    kind: ClusterConfiguration
+    patch: |
+      - op: add
+        path: /apiServer/certSANs/-
+        value: my-hostname
+
--- a/pkg/cluster/config/types.go
+++ b/pkg/cluster/config/types.go
@@ -43,6 +43,12 @@ type Node struct {
 	// Role defines the role of the nodw in the in the Kubernetes cluster managed by `kind`
 	// Defaults to "control-plane"
 	Role NodeRole
+	// CNI is the name of the desired network plugin
+	CNI string
+	// CNIConfigFIleName is the name of the config file that will be placed in /etc/cni/net.d
+	CNIConfigFileName string
+	// CNIConfig is the actual content inside the cni config file.
+	CNIConfig []string
 	// Image is the node image to use when running the cluster
 	// TODO(bentheelder): split this into image and tag?
 	Image string
@@ -74,4 +80,6 @@ const (
 	// WARNING: this node type is not yet implemented!
 	// Please note that `kind` nodes hosting external load balancer are not kubernetes nodes
 	ExternalLoadBalancerRole NodeRole = "external-load-balancer"
+	// Default path where the CNI configuration(s) are placed.
+	CNIConfigPath = "/etc/cni/net.d/"
 )
--- a/pkg/cluster/config/v1alpha2/types.go
+++ b/pkg/cluster/config/v1alpha2/types.go
@@ -42,6 +42,12 @@ type Node struct {
 	// Role defines the role of the nodw in the in the Kubernetes cluster managed by `kind`
 	// Defaults to "control-plane"
 	Role NodeRole `json:"role,omitempty"`
+	// CNI is the name of the desired network plugin
+	CNI string `json:"cni,omitempty"`
+	// CNIConfigFIleName is the name of the config file that will be placed in /etc/cni/net.d
+	CNIConfigFileName string `json:"cniConfigFileName,omitempty"`
+	// CNIConfig is the actual content inside the cni config file.
+	CNIConfig []string `json:"cniConfig,omitempty"`
 	// Image is the node image to use when running the cluster
 	// TODO(bentheelder): split this into image and tag?
 	Image string `json:"image,omitempty"`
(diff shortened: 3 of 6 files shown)


Arvinderpal:
You can try out the bridge+host-local plugin as so:
kind create cluster --config docs/user/kind-bridge-ipv4-config.yaml

pkg/cluster/config/types.go
```diff
@@ -74,4 +80,6 @@ const (
 	// WARNING: this node type is not yet implemented!
 	// Please note that `kind` nodes hosting external load balancer are not kubernetes nodes
 	ExternalLoadBalancerRole NodeRole = "external-load-balancer"
+	// Default path where the CNI configuration(s) are placed.
+	CNIConfigPath = "/etc/cni/net.d/"
```
neolit123: this probably needs separate windows / linux constants.
having said that i have no idea what the recommended path for windows is.
we can ask in #sig-windows on slack.

pkg/cluster/config/types.go
```diff
@@ -43,6 +43,12 @@ type Node struct {
 	// Role defines the role of the nodw in the in the Kubernetes cluster managed by `kind`
 	// Defaults to "control-plane"
 	Role NodeRole
+	// CNI is the name of the desired network plugin
+	CNI string
+	// CNIConfigFIleName is the name of the config file that will be placed in /etc/cni/net.d
```
neolit123: typo in comment
probably might want to omit the path and say e.g. "CNI plugin path".

pkg/cluster/config/types.go
```diff
@@ -43,6 +43,12 @@ type Node struct {
 	// Role defines the role of the nodw in the in the Kubernetes cluster managed by `kind`
 	// Defaults to "control-plane"
 	Role NodeRole
+	// CNI is the name of the desired network plugin
+	CNI string
```
neolit123: if we are going to add these they should be under a CNI object and here it should be a pointer so that we can use omitifempty with conjunction with the types.go from v1alpha2

docs/user/kind-bridge-ipv4-config.yaml
```diff
@@ -0,0 +1,41 @@
+# this config file contains all config fields with comments
```
neolit123: if this file is added should it be linked from the docs?
possibly a separate PR should add this file and link to docs?

pkg/cluster/internal/create/kubeadm-init.go
```diff
@@ -140,3 +167,20 @@ func addDefaultStorageClass(controlPlane *nodes.Node) error {
 	cmd.SetStdin(in)
 	return cmd.Run()
 }
+
+func createCNIConfigFile(config []string) (path string, err error) {
+	f, err := ioutil.TempFile("", "")
+	if err != nil {
+		return "", errors.Wrap(err, "failed to create local file for cni config")
+	}
+	path = f.Name()
```
neolit123: use `defer os.Remove(path)` instead of the later `os.Remove(path)` on error.

neolit123:
you can sign the CLA @Arvinderpal 
just follow the instructions the bot gave and after you are done call this command `/check-cla`

neolit123:
> you can sign the CLA @Arvinderpal

ok, looks like you signed it already. :)

pkg/cluster/nodes/node.go
```diff
@@ -96,6 +96,10 @@ func (n *Node) CopyFrom(source, dest string) error {
 	return docker.CopyFrom(n.nameOrID, source, dest)
 }
 
+func (n *Node) MkDirPath(dirPath string) error {
```
BenTheElder: this seems awfully specific?

pkg/cluster/config/v1alpha2/types.go
```diff
@@ -42,6 +42,12 @@ type Node struct {
 	// Role defines the role of the nodw in the in the Kubernetes cluster managed by `kind`
 	// Defaults to "control-plane"
 	Role NodeRole `json:"role,omitempty"`
+	// CNI is the name of the desired network plugin
+	CNI string `json:"cni,omitempty"`
```
BenTheElder: are we going to put different CNI per node?

pkg/cluster/config/types.go
Arvinderpal: That's a good idea. So something like?
type CNI struct{
CNIType string 
CNIConfigFileName string
CNIConfig []string
}

docs/user/kind-bridge-ipv4-config.yaml
Arvinderpal: Perhaps an "examples" folder would be more appropriate. We will likely have additional configurations for different scenarios -- e.g. IPv4/IPv6 single/multi node with various cni plugins.

pkg/cluster/config/v1alpha2/types.go
Arvinderpal: That's a good point. 
I don't think that's a very likely scenario, but at the same time I don't believe there is any strict requirement that all nodes in a cluster use the same CNI plugin. For example, a particular set of nodes may have specialized NICs for which a different CNI plugin may be desirable.

neolit123: we should definitely move the CNI object outside of the Node and into the main Config object.
CNI should be per cluster.

Arvinderpal:
Thanks for the feedback guys!
I would love to help out with this effort.

Arvinderpal:
I accidentally closed the PR. My bad!

pkg/cluster/config/v1alpha2/types.go
Arvinderpal: @neolit123 CNIConfig can contain node specific data. For example, the host-local ipam consists of fields (e.g. subnet, gateway, routes) that are node specific (https://github.com/containernetworking/plugins/tree/master/plugins/ipam/host-local) 
I suspect that this will not play well with Replicas since nodes will not be exact copies of each other. However, this is something that can be documented.

BenTheElder:
Sorry I missed this:
> I would love to help out with this effort. On that note, are there any periodic syncs/meetings where such topics are discussed?

There's a weekly SIG-Testing subproject meeting.
https://github.com/kubernetes/community/blob/master/sig-list.md

pkg/cluster/config/v1alpha2/types.go
BenTheElder: > For example, a particular set of nodes may have specialized NICs for which a different CNI plugin may be desirable.

But these are not on real NICs.

I also think having CNI config is perhaps not the right level of complexity for kind. Most CNI are deployable as a k8s manifest, I think we can support that or not installing a CNI at all and leaving it to the end user. Ref https://github.com/kubernetes-sigs/kind/issues/278

BenTheElder:
Apologies for the delay
/priority backlog
I'd prefer to let this come after the next release. So far we're testing Kubernetes / kubeadm, and apps on top. Testing CNIs with kind is neat but originally unplanned territory and we're trying to stabilize the config in particular so I'd like to think a bit more before adding anything to it.

Arvinderpal:
Hi, no worries. This does bloat the Config so if there is a cleaner solution that would be preferred. Most if not all major CNIs provide manifests. With these smaller CNIs however, there are no such manifests available but in theory it should be possible to create them. I'll look into it. :)

BenTheElder:
SGTM, let's do a new PR moving this into EG the node image build, or override the manifest in config (and just depend on pulling the images) ref: #340.

Going forward the manifest is now not pulled from weave's site except as a fallback for old images, it is pulled from the node instead.

PR #286, opened 2019-02-08
docker: Remove unneeded quotes from the inspect format string

dlespiau:
I'm guessing those quotes come from the command line examples where you have to
quote the go template:

  $ docker inspect --format='{{.LogPath}}' $INSTANCE_ID

However, we directly give the array of strings for the exec call so we don't
need to use the shell quotes here.

This removes the need to trim the quotes at the caller sites.


BenTheElder:
I don't think this can pass netlify without being rebased against master so I'm just going to manually merge it.

Thanks for the PR!

PR #291, opened 2019-02-11
add CC-BY-4.0 LICENSE to site

BenTheElder:
This follows https://github.com/kubernetes/website and our site footer.


BenTheElder:
This is a direct copy from https://github.com/kubernetes/website/blob/master/LICENSE to be more specific.

Issue #263, opened 2019-02-04
build new images


BenTheElder:
Since we don't have #197 we should push & upgrade the default image to 1.13.3 for the 0.2 release.
Long term we should definitely get #197

BenTheElder:
/lifecycle active
doing this now

BenTheElder:
images have been pushed, also updated the release notes

PR #292, opened 2019-02-11
Upgrade images

BenTheElder:
ref: #263 
This also picks up docker 18.06.2 regarding the runc escape. I am similarly pushing new images to previous default Kubernetes versions kind has shipped.

We should consider reversing course and unpinning the default from the image tag. CI could always explicitly select an exact image. For this PR though I am simply upgrading the images, as upgrading the base image it the most important change

/assign @neolit123


BenTheElder:
updated https://github.com/kubernetes-sigs/kind/releases/tag/0.0.1, will send out a note shortly.

Issue #55, opened 2018-10-03
Unstable cluster


mitar:
When running it locally on my machine the cluster seems much more unstable than on our CI. So now cluster is created inside a privileged container, but then I am getting strange errors:

```
$ kubectl cluster-info
Unable to connect to the server: unexpected EOF
```

```
$ kubectl cluster-info
Error from server (InternalError): an error on the server ("") has prevented the request from succeeding (get services)
```

```
$ kubectl cluster-info 
error: the server doesn't have a resource type "services"
```

BenTheElder:
Can you provide more details about how you're running this?

Creating the cluster locally should not require anything more than:
- have docker running
- `kind create`

I've tested regularly both on docker for mac and on docker-ce on linux. Currently it doesn't depend on any recent or advanced functionality. Sticking it in another layer of docker will likely make things less debuggable.

mitar:
So I am running inside a Docker container to simulate how my CI (gitlab) works. So it seems it just dies after some time. I am running it [more or less like this](https://github.com/kubernetes-sigs/kind/issues/52#issuecomment-426550163).

So I am not sure what to inspect when this errors start happening? I can exec into `kind-1-control-plane`?

BenTheElder:
right now `docker exec kind-1-control-plane journalctl > logs.txt` works, but I'm working on a nice `kind` command for this soon... stuck on go1.11 upgrade issues for kubernetes/kubernetes right now, xref https://github.com/kubernetes/test-infra/pull/9695

BenTheElder:
Those errors look like the API server is not actually running or the networking is not pointing at it correctly again. It's not clear from that output what else is going on beyond that yet. 

`docker exec` (not the prettiest since the container names are technically not really exposed yet) can get you onto the "node" after which normal debian debugging tools should generally work (`ps`, `journalctl`, etc.)

mitar:
OK, it is failing even if I run it directly on my laptop/host.

mitar:
See log: [log.txt](https://github.com/kubernetes-sigs/kind/files/2444414/log.txt)

BenTheElder:
Can you provide more details about your laptop/host? Docker version? Any special network settings? 

I think we'll need the API server logs as well. That will be in a location like `/var/log/containers/kube-apiserver-kind-1-control-plane.*.log` Again, I'll be adding a tool to collect these shortly.. :grimacing:

mitar:
[...]

PR #293, opened 2019-02-12
disable disk eviction by default

BenTheElder:
hopefully resolves https://github.com/kubernetes-sigs/kind/issues/55

with this PR the default configuration for kind nodes simply won't have hard eviction.

### DIFF
--- a/pkg/cluster/internal/kubeadm/config.go
+++ b/pkg/cluster/internal/kubeadm/config.go
@@ -90,6 +90,17 @@ apiServerExtraVolumes:
 # so we need to ensure the cert is valid for localhost so we can talk
 # to the cluster after rewriting the kubeconfig to point to localhost
 apiServerCertSANs: [localhost]
+kubeletConfiguration:
+  baseConfig:
+    # disable disk resource management by default
+    # kubelet will see the host disk that the inner container runtime
+    # is ultimately backed by and attempt to recover disk space.
+    # we don't want that.
+    imageGCHighThresholdPercent: 100
+    evictionHard:
+      nodefs.available: "0%"
+      nodefs.inodesFree: "0%"
+      imagefs.available: "0%"
 `
 
 // ConfigTemplateAlphaV3 is the kubadm config template for API version v1alpha3
@@ -128,9 +139,16 @@ apiEndpoint:
 apiVersion: kubeadm.k8s.io/v1alpha3
 kind: JoinConfiguration
 ---
-# no-op entry that exists soley so it can be patched
 apiVersion: kubelet.config.k8s.io/v1beta1
 kind: KubeletConfiguration
+# disable disk resource management by default
+# kubelet will see the host disk that the inner container runtime
+# is ultimately backed by and attempt to recover disk space. we don't want that.
+imageGCHighThresholdPercent: 100
+evictionHard:
+  nodefs.available: "0%"
+  nodefs.inodesFree: "0%"
+  imagefs.available: "0%"
 ---
 # no-op entry that exists soley so it can be patched
 apiVersion: kubeproxy.config.k8s.io/v1alpha1
@@ -166,9 +184,16 @@ localAPIEndpoint:
 apiVersion: kubeadm.k8s.io/v1beta1
 kind: JoinConfiguration
 ---
-# no-op entry that exists soley so it can be patched
 apiVersion: kubelet.config.k8s.io/v1beta1
 kind: KubeletConfiguration
+# disable disk resource management by default
+# kubelet will see the host disk that the inner container runtime
+# is ultimately backed by and attempt to recover disk space. we don't want that.
+imageGCHighThresholdPercent: 100
+evictionHard:
+  nodefs.available: "0%"
+  nodefs.inodesFree: "0%"
+  imagefs.available: "0%"
 ---
 # no-op entry that exists soley so it can be patched
 apiVersion: kubeproxy.config.k8s.io/v1alpha1


pkg/cluster/internal/kubeadm/config.go
```diff
@@ -169,6 +181,11 @@ kind: JoinConfiguration
 # no-op entry that exists soley so it can be patched
 apiVersion: kubelet.config.k8s.io/v1beta1
 kind: KubeletConfiguration
+evictionHard:
+  memory.available: "1"
```
BenTheElder: cc @neolit123 @dashpole for some reason cluster up fails when this is `memory.available: "0"`. I can't find any reason that should be invalid yet though.

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -90,6 +90,13 @@ apiServerExtraVolumes:
 # so we need to ensure the cert is valid for localhost so we can talk
 # to the cluster after rewriting the kubeconfig to point to localhost
 apiServerCertSANs: [localhost]
+kubeletConfiguration:
+  baseConfig:
```
BenTheElder: I _think_ this is right for the older types... https://github.com/kubernetes/kubernetes/blob/e6f6fa1f2dd136552e0d6e507b943e5adeaeae09/cmd/kubeadm/app/apis/kubeadm/v1alpha2/types.go#L286 @neolit123

neolit123: yes

BenTheElder:
I also confirmed that a default cluster from this has configz like:
```
kind create cluster --name=test
export KUBECONFIG=$(kind get kubeconfig-path --name=test)
kubectl proxy --port=8001 &
curl -sSL "http://localhost:8001/api/v1/nodes/test-control-plane/proxy/configz"
```
-> (pretty printed after the fact
<details>

```
{
  "kubeletconfig": {
    "staticPodPath": "/etc/kubernetes/manifests",
    "syncFrequency": "1m0s",
    "fileCheckFrequency": "20s",
    "httpCheckFrequency": "20s",
    "address": "0.0.0.0",
    "port": 10250,
    "tlsCertFile": "/var/lib/kubelet/pki/kubelet.crt",
    "tlsPrivateKeyFile": "/var/lib/kubelet/pki/kubelet.key",
    "rotateCertificates": true,
    "authentication": {
      "x509": {
        "clientCAFile": "/etc/kubernetes/pki/ca.crt"
      },
      "webhook": {
        "enabled": true,
        "cacheTTL": "2m0s"
      },
      "anonymous": {
        "enabled": false
      }
    },
    "authorization": {
      "mode": "Webhook",
      "webhook": {
        "cacheAuthorizedTTL": "5m0s",
        "cacheUnauthorizedTTL": "30s"
      }
    },
    "registryPullQPS": 5,
    "registryBurst": 10,
    "eventRecordQPS": 5,
    "eventBurst": 10,
    "enableDebuggingHandlers": true,
    "healthzPort": 10248,
    "healthzBindAddress": "127.0.0.1",
    "oomScoreAdj": -999,
    "clusterDomain": "cluster.local",
    "clusterDNS": [
      "10.96.0.10"
    ],
    "streamingConnectionIdleTimeout": "4h0m0s",
    "nodeStatusUpdateFrequency": "10s",
    "nodeStatusReportFrequency": "1m0s",
    "nodeLeaseDurationSeconds": 40,
    "imageMinimumGCAge": "2m0s",
    "imageGCHighThresholdPercent": 85,
    "imageGCLowThresholdPercent": 80,
    "volumeStatsAggPeriod": "1m0s",
    "cgroupsPerQOS": true,
    "cgroupDriver": "cgroupfs",
    "cpuManagerPolicy": "none",
    "cpuManagerReconcilePeriod": "10s",
    "runtimeRequestTimeout": "2m0s",
    "hairpinMode": "promiscuous-bridge",
    "maxPods": 110,
    "podPidsLimit": -1,
    "resolvConf": "/etc/resolv.conf",
    "cpuCFSQuota": true,
    "cpuCFSQuotaPeriod": "100ms",
    "maxOpenFiles": 1000000,
    "contentType": "application/vnd.kubernetes.protobuf",
    "kubeAPIQPS": 5,
    "kubeAPIBurst": 10,
    "serializeImagePulls": true,
    "evictionHard": {
      "imagefs.available": "0%",
      "memory.available": "1",
      "nodefs.available": "0%",
      "nodefs.inodesFree": "0%"
    },
    "evictionPressureTransitionPeriod": "5m0s",
    "enableControllerAttachDetach": true,
    "makeIPTablesUtilChains": true,
    "iptablesMasqueradeBit": 14,
    "iptablesDropBit": 15,
    "failSwapOn": false,
    "containerLogMaxSize": "10Mi",
    "containerLogMaxFiles": 5,
    "configMapAndSecretChangeDetectionStrategy": "Watch",
    "enforceNodeAllocatable": [
      "pods"
    ]
  }
}
```
</details>

pkg/cluster/internal/kubeadm/config.go
neolit123: i cannot find where in the source 0 would be a special case.
also this seems undocumented.

BenTheElder:
If this solves Mitar's problem the next thing to consider is smaller
thresholds rather than none.

We do not have conformance depending on this. I don't think you could make
a reasonable test for that very easily. Eviction behavior is supposed to be
tuneable.

mitar:
It does solve a problem for me and now I can run kind on my laptop.

BenTheElder:
> It does solve a problem for me and now I can run kind on my laptop.

Awesome! Now the question is which resource and should we adjust this back a bit... [the defaults](https://godoc.org/k8s.io/kubelet/config/v1beta1) are:
```yaml
evictionHard:
  memory.available:  "100Mi"
  nodefs.available:  "10%"
  nodefs.inodesFree: "5%"
  imagefs.available: "15%"
```

mitar:
See my comment in #55.

BenTheElder:
So specifically we probably just need the `imagefs.available` threshold and the rest are probably fine. Can you try:
```yaml
# config.yaml
kind: Config
- role: control-plane
  kubeadmConfigPatches:
  - |
    apiVersion: kubelet.config.k8s.io/v1beta1
    kind: KubeletConfiguration
    evictionHard:
      memory.available:  "100Mi"
      nodefs.available:  "10%"
      nodefs.inodesFree: "5%"
      imagefs.available: "0%"
```
`kind create cluster --config=config.yaml`

mitar:
I think this example config is invalid YAML.

BenTheElder:
fixed, it was missing a `nodes:` line before `- role: control-plane`

pkg/cluster/internal/kubeadm/config.go
BenTheElder: we're now not setting this key, but this is still... curious

BenTheElder:
Thanks to @mitar's testing [here](https://github.com/kubernetes-sigs/kind/issues/55#issuecomment-462622605) we're now just setting disk related thresholds, memory can be left alone for now.

neolit123:
hm, so between this in the other thread -  what are the technical details on why the kubelet is failing and what are the conditions it fails under? i saw the user's host node has a lot GB of images.
/lgtm
/hold

mitar:
So the problem is that having percentage as a limit might not be really reasonable for end-user laptop. For example, I have 2 TB drive, so 10% is 200 GB. So if I have 190 GB of free disk space (which I would claim is enough for my intended work with kind) Kubernetes would see it under the limit.

I would say that I agree with setting the limits to 0. It is not like you are loosing any data when running kind. So I do not think we should worry about available disk space. If it can run, it can. Otherwise it fails. And this is it.

Moreover I can imagine that this can also help on CI if things there become close to full as well. One does not expect that free space on your CI worker influences will your CI run fail or succeed (unless it is out of disk space).

BenTheElder:
> yet it seems to me we were passing the default thresholds.

Those thresholds are intended for use on dedicated VMs with reserved resources, kind runs on hosts that have lots of extra noise / resource usage on them.

Basically @mitar's comment above :-) https://github.com/kubernetes-sigs/kind/pull/293#issuecomment-462843546

BenTheElder:
This won't automerge due to the dual netlifys we somehow have now. I think I know how this happened (one we don't control), and will follow up... Manual merge in the meantime.

PR #299, opened 2019-02-14
convert dependency management to go modules

BenTheElder:
- we're still using vendor/ for now, all of the scripts respect this except `hack/{update,verify}-deps.sh`
- we no longer need to vendor dep
- go 1.11+ is required to manage dependencies (but was already required for proper gofmt)

TODO:
- module proof how we interact with kubernetes (finding and building it)
- eliminate vendor and switch to a module cache proxy? :thinking: 
- update docs


BenTheElder:
cc @neolit123 
I was vendoring CRI to add mounts and not happy with how that was being pruned with dep so I kinda just went all in and bit the bullet on go modules.

neolit123:
> I was vendoring CRI to add mounts and not happy with how that was being pruned with dep so I kinda just went all in and bit the bullet on go modules.

+1
the sooner modules are adapted the better.
handling k/k will be interesting. if foresee hacks.

BenTheElder:
/retest
note: working with `go.mod` requires go >= 1.11.4 (to avoid checksum mismatches, there are bugs in earlier go 1.11 versions)

Issue #148, opened 2018-12-04
Cluster doesn't restart when docker restarts


vincepri:
When docker restarts or stop/start (for any reason), the kind node containers remain stopped and aren't restarted properly. When I tried to run `docker restart <node container id>` the cluster didn't start either.

The only solution seems to recreate the cluster at this point.

/kind bug

BenTheElder:
Isn't this standard behavior with docker containers? (remaining stopped?)

kind does not run any daemon to mange the cluster, the commands create / delete "nodes" (containers) and run some tasks in them (like `kubeadm init`), they're effectively "unmanaged".

`docker restart` is not going to work, because creating a container is not just `docker run`, we need to take a few actions after creating the container.

--- 

What is the use case for this? These are meant to be transient test-clusters and it's probably not a good idea to restart the host daemon during testing.

"Restarting" a cluster is probably going to just look like delete + create of the cluster.

I'm not sure I'd even consider supporting this so much of a bug as a feature, "node" restarts are not really intended functionality currently.

neolit123:
> What is the use case for this?

+1 to this question.

docker restart in this case will act like a power grid restart on a bunch of bare metal machines.
so while those bare metal machines might come back up, not sure if we want to support this for kind.
for that to work i think some sort of state has to be stored somewhere...

vincepri:
I've been using kind locally (using Docker for Mac) and when docker reboots or stops, the cluster has to be deleted and recreated. I'm perfectly fine with it, just thought this might be something we should look into. 

The use case was to keep the cluster around even after I reboot or shut down my machine / docker.

BenTheElder:
Thanks for clarifying - this is certainly a hole in the usability but I'd hoped that clusters would be cheap enough to [create, use, delete] regularly.

This might be a little non-trivial to resolve but is probably do-able.
/priority backlog
/help

BenTheElder:
I think I know how we can do this effectively, but I have no idea what to call the command that will fit with the rest of the CLI 🙃

cc @munnerz

Something like `kind restart cluster` maybe?

vincepri:
`restart` seems it fits well with the other create/delete cluster commands, what's the idea you had? Wondering if it actually fits the `restart` word or it's something more.

BenTheElder:
It should roughly be:
- list the containers matching the cluster name
- for each ...
  - docker {re}start
  - run the pre-boot fixes (mounts)
  - signal the entrypoint to boot
[...]

PR #300, opened 2019-02-14
systemd start failure when nodes container restart

aoxn:
## Introduction
systemd start failure when nodes container restart
use host cgroup for systemd initializing to avoid this failure

## Problems
docker create a default cgroup mount for every node container instead of using host cgroup. systemd would start failure when nodes container restart (after SIGUSR1 was triggered) which due to some systemd cgroup behavior problems. Mount host cgroup to nodes container would resolve this problem, see `https://www.freedesktop.org/wiki/Software/systemd/ContainerInterface/`. 

## Reproduce
1. kind create cluster kind-xx
2. stop node containers by docker stop kind-xx
3. start node containers by docker start kind-xx
4. send SIGUSR1 signal by docker kill -s SIGUSR1 kind-xx
5. exec into kind-xx node containers with ps -ef command. 
`
root@kind-55-control-plane:/sys/fs/cgroup/unified# ps -eaf
UID        PID  PPID  C STIME TTY          TIME CMD
root         1     0  0 18:23 ?        00:00:00 /sbin/init
root        13     0  0 18:23 pts/0    00:00:00 bash
root        56    13  0 20:00 pts/0    00:00:00 ps -eaf
`
you would find systemd started without unit / service up.

Signed-off-by: Aoxn <yaoyao.aoxn@hotmail.com>

### DIFF
--- a/pkg/cluster/nodes/create.go
+++ b/pkg/cluster/nodes/create.go
@@ -121,6 +121,9 @@ func createNode(name, image, clusterLabel string, role config.NodeRole, extraArg
 		"--tmpfs", "/run", // systemd wants a writable /run
 		// some k8s things want /lib/modules
 		"-v", "/lib/modules:/lib/modules:ro",
+		// NOTE: systemd start failure when nodes container restart
+		// use host cgroup for systemd initializing to avoid this failure.
+		"-v", "/sys/fs/cgroup:/sys/fs/cgroup:ro",
 		"--hostname", name, // make hostname match container name
 		"--name", name, // ... and set the container name
 		// label the node with the cluster ID


BenTheElder:
Thanks for the PR but:
- restarting the containers is not currently supported and needs more than this, see https://github.com/kubernetes-sigs/kind/issues/148 (we would like to solve this though!)
- this won't work in CI currently
- making assumptions about the host cgroup path may be a bad idea
- the host may not use systemd

Note we are aware of https://www.freedesktop.org/wiki/Software/systemd/ContainerInterface (see the base image contents), but:
- that does not say you must mount the host cgroups, you either can or can let the container do it
- we're also running a CRI inside the container, and in CI we nest this again

/hold

neolit123:
on a related note we are adding strong assumptions in the CRI docs, that systemd is used as the init system:
https://github.com/kubernetes/website/pull/12638/files

aoxn:
@BenTheElder  Thanks, Looking forward to this new feature!

PR #304, opened 2019-02-15
fix logging in e2e.sh

BenTheElder:
- set xtrace to enable debugging
- quote export so bash doesn't get confused

### DIFF
--- a/hack/ci/e2e.sh
+++ b/hack/ci/e2e.sh
@@ -21,6 +21,7 @@
 set -o errexit
 set -o nounset
 set -o pipefail
+set -o xtrace
 
 # our exit handler (trap)
 cleanup() {
--- a/hack/ci/e2e.sh
+++ b/hack/ci/e2e.sh
@@ -26,7 +26,7 @@ set -o xtrace
 # our exit handler (trap)
 cleanup() {
     # always attempt to dump logs
-    kind export logs "${ARTIFACTS}/logs" || true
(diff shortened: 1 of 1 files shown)


hack/ci/e2e.sh
```diff
@@ -21,11 +21,12 @@
 set -o errexit
 set -o nounset
 set -o pipefail
+set -o xtrace
 
 # our exit handler (trap)
 cleanup() {
     # always attempt to dump logs
-    kind export logs "${ARTIFACTS}/logs" || true
+    kind "export" logs "${ARTIFACTS}/logs" || true
```
neolit123: yikes, why is it doing that...

BenTheElder: because "export" is a keyword in bash. possibly we could rename it... :thinking:

neolit123: but this works fine for me:
```
#!/usr/bin/env bash

set -o errexit
set -o nounset
set -o pipefail
set -o xtrace

cleanup() {
  kind export logs
}
cleanup
```

BenTheElder: hmm you're right, this fix didn't work, it must have just been a coincidence.

.. not sure if I should spend more time on this or kubetest2

BenTheElder: http://gcsweb.k8s.io/gcs/kubernetes-jenkins/logs/ci-kubernetes-kind-conformance-parallel/9528/artifacts/ <-- no logs

PR #308, opened 2019-02-15
Implements a new command "load" which allows images to be loaded 

### LINKED ISSUE #28: First class image side-loading support

stevesloka:
Fixes #28 by implementing a new `load` command which allows docker images to be loaded via an archive or by referencing an image on the host.

Example:
```
# Loads an image from an archive to cluster named `1`
$ kind load image-archive --name=1 sonoimg.tar 

# Loads an image from one that exists on the host into cluster named `1`
$ kind load docker-image --name=1 gcr.io/heptio-images/sonobuoy:kindCI 
```

It's also possible to pass a `--nodes` flag to specificy which nodes to deploy the image to, if not specified, it defaults to all nodes. 

Signed-off-by: Steve Sloka <slokas@vmware.com>

### DIFF
--- a/cmd/kind/load/docker-image/docker-image.go
+++ b/cmd/kind/load/docker-image/docker-image.go
@@ -1,0 +1,102 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package load implements the `load` command
+package load
+
+import (
+	"fmt"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+
+	"sigs.k8s.io/kind/pkg/cluster"
+	clusternodes "sigs.k8s.io/kind/pkg/cluster/nodes"
+	"sigs.k8s.io/kind/pkg/docker"
+)
+
+type flagpole struct {
+	Name  string
+	Nodes []string
+}
+
+// NewCommand returns a new cobra.Command for loading an image into a cluster
+func NewCommand() *cobra.Command {
+	flags := &flagpole{}
+	cmd := &cobra.Command{
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) < 1 {
+				return errors.New("name of image is required")
+			}
+			return nil
+		},
+		Use:   "docker-image",
+		Short: "loads docker image from host into nodes",
+		Long:  "loads docker image from host into all or specified nodes by name",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runE(flags, cmd, args)
+		},
+	}
+	cmd.Flags().StringVar(
+		&flags.Name,
+		"name",
+		cluster.DefaultName,
+		"the cluster context name",
+	)
+	cmd.Flags().StringSliceVar(
+		&flags.Nodes,
+		"nodes",
+		nil,
+		"comma seperated list of nodes to load images into",
+	)
+	return cmd
+}
+
+func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
+	imageTarName := "image.tar"
+	destinationImageTar := fmt.Sprintf("/%s", imageTarName)
+
+	// Get the image into a tar
+	err := docker.Save(args[0], imageTarName)
+	if err != nil {
+		return err
+	}
+
+	// List nodes by cluster context name
+	n, err := clusternodes.ListByCluster()
+	if err != nil {
+		return err
+	}
+	nodes, known := n[flags.Name]
+	if !known {
+		return errors.Errorf("unknown cluster %q", flags.Name)
+	}
+
+	for _, node := range nodes {
+		// Copy image tar to each node
+		if err := node.CopyTo(imageTarName, destinationImageTar); err != nil {
+			return errors.Wrap(err, "failed to copy image to node")
+		}
+
+		// Load image into each node
+		if err := node.Command(
+			"docker", "load", "--input", destinationImageTar,
+		).Run(); err != nil {
+			return errors.Wrap(err, "failed to load image")
+		}
+	}
+	return nil
+}
--- a/cmd/kind/load/image-archive/image-archive.go
+++ b/cmd/kind/load/image-archive/image-archive.go
@@ -1,0 +1,94 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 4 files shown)


BenTheElder:
thank you!!
/assign
/ok-to-test

cmd/kind/load/docker-image/docker-image.go
```diff
@@ -0,0 +1,102 @@
...
+		&flags.Name,
+		"name",
+		cluster.DefaultName,
+		"the cluster context name",
+	)
+	cmd.Flags().StringSliceVar(
+		&flags.Nodes,
+		"nodes",
+		nil,
+		"comma seperated list of nodes to load images into",
```
BenTheElder: should we mention "(default is all nodes)" or something along those lines?

cmd/kind/load/docker-image/docker-image.go
```diff
@@ -0,0 +1,102 @@
...
+	)
+	return cmd
+}
+
+func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
+	imageTarName := "image.tar"
```
BenTheElder: possibly this should be a tempfile that we then defer removing?

cmd/kind/load/image-archive/image-archive.go
```diff
@@ -0,0 +1,94 @@
...
+	// List nodes by cluster context name
+	n, err := clusternodes.ListByCluster()
+	if err != nil {
+		return err
+	}
+	nodes, known := n[flags.Name]
+	if !known {
+		return errors.Errorf("unknown cluster %q", flags.Name)
+	}
+
```
BenTheElder: we should _not_ try to solve all of this in the first PR, but some TODOs for later:
```go
// TODO: handle non-dockershim CRI inside the node
// TODO: test performance of concurrently loading to all nodes
// TODO: pipe the image directly into docker load without an intermediate copy
```

cmd/kind/load/docker-image/docker-image.go
```diff
@@ -0,0 +1,102 @@
...
+	// List nodes by cluster context name
+	n, err := clusternodes.ListByCluster()
+	if err != nil {
+		return err
+	}
+	nodes, known := n[flags.Name]
+	if !known {
+		return errors.Errorf("unknown cluster %q", flags.Name)
+	}
+
```
BenTheElder: (dupe of https://github.com/kubernetes-sigs/kind/pull/308#discussion_r257400168) - we should explore optimizations in the future, we should not prematurely optimize though :sweat_smile:

cmd/kind/load/docker-image/docker-image.go
stevesloka: Is there a good spot to work out of? I tried writing to `/tmp` but the `docker cp` wouldn't work. It wouldn't throw an error, just wouldn't copy.

cmd/kind/load/docker-image/docker-image.go
stevesloka: Yup I can remove, I just literally copy/pasted from another spot. =)

cmd/kind/load/docker-image/docker-image.go
stevesloka: 👍

cmd/kind/load/docker-image/docker-image.go
BenTheElder: Was this on Mac? The tempdir wrapper in pkg/fs works around this, we may need a tempfiles wrapper too

BenTheElder: https://github.com/kubernetes-sigs/kind/blob/e14d2d27103b4d1b4a2e4ba81d08a672cf64dec1/pkg/fs/fs.go#L30-L43

stevesloka: Just to clarify which I think I understand now is you are looking for a temp file on the `host`. I was trying to copy into a tmp folder inside the container. 

I agree the host should be done, should we also have a temp place inside the container to copy to? (e.g. `docker cp /tmp/img.tar kind-1-worker3:/tmp/img.tar`)

When I tried to copy to `/tmp` in the container, that's where I was getting the silent failure.

BenTheElder: 🤦‍♂️

BenTheElder: This comment should be on the imageTarName below, not this destinationImageTar

cmd/kind/load/docker-image/docker-image.go
```diff
@@ -0,0 +1,102 @@
...
+	)
+	return cmd
+}
+
+func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
+	imageTarName := "image.tar"
+	destinationImageTar := fmt.Sprintf("/%s", imageTarName)
+
+	// Get the image into a tar
+	err := docker.Save(args[0], imageTarName)
```
BenTheElder: So _this_ line is saving to image.tar in the current directory.
We should use a tempdir instead and defer deleting the did.

BenTheElder: Additionally we should probably add a random suffix to the destination image name to make it save to run multiple loads concurrently.

cmd/kind/load/docker-image/docker-image.go
BenTheElder: Within the container this is fine for now. We should similarly clean it up after the load though. (Or see about piping it to the load and never writing the tar to disk directly within the container)

BenTheElder:
Just checked this out locally, here's a small patch that avoids writing the image to the container filesystem (only implemented it for one of the two commands, but the same idea should apply)
```patch
diff --git a/cmd/kind/load/docker-image/docker-image.go b/cmd/kind/load/docker-image/docker-image.go
index 3529747..2401c94 100644
--- a/cmd/kind/load/docker-image/docker-image.go
+++ b/cmd/kind/load/docker-image/docker-image.go
@@ -18,7 +18,7 @@ limitations under the License.
 package load
 
 import (
-	"fmt"
+	"os"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
@@ -67,7 +67,6 @@ func NewCommand() *cobra.Command {
 
 	imageTarName := "image.tar"
-	destinationImageTar := fmt.Sprintf("/%s", imageTarName)
 
 	// Get the image into a tar
 	err := docker.Save(args[0], imageTarName)
@@ -86,17 +85,28 @@ func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
 	}
 
 	for _, node := range nodes {
-		// Copy image tar to each node
-		if err := node.CopyTo(imageTarName, destinationImageTar); err != nil {
-			return errors.Wrap(err, "failed to copy image to node")
+		if err := loadImage(imageTarName, &node); err != nil {
+			return err
 		}
+	}
+	return nil
+}
 
-		// Load image into each node
-		if err := node.Command(
-			"docker", "load", "--input", destinationImageTar,
-		).Run(); err != nil {
-			return errors.Wrap(err, "failed to load image")
-		}
+func loadImage(imageTarName string, node *clusternodes.Node) error {
+	// Copy image tar to each node
+	f, err := os.Open(imageTarName)
+	if err != nil {
+		return errors.Wrap(err, "failed to open image")
+	}
+	defer f.Close()
+
+	// Load image into each node
+	cmd := node.Command(
+		"docker", "load",
+	)
+	cmd.SetStdin(f)
+		return errors.Wrap(err, "failed to load image")
 	}
 	return nil
 }
```

BenTheElder:
actually, i'm about to reorganize the docker package a bit as well, let me go ahead and merge this to unblock people and I'll follow up with the patch above etc.

thanks again!
/lgtm
/approve
/hold cancel

PR #309, opened 2019-02-15
ensure all go get references have update flag

BenTheElder:
this one trips people up a lot, `go get` without `-u` will install from an old clone if one exists. in our install instructions when we say `go get` we want an up to date copy.

we should also add more instructions for installing from a release.


BenTheElder:
/hold cancel
> another related problem is the binary not landing into PATH for one reason or another.

yes, we will need updated instructions as go module mode becomes popular.
where the user guide mentions `go get` install we should also specify the minimum version known to work.

PR #310, opened 2019-02-16
Improve load subcommand

BenTheElder:
follow up to #308 
- use a tempdir that we clean up for the locally saved image
- pipe images to the node instead of copying them
- actually implement the `--nodes` flag
- add some limited documentation


BenTheElder:
TODO:
- move most of this logic into `pkg/...` for library usage, in particular image loading routines should be moved there so we can abstract away the node's underlying CRI and load command
- document more

Issue #154, opened 2018-12-06
support multiple container engines on the host


BenTheElder:
<strike>#151 should bring initial support for `podman` via @rhatdan</strike>

~This should mostly be a drop in replacement.~ (it's not... podman has quirks)

We should further clean up the code base a bit to help with this sort of thing. We've mostly but not quite isolated the container runtime functionality.

/assign
/assign @rhatdan 

/kind feature
/kind cleanup
/priority important-longterm

related: #153 is the same thing, but within the nodes

BenTheElder:
There's now a `container` package with `docker` as a subset, and we're working on plumbing through optional arguments to the cluster better, I expect to get limited podman support based on these pretty soon.

BenTheElder:
We're also targeting CRI functionality explicitly: https://kind.sigs.k8s.io/docs/design/principles/#target-cri-functionality

BenTheElder:
Currently working on moving more of the flags (ports) behind abstractions borrowed from CRI. 
I'd like to isolate all knowledge of docker / podman style flags to pkg/container/... by the next milestone, at which point we should be able to do podman support properly with it's own implementation of all functionality.

tinyzimmer:
This is pretty old, but in case anyone is still paying attention - I'm pretty close to having this working where basically the global `--podman` switch uses podman instead.

I had to do something along the lines of what you are describing first, @BenTheElder, where I made a generic `Container` interface and different backends (e.g. docker,podman) just need to implement the various methods.

```
package runtime

// .. import pkg/container/backends ...

type ContainerRuntime string

const Podman ContainerRuntime = "podman"
const Docker ContainerRuntime = "docker"

var currentRuntime ContainerRuntime

func SetContainerRuntime(rt ContainerRuntime) {
	currentRuntime = rt
}

func GetContainerRuntime() container.Container {
	switch currentRuntime {
	case Docker:
		return docker.New()
	case Podman:
		return podman.New()
	default:
		panic("invalid container runtime")
	}
}
```

Then replaced all the places that called the `docker` package directly with basically a `runtime.GetContainerRuntime()`

BenTheElder:
Yeah, sorry this has been drawn out. We're definitely doing this, things have settled a bit and now that I've landed some new debugging infrastructure and other major cleanups I'll be reviving this ...

That said: If you're just replacing `docker` with `podman` in all the commands that should be possible today without any changes to the binary:

```
TMP_DIR=$(mktemp -d)
[...]

PR #313, opened 2019-02-17
refactor container packages and support mounts

BenTheElder:
- move pkg/docker to pkg/container/docker, long term we can better break up docker specific vs general container functionality and support other container runtimes.
  - tracking issue: https://github.com/kubernetes-sigs/kind/issues/154
  - see also: https://github.com/kubernetes-sigs/kind/pull/151, this is work towards a cleaner alternative
  - see also: https://kind.sigs.k8s.io/docs/design/principles/#target-cri-functionality
- create a CRI package, with types (Mount related only for now) copied from upstream
  - this allows us to avoid stability concerns, if necessary we can provide our own conversion
  - I added better JSON / yaml encoding for Mount's propagation field
    - I've tested that in the future we could instead use `type Mount upstreamcri.Mount` and still support this
- add support for CRI Mounts in `docker.Run`, refactor docker.Run to take functional options
- support mounts in the node config, update control-plane and worker node creation to support this

This is an alternative to #302 where we vendor CRI

cc @neolit123


pkg/container/docker/run.go
```diff
@@ -22,19 +22,59 @@ import (
...
 	log "github.com/sirupsen/logrus"
 
+	"sigs.k8s.io/kind/pkg/container/cri"
 	"sigs.k8s.io/kind/pkg/exec"
 )
 
 // Docker container IDs are hex, more than one character, and on their own line
 var containerIDRegex = regexp.MustCompile("^[a-f0-9]+$")
 
+type RunOpt func(*runOpts) *runOpts
```
BenTheElder: TODO: need to godoc these. will fix before merge

BenTheElder: fixed

BenTheElder:
usage:
```yaml
kind: Config
nodes:
- role: control-plane
  mounts:
  - container_path: /kind-source
    host_path: /Users/bentheelder/go/src/sigs.k8s.io/kind
    readonly: true
```

Things to consider for the future (but I think I'd rather not do in the first PR to keep the diff down):
- swap out the snakeCase json fields for camelCase fields to match the rest of our config?
- resolve environment variables in `host_path` or at least resolve relative paths with respect to the `kind` working directory. Docker requires absolute paths for bind mounts but that's not super great UX for our case I think ...

pkg/cluster/config/types.go
```diff
@@ -54,6 +56,9 @@ type Node struct {
 	// KubeadmConfigPatchesJSON6902 are applied to the generated kubeadm config
 	// as patchesJson6902 to `kustomize build`
 	KubeadmConfigPatchesJSON6902 []kustomize.PatchJSON6902
+	// Mounts describes additional mount points for the node container
+	// These may be used to EG bind a hostpath
+	Mounts []cri.Mount `json:"mounts,omitempty"`
```
neolit123: there is a weird side effect when marshaling slices with omitempty, that is already a problem in k/k, kubeadm, cluster-api. i think the safe route was to use `Mounts *[]cri.Mount `json:"mounts,omitempty"` but i need to check this.

pkg/container/cri/json.go
```diff
@@ -0,0 +1,63 @@
...
+
+func (m *Mount) UnmarshalJSON(data []byte) error {
+	type Alias Mount
+	aux := &struct {
+		Propagation string `json:"propagation"`
+		*Alias
+	}{
+		Alias: (*Alias)(m),
+	}
+	if err := json.Unmarshal(data, &aux); err != nil {
```
neolit123: ideally everything "unmarshall" should be using [DisallowUnknownFields](https://gist.github.com/bndw/a95e49c8bb5581aa4685d289ec9a3aeb)
but i leave this to you.

pkg/container/cri/types.go
```diff
@@ -0,0 +1,73 @@
...
+	// Path of the mount on the host. If the hostPath doesn't exist, then runtimes
+	// should report error. If the hostpath is a symbolic link, runtimes should
+	// follow the symlink and mount the real destination to container.
+	HostPath string `protobuf:"bytes,2,opt,name=host_path,json=hostPath,proto3" json:"host_path,omitempty"`
+	// If set, the mount is read-only.
+	Readonly bool `protobuf:"varint,3,opt,name=readonly,proto3" json:"readonly,omitempty"`
+	// If set, the mount needs SELinux relabeling.
+	SelinuxRelabel bool `protobuf:"varint,4,opt,name=selinux_relabel,json=selinuxRelabel,proto3" json:"selinux_relabel,omitempty"`
+	// Requested propagation mode.
+	Propagation MountPropagation `protobuf:"varint,5,opt,name=propagation,proto3,enum=runtime.v1alpha2.MountPropagation" json:"propagation,omitempty"`
```
neolit123: looks like these variable names need "adjustments" upstream.

pkg/container/cri/types.go
```diff
@@ -0,0 +1,73 @@
...
+type MountPropagation int32
+
+const (
+	// No mount propagation ("private" in Linux terminology).
+	MountPropagation_PROPAGATION_PRIVATE MountPropagation = 0
+	// Mounts get propagated from the host to the container ("rslave" in Linux).
+	MountPropagation_PROPAGATION_HOST_TO_CONTAINER MountPropagation = 1
+	// Mounts get propagated from the host to the container and from the
+	// container to the host ("rshared" in Linux).
+	MountPropagation_PROPAGATION_BIDIRECTIONAL MountPropagation = 2
```
neolit123: i've missed if we are adding a linter exception for this file/package.

BenTheElder: We did, it's a separate commit.

pkg/container/cri/types.go
BenTheElder: Yeah I'm not sure if that's actually an option. I think it would probably be fine for us to match the types closely but use a different on-disk format.

pkg/cluster/config/types.go
neolit123:
```
package main

import (
	"fmt"
	"log"
	"sigs.k8s.io/yaml"
)

type A struct {
	Mounts []string `json:"mounts,omitempty"`
}

func main() {
	a := A{Mounts: []string{}} // <--------- [*]
	fmt.Println(a.Mounts, a.Mounts == nil)
	b, err := yaml.Marshal(a)	
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println(string(b))
	c := A{}
	err = yaml.Unmarshal(b, &c)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println(c.Mounts, c.Mounts == nil)
}
```

```
[] false
{}
[] true
```

[*]
so this problem is relevant only if we assign a meaning to the "empty" value of `Mounts` and what is the "empty" value (is it a empty slice or `nil`) and if the user is marhaling the kind Config.

if there is no default value this problem is not relevant, but just referencing this here so that we don't have this problem for the future kind config:
https://github.com/kubernetes/kubeadm/issues/1358

pkg/container/cri/types.go
neolit123: ah, saw it.

pkg/cluster/config/types.go
BenTheElder: What was it? https://play.golang.org/p/EOfDnWThpdj

pkg/container/cri/types.go
neolit123: we still have the option to shim with getters and setters. this means some extra boiler plate.
this fixes both the field names and hides the upstream config.

> create a CRI package, with types (Mount related only for now) copied from upstream
* this allows us to avoid stability concerns, if necessary we can provide our own conversion

i need to know the roadmap for CRI in terms of do they have plans to remove or rename fields, because if there are upstream API changes our local type can end up invalid and we will have to break kind users because these fields are public in the kind config.

i will leave this to you at this point if you want us to proceed, but comments from others would be appreciated too.

pkg/cluster/config/types.go
BenTheElder: I _think_ we can treat empty and nil the same, the node already has mounts (tmps and one docker volume) we're just adding additional mounts here (we should clarify that in the docs)

pkg/container/cri/types.go
BenTheElder: CRI is pretty stable at this point, this type hasn't actually changed in years.

But our local type can be a shim just closely matching their type, if they diverge we will translate. As long as we are not using the client this isn't an issue.

But also for the types we expose here for on disk purposes, what they are capable of is unlikely to change. They cannot remove support for these fields without breaking Kubernetes majorly.

pkg/cluster/config/types.go
neolit123: hm, yes. i know the existing mounts are essential for the setup to work.

so in a way we already have default mounts and these mounts are extra mounts :thinking: 
would it be clearer naming this `ExtraMounts` and including some more nodes in the field comment?

pkg/container/cri/types.go
neolit123: ok

BenTheElder: It is nice for now that they are literally the same definition, but if it diverges slightly (EG they add a field or break the type up) we can keep our changes compatible and translate.

The things we will need to expose in config like Mounts, DNS, ... those seem unlikely to change significantly.

More complex types like entire containers, those seem stable too but we don't need to expose those on disk anyhow.

neolit123:
/assign @munnerz 
hi, do you have comments on the new CRI mount support?

pkg/container/cri/json.go
BenTheElder: Yeah, trying to figure out if we should just do that inline or if there's a way to plumb through the option 🙄

pkg/cluster/config/types.go
BenTheElder: Yeah I think that might make sense.

BenTheElder: renamed it 👍

BenTheElder:
per discussion in the meeting I renamed to camel case and switched to the core/v1 VolumeMount strings for the mount enum. I think we can remove the lint exception now as well.

BenTheElder:
Re-enabled golint and fixed remaining lints.

pkg/container/cri/json.go
BenTheElder: I'm going to tackle this in a follow up, but before the next release, I'm not sure how this will interact with the rest of the decoding logic.

i'm still hopeful for getting support for that upstream in API machinery.

pkg/container/cri/types.go
BenTheElder: I adjusted our names, strangely the proto json tags were already camelCase but not the actual json tags 🤷‍♂️ 

Where possible these match user facing API fields in Kubernetes APIs

pkg/container/cri/types.go
```diff
@@ -0,0 +1,83 @@
...
+	// Path of the mount within the container.
+	ContainerPath string `protobuf:"bytes,1,opt,name=container_path,json=containerPath,proto3" json:"containerPath,omitempty"`
+	// Path of the mount on the host. If the hostPath doesn't exist, then runtimes
+	// should report error. If the hostpath is a symbolic link, runtimes should
+	// follow the symlink and mount the real destination to container.
+	HostPath string `protobuf:"bytes,2,opt,name=host_path,json=hostPath,proto3" json:"hostPath,omitempty"`
+	// If set, the mount is read-only.
+	Readonly bool `protobuf:"varint,3,opt,name=readonly,proto3,json=readOnly,proto3" json:"readOnly,omitempty"`
+	// If set, the mount needs SELinux relabeling.
+	SelinuxRelabel bool `protobuf:"varint,4,opt,name=selinux_relabel,json=selinuxRelabel,proto3" json:"selinuxRelabel,omitempty"`
```
BenTheElder: this one is not actually exposed directly in kubernetes volumes AFAICT, but from previous test-infra containerization work I remember control over this being necessary on some systems.

we should consider if we want to expose this or not, imho we can just document it well and expect the user to do the right thing, support for this shouldn't go anywhere in CRI

BenTheElder:
/hold cancel

Will follow up with considerations on EG relative paths / other UX improvements.

In the meantime this config is now possible:
```yaml
kind: Config
nodes:
- role: control-plane
  extraMounts:
  - containerPath: /kind-source
    hostPath: /Users/bentheelder/go/src/sigs.k8s.io/kind
    readOnly: true
```

PR #314, opened 2019-02-18
improve contributor docs

BenTheElder:
- ungroup tool installs, most everyone should install git... generally improve the getting started guide
- add many links and a few more details to the scope document

### DIFF
--- a/site/content/docs/contributing/getting-started.md
+++ b/site/content/docs/contributing/getting-started.md
@@ -31,8 +31,6 @@ the following command:
 ```
 git --version
 ```
-This documentation is written using `git` version `2.17.1`. 
-Your version may be different depending on your OS.
 
 ### Install Hugo
 
--- a/site/content/docs/contributing/getting-started.md
+++ b/site/content/docs/contributing/getting-started.md
@@ -20,44 +20,53 @@ an account to contribute.
 
 ## 2. Install Tools
 
-### Install Documentation Tools
+### Install git
 
-If you wish to contribute to the documentation, it is recommended but not 
-required to install [hugo], which we use to develop this site.
+Our source code is managed with [`git`][git], to develop locally you
+will need to install `git`.
 
-Please see: https://gohugo.io/getting-started/installing/
-
-### Install Developer Tools
-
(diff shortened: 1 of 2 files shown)


BenTheElder:
these docs both still need work. this should help a little.

tao12345666333:
I also happened to think of this part, I pushed an [image](https://hub.docker.com/r/taobeier/kindenv/tags) last night. 

The source code https://github.com/tao12345666333/kindenv .

Do we need to provide a standard development environment, like the moby project's development environment?

BenTheElder:
> Do we need to provide a standard development environment, like the moby project's development environment?

I don't think that's necessary currently. We just require a reasonably up to date Go install and the same docker install requirements as **using** kind. 
Git probably isn't worth containerizing just for kind, and Hugo is fairly optional since we have the previews and the contents are just markdown.

tao12345666333:
I agree with you. :+1: 

The original intention of this [project](https://github.com/tao12345666333/kindenv) was to add a timing build task for CI. Auto build node image synchronized with the upstream version of kubernetes

BenTheElder:
Ooh, related to https://github.com/kubernetes-sigs/kind/issues ? 👀

tao12345666333:
xref: https://github.com/kubernetes-sigs/kind/issues/197

site/content/docs/contributing/getting-started.md
```diff
@@ -20,44 +20,53 @@ an account to contribute.
...
-#### Install git
-Install `git` on your local machine.
 You can check if `git` is already on your system and properly installed with 
 the following command:
 
 ```
 git --version
 ```
-This documentation is written using `git` version 2.17.1. 
+This documentation is written using `git` version `2.17.1`.
```
neolit123: written -> managed?

also possibly no need to add this sentence?
users without git will land with a fairly new and compatible git package as a download.

BenTheElder: SGTM, this was already in the doc, I only changed the command / version markup to match. Will remove.

BenTheElder: done

PR #316, opened 2019-02-18
library - get role

fabriziopandini:
This PR adds a method to the `Node` object for reading the `io.k8s.sigs.kind.role` label.

This method is really useful (or I would rather say it is necessary) when using kind as a library on an existing cluster, and you must use the list of actual nodes as a source of truth for your cluster topology e.g.
```go
nodes, _ := c.ListNodes()

for _, n := range nodes {
    if n.Role() == config.ControlPlaneRole {
        // Do something on roles
    }
}
```

### DIFF
--- a/pkg/cluster/nodes/node.go
+++ b/pkg/cluster/nodes/node.go
@@ -24,12 +24,15 @@ import (
 	"path/filepath"
 	"regexp"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 
 	"k8s.io/apimachinery/pkg/util/version"
+	"sigs.k8s.io/kind/pkg/cluster/config"
+	"sigs.k8s.io/kind/pkg/cluster/constants"
 
 	"sigs.k8s.io/kind/pkg/docker"
 	"sigs.k8s.io/kind/pkg/exec"
@@ -69,6 +72,7 @@ type nodeCache struct {
 	kubernetesVersion string
 	ip                string
 	ports             map[int]int
+	role              config.NodeRole
 	containerCmder    exec.Cmder
 }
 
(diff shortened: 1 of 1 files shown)


pkg/cluster/nodes/node.go
```diff
@@ -259,6 +263,24 @@ func (n *Node) Ports(containerPort int) (hostPort int, err error) {
...
+	if n.nodeCache.role != "" {
+		return n.nodeCache.role, nil
+	}
+	// retrive the role the node using docker inspect
+	lines, err := docker.Inspect(n.nameOrID, fmt.Sprintf("{{index .Config.Labels %q}}", constants.ClusterRoleKey))
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to get %q label", constants.ClusterRoleKey)
+	}
+	if len(lines) != 1 {
+		return "", fmt.Errorf("%q label should only be one line, got %d lines", constants.ClusterRoleKey, len(lines))
```
neolit123: `errors.Errorf`.

fabriziopandini: done

pkg/cluster/nodes/node.go
```diff
@@ -259,6 +263,24 @@ func (n *Node) Ports(containerPort int) (hostPort int, err error) {
...
 }
 
+// Role returns the role of the node
+func (n *Node) Role() (role config.NodeRole, err error) {
+	// use the cached version first
+	if n.nodeCache.role != "" {
+		return n.nodeCache.role, nil
+	}
+	// retrive the role the node using docker inspect
+	lines, err := docker.Inspect(n.nameOrID, fmt.Sprintf("{{index .Config.Labels %q}}", constants.ClusterRoleKey))
```
BenTheElder: We should probably move logic to obtain labels into the docker package at some point. This is consistent for now though and this PR is nice and small 👍

fabriziopandini: :+1:

PR #317, opened 2019-02-19
Upgrade kustomize

BenTheElder:
supersedes https://github.com/kubernetes-sigs/kind/pull/138
- brings us up to an up to date version, lots of work has gone into kustomize
  - drops tons of dependencies
    - drops our binary size by ~5MB to ~31MB from ~36MB due largely to the AWS SDK being removed (by not depending on go-getter upstream)
- this required injecting a fixed name (`"config"`) into all generated config objects and the patches 

Technically this is a breaking change I think (?), since I don't think you can set other metadata fields on your patches, but that never made sense and we never documented doing this. Some of them never would have worked (`generateName` for example)


tao12345666333:
Quickly looked down. LGTM :+1:

PR #318, opened 2019-02-19
library - Refactor create options

fabriziopandini:
As discussed in last kind office hours (and in the kind slack channel), the current `library.Create` method expose options in a way that cannot be supported in the long term, because each new option will introduce a breaking change

This PR propose an alternative approach, based on a pattern already used in kind and described [here](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis)

### DIFF
--- a/pkg/cluster/context.go
+++ b/pkg/cluster/context.go
@@ -100,8 +100,26 @@ func (c *Context) ClusterName() string {
 	return c.Name()
 }
 
+//CreateOption is a Context configuration option supplied to create
+type CreateOption func(*create.Context)
+
+// Retain configures create to retain nodes after failing for debugging pourposes
+func Retain(retain bool) CreateOption {
+	return func(c *create.Context) {
+		c.Retain = retain
+	}
+}
+
+// WaitForReady configures create to use interval as maximum wait time for the control plane node to be ready
+func WaitForReady(interval time.Duration) CreateOption {
+	return func(c *create.Context) {
+		// NB. WaitForReady option does not apply directly to create, but it should be forwarded to the exec command
+		c.ExecOptions = append(c.ExecOptions, create.WaitForReady(interval))
+	}
+}
+
 // Create provisions and starts a kubernetes-in-docker cluster
-func (c *Context) Create(cfg *config.Config, retain bool, wait time.Duration) error {
+func (c *Context) Create(cfg *config.Config, options ...CreateOption) error {
 	// default config fields (important for usage as a library, where the config
 	// may be constructed in memory rather than from disk)
 	encoding.Scheme.Default(cfg)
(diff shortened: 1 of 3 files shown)


pkg/cluster/context.go
```diff
@@ -100,8 +100,26 @@ func (c *Context) ClusterName() string {
 	return c.Name()
 }
 
+//CreateOption is Context configuration option supplied to create
```
neolit123: `is a Context`

pkg/cluster/internal/create/createcontext.go
```diff
@@ -39,28 +39,42 @@ type Context struct {
 	Status *logutil.Status
 	Config *config.Config
 	*DerivedConfig
-	Retain       bool          // if we should retain nodes after failing to create.
-	WaitForReady time.Duration // Wait for the control plane node to be ready.
+	Retain      bool         // if we should retain nodes after failing to create.
+	ExecOptions []ExecOption // options to be forwarded to the exec command.
+}
+
+//ExecOption is execContext configuration option supplied to Exec
```
neolit123: `is an execContext`

neolit123:
@BenTheElder @fabriziopandini we should move wait and retain to the config only.
these are advanced use case flags.

BenTheElder:
+1 on moving those, but in be another PR.

pkg/cluster/context.go
```diff
@@ -100,8 +100,26 @@ func (c *Context) ClusterName() string {
 	return c.Name()
 }
 
+//CreateOption is a Context configuration option supplied to create
+type CreateOption func(*create.Context)
+
+// Retain configures create to retain nodes after failing for debugging pourposes
+func Retain(retain bool) CreateOption {
+	return func(c *create.Context) {
+		c.Retain = retain
```
BenTheElder: normally these would not be exported, so you can ensure that they are only set when applying the options. We can refactor further in a follow-up though

fabriziopandini:
@BenTheElder I don't think the error is related to this change...
Which version of kubeadm is testing "pull-kind-conformance-parallel"? Master?

BenTheElder:
Yes master, the failure is due to upstream. https://github.com/kubernetes-sigs/kind/issues/321

I put in a PR to test-infra to enable `/override` but I don't think it merged yet. The fix to kubeadm should be merged though
/retest

BenTheElder:
This test case needs to be fixed 🙄  https://gubernator.k8s.io/build/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_kind/318/pull-kind-conformance-parallel/488/#sig-network-service-endpoints-latency-should-not-be-very-high-conformance

/retest

BenTheElder:
I filed a bug upstream about this test

PR #319, opened 2019-02-20
print kubeadm init and join combined output for loglevel 'debug'

neolit123:
- Use --v=6 and print the combined output from 'init' and 'join'
if the user sets --loglevel=info.
- Skip printing the kubeadm joint token from init.

https://k8s-testgrid.appspot.com/sig-cluster-lifecycle-all#kubeadm-kind-master
https://k8s-testgrid.appspot.com/sig-testing-kind#conformance,%20master%20(dev)%20%5Bnon-serial%5D

^ seeing some errors.

i think this is quite needed because without it, the actual error from kubeadm init or join is not printed to the kind user host terminal. also with level=info i'm adding a pretty high klog debug level =6.
i'm ok with reducing the level 6 to something lower?

/kind feature
/important soon
/assign @BenTheElder 
cc @fabriziopandini

### DIFF
--- a/pkg/cluster/internal/create/const.go
+++ b/pkg/cluster/internal/create/const.go
@@ -29,3 +29,5 @@ metadata:
   labels:
     addonmanager.kubernetes.io/mode: EnsureExists
 provisioner: kubernetes.io/host-path`
+
+const kubeadmVerbosityFlag = "--v=6"
--- a/pkg/cluster/internal/create/kubeadm-init.go
+++ b/pkg/cluster/internal/create/kubeadm-init.go
@@ -17,6 +17,7 @@ limitations under the License.
 package create
 
 import (
+	"sigs.k8s.io/kind/pkg/exec"
 	"strings"
 	"time"
 
@@ -64,7 +65,7 @@ func runKubeadmInit(ec *execContext, configNode *NodeReplica) error {
 	}
 
 	// run kubeadm
-	if err := node.Command(
+	cmd := node.Command(
 		// init because this is the control plane node
 		"kubeadm", "init",
 		// preflight errors are expected, in particular for swap being enabled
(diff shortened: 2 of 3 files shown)


BenTheElder:
We still need a more consistent plan for log output. 

Ideally all exec commands should capture output and output them on error.

These outputs seem more like debug than info?

neolit123:
> We still need a more consistent plan for log output.

we need to catch errors like that this cycle (aka right now).

> Ideally all exec commands should capture output and output them on error.

is this blocking for this PR? can this be considered as a temporary patch and a future refactor can adapt all exec commands to a new scheme?

> These outputs seem more like debug than info?

would this change suffice `s/log.Info/log.Debug/`?

BenTheElder:
> we need to catch errors like that this cycle (aka right now).

Sure

> is this blocking for this PR? can this be considered as a temporary patch and a future refactor can adapt all exec commands to a new scheme?

Not blocking at all, just a statement of intent, we don't surface these subcommand failures well in general.

> would this change suffice s/log.Info/log.Debug/

Yeah. This is a lot of output for Info. But like I said we need to come back to the entirety of logging with better consistency soon.

neolit123:
updated `info` -> `debug`.
now i need to look at the actual problem in kubeadm :grumble:

> Not blocking at all, just a statement of intent, we don't surface these subcommand failures well in general.

yes, i find the current output from debug and info non-informative we need to show a lot more output e.g. possibly `docker exec` commands etc.

> Yeah. This is a lot of output for Info. But like I said we need to come back to the entirety of logging with better consistency soon.

yes, true.
but this means we need to modify the deployer to `debug` too; currently it's `info`.

neolit123:
test infra PR for "debug:
https://github.com/kubernetes/test-infra/pull/11379

kubeadm fix PR:
https://github.com/kubernetes/kubernetes/pull/74277

BenTheElder:
manually merged to unblock debugging, put in a PR to test-infra to make `/override` work for this repo

Issue #306, opened 2019-02-15
kind create cluster fails to create cluster on default install


bart0sh:
Running ```kind create cluster``` just after  ```go get sigs.k8s.io/kind``` install produces this error:

```
$ kind create cluster
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.13.3) 🖼 
 ✓ [control-plane] Creating node container 📦 
 ✓ [control-plane] Fixing mounts 🗻 
 ✓ [control-plane] Starting systemd 🖥
 ✓ [control-plane] Waiting for docker to be ready 🐋 
 ✓ [control-plane] Pre-loading images 🐋 
 ✓ [control-plane] Creating the kubeadm config file ⛵
ERRO[14:34:06] failed to init node with kubeadm: exit status 1  ☸ 
 ✗ [control-plane] Starting Kubernetes (this may take a minute) ☸
ERRO[14:34:06] failed to init node with kubeadm: exit status 1 
Error: failed to create cluster: failed to init node with kubeadm: exit status 1
```

System info:
* openSUSE 42.3 (x86_64)
* go1.11.2 linux/amd64
* Docker 17.09.1-ce

Debug output:
```
$ kind create cluster --loglevel debug
DEBU[14:38:01] Running: /usr/bin/docker [docker ps -q -a --no-trunc --filter label=io.k8s.sigs.kind.cluster --format {{.Names}}\t{{.Label "io.k8s.sigs.kind.cluster"}}] 
Creating cluster "kind" ...
DEBU[14:38:01] Running: /usr/bin/docker [docker inspect --type=image kindest/node:v1.13.3@sha256:d1af504f20f3450ccb7aed63b67ec61c156f9ed3e8b0d973b3dee3c95991753c] 
INFO[14:38:01] Image: kindest/node:v1.13.3@sha256:d1af504f20f3450ccb7aed63b67ec61c156f9ed3e8b0d973b3dee3c95991753c present locally 
 ✓ Ensuring node image (kindest/node:v1.13.3) 🖼
DEBU[14:38:01] Running: /usr/bin/docker [docker info --format '{{json .SecurityOptions}}'] 
[...]

PR #322, opened 2019-02-20
configure Docker daemon to use proxy

bart0sh:
This is a second fix for the proxy issues (see #306 for details)

Docker fails to pull images if running behind the proxy.
It causes weave-net pods to stuck in ImagePullBackOff state.

Configured proxy as explained in docker documentation:
https://docs.docker.com/config/daemon/systemd/#http-proxy
This should solve the issue.

Fixes: #306

### DIFF
--- a/pkg/cluster/internal/create/createcontext.go
+++ b/pkg/cluster/internal/create/createcontext.go
@@ -146,6 +146,12 @@ func (cc *Context) ProvisionNodes() (nodeList map[string]*nodes.Node, err error)
 			return nodeList, err
 		}
 
+		cc.Status.Start(fmt.Sprintf("[%s] Configuring proxy 🐋", configNode.Name))
+		if err := node.SetProxy(); err != nil {
+			// TODO: logging here
+			return nodeList, errors.Wrapf(err, "failed to set proxy for %s", configNode.Name)
+		}
+
 		cc.Status.Start(fmt.Sprintf("[%s] Starting systemd 🖥", configNode.Name))
 		// signal the node container entrypoint to continue booting into systemd
 		if err := node.SignalStart(); err != nil {
--- a/pkg/cluster/nodes/node.go
+++ b/pkg/cluster/nodes/node.go
@@ -319,3 +319,59 @@ func (n *Node) WriteKubeConfig(dest string, hostPort int) error {
 
 	return ioutil.WriteFile(dest, buff.Bytes(), 0600)
 }
+
+// WriteFile writes temporary file on the host
+// and copies it to the node
+func (n *Node) WriteFile(dest, content string) error {
+	// create destination directory
+	cmd := n.Command("mkdir", "-p", filepath.Dir(dest))
+	err := exec.RunLoggingOutputOnFail(cmd)
+	if err != nil {
+		return errors.Wrapf(err, "failed to create directory %s", dest)
+	}
+
+	// create temporary file
+	ftmp, err := ioutil.TempFile("", "")
+	if err != nil {
+		return errors.Wrap(err, "failed to create temporary file")
+	}
+
+	path := ftmp.Name()
+	defer os.Remove(path)
+
+	// write content to the temp file
+	_, err = ftmp.WriteString(content)
+	if err != nil {
+		return errors.Wrap(err, "failed to write content to the file")
+	}
+
+	// copy the temp file from the host to the node
+	if err := n.CopyTo(path, dest); err != nil {
+		return errors.Wrap(err, "failed to copy file to the node")
+	}
+	return nil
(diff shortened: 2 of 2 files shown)


bart0sh:
@BenTheElder looks like failing test is not related to my changes. It's rather related to kubeadm changes. What should I do with this?

pkg/cluster/internal/create/createcontext.go
```diff
@@ -146,6 +146,12 @@ func (cc *Context) ProvisionNodes() (nodeList map[string]*nodes.Node, err error)
 			return nodeList, err
 		}
 
+		cc.Status.Start(fmt.Sprintf("[%s] Configuring proxy 🐋", configNode.Name))
```
neolit123: had to view this RAW and search the U+... in google as none of my fonts support this glyph.
i think we need a different icon for the proxy and not a "whale".

also my suggestion to move these to constants was not accepted. :grumble:

pkg/cluster/nodes/node.go
```diff
@@ -319,3 +319,59 @@ func (n *Node) WriteKubeConfig(dest string, hostPort int) error {
 
 	return ioutil.WriteFile(dest, buff.Bytes(), 0600)
 }
+
+// WriteFile writes temporary file on the host
+// and copies it to the node
+func (n *Node) WriteFile(dest, content string) error {
```
neolit123: didn't we have a method for this already?

neolit123:
/priority important-longterm
/kind feature

> @BenTheElder looks like failing test is not related to my changes. It's rather related to kubeadm changes. What should I do with this?

yes

pkg/cluster/nodes/node.go
bart0sh: no as far as I can see. I'm planning to use this to refactor a couple of places in a separate PR.

pkg/cluster/internal/create/createcontext.go
bart0sh: It worked just fine fine for me in Mac Terminal. As for another icon, so far the code configures proxy for Docker, so I decided to use "whale" glyph.

Can you suggest where can I see a list of supported glyphs?

neolit123: i guess the whale also makes sense. i'm going to leave this topic to @BenTheElder

BenTheElder:
/assign
coming back to this in the morning, I expected to come back after the kubaadm fix upstream, but today was mostly tackling prow.k8s.io problems... 😞

pkg/cluster/internal/create/createcontext.go
BenTheElder: This is fine for now and on par with the rest. I'll probably revisit this UX late next week. Output needs cleanuo.

pkg/cluster/nodes/node.go
```diff
@@ -319,3 +319,59 @@ func (n *Node) WriteKubeConfig(dest string, hostPort int) error {
...
+	defer os.Remove(path)
+
+	// write content to the temp file
+	_, err = ftmp.WriteString(content)
+	if err != nil {
+		return errors.Wrap(err, "failed to write content to the file")
+	}
+
+	// copy the temp file from the host to the node
+	if err := n.CopyTo(path, dest); err != nil {
```
BenTheElder: I have a neat trick to avoid this tempfile but we can do it in a follow up:
`node.Command("cp", "/dev/stdin", dest).SetInput(strings.NewReader(contents)).Run()`
Will probably refactor to do this in other places to deal with docker installed via snap not having access to $TMPDIR.

pkg/cluster/internal/create/createcontext.go
BenTheElder: I'm not sure if we need a status for this step given how quick it should be, but like I said we need to look at cleaning this up anyhow.

Aside: what windows are you using Lubomir? Windows should have emoji fonts and even has some custom nonstandard emoji.

pkg/cluster/internal/create/createcontext.go
```diff
@@ -146,6 +146,12 @@ func (cc *Context) ProvisionNodes() (nodeList map[string]*nodes.Node, err error)
...
+		if err := node.SetProxy(); err != nil {
+			// TODO: logging here
+			return nodeList, errors.Wrapf(err, "failed to set proxy for %s", configNode.Name)
```
BenTheElder: Also slightly iffy on this being a creation failure rather than a warning, but we can revisit this.

pkg/cluster/nodes/node.go
bart0sh: looks nice :)

Issue #118, opened 2018-11-16
Dynamic volume provisioning support


rimusz:
Dynamic volume provisioning support would be handy to have to test apps which need persistence.

BenTheElder:
So we do have a default storage class (host-path), though I haven't really tested it out yet. This is required for some conformance tests.

https://github.com/kubernetes-sigs/kind/blob/4d7dded365359afeb1831292cd1a3a3e15fff0b2/pkg/cluster/const.go#L21-L31

BenTheElder:
I _think_ we just need to document this
/kind documentation

If not, I'll update this issue with what else is required and follow-up.
/help
/priority important-longterm

alejandrox1:
I can document this and test it out

rimusz:
yes, I there is default storage class, but it doesn’t work for dynamic volumes prisioning, deployments get stuck waiting for PVCs

BenTheElder:
Ok, thanks.
We'll need to look into options for providing something more suitable. Any
ideas @davidz627 ?

BenTheElder:
We're inherently not on a cloud provider (these clusters are locally like
minikube etc.) - is NFS the only option? Could we write one that handed out
"local disks" that are just dirs or docker volumes or something..?

I'm not the most experienced with stateful Kubernetes yet, admittedly, but
it seems like we could come up with something. 🤔

msau42:
Sorry can someone explain some context to me? Is this for testing only, or do we actually want to run real production workloads?  If it's testing only, there's a hostpath dynamic provisioner that uses the new volume topology feature to schedule correctly to nodes.  However, it doesn't handle anything like capacity isolation or accounting.

I forgot, someone at rancher was working on this project.  I can't remember the name at the moment though :(

BenTheElder:
Just testing. If anyone uses this project for production workloads they're
crazy 😆

I'll try to look around for that, if you remember or have any other ideas
that would be appreciated though. I wonder if minikube or kinvolk have
tried to sort this out yet as well.

msau42:
I found it @yasker https://github.com/rancher/local-path-provisioner

rimusz:
also https://github.com/kubernetes-incubator/external-storage/tree/master/local-volume

rimusz:
OK, been messing with storage:
1) https://github.com/kubernetes-incubator/external-storage/tree/master/nfs fails to work in `kind`
[...]

Issue #324, opened 2019-02-20
RFE: Support for a phased breakdown of different operations, specifically create


timothysc:
Currently today kind executes a series of implicit steps as part of operations such as `create`.  The purpose of this issue is to make those steps explicit and well-defined ~= `kubeadm phases` such that both developers and consumers can execute different steps as needed for their testing.  

/cc @kubernetes/sig-cluster-lifecycle

BenTheElder:
Making them well defined is the tricky part, as we've added features like HA or proxy support the required steps have changed.

An immediate alternative with even more control to iterate on kubeadm phases was suggested at the kind meeting with @fabriziopandini @neolit123 to support creating nodes but stopping short of any kubeadm provisioning etc.

timothysc:
FWIW not all phases need to be at the same level and can go through promotion, so supportability is also explicit.

BenTheElder:
Right. Thinking more in terms of how to actually break it down internally at this point. 

We will need some more refactor to clean up interdependencies between the creation steps (which we absolutely should do).

We should start by identifying what steps would be helpful.

As far as I could tell the most helpful thing would be to have control over the kubeadm steps, for which a way to provision but stop before `kubeadm init` / `kubeadm join` would probably be a quicker win in the short term.

fabriziopandini:
@BenTheElder I agree? Stopping before init/join is already a well defined use case and having a short term solution in please will be great
I'm going to send a PR for this

BenTheElder:
0.3 is overdue, punting to 0.4, but I've spent some time experimenting with options for this and what the phases should be (we need to make a few more granular).

BenTheElder:
We wound up doing https://github.com/kubernetes-sigs/kind/issues/324#issuecomment-465731832 and it has worked pretty well, as the projects have matured kinder has needed different opinionated phases (e.g. for some time using different CNI installation) anyhow, and these are not the hard party of the project generally, the tricky part is all the "in docker" "in podman" bits, which are re-usable via this mechanism (and other APIs we've exposed).

PR #325, opened 2019-02-20
Add dynamic volume provisioning support

aojea:
Dynamic volume provisioning support is nice to have in order to test applications that need persistence.

[Local Path Provisioner](https://github.com/rancher/local-path-provisioner
) provides a way for the Kubernetes users to utilize the local storage in each node ...

It requires a kubernetes version > 1.12

Ref: https://github.com/kubernetes-sigs/kind/issues/118

### DIFF
--- a/pkg/cluster/internal/create/kubeadm-init.go
+++ b/pkg/cluster/internal/create/kubeadm-init.go
@@ -120,6 +120,40 @@ func runKubeadmInit(ec *execContext, configNode *NodeReplica) error {
 		return errors.Wrap(err, "failed to add default storage class")
 	}
 
+	// add dynamic volume provisioning
+	// local-path provisioner requires kubernetes v1.12+
+	// this can be removed once kind drop support for v1.11
+	// https://github.com/rancher/local-path-provisioner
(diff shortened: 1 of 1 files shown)


aojea:
Please hold on until I can test it locally, I'm having some issues to detect the cluster version

pkg/cluster/internal/create/kubeadm-init.go
```diff
@@ -120,6 +120,40 @@ func runKubeadmInit(ec *execContext, configNode *NodeReplica) error {
 		return errors.Wrap(err, "failed to add default storage class")
 	}
 
+	// add dynamic volume provisioning
```
BenTheElder: /hold
I don't think this should be part of kubeadm init, and I think we should consider if we want to bundle this with the node image or not, and if installing this should be configurable.

aojea:
I'll close this PR and wait until https://github.com/kubernetes-sigs/kind/issues/324 that seems aligned with @neolit123 proposal

BenTheElder:
thanks, I actually want to explore this (this weekend?) and bring up at our next meeting, I do think having this either by default or extremely easy to install is important for a lot of usage.

Issue #191, opened 2018-12-20
improve instructions for windows


BenTheElder:
The CLI in particular should output valid instructions for setting KUBECONFIG on Windows ideally, this might be a bit tricky as technically you could also be in bash on windows, and this may open a whole can of worms to detect correctly :-)

/kind feature
/priority important-longterm
/help

xref #155 #190

alejandrox1:
Can I take this? Need something cool to get back from the holiday slump :smiley:

BenTheElder:
Sure.

I think we might need some tricky business for this one, depending on what shell the user is in. Can't just use POSIX :(

alejandrox1:
So I've been looking up into this. I'll add some more documentation and will focus mostly on using kind in powershell. Git bash seems to more or less work as expected and the command prompt itself behaves very much like powershell.

Please tell me if there is some other type of environment that you think should be tested :-)

BenTheElder:
I think cmd.exe and powershell are indeed the main targets, but I'm not really a windows developer, let's start there and look for feedback in the future :-)

neolit123:
with this PR the CLI prints shell instructions for Windows:
https://github.com/kubernetes-sigs/kind/pull/326

should the docs include these Windows shell instructions here?
https://github.com/kubernetes-sigs/kind/blob/master/site/content/docs/user/quick-start.md#interacting-with-your-cluster

BenTheElder:
yeah, at least they should link to another section with them, I'm not sure if we want to detail every shell inline there 🤔

BenTheElder:
(note this issue is almost solve, but I want to track which issues would be good to see if we're having them at all. if we're not we should aim to write more well-scoped issues...)

neolit123:
do we still need to track this?

BenTheElder:
I think we're in pretty good shape now.

PR #326, opened 2019-02-21
Windows: print message how to use with PowerShell, bash, cmd

neolit123:
/kind feature
/priority important-longterm

ref https://github.com/kubernetes-sigs/kind/issues/191

### DIFF
--- a/pkg/cluster/context.go
+++ b/pkg/cluster/context.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"runtime"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/cluster/context.go
```diff
@@ -188,10 +189,34 @@ func (c *Context) Create(cfg *config.Config, options ...CreateOption) error {
 		return err
 	}
 
-	fmt.Printf(
-		"Cluster creation complete. You can now use the cluster with:\n\nexport KUBECONFIG=\"$(kind get kubeconfig-path --name=%q)\"\nkubectl cluster-info\n",
-		cc.Name(),
-	)
+	// TODO: consider shell detection.
```
neolit123: although, probably best to just to print these 3.

BenTheElder: that SGTM, 100% agree. shell detection could be nice if feasible, but isn't really necessary and this should be pretty helpful

pkg/cluster/context.go
```diff
@@ -188,10 +189,34 @@ func (c *Context) Create(cfg *config.Config, options ...CreateOption) error {
...
+				"You can now use the cluster:\n"+
+				"  kubectl cluster-info\n",
+			cc.Name(),
+		)
+	} else {
+		fmt.Printf(
+			"Cluster creation complete. You can now use the cluster with:\n\n"+
+
+				"  export KUBECONFIG=\"$(kind get kubeconfig-path --name=%q)\"\n"+
+				"  kubectl cluster-info\n",
```
BenTheElder: we should probably remove these leading spaces, bash (And I think other shells) will not store things in history if they have leading spaces, so if you copy and paste these lines they won't be in your history.

PR #327, opened 2019-02-21
spelling: add a new hack/verify-spelling script

neolit123:
- pull and build misspell in a temporary dir and run tests on the
kind tree, while skipping vendor.
- update verify-all to run this test first, and move the govet test
before verify-generated.sh.

/kind feature
/area testing
/priority important-longterm
/hold

### DIFF
--- a/hack/verify-spelling.sh
+++ b/hack/verify-spelling.sh
@@ -1,0 +1,62 @@
+#!/usr/bin/env bash
+# Copyright 2019 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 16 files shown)


BenTheElder:
we have a lot of typos 😁

neolit123:
PR updated.
- made it possible to highlight each error line from misspell in e2e logs (by adding "error: ").
- added spelling fixes.

hack/verify-spelling.sh
```diff
@@ -0,0 +1,62 @@
...
+TMP_DIR=$(mktemp -d)
+
+# cleanup
+exitHandler() (
+  echo "Cleaning up ..."
+  rm -rf "${TMP_DIR}"
+)
+trap exitHandler EXIT
+
+# pull misspell
```
BenTheElder: why are we pulling it?

neolit123: should it be vendored - can you elaborate?

also this is still broken for me:
https://github.com/kubernetes/test-infra/blob/master/hack/verify-spelling.sh

BenTheElder: hmm I guess we can not vendor it, I feel like pulling will be flakier in CI. maybe what we should do is have a podutils job run an image with a misspell binary in it.

BenTheElder: vendoring it is worse for the `go get` users that are not developing kind 🤔

BenTheElder: let's go with this for now and revisit later

PR #330, opened 2019-02-22
library: breakdown create actions

fabriziopandini:
As discussed in https://github.com/kubernetes-sigs/kind/issues/324, this PR provides a quick win for achieving better control over the kubeadm steps during create, and more specifically a way to provision but stop before `kubeadm init` / `kubeadm join` 

The solution leverage on the options management and makes this feature available only at the library level, while if/how to exposed to the Users can be defined in follow-up PRs.

Rif. https://github.com/kubernetes-sigs/kind/issues/324

/assign @BenTheElder 
/cc @timothysc  
/cc @neolit123

### DIFF
--- a/pkg/cluster/context.go
+++ b/pkg/cluster/context.go
@@ -72,7 +72,9 @@ func (c *Context) KubeConfigPath() string {
 // Create provisions and starts a kubernetes-in-docker cluster
 func (c *Context) Create(cfg *config.Config, options ...create.ClusterOption) error {
 	// apply create options
-	opts := &internalcreate.Options{}
+	opts := &internalcreate.Options{
+		SetupKubernetes: true,
+	}
 	for _, option := range options {
 		opts = option(opts)
 	}
--- a/pkg/cluster/create/cluster.go
+++ b/pkg/cluster/create/cluster.go
@@ -40,3 +40,13 @@ func WaitForReady(interval time.Duration) ClusterOption {
 		return o
 	}
 }
+
+// SetupKubernetes configures create command to setup kubernetes after creating nodes containers
+// TODO: Refactor this. It is a temporary solution for a phased breakdown of different
+//      operations, specifically create. see https://github.com/kubernetes-sigs/kind/issues/324
+func SetupKubernetes(setupKubernetes bool) ClusterOption {
+	return func(o *internalcreate.Options) *internalcreate.Options {
+		o.SetupKubernetes = setupKubernetes
+		return o
+	}
+}
--- a/pkg/cluster/internal/create/create.go
+++ b/pkg/cluster/internal/create/create.go
@@ -44,6 +44,9 @@ import (
 type Options struct {
 	Retain       bool
 	WaitForReady time.Duration
+	//TODO: Refactor this. It is a temporary solution for a phased breakdown of different
+	//      operations, specifically create. see https://github.com/kubernetes-sigs/kind/issues/324
+	SetupKubernetes bool // if kind should setup kubernetes after creating nodes
 }
 
 // Cluster creates a cluster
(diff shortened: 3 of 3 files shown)


pkg/cluster/context.go
```diff
@@ -189,6 +198,15 @@ func (c *Context) Create(cfg *config.Config, options ...CreateOption) error {
 		return err
 	}
 
+	if !cc.InstallKubernetes {
+		fmt.Printf(
+			"Cluster creation complete. To setup Kubernetes:\n\n"+
```
neolit123: so i don't think anybody knows what Kubernetes is.
i think it's a Framework the produces binaries that expose an API, not the distro or a kernel as per the popular argument.

having said that we already install Kubernetes by copying it's binaries to the nodes.
would it make sense to call the option "SetupKubernetes" instead of "InstallKubernetes"?

pkg/cluster/context.go
```diff
@@ -189,6 +198,15 @@ func (c *Context) Create(cfg *config.Config, options ...CreateOption) error {
...
+				"  docker exec %s-%s kubeadm init  --config=/kind/kubeadm.conf --ignore-preflight-errors=all\n",
+			cc.Name(), cc.BootStrapControlPlane().Name,
```
neolit123: would it make sense to dump the list of `init` and `join` commands for all nodes?

pkg/cluster/context.go
BenTheElder: yeah, kubernetes is already "installed" to the nodes by this point more or less, setup makes more sense

pkg/cluster/context.go
fabriziopandini: I have added the link to the kubeadm documentation, so the sequence of operation is always updated/versioned

pkg/cluster/internal/create/createcontext.go
```diff
@@ -39,8 +39,9 @@ type Context struct {
 	Status *logutil.Status
 	Config *config.Config
 	*DerivedConfig
-	Retain      bool         // if we should retain nodes after failing to create.
-	ExecOptions []ExecOption // options to be forwarded to the exec command.
+	Retain            bool         // if we should retain nodes after failing to create.
+	InstallKubernetes bool         // if kind should install kubernetes after creating nodes
```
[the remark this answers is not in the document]
fabriziopandini: Yes it is a temporary workaround.
Added TODO + link

fabriziopandini:
@timothysc @neolit123 @BenTheElder Thanks for the feedback!
Everything should be addressed now!

neolit123:
>> would it make sense to dump the list of `init` and `join` commands for all nodes?

> I have added the link to the kubeadm documentation, so the sequence of operation is always updated/versioned

the join command requires IP of the endpoint and boostrap tokens.

some users might ask how the join command will look like if we don't call it automatically:
https://github.com/kubernetes-sigs/kind/blob/master/pkg/cluster/internal/create/kubeadm-join.go#L132-L145

pkg/cluster/internal/create/createcontext.go
```diff
@@ -41,6 +41,9 @@ type Context struct {
 	*DerivedConfig
 	Retain      bool         // if we should retain nodes after failing to create.
 	ExecOptions []ExecOption // options to be forwarded to the exec command.
+	//TODO: Refactor this. It is a temporary solution for a phased breakdown of different
```
BenTheElder: We should probably make the CreateOption be marked temporary as well, particularly since that is actually exported.

fabriziopandini: done

fabriziopandini:
@BenTheElder done!

@neolit123 done, and also added instruction for the CNI addon that has includes some fancy quoting stuff for weave

However, I'm a little bit concerned about going further (e.g. copy certs, join control-plane nodes, copy admin.conf file to the host) for a feature that ATM is only in the library and that is meant for the advanced users. If a user doesn't got the idea after init, CNI and join, probably it is better he/she sticks to the default kind behaviour

neolit123:
@fabriziopandini 

> However, I'm a little bit concerned about going further (e.g. copy certs, join control-plane nodes, copy admin.conf file to the host) for a feature that ATM is only in the library and that is meant for the advanced users. If a user doesn't get the idea after init, CNI and join, probably it is better he/she sticks to the default kind behavior

after posting my comment above, i realized that we might as well just remove such messages entirely.
we are not exposing this on the CLI so this message is for advanced users that already are using the library anyhow.

neolit123:
but this is fine like that too...

fabriziopandini:
@BenTheElder kind reminder
This will unblock work on kinder, and greatly help in simplifying manual testing for this release

PR #331, opened 2019-02-23
pre-load CNI

### LINKED ISSUE #200: pre-load CNI
... IE weave, the images and the manifest, so that we can finally run clusters offline / air-gapped

I've started on this.
/assign
/priority important-soon

BenTheElder:
With this patch a cluster can be created fully offline. Fixes #200 

A test image is at: `gcr.io/bentheelder-kind-dev/kindest/node:v1.13.3`

How this works:
- pkg/build/node now contains some private constants for the CNI, with instructions for udpating them
- we write a fixed CNI manifest to a well-known-location when building the node image
- we pull a fixed set of CNI images and pre-load them when building the node image
- when creating a cluster we detect if the manifest exists in the container, if it does we will load that, if not we will fall back to the previous CNI install mechanism

Along the way I cleaned up the image build slightly, and improved the ergonomics of the `exec.Command` interface to allow method chaining.

### DIFF
--- a/pkg/cluster/internal/create/kubeadm-init.go
+++ b/pkg/cluster/internal/create/kubeadm-init.go
@@ -17,9 +17,10 @@ limitations under the License.
 package create
 
 import (
-	"sigs.k8s.io/kind/pkg/exec"
 	"strings"
 	"time"
+
+	"sigs.k8s.io/kind/pkg/exec"
 
 	log "github.com/sirupsen/logrus"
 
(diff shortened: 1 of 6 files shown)


pkg/cluster/internal/create/kubeadm-init.go
```diff
@@ -103,11 +104,28 @@ func runKubeadmInit(ec *execContext, configNode *NodeReplica) error {
...
+	// first probe for a pre-installed manifest
+	haveDefaultCNIManifest := true
+	if err := node.Command("test", "-f", "/kind/manifests/default-cni.yaml").Run(); err != nil {
+		haveDefaultCNIManifest = false
+	}
+	if haveDefaultCNIManifest {
+		// we found the default manifest, install that
+		// the images should already be loaded along with kubernetes
+		if err := node.Command(
+			"kubectl", "create", "--kubeconfig=/etc/kubernetes/admin.conf",
```
neolit123: bellow we are using the `"/bin/sh", "-c", ...` patter for kubectl.
this seems better.

pkg/cluster/internal/create/kubeadm-init.go
```diff
@@ -103,11 +104,28 @@ func runKubeadmInit(ec *execContext, configNode *NodeReplica) error {
...
+			"-f", "/kind/manifests/default-cni.yaml",
+		).Run(); err != nil {
+			return errors.Wrap(err, "failed to apply overlay network")
+		}
+	} else {
```
neolit123: given now the images are preloaded and manifest is written on the node, shouldn't we exit with an error instead of going for the download from web method as a fallback?

pkg/cluster/internal/create/kubeadm-init.go
BenTheElder: we still pass the kubeconfig via the flag? this is the most explicit way to pass kubeconfig, I don't think using bash is better, we just need to easily exactly mimic the command they specify in their docs

pkg/cluster/internal/create/kubeadm-init.go
BenTheElder: the command below does download from the web?
kind gracefully degrades in the presence of older images without this to the behavior expected for those images  

this is in line with our principles of:
- graceful degredation
- avoiding breaking changes
- avoiding assumptions

pkg/cluster/internal/create/kubeadm-init.go
BenTheElder: I would've strongly preferred to _not_ invoke bash, but the upstream preferred installation from a non-fixed manifest is:

> `kubectl apply -f "https://cloud.weave.works/k8s/net?k8s-version=$(kubectl version | base64 | tr -d '\n')"`

https://www.weave.works/docs/net/latest/kubernetes/kube-addon/#-installation

BenTheElder:
/hold cancel
Will move that image and file a follow up to default to it.

kargakis:
@BenTheElder a howto in the docs would be nice for this - it's not clear what are the exact deps/tools we need to have offline to deploy a cluster.

BenTheElder:
@kargakis noted -- right now you just need to install from HEAD and pull the node image. `docker pull kindest/node:v1.13.3`, `kind create cluster` should work

kargakis:
Yeap, I confirm I can create offline clusters now. Thanks!

 :tada: :champagne: :tada: :champagne:

BenTheElder:
Awesome! We'll definitely document this and ways to work with it 👍

PR #332, opened 2019-02-23
nodes/tryUntil: add argument for Sleep inside the for loop

neolit123:
/kind cleanup
can adjust sleep times if needed.

### DIFF
--- a/pkg/cluster/nodes/node.go
+++ b/pkg/cluster/nodes/node.go
@@ -102,7 +102,7 @@ func (n *Node) CopyFrom(source, dest string) error {
 // WaitForDocker waits for Docker to be ready on the node
 // it returns true on success, and false on a timeout
 func (n *Node) WaitForDocker(until time.Time) bool {
-	return tryUntil(until, func() bool {
+	return tryUntil(until, time.Second, func() bool {
 		cmd := n.Command("systemctl", "is-active", "docker")
 		out, err := exec.CombinedOutputLines(cmd)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/cluster/nodes/node.go
```diff
@@ -102,7 +102,7 @@ func (n *Node) CopyFrom(source, dest string) error {
 // WaitForDocker waits for Docker to be ready on the node
 // it returns true on success, and false on a timeout
 func (n *Node) WaitForDocker(until time.Time) bool {
-	return tryUntil(until, func() bool {
+	return tryUntil(until, time.Second, func() bool {
```
BenTheElder: hmm, how did we pick these values?

the downside to this is increasing the latency from ready -> cluster up progressing
these commands are somewhat self throttling by taking time themselves.

neolit123: > hmm, how did we pick these values?

experiment from yesterday.

> the downside to this is increasing the latency from ready -> cluster up progressing
these commands are somewhat self throttling by taking time themselves.

it executes something like a 100 kubectl calls until it gets the signal that the CP is ready.

BenTheElder: > it executes something like a 100 kubectl calls until it gets the signal that the CP is ready.

sounds working as intended 😉 

the docker call should be much more throttled, this one is much slower

tao12345666333: SGTM.  :+1  Do we need to do more experiments on different performance machines? I have some local machines available for testing.

BenTheElder:
FWIW I do think this probably makes sense, but I'd like to use a more rigorous value and possibly not throttle waiting for docker 🤔

PR #334, opened 2019-02-23
upgrade default node image

BenTheElder:
picks up #331
/hold

### DIFF
--- a/pkg/cluster/config/defaults/image.go
+++ b/pkg/cluster/config/defaults/image.go
@@ -18,4 +18,4 @@ limitations under the License.
 package defaults
 
 // Image is the default for the Config.Image field, aka the default node image.
-const Image = "kindest/node:v1.13.3@sha256:d1af504f20f3450ccb7aed63b67ec61c156f9ed3e8b0d973b3dee3c95991753c"
+const Image = "kindest/node:v1.13.3@sha256:f51da441e32b69363f0979753b3e045ff7d17f7f3dbff38ca9a39ca8d920bdf1"


pkg/cluster/config/defaults/image.go
```diff
@@ -18,4 +18,4 @@ limitations under the License.
 package defaults
 
 // Image is the default for the Config.Image field, aka the default node image.
-const Image = "kindest/node:v1.13.3@sha256:d1af504f20f3450ccb7aed63b67ec61c156f9ed3e8b0d973b3dee3c95991753c"
+const Image = "kindest/node:v1.13.3"
```
BenTheElder: I'm extremely on the fence about pinning the image fully or not.

It's still possible to use a pinned image via `--image`, otoh being reproducible-ish out of the box is nice.
Any change we push to these images is arguably breaking, pinning avoids that 🤔

OTOH this allows us to push things like docker security fixes.

neolit123: i don't think pinning the image to the k8s patch release cycle is a good idea.

complex version schemas is a solution for that:
`kindest/node:v1.13.3-0.1.0`

where `0.1.0` is the kind version.
thus with a kind bump the image can be bumped too on security fixes.

BenTheElder: `kindest/node:v1.13.3-0.1.0` is no different (in fact more complex and weaker) than putting the hash

BenTheElder: at this point IMO users should not depend on the specific contents of the node image, and kind itself handles this kind of image change fine.

neolit123: it's defenitely weaker than the hash but stronger than having the k8s release only.

BenTheElder: there might be a further argument to just make the default `kindest/node:v1.13` for k8s fixes, that one seems a bit iffier though, as the k8s version is the one thing you _should_ depend on.

BenTheElder: > it's definitely weaker than the hash but stronger than having the k8s release only.

I mean if we want to pin based on kind release, we should just use the hash.
But I don't think we actually need this, kind will work with any image we've ever published. Image changes are graceful.

The things we have "broken" like docker version aren't really intended to be a public detail, though we unfortunately can't hide these.

I'll look around for prior art on this from other tools...

neolit123: > I'm extremely on the fence about pinning the image fully or not.

i'm +1 on keeping the existing behavior.
ye, let me know if you need LGTM.

BenTheElder: kubeadm itself seems to pin to `stable-1`? 
https://github.com/kubernetes/kubernetes/blob/b96378c058c99180b84efb259ec56e5ee31941ae/cmd/kubeadm/app/apis/kubeadm/v1beta1/defaults.go#L36:2

neolit123: yes, it does. but we are bound to the k8s release cycle.

BenTheElder: AFAICT the `stable-1` is not bound though, it's looked up from the published releases?
that would be like if we used `kindest/node:stable-1`

BenTheElder: i'm going to add the hash for now and we can revisit this at the meeting or something. I think it's worth considering futher before we get too many releases in.

BenTheElder:
/hold cancel
added the sha

pkg/cluster/config/defaults/image.go
neolit123: > AFAICT the `stable-1` is not bound though, it's looked up from the published releases?

it tends to resolve to latest stable - e.g. currently 1.13.x

not binding to the k8s release allows you to push immediate fixes and not wait on a upstream PATCH bump, to bump an image for a docker fix. but either way is fine, as minors get pushed relatively often.

/lgtm

Issue #166, opened 2018-12-11
Support arm64


vielmetti:
Device under test is a Packet c1.large.arm 96-core arm64 machine running Ubuntu 18.04.

```
ed@ed-2a-bcc-llvm:~$ go version
go version go1.11.2 linux/arm64
ed@ed-2a-bcc-llvm:~$ go get sigs.k8s.io/kind
ed@ed-2a-bcc-llvm:~$ go/bin/kind create cluster
Creating cluster 'kind-1' ...
 ✓ Ensuring node image (kindest/node:v1.12.2)  
 ✓ [kind-1-control-plane] Creating node container 📦 
 ✗ [kind-1-control-plane] Fixing mounts 🗻 
Error: failed to create cluster: exit status 1
Usage:  
  kind create cluster [flags]
        
Flags:  
      --config string   path to a kind config file
  -h, --help            help for cluster
      --image string    node docker image to use for booting the cluster
      --name string     cluster context name (default "1")
      --retain          retain nodes for debugging when cluster creation fails
      --wait duration   Wait for control plane node to be ready (default 0s)

Global Flags:
      --loglevel string   logrus log level [panic, fatal, error, warning, info, debug] (default "warning")

failed to create cluster: exit status 1
ed@ed-2a-bcc-llvm:~$ docker version
Client:
 Version:           18.09.0
 API version:       1.39
 Go version:        go1.10.4
 Git commit:        4d60db4
 Built:             Wed Nov  7 00:52:41 2018
 OS/Arch:           linux/arm64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          18.09.0
  API version:      1.39 (minimum version 1.12)
  Go version:       go1.10.4
  Git commit:       4d60db4
  Built:            Wed Nov  7 00:17:01 2018
  OS/Arch:          linux/arm64
  Experimental:     false
```

vielmetti:
Looking at https://github.com/kubernetes-sigs/kind/blob/master/pkg/cluster/context.go#L196-L206 where the code errors out, there's a TODO for logging from @BenTheElder .

vielmetti:
Looks like the underlying reason is that kindest/node is not multiarchitecture.

```
ed@ed-2a-bcc-llvm:~$ docker run --rm mplatform/mquery kindest/node:v1.12.2
Image: kindest/node:v1.12.2
 * Manifest List: No
 * Supports: amd64/linux
```

tao12345666333:
[...]

PR #335, opened 2019-02-23
switch off of the bazel debs

BenTheElder:
easiest answer to fixing the bazel build after cross compliation landed.
/hold

positive:
- the bazel build and the non-bazel build have the same install process now
- we can probably refactor the build to only COPY operations and support cross-compilation #166
- no longer broken by changes to the debs, the binaries are less in-flux

negative:
- no longer installing the debs

### DIFF
--- a/pkg/build/kube/bazelbuildbits.go
+++ b/pkg/build/kube/bazelbuildbits.go
@@ -17,17 +17,21 @@ limitations under the License.
 package kube
 
 import (
+	"fmt"
 	"os"
 	"path"
 	"path/filepath"
+	"runtime"
 
-	log "github.com/sirupsen/logrus"
+	"github.com/pkg/errors"
 	"sigs.k8s.io/kind/pkg/exec"
 )
 
 // BazelBuildBits implements Bits for a local Bazel build
 type BazelBuildBits struct {
 	kubeRoot string
+	// computed at build time
+	paths map[string]string
 }
 
 var _ Bits = &BazelBuildBits{}
@@ -56,16 +60,17 @@ func (b *BazelBuildBits) Build() error {
 	// make sure we cd back when done
 	defer os.Chdir(cwd)
 
+	// TODO(bentheelder): we assume the host arch, but cross compiling should
+	// be possible now
+	bazelGoosGoarch := fmt.Sprintf("linux_%s", runtime.GOARCH)
+
 	// build artifacts
 	cmd := exec.Command(
 		"bazel", "build",
-		// TODO(bentheelder): we assume linux amd64, but we could select
-		// this based on Arch etc. throughout, this flag supports GOOS/GOARCH
-		"--platforms=@io_bazel_rules_go//go/toolchain:linux_amd64",
-		// we want the debian packages
-		"//build/debs:debs",
+		fmt.Sprintf("--platforms=@io_bazel_rules_go//go/toolchain:%s", bazelGoosGoarch),
+		// node installed binaries
+		"//cmd/kubeadm:kubeadm", "//cmd/kubectl:kubectl", "//cmd/kubelet:kubelet",
 		// and the docker images
-		//"//cluster/images/hyperkube:hyperkube.tar",
 		"//build:docker-artifacts",
 	)
 	exec.InheritOutput(cmd)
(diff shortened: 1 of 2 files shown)


BenTheElder:
all the builds passed this time, pushing a slight cleanup / refactor

pkg/build/kube/bazelbuildbits.go
```diff
@@ -56,74 +60,114 @@ func (b *BazelBuildBits) Build() error {
...
 // Install implements Bits.Install
 func (b *BazelBuildBits) Install(install InstallContext) error {
-	base := install.BasePath()
-
-	// install debians
-	debs := path.Join(base, "debs", "*.deb")
-	if err := install.Run("/bin/sh", "-c", "dpkg -i "+debs); err != nil {
-		log.Errorf("Debian install failed! %v", err)
-		return err
+	kindBinDir := path.Join(install.BasePath(), "bin")
```
BenTheElder: this is the install logic from the non-bazel build, they match now

pkg/build/kube/bazelbuildbits.go
```diff
@@ -56,74 +60,114 @@ func (b *BazelBuildBits) Build() error {
 	// make sure we cd back when done
 	defer os.Chdir(cwd)
 
+	// TODO(bentheelder): we assume the host arch, but cross compiling should
+	// be possible now
```
munnerz: "we assume the host arch" won't this be invalid on OSX?

pkg/build/kube/bazelbuildbits.go
```diff
@@ -56,74 +60,114 @@ func (b *BazelBuildBits) Build() error {
...
 	cmd := exec.Command(
 		"bazel", "build",
-		// TODO(bentheelder): we assume linux amd64, but we could select
-		// this based on Arch etc. throughout, this flag supports GOOS/GOARCH
-		"--platforms=@io_bazel_rules_go//go/toolchain:linux_amd64",
+		fmt.Sprintf("--platforms=@io_bazel_rules_go//go/toolchain:%s", bazelGoosGoarch),
 		// we want the debian packages
-		"//build/debs:debs",
+		//"//build/debs:debs",
+		// host installed binaries
```
munnerz: I'm confused by this comment - why are we building kubelet for the host?

pkg/build/kube/bazelbuildbits.go
```diff
@@ -56,74 +60,114 @@ func (b *BazelBuildBits) Build() error {
 	// make sure we cd back when done
 	defer os.Chdir(cwd)
 
+	// TODO(bentheelder): we assume the host arch, but cross compiling should
+	// be possible now
+	bazelGoosGoarch := fmt.Sprintf("%s_%s", runtime.GOOS, runtime.GOARCH)
```
BenTheElder: actually GOOS should be fixed to linux only

pkg/build/kube/bazelbuildbits.go
```diff
@@ -56,74 +60,114 @@ func (b *BazelBuildBits) Build() error {
...
+
 	// build artifacts
 	cmd := exec.Command(
 		"bazel", "build",
-		// TODO(bentheelder): we assume linux amd64, but we could select
-		// this based on Arch etc. throughout, this flag supports GOOS/GOARCH
-		"--platforms=@io_bazel_rules_go//go/toolchain:linux_amd64",
+		fmt.Sprintf("--platforms=@io_bazel_rules_go//go/toolchain:%s", bazelGoosGoarch),
```
munnerz: Shouldn't this always be linux?

pkg/build/kube/bazelbuildbits.go
BenTheElder: the arch is still amd64 😅 
also kubelet's CGO binary was not buildable natively on mac with bazel unless you did some fun crosstool stuff AFAIK. I don't think this is supported yet.

pkg/build/kube/bazelbuildbits.go
BenTheElder: just fixed that!

pkg/build/kube/bazelbuildbits.go
BenTheElder: er yeah host is confusing here, what I mean is "binaries on the """"host""""" (node) as opposed to docker-wrapped binaries (the component images)

BenTheElder: re-worded the comment to reference the node

pkg/build/kube/bazelbuildbits.go
```diff
@@ -56,74 +60,112 @@ func (b *BazelBuildBits) Build() error {
...
 	// https://docs.bazel.build/versions/master/output_directories.html
 	binDir := filepath.Join(b.kubeRoot, "bazel-bin")
 	buildDir := filepath.Join(binDir, "build")
-	return map[string]string{
-		// debians
-		filepath.Join(buildDir, "debs", "kubeadm.deb"):        "debs/kubeadm.deb",
-		filepath.Join(buildDir, "debs", "kubelet.deb"):        "debs/kubelet.deb",
-		filepath.Join(buildDir, "debs", "kubectl.deb"):        "debs/kubectl.deb",
-		filepath.Join(buildDir, "debs", "kubernetes-cni.deb"): "debs/kubernetes-cni.deb",
-		filepath.Join(buildDir, "debs", "cri-tools.deb"):      "debs/cri-tools.deb",
```
munnerz: How are cri-tools and kubernetes-cni now installed? Don't see any references to them elsewhere 🙄

pkg/build/kube/bazelbuildbits.go
BenTheElder: `bazelGoosGoarch := fmt.Sprintf("linux_%s", runtime.GOARCH)` now 🙏 

this doesn't work on mac / windows anyhow though, FWIW. at least for the moment they need to not use `--type=bazel` when building kubernetes node images (we should document this in the docs!)

pkg/build/kube/bazelbuildbits.go
BenTheElder: they're installed in the base image, this is now on par with the make/bash/docker build.

these debs just wrap downloading a copy of a stable release of these based on WORKSPACE, we do ~the same in the base image already.

BenTheElder: or rather CNI is:
https://github.com/kubernetes-sigs/kind/blob/6f430d3884438622ec8e9b4483d235c35b8938e4/images/base/Dockerfile#L82-L92
but cri-tools was only to satisfy the deb install requirements, we don't use those while we're on the dockershim path.

BenTheElder: if/when we take a poke at CRI path we can install these to the base image, they have stable releases and don't change often, and IIRC are backwards/forwards compatible.

BenTheElder:
all of the builds passed again, so we're just waiting for the e2es 😅

PR #337, opened 2019-02-23
Added kind build node --type limitations

alejandrox1:
/kind documentation

### DIFF
--- a/site/content/docs/user/quick-start.md
+++ b/site/content/docs/user/quick-start.md
@@ -137,6 +137,10 @@ Currently, kind supports three different ways to build a `node-image`: via
 `apt`, or if you have the [Kubernetes][kubernetes] source in your host machine
 (`$GOPATH/src/k8s.io/kubernetes`), by using `docker` or `bazel`.
 To specify the build type use the flag `--type`.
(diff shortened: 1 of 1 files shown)


site/content/docs/user/quick-start.md
```diff
@@ -137,6 +137,10 @@ Currently, kind supports three different ways to build a `node-image`: via
 `apt`, or if you have the [Kubernetes][kubernetes] source in your host machine
 (`$GOPATH/src/k8s.io/kubernetes`), by using `docker` or `bazel`.
 To specify the build type use the flag `--type`.
+Note however that using `--type=bazel` in Windows or Mac will not work as some
```
BenTheElder: let's say "will not work currently" as this is a solve-able problem, it just isn't yet. it is also only kubelet.

```
Note however that using `--type=bazel` on Windows or MacOS will not work currently due to Kubelet using [CGO] which requires GCC/glibc for linux. A workaround may be enabled in the future.


[CGO]: https://golang.org/cmd/cgo/
```

We won't mention it here, but test-infra/planter integration is one possible option to explore later.

alejandrox1: Fixed

PR #342, opened 2019-02-24
fix TODO for setting hostport in the kubeconfig file for the host

fabriziopandini:
This PR fixes a TODO related to HA (sorry I missed that in the past)

More specifically, when copying the kubeconfig file to the host, now kind uses the load balancer `haproxy.ControlPlanePort` in case a load balancer exists, otherwise kind uses the control-plane `kubeadm.APIServerPort` (the latter is the only case managed ATM 😞)

/assign @neolit123  @BenTheElder
/priority important-soon
/kind cleanup

### DIFF
--- a/pkg/cluster/internal/create/actions/kubeadminit/init.go
+++ b/pkg/cluster/internal/create/actions/kubeadminit/init.go
@@ -25,6 +25,7 @@ import (
 	"github.com/pkg/errors"
 
 	"sigs.k8s.io/kind/pkg/cluster/internal/create/actions"
+	"sigs.k8s.io/kind/pkg/cluster/internal/haproxy"
 	"sigs.k8s.io/kind/pkg/cluster/internal/kubeadm"
 	"sigs.k8s.io/kind/pkg/cluster/nodes"
 	"sigs.k8s.io/kind/pkg/exec"
@@ -79,12 +80,9 @@ func (a *action) Execute(ctx *actions.ActionContext) error {
 	// usable with kubectl.
 	// the kubeconfig file created by kubeadm internally to the node
 	// must be modified in order to use the random host port reserved
-	// for the API server and exposed by the node
-
-	// retrives the random host where the API server is exposed
-	// TODO(fabrizio pandini): when external load-balancer will be
-	//      implemented this should be modified accordingly
-	hostPort, err := node.Ports(kubeadm.APIServerPort)
+	// for for the control plane in case a load balancer exists, or for
+	// the API server in case of single control-plane node
+	hostPort, err := getHostPort(allNodes)
 	if err != nil {
 		return errors.Wrap(err, "failed to get kubeconfig from node")
 	}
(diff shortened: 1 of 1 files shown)


pkg/cluster/internal/create/kubeadm-init.go
```diff
@@ -155,6 +154,29 @@ func runKubeadmInit(ec *execContext, configNode *NodeReplica) error {
...
+func getHostPort(ec *execContext, controlPlane *nodes.Node) (int, error) {
+	// in single control-plane cluster, the node exposing the API server to the outside world is the control-plane itself
+	node := controlPlane
+	port := kubeadm.APIServerPort
+
+	// in case an lb is defined, use the lb node instead
+	if ec.ExternalLoadBalancer() != nil {
+		loadBalancerHandle, ok := ec.NodeFor(ec.ExternalLoadBalancer())
+		if !ok {
+			return 0, errors.Errorf("unable to get the handle for operating on node: %s", ec.ExternalLoadBalancer().Name)
```
neolit123: optional nit: `unable to get the handle of load balancer node: %s` ?

pkg/cluster/internal/create/kubeadm-init.go
```diff
@@ -155,6 +154,29 @@ func runKubeadmInit(ec *execContext, configNode *NodeReplica) error {
...
+		}
+
+		node = loadBalancerHandle
+		port = haproxy.ControlPlanePort
```
neolit123: with respect to debugging e2e, would it make sense to add a `log.Info` message to indicate that we are using the LB port?

pkg/cluster/internal/create/kubeadm-init.go
```diff
@@ -90,9 +91,7 @@ func runKubeadmInit(ec *execContext, configNode *NodeReplica) error {
 	// for the API server and exposed by the node
 
 	// retrives the random host where the API server is exposed
```
neolit123: possibly remove this comment and add a comment above `getHostPort()`?

pkg/cluster/internal/create/kubeadm-init.go
fabriziopandini: done

pkg/cluster/internal/create/kubeadm-init.go
```diff
@@ -155,6 +152,32 @@ func runKubeadmInit(ec *execContext, configNode *NodeReplica) error {
...
+		node = loadBalancerHandle
+		port = haproxy.ControlPlanePort
+	}
+
+	hostPort, err := node.Ports(port)
+	if err != nil {
+		return 0, errors.Wrapf(err, "failed to get Port from node: %s", node.String())
+	}
+
+	log.Infof("kubernetes cluster is accessible from the host using port %d exposed by node %s", hostPort, node.String())
```
BenTheElder: tiny nit/suggestion: `s/is/should be/` -- we haven't actually verified that it's accessible, this kind of message can be frustrating when something isn't working 😅

fabriziopandini:
/close 
Due to https://github.com/kubernetes-sigs/kind/pull/418

pkg/cluster/internal/create/actions/kubeadminit/init.go
```diff
@@ -143,6 +141,33 @@ func (a *action) Execute(ctx *actions.ActionContext) error {
 	return nil
 }
 
+// getHostPort retrives the random host where the API server/the control plane is exposed
+func getHostPort(allNodes []nodes.Node) (int, error) {
```
BenTheElder: :facepalm: I knew this problem was familiar in #418 ...

Thanks for this. Sorry it didn't get in sooner.

PR #346, opened 2019-02-25
base-image: inherit the host environment when building

neolit123:
With modules enabled in the repo and without this change,
errors seem to be thrown when building the base image - e.g. for
GOCACHE or HOME not being defined.

Inherit the host environment when calling `go build`
for the entry point binary.

https://github.com/golang/go/issues/29267

/kind bug
/priority important-longterm


neolit123:
> I assume Go takes the last value for GOOS etc?

yes,
https://golang.org/pkg/os/exec/#Cmd

> // If Env contains duplicate environment keys, only the last
        // value in the slice for each duplicate key is used.

BenTheElder:
Thanks!
/hold cancel

Issue #340, opened 2019-02-23
RFC: v1alpha3 config proposal


BenTheElder:
I've been thinking for a while now about how to improve the ergonomics of our config and extend it for more use cases. I intend to bring this up at the next meeting. Proposal is as follows:

----

# Proposed changes

### Rationale
This is more precise, it specifically configures a kind cluster, whereas some of the sub objects configure nodes etc. Additionally, `config.Cluster` is a better library type name than `config.Config`

### Cons
This is a bikeshed. They both work. Users will need to update (but the same is pretty much true for changing the API version).

## Allow specifying extra ports on nodes

Specifically allow config of the form:

```yaml
apiVersion: kind.sigs.k8s.io/v1alpha3
kind: Cluster
nodes:
- role: worker
  nodePorts: [8080]
- role: worker
  nodePorts: [9090, 7070]
```

### Rationale

In single node mode this will be especially easy to use and allow serving traffic from workloads with Kubernetes `nodePort`s.

This should solve https://github.com/kubernetes-sigs/kind/issues/99, which is one of the earliest standing requests.

### Cons

For multi-node we will likely need to label the nodes with the ports or make the node names predictable or configurable. Suggestions are welcome.

This may also not be the most favorable solution long term versus say, implementing a form of `LoadBalancer` / `Ingress`, but should be much easier to solve and unblock usage.

## Add a `networking` struct to `Cluster`

Specifically:
```yaml
apiVersion: kind.sigs.k8s.io/v1alpha3
kind: Cluster
networking:
  # 127.0.0.1 if unset, affects all nodes
  listenAddress: '0.0.0.0'
  apiServer:
    # random if unset, applies to the control plane node or load balancer
    listenPort: 6443
nodes:
 ....
```

### Rationale

There's increasing demand for configuring this. Additionally, we need to change the defaut listen address to 127.0.0.1 for security reasons ... users will need a way to configure this.

These fields are similar to kubeadm's config.

### Cons

I'm not aware of any. Minor complexity increase?

## Drop the `load-balancer` node role

... but keep the node type internally, and mark it as "control-plane-load-balancer" (or something similarly precise)

### Rationale

We only actually support the control plane load balancer, which is implicitly necessary for multiple control-plane clusters. Nothing about the control plane is currently configurable, and if it were it would actually be cluster-wide details (like the api server listen address).

We already intend to implicitly add these to `v1alpha2` https://github.com/kubernetes-sigs/kind/pull/274

### Cons

[...]

PR #348, opened 2019-02-25
Add IPv6 support

### LINKED ISSUE #280: IPv6 Support

aojea:
This PR adds allows creating IPv6 Kubernetes clusters with `kind` and have in mind a future dual-stack implementation, considering for simplicity, only one address of each protocol.

It adds a new option `ipFamily` to the v1alpha3 API that allows choosing the IP family of the cluster.
To avoid issues with the different networking options the **podSubnet** and the **serviceSubnet** `kubeadm` values are predefined with the following values:

```
	Default PodSubnet          = "10.244.0.0/16"
	Default ServicesSubnet     = "10.96.0.0/12"
	Default PodSubnetIPv6      = "fd00:10:244::/64"
	Default ServicesSubnetIPv6 = "fd00:10:96::/112"
```

We can create a Kubernetes IPv6 cluster with the following config:

```
# necessary for conformance
kind: Cluster
networking:
  ipFamily: ipv6
nodes:
# the control plane node
- role: control-plane
- role: worker
- role: worker
```

**Test results with IPv4 and IPv6**
* [Conformance Tests](https://circleci.com/workflow-run/b898d823-9683-4ef2-b8ed-fba0fedf6dd5)

**References:** 
* https://github.com/kubernetes-sigs/kind/issues/280
* https://docs.google.com/document/d/17e3TWWLfnIZrsVxpln9wNi4x0JVn2oHIHDYjaeENdVE/edit?ts=5c9af5b4#heading=h.on33tp91ehzk

Fixes #280

### DIFF
--- a/hack/ci/e2e.sh
+++ b/hack/ci/e2e.sh
@@ -110,6 +110,7 @@ nodes:
 - role: worker
 - role: worker
 EOF
+
     # mark the cluster as up for cleanup
     # even if kind create fails, kind delete can clean up after it
     KIND_IS_UP=true
@@ -131,6 +132,34 @@ run_tests() {
     # export the KUBECONFIG
     KUBECONFIG="$(kind get kubeconfig-path)"
     export KUBECONFIG
+
+    if [[ "${IP_FAMILY:-ipv4}" == "ipv6" ]]; then
+        # Create the CoreDNS config for offline and IPv6 clusters
+        # https://github.com/coredns/coredns/issues/2494#issuecomment-457215452
+        cat <<EOF | kubectl apply -f -
+---
+apiVersion: v1
+data:
+  Corefile: |
+    .:53 {
+        errors
+        health
+        rewrite name google.com my-google.default.svc.cluster.local
+        kubernetes cluster.local in-addr.arpa ip6.arpa {
+           pods insecure
+        }
+        prometheus :9153
+        cache 30
+        reload
+        loadbalance
+    }
+kind: ConfigMap
+metadata:
+  name: coredns
+  namespace: kube-system
+---
+EOF
+    fi
 
     # base kubetest args
     KUBETEST_ARGS="--provider=skeleton --test --check-version-skew=false"
--- a/hack/ci/e2e.sh
+++ b/hack/ci/e2e.sh
@@ -102,6 +102,8 @@ create_cluster() {
 # necessary for conformance
 kind: Cluster
 apiVersion: kind.sigs.k8s.io/v1alpha3
+networking:
+  ipFamily: ${IP_FAMILY:-ipv4}
 nodes:
 # the control plane node
 - role: control-plane
--- a/pkg/cluster/config/default.go
+++ b/pkg/cluster/config/default.go
@@ -41,14 +41,32 @@ func SetDefaults_Cluster(obj *Cluster) {
 			},
 		}
 	}
-	// default to listening on 127.0.0.1:randomPort
-	// TODO(bentheelder): this defaulting will need to be ipv6 aware as well
+	if obj.Networking.IPFamily == "" {
+		obj.Networking.IPFamily = "ipv4"
+	}
+	// default to listening on 127.0.0.1:randomPort on ipv4
+	// and [::1]:randomPort on ipv6
 	if obj.Networking.APIServerAddress == "" {
 		obj.Networking.APIServerAddress = "127.0.0.1"
+		if obj.Networking.IPFamily == "ipv6" {
+			obj.Networking.APIServerAddress = "::1"
+		}
 	}
 	// default the pod CIDR
 	if obj.Networking.PodSubnet == "" {
 		obj.Networking.PodSubnet = "10.244.0.0/16"
+		if obj.Networking.IPFamily == "ipv6" {
+			obj.Networking.PodSubnet = "fd00:10:244::/64"
+		}
+	}
+	// default the service CIDR using the kubeadm default
+	// https://github.com/kubernetes/kubernetes/blob/746404f82a28e55e0b76ffa7e40306fb88eb3317/cmd/kubeadm/app/apis/kubeadm/v1beta2/defaults.go#L32
+	// Note: kubeadm is doing it already but this simplifies kind's logic
+	if obj.Networking.ServiceSubnet == "" {
+		obj.Networking.ServiceSubnet = "10.96.0.0/12"
+		if obj.Networking.IPFamily == "ipv6" {
+			obj.Networking.ServiceSubnet = "fd00:10:96::/112"
+		}
 	}
 }
 
--- a/pkg/cluster/config/types.go
+++ b/pkg/cluster/config/types.go
@@ -88,6 +88,8 @@ const (
 
 // Networking contains cluster wide network settings
 type Networking struct {
+	// IPFamily is the network cluster model, currently it can be ipv4 or ipv6
+	IPFamily ClusterIPFamily
 	// APIServerPort is the listen port on the host for the Kubernetes API Server
 	// Defaults to a random port on the host
 	APIServerPort int32
@@ -99,7 +101,20 @@ type Networking struct {
 	// PodSubnet is the CIDR used for pod IPs
 	// kind will select a default if unspecified
 	PodSubnet string
+	// ServiceSubnet is the CIDR used for services VIPs
+	// kind will select a default if unspecified
+	ServiceSubnet string
 	// If DisableDefaultCNI is true, kind will not install the default CNI setup.
 	// Instead the user should install their own CNI after creating the cluster.
 	DisableDefaultCNI bool
 }
+
+// ClusterIPFamily defines cluster network IP family
+type ClusterIPFamily string
+
+const (
+	// IPv4Family sets ClusterIPFamily to ipv4
+	IPv4Family ClusterIPFamily = "ipv4"
+	// IPv6Family sets ClusterIPFamily to ipv6
+	IPv6Family ClusterIPFamily = "ipv6"
+)
--- a/pkg/cluster/config/v1alpha3/types.go
+++ b/pkg/cluster/config/v1alpha3/types.go
@@ -88,6 +88,8 @@ const (
 
 // Networking contains cluster wide network settings
 type Networking struct {
+	// IPFamily is the network cluster model, currently it can be ipv4 or ipv6
+	IPFamily ClusterIPFamily `json:"ipFamily,omitempty"`
 	// APIServerPort is the listen port on the host for the Kubernetes API Server
 	// Defaults to a random port on the host
 	APIServerPort int32 `json:"apiServerPort,omitempty"`
@@ -99,7 +101,20 @@ type Networking struct {
 	// PodSubnet is the CIDR used for pod IPs
 	// kind will select a default if unspecified
 	PodSubnet string `json:"podSubnet,omitempty"`
+	// ServiceSubnet is the CIDR used for services VIPs
+	// kind will select a default if unspecified for IPv6
+	ServiceSubnet string `json:"serviceSubnet,omitempty"`
 	// If DisableDefaultCNI is true, kind will not install the default CNI setup.
 	// Instead the user should install their own CNI after creating the cluster.
 	DisableDefaultCNI bool `json:"disableDefaultCNI,omitempty"`
 }
+
+// ClusterIPFamily defines cluster network IP family
+type ClusterIPFamily string
+
+const (
+	// IPv4Family sets ClusterIPFamily to ipv4
+	IPv4Family ClusterIPFamily = "ipv4"
+	// IPv6Family sets ClusterIPFamily to ipv6
+	IPv6Family ClusterIPFamily = "ipv6"
+)
--- a/pkg/cluster/internal/create/nodes.go
+++ b/pkg/cluster/internal/create/nodes.go
@@ -103,9 +103,16 @@ func createNodeContainers(
 		desiredNode := desiredNode // capture loop variable
 		fns = append(fns, func() error {
 			// create the node into a container (~= docker run -d)
-			_, err := desiredNode.Create(clusterLabel)
+			node, err := desiredNode.Create(clusterLabel)
+			if err != nil {
+				return err
+			}
+			if desiredNode.IPv6 {
+				err = node.EnableIPv6()
+			}
 			return err
 		})
+
 	}
 	if err := concurrent.UntilError(fns); err != nil {
 		return err
@@ -125,6 +132,7 @@ type nodeSpec struct {
 	// TODO(bentheelder): replace with a cri.PortMapping when we have that
 	APIServerPort    int32
 	APIServerAddress string
+	IPv6             bool
 }
 
 func nodesToCreate(cfg *config.Cluster, clusterName string) []nodeSpec {
@@ -154,6 +162,11 @@ func nodesToCreate(cfg *config.Cluster, clusterName string) []nodeSpec {
 		}
 	}
 	isHA := controlPlanes > 1
+	// obtain IP family
+	ipv6 := false
+	if cfg.Networking.IPFamily == "ipv6" {
+		ipv6 = true
+	}
 
 	// add all of the config nodes as desired nodes
 	for _, configNode := range configNodes {
@@ -173,6 +186,7 @@ func nodesToCreate(cfg *config.Cluster, clusterName string) []nodeSpec {
 			ExtraMounts:      configNode.ExtraMounts,
 			APIServerAddress: apiServerAddress,
 			APIServerPort:    apiServerPort,
+			IPv6:             ipv6,
 		})
 	}
 
@@ -186,6 +200,7 @@ func nodesToCreate(cfg *config.Cluster, clusterName string) []nodeSpec {
 			ExtraMounts:      []cri.Mount{},
 			APIServerAddress: cfg.Networking.APIServerAddress,
 			APIServerPort:    cfg.Networking.APIServerPort,
+			IPv6:             ipv6,
 		})
 	}
 
--- a/pkg/cluster/internal/kubeadm/config.go
+++ b/pkg/cluster/internal/kubeadm/config.go
@@ -19,7 +19,9 @@ package kubeadm
 import (
 	"bytes"
 	"strings"
-	"text/template"
+  "text/template"
+  log "github.com/sirupsen/logrus"
+
 
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/util/version"
@@ -440,6 +442,7 @@ metadata:
 // Config returns a kubeadm config generated from config data, in particular
 // the kubernetes version
 func Config(data ConfigData) (config string, err error) {
+  log.Debugf("Configuration Input data: %v", data)
 	ver, err := version.ParseGeneric(data.KubernetesVersion)
 	if err != nil {
 		return "", err
@@ -468,6 +471,7 @@ func Config(data ConfigData) (config string, err error) {
 	err = t.Execute(&buff, data)
 	if err != nil {
 		return "", errors.Wrap(err, "error executing config template")
-	}
+  }
+  log.Debugf("Configuration generated:\n %v", buff.String())
 	return buff.String(), nil
 }
--- a/pkg/cluster/internal/kubeadm/config.go
+++ b/pkg/cluster/internal/kubeadm/config.go
@@ -45,6 +45,10 @@ type ConfigData struct {
 	Token string
 	// The subnet used for pods
 	PodSubnet string
+	// The subnet used for services
+	ServiceSubnet string
+	// IPv4 values take precedence over IPv6 by default, if true set IPv6 default values
+	IPv6 bool
 	// DerivedConfigData is populated by Derive()
 	// These auto-generated fields are available to Config templates,
 	// but not meant to be set by hand
@@ -104,9 +108,14 @@ apiServerExtraVolumes:
 # on docker for mac we have to expose the api server via port forward,
 # so we need to ensure the cert is valid for localhost so we can talk
 # to the cluster after rewriting the kubeconfig to point to localhost
-apiServerCertSANs: [localhost, {{.APIServerAddress}}]
+apiServerCertSANs: [localhost, "{{.APIServerAddress}}"]
 kubeletConfiguration:
   baseConfig:
+    # configure ipv6 addresses in IPv6 mode
+    {{ if .IPv6 -}}
+    address: "::"
+    healthzBindAddress: "::"
+    {{- end }}
     # disable disk resource management by default
     # kubelet will see the host disk that the inner container runtime
     # is ultimately backed by and attempt to recover disk space.
@@ -151,6 +160,9 @@ metadata:
 kubernetesVersion: {{.KubernetesVersion}}
 clusterName: "{{.ClusterName}}"
 controlPlaneEndpoint: {{ .ControlPlaneEndpoint }}
+networking:
+  podSubnet: "{{ .PodSubnet }}"
+  serviceSubnet: "{{ .ServiceSubnet }}"
 # we need nsswitch.conf so we use /etc/hosts
 # https://github.com/kubernetes/kubernetes/issues/69195
 apiServerExtraVolumes:
@@ -162,7 +174,7 @@ apiServerExtraVolumes:
 # on docker for mac we have to expose the api server via port forward,
 # so we need to ensure the cert is valid for localhost so we can talk
 # to the cluster after rewriting the kubeconfig to point to localhost
-apiServerCertSANs: [localhost, {{.APIServerAddress}}]
+apiServerCertSANs: [localhost, "{{.APIServerAddress}}"]
 controllerManagerExtraArgs:
   enable-hostpath-provisioner: "true"
 networking:
@@ -210,6 +222,11 @@ apiVersion: kubelet.config.k8s.io/v1beta1
 kind: KubeletConfiguration
 metadata:
   name: config
+# configure ipv6 addresses in IPv6 mode
+{{ if .IPv6 -}}
+address: "::"
+healthzBindAddress: "::"
+{{- end }}
 # disable disk resource management by default
 # kubelet will see the host disk that the inner container runtime
 # is ultimately backed by and attempt to recover disk space. we don't want that.
@@ -239,12 +256,24 @@ controlPlaneEndpoint: {{ .ControlPlaneEndpoint }}
 # so we need to ensure the cert is valid for localhost so we can talk
 # to the cluster after rewriting the kubeconfig to point to localhost
 apiServer:
-  certSANs: [localhost, {{.APIServerAddress}}]
+  certSANs: [localhost, "{{.APIServerAddress}}"]
 controllerManager:
   extraArgs:
     enable-hostpath-provisioner: "true"
+    # configure ipv6 default addresses for IPv6 clusters
+    {{ if .IPv6 -}}
+    bind-address: "::"
+    {{- end }}
+scheduler:
+  extraArgs:
+    # configure ipv6 default addresses for IPv6 clusters
+    {{ if .IPv6 -}}
+    address: "::"
+    bind-address: "::1"
+    {{- end }}
 networking:
   podSubnet: "{{ .PodSubnet }}"
+  serviceSubnet: "{{ .ServiceSubnet }}"
 ---
 apiVersion: kubeadm.k8s.io/v1beta1
 kind: InitConfiguration
@@ -290,6 +319,11 @@ apiVersion: kubelet.config.k8s.io/v1beta1
 kind: KubeletConfiguration
 metadata:
   name: config
+# configure ipv6 addresses in IPv6 mode
+{{ if .IPv6 -}}
+address: "::"
+healthzBindAddress: "::"
+{{- end }}
 # disable disk resource management by default
 # kubelet will see the host disk that the inner container runtime
 # is ultimately backed by and attempt to recover disk space. we don't want that.
@@ -319,12 +353,24 @@ controlPlaneEndpoint: {{ .ControlPlaneEndpoint }}
 # so we need to ensure the cert is valid for localhost so we can talk
 # to the cluster after rewriting the kubeconfig to point to localhost
 apiServer:
-  certSANs: [localhost, {{.APIServerAddress}}]
+  certSANs: [localhost, "{{.APIServerAddress}}"]
 controllerManager:
   extraArgs:
     enable-hostpath-provisioner: "true"
+    # configure ipv6 default addresses for IPv6 clusters
+    {{ if .IPv6 -}}
+    bind-address: "::"
+    {{- end }}
+scheduler:
+  extraArgs:
+    # configure ipv6 default addresses for IPv6 clusters
+    {{ if .IPv6 -}}
+    address: "::"
+    bind-address: "::1"
+    {{- end }}
 networking:
   podSubnet: "{{ .PodSubnet }}"
+  serviceSubnet: "{{ .ServiceSubnet }}"
 ---
 apiVersion: kubeadm.k8s.io/v1beta2
 kind: InitConfiguration
@@ -370,6 +416,11 @@ apiVersion: kubelet.config.k8s.io/v1beta1
 kind: KubeletConfiguration
 metadata:
   name: config
+# configure ipv6 addresses in IPv6 mode
+{{ if .IPv6 -}}
+address: "::"
+healthzBindAddress: "::"
+{{- end }}
 # disable disk resource management by default
 # kubelet will see the host disk that the inner container runtime
 # is ultimately backed by and attempt to recover disk space. we don't want that.
--- a/pkg/cluster/nodes/create.go
+++ b/pkg/cluster/nodes/create.go
@@ -59,11 +59,12 @@ func CreateControlPlaneNode(name, image, clusterLabel, listenAddress string, por
 		port = p
 	}
 
+	portMapping := net.JoinHostPort(listenAddress, fmt.Sprintf("%d", port)) + fmt.Sprintf(":%d", kubeadm.APIServerPort)
 	node, err = createNode(
 		name, image, clusterLabel, constants.ControlPlaneNodeRoleValue, mounts,
 		// publish selected port for the API server
 		"--expose", fmt.Sprintf("%d", port),
-		"-p", fmt.Sprintf("%s:%d:%d", listenAddress, port, kubeadm.APIServerPort),
+		"-p", portMapping,
 	)
 	if err != nil {
 		return node, err
@@ -90,11 +91,12 @@ func CreateExternalLoadBalancerNode(name, image, clusterLabel, listenAddress str
 		port = p
 	}
 
+	portMapping := net.JoinHostPort(listenAddress, fmt.Sprintf("%d", port)) + fmt.Sprintf(":%d", loadbalancer.ControlPlanePort)
 	node, err = createNode(name, image, clusterLabel, constants.ExternalLoadBalancerNodeRoleValue,
 		nil,
 		// publish selected port for the control plane
 		"--expose", fmt.Sprintf("%d", port),
-		"-p", fmt.Sprintf("%s:%d:%d", listenAddress, port, loadbalancer.ControlPlanePort),
+		"-p", portMapping,
 	)
 	if err != nil {
 		return node, err
--- a/pkg/cluster/nodes/node.go
+++ b/pkg/cluster/nodes/node.go
@@ -62,7 +62,8 @@ func (n *Node) Command(command string, args ...string) exec.Cmd {
 type nodeCache struct {
 	mu                sync.RWMutex
 	kubernetesVersion string
-	ip                string
+	ipv4              string
+	ipv6              string
 	ports             map[int32]int32
 	role              string
 }
@@ -79,10 +80,10 @@ func (cache *nodeCache) KubeVersion() string {
 	return cache.kubernetesVersion
 }
 
-func (cache *nodeCache) IP() string {
+func (cache *nodeCache) IP() (string, string) {
 	cache.mu.RLock()
 	defer cache.mu.RUnlock()
-	return cache.ip
+	return cache.ipv4, cache.ipv6
 }
 
 func (cache *nodeCache) HostPort(p int32) (int32, bool) {
@@ -147,25 +148,30 @@ func (n *Node) KubeVersion() (version string, err error) {
 }
 
 // IP returns the IP address of the node
-func (n *Node) IP() (ip string, err error) {
+func (n *Node) IP() (ipv4 string, ipv6 string, err error) {
 	// use the cached version first
-	cachedIP := n.cache.IP()
-	if cachedIP != "" {
-		return cachedIP, nil
+	cachedIPv4, cachedIPv6 := n.cache.IP()
+	// TODO: this assumes there are always ipv4 and ipv6 cached addresses
+	if cachedIPv4 != "" && cachedIPv6 != "" {
+		return cachedIPv4, cachedIPv6, nil
 	}
 	// retrive the IP address of the node using docker inspect
-	lines, err := docker.Inspect(n.name, "{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}")
+	lines, err := docker.Inspect(n.name, "{{range .NetworkSettings.Networks}}{{.IPAddress}},{{.GlobalIPv6Address}}{{end}}")
 	if err != nil {
-		return "", errors.Wrap(err, "failed to get file")
+		return "", "", errors.Wrap(err, "failed to get container details")
 	}
 	if len(lines) != 1 {
-		return "", errors.Errorf("file should only be one line, got %d lines", len(lines))
+		return "", "", errors.Errorf("file should only be one line, got %d lines", len(lines))
 	}
-	ip = lines[0]
+	ips := strings.Split(lines[0], ",")
+	if len(ips) != 2 {
+		return "", "", errors.Errorf("container addresses should have 2 values, got %d values", len(ips))
+	}
 	n.cache.set(func(cache *nodeCache) {
-		cache.ip = ip
+		cache.ipv4 = ips[0]
+		cache.ipv6 = ips[1]
 	})
-	return ip, nil
+	return ips[0], ips[1], nil
 }
 
 // Ports returns a specific port mapping for the node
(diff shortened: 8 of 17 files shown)


BenTheElder:
/ok-to-test
I'd like to make this a config field in cluster wide networking re: https://github.com/kubernetes-sigs/kind/issues/340

images/base/Dockerfile
```diff
@@ -81,10 +81,10 @@ RUN curl -fsSL "https://download.docker.com/linux/$(. /etc/os-release; echo "$ID
 
 # Install CNI binaries to /opt/cni/bin
 # TODO(bentheelder): doc why / what here
-ARG CNI_VERSION="0.6.0"
-ARG CNI_TARBALL="cni-plugins-${ARCH}-v${CNI_VERSION}.tgz"
-ARG CNI_BASE_URL="https://storage.googleapis.com/kubernetes-release/network-plugins/"
-ARG CNI_URL="${CNI_BASE_URL}${CNI_TARBALL}"
+ARG CNI_VERSION="v0.7.4"
```
neolit123: probably best to bump this only after the official packages are bumped to this version:
https://github.com/kubernetes/release/blob/4de61a4d8b5fdf2a9a5aac4013afcbc94b54b2ad/debian/xenial/kubernetes-cni/debian/rules

is it a requirement for ipv6 to work?

pkg/cluster/nodes/node.go
```diff
@@ -218,21 +219,28 @@ func (n *Node) KubeVersion() (version string, err error) {
...
 
 // IP returns the IP address of the node
-func (n *Node) IP() (ip string, err error) {
+func (n *Node) IP(isIPv6 bool) (ip string, err error) {
+	// Check the IP version
+	cachedIP := &n.nodeCache.ip
+	IPAddress := "IPAddress"
+	if isIPv6 {
+		cachedIP = &n.nodeCache.ipv6
+		IPAddress = "GlobalIPv6Address"
```
neolit123: do we need to add this separate string?

pkg/cluster/nodes/node.go
```diff
@@ -71,6 +71,7 @@ func (n *Node) Command(command string, args ...string) exec.Cmd {
 type nodeCache struct {
 	kubernetesVersion string
 	ip                string
+	ipv6              string
```
neolit123: so https://golang.org/pkg/net/#IP is technically already an abstraction between v4 and v6 can we use that?

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -74,11 +76,14 @@ clusterName: "{{.ClusterName}}"
 bootstrapTokens:
 - token: "{{ .Token }}"
 {{ if .ControlPlaneEndpoint -}}
-controlPlaneEndpoint: {{ .ControlPlaneEndpoint }}
+controlPlaneEndpoint: "{{ .ControlPlaneEndpoint }}"
 {{- end }}
 # we use a well know port for making the API server discoverable inside docker network. 
 # from the host machine such port will be accessible via a random local port instead.
```
neolit123: this comment is about the port.

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -74,11 +76,14 @@ clusterName: "{{.ClusterName}}"
...
 api:
+  {{ if .APIAdvertiseAddress -}}
+  advertiseAddress: "{{ .APIAdvertiseAddress }}"
```
neolit123: why do we need to add the advertise address as part of the this PR?
`controlPlaneEndpoint` should already handle that for both single CP and multi CP scenarios.

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -74,11 +76,14 @@ clusterName: "{{.ClusterName}}"
 bootstrapTokens:
 - token: "{{ .Token }}"
 {{ if .ControlPlaneEndpoint -}}
-controlPlaneEndpoint: {{ .ControlPlaneEndpoint }}
+controlPlaneEndpoint: "{{ .ControlPlaneEndpoint }}"
```
neolit123: was the quoting needed here?

pkg/cluster/internal/create/kubeadm-config.go
```diff
@@ -68,6 +68,12 @@ func runKubeadmConfig(ec *execContext, configNode *NodeReplica) error {
...
 
+	// get the address the API Server will advertise to other members of the cluster
+	apiAdvertiseAddress, err := node.IP(ec.Context.IPv6)
+	if err != nil {
+		return errors.Wrap(err, "failed to get IP for bootsrap node")
+	}
+
 	// get the control plane endpoint, in case the cluster has an external load balancer in
 	// front of the control-plane nodes
 	controlPlaneEndpoint, err := getControlPlaneEndpoint(ec)
```
neolit123: `getControlPlaneEndpoint` might need modifications based on IPv6.

have you tested an HA setup?

images/base/Dockerfile
aojea: it's an option, you can read more about it here https://github.com/containernetworking/cni/issues/531

I think that it can be solved within `kind` too

pkg/cluster/nodes/node.go
aojea: yeah, this is the snippet of a `docker inspect`output
```   
                   "IPAddress": "172.17.0.2",
                    "IPPrefixLen": 16,
                    "IPv6Gateway": "2001:db8:1::1",
                    "GlobalIPv6Address": "2001:db8:1::242:ac11:2",
 ```

we need to use a different string depending on the protocol

File: pkg/cluster/nodes/node.go (hunk 3 above)
Comment (aojea) [reply pkg/cluster/nodes/node.go]: I was thinking that if we'll need to support dual stack it can be useful to have both fields.
Also, I think that will imply more changes in other parts of the code that are not relevant to this PR

File: pkg/cluster/internal/kubeadm/config.go (hunk 4 above)
Comment (aojea) [reply pkg/cluster/internal/kubeadm/config.go]: don't understand the comment, I had to quote the ControlPlaneEndpoint because it failed to create the JSON from the template using IPv6 addresses

File: pkg/cluster/internal/create/kubeadm-config.go (hunk 7 above)
Comment (aojea) [reply pkg/cluster/internal/create/kubeadm-config.go]: didn't test it, but will do.
I've already modified the function `func getControlPlaneEndpoint(` to support IPv6, check L135

File: pkg/cluster/internal/kubeadm/config.go (hunk 5 above)
Comment (aojea) [reply pkg/cluster/internal/kubeadm/config.go]: I had to add the `advertiseAddress ` because `etcd` , the `kube-apiserver` and other components were using the interface IPv4 address, also the `sevrer` field in `/etc/kubernetes/kubelet.conf` was using the IPv4 address

File: pkg/cluster/internal/kubeadm/config.go (hunk 6 above)
Comment (aojea) [reply pkg/cluster/internal/kubeadm/config.go]: I had to quote the ControlPlaneEndpoint because it failed to create the JSON from the template using IPv6 addresses

**Comment by aojea** [comment]
@BenTheElder it's failing to verify the docker image `-var _imagesBaseDockerfile = ` but that header's file says
```
// Code generated by go-bindata.
// sources:
// ../../../../images/base/Dockerfile
// ../../../../images/base/clean-install
// ../../../../images/base/entrypoint/main.go
// DO NOT EDIT!
```

Do I need to recreate the file? how can I do it?

File: pkg/cluster/internal/kubeadm/config.go (hunk 4 above)
Comment (neolit123) [reply pkg/cluster/internal/kubeadm/config.go]: i wanted to say that the comment `# we use a well know port for making...` is now above advertiseAddress, while it applies to the port bellow.

File: pkg/cluster/internal/kubeadm/config.go (hunk 5 above)
Comment (neolit123) [reply pkg/cluster/internal/kubeadm/config.go]: hmm, not sure how should we approach this problem.
cc @fabriziopandini

File: pkg/cluster/internal/kubeadm/config.go (hunk 6 above)
Comment (neolit123) [reply pkg/cluster/internal/kubeadm/config.go]: ok, understood.

**Comment by neolit123** [comment]
> Do I need to recreate the file? how can I do it?

`./hack/update-generated.sh` ?

File: pkg/cluster/internal/kubeadm/config.go (hunk 4 above)
Comment (aojea) [reply pkg/cluster/internal/kubeadm/config.go]: I think that the comment is ok, it's above the `api` section not above the port field, to explain that the `apiendpoint` uses a well know port.

File: pkg/cluster/internal/kubeadm/config.go (hunk 5 above)
Comment (aojea) [reply pkg/cluster/internal/kubeadm/config.go]: I have a similar issue when adding multiple control nodes with IPv6, first control node register successfully but the second one falls back to IPb4 and fails

```
EBU[00:13:45] Running: /usr/bin/docker [docker exec --privileged -t kind-control-plane2 kubeadm join [2001:db8:1::242:ac11:2]:6443 --experimental-control-plane --token abcdef.0123456789abcdef --discovery-token-unsafe-skip-ca-verification --ignore-preflight-errors=all --v=6]
⢎⡰ [control-plane2] Joining control-plane node to Kubernetes ☸
DEBU[00:15:06] I0225 23:13:46.054405    1008 join.go:299] [join] found NodeName empty; using OS hostname as NodeName
I0225 23:13:46.054436    1008 join.go:303] [join] found advertiseAddress empty; using default interface's IP address as advertiseAddress
I0225 23:13:46.054653    1008 interface.go:384] Looking for default routes with IPv4 addresses
I0225 23:13:46.054665    1008 interface.go:389] Default route transits interface "eth0"
I0225 23:13:46.054758    1008 interface.go:196] Interface eth0 is up
I0225 23:13:46.054792    1008 interface.go:244] Interface "eth0" has 3 addresses :[172.17.0.4/16 2001:db8:1::242:ac11:4/64 fe80::42:acff:fe11:4/64].
I0225 23:13:46.054808    1008 interface.go:211] Checking addr  172.17.0.4/16.
I0225 23:13:46.054814    1008 interface.go:218] IP found 172.17.0.4
I0225 23:13:46.054820    1008 interface.go:250] Found valid IPv4 address 172.17.0.4 for interface "eth0".
I0225 23:13:46.054827    1008 interface.go:395] Found active IP 172.17.0.4
[preflight] Running pre-flight checks
I0225 23:13:46.054858    1008 join.go:328] [preflight] Running general checks
```

File: pkg/cluster/internal/create/kubeadm-config.go (hunk 7 above)
Comment (aojea) [reply pkg/cluster/internal/create/kubeadm-config.go]: @ncdc results of an HA setup test

```
 # docker ps
CONTAINER ID        IMAGE                  COMMAND                  CREATED             STATUS              PORTS                                NAMES
7b37bcbcf3b6        kindest/node:v1.13.3   "/usr/local/bin/entr…"   11 minutes ago      Up 11 minutes                                            kind-worker2
89fa42fd404e        kindest/node:v1.13.3   "/usr/local/bin/entr…"   12 minutes ago      Up 12 minutes                                            kind-worker1
0d0caea990ca        kindest/node:v1.13.3   "/usr/local/bin/entr…"   12 minutes ago      Up 12 minutes       33453/tcp, 0.0.0.0:33453->6443/tcp   kind-control-plane3
422bd7c52d53        kindest/node:v1.13.3   "/usr/local/bin/entr…"   13 minutes ago      Up 12 minutes       34151/tcp, 0.0.0.0:34151->6443/tcp   kind-control-plane2
780f8cde9d6d        kindest/node:v1.13.3   "/usr/local/bin/entr…"   13 minutes ago      Up 13 minutes       37927/tcp, 0.0.0.0:37927->6443/tcp   kind-control-plane1
3f2e867b33b6        kindest/node:v1.13.3   "/usr/local/bin/entr…"   13 minutes ago      Up 13 minutes       40189/tcp, 0.0.0.0:40189->6443/tcp   kind-lb
```
```
# kubectl get endpoints --all-namespaces
NAMESPACE     NAME                      ENDPOINTS                                                                                   AGE
default       kubernetes                [2001:db8:1::242:ac11:3]:6443,[2001:db8:1::242:ac11:4]:6443,[2001:db8:1::242:ac11:5]:6443   11m
kube-system   kube-controller-manager   <none>                                                                                      10m
kube-system   kube-dns                  10.32.0.2:53,10.32.0.3:53,10.32.0.2:53 + 1 more...                                          10m
kube-system   kube-scheduler            <none>                                                                                      10m
```

with current parameters:
`./kind create cluster --ipv6 --config config-ipv6.yaml`

and config-ipv6.yaml

```
# this config file contains all config fields with comments
kind: Config
# 3 control plane node and 3 workers
nodes:
# the control plane node config
- role: control-plane
  replicas: 3
  # patch the generated kubeadm config with some extra settings
  kubeadmConfigPatches:
  - |
    kind: ClusterConfiguration
    metadata:
      name: config
    networking:
      serviceSubnet: "fd00:1234::/112"
# the three workers
- role: worker
  # replicas specifes the number of nodes to create with this configuration
  replicas: 2
- role: external-load-balancer
```

@BenTheElder I'd like to add some e2e scenarios, could you help me or guide me?
Should those tests be part of this PR?

File: pkg/cluster/internal/create/kubeadm-join.go [hunk 8]
```diff
@@ -142,6 +148,8 @@ func runKubeadmJoinControlPlane(ec *execContext, configNode *NodeReplica) error
 		// preflight errors are expected, in particular for swap being enabled
 		// TODO(bentheelder): limit the set of acceptable errors
 		"--ignore-preflight-errors=all",
+		// use advertise address to avoid issues with Ipv6
+		"--apiserver-advertise-address", advertisedAddress,
```
aojea: Seems that this option is not available in v1.11 and this causes the gate job failure.
Appreciate any help/idea to overcome this, are there any alternatives in `kubeadm` version 1.11 or do we simply can skip this option and IPv6 in v1.11? @neolit123

neolit123: 1.11 is going outside of the support skew with the release of 1.14 (late march), so +1.

aojea: :thinking: seems that v1.11 is going away soon https://github.com/kubernetes/test-infra/pull/11476/files

pkg/cluster/internal/create/kubeadm-config.go
neolit123: @aojea 
hm, given weave supposedly does not support ipv6, how are you testing this?

> ./kind create cluster --ipv6

we need to remove the `--ipv6` flag from the CLI, should be config only.
the CLI is only for the basic functionality.

> @BenTheElder I'd like to add some e2e scenarios, could you help me or guide me?
Should those tests be part of this PR?

let's get the support in kind first, and then we can think about e2e.
ideally things like ipv6 should be owned by sig-network, but i can think we might be able to add this in the sig-cluster-lifecycle dashboard too.

i will try your PR later today myself, hopefully.

aojea: @neolit123 seems that the cluster control plane is ipv6 abut the data plane is ipv4, let me do more tests changing to the cni bridge plugin and come back to you

BenTheElder:
[intend to start doing some more tests around this ~tomorrow]

Thanks again for working on this, apologies for the review latency on my end. I've been looking deeper into the CNI options FWIW, intend to do some more testing with calico 👍

aojea:
@BenTheElder after playing with different CNI plugins with unsatisfactory results  I explored other options.
I've found that @leblancd had a[ k8s cluster working with IPv6 ](https://github.com/leblancd/kube-v6-test/tree/master)using static routes and the standard CNI `bridge` plugin, so I tried to add the logic to `kind` but I've discarded that because I've found it will be very hard to maintain and to scale, specially for having multiple clusters in the same host.
I ended creating a [Poor Man CNI plugin](https://github.com/aojea/kindnet) that I called `kindnet` leveraging the standard CNI plugins, it's very simple, it has a daemon running in each node that polls the API to get the podCIDR assigned to each node and add a static route.
I have to rebase the code and still need to do more testing but I wanted to start to clean up things and get the code ready to merge, can you let me know your opinion about `kindnet`?
What will be the best option to select IPv6 or IPv4? an option in the config file?

aojea:
jobs are failing because of 

```W0304 12:03:42.352] Error: failed to create cluster: failed to generate kubeadm config content: failed to find an object with kubeadm.k8s.io_v1alpha3_ClusterConfiguration|config to apply the patch```

it seems related to this https://github.com/kubernetes-sigs/kind/issues/126 :thinking:

File: pkg/cluster/internal/create/kubeadm-config.go (hunk 7 above)
Comment (aojea) [reply pkg/cluster/internal/create/kubeadm-config.go]: @neolit123 I can access IPv6 services with the new [CNI plugin I've created](https://github.com/aojea/kindnet) :)

To test it just update the `node-image` with the new `kind` binary built from this PR
`kind build node-image --base-image kindest/base:latest --type bazel`
And spawn a cluster with the following config file 
```
# this config file contains all config fields with comments
kind: Config
# 3 control plane node and 3 workers
nodes:
# the control plane node config
- role: control-plane
  replicas: 1
  # patch the generated kubeadm config with some extra settings
  kubeadmConfigPatches:
  - |
    kind: ClusterConfiguration
    metadata:
      name: config
    networking:
      podSubnet: "fd00:100::/64"
      serviceSubnet: "fd00:1234::/112"
# the three workers
- role: worker
  # replicas specifes the number of nodes to create with this configuration
  replicas: 2
```

`kind create cluster --ipv6 --config config-ipv6-ci.yaml --loglevel debug --image kindest/node:latest`

and you can have an IPv6 k8s cluster:

```
 kubectl get nodes -o wide
NAME                 STATUS    ROLES     AGE       VERSION          INTERNAL-IP              EXTERNAL-IP   OS-IMAGE             KERNEL-VERSION                CONTAINER-RUNTIME
kind-control-plane   Ready     master    3m9s      v1.14.0-beta.1   2001:db8:1::242:ac11:3   <none>        Ubuntu 18.04.1 LTS   4.12.14-lp150.12.45-default   docker://18.6.3
kind-worker1         Ready     <none>    2m31s     v1.14.0-beta.1   2001:db8:1::242:ac11:4   <none>        Ubuntu 18.04.1 LTS   4.12.14-lp150.12.45-default   docker://18.6.3
kind-worker2         Ready     <none>    2m16s     v1.14.0-beta.1   2001:db8:1::242:ac11:5   <none>        Ubuntu 18.04.1 LTS   4.12.14-lp150.12.45-default   docker://18.6.3
```
with pod IPv6 connectivity and services
```
kubectl -n kube-system get ep -o wide
NAME                      ENDPOINTS                                                        AGE
kube-controller-manager   <none>                                                           4m34s
kube-dns                  [fd00:100::2]:53,[fd00:100::3]:53,[fd00:100::2]:53 + 3 more...   4m17s
kube-scheduler            <none>                                                           4m36s
```
The pods obtain IPv6 addresses:

```
 kubectl run -i --tty busybox12 --image=busybox --restart=Never -- sh
If you don't see a command prompt, try pressing enter.
/ # ip -6 a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qlen 1000
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
5: eth0@if6: <BROADCAST,MULTICAST,UP,LOWER_UP,M-DOWN> mtu 1500
    inet6 fd00:100::1:0:0:2/80 scope global
       valid_lft forever preferred_lft forever
    inet6 fe80::5c62:8aff:fe75:7ce0/64 scope link
       valid_lft forever preferred_lft forever
```
with pod connectivity between nodes

```
/ # ping fd00:100::2
PING fd00:100::2 (fd00:100::2): 56 data bytes
64 bytes from fd00:100::2: seq=0 ttl=62 time=0.181 ms
64 bytes from fd00:100::2: seq=1 ttl=62 time=0.151 ms
```

and pod to service connectivity
```
/ # nc -zv fd00:1234::a 53
fd00:1234::a ([fd00:1234::a]:53) open
```

Comment (neolit123) [reply pkg/cluster/internal/create/kubeadm-config.go]: @aojea nice.

>       podSubnet: "fd00:100::/64"
>      serviceSubnet: "fd00:1234::/112"

do we have to instruct the users to patch the config with IPv6 CIDRs if they want IPv6 support.

> kind create cluster --ipv6 

we might want to move this a flag in the kind Config.

Comment (aojea) [reply pkg/cluster/internal/create/kubeadm-config.go]: Where should be the best place to add the ipv6 flag?
What do you think to add it in the new `networking` section of the v1alpha3 API https://github.com/kubernetes-sigs/kind/issues/340 ?

Comment (neolit123) [reply pkg/cluster/internal/create/kubeadm-config.go]: > What do you think to add it in the new networking section of the v1alpha3 API

yes, makes sense.

and how about this?
> do we have to instruct the users to patch the config with IPv6 CIDRs if they want IPv6 support?

Comment (aojea) [reply pkg/cluster/internal/create/kubeadm-config.go]: @neolit123 that's a good question but I miss context here, but I think that maybe can be a topic to be discussed in `kubeadm`.
I could check that `kubeadm` sets a default `servicesubnet` if none is provided, should we do the same for IPv6? if that's the case we should add a flag to `kubeadm` then

Comment (neolit123) [reply pkg/cluster/internal/create/kubeadm-config.go]: > I could check that kubeadm sets a default servicesubnet if none is provided, should we do the same for IPv6? if that's the case we should add a flag to kubeadm then

here are kubeadm defaults for these CIDRs:
https://github.com/kubernetes/kubernetes/blob/b66e332d3c19ce78e00dd7c904fa29e8f6784ba0/cmd/kubeadm/app/apis/kubeadm/v1beta1/defaults.go#L31-L34

it not very likely that we will add a flag or a field in the kubeadm config for this.
IPv6 is explicit on the kubeadm side and up to the administrator to decide what stack to use. so kind as a kubeadm wrapper might have to handle this explicitly (as your patch is doing here).

at least that's how i see it for now.

**Comment by BenTheElder** [comment]
[sorry I've been trying to get #360 in knowing full well that we'll need to rebase PRs like this on it]

**Comment by BenTheElder** [comment]
Let's add ipv6 as an option in the new API, now that #360 is in it should be a lot more straightforward to implement #340.

Not sure about kindnet yet 😅 

> are the CI jobs using a different kubeadm version?

yes we run CI for kind with kubernetes back to 1.11 currently. 1.11, 1.12, 1.13, ...

File: hack/ci/e2e.sh [hunk 9]
```diff
@@ -103,6 +103,16 @@ apiVersion: kind.sigs.k8s.io/v1alpha2
...
 - role: control-plane
+  # patch the generated kubeadm config with some extra settings
+  kubeadmConfigPatches:
+  - |
+    apiVersion: kubeadm.k8s.io/v1alpha3
+    kind: ClusterConfiguration
+    metadata:
+      name: config
+    networking:
+      podSubnet: "10.192.0.0/16"
```
BenTheElder: general kind CI should work without a patch like this.

aojea: AFAIK podSubnet is required by Calico and Flannel

neolit123: yes, unlike weave, some CNIs need pod network CIDRs.

also e.g. Calico users that want IPv6 need to patch with an IPv6 pod-net CIDR...
it's messy.

aojea:
[`kind` multinode cluster with ipv6 automated test](https://circleci.com/gh/aojea/kindnet/92) 😄 

```
Summarizing 1 Failure:

[Fail] [sig-network] Networking [It] should provide Internet connection for containers [Feature:Networking-IPv6][Experimental] 
/workspace/anago-v1.13.0-rc.2.1+ddf47ac13c1a94/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/network/networking.go:56

Ran 24 of 1946 Specs in 1364.936 seconds
FAIL! -- 23 Passed | 1 Failed | 0 Pending | 1922 Skipped --- FAIL: TestE2E (1365.23s)
FAIL

```

aojea:
/hold 

I've tried to rebase but it's a nightmare to carry over the CLI options :sweat_smile: 
@BenTheElder I will wait for the new API to use the IPv6 and CNI fields in the cluster-wide configurations if that's acceptable

aojea:
WIP

pkg/cluster/config/types.go
```diff
@@ -82,3 +84,13 @@ const (
...
 )
+
+// Network defines networking cluster wide parameters
+type Network struct {
+	// range of IP addresses for the pod network
+	PodSubnet string
+	// range of IP address for service VIPs
+	ServiceSubnet string
+	// true if is IPv6
+	IPv6 bool
```
BenTheElder: only ipv6? what about dual stack? :thinking:

BenTheElder:
let's discuss this for 0.3 at the next meeting, 0.2 is about done and this will need to catch up with the networking settings introduced :sweat_smile:

pkg/cluster/config/types.go
aojea: yeah, I think that this needs better planning, let me try to create a google doc to discuss this implementation

aojea:
TODO:

- [ ] Add NAT64 and DNS64 container
- [ ] Add podSubnet and service subnet to kubeadm v1alpha2

pkg/cluster/config/v1alpha3/types.go
```diff
@@ -88,6 +88,8 @@ const (
 
 // Networking contains cluster wide network settings
 type Networking struct {
+	// IpFamily is the network cluster model, it can be ipv4, ipv6 or dual-stack
+	IpFamily ClusterIpFamily `json:"ipFamily,omitempty"`
```
BenTheElder: nit: IPFamily

pkg/cluster/config/v1alpha3/types.go
```diff
@@ -97,3 +99,12 @@ type Networking struct {
 	// Defaults to 127.0.0.1
 	APIServerAddress string `json:"apiServerAddress,omitempty"`
 }
+
+// Networking contains cluster wide network settings
+type ClusterIpFamily string
```
BenTheElder: ClusterIPFamily

pkg/cluster/internal/create/nodes.go
```diff
@@ -153,14 +153,35 @@ func fixupNode(node *nodes.Node) error {
 		}
 	}
 
+	if isIPv6 {
+		// Enable IPv6 in the inner docker daemon
+		if err := node.EnableDinDIPv6(); err != nil {
```
BenTheElder: `node.EnableIPv6()`

pkg/cluster/nodes/node.go
```diff
@@ -460,3 +466,16 @@ func getProxyDetails() proxyDetails {
...
 }
+
+// EnableDinDIPv6 enables IPv6 in the inner docker daemon
+func (n *Node) EnableDinDIPv6() error {
+	// configure Docker daemon to use ipv6
+	// we take an unused range otherwise the daemon refuse to start
+	err := n.WriteFile("/etc/docker/daemon.json",
+		"{\n\t\"ipv6\": true,\n\t\"fixed-cidr-v6\": \"fc00:db8:1::/64\"\n}")
+	if err != nil {
+		errors.Wrap(err, "failed to create docker daemon.json")
```
BenTheElder: this should have a return

pkg/cluster/nodes/create.go
```diff
@@ -135,6 +135,10 @@ func createNode(name, image, clusterLabel, role string, mounts []cri.Mount, extr
 		"--tmpfs", "/run", // systemd wants a writable /run
 		// some k8s things want /lib/modules
 		"-v", "/lib/modules:/lib/modules:ro",
+		// current calico config needs ipv6 enable
+		// run the node as dual stack to support all scenarios
+		"--sysctl", "net.ipv6.conf.all.disable_ipv6=0",
+		"--sysctl", "net.ipv6.conf.all.forwarding=1",
```
BenTheElder: this seems suspiciously likely to break things

aojea: yeah, I reverted it

pkg/cluster/nodes/node.go
aojea: fixed, thanks for catching it

pkg/cluster/internal/create/nodes.go
aojea: done

pkg/cluster/config/v1alpha3/types.go
aojea: done

pkg/cluster/config/v1alpha3/types.go
aojea: done

aojea:
Tested with calico and calico-ipv6 cni plugins https://circleci.com/workflow-run/1d88e726-b110-46e7-9c22-f770c5de9b84

aojea:
@BenTheElder I cleaned and reorganized the git history, and properly commented each commit.

aojea:
@BenTheElder @neolit123 I've taken the approach we decided during the last meeting and ran conformance tests in circle ci with ipv4 and ipv6 https://circleci.com/workflow-run/f9cabd29-173b-4198-b7d8-398d8c093b22

aojea:
/hold 
wait until #524 and #500 are merged

pkg/cluster/internal/create/nodes.go
```diff
@@ -154,6 +162,7 @@ func nodesToCreate(cfg *config.Cluster, clusterName string) []nodeSpec {
 		}
 	}
 	isHA := controlPlanes > 1
+	ipv6 := "ipv6" == cfg.Networking.IPFamily
```
aojea: Job is failing because of this

aojea:
/hold
loadbalancer is not working with ipv6

pkg/cluster/config/types.go
```diff
@@ -99,7 +101,22 @@ type Networking struct {
...
+// ClusterIPFamily defines cluster network IP family
+type ClusterIPFamily string
+
+const (
+	// IPv4Family sets ClusterIPFamily to ipv4
+	IPv4Family ClusterIPFamily = "ipv4"
+	// IPv6Family sets ClusterIPFamily to ipv6
+	IPv6Family ClusterIPFamily = "ipv6"
+	// DualStackFamily sets ClusterIPFamily to ipv4 and ipv6
+	DualStackFamily ClusterIPFamily = "ipv4-ipv6"
```
BenTheElder: Do we support this yet? If not we should drop this and add it later.

pkg/cluster/internal/create/nodes.go
BenTheElder: Which one? Is this staticcheck? This code looks legitimate to me

pkg/cluster/nodes/node.go
```diff
@@ -275,3 +281,32 @@ func getProxyDetails() proxyDetails {
...
+	if err != nil {
+		return errors.Wrap(err, "failed to enable ipv6 forwarding")
+	}
+	// enable ipv6 kubelet
+	_, nodeIPv6, err := n.IP()
+	if err != nil {
+		return errors.Wrap(err, "failed to get IP for node")
+	}
+
+	kubeletExtraConfig := fmt.Sprintf("KUBELET_EXTRA_ARGS=\"--fail-swap-on=false --node-ip=%s\"", nodeIPv6)
```
BenTheElder: can we set node-ip in kuebadm config?

BenTheElder: I guess not right now since we'd have to refactor to make it per-node, but that could be arranged.

I ran into some issues trying to get the bootstrap token stuff moved into config, and `kubeadm join` doesn't allow mixing that flag and config. cc @neolit123

neolit123: xref https://godoc.org/k8s.io/kubelet/config/v1beta1#KubeletConfiguration

not sure if KubeletConfiguration Address is the same as node-ip.
the docs are confusing.

as far as FailSwapOn it's possible from KubeletConfiguration, but i think we had a bug in kubeadm related to that.

`--fail-swap-on=false --node-ip` needs testing

neolit123: EDIT: but yeah, each node has unique IP so the flag override is best at this point.

BenTheElder: yeah but we can't do it via kubeadm, so now we have to hope that new image never need to add another flag to this file since we're overwriting it at runtime.

BenTheElder: checking if we can write to a second file instead

BenTheElder: no dice, we only read this file in the dropin, and if we set `KUBELET_EXTRA_ARGS` again on another line it will redefine it wholesale. we could instead edit this line but that's a little more complex.

BenTheElder: we can clean this up in a follow-up

pkg/cluster/config/types.go
```diff
@@ -99,7 +101,20 @@ type Networking struct {
 	// PodSubnet is the CIDR used for pod IPs
 	// kind will select a default if unspecified
 	PodSubnet string
+	// ServiceSubnet is the CIDR used for services VIPs
+	// kind will select a default if unspecified for IPv6
```
BenTheElder:
```suggestion
	// kind will select a default if unspecified
```

we're always selecting this. not a big deal though

pkg/cluster/config/default.go
```diff
@@ -41,14 +41,30 @@ func SetDefaults_Cluster(obj *Cluster) {
...
-	// TODO(bentheelder): this defaulting will need to be ipv6 aware as well
-	if obj.Networking.APIServerAddress == "" {
+	if obj.Networking.IPFamily == "" {
+		obj.Networking.IPFamily = "ipv4"
+	}
+	// default to listening on 127.0.0.1:randomPort on ipv4
+	// docker has limitations to expose ports on IPv6 addresses
+	// TODO(bentheelder): restrict default for ipv6 clusters
+	if obj.Networking.APIServerAddress == "" && obj.Networking.IPFamily == "ipv4" {
 		obj.Networking.APIServerAddress = "127.0.0.1"
```
BenTheElder: we should default ipv6 to localhost as well?

BenTheElder:
I've got this working and passing tests locally, but I needed an additional tiny patch (+14/-8) to fixup a few more ip:port locations and default the advertise address.

pkg/cluster/internal/create/nodes.go
aojea: It was because it doesn´t allow Yoda conditions 👽  :) I´ve alredy fixed it

pkg/cluster/config/default.go
aojea: from a UX point of view, I suggest to keep it this way at least during some time.
It took me a lot of time today to realize that I can´t use localhost from the host to reach the endpoints, you have to use [::1] instead 😅 
This way you can reach the API using an IPv4 or an IPv6 socket this way, and the main goal now is to test the kubernetes IPv6 support, I think that we are good with the TODO

BenTheElder:
https://github.com/kubernetes-sigs/kind/compare/master...BenTheElder:pr348?expand=1 has the fixes

pkg/cluster/config/default.go
BenTheElder: if we don't set this it will bind to 0.0.0.0 which is insecure

BenTheElder:
specifically https://github.com/kubernetes-sigs/kind/commit/a5e9d4d2622c447faa2502d30580e0174a13cbe5

with that after enabling ipv6 on my local docker I can run `FOCUS="Networking.Granular.Checks" SKIP='\[Slow\]|\[Serial\]|\[Flaky\]|\[Feature:.*\]' IP_FAMILY=ipv6 $HOME/go/src/sigs.k8s.io/kind/hack/ci/e2e.sh` from a local kubernetes master checkout. all 4 tests passed

pkg/cluster/config/default.go
aojea: yeah, I always forget, I added your fix to the PR

BenTheElder: thanks :-)

BenTheElder:
sanity checking again locally with the latest changes. so far so good! exciting stuff :-)

pkg/cluster/nodes/node.go
BenTheElder: this is also only an issue if we _do_ need to add another flag to this file for some reason, in the future we can instead _not_ do that and bite the bullet on generating a per-node kubeadm config file. we just need to figure out the bootstrap token part in particular :confused:

BenTheElder:
passed locally with both ipv4 and ipv6

BenTheElder:
after some more testing https://github.com/kubernetes-sigs/kind/pull/538

BenTheElder:
https://github.com/kubernetes/test-infra/pull/12669 hopefully is the first step towards being able to run this in CI

BenTheElder:
so the ip-masq-agent does not work properly with ipv6
```
$ kubectl -n=kube-system logs ip-masq-agent-8586w
E0515 23:02:37.493534       1 ip-masq-agent.go:146] error syncing masquerade rules: exit status 2 (iptables-restore v1.6.0: invalid mask `64' specified
Error occurred at line: 4
Try `iptables-restore -h' or 'iptables-restore --help' for more information.
)
```
we could resume using `"ipMasq": true` on the CNI bridge instead.

BenTheElder:
So I think everything here is fine in IPv4 mode, and we can follow up on more IPv6 improvements in more PRs instead of making this one even more massive :^)

aojea:
Automated testing working in circle ci https://circleci.com/gh/aojea/kubernetes/112

pkg/cluster/config/default.go
aojea: @BenTheElder this requires having localhost as an entry for `::1` , otherwise it fails
I hit this in the CircleCI and seems that's missing in old linux versions distributions.

```
# IP-Address  Full-Qualified-Hostname  Short-Hostname
#

127.0.0.1       localhost

# special IPv6 addresses
::1             localhost ipv6-localhost ipv6-loopback
```

aojea:
> so the ip-masq-agent does not work properly with ipv6
> 
> ```
> $ kubectl -n=kube-system logs ip-masq-agent-8586w
> E0515 23:02:37.493534       1 ip-masq-agent.go:146] error syncing masquerade rules: exit status 2 (iptables-restore v1.6.0: invalid mask `64' specified
> Error occurred at line: 4
> Try `iptables-restore -h' or 'iptables-restore --help' for more information.
> )
> ```
> 
> we could resume using `"ipMasq": true` on the CNI bridge instead.

seems easy to fix in the ip-masq-agent https://github.com/kubernetes-incubator/ip-masq-agent/issues/43

pkg/cluster/config/default.go
BenTheElder: We just need to write the KUBECONFIG appropriately

aojea: I have doubts, I was thinking the same at the beginning,
 but I think that localhost resolve to `::1`  too, I'm leaning keep it as is.

BenTheElder: in KUBECONFIG we can write the ip instead of localhost

aojea:
@BenTheElder this is working to me with the new kindnet 0.2.0
https://circleci.com/workflow-run/7422f791-7c01-4177-850f-418c874d9c7f

BenTheElder:
[for anyone following along]
we've been in close contact on this, even if there's not much activity on this PR, the next step is new images with something like #545 to handle ipv6 masquerade. then we will leverage that here and fixup kind's config and internals to handle ipv6 addresses properly in this PR. then ipv6 will ~work

switching to kindnetd in 0.3 got us much closer, 0.4 is slated to land this after we get #545 in and robust.

BenTheElder:
that test won't pass until it is updated to use the correct node pool over in test-infra

BenTheElder:
we also have to add the appropriate modprobes to the ipv6 setup

BenTheElder:
this [actually ran](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_kind/348/pull-kind-conformance-parallel-ipv6/1136793268353241088/) but as expected some tests fail.

aojea:
rebased over the kindnet 0.2.0 branch and squashed, we need the new kindest/kindnetd:0.2.0 images in order to test it

BenTheElder:
this needs a rebase now that #632 / #616 is in, but that puts us one step closer :sweat_smile:

aojea:
/hold

> 09:55:14.519] Error: failed to create cluster: failed to generate kubeadm config content: rawResources failed to read Resources: YAML file [resource-0.yaml] encounters a format error.
W0621 09:55:14.520] error converting YAML to JSON: yaml: line 7:

hack/ci/e2e.sh
```diff
@@ -130,6 +133,34 @@ run_tests() {
     KUBECONFIG="$(kind get kubeconfig-path)"
     export KUBECONFIG
 
+    if [[ "${IP_FAMILY:-ipv4}" == "ipv6" ]]; then
```
BenTheElder: this is not great but not exactly our fault and we can revisit it later and it's orthogonal to the actual CLI work for ipv6

pkg/cluster/config/default.go
BenTheElder: we do write the IP now

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -319,12 +355,24 @@ controlPlaneEndpoint: {{ .ControlPlaneEndpoint }}
 # so we need to ensure the cert is valid for localhost so we can talk
 # to the cluster after rewriting the kubeconfig to point to localhost
 apiServer:
-  certSANs: [localhost, {{.APIServerAddress}}]
+  certSANs: [localhost, "{{.APIServerAddress}}"]
```
BenTheElder: controlPlaneEndpoint above needs quoting

BenTheElder:
I had to add a tiny commmit to fix the quoting in the config, but otherwise this is good to go. It's merged now! :tada:

PR #352, opened 2019-02-25
Refactored load cmd to simplify library usage

### LINKED ISSUE #351: Improve kind as a library for loading images
kind's load command wraps the `docker save` command to save a container image as a tar archive and 
https://github.com/kubernetes-sigs/kind/blob/2c5195712b0b0edbe06d5ca5e777ab97d11cb029/cmd/kind/load/docker-image/docker-image.go#L103-L122

The `loadImage` function provides a good API for using kind as a library, however, this leaves the user to "docker save" any container image. My proposal is to decouple saving-image from the runE command so that it can be used as:

```go
import (
    "sigs.k8s.io/kind/load/cmd/kind/load"
)

func main() {
     ctx := cluster.NewContext(clusterName)                                     
                                                                                
    // Create a kind cluster and all that...                                            
                                                                                
    nodes, _ := ctx.ListNodes()
    // "docker save" and "docker load".
    load.SaveImage("image:tag", nodes)
```

Also, the code for `loadImage` is replicated in [docker-image](https://github.com/kubernetes-sigs/kind/blob/master/cmd/kind/load/docker-image/docker-image.go#L125) and [image-archive](https://github.com/kubernetes-sigs/kind/blob/master/cmd/kind/load/image-archive/image-archive.go#L109).
Would it be good to move the code somewhere like `sigs.k8s.io/kind/cmd/kind/load/utils` (please suggest a name that is not "utils").

/assign
cc @BenTheElder @neolit123

alejandrox1:
Resolves #351
/kind cleanup

### DIFF
--- a/cmd/kind/load/docker-image/docker-image.go
+++ b/cmd/kind/load/docker-image/docker-image.go
@@ -19,7 +19,6 @@ package load
 
 import (
 	"os"
-	"path/filepath"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
@@ -27,7 +26,6 @@ import (
 	"sigs.k8s.io/kind/pkg/cluster"
 	clusternodes "sigs.k8s.io/kind/pkg/cluster/nodes"
 	"sigs.k8s.io/kind/pkg/container/docker"
-	"sigs.k8s.io/kind/pkg/fs"
 )
 
 type flagpole struct {
@@ -100,43 +98,19 @@ func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
 		}
 	}
 
-	// Save the image into a tar
-	dir, err := fs.TempDir("", "image-tar")
-	if err != nil {
-		return errors.Wrap(err, "failed to create tempdir")
-	}
-	defer os.RemoveAll(dir)
-	imageTarPath := filepath.Join(dir, "image.tar")
-
-	err = docker.Save(args[0], imageTarPath)
+	// Save the image into a tar.
+	dir, imageTarPath, err := docker.SaveToTarball(args[0])
 	if err != nil {
 		return err
 	}
+	defer os.RemoveAll(dir)
 
-	// Load the image into every node
+	// Load the image into every selected node.
 	for _, node := range selectedNodes {
-		if err := loadImage(imageTarPath, &node); err != nil {
+		if err := node.LoadTarImage(imageTarPath); err != nil {
 			return err
 		}
 	}
-	return nil
-}
 
-func loadImage(imageTarName string, node *clusternodes.Node) error {
-	// Copy image tar to each node
-	f, err := os.Open(imageTarName)
-	if err != nil {
-		return errors.Wrap(err, "failed to open image")
-	}
-	defer f.Close()
-
-	// Load image into each node
-	cmd := node.Command(
-		"docker", "load",
-	)
-	cmd.SetStdin(f)
-	if err := cmd.Run(); err != nil {
-		return errors.Wrap(err, "failed to load image")
-	}
 	return nil
 }
--- a/cmd/kind/load/image-archive/image-archive.go
+++ b/cmd/kind/load/image-archive/image-archive.go
@@ -18,8 +18,6 @@ limitations under the License.
 package load
 
 import (
-	"os"
-
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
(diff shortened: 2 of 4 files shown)


alejandrox1:
https://go.googlesource.com/tools/ is a bit flaky today
/retest

cmd/kind/load/image-archive/image-archive.go
```diff
@@ -99,28 +98,9 @@ func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
 
 	// Load the image into every node
 	for _, node := range selectedNodes {
-		if err := loadImage(args[0], &node); err != nil {
+		if err := loadutils.LoadImage(args[0], &node); err != nil {
```
neolit123: side note: do we need to check if `len(args) > 0`?
[1]

cmd/kind/load/docker-image/docker-image.go
```diff
@@ -100,43 +96,11 @@ func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
...
-	dir, err := fs.TempDir("", "image-tar")
-	if err != nil {
-		return errors.Wrap(err, "failed to create tempdir")
-	}
-	defer os.RemoveAll(dir)
-	imageTarPath := filepath.Join(dir, "image.tar")
-
-	err = docker.Save(args[0], imageTarPath)
+	// Save the image and load it into every selected node.
+	err = loadutils.SaveAndLoadImage(args[0], selectedNodes)
```
neolit123: [1]

cmd/kind/load/image-archive/image-archive.go
alejandrox1: not really, the `NewCommand` func already does this:
```go
cmd := &cobra.Command{                                                      
    Args: func(cmd *cobra.Command, args []string) error {                   
        if len(args) < 1 {                                                  
            return errors.New("name of image archive is required")          
        }
    ...                                                                   
```

cmd/kind/load/docker-image/docker-image.go
```diff
@@ -18,16 +18,12 @@ limitations under the License.
 package load
 
 import (
-	"os"
-	"path/filepath"
-
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
+	loadutils "sigs.k8s.io/kind/cmd/kind/load/load-utils"
```
BenTheElder: load logic should probably live in `pkg/`, importing kind as a library shouldn't require looking in `cmd` imo.

we can do something like pkg/cluster context.LoadImageTar and combine that with a util in pkg/container/docker to save tars

BenTheElder: Ideally we should probably refactor & document so that something like:
- pkg/cluster
- pkg/cluster/nodes
- pkg/cluster/config (and api versions)
- NOT pkg/cluster/\<anything else\>

Are used for library usage, and the rest are just used internally or are the commands themselves.

alejandrox1: Done. Ended up moving the functionality into /pkg/cluster/nodes to maintain the load cmd logic as is. Please tell me what you think

cmd/kind/load/image-archive/image-archive.go
leodido: Why not simply using `cobra.MinimumNArgs(int)` ?

alejandrox1:
@BenTheElder @neolit123 could you please take another look? :slightly_smiling_face:

cmd/kind/load/docker-image/docker-image.go
```diff
@@ -100,43 +95,12 @@ func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
...
-	if err != nil {
-		return err
-	}
-
-	// Load the image into every node
+	// Save the image and load it into every selected node.
 	for _, node := range selectedNodes {
-		if err := loadImage(imageTarPath, &node); err != nil {
-			return err
+		if err := node.LoadContainerImage(args[0]); err != nil {
```
BenTheElder: this is a pretty major performance regression because we now ave the image to a tarball once per node

alejandrox1: @BenTheElder thank you for catching that!
I think I fixed it

BenTheElder:
I had to refactor some of this for https://github.com/kubernetes-sigs/kind/pull/461

BenTheElder:
thanks again, sorry about the delay in those conflicting changes, I spent a lot of time trying to make them non-breaking before deciding we should just move forward since we're in alpha anyhow :sweat_smile:

PR #357, opened 2019-03-01
refactor: break up cluster context

BenTheElder:
This is part one of refactoring cluster creation.

- create an internal context package that can be passed into implementations of complex methods like Create()
- make create & delete consume this package, move them to separate internal packages
- make the user facing context wrap this context and the create and delete packages
  - the user facing pkg/cluster.Context object now only contains shim methods
- move create options to user facing `pkg/cluster/create.ClusterOptions`
- move more public cluster constants to the constants package

This contains NO logic changes, only re-organizing some of the types. The next cleanups will:
- decouple slightly from config
- refactor the internal create package

These are a blocker to v1alpha3, and more generally making Creation extensible. We need access to a manageable common internal context and separation between user facing options and the internals

This also means less non trivial logic is crammed into the nebulous `cluster` package, but the package is just as powerful to the user.


BenTheElder:
fixed a lint error with a comment.

PR #320, opened 2019-02-20
Remove ARCH specific apt repo info to enable arm support

hh:

PR #358, opened 2019-03-04
Support linux/arm64 in kind

dims:
Picked up a commit by @hh from https://github.com/kubernetes-sigs/kind/pull/320

Install go-1.12 from https://storage.googleapis.com/golang/go1.12.linux-arm64.tar.gz
Install docker-ce using https://docs.docker.com/install/linux/docker-ce/ubuntu/ 

```
cd $GOPATH/src/sigs.k8s.io/kind && go install .
kind build base-image --image base:latest --source $GOPATH/src/sigs.k8s.io/kind/images/base --loglevel debug
kind build node-image --type apt --base-image base:latest --kube-root=$GOPATH/src/k8s.io/kubernetes/
kind create cluster --image kindest/node:latest
```

Notes:
- ONLY `--type apt` works, need to get `docker`/`make` working as that will need changes in kubernetes/kubernetes


dims:
cc @hh

pkg/util/osarch.go
```diff
@@ -0,0 +1,68 @@
...
+	case "linux":
+		switch runtime.GOARCH {
+		case "amd64":
+			return "linux" + separator + "amd64"
+		case "arm64":
+			return "linux" + separator + "arm64"
+		}
+	case "windows":
+		if runtime.GOARCH == "amd64" {
+			return "darwin" + separator + "amd64"
```
tao12345666333: it should be windows ？

pkg/util/osarch.go
```diff
@@ -0,0 +1,68 @@
...
+		return "amd64"
+	case "arm64":
+		return "arm64"
+	}
+	panic(fmt.Sprintf("unsupported architecture %s", runtime.GOARCH))
+}
+
+// GetOSandArch validates/returns the current os/arch combination if supported and panics otherwise
+func GetOSandArch(separator string) string {
+	switch runtime.GOOS {
```
tao12345666333: why don’t we call those two functions directly?

dims: @tao12345666333 because then we will allow windows+arm64 (for example) which is not something we are likely to support

pkg/build/kube/bazelbuildbits.go
```diff
@@ -60,9 +60,7 @@ func (b *BazelBuildBits) Build() error {
 	// make sure we cd back when done
 	defer os.Chdir(cwd)
 
-	// TODO(bentheelder): we assume the host arch, but cross compiling should
```
BenTheElder: This TODO is still true

dims: Sorry, added back.

dims:
i've opened an issue against k/k for the "docker"/"make" types:
https://github.com/kubernetes/kubernetes/issues/74965

I could not even poke at the bazel one, will leave it to others to poke at it. I have a machine if someone wants to try.

dims:
for the record `bazel build --platforms=@io_bazel_rules_go//go/toolchain:linux_arm64 //cmd/kubeadm:kubeadm //cmd/kubectl:kubectl //cmd/kubelet:kubelet //build:docker-artifacts` works fine in k/k

Though the caveat is that bazel needs to be installed from source
```
wget https://github.com/bazelbuild/bazel/releases/download/0.23.1/bazel-0.23.1-dist.zip
sudo apt-get install openjdk-8-jdk pkg-config zip g++ zlib1g-dev unzip
./compile.sh
```

pkg/build/kube/bazelbuildbits.go
```diff
@@ -62,7 +62,7 @@ func (b *BazelBuildBits) Build() error {
 
 	// TODO(bentheelder): we assume the host arch, but cross compiling should
 	// be possible now
-	bazelGoosGoarch := fmt.Sprintf("linux_%s", runtime.GOARCH)
+	bazelGoosGoarch := util.GetOSandArch("_")
```
BenTheElder: whoops, this sort of thing is wrong on mac.

PR #364, opened 2019-03-07
Refactor part 3: speed up creation 💨

BenTheElder:
- create node containers concurrently
- make worker join concurrent
- make image tarball loading concurrent
- minor status tweaks (given that we've changed the scope / ordering, I tweaked these)

### DIFF
--- a/pkg/cluster/internal/create/actions/kubeadmjoin/join.go
+++ b/pkg/cluster/internal/create/actions/kubeadmjoin/join.go
@@ -100,9 +100,26 @@ func joinWorkers(
 	defer ctx.Status.End(false)
 
 	// TODO(bentheelder): this should be concurrent
+	errChan := make(chan error, len(workers))
+	defer close(errChan)
 	for _, node := range workers {
-		if err := runKubeadmJoin(ctx, allNodes, &node); err != nil {
+		node := node // capture loop variable
+		go func() {
+			errChan <- runKubeadmJoin(ctx, allNodes, &node)
+		}()
+	}
+
+	// watch for all worker joins to finish
+	// NOTE: we don't use a waitgroup because we want to exit early
+	count := 0
+	for err := range errChan {
+		// if any errored, bail out
+		if err != nil {
 			return err
+		}
+		count++
+		if count == len(workers) {
+			break
 		}
 	}
 
--- a/pkg/cluster/internal/create/actions/kubeadmjoin/join.go
+++ b/pkg/cluster/internal/create/actions/kubeadmjoin/join.go
@@ -77,7 +77,7 @@ func joinSecondaryControlPlanes(
 	allNodes []nodes.Node,
 	secondaryControlPlanes []nodes.Node,
 ) error {
-	ctx.Status.Start("Joining more control plane nodes 🎮")
+	ctx.Status.Start("Joining more control-plane nodes 🎮")
 	defer ctx.Status.End(false)
 
 	// TODO(bentheelder): this should be concurrent
--- a/pkg/cluster/internal/create/nodes.go
+++ b/pkg/cluster/internal/create/nodes.go
@@ -225,7 +225,7 @@ func (d *nodeSpec) Create(clusterLabel string) (node *nodes.Node, err error) {
 	case constants.WorkerNodeRoleValue:
 		node, err = nodes.CreateWorkerNode(d.Name, d.Image, clusterLabel, d.ExtraMounts)
 	default:
-		return nil, errors.Errorf("unkown node role: %s", d.Role)
+		return nil, errors.Errorf("unknown node role: %s", d.Role)
 	}
 	return node, err
 }
--- a/pkg/cluster/internal/create/nodes.go
+++ b/pkg/cluster/internal/create/nodes.go
@@ -103,7 +103,7 @@ func createNodeContainers(
 
 	// create all of the node containers, concurrently
 	desiredNodes := nodesToCreate(cfg, clusterName)
-	status.Start("Preparing node containers... " + strings.Repeat("📦", len(desiredNodes)))
+	status.Start("Preparing nodes " + strings.Repeat("📦", len(desiredNodes)))
 	nodeChan := make(chan *nodes.Node, len(desiredNodes))
 	errChan := make(chan error)
 	defer close(nodeChan)
--- a/pkg/cluster/internal/create/nodes.go
+++ b/pkg/cluster/internal/create/nodes.go
@@ -19,6 +19,7 @@ package create
 import (
 	"fmt"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
@@ -26,6 +27,7 @@ import (
 	"sigs.k8s.io/kind/pkg/cluster/config"
 	"sigs.k8s.io/kind/pkg/cluster/constants"
 	"sigs.k8s.io/kind/pkg/cluster/nodes"
+	"sigs.k8s.io/kind/pkg/container/cri"
 	logutil "sigs.k8s.io/kind/pkg/log"
 )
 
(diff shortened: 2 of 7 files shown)


pkg/cluster/internal/create/actions/kubeadmjoin/join.go
```diff
@@ -100,10 +100,27 @@ func joinWorkers(
...
+	// watch for all worker joins to finish
+	// NOTE: we don't use a waitgroup because we want to exit early
+	count := 0
+	for err := range errChan {
+		// if any errored, bail out
+		if err != nil {
 			return err
 		}
+		count++
+		if count == len(workers) {
```
munnerz: Does ctx already have a timeout? Do we need to also wait for `ctx.Done` using a switch?

pkg/cluster/internal/create/nodes.go
```diff
@@ -82,85 +84,148 @@ func convertReplicas(nodes []config.Node) []config.Node {
...
+	errChan := make(chan error)
+	defer close(nodeChan)
+	defer close(errChan)
+	for _, desiredNode := range desiredNodes {
+		desiredNode := desiredNode // capture loop variable
+		go func() {
+			// create the node into a container (docker run, but it is paused, see createNode)
+			node, err := desiredNode.Create(clusterLabel)
+			if err != nil {
+				errChan <- err
```
munnerz: Missing returns here? Else we'll put multiple errors onto the chans

pkg/cluster/nodes/node.go
```diff
@@ -163,10 +163,9 @@ func tryUntil(until time.Time, try func() bool) bool {
 func (n *Node) LoadImages() {
 	// load images cached on the node into docker
 	if err := n.Command(
-		"find",
-		"/kind/images",
-		"-name", "*.tar",
-		"-exec", "docker", "load", "-i", "{}", ";",
+		"/bin/bash", "-c",
+		// use xargs to load images in parallel
+		`find /kind/images -name *.tar -print0 | xargs -0 -n 1 -P $(nproc) docker load -i`,
```
munnerz: /this-is-fine

pkg/cluster/internal/create/nodes.go
BenTheElder: yes this should return 👍

pkg/cluster/nodes/node.go
BenTheElder: yeah.

So this means that we list everything, collect it up in xargs, and then run one docker load per core.

It's not actually terribly important how many cores, mostly that we run multiple in parallel because that's definitely faster, non-scientific experimentation suggested nproc is good enough for now.

pkg/cluster/internal/create/actions/kubeadmjoin/join.go
BenTheElder: it does not yet. 
I don't think we care though, we'll list up the containers and forcibly delete them on exit if retain is not set, if it is set we wouldn't have done anything either way, we'll just leave them around to poke at for debugging

pkg/cluster/internal/create/nodes.go
BenTheElder: returns added

pkg/cluster/nodes/node.go
BenTheElder: FWIW this is actually pretty simple:

`find`: we're using find to locate files
`/kind/images`: in this directory
`-name *.tar`: named `*.tar`
`-print0`: print the names out null terminated, so we can support files with spaces etc properly
`| xargs`: pipe the results to xargs for running over each file
`-0`: tell xargs to expect null terminated inputs
`-n 1`: tell xargs to run each input individually (IE one file at a time, since `docker load` requires this)
`-P $(nproc)`: tell xargs to run in parallel with at most one process per core
`docker load -i`: xargs should run `docker load -i ${input}` to load the image, `-i` tells docker load to use a file name instead of stdin.

BenTheElder:
flaky dependency check stuff ... 😞 
/retest

pkg/cluster/nodes/node.go
neolit123: would be nice to have this as a comment above the function.
but also why not call `docker load...` in go routines. would that work?

BenTheElder: that's slower and more code. if someone can't read this snippet they shouldn't touch it :^)

pkg/cluster/internal/create/actions/kubeadmjoin/join.go
```diff
@@ -77,7 +77,7 @@ func joinSecondaryControlPlanes(
 	allNodes []nodes.Node,
 	secondaryControlPlanes []nodes.Node,
 ) error {
-	ctx.Status.Start("Joining more control plane nodes 🎮")
+	ctx.Status.Start("Joining more control-plane nodes 🎮")
```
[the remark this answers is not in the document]
BenTheElder: Sets are usually referred to in plural regardless of the size. We're joining whichever exist.

Kubeadm doesn't support 2 CPs does it? IIRC it must be 3

pkg/cluster/nodes/node.go
BenTheElder: to be clear: there's always been a comment here about what we're doing, before anyone went to edit this I'd expect them to understand what it does and why. It shouldn't be a big deal to check the docs for these commands and the only bash syntax we use here is a pipe.

pkg/cluster/internal/create/nodes.go
```diff
@@ -82,85 +84,148 @@ func convertReplicas(nodes []config.Node) []config.Node {
...
+			nodeChan <- node
+		}()
+	}
+
+	// collect nodes
+	allNodes := []nodes.Node{}
+	for {
+		select {
+		case node := <-nodeChan:
 			// TODO(bentheelder): nodes should maybe not be pointers /shrug
```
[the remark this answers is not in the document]
BenTheElder: the intention originally with the slice was to return it to passing for other methods that accept []Node
slices are already reference types so this is redundant, and copying Node is safe now, I should probably change some of these to drop the pointer on Node but it works fine either way.

pkg/cluster/internal/create/nodes.go
```diff
@@ -82,85 +84,148 @@ func convertReplicas(nodes []config.Node) []config.Node {
...
-		status.Start(fmt.Sprintf("[%s] Waiting for Docker to be ready 🐋", name))
-		// wait for docker to be ready
-		if !node.WaitForDocker(time.Now().Add(time.Second * 30)) {
-			// TODO(bentheelder): logging here
-			return allNodes, errors.New("timed out waiting for docker to be ready on node")
-		}
+	// wait for docker to be ready
+	if !node.WaitForDocker(time.Now().Add(time.Second * 30)) {
+		// TODO(bentheelder): logging here
+		return errors.New("timed out waiting for docker to be ready on node")
```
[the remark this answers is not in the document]
BenTheElder: done, also on the proxy error

pkg/cluster/internal/create/actions/kubeadmjoin/join.go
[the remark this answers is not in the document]
neolit123: it does support 2.

BenTheElder:
hmm did I goof the config? https://github.com/kubernetes/test-infra/pull/11702

/test pull-kind-verify

neolit123:
> error: pkg/cluster/internal/create/nodes.go:228:29: "unkown" is a misspelling of "unknown"
Found spelling errors!

there is a typo, but not sure why it's not highlighted here https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/364/pull-kind-verify/617

BenTheElder:
hmm so that's true, but the job also still seems to be running the dep verification which should not be running in that job anymore !? 😕

BenTheElder:
the config updater didn't run

Issue #354, opened 2019-02-26
handle lowercase _proxy variables


BenTheElder:
see previous discussion: https://github.com/kubernetes-sigs/kind/issues/306

BenTheElder:
This should be a very easy change, see [the nodes package](https://github.com/kubernetes-sigs/kind/blob/master/pkg/cluster/nodes/node.go), `NeedProxy` and `SetProxy` methods, Ideally these should be refactored to:
- move the proxy environment variable strings & detection into the common function
  - return a private struct with proxy metadata
  - make both `NeedProxy` and `SetProxy` use this method
- support lowercase versions of `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`

There will need to be a priority between these values, arbitrarily I will say uppercase should be preferred over lowercase, since uppercase variables are common practice for exported environment variables.


/good-first-issue

aojea:
@BenTheElder PR was merged, can we close it?

PR #372, opened 2019-03-11
handle lowercase proxy variables

aojea:
Use an internal struct to store the proxy environment variables.

The function `getProxyDetails()` populates the map inside the internal struct with the proxy environment variables giving preference to the uppercase one.

The function `SetProxy()` consume the map inside the struct to populate the systemd drop-in file with the proxy environment variables

Reference #354

### DIFF
--- a/pkg/cluster/internal/create/nodes.go
+++ b/pkg/cluster/internal/create/nodes.go
@@ -155,7 +155,7 @@ func fixupNode(node *nodes.Node) error {
 	if nodes.NeedProxy() {
 		if err := node.SetProxy(); err != nil {
 			// TODO: logging here
-			return errors.Errorf("failed to set proxy for %s", node.Name())
+			return errors.Wrapf(err, "failed to set proxy for node %s", node.Name())
 		}
 	}
 
--- a/pkg/cluster/internal/create/nodes.go
+++ b/pkg/cluster/internal/create/nodes.go
@@ -155,7 +155,7 @@ func fixupNode(node *nodes.Node) error {
 	if nodes.NeedProxy() {
 		if err := node.SetProxy(); err != nil {
 			// TODO: logging here
-			return errors.Wrapf(err, "failed to set proxy for node %s", node.Name())
+			return errors.Errorf("failed to set proxy for %s", node.Name())
 		}
 	}
 
--- a/pkg/cluster/nodes/create.go
+++ b/pkg/cluster/nodes/create.go
@@ -141,12 +141,10 @@ func createNode(name, image, clusterLabel string, role config.NodeRole, mounts [
 
 	// pass proxy environment variables to be used by node's docker deamon
 	if NeedProxy() {
-		details := getProxyDetails()
-		proxies := ""
-		for key, val := range details {
-			proxies += fmt.Sprintf("-e %s=%s ", key, val)
+		proxyDetails := getProxyDetails()
+		for key, val := range proxyDetails.Envs {
+			runArgs = append(runArgs, "-e", fmt.Sprintf("%s=%s", key, val))
 		}
-		runArgs = append(runArgs, proxies)
 	}
 
 	// adds node specific args
--- a/pkg/cluster/nodes/create.go
+++ b/pkg/cluster/nodes/create.go
@@ -19,7 +19,6 @@ package nodes
 import (
 	"fmt"
 	"net"
-	"os"
 
 	"github.com/pkg/errors"
 	"sigs.k8s.io/kind/pkg/cluster/config"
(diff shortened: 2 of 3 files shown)


pkg/cluster/nodes/node.go
```diff
@@ -403,42 +403,46 @@ func (n *Node) WriteFile(dest, content string) error {
...
+		"[Service]\nEnvironment="+proxies)
+	if err != nil {
+		errors.Wrap(err, "failed to create http-proxy drop-in")
 	}
 
 	return nil
 }
 
-// NeedProxy returns true if the host environment appears to have proxy settings
+// NeedProxy returns a struct with the host environment appears to have proxy settings
```
BenTheElder: This should be renamed if we're going to refactor it this way.

I also think the map thing should be an internal detail, we may need proxy options in the future that are not environment variable based, in which case making the map public will make fixing that a breaking change.

BenTheElder:
```go
// NeedProxy returns true if the host environment appears to have proxy settings
func NeedProxy() bool {
    details := getProxyDetails()
    return len(details) > 0
}

func (n *Node) SetProxy() {
    details := getProxyDetails()
    // now set the the proxy settings
}

func getProxyDetails() map[string]string {

}

```

pkg/cluster/internal/create/nodes.go
```diff
@@ -155,7 +155,7 @@ func fixupNode(node *nodes.Node) error {
 	if nodes.NeedProxy() {
 		if err := node.SetProxy(); err != nil {
 			// TODO: logging here
-			return errors.Wrapf(err, "failed to set proxy for node %s", node.Name())
+			return errors.Errorf("failed to set proxy for %s", node.Name())
```
neolit123: could you please explain this change?

aojea: it's a leftover of the first PR, I force pushed and deleted the history. let me fix it
Thanks for catching it

pkg/cluster/nodes/node.go
```diff
@@ -403,42 +403,53 @@ func (n *Node) WriteFile(dest, content string) error {
...
+		"[Service]\nEnvironment="+proxies)
+	if err != nil {
+		errors.Wrap(err, "failed to create http-proxy drop-in")
 	}
 
 	return nil
 }
 
-// NeedProxy returns true if the host environment appears to have proxy settings
+// getProxyDetails returns a struct with the host environment proxy settings
```
BenTheElder: s/struct with/map containing/, this is not a struct :upside_down_face: 
```suggestion
// getProxyDetails returns a map containing the host environment proxy settings
```
it's also worth considering making this _actually_ return a struct instead and ignoring the comment about renaming above then

```go
// proxyDetails contains proxy settings discovered on the host
type proxyDetails struct {
    Envs map[string]string
    // future proxy details here
}

func getProxyDetails() proxyDetails {
```

pkg/cluster/nodes/create.go
```diff
@@ -141,18 +140,13 @@ func createNode(name, image, clusterLabel string, role config.NodeRole, mounts [
...
-	}
-
-	noProxy := os.Getenv("NO_PROXY")
-	if noProxy != "" {
-		runArgs = append(runArgs, "-e", "NO_PROXY="+noProxy)
+	if NeedProxy() {
+		details := getProxyDetails()
+		proxies := ""
+		for key, val := range details {
+			proxies += fmt.Sprintf("-e %s=%s ", key, val)
```
BenTheElder: this is wrong, we should be building an argument vector here in `runArgs`
```suggestion
			runArgs = append(runArgs, "-e", fmt.Sprintf(%s=%s ", key, val))
```

We should probably rename `getProxyDetails` to `getProxyEnvs` to be clear what this map is and how we're using it

pkg/cluster/nodes/node.go
```diff
@@ -403,42 +403,59 @@ func (n *Node) WriteFile(dest, content string) error {
...
+
+// getProxyDetails returns a struct with the host environment proxy settings
 // that should be passed to the nodes
-func NeedProxy() bool {
-	for _, env := range proxyEnvs {
-		if os.Getenv(env) != "" {
-			return true
+func getProxyDetails() proxyDetails {
+	var proxyEnvs = []string{"HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY"}
+	var val string
```
BenTheElder: tiny nit but I want to get this in 🙃 -- this should not be hoisted out of the loop here.

this is functionally OK though, so I won't block merging on it.

PR #373, opened 2019-03-11
implement v1alpha3

BenTheElder:
- create v1alpha3
- drop v1alpha1
- update internal types to match v1alpha3
- update conversions
- fixup internal logic for the new format
- make load balancers implicit
- remove ExternalEtcdRole from config until it is implemented

TODO:
- add more of the cluster fields to v1alpha3, this focuses on the fields necessary to convert the existing fields
- add a structure to v1alpha3 for controlling the load balancer settings (for now just the image)

/assign @munnerz


pkg/cluster/config/v1alpha2/conversion.go
```diff
@@ -0,0 +1,66 @@
...
+
+package v1alpha2
+
+import (
+	conversion "k8s.io/apimachinery/pkg/conversion"
+	config "sigs.k8s.io/kind/pkg/cluster/config"
+)
+
+func Convert_v1alpha2_Config_To_config_Cluster(in *Config, out *config.Cluster, s conversion.Scope) error {
+	// TODO(bentheelder): try to convert kubeadm config patches?
```
BenTheElder: we should definitely convert them going in this direction
/hold

cc @munnerz , this file is the hairy part. Technically we don't do round-tripping 1:1, but that's probably OK since we don't actually ever convert back to an external type and the types are alpha.

BenTheElder:
I think that's a prow flake actually ??
/retest

BenTheElder:
confirmed locally that with multiple control planes in a v1alpha2 config an external LB is created and works.

pkg/cluster/config/v1alpha2/conversion.go
```diff
@@ -0,0 +1,81 @@
...
+	for _, node := range in.Nodes {
+		// TODO(bentheelder): if v1alpha3 does re-gain per-node patches, drop
+		// this snippet
+		// Copy config patches from nodes to the cluster level, since we only
+		// generate a cluster-wide config anyhow
+		for _, patch := range node.KubeadmConfigPatches {
+			out.KubeadmConfigPatches = append(out.KubeadmConfigPatches, patch)
+		}
+		for _, patch := range node.KubeadmConfigPatchesJSON6902 {
+			out.KubeadmConfigPatchesJSON6902 = append(out.KubeadmConfigPatchesJSON6902, patch)
```
neolit123: this will apply patches from multiple nodes that previously could have been unique, to all nodes in v1alpha3.
that said:
- this is probably better than discarding patches and e.g. taking only the patches from the first CP node.
- we probably don't have that many users that use patches.

pkg/cluster/config/v1alpha2/conversion.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func Convert_v1alpha2_Node_To_config_Node(in *Node, out *config.Node, s conversion.Scope) error {
+	return autoConvert_v1alpha2_Node_To_config_Node(in, out, s)
+}
+
+func Convert_config_Cluster_To_v1alpha2_Config(in *config.Cluster, out *Config, s conversion.Scope) error {
+	// TODO(bentheelder): try to convert kubeadm config patches?
+	// Not sure if that makes sense since these aren't cluster wide, and we don't
+	// actually roundtrip anywho? :thinking:
```
neolit123: seems OK to me to exclude the patches..
proper rountrips would be nice to have in future versions though.

pkg/cluster/config/v1alpha2/types.go
```diff
@@ -69,10 +69,6 @@ const (
 	ControlPlaneRole NodeRole = "control-plane"
 	// WorkerRole identifies a node that hosts a Kubernetes worker
 	WorkerRole NodeRole = "worker"
-	// ExternalEtcdRole identifies a node that hosts an external-etcd instance.
-	// WARNING: this node type is not yet implemented!
-	// Please note that `kind` nodes hosting external etcd are not kubernetes nodes
-	ExternalEtcdRole NodeRole = "external-etcd"
```
neolit123: i don't think we should remove fields from alpha2 as part of the alpha3 introduction.

pkg/cluster/config/v1alpha3/types.go
```diff
@@ -0,0 +1,79 @@
...
+	ExtraMounts []cri.Mount `json:"extraMounts,omitempty"`
+}
+
+// NodeRole defines possible role for nodes in a Kubernetes cluster managed by `kind`
+type NodeRole string
+
+const (
+	// ControlPlaneRole identifies a node that hosts a Kubernetes control-plane.
+	//
+	// NOTE: in single node clusters, control-plane nodes act also as a worker nodes
```
neolit123: possibly use two lines only:
```
// ControlPlaneRole identifies a node that hosts a Kubernetes control-plane.
// NOTE: in single node clusters, control-plane nodes will allow workloads.
```
^ proposed minor change to the note.

pkg/cluster/config/v1alpha2/types.go
BenTheElder: this is a value, not a field, and it is rejected during validation, we're exporting an invalid value and then validating that it is invalid currently 🤷‍♂️

pkg/cluster/config/v1alpha2/conversion.go
BenTheElder: > this will apply patches from multiple nodes that previously could have been unique, to all nodes in v1alpha3.

we only generate one config file currently, so if anyone was using them they were already effectively global patches

pkg/cluster/config/v1alpha2/conversion.go
BenTheElder: we could include them currently by shoe-horning them onto one of the nodes, but I'd rather revisit it later given we don't depend on this and that might not be the right answer 🤔

pkg/cluster/config/v1alpha3/types.go
BenTheElder: sure, SGTM

BenTheElder: done

pkg/cluster/config/v1alpha2/conversion.go
BenTheElder: I have a WIP to generate a config file per node, for better version skew. I'll revisit per-node patches as an _actually_ supported thing in addition to global patches as part of that.

tao12345666333:
Looked quickly, it looks great to me. Thanks. :+1:

pkg/cluster/config/default.go
```diff
@@ -26,8 +26,8 @@ func addDefaultingFuncs(scheme *runtime.Scheme) error {
 	return RegisterDefaults(scheme)
 }
 
-// SetDefaults_Config sets uninitialized fields to their default value.
-func SetDefaults_Config(obj *Config) {
+// SetDefaults_Cluster sets uninitialized fields to their default value.
+func SetDefaults_Cluster(obj *Cluster) {
```
munnerz: 🤔 I didn't think defaults were usually applied as part of the internal API version? Although I could definitely be wrong 😄 (/me goes away to dig into Kubernetes)

pkg/cluster/config/types.go
```diff
@@ -25,58 +25,54 @@ import (
...
+type Cluster struct {
 	// TypeMeta representing the type of the object and its API schema version.
-	metav1.TypeMeta
+	metav1.TypeMeta `json:",inline"`
 
-	// Nodes contains the list of nodes defined in the `kind` Config
-	Nodes []Node `json:"nodes,"`
-}
+	// Nodes contains the list of nodes defined in the `kind` Cluster
+	Nodes []Node `json:"nodes"`
```
munnerz: Internal API types shouldn't have JSON tags. They will never actually be serialised/deserialised, and historically certain generators have tripped up on this (they've used the presence of even a single JSON tag to denote the API types as 'external')

pkg/cluster/config/types.go
```diff
@@ -25,58 +25,54 @@ import (
...
+	Role NodeRole `json:"role,omitempty"`
+
+	// Image is the node image to use when creating this node
+	Image string `json:"image,omitempty"`
+
+	/* Advanced fields */
+
 	// ExtraMounts describes additional mount points for the node container
-	// These may be used to bind a hostpath
+	// These may be used to bind a hostPath
```
munnerz: `bind a hostPath into a single node container`?

pkg/cluster/config/types.go
```diff
@@ -25,58 +25,54 @@ import (
...
 }
 
 // NodeRole defines possible role for nodes in a Kubernetes cluster managed by `kind`
 type NodeRole string
 
 const (
-	// ControlPlaneRole identifies a node that hosts a Kubernetes control-plane
-	// NB. in single node clusters, control-plane nodes act also as a worker nodes
+	// ControlPlaneRole identifies a node that hosts a Kubernetes control-plane.
+	// NOTE: in single node clusters, control-plane nodes act also as a worker nodes
```
munnerz: nit (as it's not introduced in this PR anyway) - maybe clarify "... by removing the taint from the control plane node"?

pkg/cluster/config/v1alpha2/conversion.go
```diff
@@ -0,0 +1,81 @@
...
+		for _, patch := range node.KubeadmConfigPatches {
+			out.KubeadmConfigPatches = append(out.KubeadmConfigPatches, patch)
+		}
+		for _, patch := range node.KubeadmConfigPatchesJSON6902 {
+			out.KubeadmConfigPatchesJSON6902 = append(out.KubeadmConfigPatchesJSON6902, patch)
+		}
+
+		// skip now-implicit external load balancers
+		if node.Role == ExternalLoadBalancerRole {
+			continue
```
munnerz: I don't think we can just skip this altogether here - it will mean that a conversion from v1alpha2->internal->v1alpha3->internal->v1alpha2 will not be bijective, and will produce different results.

This is an important gotcha of writing conversions - in practice, I'm not sure what the effect of *not* doing it is, but the apimachinery assumes that this *is* the case, so we are breaking the rules here 🙄 

Instead here, we would usually store some additional information on annotations on the new resource, so that we can convert *back* from that format.

pkg/cluster/config/v1alpha2/conversion.go
```diff
@@ -0,0 +1,81 @@
...
+		}
+
+		// always convert the node
+		convertedNode := config.Node{}
+		if err := Convert_v1alpha2_Node_To_config_Node(&node, &convertedNode, s); err != nil {
+			return err
+		}
+		out.Nodes = append(out.Nodes, convertedNode)
```
munnerz: What if someone specified `replicas: 0`?

pkg/cluster/config/v1alpha3/types.go
```diff
@@ -0,0 +1,78 @@
...
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// Cluster contains kind cluster configuration
+type Cluster struct {
+	// TypeMeta representing the type of the object and its API schema version.
+	metav1.TypeMeta `json:",inline"`
+
+	// Nodes contains the list of nodes defined in the `kind` Cluster
+	Nodes []Node `json:"nodes"`
```
munnerz: Add a note here describing the defaulting behaviour (i.e. a single control-plane node will be used if not specified)

pkg/cluster/config/v1alpha3/types.go
```diff
@@ -0,0 +1,78 @@
...
+// required for the assigned role in the Kubernetes cluster
+type Node struct {
+	// Role defines the role of the node in the in the Kubernetes cluster
+	// created by kind
+	//
+	// Defaults to "control-plane"
+	Role NodeRole `json:"role,omitempty"`
+
+	// Image is the node image to use when creating this node
+	Image string `json:"image,omitempty"`
```
munnerz: Add note about default?

pkg/cluster/config/v1alpha3/types.go
```diff
@@ -0,0 +1,78 @@
...
+
+	/* Advanced fields */
+
+	// ExtraMounts describes additional mount points for the node container
+	// These may be used to bind a hostPath
+	ExtraMounts []cri.Mount `json:"extraMounts,omitempty"`
```
munnerz: Not for part of this PR, but it's just highlighted it to me... we don't currently version the `cri.Mount` structure - we should be careful here, as a change to the `cri` package could inadvertently affect users and sneak through review (as it's not in the `config` package)

pkg/cluster/internal/create/nodes.go
```diff
@@ -209,7 +214,16 @@ func nodesToCreate(cfg *config.Config, clusterName string) []nodeSpec {
 		})
 	}
 
-	// TODO(bentheelder): handle implicit nodes as well
+	// add an external load balancer if there are multiple control planes
+	if controlPlanes > 1 {
```
munnerz: There should be a note, probably on the Role field or on the Nodes field, to explain that if you specify more than 1 control plane node then a load balancer will be provisioned.

IMO, API docs should themselves serve users quite well as a reference for how things work (even if we also have additional proper documentation for this too).

pkg/cluster/config/v1alpha3/types.go
BenTheElder: Yes. At this point that structure is expected to be stable, but we should be explicit about that.

pkg/cluster/config/v1alpha2/conversion.go
BenTheElder: What a jerk /s
We should test how 0.1.0 behaves specifically and reflect that.

pkg/cluster/config/v1alpha2/conversion.go
BenTheElder: Er so maybe this is the wrong place to do this but effectively I do want to drop this. Previously you needed to specify this and now it is implicit.

We don't support roundtripping properly because we did actually break between versions which is only allowed because the types are alpha by my understanding.

We also don't ever convert backwards yet. Only to internal.

pkg/cluster/config/types.go
BenTheElder: ER isn't it implied that they're bound to the current node?

pkg/cluster/config/types.go
BenTheElder: Interesting. Will drop these.

pkg/cluster/config/default.go
BenTheElder: We have to default the internal type as well currently because its used by library consumers 🤔

pkg/cluster/config/types.go
neolit123: yep, i missed that.

pkg/cluster/config/v1alpha2/conversion.go
neolit123: something similar was discussed in one of my comments earlier.
we are indeed breaking the rountrip rules a little here, so it's a question if we want the full rountrip support now, or close our eyes for the alpha2 -> alpha3 transition, but consider the proper conversations for say alpha3->next.

cc @fabriziopandini

pkg/cluster/config/v1alpha2/conversion.go
neolit123: i did test it at some point and i think it was silently existing. :)

pkg/cluster/config/v1alpha2/conversion.go
BenTheElder: yeah, I think roundtrip support is important long term, but we're intentionally breaking compatibility between these now in alpha, you cannot express roundtrip because of this. we could have just deleted all old alphas.

BenTheElder: beta and onwards versions should be round trippable without issue.

pkg/cluster/internal/create/nodes.go
BenTheElder: done. the API docs definitely need work

pkg/cluster/config/v1alpha3/types.go
BenTheElder: done

pkg/cluster/config/v1alpha3/types.go
BenTheElder: done

pkg/cluster/config/types.go
BenTheElder: done

pkg/cluster/config/types.go
BenTheElder: done

pkg/cluster/config/v1alpha2/conversion.go
BenTheElder:
```
$ kind version
0.1.0

$ cat $HOME/kind-config.yaml
kind: Config
nodes:
- role: control-plane
  replicas: 0

$ kind create cluster --config=$HOME/kind-config.yaml
Error: failed to create cluster: please add at least one node with role "control-plane"

Usage:
  kind create cluster [flags]

Flags:
      --config string   path to a kind config file
  -h, --help            help for cluster
      --image string    node docker image to use for booting the cluster
      --name string     cluster context name (default "1")
      --retain          retain nodes for debugging when cluster creation fails
      --wait duration   Wait for control plane node to be ready (default 0s)

Global Flags:
      --loglevel string   logrus log level [panic, fatal, error, warning, info, debug] (default "warning")

failed to create cluster: please add at least one node with role "control-plane"
```

BenTheElder: so we should actually treat it as 0 and skip the node I guess.

BenTheElder: will handle this in a follow-up once we decide

PR #376, opened 2019-03-12
improve patch roundtripping

BenTheElder:
Follow up to https://github.com/kubernetes-sigs/kind/pull/373

In kind versions we shipped with the v1alpha2 config the kubeadm config patches were only respected if they were on the bootstrap control plane node. this re-introduces that behavior when converting from v1alpha2 to internal.

This allows us to preserve the behavior of existing config files for now, while leaving room to introduce per-node patches that are actually used going forward.

### DIFF
--- a/pkg/cluster/config/v1alpha2/conversion.go
+++ b/pkg/cluster/config/v1alpha2/conversion.go
@@ -24,16 +24,11 @@ import (
 )
 
 func Convert_v1alpha2_Config_To_config_Cluster(in *Config, out *config.Cluster, s conversion.Scope) error {
+	foundFirstControlPlane := false // in v1alpha2 the first control plane is special
 	for _, node := range in.Nodes {
-		// TODO(bentheelder): if v1alpha3 does re-gain per-node patches, drop
-		// this snippet
-		// Copy config patches from nodes to the cluster level, since we only
-		// generate a cluster-wide config anyhow
-		for _, patch := range node.KubeadmConfigPatches {
-			out.KubeadmConfigPatches = append(out.KubeadmConfigPatches, patch)
-		}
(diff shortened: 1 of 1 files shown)


BenTheElder:
FYI @neolit123 @munnerz 

This should resolve the roundtripping concerns as well, patches should roundtrip properly now.

pkg/cluster/config/v1alpha2/conversion.go
```diff
@@ -63,14 +73,27 @@ func Convert_v1alpha2_Node_To_config_Node(in *Node, out *config.Node, s conversi
...
-	// TODO(bentheelder): try to convert kubeadm config patches?
-	// Not sure if that makes sense since these aren't cluster wide, and we don't
-	// actually roundtrip anywho? :thinking:
+	foundFirstControlPlane := false // in v1alpha2 the first control plane is special
 	for _, node := range in.Nodes {
 		convertedNode := Node{}
 		if err := Convert_config_Node_To_v1alpha2_Node(&node, &convertedNode, s); err != nil {
 			return err
 		}
+		// the first control plane is treated specially
```
BenTheElder: specifically this new snippet means we roundtrip the global patches to v1alpha2

we should get people off of v1alpha2 though and we should not actually roundtrip to it, because per-node patches will NOT roundtrip correctly as they were ignored in v1alpha2

BenTheElder:
> like i mentioned in the previous PR; we could have closed our eyes on this one. :)

probably, but I want to improve the patch behavior next and avoid breaking existing users :sweat_smile:

munnerz:
It'll be great to get some fuzz tests in this area of the codebase at some point. I'd love to see some signal on whether we're actually getting these conversions correct or not before we starting utilising them more seriously in later api versions 😄 

/lgtm

Issue #377, opened 2019-03-13
Errors running kind on windows 7 using docker toolbox


aojea:
I was trying to use ` kind` on Windows 7 with Docker-toolbox and I've found the following issues that I think makes this combination hard or very difficult to support:

1. [Docker is not mapping ports to localhost](https://github.com/docker/for-win/issues/204), so you have to locate your docker VM IP to access the port forwarded 
```
$ docker-machine ip
192.168.99.100
```

2. `kind` uses privileged containers and remount directories, this screws up the docker-toolbox VM and you can't use docker commands that fails with the following error
 
[docker: Error response from daemon: cgroups: cannot found cgroup mount destination: unknown.](https://github.com/boot2docker/boot2docker/issues/1301)

A workaround is resetting the VirtualBox VM

3. ` kind` fails to create the `kubedm config`, seems that's related to the way that [go handles windows directories](https://github.com/golang/go/commit/153c0da89bca6726545cf4451053235b552d3d51), 

```
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.13.3) 🖼
 ✓ Preparing nodes 📦
 ✗ Creating kubeadm config 📜
Error: failed to create cluster: failed to generate kubeadm config content: got file '\', but 'C:\\' must be a directory to be a root

```

![image](https://user-images.githubusercontent.com/6450081/54269241-473e3b00-457d-11e9-8db2-4439d6e16945.png)

however, is easy to fix with the [following patch](https://github.com/aojea/kind/commit/a7db747bf466a35f6afb639dc1040c5086f35012)

4. Once fixed 3 ` kind` fails with the following error .... :
```
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.13.3) 🖼
 ✓ Preparing nodes 📦
 ✓ Creating kubeadm config 📜
 ✗ Starting control-plane 🕹️
Error: failed to create cluster: failed to init node with kubeadm: exit status 1
```

and then I gave up :)

neolit123:
> however, is easy to fix with the following patch

this seems like a viable fix to me. thank you.
the rest looks like things we need to document.

/kind documentation
/priority important-longterm

BenTheElder:
So FWIW I use [docker desktop](https://www.docker.com/products/docker-desktop) and haven't had issues. It's been a while since I ran on windows which might explain the patching regression.

aojea:
[...]

PR #378, opened 2019-03-13
windows: Fix running kind: "C:\\ must be a directory to be a root"

steffengy:
When running kind on windows, "C:\\ must be a directory to be a root" in kustomize is returned as an error.
This is since `C:\\` in the docstring syntax is expanded to "C:\\\\" which is not what we want here.


aojea:
Fixes some of the issues in #377

steffengy:
@neolit123 How so? I haven't found (this) kustomize.go upstream.

neolit123:
> @neolit123 How so? I haven't found (this) kustomize.go upstream.

never mind. it's our own wrapper file.

Issue #380, opened 2019-03-13
Make kind load docker-image only load if the image is not already on the nodes


bouk:
Currently `kind load docker-image` always tars up the image and sends it, even if the nodes already contains the exact image. To make image loading faster, Kind should look at the image ID and only load over the image if the nodes don't have it yet.

BenTheElder:
So the tar loading on the node should more or less skip reading images that are already present, for large enough images skipping the tar on the host after checking may still speed up a bit though.

/kind feature
/priority important-longterm

rohansingh:
I'm not sure if this should be a separate issue, but I think `kind` should also only load layers that are not already present on any given node. Here's an example of that kind of incremental load logic:
https://github.com/bazelbuild/rules_docker/blob/master/container/incremental_load.sh.tpl

I'm using that internally to wrap `kind load image-archive` and only providing layers that are missing in a single-node cluster, and it is significantly faster when pushing changes.

rohansingh:
Here's the key insight from that script, by the way:

> Docker elides layer reads from the tarball when it already has a copy of the layer with the same basis as it has within the tarball.  This means that once we have found the lowest layer in our image of which Docker is unaware we must load all of the remaining layers.  So to determine existence, iterate through the layers attempting to load the image without it's tarball.  As soon as one fails, break and synthesize a "docker save" tarball of all of the remaining layers.

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/382 fixed this.

PR #382, opened 2019-03-13
load image only if not present

aojea:
1. Load a docker image only on those nodes that don't have it when using `kind load docker-image` 
2. Add sanity checks to `kind load` to verify that the docker-image or the tar-image exists.
3. Load concurrently the images on the nodes

https://github.com/kubernetes-sigs/kind/issues/380

### DIFF
--- a/cmd/kind/load/docker-image/docker-image.go
+++ b/cmd/kind/load/docker-image/docker-image.go
@@ -22,10 +22,12 @@ import (
 	"path/filepath"
 
 	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
 	"sigs.k8s.io/kind/pkg/cluster"
 	clusternodes "sigs.k8s.io/kind/pkg/cluster/nodes"
+	"sigs.k8s.io/kind/pkg/concurrent"
 	"sigs.k8s.io/kind/pkg/container/docker"
 	"sigs.k8s.io/kind/pkg/fs"
 )
@@ -68,6 +70,12 @@ func NewCommand() *cobra.Command {
 }
 
 func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
+	imageName := args[0]
+	// Check that the image exists locally, if not return error
+	_, err := docker.ImageInspect(imageName)
+	if err != nil {
+		return errors.Errorf("Image: %q not present locally", imageName)
+	}
 	// Check if the cluster name exists
 	known, err := cluster.IsKnown(flags.Name)
 	if err != nil {
@@ -93,16 +101,31 @@ func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
 
 	// pick only the user selected nodes and ensure they exist
 	// the default is all nodes unless flags.Nodes is set
-	selectedNodes := nodes
+	candidateNodes := nodes
 	if len(flags.Nodes) > 0 {
-		selectedNodes = []clusternodes.Node{}
+		candidateNodes = []clusternodes.Node{}
 		for _, name := range flags.Nodes {
 			node, ok := nodesByName[name]
 			if !ok {
-				return errors.Errorf("unknown node: %s", name)
+				return errors.Errorf("unknown node: %q", name)
 			}
-			selectedNodes = append(selectedNodes, node)
+			candidateNodes = append(candidateNodes, node)
 		}
+	}
+
+	// pick only the nodes that don't have the image
+	selectedNodes := []clusternodes.Node{}
+	for _, node := range candidateNodes {
+		// TODO: move under pkg/
+		cmdNode := node.Command("crictl", "-r", "/var/run/containerd/containerd.sock", "inspecti", imageName)
+		if err := cmdNode.Run(); err != nil {
+			selectedNodes = append(selectedNodes, node)
+			log.Debugf("Image: %q not present on node %q", imageName, node.String())
+		}
+	}
+
+	if len(selectedNodes) == 0 {
+		return nil
 	}
 
 	// Save the image into a tar
@@ -113,21 +136,23 @@ func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
 	defer os.RemoveAll(dir)
 	imageTarPath := filepath.Join(dir, "image.tar")
 
-	err = docker.Save(args[0], imageTarPath)
+	err = docker.Save(imageName, imageTarPath)
 	if err != nil {
 		return err
 	}
 
-	// Load the image into every node
-	// TODO(bentheelder): this should probably be concurrent
-	for _, node := range selectedNodes {
-		if err := loadImage(imageTarPath, &node); err != nil {
-			return err
-		}
+	// Load the image on the selected nodes
+	fns := []func() error{}
+	for _, selectedNode := range selectedNodes {
+		selectedNode := selectedNode // capture loop variable
+		fns = append(fns, func() error {
+			return loadImage(imageTarPath, &selectedNode)
+		})
 	}
-	return nil
+	return concurrent.UntilError(fns)
 }
 
+// loads an image tarball onto a node
 func loadImage(imageTarName string, node *clusternodes.Node) error {
 	f, err := os.Open(imageTarName)
 	if err != nil {
--- a/cmd/kind/load/docker-image/docker-image.go
+++ b/cmd/kind/load/docker-image/docker-image.go
@@ -116,9 +116,8 @@ func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
 	// pick only the nodes that don't have the image
 	selectedNodes := []clusternodes.Node{}
 	for _, node := range candidateNodes {
-		// TODO: move under pkg/
-		cmdNode := node.Command("crictl", "-r", "/var/run/containerd/containerd.sock", "inspecti", imageName)
-		if err := cmdNode.Run(); err != nil {
+		_, err := node.ImageInspect(imageName)
+		if err != nil {
 			selectedNodes = append(selectedNodes, node)
 			log.Debugf("Image: %q not present on node %q", imageName, node.String())
 		}
--- a/cmd/kind/load/image-archive/image-archive.go
+++ b/cmd/kind/load/image-archive/image-archive.go
@@ -25,6 +25,7 @@ import (
 
 	"sigs.k8s.io/kind/pkg/cluster"
 	clusternodes "sigs.k8s.io/kind/pkg/cluster/nodes"
+	"sigs.k8s.io/kind/pkg/concurrent"
 )
 
 type flagpole struct {
@@ -65,6 +66,11 @@ func NewCommand() *cobra.Command {
 }
 
 func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
+	imageTarPath := args[0]
+	// Check if file exists
+	if _, err := os.Stat(imageTarPath); err != nil {
+		return err
+	}
 	// Check if the cluster name exists
 	known, err := cluster.IsKnown(flags.Name)
 	if err != nil {
@@ -101,17 +107,18 @@ func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
 			selectedNodes = append(selectedNodes, node)
 		}
 	}
-
-	// Load the image into every node
-	// TODO(bentheelder): this should probably be concurrent
-	for _, node := range selectedNodes {
-		if err := loadImage(args[0], &node); err != nil {
-			return err
-		}
+	// Load the image on the selected nodes
+	fns := []func() error{}
+	for _, selectedNode := range selectedNodes {
+		selectedNode := selectedNode // capture loop variable
+		fns = append(fns, func() error {
+			return loadImage(imageTarPath, &selectedNode)
+		})
 	}
-	return nil
+	return concurrent.UntilError(fns)
 }
 
+// loads an image tarball onto a node
 func loadImage(imageTarName string, node *clusternodes.Node) error {
 	f, err := os.Open(imageTarName)
 	if err != nil {
--- a/pkg/cluster/nodes/node.go
+++ b/pkg/cluster/nodes/node.go
@@ -237,7 +237,7 @@ func (n *Node) WriteFile(dest, content string) error {
 // ImageInspect return low-level information on containers images inside a node
 func (n *Node) ImageInspect(containerNameOrID string) ([]string, error) {
 	cmd := n.Command(
-		"crictl", "-r", "/var/run/containerd/containerd.sock", "inspecti", containerNameOrID,
+		"crictl", "inspecti", containerNameOrID,
 	)
 	return exec.CombinedOutputLines(cmd)
 }
--- a/pkg/cluster/nodes/node.go
+++ b/pkg/cluster/nodes/node.go
@@ -234,6 +234,14 @@ func (n *Node) WriteFile(dest, content string) error {
 	return n.Command("cp", "/dev/stdin", dest).SetStdin(strings.NewReader(content)).Run()
 }
 
+// ImageInspect return low-level information on containers images inside a node
+func (n *Node) ImageInspect(containerNameOrID string) ([]string, error) {
+	cmd := n.Command(
+		"crictl", "-r", "/var/run/containerd/containerd.sock", "inspecti", containerNameOrID,
+	)
+	return exec.CombinedOutputLines(cmd)
+}
+
 // LoadImageArchive will load the image contents in the image reader to the
 // k8s.io namespace on the node such that the image can be used from a
 // Kubernetes pod
--- a/pkg/container/docker/image.go
+++ b/pkg/container/docker/image.go
@@ -1,0 +1,30 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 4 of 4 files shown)


aojea:
/hold 

I still have to test it

bouk:
I can confirm that this makes it faster! :smile:

cmd/kind/load/docker-image/docker-image.go
```diff
@@ -68,6 +69,17 @@ func NewCommand() *cobra.Command {
 }
 
 func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
+	imageName := args[0]
+	// Obtain image id from local host
+	lines, err := docker.Images(imageName)
+	if err != nil {
+		return errors.Wrap(err, "failed to get image id")
```
neolit123: `failed to get the id of image %q`

cmd/kind/load/docker-image/docker-image.go
```diff
@@ -68,6 +69,17 @@ func NewCommand() *cobra.Command {
...
+	}
+	if len(lines) != 1 {
+		return errors.Errorf("failed to get image id with image name %s", imageName)
```
neolit123: `failed to get the id of image %q`

cmd/kind/load/docker-image/docker-image.go
```diff
@@ -88,18 +100,35 @@ func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
...
 				return errors.Errorf("unknown node: %s", name)
 			}
+			candidateNodes = append(candidateNodes, node)
+		}
+	}
+
+	// pick only the nodes that don't have the image
+	selectedNodes := []clusternodes.Node{}
+	for _, node := range candidateNodes {
+		cmd := node.Command("docker", "images", "-q", imageName)
```
neolit123: we just added the `docker.Images(imageName)` utility, yet in the case of nodes we still have to manually bake a command. can these two cases use the same utility?

pkg/container/docker/images.go
```diff
@@ -0,0 +1,27 @@
...
+*/
+
+package docker
+
+import (
+	"sigs.k8s.io/kind/pkg/exec"
+)
+
+// Images returns the images ID as in docker images
+func Images(image string) ([]string, error) {
```
neolit123: possible rename here - `ImageID`:

```
// ImageID returns a slice of image IDs for a given image[:tag]
func ImageID(image string) ([]string, error) {
```

cmd/kind/load/docker-image/docker-image.go
```diff
@@ -68,6 +69,17 @@ func NewCommand() *cobra.Command {
 }
 
 func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
+	imageName := args[0]
```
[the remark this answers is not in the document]
aojea: that's checked in L44

cmd/kind/load/docker-image/docker-image.go
aojea: I was checking how it was done in the rest of the code and my understanding is that, when running commands on nodes, we always use this way.
However, if you point me to an example of using the ` docker.` utilities to run the command on nodes I'm happy to change it.

aojea:
@neolit123 I've changed the `docker.ImageID` function to return only the image ID string or fail if it doesn't exist

pkg/container/docker/images.go
```diff
@@ -0,0 +1,35 @@
...
+
+// ImageID returns the image ID for a given image[:tag]
+func ImageID(image string) (string, error) {
+	cmd := exec.Command("docker", "images", "-q", image)
+	lines, err := exec.CombinedOutputLines(cmd)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to get the id of name %q", image)
+	}
+	if len(lines) != 1 {
+		return "", errors.Errorf("failed to get the id of name %q", image)
```
BenTheElder: if we're going to return an error for this case it should be a well known / exporter error and documented so that callers can check for this not-surprising case.

alternatively, `"", nil` could signal there isn't one, since it hsouldn't be possible to have an image without an id?

pkg/container/docker/images.go
```diff
@@ -0,0 +1,35 @@
...
+package docker
+
+import (
+	"github.com/pkg/errors"
+	"sigs.k8s.io/kind/pkg/exec"
+)
+
+// ImageID returns the image ID for a given image[:tag]
+func ImageID(image string) (string, error) {
+	cmd := exec.Command("docker", "images", "-q", image)
```
BenTheElder: we should document that we are relying on this behavior when multiple IDs are returned:
```
$ docker images -q kindest/node
2f48f23e3589
f30a33de05b8
f934a127ad69
4a61fa22c039
4a61fa22c039
2d69a822937a
67409709c324
0d16a995491a
```

```
$ docker images -q kindest/node:latest
2f48f23e3589
```

aojea: I think that's better to default to `latest` if not tag is specified, at least is what I will expect as an user. Let me change it

pkg/container/docker/images.go
aojea: you can pass a  non-existent image name, I think we need to fail in that case.

```
./kind load docker-image nonexistanteiamge
Error: There is no image id with name "nonexistanteiamge:latest"
```

what do you mean by this? 

> well known / exporter error and documented

pkg/container/docker/images.go
aojea: and seems that's the logic implemented in `docker image inspect` 
```

# docker images weaveworks/weave-npc
REPOSITORY             TAG                 IMAGE ID            CREATED             SIZE
weaveworks/weave-npc   2.5.1               789b7f496034        8 weeks ago         49.6MB
# docker image inspect weaveworks/weave-npc
[]
Error: No such image: weaveworks/weave-npc
```

aojea:
@BenTheElder is this ok or do you want me to rework something?

pkg/container/docker/images.go
BenTheElder: what i mean is, we depend on the ordering of the output of by `docker images`

BenTheElder: we should at least put a comment at the call site about this

cmd/kind/load/docker-image/docker-image.go
```diff
@@ -88,18 +96,35 @@ func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
 
 	// pick only the user selected nodes and ensure they exist
 	// the default is all nodes unless flags.Nodes is set
-	selectedNodes := nodes
+	candidateNodes := nodes
 	if len(flags.Nodes) > 0 {
-		selectedNodes := []clusternodes.Node{}
+		candidateNodes := []clusternodes.Node{}
```
BenTheElder: this has the same bug as the old code, this should be `=` not `:=`

cmd/kind/load/docker-image/docker-image.go
BenTheElder: we should add a `node.Images()` utility to list images available to kubernetes. this should use `ctr --namespace=k8s.io images list`

aojea: I'm afraid that we have to use ` crictl -r /var/run/containerd/containerd.sock` instead of 
`ctr --namespace=k8s.io images` , I couldn't find same hashes or ids between `docker` and `ctr`

pkg/container/docker/images.go
aojea: this no longer applies

cmd/kind/load/docker-image/docker-image.go
aojea: done

cmd/kind/load/docker-image/docker-image.go
```diff
@@ -88,18 +98,37 @@ func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
...
+
+	// pick only the nodes that don't have the image
+	selectedNodes := []clusternodes.Node{}
+	for _, node := range candidateNodes {
+		// Don't load image on external nodes like the load balancer
+		if strings.Contains(node.String(), "external") {
+			continue
+		}
+
+		cmdNode := node.Command("crictl", "-r", "/var/run/containerd/containerd.sock", "inspecti", imageName)
```
aojea: is this ok?

cmd/kind/load/docker-image/docker-image.go
```diff
@@ -68,6 +72,12 @@ func NewCommand() *cobra.Command {
 }
 
 func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
+	imageName := args[0]
+	// Check that the image exists locally, if not return error
+	cmdDocker := exec.Command("docker", "inspect", "--type=image", imageName)
```
BenTheElder: This logic should be in the docker package

cmd/kind/load/docker-image/docker-image.go
BenTheElder: Put a TODO to move it out at least, cmd/ should only have application boilerplate, pkg/... should be providing this logic

cmd/kind/load/image-archive/image-archive.go
```diff
@@ -96,17 +103,50 @@ func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
...
+// loadImageOnNodes loads concurrently an image tarball in a group of nodes
+func loadImageOnNodes(imageTarName string, nodes []clusternodes.Node) error {
+	// Load the image into every node concurrently
+	// create all of the node containers, concurrently
+	errChan := make(chan error, len(nodes))
+	var wg sync.WaitGroup
+	for _, node := range nodes {
+		node := node // capture loop variable
+		// Don't load image on external nodes like the load balancer
+		if strings.Contains(node.String(), "external") {
```
BenTheElder: this is a hack that will bite us later. Should be using the node role

aojea: done

cmd/kind/load/docker-image/docker-image.go
aojea: done

cmd/kind/load/docker-image/docker-image.go
aojea: done

cmd/kind/load/docker-image/docker-image.go
BenTheElder: using crictl and CRI in general where possible is ideal, mainly we just needed to not use `docker`, we should check the ctr output though at some point ourselves just as a sanity check, using crictl here is ideal.

cmd/kind/load/image-archive/image-archive.go
```diff
@@ -96,17 +103,54 @@ func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
...
+	go func() {
+		defer close(errChan)
+		wg.Wait()
+	}()
 
-	// Load the image into every node
-	// TODO(bentheelder): this should probably be concurrent
-	for _, node := range selectedNodes {
-		if err := loadImage(args[0], &node); err != nil {
+	// return the first error encountered if any
```
BenTheElder: we've passed [WET](https://dev.to/wuz/stop-trying-to-be-so-dry-instead-write-everything-twice-wet-5g33), sending a PR to de-dupe this logic finally :-)

BenTheElder: https://github.com/kubernetes-sigs/kind/pull/481

aojea:
/hold

wait until #481 and use the new concurrent pkg

cmd/kind/load/docker-image/docker-image.go
```diff
@@ -68,6 +70,13 @@ func NewCommand() *cobra.Command {
 }
 
 func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
+	imageName := args[0]
+	// Check that the image exists locally, if not return error
+	// {{.Os}} seems to be an attribute specific to images only
```
BenTheElder: seems a bit brittle, I wonder if there's a better option... 🤔

BenTheElder: https://docs.docker.com/engine/reference/commandline/image_inspect/ ?

tao12345666333: It should be possible to use `docker image inspect` or `docker image ls`. In terms of meaning, `docker image ls` should be more appropriate.

cmd/kind/load/docker-image/docker-image.go
```diff
@@ -93,18 +102,33 @@ func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
...
 			}
+			candidateNodes = append(candidateNodes, node)
+		}
+	}
+
+	// pick only the nodes that don't have the image
+	selectedNodes := []clusternodes.Node{}
+	for _, node := range candidateNodes {
+		// TODO: move under pkg/
+		cmdNode := node.Command("crictl", "-r", "/var/run/containerd/containerd.sock", "inspecti", imageName)
```
aojea: @BenTheElder I'll move this under pkg, what name do you prefer pkg/container/cri?

cmd/kind/load/docker-image/docker-image.go
aojea: `docker image ls` doesn´t discriminate by tag, I find more appropriate the behavior of `image inspect` that only returns the matching image with the tag included

cmd/kind/load/docker-image/docker-image.go
aojea: moved to a node method

cmd/kind/load/docker-image/docker-image.go
tao12345666333: > `docker image ls` doesn´t discriminate by tag,

It can be distinguished by tag. i.e.

```
(MoeLove) ➜  ~ docker image ls debian
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
debian              latest              8d31923452f8        5 weeks ago         101MB
debian              stretch             de8b49d4b0b3        5 months ago        101MB
(MoeLove) ➜  ~ docker image ls debian:stretch
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
debian              stretch             de8b49d4b0b3        5 months ago        101MB
(MoeLove) ➜  ~ docker image ls debian:latest 
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
debian              latest              8d31923452f8        5 weeks ago         101MB
```

or

```
(MoeLove) ➜  ~ docker image ls -f reference=debian
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
debian              latest              8d31923452f8        5 weeks ago         101MB
debian              stretch             de8b49d4b0b3        5 months ago        101MB
(MoeLove) ➜  ~ docker image ls -f reference=debian:stretch
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
debian              stretch             de8b49d4b0b3        5 months ago        101MB
(MoeLove) ➜  ~ docker image ls -f reference=debian:latest 
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
debian              latest              8d31923452f8        5 weeks ago         101MB

```

aojea: @tao12345666333 my bad 😅 ,  I didn't explain it correctly, my point was that the user should expect to load only one image, and use ` latest` in absence of any tag, similar to another docker commands

``` 
docker image ls nginx
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
nginx               1.17-alpine         bfba26ca350c        7 days ago          20.5MB
nginx               1.15.12-alpine      0be75340bd9b        8 weeks ago         16.1MB
nginx               1.14-alpine         cafef9fe265b        3 months ago        16MB
```

pkg/cluster/nodes/node.go
```diff
@@ -234,6 +234,14 @@ func (n *Node) WriteFile(dest, content string) error {
 	return n.Command("cp", "/dev/stdin", dest).SetStdin(strings.NewReader(content)).Run()
 }
 
+// ImageInspect return low-level information on containers images inside a node
+func (n *Node) ImageInspect(containerNameOrID string) ([]string, error) {
+	cmd := n.Command(
+		"crictl", "-r", "/var/run/containerd/containerd.sock", "inspecti", containerNameOrID,
```
BenTheElder: we don't need to set the socket, kind images have included this pre-configured for a while. not specifying it buys us free future {insert-cri-here} support for this feature

```
$ docker exec kind-control-plane crictl ps
CONTAINER ID        IMAGE               CREATED             STATE               NAME                      ATTEMPT             POD ID
33f1589cc2129       b3b94275d97cb       22 hours ago        Running             coredns                   0                   7ef4face252c1
c9a518029909c       b3b94275d97cb       22 hours ago        Running             coredns                   0                   2863336690554
a7a5abebcf708       f227066bdc5f9       22 hours ago        Running             kindnet-cni               0                   7b5c8553efd43
04ecc7a8c3c30       19bb968f77bba       22 hours ago        Running             ip-masq-agent             0                   f28561ce51146
6efcff288f6b7       6d859c5ba0871       22 hours ago        Running             kube-proxy                0                   50869b3535504
176dbdf4dd874       b8df3b177be23       22 hours ago        Running             etcd                      0                   dde1d7249abb4
69b32e5f5e6dd       b2e25e147ed1c       22 hours ago        Running             kube-apiserver            0                   7c095b1397162
30fc0b2cf24e6       b91e0e16786e3       22 hours ago        Running             kube-scheduler            0                   326d485d5cb91
081b277edb8ef       3eb89ae87afc9       22 hours ago        Running             kube-controller-manager   0                   12df50441fdbf
```

aojea: done

Issue #266, opened 2019-02-04
document how to use HA


neolit123:
mostly document how to to use the config for HA at this point.
/assign
/kind documentation
/priority important-soon

neolit123:
will add a small config example for HA somewhere under here, once alpha3 is done:
https://kind.sigs.k8s.io/docs/user/quick-start/#configuring-your-kind-cluster

tao12345666333:
#373 has been merged. :+1:  
Maybe we can add config example for HA.

neolit123:
will do tomorrow. :+1:

tao12345666333:
> EDIT: i suggest we add this to the 0.2.0 milestone?

+1

PR #384, opened 2019-03-14
site: add examples for v1alpha3

neolit123:
- add minimal multi CP example
- update the example YAML to v1alpha3
- add API redirect

fixes https://github.com/kubernetes-sigs/kind/issues/266

/kind documentation
/priority important-longterm

### DIFF
--- a/site/content/docs/user/kind-example-config.yaml
+++ b/site/content/docs/user/kind-example-config.yaml
@@ -1,30 +1,29 @@
 # this config file contains all config fields with comments
-kind: Config
-apiVersion: kind.sigs.k8s.io/v1alpha2
+kind: Cluster
+apiVersion: kind.sigs.k8s.io/v1alpha3
+# patch the generated kubeadm config with some extra settings
+kubeadmConfigPatches:
+- |
+  apiVersion: kubeadm.k8s.io/v1beta1
+  kind: ClusterConfiguration
+  metadata:
+    name: config
+  networking:
+    serviceSubnet: 10.0.0.0/16
+# patch it further using a JSON 6902 patch
+kubeadmConfigPatchesJson6902:
+- group: kubeadm.k8s.io
+  version: v1beta1
+  kind: ClusterConfiguration
+  patch: |
+    - op: add
+      path: /apiServer/certSANs/-
+      value: my-hostname
 # 1 control plane node and 3 workers
 nodes:
 # the control plane node config
 - role: control-plane
-  # patch the generated kubeadm config with some extra settings
-  kubeadmConfigPatches:
-  - |
-    apiVersion: kubeadm.k8s.io/v1beta1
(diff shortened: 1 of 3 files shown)


site/content/docs/user/kind-example-config.yaml
```diff
@@ -1,30 +1,29 @@
...
+kind: Cluster
+apiVersion: kind.sigs.k8s.io/v1alpha3
+# patch the generated kubeadm config with some extra settings
+kubeadmConfigPatches:
+- |
+  apiVersion: kubeadm.k8s.io/v1beta1
+  kind: ClusterConfiguration
+  metadata:
+    name: config
+  networking:
```
aojea: v1alpha3 is going to have a new `networking` field #340 .

I've sent a proposal (#348)  using ` serviceSubnet` and `podSubnet` in the new `networking` section.

Does it make sense?

neolit123: depends on what we want in the kind example config file(s).

site/content/docs/user/quick-start.md
```diff
@@ -202,12 +202,26 @@ In particular, many users may be interested in multi-node clusters. A simple
 configuration for this can be achieved with the following config file contents:
 ```yaml
 # three node (two workers) cluster config
```
BenTheElder: this comment doesn't match now

neolit123: hm, it says 3 nodes total, with 2 workers.
```
# three node (two workers) cluster config
kind: Cluster
nodes:
- role: control-plane
- role: worker
- role: worker
```

should this be changed?

BenTheElder: Derp, thanks

site/content/docs/user/kind-example-config.yaml
BenTheElder: Let's only add what is merged for now with the API server options and come back to discussing these other options (next meeting / 0.3 planning perhaps?)

neolit123: ok, i'd prefer if we talk about this PR the next meeting and see if this config is what we want for the next release.

BenTheElder: SGTM, 0.2 is about ready to go, would be good to document v1alpha3 for that and add in these extra fields later.

BenTheElder: .2 mostly needs relnotes and https://github.com/kubernetes-sigs/kind/pull/394 https://github.com/kubernetes-sigs/kind/pull/393

neolit123: so is this PR good for 0.2?

BenTheElder:
let's add our networking field in a follow-up

PR #390, opened 2019-03-19
add v1alpha3 network options

BenTheElder:
fixes https://github.com/kubernetes-sigs/kind/issues/340

- v1alpha2 will continue to bind to 0.0.0.0 by default (as is implicit if not specified to docker)
- v1alpha3 will bind to 127.0.0.1 by default, but has a knob to override it in `Cluster.Network`
- v1alpha3 can also set a non-random api server port on the host, the default is random

this change is not quite as clean as I hoped, I have a more complex follow-up to add cri.PortMapping internally, but that shouldn't affect the external behavior and 0.2.0 is behind.


pkg/cluster/config/types.go
```diff
@@ -38,6 +38,9 @@ type Cluster struct {
 
 	/* Advanced fields */
 
+	// Network contains cluster wide network settings
+	Network Network
```
neolit123: optionally this can be?
Network -> Networking

kubeadm:
https://godoc.org/k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm/v1beta1#Networking

docker:
https://docs.docker.com/config/containers/container-networking/

pkg/cluster/config/v1alpha3/types.go
```diff
@@ -82,3 +85,14 @@ const (
...
+
+// Network contains cluster wide network settings
+type Network struct {
+	// APIServerPort is the listen port on the host for the Kubernetes API Server
+	// Defaults to a random port on the host
+	APIServerPort int32 `json:"apiServerHostPort,omitempty"`
+	// APIServerListenAddress is the listen address on the host for the Kubernetes API Server
+	// This should be an IP address
+	// Defaults to 127.0.0.1
+	APIServerListenAddress string `json:"apiServerAddress,omitempty"`
```
neolit123: my preference would be to match the json tag names and the go field names.

also possibly use `Host` for both?
```
APIServerHostAddress `json:"apiServerHostAddress,omitempty"`
APIServerHostPort `json:"apiServerHostPort,omitempty"`
```

or use `Listen` for both.

but not strong opinions as long as the comments explain the fields.

here is the naming in kubeadm:
https://godoc.org/k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm/v1beta1#APIEndpoint

but that's a bit different as we try to map to the kube-apiserver CLI flags:
https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/

BenTheElder: +1 to making them match json, this field is a bit different as well because it's a host mapping to the actual one, the kubeadm config value should NOT match this value in most cases.

pkg/cluster/config/types.go
BenTheElder: sure, no strong opinions on network vs networking here.

pkg/cluster/config/v1alpha3/types.go
```diff
@@ -82,3 +85,15 @@ const (
 	// WorkerRole identifies a node that hosts a Kubernetes worker
 	WorkerRole NodeRole = "worker"
 )
+
+// Networking contains cluster wide network settings
+type Networking struct {
+	// APIServerPort is the listen port on the host for the Kubernetes API Server
+	// Defaults to a random port on the host
+	APIServerPort int32 `json:"apiServerHostPort,omitempty"`
```
neolit123: `json: apiServerPort` to match?

pkg/cluster/config/types.go
BenTheElder: amusingly s/Network/Networking broke finding the ports because it's in NetworkSettings

pkg/cluster/config/v1alpha3/types.go
BenTheElder: done

pkg/cluster/config/v1alpha3/types.go
BenTheElder: done

pkg/cluster/config/types.go
BenTheElder: done

pkg/cluster/nodes/node.go
```diff
@@ -265,7 +265,7 @@ func (n *Node) IP() (ip string, err error) {
 		return cachedIP, nil
 	}
 	// retrive the IP address of the node using docker inspect
-	lines, err := docker.Inspect(n.name, "{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}")
+	lines, err := docker.Inspect(n.name, "{{range .NetworkSettings.Network}}{{.IPAddress}}{{end}}")
```
BenTheElder: gah, this is still wrong. find and replace was a mistake

neolit123:
/lgtm
tested with a 3CP + 3W setup.

pkg/cluster/internal/create/nodes.go
```diff
@@ -206,11 +206,26 @@ func nodesToCreate(cfg *config.Cluster, clusterName string) []nodeSpec {
...
+
+	// add all of the config nodes as desired nodes
+	for _, configNode := range configNodes {
+		role := string(configNode.Role)
+		apiServerPort := cfg.Networking.APIServerPort
+		apiServerAddress := cfg.Networking.APIServerAddress
+		// only the external LB should reflect the port if we have
+		// multiple control planes
+		if isHA && role != constants.ExternalLoadBalancerNodeRoleValue {
+			apiServerPort = 0
```
BenTheElder: so actually this should also probably set `apiServerAddress` to `127.0.0.1`, only the single CP or LB needs to be listening on the configured addr and port, the rest should be local + random 🤔

neolit123: hm, i think you are right about this.
cc @fabriziopandini 

btw, `isHA` should probably be `hasMultipleCP` or something like that.
because HA is a bit ambiguous, e.g. HA is at least 3 CPs, behind a LB.

BenTheElder: or really `wantLoadBalancer`, since we may eventually unblock an option to have multiple CP but no LB at some point, we'll need to think these things through for future config options

Issue #391, opened 2019-03-20
update CNI version


aojea:
Seems that[ kubernetes bump the CNI version](https://github.com/kubernetes/kubernetes/pull/75455/files) to [0.7.5 ](https://github.com/containernetworking/plugins/releases )to deal with some portmap bugs, it also fixes the Docker IPv6 problem.

I think that we should do the same https://github.com/kubernetes-sigs/kind/blob/0c1210661b5c796eb8953d5dc3b158661e6236cb/images/base/Dockerfile#L80

BenTheElder:
any breaking changes for older kubernetes?

aojea:
this update installs the latest CNI plugins, it doesn't affect Kubernetes per se, only the network plugins that use the standard CNI plugins.
In the `kind` case it's good because `weave` seems to use the standard ` portmap` plugin and this version address some issues with it

PR #392, opened 2019-03-20
Bump CNI version to 0.7.5

aojea:
Update kind to use the CNI plugins version 0.7.5 per https://github.com/kubernetes/kubernetes/pull/75455

Refernce: #391


neolit123:
the kubeadm debs/rpms have not done this yet.
i'm messaging people right now if the packages should do this or not.

tao12345666333:
When you update the base/Dockerfile, you need to run hack/update-generated. :)

aojea:
ahhg, my bad, I always forgot that part :(

pkg/build/node/node.go
```diff
@@ -22,10 +22,11 @@ import (
 	"os"
 	"path"
 	"regexp"
-	"sigs.k8s.io/kind/pkg/util"
 	"strings"
 	"time"
 
+	"sigs.k8s.io/kind/pkg/util"
```
BenTheElder: aside:
will re-arrange this later, imho this is a more correct place, but still not the most, it should be in the bottom block with our own imports:
- stdlib
- third party
- kubernetes
- kind

import grouping might be something to doc somewhere, I wonder if any of kubernetes's guides cover this...

PR #396, opened 2019-03-21
Build images from tarballs

### LINKED ISSUE #381: support building node images from release tarballs
We should implement `kind build node-image --type=release-tar /path/to/release.tar` and use this for CI testing to dedupe building the binaries etc.

/kind feature
/priority important-soon
/assign

akutz:
Hi all, 

I'd love to get some feedback on this. I have a need for it, and I discussed it briefly with @BenTheElder on Slack. I'm aware others have ideas about it, and I'd love to kick off the discussion with this PR. Thanks all!

This PR introduces support for building images from tarballs and fixes #381. For example:

```shell
$ KIND_TARBITS=$HOME/.kind/tarbits GOOS=linux ./kind build node-image --type tar --kube-root ci/latest --image kindest/node:ci-latest
Building version v1.15.0-alpha.0.1381+8af1bf313efeea
Created symlink /etc/systemd/system/multi-user.target.wants/kubelet.service → /kind/systemd/kubelet.service.
Created symlink /etc/systemd/system/kubelet.service → /kind/systemd/kubelet.service.
Pulling: k8s.gcr.io/pause:3.1
Pulling: k8s.gcr.io/etcd:3.3.10
Pulling: k8s.gcr.io/coredns:1.3.1
Pulling: weaveworks/weave-kube:2.5.1
Pulling: weaveworks/weave-npc:2.5.1
sha256:99649ee63ff408e72c0261858c9a13edc17a5701f040e9167ab84606dede10a9
```

```shell
$ ls $HOME/.kind/tarbits/v1.15.0-alpha.0.1437+b3824cd094f73d
total 24
drwxr-xr-x  6 akutz  staff   192B Mar 22 14:36 ./
drwxr-xr-x  6 akutz  staff   192B Mar 22 14:41 ../
-rw-r--r--  1 akutz  staff   898B Mar 22 14:36 10-kubeadm.conf
drwxr-xr-x  3 akutz  staff    96B Mar 22 14:35 bin/
-rw-r--r--  1 akutz  staff   227B Mar 22 14:36 kubelet.service
-rw-r--r--  1 akutz  staff    35B Mar 22 14:35 version
```

```shell
$ ls $HOME/.kind/tarbits/v1.15.0-alpha.0.1437+b3824cd094f73d/bin/linux/amd64
total 1514816
drwxr-xr-x  9 akutz  staff   288B Mar 22 14:36 ./
drwxr-xr-x  3 akutz  staff    96B Mar 22 14:35 ../
-rw-r--r--  1 akutz  staff   201M Mar 22 14:35 kube-apiserver.tar
-rw-r--r--  1 akutz  staff   152M Mar 22 14:36 kube-controller-manager.tar
-rw-r--r--  1 akutz  staff    80M Mar 22 14:36 kube-proxy.tar
-rw-r--r--  1 akutz  staff    79M Mar 22 14:36 kube-scheduler.tar
-rwxr-xr-x  1 akutz  staff    38M Mar 22 14:36 kubeadm*
-rwxr-xr-x  1 akutz  staff    41M Mar 22 14:35 kubectl*
-rwxr-xr-x  1 akutz  staff   122M Mar 22 14:35 kubelet*
```

```shell
$ kind create cluster --image kindest/node:ci-latest
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:ci-latest) 🖼
 ✓ Preparing nodes 📦 
 ✓ Creating kubeadm config 📜 
 ✓ Starting control-plane 🕹️ 

kubectl cluster-info
```

```
$ kubectl --kubeconfig $(kind get kubeconfig-path --name="kind") version
Client Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.3", GitCommit:"a4529464e4629c21224b3d52edfe0ea91b072862", GitTreeState:"clean", BuildDate:"2018-09-09T18:02:47Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"15+", GitVersion:"v1.15.0-alpha.0.1437+b3824cd094f73d", GitCommit:"8af1bf313efeeafb94c01ce04965ad8fe9cea21a", GitTreeState:"clean", BuildDate:"2019-03-21T03:35:37Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"linux/amd64"}
```

The `-type` is `tar` and `-kube-root` may be set to one of three values:

1. A local filesystem path containing the files listed above structure that matches the file structure of the public, GCS buckets.
2. An HTTP address that contains the files listed above in a structure that adheres to the layout of the public GCS buckets kubernetes-release and kubernetes-release-dev. For example, if tarRoot is set to https://k8s.ci/v1.13 then the following URIs should be valid:
    * https://k8s.ci/v1.13/kubernetes.tar.gz
    * https://k8s.ci/v1.13/bin/GOOS/GOARCH/kube-apiserver.tar
    * https://k8s.ci/v1.13/bin/GOOS/GOARCH/kube-controller-manager.tar
    * https://k8s.ci/v1.13/bin/GOOS/GOARCH/kube-scheduler.tar
    * https://k8s.ci/v1.13/bin/GOOS/GOARCH/kube-proxy.tar
    * https://k8s.ci/v1.13/bin/GOOS/GOARCH/kubeadm
    * https://k8s.ci/v1.13/bin/GOOS/GOARCH/kubectl
    * https://k8s.ci/v1.13/bin/GOOS/GOARCH/kubelet

3. A valid semantic version for a released version of Kubernetes or begins with `ci/` or `release/`. If the string matches any of these then the value is presumed to be a CI or release build hosted on one of the public GCS buckets. 

    This option also supports values ended with `.txt`, ex. `ci/latest.txt`. In fact, all values beginning with `ci/` or `release/` are first checked to see if there's a matching txt file for that value. For example, if `ci/latest` is provided then before assuming that is a directory, `ci/latest.txt` is queried.

Some additional notes:
* In order to prevent the second two options from conflicting with the first, a local file path, using the prefix `file://` explicitly indicates the given tarRoot is a local file path.
* If no explicit local path is provided then remote files are downloaded to a `NEW_TEMP_DIR/VERSION/GOOS/GOARCH` where `VERSION` is the value extracted from the "version" file in the `kubernetes.tar.gz` tarball. If `$KIND_TARBITS` is defined it will be used instead of a new temporary directory.
* If a file is already present on disk with the same size as the remote equivalent, the file will not be downloaded again.

Finally, per @BenTheElder's `TODO`:
> // TODO(bentheelder): probably we should use our own config instead :-)

The files `kubelet.service` and `10-kubeadm.conf` are written to disk by `TarBits` from string constants in `pkg/build/kube/files.go`

## TODO
1. I'm almost positive I did the wrong thing with the `version` file. I realize that the existing version logic uses extended information from the Git version, but that isn't readily available (or rather readily apparent to me if it **is** available) in the tar files.

### DIFF
--- a/pkg/build/kube/files.go
+++ b/pkg/build/kube/files.go
@@ -1,0 +1,84 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kube
+
+import (
+	"io/ioutil"
+	"net/http"
+
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+)
+
+const kubeletServiceURI = `https://raw.githubusercontent.com/kubernetes/release/master/debian/xenial/kubelet/lib/systemd/system/kubelet.service`
+
+const kubeletService = `[Unit]
+Description=kubelet: The Kubernetes Node Agent
+Documentation=http://kubernetes.io/docs/
+
+[Service]
+ExecStart=/usr/bin/kubelet
+Restart=always
+StartLimitInterval=0
+RestartSec=10
+
+[Install]
+WantedBy=multi-user.target
+`
+
+const tenKubeadmConfURI = `https://raw.githubusercontent.com/kubernetes/release/master/debian/xenial/kubeadm/channel/stable/etc/systemd/system/kubelet.service.d/post-1.10/10-kubeadm.conf`
+
+const tenKubeadmConf = `# Note: This dropin only works with kubeadm and kubelet v1.11+
+[Service]
+Environment="KUBELET_KUBECONFIG_ARGS=--bootstrap-kubeconfig=/etc/kubernetes/bootstrap-kubelet.conf --kubeconfig=/etc/kubernetes/kubelet.conf"
+Environment="KUBELET_CONFIG_ARGS=--config=/var/lib/kubelet/config.yaml"
+# This is a file that "kubeadm init" and "kubeadm join" generates at runtime, populating the KUBELET_KUBEADM_ARGS variable dynamically
+EnvironmentFile=-/var/lib/kubelet/kubeadm-flags.env
+# This is a file that the user can use for overrides of the kubelet args as a last resort. Preferably, the user should use
+# the .NodeRegistration.KubeletExtraArgs object in the configuration files instead. KUBELET_EXTRA_ARGS should be sourced from this file.
+EnvironmentFile=-/etc/default/kubelet
+ExecStart=
+ExecStart=/usr/bin/kubelet $KUBELET_KUBECONFIG_ARGS $KUBELET_CONFIG_ARGS $KUBELET_KUBEADM_ARGS $KUBELET_EXTRA_ARGS
+`
+
+func getKubeletServiceBytes() []byte {
+	return getRemoteOrDefaultBytes(kubeletServiceURI, kubeletService)
+}
+
+func getTenKubeadmConfBytes() []byte {
+	return getRemoteOrDefaultBytes(tenKubeadmConfURI, tenKubeadmConf)
+}
+
+func getRemoteOrDefaultBytes(uri, defaultContent string) []byte {
+	resp, err := http.Get(uri)
+	if err == nil && resp.StatusCode != 200 {
+		err = errors.Errorf("%s", resp.Status)
+	}
+	if err != nil {
+		log.WithError(err).Debugf(
+			"error reading %s; using default content", uri)
+		return []byte(defaultContent)
+	}
+	defer resp.Body.Close()
+	buf, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		log.WithError(err).Debugf(
+			"error reading %s; using default content", uri)
+		return []byte(defaultContent)
+	}
+	return buf
+}
--- a/pkg/build/kube/tarbits.go
+++ b/pkg/build/kube/tarbits.go
@@ -62,6 +62,21 @@ func init() {
 	RegisterNamedBits("tar", NewTarBits)
 }
 
+// TarBitsExtract extracts bits for a given buildID to the given extractPath.
+// Please see the NewTarBits and its explanation of "kubeRoot" for information
+// on valid buildID values.
+func TarBitsExtract(buildID, extractPath string) error {
+	bits, err := NewTarBits(buildID)
+	if err != nil {
+		return err
+	}
+	tarBits := bits.(*TarBits)
+	if _, err := tarBits.extract(tarBits.kubeRoot, extractPath); err != nil {
+		return err
+	}
+	return nil
+}
+
 // NewTarBits returns a new Bits backed by a version file, the release image
 // archives and the kubeadm, kubectl, and kubelet binaries. The given kubeRoot
 // may be defined three ways:
--- a/pkg/build/kube/tarbits.go
+++ b/pkg/build/kube/tarbits.go
@@ -155,6 +155,14 @@ func (b *TarBits) Build() error {
 	return nil
 }
 
+// Clean implements the Bits.Clean interface.
+func (b *TarBits) Clean() {
+	if b.tempData {
+		os.RemoveAll(b.dataPath)
+		log.WithField("path", b.dataPath).Debug("Removed temporary files")
+	}
+}
+
 // Paths implements Bits.Paths
 func (b *TarBits) Paths() map[string]string {
 	return b.paths
--- a/pkg/build/kube/tarbits.go
+++ b/pkg/build/kube/tarbits.go
@@ -1,0 +1,543 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kube
+
+import (
+	"archive/tar"
+	"bytes"
+	"compress/gzip"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"os"
+	"path"
+	"strings"
+
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+	apimv "k8s.io/apimachinery/pkg/util/version"
+
+	"sigs.k8s.io/kind/pkg/util"
+)
+
+const (
+	tarBitsCIBuildURI      = "https://storage.googleapis.com/kubernetes-release-dev"
+	tarBitsReleaseBuildURI = "https://storage.googleapis.com/kubernetes-release"
+)
+
+// TarBitsExtractorFunc is the signature for functions that are able to extract
+// the bits from a source location given by src to a destination location given
+// by dst. The map returned should adhere to the same key/value rules as the
+// map returned by Bits.Paths.
+type TarBitsExtractorFunc func(src, dst string) (map[string]string, error)
+
+// TarBits implements Bits for the tarballs.
+type TarBits struct {
+	dataPath string
+	extract  TarBitsExtractorFunc
+	kubeRoot string
+	paths    map[string]string
+	tempData bool
+}
+
+var _ Bits = &TarBits{}
+
+func init() {
+	RegisterNamedBits("tar", NewTarBits)
+}
+
+// NewTarBits returns a new Bits backed by a version file, the release image
+// archives and the kubeadm, kubectl, and kubelet binaries. The given kubeRoot
+// may be defined three ways:
+//
+//   1. A local filesystem path containing the files listed above in a flat
+//      structure.
+//
+//   2. An HTTP address that contains the files listed above in a structure
+//      that adheres to the layout of the public GCS buckets kubernetes-release
+//      and kubernetes-release-dev. For example, if kubeRoot is set to
+//      https://k8s.ci/v1.13 then the following URIs should be valid:
+//
+//        * https://k8s.ci/v1.13/kubernetes.tar.gz
+//        * https://k8s.ci/v1.13/bin/GOOS/GOARCH/kube-apiserver.tar
+//        * https://k8s.ci/v1.13/bin/GOOS/GOARCH/kube-controller-manager.tar
+//        * https://k8s.ci/v1.13/bin/GOOS/GOARCH/kube-scheduler.tar
+//        * https://k8s.ci/v1.13/bin/GOOS/GOARCH/kube-proxy.tar
+//        * https://k8s.ci/v1.13/bin/GOOS/GOARCH/kubeadm
+//        * https://k8s.ci/v1.13/bin/GOOS/GOARCH/kubectl
+//        * https://k8s.ci/v1.13/bin/GOOS/GOARCH/kubelet
+//
+//   3. A valid semantic version for a released version of Kubernetes or
+//      begins with "ci/" or "release/". If the string matches any of these
+//      then the value is presumed to be a CI or release build hosted on one
+//      of the public GCS buckets.
+//
+//      This option also supports values ended with ".txt", ex. "ci/latest.txt".
+//      In fact, all values beginning with "ci/" or "release/" are first checked
+//      to see if there's a matching txt file for that value. For example,
+//      if "ci/latest" is provided then before assuming that is a directory,
+//      "ci/latest.txt" is queried.
+//
+// In order to prevent the second two options from conflicting with the first,
+// a local file path, using the prefix "file://" explicitly indicates the given
+// kubeRoot is a local file path.
+func NewTarBits(kubeRoot string) (bits Bits, err error) {
+	tarBits := &TarBits{
+		kubeRoot: kubeRoot,
+	}
+
+	if strings.HasPrefix(kubeRoot, "file://") {
+		tarBits.dataPath = kubeRoot
+		tarBits.extract = tarBits.extractFromLocalDir
+	} else if strings.HasPrefix(kubeRoot, "http:") ||
+		strings.HasPrefix(kubeRoot, "https:") {
+		tarBits.extract = tarBits.extractFromHTTP
+	} else if _, err := apimv.ParseGeneric(kubeRoot); err == nil {
+		tarBits.extract = tarBits.extractFromSemVer
+	} else if strings.HasPrefix(kubeRoot, "ci/") {
+		tarBits.extract = tarBits.extractFromCIBuild
+	} else if strings.HasPrefix(kubeRoot, "release/") {
+		tarBits.extract = tarBits.extractFromReleaseBuild
+	} else {
+		tarBits.dataPath = kubeRoot
+		tarBits.extract = tarBits.extractFromLocalDir
+	}
+
+	// tarBits.dataPath is the root directory to which the bits are extracted.
+	// If a local extractor is used then the data path is the same as the
+	// given kubeRoot.
+	//
+	// Otherwise dataPath may be defined explicitly via $KIND_TARBITS.
+	//
+	// If neither the above is true then dataPath is set to a temp directory
+	// and its bits are removed after installation.
+	if tarBits.dataPath == "" {
+		tarBits.dataPath = os.Getenv("KIND_TARBITS")
+		if tarBits.dataPath != "" {
+			os.MkdirAll(tarBits.dataPath, 0755)
+		} else {
+			tempDir, err := ioutil.TempDir("", "")
+			if err != nil {
+				return nil, errors.Wrap(err, "error creating tarbits temp dir")
+			}
+			tarBits.tempData = true
+			tarBits.dataPath = tempDir
+		}
+	}
+
+	return tarBits, nil
+}
+
+// Build implements Bits.Build
+func (b *TarBits) Build() error {
+	paths, err := b.extract(b.kubeRoot, b.dataPath)
+	if err != nil {
+		return errors.Wrapf(
+			err, "error extracting bits from %s to %s", b.kubeRoot, b.dataPath)
+	}
+	b.paths = paths
+	return nil
+}
+
+// Paths implements Bits.Paths
+func (b *TarBits) Paths() map[string]string {
+	return b.paths
+}
+
+// Install implements Bits.Install
+func (b *TarBits) Install(install InstallContext) error {
+	kindBinDir := path.Join(install.BasePath(), "bin")
+
+	// symlink the kubernetes binaries into $PATH
+	binaries := []string{"kubeadm", "kubelet", "kubectl"}
+	for _, binary := range binaries {
+		if err := install.Run("ln", "-s",
+			path.Join(kindBinDir, binary),
+			path.Join("/usr/bin/", binary),
+		); err != nil {
+			return errors.Wrap(err, "failed to symlink binaries")
+		}
+	}
+
+	// enable the kubelet service
+	kubeletService := path.Join(install.BasePath(), "systemd/kubelet.service")
+	if err := install.Run("systemctl", "enable", kubeletService); err != nil {
+		return errors.Wrap(err, "failed to enable kubelet service")
+	}
+
+	// setup the kubelet dropin
+	kubeletDropinSource := path.Join(install.BasePath(), "systemd/10-kubeadm.conf")
+	kubeletDropin := "/etc/systemd/system/kubelet.service.d/10-kubeadm.conf"
+	if err := install.Run("mkdir", "-p", path.Dir(kubeletDropin)); err != nil {
+		return errors.Wrap(err, "failed to configure kubelet service")
+	}
+	if err := install.Run("cp", kubeletDropinSource, kubeletDropin); err != nil {
+		return errors.Wrap(err, "failed to configure kubelet service")
+	}
+
+	return nil
+}
+
+func (b TarBits) extractFromSemVer(
+	src, dst string) (map[string]string, error) {
+
+	return b.extractFromReleaseBuild(
+		fmt.Sprintf("%s/release/%s", tarBitsReleaseBuildURI, src), dst)
+}
+
+func (b TarBits) extractFromCIBuild(
+	src, dst string) (map[string]string, error) {
+
+	uri, err := b.resolveBuildURI(
+		fmt.Sprintf("%s/%s", tarBitsCIBuildURI, src), true)
+	if err != nil {
+		return nil, errors.Wrapf(
+			err, "error resolving CI build uri %s", src)
+	}
+	return b.extractFromHTTP(uri, dst)
+}
+
+func (b TarBits) extractFromReleaseBuild(
+	src, dst string) (map[string]string, error) {
+
+	uri, err := b.resolveBuildURI(
+		fmt.Sprintf("%s/%s", tarBitsReleaseBuildURI, src), false)
+	if err != nil {
+		return nil, errors.Wrapf(
+			err, "error resolving release build uri %s", src)
+	}
+	return b.extractFromHTTP(uri, dst)
+}
+
+func (b TarBits) extractFromHTTP(
+	src, dst string) (map[string]string, error) {
+
+	// Read the version from the remote kubernetes tarball.
+	var version []byte
+	{
+		kubeTarballURI := src + "/kubernetes.tar.gz"
+		resp, err := http.Get(kubeTarballURI)
+		if err != nil {
+			return nil, errors.Wrapf(err, "error reading %s", kubeTarballURI)
+		}
+		if resp.StatusCode != http.StatusOK {
+			return nil, errors.Errorf(
+				"error reading %s: %s", kubeTarballURI, resp.Status)
+		}
+		defer resp.Body.Close()
+
+		buf, err := b.readVersionFromKubeTarball(resp.Body)
+		if err != nil {
+			return nil, errors.Wrapf(
+				err, "error reading version from %s", kubeTarballURI)
+		}
+		version = buf
+	}
+
+	// Update the destination path to include the version.
+	dst = path.Join(dst, string(version))
+	os.MkdirAll(dst, 0755)
+
+	// Write the version file.
+	versionPath := path.Join(dst, "version")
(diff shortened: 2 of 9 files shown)


pkg/build/kube/files.go
```diff
@@ -0,0 +1,44 @@
...
+[Service]
+Environment="KUBELET_KUBECONFIG_ARGS=--bootstrap-kubeconfig=/etc/kubernetes/bootstrap-kubelet.conf --kubeconfig=/etc/kubernetes/kubelet.conf"
+Environment="KUBELET_CONFIG_ARGS=--config=/var/lib/kubelet/config.yaml"
+# This is a file that "kubeadm init" and "kubeadm join" generates at runtime, populating the KUBELET_KUBEADM_ARGS variable dynamically
+EnvironmentFile=-/var/lib/kubelet/kubeadm-flags.env
+# This is a file that the user can use for overrides of the kubelet args as a last resort. Preferably, the user should use
+# the .NodeRegistration.KubeletExtraArgs object in the configuration files instead. KUBELET_EXTRA_ARGS should be sourced from this file.
+EnvironmentFile=-/etc/default/kubelet
+ExecStart=
+ExecStart=/usr/bin/kubelet $KUBELET_KUBECONFIG_ARGS $KUBELET_CONFIG_ARGS $KUBELET_KUBEADM_ARGS $KUBELET_EXTRA_ARGS
```
neolit123: baking in the unit/drop-in is not ideal. because kind now has to maintain and sync with the upstream systemd files...
we have deb/rpm specs scattered in k/k and k/release. kind *probably* shouldn't manage it's own copies.

i would HTTP download them from github:
https://github.com/kubernetes/release/blob/master/debian/xenial/kubelet/lib/systemd/system/kubelet.service
https://github.com/kubernetes/release/blob/master/debian/xenial/kubeadm/channel/stable/etc/systemd/system/kubelet.service.d/post-1.10/10-kubeadm.conf

(1.15 is hopefully going to nuke the above, though).

pkg/build/kube/tarbits.go
```diff
@@ -0,0 +1,425 @@
...
+
+	if strings.HasPrefix(tarRoot, "file://") {
+		tarBits.tarRoot = strings.Replace(tarRoot, "file://", "", 1)
+		tarBits.tbitsType = tarBitsLocal
+	} else if strings.HasPrefix(tarRoot, "http:") ||
+		strings.HasPrefix(tarRoot, "https:") {
+		tarBits.remoteURI = tarRoot
+		tarBits.tbitsType = tarBitsHTTP
+	} else if ok, _ := regexp.MatchString(
+		`^(v\d+\.\d+\.\d+[\w.\-+]*)(::.+)?$`, tarRoot); ok {
```
neolit123: better to use the functions from here:
https://github.com/kubernetes/apimachinery/blob/master/pkg/util/version/version.go

the package is already vendored in kind.

pkg/build/kube/tarbits.go
```diff
@@ -0,0 +1,425 @@
...
+		if goarch == "" {
+			goarch = util.GetArch()
+		}
+
+		// Get the version from the remote "kubernetes.tar.gz"
+		version, err := b.getVersion()
+		if err != nil {
+			return err
+		}
+		fmt.Printf("Building version %s\n", version)
```
neolit123: the other "bits builders" seem to not print anything to stdout.
but if they do probably should use logrus:
https://github.com/kubernetes-sigs/kind/blob/7e62c2124607f40221f3ef80bd27482bfa2c1d4e/pkg/build/kube/aptbits.go#L24

`e.g. log.Debugf(....)`

pkg/build/kube/tarbits.go
```diff
@@ -0,0 +1,425 @@
...
+
+		// Get the location where to the remote files are cached locally.
+		if b.tarRoot == "" {
+			currentUser, err := user.Current()
```
neolit123: this is entering a retrospective minefield of the standard go library.
nowadays this should be portable to all the OSes..

pkg/build/kube/tarbits.go
```diff
@@ -0,0 +1,425 @@
...
+			if err != nil {
+				return err
+			}
+			b.tarRoot = path.Join(
```
neolit123: probably best to not modify the original b.tarRoot field here and use a temp var instead.

pkg/build/kube/tarbits.go
```diff
@@ -0,0 +1,425 @@
...
+			fmt.Sprintf("%s/kubeadm", kubeOSAndArchURI),
+			fmt.Sprintf("%s/kubectl", kubeOSAndArchURI),
+			fmt.Sprintf("%s/kubelet", kubeOSAndArchURI),
+		}
+
+		// Download the files.
+		for _, uri := range downloadables {
+			dst := path.Join(b.tarRoot, path.Base(uri))
+			if err := b.copyFromURI(uri, dst); err != nil {
+				return fmt.Errorf("failed to copy %s to %s", uri, dst)
```
neolit123: `errors.Errorf(`

errors here is github.com/pkg/errors.

pkg/build/kube/tarbits.go
```diff
@@ -0,0 +1,425 @@
...
+func (b *TarBits) copyFromURI(src, dst string) error {
+	n, r, err := b.getReaderForURI(src)
+	if err != nil {
+		return err
+	}
+	defer r.Close()
+
+	// If the file already exists and has the same size as the remote
+	// content then do not redownload it.
+	if f, err := os.Stat(dst); err == nil {
```
neolit123: needs `os.IsNotExist` here

pkg/build/kube/tarbits.go
```diff
@@ -0,0 +1,425 @@
...
+		b.remoteURI = fmt.Sprintf("%s/release/%s", releaseURIPrefix, version)
+	}
+
+	return nil
+}
+
+func (b *TarBits) getReaderForURI(uri string) (int64, io.ReadCloser, error) {
+	resp, err := http.Get(uri)
+	if err != nil {
+		return 0, nil, fmt.Errorf("failed to read %s: %v", uri, err)
```
neolit123: errors.Wrapf

pkg/build/kube/tarbits.go
```diff
@@ -0,0 +1,425 @@
...
+	}
+	if resp.StatusCode != http.StatusOK {
+		return 0, nil, fmt.Errorf("%s: %s", uri, resp.Status)
```
neolit123: errors.Errorf

pkg/build/kube/tarbits.go
```diff
@@ -0,0 +1,425 @@
...
+
+	// Create a tar reader to process the gzip reader.
+	tarReader := tar.NewReader(gzipReader)
+
+	// Iterate until able to read the version file.
+	for {
+		header, err := tarReader.Next()
+		if err != nil {
+			if err == io.EOF {
+				return nil, fmt.Errorf("failed to find version in %s", uri)
```
neolit123: errors.Errorf

pkg/build/kube/files.go
akutz: Hi @neolit123,

I was just following @BenTheElder's `// TODO` from the BazelBits code. Still, I've no issue downloading them either. Ben?

pkg/build/kube/tarbits.go
akutz: Hi @neolit123,

I was *going* to use logrus, but then I noticed there does appear to be non-logger output:

```shell
Created symlink /etc/systemd/system/multi-user.target.wants/kubelet.service → /kind/systemd/kubelet.service.
Created symlink /etc/systemd/system/kubelet.service → /kind/systemd/kubelet.service.
```

I just thought it would be good to know what version is being built. Still, I think it makes sense to write it via the logger instead. Thanks!

pkg/build/kube/tarbits.go
akutz: Hi @neolit123,

I'm not sure if your comment was more opining the portability or if you're suggesting a change?

FWIW the `user.Current()` call *is* [portable](https://golang.org/src/os/user/). I had originally used [`os.UserHomeDir`](https://golang.org/pkg/os/#UserHomeDir), but the CI builds failed. I don't know when `UserHomeDir` was introduced, but it must be in a version later than the one used to build Kind.

pkg/build/kube/tarbits.go
akutz: Hi @neolit123,

Hmm, I think I disagree. This case is only activated when `tarRoot == ""`.

pkg/build/kube/tarbits.go
akutz: Hi @neolit123,

So I'm aware of the function, but I explicitly opted not to use it because in this case I don't care *why* the `stat` failed. If the `stat` fails I want to try and create and write to the file. If the write fails, then an error will be returned. It's a case of preferring `create -> success|fail` versus `check -> create -> success|fail`. If the `check` step fails for any reason then `create` should at least be attempted. Therefore I'm just optimizing away the initial `check` step.

pkg/build/kube/files.go
akutz: Hi @neolit123,

I've addressed your feedback here. I first try to get the files at the remote URIs, and if that fails I default back to the constant values.

neolit123: waiting on Ben's comments on this one. pulling from the release repo is not ideal too.

pkg/build/kube/tarbits.go
neolit123: it was just a comment.
the go in the CI that builds kind *should be* 1.12.1.

alternatively this can be just a mandatory folder specified by the user or defaulting to `./` instead of defaulting to the user home path.

pkg/build/kube/files.go
akutz: Hi @neolit123,

Sounds good. I'm happy to take this in any direction. I figured the constants were what @BenTheElder was thinking since Kind already [embeds the Weave manifest as a constant](https://github.com/kubernetes-sigs/kind/blob/4119e0bb893fd32b85883c1b7cecaab73cb71a54/pkg/build/node/const.go).

pkg/build/kube/tarbits.go
akutz: Hi @neolit123,

I opted to avoid adding any new flags since @BenTheElder indicated he would like all the flags to apply to all the commands.

I'm fervently against the idea of `./` because of the size of what's being downloaded. It's much friendlier to local development to support a well-defined path for the tarbits to avoid re-downloading files when they're already cached locally.

I really don't think it's a problem to use the user's home directory. It's a fairly standard approach for storing app data.

pkg/build/kube/tarbits.go
akutz: Hi @neolit123,

By the way, using `os.IsNotExist` here is also unnecessary since I only care if it exists, not if it doesn't. It's a fine distinction, but `os.Stat` returns a `nil` error if a file exists. The only time you need `os.IsNotExist` is when `os.Stat` *does* return an error.

It's a common misconception that `os.IsExist` is the opposite of `os.IsNotExist` when that's not accurate.The `os.IsExist` function is for when a file operation fails because the operation is attempting to modify an existing file in a disallowed manner. For example, if `os.Symlink(oldname, newname string)` is invoked and the given `newname` points to an existing file or directory, an error will be returned such that `os.IsExist(err) == true`.

The `os.IsNotExistFor` function is for determining whether or not the reason the `os.Stat` or any other operation that accesses a file fails because the file doesn't actually exist -- not whether or not the process can access the file. For example, if the executing process lacks permission to access the file then `os.Stat` returns an error such that `os.IsPermission(err) == true`. In this case `os.IsNotExist(err) == false`, but the process is still unable to stat the file.

fabriziopandini:
My two cents
I really like the idea of behind this PR, but It would be great if we can achieve a better separation between
1. The "extract" logic (the logic that fetches tars/binaries etc)
2. The new bit that takes in input the output of the extract logic for building node image

More specifically if we can move the extract logic moved in a separated package / decoupled with an interface this will cleanup the code and additionally unblock other issues/feature, like e.g extract the second Kubernetes version required for testing upgrades in kind

neolit123:
> More specifically if we can move the extract logic moved in a separated package / decoupled with an interface this will cleanup the code and additionally unblock other issues/feature, like e.g extract the second Kubernetes version required for testing upgrades in kind

+1 i'm all in favor of whatever facilitates upgrades.

akutz:
Hi @fabriziopandini,

> My two cents
> I really like the idea of behind this PR, but It would be great if we can achieve a better separation between
> 
> 1. The "extract" logic (the logic that fetches tars/binaries etc)
> 2. The new bit that takes in input the output of the extract logic for building node image
> 
> More specifically if we can move the extract logic moved in a separated package / decoupled with an interface this will cleanup the code and additionally unblock other issues/feature, like e.g extract the second Kubernetes version required for testing upgrades in kind

Let me preface my response by saying I like where your head is at. I'm a big fan of abstraction layers. I do think there's merit in what you're suggesting, and I think it's possible to get there with some future iterations. But... (there's always a but, isn't there?) ... I'm also not sure it's necessary at this time.

The original commit tree for this PR had two new Bit implementations, `TarBits` and `HttpBits` which did pretty much what you say. I was in the process of adding a third called `PrebuiltBits` or `GCSBits` when I realized that the result was a pretty subpart UX.

From what I've seen, the `Build` functions for Bits implementations are pseudo-atomic -- each bits impl is responsible for fetching/downloading anything and then building it. In fact, Bits doesn't even have the concept of the fetch phase -- it's baked into `Build`; it's even in the function comment in the interface definition:

> ```go
> // Bits provides the locations of Kubernetes Binaries / Images
> // needed on the cluster nodes
> // Implementations should be registered with RegisterNamedBits
> type Bits interface {
> 	// Build returns any errors encountered while building it Kubernetes.
> 	// Some implementations (upstream binaries) may use this step to obtain
> 	// an existing build isntead
> 	Build() error
> ```

I've no issue seeing the evolution of the `Bits` interface adopt the concept of a `BeforeBuild` or `Dependencies` phase that is directly responsible for downloading the bits. In fact, you could convince me of separating the two interfaces entirely: a dependencies resolver and a builder. 

Or perhaps you were you suggesting that `Bits` remain as is and instead something like this be defined alongside `TarBits`:

```go
// TarBitsTractorBeam is implemented by types that pull the image archives and binaries
// from a remote location.
type TarBitsTractorBeam interface {
	// CanLockOn returns a flag that indicates whether this tractor beam is able to pull
	// the bits from the given source.
	CanLockOn(src string) bool

	// Pull engages the tractor beam to download the bits from the remote source
	// and store them in the provided destination.
	Pull(src, dst string) error
}
```

There are half a dozen to 3,000 ways to design most things, and even after the 20+ years I've been doing this, I'd be lying if I didn't say that I **still** tend to delay features until I think they're "perfect." In reality nothing ever is. In fact, one of the lessons I'm still learning nearly two decades later is that it's sometimes best to start with what works and iterate on it, time and resources permitting. This PR works, and so I'm a little hesitant to delay such an important feature for a design change that I'm unsure is entirely necessary at this time.

Still, I'd love to hear more about what you were thinking. I *think* you were referring to the tractor beam idea, but I don't want to assume. Thanks!

-- 
-a

p.s.

>  this will cleanup the code

Excuse me? I'll have you know that:

My code is pretty, and witty and gay
And I pity
Any code that isn’t my code today

:)

akutz:
Hi @fabriziopandini,

May I ask if you were thinking something along the lines of:

```golang
// GetTarBitsExtractorFor returns the TarBitsExtractor for the provided
// kubeRoot value. A nil value is returned if there is no extractor
// available for the provided kubeRoot.
func GetTarBitsExtractorFor(kubeRoot string) TarBitsExtractor {
	if strings.HasPrefix(kubeRoot, "file://") {
		return TarBitsFileExtractor{}
	} else if strings.HasPrefix(kubeRoot, "http:") ||
		strings.HasPrefix(kubeRoot, "https:") {
		return TarBitsHTTPExtractor{}
	} else if _, err := apimv.ParseGeneric(kubeRoot); err == nil {
		return TarBitsSemVerExtractor{}
	} else if strings.HasPrefix(kubeRoot, "ci/") {
		return TarBitsCIBuildExtractor{}
	} else if strings.HasPrefix(kubeRoot, "release/") {
		return TarBitsReleaseBuildExtractor{}
	}
	return TarBitsFileExtractor{}
}

// TarBitsExtractor is implemented by types that can extract the pre-built
// image archives (tar files), the version file from kubernetes.tar.gz, and
// the kubeadm, kubectl, and kubelet binaries.
//
// Files are extracted to $KIND_TARBITS/VERSION. The default value of
// $KIND_TARBITS is $HOME/.kind/tarbits.
type TarBitsExtractor interface {
	// Extract extracts the pre-built image archives (tar files), the
	// version file from kubernetes.tar.gz, and the kubeadm, kubectl, and
	// kubelet binaries. The returned map adheres to the same rules as the
	// map returned by Bits.Paths.
	Extract(src string) (map[string]string, error)
}

// TarBitsFileExtractor verifies the extracted bits on the local disk.
type TarBitsFileExtractor struct{}

// Extract implements TarBitsExtractor.Extract.
func (b TarBitsFileExtractor) Extract(src string) (map[string]string, error) {
	return nil, nil
}

// TarBitsHTTPExtractor extracts the bits for a given HTTP endpoint.
type TarBitsHTTPExtractor struct{}

// Extract implements TarBitsExtractor.Extract.
func (b TarBitsHTTPExtractor) Extract(src string) (map[string]string, error) {
	return nil, nil
}

// TarBitsSemVerExtractor extracts the bits for a release by the given semantic
// version.
type TarBitsSemVerExtractor struct{}

// Extract implements TarBitsExtractor.Extract.
func (b TarBitsSemVerExtractor) Extract(src string) (map[string]string, error) {
	return nil, nil
}

// TarBitsCIBuildExtractor extracts the bits for a given CI build.
type TarBitsCIBuildExtractor struct{}

// Extract implements TarBitsExtractor.Extract.
func (b TarBitsCIBuildExtractor) Extract(src string) (map[string]string, error) {
	return nil, nil
}

// TarBitsReleaseBuildExtractor extracts the bits for a given release build.
type TarBitsReleaseBuildExtractor struct{}

// Extract implements TarBitsExtractor.Extract.
func (b TarBitsReleaseBuildExtractor) Extract(src string) (map[string]string, error) {
	return nil, nil
}

```

I will say that this has made me consider that it's unnecessary to redownload `kubernetes.tar.gz` for different `GOOS_GOARCH` combinations as the code currently does. Given this design I'd likely stick to the remote structure locally as well.

fabriziopandini:
@akutz thanks for your detailed answers. I really appreciate the effort you did for understanding my rant.

Your second comment is definitely hitting the point.

If it is possible to create a new _public_ package `extractor`, exposing `TarBitsExtractor` it will be awesome. The name can be eventually further simplified into `Extractor`, because it can be now consumed by things different than the `TarBits`, but those are details.

What is important is that the code of your `TarBits` will be freed by all the http/GCS/ci/release nuances, while the `extract` package can be re-used for implementing a step for upgrades in kind. Great!

pkg/build/kube/files.go
BenTheElder: The weave constant is not per-kubernetes version. This one is a bit more problematic.

pkg/build/kube/tarbits.go
BenTheElder: Why should they be cached at all? When do entries get evicted?

pkg/build/kube/files.go
BenTheElder: Downloading it separately is also not super ideal, will have to poke around the tarballs a bit. There should be a way to consistently obtain this 🤔

BenTheElder: The TODO about our own config is stale and wrong but the overall situation for these has not been good :(

BenTheElder:
I don't see why you would even need an extractor object / interface? Just have a function to map a source spec to downloading a tarball to a specified location on disk, and then you can have unrelated code extract the tarball and pick out files. That can be pulled out, but I don't think kind should be in the business of offerring this publicly. Perhaps a copy in another repo?

I also don't think this should extract in $HOME. If we want to repeatedly use ci/latest, reuse the built image? This should probably get downloaded to a tempdir and removed on exit.

fabriziopandini:
@BenTheElder I agree extract is base function that is not kind specific and can be reused across projects.
Might be wg component  standard can be home for this. @luxas opinions?
In the meantime, an interface can help in making the transition smoother

akutz:
> I don't see why you would even need an extractor object / interface? Just have a function to map a source spec to downloading a tarball to a specified location on disk, and then you can have unrelated code extract the tarball and pick out files. That can be pulled out, but I don't think kind should be in the business of offerring this publicly. Perhaps a copy in another repo?
> 
> I also don't think this should extract in $HOME. If we want to repeatedly use ci/latest, reuse the built image? This should probably get downloaded to a tempdir and removed on exit.

Hi @BenTheElder,

HA! I was just writing this when I saw your comment:

> ```golang
> // TarBitsExtractorFunc is the signature for functions that are able to
> // extract tar bits from a given src to a given dst.
> type TarBitsExtractorFunc func(src, dst string) error
> ```

It also occurred to me all you really need is a series of functions, not an interface.

akutz:
> I also don't think this should extract in $HOME. If we want to repeatedly use ci/latest, reuse the built image? This should probably get downloaded to a tempdir and removed on exit.

Hi @BenTheElder,

Except if the primary use case is for building in Prow jobs, who cares if they're not removed? I agree when building locally they should be, so perhaps anything extracted from `ci/` is automatically removed?

pkg/build/kube/files.go
akutz: Hi @BenTheElder,

> The weave constant is not per-kubernetes version. This one is a bit more problematic.

And yet I also cannot find the affected files in any of the k/k repos except for release. So while they may be specific to the k/k version, they're not seemingly available per build. I initially wanted to use the version grokked from `kubernetes.tar.gz` and get the service and kubadm conf file specific for that version. Then I realized those were not available.

pkg/build/kube/tarbits.go
akutz: Hi @BenTheElder,

See my comment above. I agree that perhaps caching `ci/` builds is unnecessary. Caching release builds, however, seems to be not the worst idea. At the very minimum having an opt-in mode for caching would be nice for local devs.

akutz:
Hi @BenTheElder,

One more item regarding the caching. For purposes of local dev, what about using `KIND_TARBITS` to set the root directory to which bits are downloaded. If not set it will use a temp directory and remove it afterwards. If set it will not remove it afterwards.

akutz:
Hi @fabriziopandini,

Was that you on a "rant"? If so, you're the friendliest "angry" person I've ever met :) No, I didn't take it as a rant. It was legitimate feedback, and I appreciate it. In fact, I just submitted a refactor to the PR that was inspired by your suggestions. In other words, please keep them coming!

akutz:
Hi @BenTheElder,

I just submitted a refactor that I think will help future changes. I'm concerned about the caching, however. The process now defaults to using temp space, but even when trying to put a `os.RemoveAll` at the end of the `Install` function, it's too early. The build fails because the files are accessed post-Install. Is there some way to tie back into the bits implementation to allow it to clean up after itself?

akutz:
Hi @BenTheElder, et. al,

I also updated the way the local file structure is laid out. Instead of a flat structure it now adheres to the same structure as the GCS buckets. This way it's possible to use the file extractor against a local copy of a pre-built distribution.

akutz:
Hi @BenTheElder,

Okay, for better or worse I added a new commit that adds a `Clean` function to the `Bits` interface to enable a Bits implementation to clean up after the image is built.

akutz:
> @akutz thanks for this update!
> 
> I like the new code structure, but unfortunately the resulting code is still difficult to reuse as a building piece for the kubeadm-upgrade use case because e.g.
> 
> * `extractFromCIBuild` func, `extractFromReleaseBuild` func etc. are private and tied to the `TarBits`  component
> * the piece of code that select the right extractor given an url is hidden into the `NewTarBits` func
> 
> If we can make a last step for enabling reuse I will really appreciate this, but I leave the final call on @BenTheElder

Hi @fabriziopandini,

I don't think the intent was ever to make any of the above accessible by external tooling. I understood the purpose of the extraction simply as a means to make things easier to maintain. I was not under the impression you wanted to use this code for something outside of Kind.

If you're referring to `kubeadm-upgrade` simply as a piece of Kind and not in the kubeadm project, then sure, I can make the extractors public as well as a function to parse which one to return.

akutz:
Hi @fabriziopandini,

I was also thinking about what you want to do and how refactoring `TarBits` might help. I think there may be a much simpler solution. Hang on.

akutz:
Hi @fabriziopandini,

Hi @fabriziopandini,

Referring to your desire to use the logic externally, what do you think of the new `Extract` function on `TarBits`? It gets the bits locally and handles everything up to, excluding the Build step. Is that what you want? A locally staged set of bits? If not then perhaps I am unclear on your goals.

akutz:
Hi @fabriziopandini,

Thank you for taking the time to explain things on Slack. FWIW, I've updated the PR so that the function `pkg/build/kube.TarBitsExtract` is now available:

```golang
// TarBitsExtract extracts bits from a given kubeRoot to the given extractPath
// Please see NewTarBits for information on the valid kubeRoot values.
func TarBitsExtract(kubeRoot, extractPath string) error {
	bits, err := NewTarBits(kubeRoot)
	if err != nil {
		return err
	}
	tarBits := bits.(*TarBits)
	if _, err := tarBits.extract(tarBits.kubeRoot, extractPath); err != nil {
		return err
	}
	return nil
}
```

BenTheElder:
I'm OK to leave it for this PR I guess, but I think the `kube-root` abstraction doesn't hold up for this.

I also kinda think exported libraries for extracting should be more general and live in a shared location for eg kubetest to consume, but I'm not sure where that would be right now, and if we're _just_ exporting for kinder to vendor without stability guarantees I guess that's ok.

akutz:
Hi @BenTheElder,

> I think the `kube-root` abstraction doesn't hold up for this

I renamed `kubeRoot` to `buildID` for the `TarBitsExtract` function.

> I also kinda think exported libraries for extracting should be more general and live in a shared location for eg kubetest to consume, but I'm not sure where that would be right now, and if we're _just_ exporting for kinder to vendor without stability guarantees I guess that's ok.

I concur, and I'd happily spend an hour moving the `ResourceURLFromBuildID` and `Extract` functions into the kind `utils` package, but I'm unsure if the Kind project itself should be seen or used as a source of common code not specific to Kind.

pkg/build/node/node.go
```diff
@@ -315,6 +315,9 @@ func (c *BuildContext) buildImage(dir string) error {
 		return err
 	}
 
+	// Let Bits clean up after itself.
+	c.bits.Clean()
```
BenTheElder: this won't run if we return early

akutz: Hi @BenTheElder,

I considered that, and I thought that perhaps it would be desirable to leave the bits/data that failed to build for purposes of figuring out what went wrong...

pkg/build/kube/files.go
BenTheElder: > And yet I also cannot find the affected files in any of the k/k repos except for release.

there are systemd unit files in kubernetes/kubernetes in the build directory, the non-versioned k/release files are considered a bug and have caused many problems (and they may change inbetween releases happening, so checking in a static version here may not be right)

pkg/build/node/node.go
BenTheElder: Seems reasonable for now. Might revisit later 👍

aojea:
@BenTheElder is this something that we want to add to 0.5?

sozercan:
any updates on this PR?

PR #397, opened 2019-03-22
feature: enable hostpath provisioner

### LINKED ISSUE #118: Dynamic volume provisioning support

joejulian:
kind creates a default host-path StorageClass but cannot create
PersistentVolumes because the `enable-hostpath-provisioner` flag is not
set on kube-controller-manager.

This configures kubeadm to enable that feature.

xref #118

### DIFF
--- a/pkg/cluster/internal/kubeadm/config.go
+++ b/pkg/cluster/internal/kubeadm/config.go
@@ -103,6 +103,8 @@ kubeletConfiguration:
       nodefs.available: "0%"
       nodefs.inodesFree: "0%"
       imagefs.available: "0%"
+controllerManagerExtraArgs:
+  enable-hostpath-provisioner: "true"
 `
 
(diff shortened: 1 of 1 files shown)


BenTheElder:
/ok-to-test
thanks!

BenTheElder:
we probably should note somewhere that this is not going to work well with multinode. https://github.com/kubernetes/kubernetes/blob/c48ed1c5aca6e06b6a3d70d5f189216df3150379/cmd/kube-controller-manager/app/options/persistentvolumebindercontroller.go#L40, we probably still want a multi-node aware option for #118

BenTheElder:
/retest
build flake

joejulian:
Right. The intent of this PR is to fix the existing implementation. This _feature_ should probably be redone to use local type.

BenTheElder:
(and thanks for fixing it! :smile:)

PR #399, opened 2019-03-22
Update hack/ci/e2e.sh for alpha3 changes

akutz:
This patch updates the file `hack/ci/e2e.sh` to use `v1alpha3` in the generated kind config yaml as well as removes the use of `replicas` and manually adds a second worker node entry to the same yaml.

Discovered at https://kubernetes.slack.com/archives/CEKK1KTN2/p1553287552595400?thread_ts=1553287191.592800&cid=CEKK1KTN2.


tao12345666333:
/lgtm

Thanks. :+1:

Issue #401, opened 2019-03-25
Support for stdin config


josdotso:
Is there some way other than shell hacks to do stdin config?

This didn't work for me:
```bash
kind create cluster --config - <<EOF
kind: Cluster
apiVersion: kind.sigs.k8s.io/v1alpha3

kubeadmConfigPatches:
- |
  apiVersion: kubeadm.k8s.io/v1beta1
  kind: ClusterConfiguration
  metadata:
    name: config
  controllerManager:
    extraArgs:
      enable-hostpath-provisioner: "true"
EOF
# Error: error loading config: open -: no such file or directory
```

BenTheElder:
there is not currently, but this would be pretty easy to add.

BenTheElder:
fixed in https://github.com/kubernetes-sigs/kind/pull/402

josdotso:
Thanks all!

PR #402, opened 2019-03-25
Add support for stdin config

aojea:
It allows to read the config data from stdin when using a dash `-`  as filename for the `--config` flag

https://github.com/kubernetes-sigs/kind/issues/401

### DIFF
--- a/pkg/cluster/config/encoding/scheme.go
+++ b/pkg/cluster/config/encoding/scheme.go
@@ -18,6 +18,7 @@ package encoding
 
 import (
 	"io/ioutil"
+	"os"
 
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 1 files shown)


pkg/cluster/config/encoding/scheme.go
```diff
@@ -53,10 +54,16 @@ func AddToScheme(scheme *runtime.Scheme) {
 // If path == "" then the default config is returned
```
BenTheElder: the "-" behavior should be in the method documentation

pkg/cluster/config/encoding/scheme.go
```diff
@@ -53,10 +54,16 @@ func AddToScheme(scheme *runtime.Scheme) {
...
 func Load(path string) (*config.Cluster, error) {
 	var latestPublicConfig = &v1alpha3.Cluster{}
+	var err error
+	var contents []byte
```
BenTheElder: nit: we should not hoist these to a wider scope unnecessarily. these can be inside the `if path != "" {` block

pkg/cluster/config/encoding/scheme.go
```diff
@@ -51,12 +52,21 @@ func AddToScheme(scheme *runtime.Scheme) {
...
 
 	if path != "" {
-		// read in file
-		contents, err := ioutil.ReadFile(path)
+		var err error
+		var contents []byte
+
+		if path == "-" {
+			// read in stdin
+			contents, err = ioutil.ReadAll(os.Stdin)
```
BenTheElder: this is slightly odd from a seperation of concerns 
(For some deep library method to handle "-" and read stdin) but I don't see a great alternative for now

PR #404, opened 2019-03-26
WIP Switch default CNI to Calico

aojea:
Kind should have a CNI that allows him to support IPv4, IPv6 IP families.
From current CNI plugins, seems that only Calico meets that requirements.

However, there are several issues that still need to be resolved

- [x] Calico depends on the cluster-cidr parameter, and it should match at installation time
> This should fall within --cluster-cidr.  - name: CALICO_IPV4POOL_CIDR value: "192.168.0.0/16"

- [x] Test IPv6 support with kind

- [x] Architecture support, Calico works on amd64, arm64, and ppc64le, is this enough?

- [ ] This PR keeps weave as fallback in case the default-cni-manifest is not found in the node-image, because weave doesn't depend on the cluster parameters, is this ok?

### DIFF
--- a/pkg/build/node/const.go
+++ b/pkg/build/node/const.go
@@ -24,225 +24,735 @@ const (
 )
 
 /*
-The default CNI manifest and images are from weave currently.
+The default CNI manifest and images are from calico currently.
 
 To update these:
-- find the latest stable release at https://github.com/weaveworks/weave/releases
-- copy the weave-daemonset-k8s-1.8.yaml to the defaultCNIManifest string
+- find the latest stable release at https://github.com/projectcalico/calico
+- copy the maniftest from https://docs.projectcalico.org/v3.6/getting-started/kubernetes/installation/calico
+ to the defaultCNIManifest string
 - add a comment to the beginning of the string with the source URL
 - update the defaultCNIImages array to include the images in the manifest
 - update the comment below with the release URL
 
-Current version: https://github.com/weaveworks/weave/releases/tag/v2.5.1
+Current version: https://github.com/projectcalico/calico/releases/tag/v3.6.1
 */
 
-var defaultCNIImages = []string{"weaveworks/weave-kube:2.5.1", "weaveworks/weave-npc:2.5.1"}
+var defaultCNIImages = []string{"calico/cni:v3.6.1", "calico/node:v3.6.1", "calico/kube-controllers:v3.6.1", "calico/ctl:v3.6.1"}
 
-const defaultCNIManifest = `# https://github.com/weaveworks/weave/releases/download/v2.5.1/weave-daemonset-k8s-1.8.yaml
+const defaultCNIManifest = `# https://docs.projectcalico.org/v3.6/getting-started/kubernetes/installation/hosted/kubernetes-datastore/calico-networking/1.7/calico.yaml
+---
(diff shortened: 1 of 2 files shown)


pkg/build/node/const.go
```diff
@@ -24,225 +24,766 @@ const (
...
+            - name: FELIX_IPINIPMTU
+              valueFrom:
+                configMapKeyRef:
+                  name: calico-config
+                  key: veth_mtu
+            # The default IPv4 pool to create on startup if none exists. Pod IPs will be
+            # chosen from this range. Changing this value after installation will have
+            # no effect. This should fall within --cluster-cidr.
+            - name: CALICO_IPV4POOL_CIDR
+              value: "192.168.0.0/16"
```
BenTheElder: hmm, looks like we aren't setting one right now. 🤔

aojea: yeah, that's something I'm trying to understand. 
I've created a multinode cluster with the current Calico configuration and the communication between pods and nodes are working, will try to run the e2e tests to see what happens.

BenTheElder:
I did some testing before this commit, but found #406, which I'm going to get patched first!! :sweat_smile:

pkg/build/node/const.go
aojea: From Calico documentation, I think that these variables are a great addon
> | CALICO_IPV4POOL_CIDR | The IPv4 Pool to create if none exists at start up. It is invalid to define this variable and NO_DEFAULT_POOLS. | IPv4 CIDR | 192.168.0.0/16 | |
| CALICO_IPV6POOL_CIDR | The IPv6 Pool to create if none exists at start up. It is invalid to define this variable and NO_DEFAULT_POOLS. | IPv6 CIDR | fd80:24e2:f998:72d6::/64 | |

I still need to explore more Calico or maybe we should reach out to Calico's community to ask for advice

aojea: Ok, I´ve found how to overcome this, we can leverage k8s and the host-local ipam plugin to avoid to create pools in Calico
By the way, I just hit this https://github.com/projectcalico/calico/issues/2458

BenTheElder: I had node bringup failures when previously testing this with mulit-node FYI.

Looked into clusterCIDR etc. quite a bit more and options for that. not overly thrilled with any of them jus yet.

> Ok, I´ve found how to overcome this, we can leverage k8s and the host-local ipam plugin to avoid to create pools in Calico

can you elaborate some more on this?

aojea: Setting CALICO_IPV4POOL_CIDR  to one CIDR implies to set `podSubnet` in `kubeadm` to the same value, creating a dependency between ` kind` and `calico`. Basically, I'm trying to decouple the IPAM configuration from Calico leveraging k8s IPAM and the node podCIDR field annotation.
This way we can have a static Calico configuration and, in the future, being able to add new CNIs without requiring any changes to ` kind`, just only providing an installation manifest.
Another interesting debate is that we should set the ` podSubnet` field, [are we happy predefining its value in ` kind` ? seems that is not clear this point in kubeadm](https://github.com/kubernetes/kubeadm/issues/1011)

aojea:
Closing as it´s duplicated in #348

PR #408, opened 2019-03-27
[WIP] Add restart cluster command

tao12345666333:
xref: #148 

- [x] Add restart cluster command
- [ ] The container's IP will be change, when restart Nodes(containers). We need update the `admin.conf` and others.
    - [x] Change HAProxy's config file. Make it can forward request to correct control plane.
    - [ ] Make control plane can work correctly. Include `api-server` , `etcd` and some certs.

### DIFF
--- a/pkg/cluster/internal/create/actions/loadbalancer/haproxy.go
+++ b/pkg/cluster/internal/create/actions/loadbalancer/haproxy.go
@@ -65,11 +65,22 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 		return err
 	}
 
+	// starts a docker container with HA proxy load balancer
+	if err := loadBalancerNode.Command(
+		"/bin/sh", "-c",
+		fmt.Sprintf(
+			"docker run -d -v /kind/haproxy.cfg:/usr/local/etc/haproxy/haproxy.cfg:ro --network host --restart always --name haproxy %s",
+			haproxy.Image,
+		),
+	).Run(); err != nil {
+		return errors.Wrap(err, "failed to start haproxy")
+	}
+
 	ctx.Status.End(true)
 	return nil
 }
 
-// ConfigHAProxy is common method for HAProxy
+// ConfigHAProxy is common method for write HAProxy's config file
 func ConfigHAProxy(loadBalancerNode *nodes.Node, allNodes []nodes.Node) error {
 	// collect info about the existing controlplane nodes
 	var backendServers = map[string]string{}
@@ -101,17 +112,6 @@ func ConfigHAProxy(loadBalancerNode *nodes.Node, allNodes []nodes.Node) error {
 	if err := loadBalancerNode.WriteFile("/kind/haproxy.cfg", haproxyConfig); err != nil {
 		// TODO: logging here
 		return errors.Wrap(err, "failed to copy haproxy config to node")
-	}
-
-	// starts a docker container with HA proxy load balancer
-	if err := loadBalancerNode.Command(
-		"/bin/sh", "-c",
-		fmt.Sprintf(
-			"docker run -d -v /kind/haproxy.cfg:/usr/local/etc/haproxy/haproxy.cfg:ro --network host --restart always --name haproxy %s",
-			haproxy.Image,
-		),
-	).Run(); err != nil {
-		return errors.Wrap(err, "failed to start haproxy")
 	}
 
 	return nil
--- a/pkg/cluster/internal/create/actions/loadbalancer/haproxy.go
+++ b/pkg/cluster/internal/create/actions/loadbalancer/haproxy.go
@@ -60,6 +60,17 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 	ctx.Status.Start("Starting the external load balancer ⚖️")
 	defer ctx.Status.End(false)
 
+	err = ConfigHAProxy(loadBalancerNode, allNodes)
+	if err != nil {
+		return err
+	}
+
+	ctx.Status.End(true)
+	return nil
+}
+
+// ConfigHAProxy is common method for HAProxy
+func ConfigHAProxy(loadBalancerNode *nodes.Node, allNodes []nodes.Node) error {
 	// collect info about the existing controlplane nodes
 	var backendServers = map[string]string{}
 	controlPlaneNodes, err := nodes.SelectNodesByRole(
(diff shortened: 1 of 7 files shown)


tao12345666333:
>is there a way to make docker restart keep the old IP/port?

if we use docker run with --ip flag maybe resolve it.

pkg/cluster/internal/restart/restart.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+// Cluster restarts the cluster identified by ctx
+func Cluster(c *context.Context) error {
+	n, err := c.ListNodes()
+	if err != nil {
+		return errors.Wrap(err, "error listing nodes")
+	}
+
+	return nodes.Restart(n...)
```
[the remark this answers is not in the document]
tao12345666333: 👌 will update.

pkg/cluster/nodes/nodes.go
```diff
@@ -134,3 +134,49 @@ func WaitForReady(node *Node, until time.Time) bool {
...
+	for _, node := range nodes {
+		ids = append(ids, node.name)
+	}
+	cmd := exec.Command(
+		"docker",
+		append(
+			[]string{
+				"restart",
+				// the default wait time is 10s
+				"--time=1",
```
vincepri: Is 1s enough wait for a successful restart?

neolit123: to my understanding it's 1 second before killing the container.
https://docs.docker.com/engine/reference/commandline/restart/

tao12345666333: yep, docker will send `SIGTERM` and after the wait time will send `SIGKILL`. 
set it to `1s` just for speed up.

pkg/cluster/internal/create/actions/loadbalancer/haproxy.go
```diff
@@ -60,6 +60,16 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
...
+	err = ConfigHAProxy(loadBalancerNode, allNodes)
+	if err != nil {
+		return err
+	}
+
+	ctx.Status.End(true)
+	return nil
+}
+
+func ConfigHAProxy(loadBalancerNode *nodes.Node, allNodes []nodes.Node) error {
```
aojea: linter is failing in this line with the following error
> /home/prow/go/src/sigs.k8s.io/kind/pkg/cluster/internal/create/actions/loadbalancer/haproxy.go:72:1: exported function ConfigHAProxy should have comment or be unexported

tao12345666333: Thanks. :+1:  Addressed.

tao12345666333:
after https://github.com/kubernetes-sigs/kind/pull/461 be merged we can make this simple. although the core issue to be addressed is still IP changes.

I will complete this PR as soon as possible after the end of this holiday. 

I need to complete the following sections.

* change all old IP to new IP at /etc/kubernetes.
* regenerate apiserver and etcd peer certs

aojea:
just sharing my thoughts,  what's the best option regarding the IP address problem?
1. Try to keep the same ip addresses on all the nodes after restart?
2. Regenerate all certificates after restart matching the new ip addresses?

I guess that 1 is the option closer to real user scenarios, and looking at docker networking seems possible to assign static ip addresses to the node so a possible solution can be.
1. Get node ip address
2. Stop node
3. Start node with address obtained in 1

tao12345666333:
Thanks for your point. 

My current practice is the first one.  

Because the scenario we first encountered was due to "docker restart",  #148 this does not guarantee that the original IP is not already occupied. It may not be suitable to start it with the original static IP.


Of course, if the restart is done in 1 , then we should select the higher IP in the IP address pool at startup to avoid being occupied.

tao12345666333:
Do you have any suggestions?  @BenTheElder @neolit123  I think it would be easier to do this by using kubeadm.

pkg/cluster/nodes/nodes.go
```diff
@@ -134,3 +134,49 @@ func WaitForReady(node *Node, until time.Time) bool {
...
+			// TODO: logging here
+			return err
+		}
+
+		if err := node.SignalStart(); err != nil {
+			// TODO: logging here
+			return err
+		}
+
+		if !node.WaitForDocker(time.Now().Add(time.Second * 60)) {
```
neolit123: i think this PR will be affected by the changes in:
https://github.com/kubernetes-sigs/kind/pull/461

tao12345666333: yes. after the PR has been merged these codes remove.

BenTheElder: I need to rework HAproxy in the new world :tm: as well, so those parts will be getting a PR soon and also will need updating... sorry about that!

tao12345666333: do not mind. as long as we can push things forward, it will suffice.

aojea:
@tao12345666333 I think that the problem is that we are using `docker restart` that's adding the complexity of dealing with the IP assignment.
IIUIC the goal is to simulate a node restart. Since the nodes are using systemd I think that it could be possible to implement a "simulated" restart just restarting all service inside the container `systemctl restart kubelet containerd ...` getting rid of the `docker restart` command.
@neolit123 @fabriziopandini @BenTheElder any thoughts?

neolit123:
the original issue https://github.com/kubernetes-sigs/kind/issues/148#issue-387428987
talks about docker on the host being restarted which renders the cluster broken.

with this comment:
https://github.com/kubernetes-sigs/kind/issues/148#issuecomment-488167271
we are one step further, but i don't think the IP update is avoidable here.

can we just use container networking instead of IPs (e.g. `run --network`).
kubeadm should work be fine with that...

BenTheElder:
- we should explore networks with --network
- the primary goal AFAIK is to survive container restarts, mostly from users  on docker for mac after restarting the daemon i'd guess.
- after we sort out networking, this should just be starting the containers with matching labels, fixmounts etc. is gone now :-)

tao12345666333:
I will open another PR to handle the network problem(using `--network` instead of IPs). After the processing is completed, I will process this PR. 

Thanks all. :+1:

Ilyes512:
> removed this from the v0.5.0 milestone 26 days ago
It would be really awesome if it was possible to restart kind cluster(s) after for example rebooting (MacOS) your host. I (want to) use kind for local development. My end goal is to replace minikube.

Luckily recreating the cluster(s) is really fast, but it would still be nice to just (re)start it. I am currently looking at Velero and see if I can use that to restore the state after recreation.

BenTheElder:
this will need to be revisited on top of the provider work after we figure out the network issues

thanks again for the PR, we'll surely reference this when we've resolved the network part and come to revive restart support.

Issue #409, opened 2019-03-27
node-image build failure due to unavailability of kubeadm build artifact


mgdevstack:
Building node-image is getting failed as `kubeadm` build artifact is not getting copied from the container to host machine.
**Note: Observed failure on MacOS but same is getting succeeded from Ubuntu.**

Is there any way we can find the root cause of the issue?

Following are the logs from system (MacOS)
```
➜  kubernetes (master) ✔ make
+++ [0327 17:26:49] Building go targets for darwin/amd64:
    ./vendor/k8s.io/code-generator/cmd/deepcopy-gen
+++ [0327 17:27:00] Building go targets for darwin/amd64:
    ./vendor/k8s.io/code-generator/cmd/defaulter-gen
+++ [0327 17:27:06] Building go targets for darwin/amd64:
    ./vendor/k8s.io/code-generator/cmd/conversion-gen
+++ [0327 17:27:22] Building go targets for darwin/amd64:
    ./vendor/k8s.io/kube-openapi/cmd/openapi-gen
+++ [0327 17:27:34] Building go targets for darwin/amd64:
    ./vendor/github.com/jteeuwen/go-bindata/go-bindata
+++ [0327 17:27:37] Building go targets for darwin/amd64:
    cmd/kube-proxy
    cmd/kube-apiserver
    cmd/kube-controller-manager
    cmd/cloud-controller-manager
    cmd/kubelet
    cmd/kubeadm
    cmd/hyperkube
    cmd/kube-scheduler
    vendor/k8s.io/apiextensions-apiserver
    cluster/gce/gci/mounter
    cmd/kubectl
    cmd/gendocs
    cmd/genkubedocs
    cmd/genman
    cmd/genyaml
    cmd/genswaggertypedocs
    cmd/linkcheck
    vendor/github.com/onsi/ginkgo/ginkgo
    test/e2e/e2e.test
    cmd/kubemark
    vendor/github.com/onsi/ginkgo/ginkgo
➜  kubernetes (master) ✔ kind build base-image
Sending build context to Docker daemon  2.187MB
Step 1/17 : ARG BASE_IMAGE="ubuntu:18.04"
Step 2/17 : FROM ${BASE_IMAGE}
 ---> 94e814e2efa8
Step 3/17 : ARG DEBIAN_FRONTEND=noninteractive
 ---> Using cache
 ---> b3f4877772f1
Step 4/17 : COPY clean-install /usr/local/bin/clean-install
 ---> Using cache
[...]

PR #413, opened 2019-03-27
always build kubernetes for linux

BenTheElder:
fixes https://github.com/kubernetes-sigs/kind/pull/358/files#r269746243

kind only supports linux "nodes" currently, we can just use the host arch, but we need to build for linux even on mac 🙃

### DIFF
--- a/pkg/build/kube/dockerbuildbits.go
+++ b/pkg/build/kube/dockerbuildbits.go
@@ -21,8 +21,9 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"sigs.k8s.io/kind/pkg/util"
 	"strings"
+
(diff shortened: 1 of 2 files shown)


BenTheElder:
also likely fixes #409

pkg/build/kube/dockerbuildbits.go
```diff
@@ -86,6 +87,10 @@ func (b *DockerBuildBits) Build() error {
 	return buildVersionFile(b.kubeRoot)
 }
 
+func dockerBuildOsAndArch() string {
```
neolit123: wonder if `Os` will pass golint.
other than that LGTM.

BenTheElder:
can confirm this fixes mac with `kind build node-image`

PR #416, opened 2019-03-28
CLI: Add shell auto completions

paranoidaditya:
Adds a subcommand to dump out bash and zsh autocompletion files

### DIFF
--- a/cmd/kind/completion/bash/bash.go
+++ b/cmd/kind/completion/bash/bash.go
@@ -1,0 +1,37 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


cmd/kind/completion/bash/bash.go
```diff
@@ -0,0 +1,37 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
tao12345666333: change it to 2019. [1]

cmd/kind/completion/completion.go
```diff
@@ -0,0 +1,55 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
tao12345666333: [1]

cmd/kind/completion/zsh/zsh.go
```diff
@@ -0,0 +1,37 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
tao12345666333: [1]

paranoidaditya:
Autocompletion is definitely useful imo, we could either do it this way or have static files up for grabs?

BenTheElder:
yep, apologies for the latency, couple other things have been eating into review time etc.

reading it now the code looks great, thanks.

BenTheElder:
@neolit123 maybe we should just add this and later we can improve the root command help to organize command groupings?

BenTheElder:
/hold cancel
thanks again @paranoidaditya !

PR #424, opened 2019-04-02
Document kind proxy behaviour

### LINKED ISSUE #347: document / update known proxy related issues
We should document somewhere:
- kind's behavior W.R.T. *_PROXY environment variables
- probably also that you may need to set / unset these
/kind documentation

aojea:
Document how to use a proxy with kind

Fixes #347

### DIFF
--- a/site/content/docs/user/quick-start.md
+++ b/site/content/docs/user/quick-start.md
@@ -227,7 +227,7 @@ nodes:
 ```
 
 ### Configure kind to use a proxy
-If you are running kind in an environment that requires a proxy to connect outside you may need to use it.
+If you are running kind in an environment that requires a proxy, you may need to configure kind to use it.
 
 You can configure kind to use a proxy using the environment variables (uppercase takes precedence):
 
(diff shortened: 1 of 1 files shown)


site/content/docs/user/quick-start.md
```diff
@@ -226,6 +226,18 @@ nodes:
 - role: worker
 ```
 
+### Configure kind to use a proxy
+If you are running kind in an environment that requires a proxy to connect outside you may need to use it.
```
BenTheElder:
```suggestion
If you are running kind in an environment that requires a proxy, you may need to configure kind to use it.
```

site/content/docs/user/quick-start.md
```diff
@@ -226,6 +226,18 @@ nodes:
...
+
+You can configure kind to use a proxy using the environment variables (uppercase takes precedence):
+
+* HTTP_PROXY or http_proxy
+* HTTPS_PROXY or https_proxy
+* NO_PROXY or no_proxy
```
BenTheElder: It would probably be helpful to link to some sort of reference regarding what these usually mean and what their values look like. perhaps docker's docs.

tao12345666333: maybe we can link to https://docs.docker.com/network/proxy/

PR #428, opened 2019-04-03
Enable containerd as an option for the CRI

dims:
- We use the version of containerd and ctr shipped with Docker 18.09
- Use the defaults from running "containerd config default" except for the snapshotter
- We use "native" as the snapshotter instead of overlayfs
- Download "crictl" command as that is needed by kubeadm when we specify criSocket in the kubeadm configuration
- Pull images from docker and register with containerd using "ctr" command

To enable containerd to be used, you can use the criSocket configuration
option in kind's configuration (which is passed into kubeadm) as shown below:
```
# this config file contains all config fields with comments
kind: Cluster
# patch the generated kubeadm config with some extra settings
kubeadmConfigPatches:
- |
  kind: InitConfiguration
  metadata:
    name: config
  nodeRegistration:
nodes:
- role: control-plane
```

To verify that containerd can see the images try both
- `crictl -r /run/containerd/containerd.sock images`
- `ctr --namespace=k8s.io images ls`

Change-Id: I57e85eeb196eb4336c0dc98d4f9ff07b8c533190


nikhita:
Hmm, why isn't Netlify running here...

images/base/Dockerfile
```diff
@@ -92,6 +92,11 @@ RUN export ARCH=$(dpkg --print-architecture) \
     && tar -C /opt/cni/bin -xzf /tmp/cni.tgz \
     && rm -rf /tmp/cni.tgz
 
+# Install crictl to /usr/local/bin
+ARG CRICTL_VERSION="v1.13.0"
+RUN export ARCH=$(dpkg --print-architecture) \
+    && curl -fSL "https://github.com/kubernetes-sigs/cri-tools/releases/download/${CRICTL_VERSION}/crictl-${CRICTL_VERSION}-linux-${ARCH}.tar.gz" | tar xzC /usr/local/bin
```
BenTheElder: I have a feeling we will also need `ctr` cc @Random-Liu

dims: i agree that we will need `ctr`, will rework this

/hold

dims:
testing using kind config

```
# this config file contains all config fields with comments
kind: Cluster
# patch the generated kubeadm config with some extra settings
kubeadmConfigPatches:
- |
  kind: InitConfiguration
  metadata:
    name: config
  nodeRegistration:
nodes:
- role: control-plane
```

images/base/Dockerfile
```diff
@@ -22,7 +22,12 @@
 # https://developers.redhat.com/blog/2016/09/13/running-systemd-in-a-non-privileged-container/
 
 ARG BASE_IMAGE="ubuntu:18.04"
+
+FROM golang:latest as fetch_ctr
+RUN go get github.com/containerd/containerd/cmd/ctr
```
BenTheElder: so @Random-Liu says that docker 18.09+ should actually have ctr already? it's part of the containerd release

pkg/cluster/nodes/node.go
```diff
@@ -167,7 +167,17 @@ func (n *Node) LoadImages() {
...
+		return
+	}
+
+	// load images cached on the node into containerd
+	if err := n.Command(
+		"/bin/bash", "-c",
+		// use xargs to load images in parallel
+		`docker images --format "{{.Repository}}:{{.Tag}}" | while read -r line; do echo "$line" && docker save "$line" > tmp.tar && ctr --namespace=k8s.io images import tmp.tar && rm tmp.tar; done`,
+	).Run(); err != nil {
+		log.Warningf("Failed to preload images into docker: %v", err)
```
BenTheElder: s/docker/containerd

dims:
latest error:
```
Apr 04 16:52:26 kind-control-plane kubelet[2270]: E0404 16:52:26.312438    2270 remote_runtime.go:109] RunPodSandbox from runtime service failed: rpc error: code = Unknown desc = failed to start sandbox container: failed to create containerd task: failed to mount rootfs component &{overlay overlay [workdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/19/work upperdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/19/fs lowerdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/1/fs]}: invalid argument: unknown
Apr 04 16:52:26 kind-control-plane kubelet[2270]: E0404 16:52:26.312554    2270 kuberuntime_sandbox.go:68] CreatePodSandbox for pod "kube-scheduler-kind-control-plane_kube-system(6965a328cde2418fb9eb4d1ed720cb6b)" failed: rpc error: code = Unknown desc = failed to start sandbox container: failed to create containerd task: failed to mount rootfs component &{overlay overlay [workdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/19/work upperdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/19/fs lowerdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/1/fs]}: invalid argument: unknown
Apr 04 16:52:26 kind-control-plane kubelet[2270]: E0404 16:52:26.312605    2270 kuberuntime_manager.go:693] createPodSandbox for pod "kube-scheduler-kind-control-plane_kube-system(6965a328cde2418fb9eb4d1ed720cb6b)" failed: rpc error: code = Unknown desc = failed to start sandbox container: failed to create containerd task: failed to mount rootfs component &{overlay overlay [workdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/19/work upperdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/19/fs lowerdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/1/fs]}: invalid argument: unknown
Apr 04 16:52:26 kind-control-plane kubelet[2270]: E0404 16:52:26.312704    2270 pod_workers.go:190] Error syncing pod 6965a328cde2418fb9eb4d1ed720cb6b ("kube-scheduler-kind-control-plane_kube-system(6965a328cde2418fb9eb4d1ed720cb6b)"), skipping: failed to "CreatePodSandbox" for "kube-scheduler-kind-control-plane_kube-system(6965a328cde2418fb9eb4d1ed720cb6b)" with CreatePodSandboxError: "CreatePodSandbox for pod \"kube-scheduler-kind-control-plane_kube-system(6965a328cde2418fb9eb4d1ed720cb6b)\" failed: rpc error: code = Unknown desc = failed to start sandbox container: failed to create containerd task: failed to mount rootfs component &{overlay overlay [workdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/19/work upperdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/19/fs lowerdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/1/fs]}: invalid argument: unknown"
```

BenTheElder:
@dims note that the presubmits do not build the base image currently

dims:
@BenTheElder ack, i am testing locally

```
cd $GOPATH/src/sigs.k8s.io/kind && go install .
kind build base-image --image kindest/base:latest --source $GOPATH/src/sigs.k8s.io/kind/images/base --loglevel debug
kind build node-image --type bazel --image kindest/node:latest --base-image kindest/base:latest --kube-root=$GOPATH/src/k8s.io/kubernetes/
kind create cluster --image kindest/node:latest --config ~/junk/kind.conf
```

neolit123:
> latest error:

@dims did you check if this works outside of the kind nodes - e.g using kubeadm/kubelet on the host?

dims:
@neolit123 good point, i haven't yet

dims:
found a toggle!!

```
root@kind-control-plane:/# crictl ps
CONTAINER ID        IMAGE               CREATED             STATE               NAME                      ATTEMPT             POD ID
cdd3855fb39a9       eb516548c180f       3 minutes ago       Running             coredns                   0                   968c945b60c2e
efd2aa48d8ba5       eb516548c180f       3 minutes ago       Running             coredns                   0                   68fb982c5068f
3beb91d375965       1f394ae9e2260       3 minutes ago       Running             weave                     0                   d5e5632e06145
1599de9fa1f3f       986631f43652b       3 minutes ago       Running             kube-proxy                0                   37dceb7b0f71b
6ae8e7fd73767       849273a4e2ed3       3 minutes ago       Running             kube-apiserver            0                   4187e8d8ac56d
9ab68b065a9ec       bd0442844cc94       3 minutes ago       Running             kube-controller-manager   0                   b5f62c53df3bf
a7507ee1ae074       2c4adeb21b4ff       3 minutes ago       Running             etcd                      0                   231ea0c75c2ab
96be9bd1db014       d2b12e84e2342       3 minutes ago       Running             kube-scheduler            0                   218a70f4a940e
root@kind-control-plane:/# crictl version
Version:  0.1.0
RuntimeName:  containerd
RuntimeVersion:  1.2.5
RuntimeApiVersion:  v1alpha2
root@kind-control-plane:/# crictl ps
CONTAINER ID        IMAGE               CREATED             STATE               NAME                      ATTEMPT             POD ID
cdd3855fb39a9       eb516548c180f       3 minutes ago       Running             coredns                   0                   968c945b60c2e
efd2aa48d8ba5       eb516548c180f       3 minutes ago       Running             coredns                   0                   68fb982c5068f
3beb91d375965       1f394ae9e2260       3 minutes ago       Running             weave                     0                   d5e5632e06145
1599de9fa1f3f       986631f43652b       3 minutes ago       Running             kube-proxy                0                   37dceb7b0f71b
6ae8e7fd73767       849273a4e2ed3       4 minutes ago       Running             kube-apiserver            0                   4187e8d8ac56d
9ab68b065a9ec       bd0442844cc94       4 minutes ago       Running             kube-controller-manager   0                   b5f62c53df3bf
a7507ee1ae074       2c4adeb21b4ff       4 minutes ago       Running             etcd                      0                   231ea0c75c2ab
96be9bd1db014       d2b12e84e2342       4 minutes ago       Running             kube-scheduler            0                   218a70f4a940e
```

dims:
Added lots of notes in the PR body. Can someone please kick the tires?

BenTheElder:
I will! Had to go take care of some other things, back on this now!

images/base/Dockerfile
```diff
@@ -78,6 +78,7 @@ RUN curl -fsSL "https://download.docker.com/linux/$(. /etc/os-release; echo "$ID
 # Enable CRI for containerd
 RUN mkdir -p /etc/docker && echo '{"cri-containerd": true}' > /etc/docker/daemon.json
 RUN mkdir -p /etc/containerd && containerd config default > /etc/containerd/config.toml
+RUN sed -i "s/overlayfs/native/" /etc/containerd/config.toml
```
neolit123: did we find out why it's needed? i personally have no idea ¯\_(ツ)_/¯

pkg/cluster/nodes/node.go
```diff
@@ -167,7 +167,17 @@ func (n *Node) LoadImages() {
...
-		log.Warningf("Failed to preload docker images: %v", err)
+		log.Warningf("Failed to preload images into docker: %v", err)
+		return
+	}
+
+	// load images cached on the node into containerd
+	if err := n.Command(
+		"/bin/bash", "-c",
+		// use xargs to load images in parallel
+		`docker images --format "{{.Repository}}:{{.Tag}}" | while read -r line; do echo "$line" && docker save "$line" > tmp.tar && ctr --namespace=k8s.io images import --snapshotter=native tmp.tar && rm tmp.tar; done`,
```
neolit123: i don't understand very well how having the images loaded in both docker and containerd affects the node in terms of e.g. memory footprint. maybe there is some magic behind the scenes to minimize this, but i wonder if we have a case where we now have the images loaded twice?

do we need a knob in kind to do that.
the kubeadm config socket is one way, but for that:
- kind needs to process the kubeadm config to check if the socket is different from the default
- we need to propagate this "knob" here so that the Node object and LoadImages know about it.

BenTheElder: we won't leave it this way, but this is a reasonable enough answer until we add the more complex answer I think

BenTheElder: first kind will need to know if the image supports this, to do that we'll start detecting metadata from the image (labels?) to identify features

then kind can simply skip the docker load when it detects that it doesn't need to use dockershim
the `kind load` commands will also need to support this, and the proxy configuration will need to handle containerd.

images/base/Dockerfile
BenTheElder: I suspect we'll need further changes around this, kind needs to use a volume where the snapshotter will be used, we do this for docker but containerd may use another path.

BenTheElder: so containerd is running _alongside_ dockerd, and uses `/var/lib/containerd` instead of `/var/lib/docker` for the snapshotters

pkg/cluster/nodes/node.go
BenTheElder: FTR this increases disk footprint at runtime, but not so much cpu / memory.

BenTheElder:
[FYI I am doing local testing and working on a couple improvements / avoiding some regression]

dims:
Excellent! thanks @BenTheElder feel free to adjust this PR or a new one.

images/base/Dockerfile
aojea: yeah, I could check that if you set `root = "/var/lib/docker"` in `/etc/containerd/config.toml` 
 containerd allows using the overlayfs snapshotter

`sed -i 's/lib\/containerd/lib\/docker/' config.toml`

BenTheElder:
Further experimentation shows that docker + cli is > 200 MB of additional weight in the base image which is considerable, an ubuntu 19.04 containerd image is ~327MB, a docker image image was > 550MB.

aojea:
I think that's been superseded by https://github.com/kubernetes-sigs/kind/pull/461

dims:
@aojea indeed! thanks!

/close

PR #429, opened 2019-04-03
patch e2e.sh now that kubernetes uses go modules

BenTheElder:
/priority critical-urgent


BenTheElder:
Will put in a better patch soon. Need to decide how best to determine we actually found the source.

PR #434, opened 2019-04-05
lookup kuberoot correctly

BenTheElder:
this _should_ fix our issues with finding k8s.io/kubernetes in module mode afaict, however this does still require that the user checkout kubernetes to `$(go env GOPATH)/src/k8s.io/kubernetes` or else specify `--kube-root`


BenTheElder:
@neolit123 go computes this as either $GOPATH or the platform default. If it's not there the user will have to specify, but also lots of the main repo tools will break anyhow.

PR #439, opened 2019-04-09
haproxy: prepull the image during build

neolit123:
Currently the haproxy image is being pulled when
haproxy is being run on the LB nodes.

To prevent that, prepull the image and support air gapped
HA setups. This requires finding what tag was preloaded
for the haproxy image.

In case no haproxy tags are found in the node-image
pull the default tag.

/kind feature
/priority important-longterm

### DIFF
--- a/pkg/build/node/const.go
+++ b/pkg/build/node/const.go
@@ -23,6 +23,9 @@ const (
 	defaultCNIManifestLocation = "/kind/manifests/default-cni.yaml"
 )
 
+// defaultHAProxyImage defines the haproxy image:tag
+const defaultHAProxyImage = "haproxy:1.8.14-alpine"
+
 /*
 The default CNI manifest and images are from weave currently.
 
--- a/pkg/build/node/node.go
+++ b/pkg/build/node/node.go
@@ -391,8 +391,11 @@ func (c *BuildContext) prePullImages(dir, containerID string) error {
 		return err
 	}
 
-	// all builds should isntall the default CNI images currently
+	// all builds should install the default CNI images currently
 	requiredImages = append(requiredImages, defaultCNIImages...)
+
+	// install the haproxy image
+	requiredImages = append(requiredImages, defaultHAProxyImage)
 
 	// Create "images" subdir.
 	imagesDir := path.Join(dir, "bits", "images")
--- a/pkg/cluster/internal/create/actions/loadbalancer/haproxy.go
+++ b/pkg/cluster/internal/create/actions/loadbalancer/haproxy.go
@@ -21,12 +21,14 @@ import (
 	"fmt"
 
 	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
 
 	"sigs.k8s.io/kind/pkg/cluster/constants"
 	"sigs.k8s.io/kind/pkg/cluster/internal/create/actions"
 	"sigs.k8s.io/kind/pkg/cluster/internal/haproxy"
 	"sigs.k8s.io/kind/pkg/cluster/internal/kubeadm"
 	"sigs.k8s.io/kind/pkg/cluster/nodes"
+	"sigs.k8s.io/kind/pkg/exec"
 )
 
 // Action implements and action for configuring and starting the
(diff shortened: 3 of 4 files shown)


pkg/build/node/const.go
```diff
@@ -23,6 +23,9 @@ const (
 	defaultCNIManifestLocation = "/kind/manifests/default-cni.yaml"
 )
 
+// DefaultHAProxyImage defines the haproxy image:tag
+const DefaultHAProxyImage = "haproxy:1.8.14-alpine"
```
neolit123: should this be in a better location?

pkg/build/node/node.go
```diff
@@ -391,9 +391,12 @@ func (c *BuildContext) prePullImages(dir, containerID string) error {
 		return err
 	}
 
-	// all builds should isntall the default CNI images currently
+	// all builds should install the default CNI images currently
 	requiredImages = append(requiredImages, defaultCNIImages...)
 
+	// install the haproxy image
+	requiredImages = append(requiredImages, DefaultHAProxyImage)
```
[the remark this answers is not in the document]
tao12345666333: should we add logic to distinguish the node's role? 🤔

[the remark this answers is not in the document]
BenTheElder: different node images would be more expensive unless we were careful to actually use layers. the current build does not

pkg/build/node/const.go
BenTheElder: this should not be exported.

pkg/cluster/internal/create/actions/loadbalancer/haproxy.go
```diff
@@ -97,7 +98,7 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 		"/bin/sh", "-c",
 		fmt.Sprintf(
 			"docker run -d -v /kind/haproxy.cfg:/usr/local/etc/haproxy/haproxy.cfg:ro --network host --restart always --name haproxy %s",
-			haproxy.Image,
+			node.DefaultHAProxyImage,
```
BenTheElder: if we're going to pre-pull, this needs to come from the node. otherwise this will break between kind versions.

BenTheElder:
thanks. I think this needs some changes though.

also fwiw, unrelated to this PR but I'm not a big fan of using dind to run this, there have to be better options.

pkg/cluster/internal/create/actions/loadbalancer/haproxy.go
neolit123: do you mean that the node image build has to import the haproxy package to get the image constant; not the other way around?

neolit123:
> also fwiw, unrelated to this PR but I'm not a big fan of using dind to run this, there have to be better options.

HA in general?

BenTheElder:
Haproxy

pkg/cluster/internal/create/actions/loadbalancer/haproxy.go
BenTheElder: i mean that when we run this container it should match what the image was built with

neolit123: ok i see. so this is a problem..
one possible way of solving that is writing the haproxy image+tag on the node in here:
https://github.com/kubernetes-sigs/kind/blob/978f4a0bc666cc701680eb99da621e1e88448206/pkg/build/node/node.go#L416
(instead of %d)

and then the haproxy container is run we can use the image+tag from there.
but kind of messy.

neolit123: added a different workaround.

neolit123:
@BenTheElder @aojea 

> running haproxy as a process instead of inside a container? or using other LB?

so i had a look at this again.

if we have consensus on haproxy as a process that it is doable, yet it's unclear what should be used to install the haproxy binary apt-get vs wget + dpkg?

...but if we run haproxy as a process we break compatibility between the kind master branch and users of the latest node-image @ dockerhub. are we OK with that?

in terms of other popular LB solutions, most of those are containerized too.
in the meantime i made the code guess which haproxy version is loaded on the node and use that tag instead of pulling a new image.

BenTheElder:
We run an nginx image as the loadbalancer node now, users never had any guarantee about how this was implemented.

PR #446, opened 2019-04-17
No proxy docker network and localhost addresses

aojea:
On environments that require proxies to work, is a common issue
to forget setting the NO_PROXY environment variable to avoid the
kind nodes to try to use the proxy for the internal communications.
This patch automatically adds the docker network and the localhost
addresses to the NO_PROXY variable in case the environment needs
a proxy.

Ref https://github.com/kubernetes-sigs/kind/issues/440#issuecomment-483867395


aojea:
@BenTheElder please check if this makes sense, it improves the UX based on the last issues related to the proxy
/assign @BenTheElder

pkg/cluster/nodes/node.go
```diff
@@ -279,6 +286,28 @@ func (n *Node) IP() (ip string, err error) {
...
+func (n *Node) IPPrefix() (ipPrefix string, err error) {
+	// use the cached version first
+	cachedIPPrefix := n.cache.IPPrefix()
+	if cachedIPPrefix != "" {
+		return cachedIPPrefix, nil
+	}
+	// retrive the subnet prefix of the node using docker inspect
+	lines, err := docker.Inspect(n.name, "{{range .NetworkSettings.Networks}}{{.IPPrefixLen}}{{end}}")
+	if err != nil {
+		return "", errors.Wrap(err, "failed to get file")
```
BenTheElder: Wrong error message?

pkg/cluster/nodes/node.go
```diff
@@ -279,6 +286,28 @@ func (n *Node) IP() (ip string, err error) {
...
+	}
+	if len(lines) != 1 {
+		return "", errors.Errorf("file should only be one line, got %d lines", len(lines))
```
BenTheElder: It's not a file?

pkg/cluster/nodes/node.go
```diff
@@ -398,13 +427,29 @@ func NeedProxy() bool {
 // See also: NeedProxy and getProxyDetails
 func (n *Node) SetProxy() error {
 	details := getProxyDetails()
+	// don't proxy localhost and docker network
+	ip, err := n.IP()
```
BenTheElder: Why should we need to do the IP and the prefix? 🤔

pkg/cluster/nodes/node.go
aojea: yeah, I rushed it a bit, copy paste thing

pkg/cluster/nodes/node.go
aojea: ditto

pkg/cluster/nodes/node.go
aojea: I thin kthat we need to obtain the docker network CIDR so we can add it to the NO_PROXY var , and because we are into a node context the only info that we have is the current node ip and the prefix len, this way we can build the docker network as `IPAddress/IPPrefixLen`

```
        "NetworkSettings": {
            "Bridge": "",
(snip)
            "Gateway": "172.17.0.1",
            "GlobalIPv6Address": "2001:db8:1::242:ac11:4",
            "GlobalIPv6PrefixLen": 64,
            "IPAddress": "172.17.0.4",
            "IPPrefixLen": 16,
(snip)
```

From a networking point of view, it´s perfectly valid as CIDR, don´t know if this can cause any issue with docker

pkg/cluster/nodes/node.go
```diff
@@ -279,6 +286,28 @@ func (n *Node) IP() (ip string, err error) {
 	return ip, nil
 }
 
+// IPPrefix returns the Subnet prefix of the node
+func (n *Node) IPPrefix() (ipPrefix string, err error) {
```
BenTheElder: The docker package should implement IPPrefix, at some point the nodes package should only use a shim interface to {docker,podman,a fake...}

pkg/cluster/nodes/node.go
BenTheElder: Ok I see that now, can we explicitly build the CIDR in a separate function and just consume than here?

aojea:
@BenTheElder can you take a look at the new approach?

aojea:
No longer needed after https://github.com/kubernetes-sigs/kind/pull/461

BenTheElder:
I think we still need to add these values to `NO_PROXY` before passing it to the node 🤔

aojea:
Do you think that is possible to add it to the entrypoint script?
https://github.com/kubernetes-sigs/kind/blob/60e4f1a560f48faec2afb68053cfc7db3d7e3600/images/base/entrypoint#L27

Something like:
```
    # Don't use proxy for the local network
    LOCAL_NETWORK=$(ip route list dev eth0 proto kernel | awk '{ print $1 }')

    if [[ -z "${NO_PROXY}" ]]; then
        NO_PROXY="${LOCAL_NETWORK}"
    else
        NO_PROXY="${NO_PROXY},${LOCAL_NETWORK}"
    fi
```

BenTheElder:
> Do you think that is possible to add it to the entrypoint script?

ohhh interesting.

do you expect there'd ever be a case where we'd want to let the user bypass the values computed in the entrypoint?

images/base/entrypoint
```diff
@@ -19,12 +19,15 @@ set -o nounset
...
+  # Don't use proxy for the local network
+  LOCAL_NETWORK=$(ip route list dev eth0 proto kernel | awk '{ print $1 }')
+
   mkdir -p /etc/systemd/system/containerd.service.d/
   cat <<EOF >/etc/systemd/system/containerd.service.d/http-proxy.conf
 [Service]
 Environment="HTTP_PROXY=${HTTP_PROXY:-}"
 Environment="HTTPS_PROXY=${HTTPS_PROXY:-}"
-Environment="NO_PROXY=${NO_PROXY:-}"
+Environment="NO_PROXY=${NO_PROXY:-localhost},${LOCAL_NETWORK}"
```
BenTheElder: - probably this should be localhost and 127... ?
- do we expect the user to ever want to be able to not NO_PROXY LOCAL_NETWORK for some reason?
- do we ever expect to want to proxy localhost / 127?

aojea: I wouldn´t be surprised if there is some application in the node trying to reach localhost and the connection get proxied, maybe etcd or kubelet, I think that´s harmless anyway and it simplifies the script using them as default value.

LOCAL_NETWORK is directly connected, also is a docker network only with containers. I can´t see any reason to use a proxy to reach those IPs (however, I´m sure some user will find an exception ;) )

BenTheElder: OK, SGTM :-)

BenTheElder:
I will push a new commit with a base image from this so we can test that just to be safe 😉

images/base/entrypoint
```diff
@@ -19,12 +19,15 @@ set -o nounset
 set -o pipefail
 
 configure_proxy() {
+  # Don't use proxy for the local network
+  LOCAL_NETWORK=$(ip route list dev eth0 proto kernel | awk '{ print $1 }')
```
BenTheElder: minor q: is `ip route list` the same as `ip route show`? so far can't find manpages for list :^)

BenTheElder:
rebased and then added a commit to bump the base image.

images/base/entrypoint
aojea: I'm more used to `ip route show` too, however the output of the ip help doesn't mention show
```
ip route help
Usage: ip route { list | flush } SELECTOR
       ip route save SELECTOR
       ip route restore
       ip route showdump
       ip route get [ ROUTE_GET_FLAGS ] ADDRESS
```

PR #453, opened 2019-04-24
Ppc64le support

mkumatag:
This PR will
- Enable ppc64le architecture
- Expose docker-version for base-image command which will enable the user to select the docker version he wants to install.


mkumatag:
> what motivated you to expose the docker version?
> @BenTheElder has expressed that the way he wants users to customize the base image is to create a new base image.

@neolit123 the reason behind exposing docker-version is because we don't have that latest version docker on ppc64le platform, so basically I didn't want to have a hardcode some other version of docker for ppc64le instead what to give luxury to the user to decide what version of docker they really want to use.

This will even open up user to select their own version of docker if someone is not really comfortable with the one what we are shipping!

mkumatag:
> @BenTheElder has expressed that the way he wants users to customize the base image is to create a new base image.

And here also user will end up building the `base-image` with docker-version he wants to use(build time argument to `build base-image` command)

BenTheElder:
I'd rather not make docker-version a flag, if only because we're about to switch, I will put up a WIP PR later, the only thing that has prevented me from submitting has been continuing to try to avoid this being a breaking change. The upsides are probably worth breaking though since we're in alpha so perhaps it's time to stop delaying that.

images/base/Dockerfile
```diff
@@ -76,14 +76,16 @@ RUN curl -fsSL "https://download.docker.com/linux/$(. /etc/os-release; echo "$ID
     && clean-install "docker-ce=${DOCKER_VERSION}"
 
 # Enable CRI for containerd
-RUN mkdir -p /etc/docker && echo '{"cri-containerd": true}' > /etc/docker/daemon.json
-RUN mkdir -p /etc/containerd && containerd config default > /etc/containerd/config.toml
+RUN if [ -x "$(command -v containerd)" ]; then \
```
BenTheElder: does PPCLE not have recent enough packages? I'm also planning to upgrade the _Actual_ base image in that WIP containerd upgrade. (to 19.04)

images/base/Dockerfile
```diff
@@ -76,14 +76,16 @@ RUN curl -fsSL "https://download.docker.com/linux/$(. /etc/os-release; echo "$ID
...
+  mkdir -p /etc/docker && echo '{"cri-containerd": true}' > /etc/docker/daemon.json;\
+  mkdir -p /etc/containerd && containerd config default > /etc/containerd/config.toml;\
+fi
 
 # Install CNI binaries to /opt/cni/bin
 # TODO(bentheelder): doc why / what here
 ARG CNI_VERSION="0.7.5"
 ARG CNI_BASE_URL="https://storage.googleapis.com/kubernetes-release/network-plugins/"
-RUN export ARCH=$(dpkg --print-architecture) \
+RUN export ARCH=$(dpkg --print-architecture| sed 's/ppc64el/ppc64le/') \
```
BenTheElder: oh that is wildly irritating. can we get that fixed??

mkumatag: there is been already concerned raised about this and here is the explanation given - https://lists.debian.org/debian-powerpc/2014/08/msg00042.html, we don’t have any other go, will have to live with this as of now.

images/base/Dockerfile
mkumatag: docker-ce 18.06.3 is the latest package we have so far in the docker website for power and we have opened and we have opened an issue with docker community to fix this issue - https://github.com/docker/for-linux/issues/629

BenTheElder:
I filed an updated change here https://github.com/kubernetes-sigs/kind/pull/463

Hopefully that should be sufficient! AFAICT all of the packages should be available on PPC upstream now :sweat_smile:

PR #454, opened 2019-04-25
Use kind as product_name in the nodes

### LINKED ISSUE #426: kubelet may detect GCE when running kind on GCE
**What happened**: kubelet appears to have read `/sys/class/dmi/id/product_name` when running kind on GCE and detected "Google Compute Engine", it appears to fail if it can't talk to the metadata server.

**What you expected to happen**: Kubernetes should not "be in cloud mode" so to speak.

**How to reproduce it (as minimally and precisely as possible)**: run a kind cluster on GCE

**Environment:**

- kind version: (use `kind version`): 0.2.1
- Kubernetes version: (use `kubectl version`): 1.13.4

aojea:
Kubernetes clusters use the  `product_name` for some features and that makes `kind` fail

Fixes #426

### DIFF
--- a/pkg/cluster/nodes/node.go
+++ b/pkg/cluster/nodes/node.go
@@ -231,6 +231,10 @@ func (n *Node) FixMounts() error {
 	if err := n.Command("mount", "--make-shared", "/var/lib/docker").Run(); err != nil {
 		return err
 	}
+	// use kind as product name to avoid issues with features that depends on it
+	if err := n.Command("mount", "-o", "ro,bind", "/kind/product_name", "/sys/class/dmi/id/product_name").Run(); err != nil {
+		return err
+	}
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/cluster/nodes/node.go
```diff
@@ -231,6 +231,10 @@ func (n *Node) FixMounts() error {
 	if err := n.Command("mount", "--make-shared", "/var/lib/docker").Run(); err != nil {
 		return err
 	}
+	// use kind as product name to avoid issues with features that depends on it
+	if err := n.Command("mount", "-o", "ro,bind", "/kind/product_name", "/sys/class/dmi/id/product_name").Run(); err != nil {
```
BenTheElder: this will fail on old node images awkwardly. worth considering

aojea: @BenTheElder fixed in the new commit

BenTheElder:
thanks for this, per your suggestion implemented in the entrypoint changes in  https://github.com/kubernetes-sigs/kind/pull/461

Issue #456, opened 2019-04-26
fix the flakes in "kind export logs"


neolit123:
**What happened**:
`kind export logs` can flake on very rare occasions.
me and @BenTheElder have seen this in CI before and it still happens.

**What you expected to happen**:
no flakes.

**How to reproduce it (as minimally and precisely as possible)**:
- create a 1CP2W cluster using `kind create cluster`
- write this to a bash script:

```bash
#!/bin/bash
TMPDIR="$(mktemp -d)"
for i in {1..1000}; do
    echo "----------------- $i"
    kind export logs "${TMPDIR}" || exit 1
    rm -rf "${TMPDIR}/*.*"
done
```
- call this script and check on it periodically (it will run for a while...)
- the command will fail on very rare occasions
- if i extend the logging a bit i'm seeing:
```
Error: error in execToPath for: kind-control-plane/pods/0820b847-67a1-11e9-b78f-0242c21cba65/coredns/0.log: exit status 126
```
or
```
Error: error in execToPath for: kind-worker2/kubernetes-version.txt: exit status 126
```

i think the 126 comes from:
https://github.com/kubernetes-sigs/kind/blob/master/pkg/cluster/logs/logs.go#L113
but not sure why, possibly a file gets an exclusive lock for a period of time.

**Anything else we need to know?**:
no

**Environment:**
any?

- kind version: latest master
- Kubernetes version: does not matter
- Docker version: doesn't seem to matter
- OS doesn't seem to matter

/assign
/kind bug
/priority important-longterm

aojea:
googling the error exit status 126 says
>  126 as exit status which means "Command was found but could not be executed".
it can be possible that the errors is because we are not using the full path in the commands?

neolit123:
126 can also be related to permissions.
https://www.tldp.org/LDP/abs/html/exitcodes.html

> Permission problem or command is not an executable

BenTheElder:
https://docs.docker.com/engine/reference/run/#/exit-status

aojea:
I´m trying to reproduce it without success, my observation is that `kind` runs in parallel several goroutines 
```
21802 pts/4    Sl+    0:00  |       \_ ./kind export logs /tmp/tmp.KIXat1dxEE
22151 pts/4    Sl+    0:00  |           \_ docker exec --privileged -t kind-control-plane cat /var/log/pods/514670304442e821ba1e3d444ab0a987/kube-apiserver/0.log
[...]

PR #458, opened 2019-04-27
Use tar to archive logs

aojea:
`kind export logs` can flake on very rare occasions.
In order to export all logs and follow the symlinks we are exporting
one file at a time. This seems to cause sporadic issues with
`docker exec` logging exit status 126 errors.

This patch exports the /var/log folder of the nodes using tar
to reduce the number of streams needed.

Partially solves #456

### DIFF
--- a/pkg/cluster/logs/logs.go
+++ b/pkg/cluster/logs/logs.go
@@ -17,14 +17,17 @@ limitations under the License.
 package logs
 
 import (
+	"archive/tar"
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+	"io"
 	"os"
+	osexec "os/exec"
 	"path/filepath"
-	"strings"
-	"sync"
-
 	"sigs.k8s.io/kind/pkg/cluster/nodes"
 	"sigs.k8s.io/kind/pkg/exec"
 	"sigs.k8s.io/kind/pkg/util"
+	"sync"
 )
 
 type errFn func() error
@@ -65,8 +68,24 @@ func Collect(nodes []nodes.Node, dir string) error {
 	// add a log collection method for each node
 	for _, n := range nodes {
 		node := n // https://golang.org/doc/faq#closures_and_goroutines
+		name := node.String()
+		// grab all logs under /var/log (pods and containers)
+		cmd := osexec.Command("docker", "exec", name, "tar", "--hard-dereference", "-C", "/var/log", "-chf", "-", ".")
+		stdout, err := cmd.StdoutPipe()
+		if err != nil {
+			return err
+		}
+		if err := cmd.Start(); err != nil {
+			return err
+		}
+		log.Debugf("Running: %v %v", cmd.Path, cmd.Args)
+		if err := untar(stdout, filepath.Join(dir, name)); err != nil {
+			return errors.Wrapf(err, "Untarring %q: %v", name, err)
+		}
+		if err := cmd.Wait(); err != nil {
+			return err
+		}
 		fns = append(fns, func() error {
-			name := node.String()
 			return coalesce(
 				// record info about the node container
 				execToPathFn(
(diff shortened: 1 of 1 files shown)


aojea:
I have to test it more but during my first tests it improves the time to export too

pkg/cluster/logs/logs.go
```diff
@@ -152,3 +145,92 @@ func coalesce(fns ...errFn) error {
...
+			}
+			nFiles++
+		case mode.IsDir():
+			if err := os.MkdirAll(abs, 0755); err != nil {
+				return err
+			}
+			madeDir[abs] = true
+		case mode&os.ModeSymlink != 0:
+			// TODO: ignore these for now. They were breaking x/build tests.
+			// Implement these if/when we ever have a test that needs them.
```
BenTheElder: I'm not sure we can just "adapt" this much code under our own license...

neolit123: this can be used instead:
https://medium.com/@skdomino/taring-untaring-files-in-go-6b07cf56bc07

aojea: Yeah, indeed seems too restrictive https://github.com/golang/build/blob/master/LICENSE, however is not a big chunk of code and there are other several projects using the same structure. Will refactor it, I don't think it can be considered as derivatives work

aojea: updated de untar code based on @neolit123 suggestion

pkg/cluster/logs/logs.go
```diff
@@ -65,8 +68,24 @@ func Collect(nodes []nodes.Node, dir string) error {
 	// add a log collection method for each node
 	for _, n := range nodes {
 		node := n // https://golang.org/doc/faq#closures_and_goroutines
+		name := node.String()
+		// grab all logs under /var/log (pods and containers)
+		cmd := osexec.Command("docker", "exec", node.Name(), "tar", "--hard-dereference", "-C", "/var/log", "-chf", "-", ".")
```
BenTheElder: why aren't we using the container cmder here?

BenTheElder: I would prefer to isolate which code execs to docker, we're going to shim that soon and it will be much easier to fix if it is in less places.

BenTheElder:
please squash these :sweat_smile:

BenTheElder:
/ok-to-test
you should apply to join kubernetes-sigs :-)

pkg/cluster/logs/logs.go
aojea: There are 2 reasons because I couldn´t use the container cmder:
* it doesn´t have implemented .StdoutPipe()
* it executes docker with the `-t` that allocates a tty and  makes the `tar` command to fail

Do you see an alternative to implementing StdoutPipe() ?

> I would prefer to isolate which code execs to docker, we're going to shim that soon and it will be much easier to fix if it is in less places.

You mean docker insider the nodes or docker everywhere?

aojea: > /ok-to-test
> you should apply to join kubernetes-sigs :-)

Sure, that will be great, do you know if is this process the one I have to follow? https://github.com/kubernetes/community/blob/master/community-membership.md

Do I need 2 sponsors?

aojea:
> /ok-to-test
> you should apply to join kubernetes-sigs :-)

Sure, that will be great, do you know if is this process the one I have to follow? https://github.com/kubernetes/community/blob/master/community-membership.md

Do I need 2 sponsors?

aojea:
> i would have approached the logging code in the following way:
> 
> * spawn N go routines for N nodes
> * in each go routine:
>   
>   * docker exec write journalctl, docker info and other logs in a tmp folder
>   * docker exec tar everything + the /var/log folder to an archive
>   * docker cp the archive from node to host
>   * delete tmp files and tar from container
> 
> it might end up being slightly slower than the current code.

@neolit123 you mean to not untar the files on the host? is this expected for the CI jobs?

neolit123:
@aojea 
> @neolit123 you mean to not untar the files on the host? is this expected for the CI jobs?

missed a step. untar to the host.

pkg/cluster/logs/logs.go
BenTheElder: > Do you see an alternative to implementing StdoutPipe() ?

stdout should not be a problem to pipe, we collect stdout regularly.

> it executes docker with the -t that allocates a tty and makes the tar command to fail

the actual code does:
```go
	if c.stderr != nil || c.stdout != nil {
		args = append(args,
			"-t", // use a tty so we can get output
		)
	}
```

we can make this code handle our case

BenTheElder: > You mean docker insider the nodes or docker everywhere?

everywhere. we need to be able to use a shim eventually so the code is testable and can support podman properly.

using the node cmder will also make it easier to fake out the node.

BenTheElder: in what way does tar fail with a TTY?

pkg/cluster/logs/logs.go
```diff
@@ -65,8 +68,24 @@ func Collect(nodes []nodes.Node, dir string) error {
...
+		// grab all logs under /var/log (pods and containers)
+		cmd := osexec.Command("docker", "exec", name, "tar", "--hard-dereference", "-C", "/var/log", "-chf", "-", ".")
+		stdout, err := cmd.StdoutPipe()
+		if err != nil {
+			return err
+		}
+		if err := cmd.Start(); err != nil {
+			return err
+		}
+		log.Debugf("Running: %v %v", cmd.Path, cmd.Args)
```
BenTheElder: cmder should already be logging this to debug?

pkg/cluster/logs/logs.go
BenTheElder: I got this working pretty nicely with the node cmder, breaking it into multiple PRs, first https://github.com/kubernetes-sigs/kind/pull/473

BenTheElder: see: https://github.com/kubernetes-sigs/kind/compare/master...BenTheElder:pr458?expand=1

https://github.com/kubernetes-sigs/kind/pull/474 adds the other required peice

aojea:
continue on https://github.com/kubernetes-sigs/kind/compare/master...BenTheElder:pr458?expand=1

PR #461, opened 2019-04-29
Images Improvements [Breaking Changes]

BenTheElder:
Overall this PR will make cluster boot faster and lighter, however it is a breaking change to the images versus the kind binary, newer images will require a newer kind binary and a newer kind binary will require newer images.

Base Image Changes:
- switch to Ubuntu 19.04 base (newer dependencies)
- switch from dockerd to containerd (lighter, faster, easier to debug)
- remove unnecessary packages
- add crictl v1.14.0
- remove more unnecessary files in clean-install & fix shellcheck lint
- entrypoint is now a small shellscript instead of a go binary
  - easier portability / reproducibility (COPY only, no host building)
  - also smaller, less dependency

Node Image Changes:
- preload images in to containerd's content store instead of storing archives (smaller & faster!)

Cluster Create Changes:
- stop waiting for dockerd to be ready (containerd boots ~instantly)
- stop loading tar archives (already loaded at build time)
- No more node fixups on the host side (proxy settings, mounts etc.. entrypoint now handles these)
  - No more SIGUSR1
- product_name is overridden

The entrypoint changes should make #148 much simpler, we only need to tackle the IP issues.


BenTheElder:
Just making this a breaking change cut down complexity quite a bit, will clean this up some more along with some other fixes and get it in to unblock other work like #453

BenTheElder:
[these will fail until a base image is pushed, also I think < 1.12 bazel
build image tags needs addressing, though perhaps we don't need to keep
supporting v1.11?]

images/base/Dockerfile
```diff
@@ -34,18 +32,18 @@ RUN chmod +x /usr/local/bin/clean-install
...
 # Then we cleanup (removing unwanted systemd services)
 # Finally we disable kmsg in journald
 # https://developers.redhat.com/blog/2014/05/05/running-systemd-within-docker-container/
 RUN clean-install \
-      apt-transport-https ca-certificates curl software-properties-common gnupg2 lsb-release \
       systemd systemd-sysv libsystemd0 \
-      conntrack iptables iproute2 ethtool socat util-linux mount ebtables udev kmod aufs-tools \
-      bash rsync \
+      containerd \
+      conntrack iptables iproute2 ethtool socat util-linux mount ebtables udev kmod \
```
aojea: we need gnupg2 in order to build the node images with apt

```
/kind build node-image --base-image kindest/base:containerd --image kindest/node:containerd --type apt
E: gnupg, gnupg2 and gnupg1 do not seem to be installed, but one of them is required for this operation
(23) Failed writing body
ERRO[22:37:41] Adding Kubernetes apt key failed! exit status 255
ERRO[22:37:41] Image build Failed! exit status 255
Error: error building node image: exit status 255
```

BenTheElder:
We need to remove the apt node tbh.

pkg/build/node/node.go
```diff
@@ -308,7 +308,19 @@ func (c *BuildContext) buildImage(dir string) error {
 	}
```
aojea: We should enable containerd service before this, shouldn´t we?
`systemctl enable containerd`

BenTheElder: we _did_ need to enable containerd but _not_ for this. we run containerd one-off during the node image build, not under systemd. less moving parts, in the future we'll only do the image import using the API directly

pkg/build/node/node.go
```diff
@@ -255,7 +255,7 @@ func (c *BuildContext) buildImage(dir string) error {
 		}()
 	}
 	if err != nil {
-		log.Errorf("Image build Failed! %v", err)
+		log.Errorf("Image build Failed! Failed to create build container: %v", err)
```
BenTheElder: I wound up adding some more details to these to aid debugging while working on this. They're not strictly related, but it's a small change.

pkg/cluster/nodes/node.go
```diff
@@ -221,16 +154,6 @@ func (n *Node) FixMounts() error {
 	if err := n.Command("mount", "-o", "remount,ro", "/sys").Run(); err != nil {
```
BenTheElder: this should probably move into ENTRYPOINT and we can get rid of SIGUSR1, unfortunately there's the userns remap hackery so we need to plumb through info there first.

might do that in this PR so we can have a single breaking image change point

pkg/cluster/internal/create/nodes.go
```diff
@@ -159,15 +158,6 @@ func fixupNode(node *nodes.Node) error {
...
 	}
 
-	// wait for docker to be ready
-	if !node.WaitForDocker(time.Now().Add(time.Second * 60)) {
-		// TODO(bentheelder): logging here
-		return errors.Errorf("timed out waiting for docker to be ready on node %s", node.Name())
-	}
-
-	// load the docker image artifacts into the docker daemon
-	node.LoadImages()
```
BenTheElder: the equivilant is now done at image build time

pkg/cluster/internal/create/actions/config/config.go
```diff
@@ -57,7 +57,7 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 	kubeVersion, err := node.KubeVersion()
 	if err != nil {
 		// TODO(bentheelder): logging here
-		return errors.Wrap(err, "failed to get kubernetes version from node: %v")
+		return errors.Wrap(err, "failed to get kubernetes version from node")
```
BenTheElder: this formatting was incorrect

pkg/cluster/nodes/node.go
BenTheElder: I did this

images/base/Dockerfile
BenTheElder: alternatively type=apt can install gnupg2 first if we decide not to get rid of it.

pkg/cluster/nodes/nodes.go
```diff
@@ -134,3 +134,14 @@ func WaitForReady(node *Node, until time.Time) bool {
 		return true
 	})
 }
+
+// helper that calls `try()`` in a loop until the deadline `until`
+// has passed or `try()`returns true, returns wether try ever returned true
+func tryUntil(until time.Time, try func() bool) bool {
```
BenTheElder: this is the same code, but the only user is over here now (directly above) so I moved it.

pkg/cluster/nodes/create.go
```diff
@@ -182,16 +180,5 @@ func createNode(name, image, clusterLabel, role string, mounts []cri.Mount, extr
 		return handle, errors.Wrap(err, "docker run error")
 	}
 
-	// Deletes the machine-id embedded in the node image and regenerate a new one.
-	// This is necessary because both kubelet and other components like weave net
-	// use machine-id internally to distinguish nodes.
-	if err := handle.Command("rm", "-f", "/etc/machine-id").Run(); err != nil {
```
BenTheElder: entrypoint does this now

pkg/cluster/internal/create/nodes.go
```diff
@@ -137,40 +131,6 @@ func createNodeContainers(
 	return nil
 }
 
-func fixupNode(node *nodes.Node) error {
-	// we need to change a few mounts once we have the container
-	// we'd do this ahead of time if we could, but --privileged implies things
-	// that don't seem to be configurable, and we need that flag
-	if err := node.FixMounts(); err != nil {
```
BenTheElder: entrypoint does this now

pkg/cluster/internal/create/nodes.go
```diff
@@ -137,40 +131,6 @@ func createNodeContainers(
...
-		// TODO(bentheelder): logging here
-		return err
-	}
-
-	if nodes.NeedProxy() {
```
BenTheElder: entrypoint does this now

pkg/cluster/internal/create/nodes.go
```diff
@@ -137,40 +131,6 @@ func createNodeContainers(
...
-		if err := node.SetProxy(); err != nil {
-			// TODO: logging here
-			return errors.Wrapf(err, "failed to set proxy for node %s", node.Name())
-		}
-	}
-
-	// signal the node container entrypoint to continue booting into systemd
-	if err := node.SignalStart(); err != nil {
```
BenTheElder: there's nothing to signal now

pkg/cluster/nodes/node.go
```diff
@@ -116,12 +112,6 @@ func (n *Node) Name() string {
 	return n.name
 }
 
-// SignalStart sends SIGUSR1 to the node, which signals our entrypoint to boot
-// see images/node/entrypoint
-func (n *Node) SignalStart() error {
-	return docker.Kill("SIGUSR1", n.name)
```
BenTheElder: we should probably remove the docker.Kill call since this was the only user afaik. that can maybe be another PR though, this PR is already a bit big

pkg/cluster/nodes/node.go
```diff
@@ -135,105 +125,6 @@ func (n *Node) CopyFrom(source, dest string) error {
 	return docker.CopyFrom(n.name, source, dest)
 }
 
-// WaitForDocker waits for Docker to be ready on the node
-// it returns true on success, and false on a timeout
-func (n *Node) WaitForDocker(until time.Time) bool {
```
BenTheElder: unnecessary now, containerd starts very very very fast in testing so far (not worth measuring), also we don't need to wait to load images, because we don't need to load images

pkg/cluster/nodes/node.go
```diff
@@ -135,105 +125,6 @@ func (n *Node) CopyFrom(source, dest string) error {
...
-	for until.After(time.Now()) {
-		if try() {
-			return true
-		}
-	}
-	return false
-}
-
-// LoadImages loads image tarballs stored on the node into docker on the node
-func (n *Node) LoadImages() {
```
BenTheElder: image loading is at build time now

pkg/cluster/nodes/node.go
```diff
@@ -135,105 +125,6 @@ func (n *Node) CopyFrom(source, dest string) error {
...
-		fmt.Sprintf(`docker images --format='{{.Repository}}:{{.Tag}}' | grep -v %s | xargs -L 1 -I '{}' /bin/bash -c 'docker tag "{}" "$(echo "{}" | sed s/:/-%s:/)"'`,
-			util.GetArch(), util.GetArch()),
-	).Run(); err != nil {
-		log.Warningf("Failed to re-tag docker images: %v", err)
-	}
-}
-
-// FixMounts will correct mounts in the node container to meet the right
-// sharing and permissions for systemd and Docker / Kubernetes
-func (n *Node) FixMounts() error {
```
BenTheElder: done in entrypoint now

pkg/cluster/nodes/node.go
```diff
@@ -384,35 +275,6 @@ func (n *Node) WriteFile(dest, content string) error {
 	return n.Command("cp", "/dev/stdin", dest).SetStdin(strings.NewReader(content)).Run()
 }
 
-// NeedProxy returns true if the host environment appears to have proxy settings
-func NeedProxy() bool {
```
BenTheElder: we handle this in create / entrypoint now

pkg/cluster/nodes/node.go
```diff
@@ -384,35 +275,6 @@ func (n *Node) WriteFile(dest, content string) error {
...
-	return len(details.Envs) > 0
-}
-
-// SetProxy configures proxy settings for the node
-//
-// Currently it only creates systemd drop-in for Docker daemon
-// as described in Docker documentation: https://docs.docker.com/config/daemon/systemd/#http-proxy
-//
-// See also: NeedProxy and getProxyDetails
-func (n *Node) SetProxy() error {
```
BenTheElder: we handle this in create / entrypoint now

BenTheElder:
pushed a base image, this shed ~60MB **compressed** versus the last base image (according to docker hub)

BenTheElder:
v1.11.1 + bazel requires some work around for the image names. We could stop supporting 1.11 but we should probably declare a support policy other than 1.11+ then...

images/base/entrypoint
```diff
@@ -0,0 +1,58 @@
...
+    chmod -s /bin/mount
+
+    # systemd-in-a-container should have read only /sys
+    # https://www.freedesktop.org/wiki/Software/systemd/ContainerInterface/
+    # however, we need other things from `docker run --privileged` ...
+    # and this flag also happens to make /sys rw, amongst other things
+    mount -o remount,ro /sys
+}
+
+fix_machine_id() {
```
aojea: @BenTheElder I think it will be great if you can add a new function here to fix the `product_name` issue and get rid of https://github.com/kubernetes-sigs/kind/pull/454

pkg/cluster/internal/create/actions/kubeadmjoin/join.go
```diff
@@ -36,11 +36,11 @@ import (
 	"sigs.k8s.io/kind/pkg/fs"
 )
 
-// Action implements action for creating the kubeadm config
+// Action implements action for creating the kubeadm join
 // and deployng it on the bootrap control-plane node.
 type Action struct{}
 
-// NewAction returns a new action for creating the kubadm config
+// NewAction returns a new action for creating the kubeadm jion
```
aojea: nit: typo s/jion/join/

BenTheElder:
/hold
need to fix more things :-)

pkg/cluster/internal/create/actions/kubeadmjoin/join.go
BenTheElder: thanks

images/base/entrypoint
BenTheElder: SGTM, will do this tomorrow

BenTheElder:
The v1.11.X + bazel fix is working now, we rewrite the image archives to have corrected tags.

TODO:
- [ ] fixup `kind load ...`
- [ ] fixup HAproxy (might leave that for a followup)

images/base/entrypoint
BenTheElder: done

BenTheElder:
/hold cancel
will need follow-ups but this should move things forward a lot and help unblock, will follow-up with:

- code cleanup (mostly removing some now dead-code)
- HA rework
- docs updates
- `kind build node-image --type=apt` fixes
- ...

pkg/cluster/nodes/node.go
```diff
@@ -135,105 +126,6 @@ func (n *Node) CopyFrom(source, dest string) error {
...
-	}
-
-	// systemd-in-a-container should have read only /sys
-	// https://www.freedesktop.org/wiki/Software/systemd/ContainerInterface/
-	// however, we need other things from `docker run --privileged` ...
-	// and this flag also happens to make /sys rw, amongst other things
-	if err := n.Command("mount", "-o", "remount,ro", "/sys").Run(); err != nil {
-		return err
-	}
-	// kubernetes needs shared mount propagation
```
nickolaev: @BenTheElder these are omitted in `fix_mount` in the entrypoint. We are adopting `kind` as a cluster provider in Network Service Mesh https://github.com/networkservicemesh/networkservicemesh/blob/master/docs/guide-kind.md and the latest version is not working for us. The issues is that we are using `/var/lib` mounts which is relying on `/` being a shared mount.
What would be the best approach here? Shall we consider updating `fix_mounts`?

BenTheElder: we can update `fix_mounts`, sorry about that!

These were a hold over from some previous docker-in-docker tooling predating kind, I haven't found much (anything?) documenting why you'd need to do this on the kubernetes host and conformance was passing without them, so I removed them as voodoo figuring kubelet must handle this 🤦‍♂ 
Looking again now I do see: https://kubernetes.io/docs/concepts/storage/volumes/#configuration 

-----

side note:

> It is worth noting that kind as any other Kubernetes deployment tool would expect that the machine that hosts the Docker has at least 4 CPU cores and 4 GB of RAM. That is specifically pointed for OSX users in the official docs.

Filed https://github.com/kubernetes-sigs/kind/issues/485, we should make those more accurate in our docs and then send y'all an update 🙃

BenTheElder: filed https://github.com/kubernetes-sigs/kind/issues/486, will send a fix momentarily

nickolaev:
Thanks, @BenTheElder, indeed I have found these mentions about the `MountFlags=shared` all over the place but was puzzled how this maps to `kind`. Apparently, it needs to be done manually here.

PR #463, opened 2019-04-30
fixes for ppc64le

BenTheElder:
based on https://github.com/kubernetes-sigs/kind/pull/453, but updated after the breaking image changes

FYI @dims @mkumatag


mkumatag:
@BenTheElder This is great thanks a lot. One minor comment, you may need to sed https://github.com/kubernetes-sigs/kind/blob/master/images/base/Dockerfile#L77 as well

BenTheElder:
ah yes, we should fix that :sweat_smile:

PR #465, opened 2019-05-01
remove skip from e2e script

BenTheElder:
/hold

### DIFF
--- a/hack/ci/e2e.sh
+++ b/hack/ci/e2e.sh
@@ -142,12 +142,16 @@ run_tests() {
     )"
 
     # ginkgo regexes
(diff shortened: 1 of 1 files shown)


BenTheElder:
It's not, just serial and focus. Around 1.12 skip was removed from conformance finally.

neolit123:
> Around 1.12 skip was removed from conformance finally.

you mean tests with tags like Alpha, Kubectl, Disruptive were removed?

BenTheElder:
I mean conformance was only supposed to have a focus regex, not a skip, but
it did until 1.12 because some tests managed to get tagged [Conformance]
but also with some of those undersireable tags.

When submitting results we've had to make sure not to include the skip, I
accidentally did when first submitting 1.14

PR #467, opened 2019-05-01
more cleanup

BenTheElder:
- improve some code comments / add more
- trim the metadata we set for kustomize patching from the generated kubeadm config output before using it

### DIFF
--- a/pkg/cluster/internal/create/actions/config/config.go
+++ b/pkg/cluster/internal/create/actions/config/config.go
@@ -19,6 +19,7 @@ package config
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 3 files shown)


BenTheElder:
ah, older Go does not have strings.ReplaceAll, switched to strings.Replace
with -1

pkg/cluster/internal/create/actions/config/config.go
```diff
@@ -111,7 +112,24 @@ func getKubeadmConfig(cfg *config.Cluster, data kubeadm.ConfigData) (path string
...
+	patched, err := kustomize.Build([]string{config}, patches, jsonPatches)
+	if err != nil {
+		return "", err
+	}
+	return removeMetadata(patched), nil
+}
+
+// trims out the metadata.name we put in the config for kustomize matching,
+// kubeadm will complain about this otherwise
+func removeMetadata(kustomized string) string {
```
neolit123: nice, this should get rid of the unmarshal warnings.

pkg/container/docker/archive.go
```diff
@@ -77,8 +77,17 @@ func GetArchiveTags(path string) ([]string, error) {
...
 }
 
-// EditArchiveRepositories applies edit to reader's repositories file
-func EditArchiveRepositories(reader io.Reader, writer io.Writer, edit func(string) string) error {
+// EditArchiveRepositories applies edit to reader's image repositories,
+// IE the repository part of repository:tag in image tags
+// This supports v1 / v1.1 / v1.2 Docker Image Archives
+//
+// editRepositories should be a function that returns the input or an edited
+// form, where hte input is the image repository
```
neolit123: typo hte

BenTheElder: fixed

PR #469, opened 2019-05-01
[WIP] Run HAProxy as a node

BenTheElder:
Follow-up to #461, this unbreaks HA mode and simplifies it.

Now we run an haproxy image directly as the external load-balancer node.

To do this we have to get around a few hurdles:
- #468 so we can avoid specifying the entrypoint for nodes (and use the normal haproxy entrypoint)
- we need to have a valid config in the haproxy image always or the container will exit (WIP: our own image with a minimal config added, per the upstream docs https://hub.docker.com/_/haproxy?tab=description)
- we need to send SIGHUP to the node to reload the config after writing it instead of starting a container on the node

### DIFF
--- a/pkg/cluster/internal/haproxy/const.go
+++ b/pkg/cluster/internal/haproxy/const.go
@@ -20,4 +20,4 @@ package haproxy
 const ControlPlanePort = 6443
 
 // Image defines the haproxy image:tag
-const Image = "haproxy:1.8.14-alpine"
+const Image = "kindest/haproxy:0.1.0" //"haproxy:1.8.14-alpine"
--- a/pkg/cluster/nodes/create.go
+++ b/pkg/cluster/nodes/create.go
@@ -141,8 +141,6 @@ func createNode(name, image, clusterLabel, role string, mounts []cri.Mount, extr
 		"--label", clusterLabel,
 		// label the node with the role ID
 		"--label", fmt.Sprintf("%s=%s", constants.NodeRoleKey, role),
-		// explicitly set the entrypoint
-		"--entrypoint=/usr/local/bin/entrypoint",
 	}
 
 	// pass proxy environment variables to be used by node's docker deamon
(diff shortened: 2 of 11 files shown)


BenTheElder:
FYI @fabriziopandini, this will be more straightforward I think and more inline with your initial concept, but it will be another breaking change for kinder I suspect.

TODO is publishing a cross-platform haproxy-with-minimal-config image and referencing that in the code rather than my local WIP build.

cc also @neolit123

pkg/container/docker/kill.go
```diff
@@ -0,0 +1,31 @@
+/*
```
BenTheElder: this file is not actually new, it was removed due to being dead code and brought back now so we can use it for haproxy. I'll probably make this a node level API instead in a follow-up cleanup.

pkg/cluster/nodes/util.go
```diff
@@ -31,14 +31,16 @@ func GetControlPlaneEndpoint(allNodes []Node) (string, error) {
...
 		return "", err
 	}
-	// if there is no external load balancer
+	// if there is no external load balancer return the empty string
 	if node == nil {
 		return "", nil
 	}
-	// gets the IP of the load balancer
+
+	// get the IP and port for the load balancer
```
BenTheElder: TODO: seems to be issues if we use the mapped port instead of 6443. need to look into that, this is fine for now though (we just "get" the fixed port)

pkg/cluster/nodes/create.go
```diff
@@ -141,8 +141,6 @@ func createNode(name, image, clusterLabel, role string, mounts []cri.Mount, extr
 		"--label", clusterLabel,
 		// label the node with the role ID
 		"--label", fmt.Sprintf("%s=%s", constants.NodeRoleKey, role),
-		// explicitly set the entrypoint
-		"--entrypoint=/usr/local/bin/entrypoint",
```
BenTheElder: we can remove this and the args below now that we have #468

this allows us to start a loadbalancer image identically to the other nodes

pkg/cluster/internal/haproxy/const.go
```diff
@@ -20,4 +20,4 @@ package haproxy
 const ControlPlanePort = 6443
 
 // Image defines the haproxy image:tag
-const Image = "haproxy:1.8.14-alpine"
+const Image = "my-haproxy:latest" //"haproxy:1.8.14-alpine"
```
BenTheElder: TODO: clean up this image in images/ and push it, then update this ref
/hold

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/470 seems much better

PR #470, opened 2019-05-01
nginx external loadbalancer

BenTheElder:
- use nginx for load balancing (smaller config and the image handles delayed creation of config cleanly)
- run the load balancer image directly instead of using docker in docker
- eliminate trying to capture the containerd id from the stdout of `docker run` (unrelated bug discovered while using a not-yet pulled image while building this)
 - use a UUID for the node image build name and use the name instead of ID
- s/haproxy/loadbalancer/ widely throughout the code

Thanks to @aojea for the suggestion, this is #469 rewritten to just use nginx instead :-)


BenTheElder:
FYI @neolit123 @fabriziopandini

BenTheElder:
this is required to unblock 0.3, as currently HA is broken

BenTheElder:
follow-up to #461

PR #475, opened 2019-05-02
Use Tar to Archive logs

BenTheElder:
filing this based on https://github.com/kubernetes-sigs/kind/pull/458 + leveraging https://github.com/kubernetes-sigs/kind/pull/473 and https://github.com/kubernetes-sigs/kind/pull/474 per @aojea


aojea:
/lgtm
can we consider that fixes #456 ?

BenTheElder:
I think so :-)

Issue #466, opened 2019-05-01
cleanup image import at build time


BenTheElder:
**What should be cleaned up or changed**: Instead of ignoring all errors (due to expected unpacking errors) when running `ctr image import` at node-image build time we should leverage https://github.com/containerd/containerd/pull/3259 to import only without unpack when that or an alternative is available. 

See https://github.com/containerd/containerd/issues/3258 for the upstream feature request.

**Why is this needed**: Node image building has very spammy logs right now, and we should not be ignoring errors

/priority important-soon
/assign
cc @Random-Liu

BenTheElder:
If we can't get this feature into the `ctr` command line, we can instead build our own tiny client binary for this. The API client in fact has `Import` and `image.Unpack` as separate API calls, the handy upstream CLI just happens to do both, as it probably makes sense in most use cases.

Right now we effectively only do `Import` as the way we run the build ensures that `Unpack` will fail and ~ no-op, as a result we potentially ignore legitimate import errors however.

PR #477, opened 2019-05-03
clean up image importing in node-build

BenTheElder:
This PR:
- adds tooling (scripts / cloudbuild) used to publish `ctr` binaries for all architectures built from https://github.com/BenTheElder/containerd/tree/kind which contains https://github.com/BenTheElder/containerd/commit/cb7c780af2394ab08d5d8a3932ca7437074ae179 (while we wait for this to upstream and eventually be packaged)
- updates the base image to install this `ctr` binary
- bumps the base image
- updates the node-image build to use the `--no-unpack` option added in these binaries
- disables some containerd plugins that we do not use that will not load (snapshotters for zfs etc.)
  - should reduce logspam and drop some (probably trivial) resource consumption

fixes: #466


hack/build/ctr/cloudbuild.yaml
```diff
@@ -0,0 +1,28 @@
+# simple cloudbuild to build github.com/contianerd/containerd/cmd/ctr
```
BenTheElder: I don't expect anyone else to need to use `hack/build/ctr/...`, but in the interests of transparency I'm attempting to create a reproducible (?) cross-platform build of `ctr` with the tiny patch we are waiting on for use in the base image, and NOT build it from my workstation :^)

BenTheElder: also: we will stop using these once we can get this patch packaged upstream, or we will figure something else out. 

being able to use this patch now means we can do image import properly without errors and handle errors.

BenTheElder:
no more spurious errors in the image build logs :^) 🎉

images/base/Dockerfile
```diff
@@ -74,7 +98,7 @@ RUN export ARCH=$(dpkg --print-architecture| sed 's/ppc64el/ppc64le/') \
 
 # Install crictl to /usr/local/bin
 ARG CRICTL_VERSION="v1.14.0"
-RUN export ARCH=$(dpkg --print-architecture) \
+RUN export ARCH=$(dpkg --print-architecture | sed 's/ppc64el/ppc64le/') \
```
BenTheElder: FYI @mkumatag I think I fixed the last of these, the `ctr` binary here is also built by us and _should_ have a working PPC64LE download, but I can't test it :^)

BenTheElder: (the existing binary and package from canonical should also already support PPC64LE, but I took care to make sure our temporary custom one does too)

BenTheElder:
We have 3 upstream maintainer LGTMs on the patch now, but it will be some time before that merges / is packaged.

hack/build/ctr/run.sh
```diff
@@ -0,0 +1,72 @@
...
+cd "${CONTAINERD_SOURCE}"
+git checkout "${CONTAINERD_BRANCH}"
+
+# TODO(bentheelder): this is hacky, compute what the makefile would have computed
+# so we can just go build the one package with our own flags
+# https://github.com/containerd/containerd/blob/a17c8095716415cebb1157a27db5fccace56b0fc/Makefile#L22-L24
+VERSION=$(git describe --match 'v[0-9]*' --dirty='.m' --always)
+REVISION=$(git rev-parse HEAD)$(if ! git diff --no-ext-diff --quiet --exit-code; then echo .m; fi)
+
+# we need to upload .git for version detection, so add a blank .gcloudignore
```
BenTheElder: we actually don't know, but i don't think changing this right now is worth rebuilding the binaries

hack/build/ctr/run.sh
```diff
@@ -0,0 +1,72 @@
...
+# options
+PROJECT="${PROJECT:-bentheelder-kind-dev}"
+BUCKET="${BUCKET:-bentheelder-kind-dev/containerd}"
+# https://github.com/BenTheElder/containerd/tree/kind
+# https://github.com/containerd/containerd/pull/3259
+CONTAINERD_SOURCE="${CONTAINERD_SOURCE:-$(go env GOPATH)/src/github.com/containerd/containerd}"
+CONTAINERD_BRANCH="${CONTAINERD_BRANCH:-kind}"
+CTR_CLOUDBUILD="${REPO_ROOT}/hack/build/ctr/cloudbuild.yaml"
+
+# make sure we submit the build to the righr project
```
neolit123: `right`

neolit123: not sure why check-spelling doesn't catch some of these.
feel free to ignore.

/lgtm
/hold

BenTheElder: fixed, went ahead and removed the .gcloudignore bit as well since we no longer need that

hack/build/ctr/run.sh
BenTheElder: fixed

neolit123:
/lgtm
read the upstream patch, SGTM.

BenTheElder:
Will incrementally switch back to upsteam ASAP,
First build from upsteam directly after merge, then back to the packaged binary once that's new enough.

Issue #111, opened 2018-11-14
Use container IP address in server field in kubeconfig for kind cluster


font:
When working with multiple clusters such as with https://github.com/kubernetes-sigs/federation-v2, we provide the cluster info to the federation controller running in one of the clusters, to talk to all the other clusters. That cluster info comes from the kubeconfig provided by the user. The current `kind create cluster` command will create a kubeconfig with the `server` field using `localhost` as shown below:

```bash
$ KUBECONFIG=$(kind get kubeconfig-path) kubectl config view | grep server
    server: https://localhost:38785
```

This makes it problematic when one kind cluster needs to access the kube API server in another kind cluster because `localhost` for each kind cluster refers to itself. Since the kind clusters share a common docker network bridge, we can use the IP address for each container. So what is needed is for the container IP address to be used in the server field in the kubeconfig for the kind cluster as shown below:

```bash
$ KUBECONFIG=$(kind get kubeconfig-path) kubectl config view | grep server
    server: https://172.17.0.2:38785
```

The host would still be able to talk to the kube API servers using their container IP addresses. Is this something that is desirable? Is there a reason to use `localhost` instead?

munnerz:
On OSX at least, it's not possible to access the docker container network 😞

If we want to make this configurable, there's 3 options I can think of:

- Add a boolean/toggle to the kind Config type
- Add a flag to `kind create cluster` to toggle the behaviour
- Add a new `kind get kubeconfig` to grab a kubeconfig for an existing cluster, and add a flag to *that*

Not sure what's most desirable here? 😄

BenTheElder:
Each cluster has a different port for the API server so it should still be possible for them to talk to each other (?), the reason for localhost is that windows / mac do not support anything else as far as I know, because the container network is actually in a light VM.

BenTheElder:
It's also possible that using a single network bridge will be problematic in the future, once we get multiple nodes. Using localhost and a random port makes things pretty portable.

Can federation support keying off of the address + port instead of just the address instead?

It would be quite easy to create a kubeconfig with the container IP address when on linux, but so far I've preferred keeping the environment as consistent as possible. Ideally users should be able to replicate CI very closely on their local machines, which is a major goal for kind.

font:
> Each cluster has a different port for the API server so it should still be possible for them to talk to each other (?), the reason for localhost is that windows / mac do not support anything else as far as I know, because the container network is actually in a light VM.

[...]

PR #478, opened 2019-05-03
Add option to print raw kubeconfig

aojea:
For some use cases, i.e federation and multinode clusters, the kubeconfig with the internal ip addresses is needed. Currently, kind exports the kubeconfig replacing the node ip address with localhost and the port that is forwarded to the host.

This adds an option to print the raw kubeconfig from a cluster with 
`kind get kubeconfig --name cluster` if you want the host kubeconfig or

`kind get kubeconfig --name cluster --internal` if you want the kubeconfig inside the nodes (with the internal  IP addresses)

Fixes https://github.com/kubernetes-sigs/kind/issues/111 #566

### DIFF
--- a/cmd/kind/get/get.go
+++ b/cmd/kind/get/get.go
@@ -21,6 +21,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"sigs.k8s.io/kind/cmd/kind/get/clusters"
+	"sigs.k8s.io/kind/cmd/kind/get/kubeconfig"
 	"sigs.k8s.io/kind/cmd/kind/get/kubeconfigpath"
 	"sigs.k8s.io/kind/cmd/kind/get/nodes"
 )
@@ -31,12 +32,13 @@ func NewCommand() *cobra.Command {
 		Args: cobra.NoArgs,
 		// TODO(bentheelder): more detailed usage
 		Use:   "get",
-		Short: "Gets one of [clusters, nodes, kubeconfig-path]",
-		Long:  "Gets one of [clusters, nodes, kubeconfig-path]",
+		Short: "Gets one of [clusters, nodes, kubeconfig, kubeconfig-path]",
+		Long:  "Gets one of [clusters, nodes, kubeconfig, kubeconfig-path]",
 	}
 	// add subcommands
 	cmd.AddCommand(clusters.NewCommand())
 	cmd.AddCommand(nodes.NewCommand())
+	cmd.AddCommand(kubeconfig.NewCommand())
 	cmd.AddCommand(kubeconfigpath.NewCommand())
 	return cmd
 }
--- a/cmd/kind/get/kubeconfig/kubeconfig.go
+++ b/cmd/kind/get/kubeconfig/kubeconfig.go
@@ -1,0 +1,95 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package kubeconfig implements the `kubeconfig` command
+package kubeconfig
+
+import (
+	"fmt"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"io/ioutil"
+
+	"sigs.k8s.io/kind/pkg/cluster"
+	clusternodes "sigs.k8s.io/kind/pkg/cluster/nodes"
+	"sigs.k8s.io/kind/pkg/exec"
+)
+
+type flagpole struct {
+	Name     string
+	Internal bool
+}
+
+// NewCommand returns a new cobra.Command for getting the kubeconfig with the internal node IP address
+func NewCommand() *cobra.Command {
+	flags := &flagpole{}
+	cmd := &cobra.Command{
+		Args:  cobra.NoArgs,
+		Use:   "kubeconfig",
+		Short: "prints cluster kubeconfig",
+		Long:  "prints cluster kubeconfig",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runE(flags, cmd, args)
+		},
+	}
+	cmd.Flags().StringVar(
+		&flags.Name,
+		"name",
+		cluster.DefaultName,
+		"the cluster context name",
+	)
+	cmd.Flags().BoolVar(
+		&flags.Internal,
+		"internal",
+		false,
+		"get internal kubeconfig",
+	)
+	return cmd
+}
+
+func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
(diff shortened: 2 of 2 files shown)


cmd/kind/get/kubeconfig/kubeconfig.go
```diff
@@ -0,0 +1,72 @@
...
+		"name",
+		cluster.DefaultName,
+		"the cluster context name",
+	)
+	return cmd
+}
+
+func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
+	// List nodes by cluster context name
+	n, err := clusternodes.ListByCluster("label=io.k8s.sigs.kind.role=control-plane")
```
BenTheElder: CMD shouldn't know about docker code, we should use the pattern used in actions.

cmd/kind/get/kubeconfig/kubeconfig.go
```diff
@@ -0,0 +1,72 @@
...
+	if err != nil {
+		return err
+	}
+	nodes, known := n[flags.Name]
+	if !known {
+		return errors.Errorf("unknown cluster %q", flags.Name)
+	}
+	// grab kubeconfig version from one of the control plane nodes
+	cmdNode := nodes[0].Command("cat", "/etc/kubernetes/admin.conf")
```
BenTheElder: The default output should not be raw, it should be what we'd write to disk normally when creating the cluster

aojea: Can you elaborate more?
I did it so you can pipe the output 
```
$ ./kind get kubeconfig  >  test.txt
$ cat test.txt
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUN5RENDQWJDZ0F3SUJBZ0lCQURBTkJna3Foa2lHOXcwQkFRc0ZBREFWTVJNd0VRWURWUVFERXdwcmRXSmwK
Y201bGRHVnpNQjRYRFRFNU1EVXdNekV3TVRZME5Wb1hEVEk1TURRek1ERXdNVFkwTlZvd0ZURVRNQkVHQTFVRQpBeE1LYTNWaVpYSnVaWFJsY3pDQ0FTSXdEUVlKS29aSWh2Y05BUUVCQlFBRGdnRVBBREND
QVFvQ2dnRUJBTjBlClJ3RElkTEF2Vll4SEpKeDhyeFFWME5VcjZ4OHlTczgwN1RGbWhBWDRqbDFJSFZ6Q0pFM3pNd2E2WDM0WXpaTTcKbUVPeC9ySEpTelFtMEZVRWhZSlN5UzIwQWZWbTdkOGZrV3yVmxPZkMK
U3VXbHl3cExUSU9DRTQ1NVpwNGFzek1TeUEvakdiUi8rSklpUmZpTFNWa0RXdTU1VlpGUUdLNEp4V05zRjJhcwo0WSt2dGVRUWlKdGZ6K3N5d09hZU5lTFhyZWNKaU9LRkRZeVJqenJibjhKaDEvRUE2Ym1P
WkZLdUk5THlUcTgyCnpmLy9SYnlYVmxwZGpzRXhQNFN5dFlSWE1KSTJsS2VybjBpTnozOVp5ZlhRL3ozT3kyZERqUm9JK2lkMEVqak8KcEFkblhKZ3lZK0tnSjYxZ3hERGFZcDZmZk9ZOFUzbnhsVjUydGpS
    server: https://172.17.0.2:6443
  name: kind
contexts:
- context:
    cluster: kind
    user: kubernetes-admin
  name: kubernetes-admin@kind
current-context: kubernetes-admin@kind
kind: Config
preferences: {}
users:
- name: kubernetes-admin
  user:
```

cmd/kind/get/kubeconfig/kubeconfig.go
aojea: done

cmd/kind/get/kubeconfig/kubeconfig.go
```diff
@@ -0,0 +1,77 @@
...
+	return cmd
+}
+
+func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
+	// List nodes by cluster context name
+	n, err := clusternodes.ListByCluster()
+	if err != nil {
+		return err
+	}
+	nodes, known := n[flags.Name]
```
tao12345666333: thanks. we can just use  cluster.IsKnown(flags.Name) to do this.

cmd/kind/get/kubeconfig/kubeconfig.go
BenTheElder: we should have two modes:
- the same contents that kind would write during `kind create cluster`
- the raw contents from within the cluster, no kind specific modifications
   - I would rename "raw" to "internal", to clarify the purpose, we may still tweak it for other reasons in the future (perhaps rewriting the context name, user...)

cmd/kind/get/kubeconfig/kubeconfig.go
BenTheElder: IsKnown internally needs to do this. We should come back to this and make ListByCluster() accept a cluster name to filter on instead of arbitrary filters. this would be more efficient, but still put a precise API in pkg/

cmd/kind/get/get.go
```diff
@@ -32,11 +33,12 @@ func NewCommand() *cobra.Command {
 		// TODO(bentheelder): more detailed usage
 		Use:   "get",
 		Short: "Gets one of [clusters, nodes, kubeconfig-path]",
```
BenTheElder: this should also be in short. We should rewrite long at some point

cmd/kind/get/kubeconfig/kubeconfig.go
aojea: done

cmd/kind/get/get.go
aojea: done

aojea:
this is a small addition and seems that could solve some issues expressed in the channel lately

cmd/kind/get/kubeconfig/kubeconfig.go
```diff
@@ -0,0 +1,95 @@
...
+	if flags.Internal {
+		// grab kubeconfig version from one of the control plane nodes
+		cmdNode := node.Command("cat", "/etc/kubernetes/admin.conf")
+		exec.InheritOutput(cmdNode)
+		if err := cmdNode.Run(); err != nil {
+			return errors.Wrap(err, "failed to get cluster internal kubeconfig")
+		}
+	} else {
+		ctx := cluster.NewContext(flags.Name)
+		out, err := ioutil.ReadFile(ctx.KubeConfigPath())
```
BenTheElder: we shouldn't read this in only to turn around and print it back out, we can copy from the file directly to stdout. that can be a follow up though

cmd/kind/get/kubeconfig/kubeconfig.go
```diff
@@ -0,0 +1,95 @@
...
+		&flags.Name,
+		"name",
+		cluster.DefaultName,
+		"the cluster context name",
+	)
+	cmd.Flags().BoolVar(
+		&flags.Internal,
+		"internal",
+		false,
+		"get internal kubeconfig",
```
BenTheElder: TODO: a better description :sweat_smile:

BenTheElder:
> this is a small addition and seems that could solve some issues expressed in the channel lately

I agree the change is small but... what issue was expressed today that this solves? I must have missed this

PR #480, opened 2019-05-04
collect node serial logs

BenTheElder:
pretty small change, in particular this lets us collect the systemd init output.

### DIFF
--- a/pkg/cluster/logs/logs.go
+++ b/pkg/cluster/logs/logs.go
@@ -88,6 +88,10 @@ func Collect(nodes []nodes.Node, dir string) error {
 				),
 				// grab all of the node logs
 				execToPathFn(
+					exec.Command("docker", "logs", name),
+					filepath.Join(name, "serial.log"),
+				),
+				execToPathFn(
 					node.Command("cat", "/kind/version"),
 					filepath.Join(name, "kubernetes-version.txt"),
 				),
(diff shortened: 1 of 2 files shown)


pkg/cluster/logs/logs.go
```diff
@@ -87,6 +87,10 @@ func Collect(nodes []nodes.Node, dir string) error {
 					filepath.Join(name, "inspect.json"),
 				),
 				// grab all of the node logs
+				execToPathFn(
+					exec.Command("docker", "logs", name),
```
aojea: should we add `-t` to the command to have the timestamps?
I find them very useful

BenTheElder:
I think we want consistent logging across daemons, they're unlikely to be
useful in this scenario as they're all quickly after start (it's just init
messages, then nothing). The runs from this PR should have examples

*From: *Antonio Ojea <notifications@github.com>
*Date: *Fri, May 3, 2019, 23:38
*To: *kubernetes-sigs/kind
*Cc: *Benjamin Elder, Author

*@aojea* commented on this pull request.

pkg/cluster/logs/logs.go
BenTheElder: https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_kind/480/pull-kind-conformance-parallel-1-14/1124528321279823874/artifacts/logs/kind-control-plane/serial.log

probably not so useful. I do think we might want to tweak the init output though.

BenTheElder: also definitely need to refactor this out to pkg/container, have a bigger PR to refactor this code later.

PR #481, opened 2019-05-04
refactor and dedupe concurrency logic

BenTheElder:
we use these in a few places, this kind of code tends to be error prone, so let's dedupe it.

### DIFF
--- a/pkg/concurrent/concurrent.go
+++ b/pkg/concurrent/concurrent.go
@@ -1,0 +1,70 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


BenTheElder:
cc @aojea

BenTheElder:
naming is hard, I wanted to name this package `go` so we could do `go.UntilError` but of course that's reserved, so `concurrent.UntilError` it is. perhaps we can come up with more precise names in a follow up 🙃

pkg/concurrent/concurrent.go
```diff
@@ -0,0 +1,70 @@
...
+// first non-nil error returned from funcs, or nil if all funcs return nil
+func UntilError(funcs []func() error) error {
+	errCh := make(chan error, len(funcs))
+	for _, f := range funcs {
+		f := f // capture f
+		go func() {
+			errCh <- f()
+		}()
+	}
+	for i := 0; i < len(funcs); i++ {
```
aojea: just curious, this `for` loop is the one that waits for all goroutines to finish, isn´t it?

BenTheElder: UntilError returns the first error immediately without waiting.

aojea: ic, and if no errors it waits for all functions to send the err results. Thanks

BenTheElder: right, we can't return `nil` correctly until every func has returned, but we can return an error immediately, this one is meant for quick bail-out (EG for cluster creation code), while we have `Coalesce` for when we want to continue and collect the errors (as in collecting logs).

PR #482, opened 2019-05-04
Add ppc64le to release script

mkumatag:
/cc @BenTheElder @dims


BenTheElder:
Thanks, do we know that it works on PPC now? 😁

mkumatag:
> hoping for confirmation that it is in fact working, but we can always release with a caveat in the release notes 😅

good question! I see few issues, will submit the PRs once I'm able to debug.

PR #483, opened 2019-05-04
add systemd restart override

BenTheElder:
In case the service exits, we want to ensure that it restarts.

The ubuntu package currently has (appears to be very similar to upstream):
```
[Unit]
Description=containerd container runtime
Documentation=https://containerd.io
After=network.target

[Service]
ExecStartPre=-/sbin/modprobe overlay
ExecStart=/usr/bin/containerd

Delegate=yes
KillMode=process
# Having non-zero Limit*s causes performance problems due to accounting overhead
# in the kernel. We recommend using cgroups to do container-local accounting.
LimitNPROC=infinity
LimitCORE=infinity
LimitNOFILE=infinity
# Comment TasksMax if your systemd version does not supports it.
# Only systemd 226 and above support this version.
TasksMax=infinity

[Install]
WantedBy=multi-user.target
```

cc @Random-Liu, perhaps something else to push upstream at some point ...


images/base/10-restart.conf
```diff
@@ -0,0 +1,8 @@
+# ensure the service is restarted
+[Service]
+Restart=always
+RestartSec=1
+
+# disable rate limiting
```
aojea: why disabling the rate limiting?
I experienced problems because a service is restarting continuously but the check wasn´t able to detect it down , can this happen to us?

BenTheElder: It starts in under a second, the default rate limit of 5 in 10s doesn't seem useful in the context of kind, if we can determine a better one for kind we should do that.

We have no checks for this, instead we're observing if the workloads are up.

kubelet and containerd are our only services.

PR #484, opened 2019-05-05
Using bridge network for nodes

tao12345666333:
ref: https://github.com/kubernetes-sigs/kind/pull/408#issuecomment-488653739 and https://github.com/kubernetes-sigs/kind/pull/408#issuecomment-488765565

we can using bridge network for nodes. after this we can using hostname instead of IPs. 
It will push #408 forward.

things need todo:

- [x] create bridge network
- [x] create nodes using bridge netwrok

<s>- [ ] using hostname instead of IPs (config files.)</s>

- [x] load-balancer config

- [x] delete network while deleting cluster.


BenTheElder:
https://github.com/kubernetes-sigs/kind/issues/273#issuecomment-488874160

BenTheElder:
The custom bridge network's magical hostname resolution is causing coreDNS to detect a loop and fail.

tao12345666333:
It seems that I missed some useful information. 😯

neolit123:
> The custom bridge network's magical hostname resolution is causing coreDNS to detect a loop and fail.

hm, the following is *not* an option:
- editing the coredns Corefile to modify the forward or disable the coredns loop detection.

we need to think of a way to avoid the loop.
docker dns configuration flags:
https://docs.docker.com/v17.09/engine/userguide/networking/default_network/configure-dns/

/cc
/cc @aojea

aojea:
The DNS loop seems caused because the docker DNS uses always the address `127.0.0.11` as quoted from the docker docs

> Note: The DNS server is always at 127.0.0.11

and that´s the address configured on the node`s `resolv.conf`
```

nameserver 127.0.0.11
options ndots:0
```
CoreDNS documentation explains why such addresses causes a loop https://github.com/coredns/coredns/blob/master/plugin/loop/README.md

> A common cause of forwarding loops in Kubernetes clusters is an interaction with a local DNS cache on the host node (e.g. systemd-resolved). For example, in certain configurations systemd-resolved will put the loopback address 127.0.0.53 as a nameserver into /etc/resolv.conf. Kubernetes (via kubelet) by default will pass this /etc/resolv.conf file to all Pods using the default dnsPolicy rendering them unable to make DNS lookups (this includes CoreDNS Pods). CoreDNS uses this /etc/resolv.conf as a list of upstreams to forward requests to. Since it contains a loopback address, CoreDNS ends up forwarding requests to itself.

It offers 3 solutions, but personally I only see feasible the 1st want mentioned modifying the kubelet yaml. Let me try to see what´s needed to implement it and come back to you.

In the other hand, I don´t know if creating a different bridge per cluster will break some functionality, I´m especially concerned about the multicluster/federation scenarios. I think it will be safer first to create a network for kind and then see if we need to move to a network per cluster

aojea:
@tao12345666333 in order to keep backward compatibility and avoid the DNS crash we should:
1. Copy the resolv.conf from the host to /kind/resolv.conf per example filtering all the loopback addresses (this is what Docker is doing when using the default bridge https://docs.docker.com/v17.09/engine/userguide/networking/default_network/configure-dns/)
> When creating the container’s /etc/resolv.conf, the daemon filters out all localhost IP address nameserver entries from the host’s original file.

2. Tell kubelet to use  /kind/resolv.conf for Pods.
Seems [kubeadm has a flag to configure this,](https://github.com/kubernetes/kubernetes/blob/c103ba914e53e15dfdd74a01aad1ebdfdf2bcccd/cmd/kubeadm/app/util/config/testdata/conversion/controlplane/v1beta1_non_linux.yaml#L162),

tao12345666333:
@aojea Thank you for the information, I will try it later.

BenTheElder:
does docker only rewrite etc/resolv.conf on container creation or continually? if the former, then we can just overwrite that?

otherwise we will have inconsistent DNS for system components.

aojea:
> does docker only rewrite etc/resolv.conf on container creation or continually? if the former, then we can just overwrite that?
> 
> otherwise we will have inconsistent DNS for system components.

if you overwrite the resolv.conf of the nodes you lose the Docker DNS and the host discovery. I mean, there is no reason to move to user defined bridges

BenTheElder:
Getting our own IP space for the nodes is still useful, and we could
selectively merge the resolv.conf

*From: *Antonio Ojea <notifications@github.com>
*Date: *Sun, May 5, 2019, 23:16
*To: *kubernetes-sigs/kind
*Cc: *Benjamin Elder, Review requested

does docker only rewrite etc/resolv.conf on container creation or

tao12345666333:
This fix can work at my side.

pkg/cluster/nodes/create.go
```diff
@@ -75,6 +75,11 @@ func CreateControlPlaneNode(name, image, clusterLabel, listenAddress string, por
 		cache.ports = map[int32]int32{kubeadm.APIServerPort: port}
 	})
 
+	// to avoid the DNS crash we need copy host's /etc/resolv.conf to node
+	// ref: https://github.com/kubernetes-sigs/kind/pull/484#issuecomment-489469044
+
+	node.CopyTo("/etc/resolv.conf", "/kind/resolv.conf")
```
aojea: we should filter the loopback addresses instead of copying it directly, or people that have loopback addresses in their resolv.conf will hit the DNS crass due to the loop.

tao12345666333: thanks will update

aojea:
It seems I misunderstood the use of the resolv-conf flag in kubelet 😅 

>May 06 11:05:34 kind-worker kubelet[200]: E0506 11:05:34.447977     200 kuberuntime_manager.go:693] createPodSandbox for pod "weave-net-khpnb_kube-system(0ebd9f5c-6fec-11e9-8010-0242ac120002)" failed: open /kind/resolv.conf: no such file or directory

The pods are not able to boot because they can´t find the new resolv.conf file. I was thinking that the file was copied inside each pod

pkg/cluster/nodes/create.go
aojea: from the errors found in the logs, it seems that has to be copied in all nodes, control plane and workers too

> ay 06 11:05:34 kind-worker kubelet[200]: E0506 11:05:34.447977     200 kuberuntime_manager.go:693] createPodSandbox for pod "weave-net-khpnb_kube-system(0ebd9f5c-6fec-11e9-8010-0242ac120002)" failed: open /kind/resolv.conf: no such file or directory

tao12345666333: Addressed. Thanks. :+1:

pkg/cluster/nodes/util.go
```diff
@@ -44,3 +46,23 @@ func GetControlPlaneEndpoint(allNodes []Node) (string, error) {
...
+// to avoid the DNS crash we need copy host's /etc/resolv.conf to node
+// ref: https://github.com/kubernetes-sigs/kind/pull/484#issuecomment-489469044
+func addResolve(node *Node) error {
+	resolv, err := ioutil.ReadFile("/etc/resolv.conf")
+	if err != nil {
+		return errors.Wrap(err, "failed to read /etc/resolv.conf")
+	}
+
+	// filter the loopback addresses
+	re := regexp.MustCompile("(?m)[\r\n]+^.*127.0.0.1.*$")
```
aojea: the loopback address network is 127.0.0.0/8 so we should filter all the address in that range, ie. from 127.0.0.1 to 127.255.255.255
I think that it should be something like this 127.\d+.\d+.\d+

tao12345666333: :+1:  Thank you for correcting me, I missed it.

BenTheElder:
> In the other hand, I don´t know if creating a different bridge per cluster will break some functionality, I´m especially concerned about the multicluster/federation scenarios. I think it will be safer first to create a network for kind and then see if we need to move to a network per cluster

If you look at my strawman proposal, the idea is to make it configurable so federation can configure the same bridge for all.

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -237,6 +239,7 @@ evictionHard:
   nodefs.available: "0%"
   nodefs.inodesFree: "0%"
   imagefs.available: "0%"
+resolvConf: /kind/resolv.conf
```
BenTheElder: I really don't like the idea that containerd and debugging tools will have a different resolv conf than kubelet.

tao12345666333: as you said 
> we could selectively merge the resolv.conf

tao12345666333: After considering it, I think the current practice is relatively simple. If we do resolv.conf merge, we need a lot of testing. Why don't we put it later? :thinking: 

Keep this practice for a while, then change it after the verification is passed.

I hope we can move forward. :heart:

BenTheElder: this means other components have an inconsistent resolv conf and we're not actually solving the issue. if we're going to use hostnames we need kubelet and everything else to have the same lookup.

pkg/cluster/internal/create/actions/loadbalancer/loadbalancer.go
```diff
@@ -71,11 +71,7 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 		return err
 	}
 	for _, n := range controlPlaneNodes {
-		controlPlaneIP, err := n.IP()
-		if err != nil {
-			return errors.Wrapf(err, "failed to get IP for node %s", n.Name())
-		}
-		backendServers[n.Name()] = fmt.Sprintf("%s:%d", controlPlaneIP, kubeadm.APIServerPort)
+		backendServers[n.Name()] = fmt.Sprintf("%s:%d", n.Name(), kubeadm.APIServerPort)
```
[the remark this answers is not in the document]
aojea: awesome, that will simplify a lot the IPv6 work too

tao12345666333:
The most basic features of this PR expectation have been achieved. Includes:

- [x] create bridge network
- [x] create nodes using bridge netwrok
- [x] load-balancer config
- [x] delete network while deleting cluster.

By using the bridge network, we can use Docker's embedded DNS to help push #408 forward.

After this PR, we only need to modify `/etc/kubernetes/pki/admin.conf` in order to deal with #408. We no longer need to regenerate the certificate because the current certificate is still available(with this PR).

i.e.

```
root@moe-control-plane3:/etc/kubernetes# openssl x509 -in pki/apiserver.crt -noout -text                        
Certificate:                                                                                                     
    Data:                                                                                        
        Version: 3 (0x2)                                                                                         
        Serial Number: 4829775455184542867 (0x4306cf2fef2cdc93)                                                 
        Signature Algorithm: sha256WithRSAEncryption                                                             
        Issuer: CN = kubernetes                                    
        Validity                                                                                                 
            Not Before: May  8 13:55:37 2019 GMT                             
            Not After : May  7 13:56:57 2020 GMT                                                                 
        Subject: CN = kube-apiserver                                              
        Subject Public Key Info:                                                                                 
            Public Key Algorithm: rsaEncryption                                           
                RSA Public-Key: (2048 bit)                                                                      
                Modulus:                                                                                        
                    00:c3:cb:9e:9f:e5:d4:bf:c7:b1:a8:77:98:95:e5:  
                    (ignore)
                    fb:66:2a:d0:ab:b1:23:30:25:39:e5:91:91:6a:f6:
                    a9:15
                Exponent: 65537 (0x10001)
        X509v3 extensions:
            X509v3 Key Usage: critical
                Digital Signature, Key Encipherment
            X509v3 Extended Key Usage:
                TLS Web Server Authentication
            X509v3 Subject Alternative Name:
                DNS:moe-control-plane3, DNS:kubernetes, DNS:kubernetes.default, DNS:kubernetes.default.svc, DNS:k
ubernetes.default.svc.cluster.local, DNS:localhost, IP Address:10.96.0.1, IP Address:172.28.0.6, IP Address:172.2
8.0.2
    Signature Algorithm: sha256WithRSAEncryption
         70:65:07:9c:54:50:24:9f:59:59:8b:de:4f:60:c9:4e:47:5f:
         (ignore)

```

I want to know your thoughts. @BenTheElder @aojea @neolit123 Thanks.

aojea:
Ben created a proposal, I think that only the labeling networks thing is missing 

https://github.com/kubernetes-sigs/kind/issues/273#issuecomment-488874160

> Strawman:
> 
> if the specified network exists, we use it without creating
> if it doesn't exist, we create it, and label it as belonging to the cluster
> on cluster delete, we list networks labeled with the cluster, and delete only those (so not just whatever the containers use, only if we labeled it)
> we name this field / functionality somehow such that it is clear that this feature is docker specific, leaving room for podman etc. in the immediate future xref #154.

tao12345666333:
Yes, I have seen this proposal.

Ben also reminded me. https://github.com/kubernetes-sigs/kind/pull/484#discussion_r281015531

 I am also very interested in implementing it. 

Just before I want to confirm your thoughts to ensure that our goals or directions are consistent.

pkg/cluster/internal/create/actions/loadbalancer/loadbalancer.go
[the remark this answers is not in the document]
BenTheElder: @aojea hmm, perhaps dual-stack support can just patch this later and have reduced restart support?

pkg/cluster/internal/create/nodes.go
```diff
@@ -88,6 +93,18 @@ func provisionNodes(
 	return nil
 }
 
+func createNetwork(status *logutil.Status, clusterName string) error {
+	defer status.End(false)
+
+	status.Start("Creating bridge network 🌉")
```
BenTheElder: excellent emoji :^)

pkg/cluster/internal/delete/delete.go
```diff
@@ -46,5 +46,10 @@ func Cluster(c *context.Context) error {
 		fmt.Printf("$KUBECONFIG is still set to use %s even though that file has been deleted, remember to unset it\n", c.KubeConfigPath())
 	}
 
-	return nodes.Delete(n...)
+	err = nodes.Delete(n...)
+	if err != nil {
```
BenTheElder: we should probably collect this error but still try to delete the network

pkg/cluster/nodes/create.go
```diff
@@ -142,6 +143,9 @@ func createNode(name, image, clusterLabel, role string, mounts []cri.Mount, extr
 		"--label", clusterLabel,
 		// label the node with the role ID
 		"--label", fmt.Sprintf("%s=%s", constants.NodeRoleKey, role),
+		// connect node to network
+		// TODO(Jintao Zhang): split just for get clustername.
+		"--network", strings.Split(name, "-")[0],
```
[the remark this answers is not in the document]
BenTheElder: clarifying: we compute this name mapping in a few places, we should have one function for this.

pkg/cluster/internal/delete/delete.go
```diff
@@ -46,5 +46,10 @@ func Cluster(c *context.Context) error {
...
+		return errors.Wrap(err, "error deleting nodes")
+	}
+	// after all nodes have been deleted, we need delete the network
+	return nodes.DeleteNetwork(c.Name())
```
BenTheElder: IMO this should be a `docker.DeleteNetWork`, why is this in nodes?

pkg/cluster/nodes/create.go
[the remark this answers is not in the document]
BenTheElder: re-reading: this should be passed in, instead of relying on splitting the name. this is brittle

pkg/cluster/internal/delete/delete.go
tao12345666333: the reason I originally wrote this is because if the node delete fails, it is very likely that its state is still active (Up). 
then deleting the network will also fail because `docker network rm` does not allow the removal of network with active endpoints.

pkg/cluster/internal/delete/delete.go
tao12345666333: Ok, then I moved both `CreateNetwork` and `DeleteNetwork` from `pkg/cluster/nodes` to `pkg/container/docker`.

pkg/cluster/internal/create/actions/loadbalancer/loadbalancer.go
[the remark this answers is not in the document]
aojea: yeah, sorry, I just wanted to raise the concern, using hostnames we are going to rely on the OS resolver and we should have predictable behaviors.
Can we add a TODO to remember it?
`// TODO(aojea): review for dual-stack`

pkg/cluster/internal/delete/delete.go
tao12345666333: Addressed.

pkg/cluster/internal/kubeadm/config.go
aojea: technically having a different resolv.conf doesn't mean that have a different lookup, just that the pods will not be able to resolv the nodes names, the other requests will use the same dns server, however I agree that reality is always different and we can be broken because docker changes on the DNS behaviour, that will not be surprising 😅 

> Docker daemon runs an embedded DNS server which provides DNS resolution among containers connected to the same user-defined network, so that these containers can resolve container names to IP addresses. If the embedded DNS server is unable to resolve the request, it will be forwarded to any external DNS servers configured for the container. To facilitate this when the container is created, only the embedded DNS server reachable at 127.0.0.11 will be listed in the container’s resolv.conf file. 

I agree with tao here, I think that we should left until last the DNS exercise

pkg/cluster/internal/delete/delete.go
neolit123: the error can be scoped:
```
if err := nodes.Delete(n...); err != nil {
...
}
```

pkg/cluster/internal/kubeadm/config.go
neolit123: +1 on merging the resolv.confs if feasble.

pkg/cluster/nodes/create.go
```diff
@@ -74,6 +75,11 @@ func CreateControlPlaneNode(name, image, clusterLabel, listenAddress string, por
 		cache.ports = map[int32]int32{kubeadm.APIServerPort: port}
 	})
 
+	err = addResolve(node)
```
neolit123: `if err := addResolve(node); err != nil {`

pkg/cluster/nodes/create.go
```diff
@@ -114,6 +120,12 @@ func CreateWorkerNode(name, image, clusterLabel string, mounts []cri.Mount) (nod
 	if err != nil {
 		return node, err
 	}
+
+	err = addResolve(node)
```
neolit123: `if err := addResolve(node); err != nil {`

pkg/container/docker/network.go
```diff
@@ -0,0 +1,42 @@
...
+	"sigs.k8s.io/kind/pkg/exec"
+)
+
+// CreateNetwork create a bridge network
+func CreateNetwork(networkName string) error {
+	cmd := exec.Command(
+		"docker", "network",
+		"create",
+		"--driver=bridge",
+		networkName,
```
aojea: we mentioned at same place [using labels for kind networks ](https://docs.docker.com/engine/reference/commandline/network_create/)

`docker network create ... --label k`

tao12345666333: sure, thanks. :+1:

pkg/cluster/nodes/create.go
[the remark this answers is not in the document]
tao12345666333: Thanks. I understand what you mean. I can continue to implement the remaining features of this PR.

There is a question. According to your proposal https://github.com/kubernetes-sigs/kind/issues/273#issuecomment-488874160, should I configure the name of the network in Networking (in the configuration file)? https://github.com/kubernetes-sigs/kind/blob/master/pkg/cluster/config/v1alpha3/types.go#L90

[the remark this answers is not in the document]
tao12345666333: now pass the clusterName as an argument.

pkg/container/docker/network.go
tao12345666333: Addressed.

pkg/cluster/nodes/create.go
tao12345666333: Thanks. Addressed.

tao12345666333:
Now this PR creates a network for each cluster.

Whether we should write the network to the configuration file as Ben’s proposal says? https://github.com/kubernetes-sigs/kind/issues/273#issuecomment-488874160

> Strawman:
> 
> * if the specified network exists, we use it without creating
> * if it doesn't exist, we create it, and label it as belonging to the cluster
> * on cluster delete, we list networks labeled with the cluster, and delete only those (so not just whatever the containers use, only if we labeled it)
> * we name this field / functionality somehow such that it is clear that this feature is docker specific, leaving room for podman etc. in the immediate future xref #154.

The configuration file looks like this: 

```YAML
kind: Cluster
nodes:
- role: control-plane
- role: control-plane
- role: control-plane
- role: worker
networking:
  name: test-network
```

BenTheElder:
We should rename the "name" field to be more explicit. 🤔

tao12345666333:
> We should rename the "name" field to be more explicit.

Modified to `bridgeName` what do you think?

pkg/cluster/nodes/create.go
```diff
@@ -142,6 +151,8 @@ func createNode(name, image, clusterLabel, role string, mounts []cri.Mount, extr
 		"--label", clusterLabel,
 		// label the node with the role ID
 		"--label", fmt.Sprintf("%s=%s", constants.NodeRoleKey, role),
+		// connect node to network
+		"--network", clusterName,
```
BenTheElder: what are valid values for this? does clusterName and network name really have the same restrictions? EG for containers we always append something so it will pass validation differently

pkg/cluster/nodes/create.go
```diff
@@ -109,19 +113,24 @@ func CreateExternalLoadBalancerNode(name, image, clusterLabel, listenAddress str
 }
 
 // CreateWorkerNode creates a worker node
-func CreateWorkerNode(name, image, clusterLabel string, mounts []cri.Mount) (node *Node, err error) {
-	node, err = createNode(name, image, clusterLabel, constants.WorkerNodeRoleValue, mounts)
+func CreateWorkerNode(name, image, clusterName, clusterLabel string, mounts []cri.Mount) (node *Node, err error) {
```
BenTheElder: this API is definitely going to need rework soon, these functions take too many parameters.

this new parameter is really the network name.

pkg/cluster/nodes/util.go
```diff
@@ -44,3 +46,23 @@ func GetControlPlaneEndpoint(allNodes []Node) (string, error) {
 
 	return fmt.Sprintf("%s:%d", loadBalancerIP, loadbalancer.ControlPlanePort), nil
 }
+
+// to avoid the DNS crash we need copy host's /etc/resolv.conf to node
+// ref: https://github.com/kubernetes-sigs/kind/pull/484#issuecomment-489469044
+func addResolve(node *Node) error {
+	resolv, err := ioutil.ReadFile("/etc/resolv.conf")
```
BenTheElder: windows has no /etc/resolv.conf.

BenTheElder: this is a blocker.
/hold

pkg/cluster/nodes/create.go
tao12345666333: it only need `string` https://github.com/moby/moby/blob/master/api/types/types.go#L475

pkg/cluster/internal/create/nodes.go
```diff
@@ -88,6 +94,18 @@ func provisionNodes(
...
+	if err := docker.CreateNetwork(clusterName); err != nil {
```
BenTheElder: we still need to handle if the network already exists.

tao12345666333: Yes, I am doing this and will submit it immediately.

pkg/cluster/nodes/create.go
tao12345666333: thanks.I modified them yesterday and have not submitted them yet. :laughing:

BenTheElder:
we should not be copying the hosts's resolv.conf ourselves, that's a simplification of what docker does and may not match. instead we should manage the node's resolv.conf appropriately.

I'm also not sure how many situations there are where actually using the host's DNS is desirable. I bet minikube does not do this, and we've run into a lot of problems with this EG when running in another cluster.

pkg/cluster/nodes/util.go
tao12345666333: > windows has no /etc/resolv.conf.

Then I will judge through the OS first, skip the behavior of windows. I can't accurately judge the behavior of windows.

pkg/cluster/internal/create/nodes.go
tao12345666333: > we still need to handle if the network already exists.

I am using `docker.IsNetworkExist` to check if the network is exist.  https://github.com/kubernetes-sigs/kind/pull/484/files#diff-62943c16bb4ee62933f24631e2d3d507R84

pkg/cluster/nodes/util.go
tao12345666333: just skip Windows. :speak_no_evil:

tao12345666333:
On the processing of `resolv.conf`, I skipped the Windows system directly. 
Because I don't have a Windows machine that can be used to test this feature. :worried:

aojea:
> On the processing of `resolv.conf`, I skipped the Windows system directly.
> Because I don't have a Windows machine that can be used to test this feature. 😟 And I can't accurately judge the behavior of windows.

I have one, but I'm using WSL (Windows Subsystem for Linux) to run kind on it and it does have resolv.conf

neolit123:
info on the the windows host file:
https://gist.github.com/zenorocha/18b10a14b2deb214dc4ce43a2d2e2992

some notes:
- syntax: IP is first, name second
- path is here: c:\Windows\System32\Drivers\etc\hosts, requires Administrator privileges.
- i don't recommended using https://docs.microsoft.com/en-us/windows/desktop/api/windns/nf-windns-dnsmodifyrecordsinset_a or powershell, but instead parse the file and add entries manually.

meta question:
- why did we end up modifying the DNS resolver on the host - can we avoid that?

aojea:
> meta question:
> why did we end up modifying the DNS resolver on the host - can we avoid that?

is about copying the resolv.conf from the hosts and filtering the loopback addresses to avoid the CoreDNS crash loop, I think that this [was the solution ](https://github.com/kubernetes-sigs/kind/pull/484#issuecomment-489469044 )to unblock the work.

[Using user defined bridges docker creates a resolv.conf that points to the docker internal DNS with address 127.0.0.11 ](https://github.com/kubernetes-sigs/kind/pull/484#issuecomment-489431439)

I guess that now is time to decide how do we want to move forward :)

pkg/cluster/nodes/util.go
```diff
@@ -44,3 +47,29 @@ func GetControlPlaneEndpoint(allNodes []Node) (string, error) {
...
+	if err != nil {
+		return errors.Wrap(err, "failed to read /etc/resolv.conf")
+	}
+
+	// filter the loopback addresses
+	re := regexp.MustCompile("(?m)[\r\n]+^.*((127.([0-9]{1,3}.){2}[0-9]{1,3})|(::1)).*$")
+	content := re.ReplaceAllString(string(resolv), "")
+
+	err = node.WriteFile("/kind/resolv.conf", content)
```
yeya24: This error can be scoped. 
``` Go
if err := node.WriteFile(); err != nil {
}
```

tao12345666333: Addressed. Thanks.

pkg/cluster/internal/delete/delete.go
```diff
@@ -46,5 +47,15 @@ func Cluster(c *context.Context) error {
...
+	networkName, err := docker.Inspect(n[0].Name(), "{{range $name, $_:=.NetworkSettings.Networks}}{{$name}}{{end}}")
+	if err != nil {
+		return errors.Wrap(err, "error inspect nodes")
+	}
+
+	if err := nodes.Delete(n...); err != nil {
+		return errors.Wrap(err, "error deleting nodes")
+	}
+	// after all nodes have been deleted, we need delete the network
+	return docker.DeleteNetwork(networkName[0])
```
DazWorrall: I believe an objective here was to _only_ delete networks which kind creates. Will this honour that?

tao12345666333: yep. it should filter by labels.  I will update it ASAP. thanks.

tao12345666333:
It seems that the docker related commands cannot be tested in the CI environment, so I deleted the test case for the docker network.

tao12345666333:
I think we should do something to push this PR forward.

/assign @BenTheElder

pkg/cluster/nodes/util.go
BenTheElder: this is the first time kind does not behave the same in all environments, not really a fan of that.

also, I don't think we should be reading the host's resolv.conf at all

pkg/cluster/nodes/util.go
```diff
@@ -44,3 +47,28 @@ func GetControlPlaneEndpoint(allNodes []Node) (string, error) {
...
+		return nil
+	}
+
+	resolv, err := ioutil.ReadFile("/etc/resolv.conf")
+	if err != nil {
+		return errors.Wrap(err, "failed to read /etc/resolv.conf")
+	}
+
+	// filter the loopback addresses
+	re := regexp.MustCompile("(?m)[\r\n]+^.*((127.([0-9]{1,3}.){2}[0-9]{1,3})|(::1)).*$")
```
BenTheElder: this shouldn't even be necessary in the hosts's resolv.conf

pkg/container/docker/network.go
```diff
@@ -0,0 +1,83 @@
...
+		"--label="+fmt.Sprintf("%s=%s", constants.ClusterLabelKey, networkName),
+		networkName,
+	)
+	return cmd.Run()
+}
+
+// DeleteNetwork delete the special network
+// only when the network was created by kind
+func DeleteNetwork(networkName string) error {
+	if isNetworkCreatedByKind(networkName) {
```
BenTheElder: this is orthogonal to the functionality of delete network. this package should be kind agnostic. we don't do this for delete containers

pkg/container/docker/network.go
```diff
@@ -0,0 +1,83 @@
...
+	"sigs.k8s.io/kind/pkg/exec"
+)
+
+// CreateNetwork create a bridge network
+func CreateNetwork(networkName string) error {
+	cmd := exec.Command(
+		"docker", "network",
+		"create",
+		"--driver=bridge",
+		"--label="+fmt.Sprintf("%s=%s", constants.ClusterLabelKey, networkName),
```
BenTheElder: this package should not be aware of the cluster label key. this package just implements docker functionality

pkg/container/docker/network.go
```diff
@@ -0,0 +1,83 @@
...
+	)
+	if err := cmd.Run(); err != nil {
+		return false
+	}
+
+	return true
+}
+
+// isNetworkCreatedByKind checks if it was created by kind
+func isNetworkCreatedByKind(networkName string) bool {
```
BenTheElder: this should live somewhere in cluster/

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -181,6 +182,7 @@ evictionHard:
...
```
BenTheElder: this is not being done for all api versions

pkg/cluster/internal/create/nodes.go
```diff
@@ -193,16 +214,16 @@ func nodesToCreate(cfg *config.Cluster, clusterName string) []nodeSpec {
 }
 
 // TODO(bentheelder): remove network in favor of []cri.PortMapping when that is in
-func (d *nodeSpec) Create(clusterLabel string) (node *nodes.Node, err error) {
+func (d *nodeSpec) Create(networkName, clusterLabel string) (node *nodes.Node, err error) {
```
BenTheElder: why isn't this in the nodeSpec?

pkg/cluster/config/types.go
```diff
@@ -102,4 +102,7 @@ type Networking struct {
 	// If DisableDefaultCNI is true, kind will not install the default CNI setup.
 	// Instead the user should install their own CNI after creating the cluster.
 	DisableDefaultCNI bool
+	// BridgeName is the bridge network name.
+	// Defaults to same as cluster name.
+	BridgeName string
```
BenTheElder: https://github.com/kubernetes-sigs/kind/issues/273#issuecomment-488874160

BenTheElder: (this config field name is not suitable, we need to separate docker specific functionality and make that clear)

BenTheElder:
> It seems that the docker related commands cannot be tested in the CI environment, so I deleted the test case for the docker network.

they absolutely can, but the _unit tests_ should be unit tests, not invoking docker commands (!)

this is why we need #528 

> I think we should do something to push this PR forward.


@tao12345666333 unfortunately I need to prioritize some other fixes etc related to testing Kubernetes (EG the HA fixes), and we have many open PRs.. This is pretty far into P3 https://kind.sigs.k8s.io/docs/contributing/project-scope/

I'm still not convinced that the DNS approach is a good one to solve this problem and may introduce more problems, and the field name needs thought / is not in line with the proposal. Left some more specific review feedback.

BenTheElder:
let's revisit a more complete design in the 0.7ish timeframe with the provider API in mind and more attention to dns config handling across platforms etc.

aojea:
> let's revisit a more complete design in the 0.7ish timeframe with the provider API in mind and more attention to dns config handling across platforms etc.

do we want to use docker embedded dns for the nodes and use the autodiscovery or are we fine without that functionality? 
The later will simplify a lot the implementations because will only requires to deal with the resolv.conf merging.
The downside is that it won't allow using the dns names to configure the cluster, but with all the ipv6 and dual-stack features coming in, I really would love not to depend on the resolver in those environments 😅

tao12345666333:
In my opinion, if we don't use docker  embedded dns and only use the new bridge network, then we may be able to reduce a lot of trouble.😄

Issue #485, opened 2019-05-05
determine and document resource requirements


BenTheElder:
**What would you like to be documented**: A more accurate lower bound on resources when using kind with docker desktop. Currently we suggest 4GB / 4 CPU which while probably accurate for building kubernetes should be more than we need to run a node. 

https://kind.sigs.k8s.io/docs/user/quick-start/#creating-a-cluster

**Why is this needed**: We don't want to overstate requirements dramatically and scare off potential users :-)

We'll need to do some testing to determine the threshold. It should be reduced at HEAD, it might also be interesting to check if that is true 🙃

nickolaev:
I am trying `kind` in the [CircleCI machine executor](https://circleci.com/docs/2.0/executor-types/#using-machine) which seems to be perfectly fine with its 2 CPU cores. Hope this helps.

BenTheElder:
thanks!

can confirm that at HEAD even the smallest settings docker desktop offers currently work fine for `kind create cluster`:
<img width="612" alt="Screen Shot 2019-05-05 at 3 24 25 PM" src="https://user-images.githubusercontent.com/917931/57201162-f417b180-6f49-11e9-8363-8bbd39719954.png">

```console
$ kind create cluster
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.14.1) 🖼
 ✓ Preparing nodes 📦 
 ✓ Creating kubeadm config 📜 
 ✓ Starting control-plane 🕹️ 
 ✓ Installing CNI 🔌 
 ✓ Installing StorageClass 💾 
Cluster creation complete. You can now use the cluster with:

export KUBECONFIG="$(kind get kubeconfig-path --name="kind")"
kubectl cluster-info

$ kubectl get no
NAME                 STATUS   ROLES    AGE   VERSION
kind-control-plane   Ready    master   20s   v1.14.1

$ kubectl get po --all-namespaces
NAMESPACE     NAME                      READY   STATUS    RESTARTS   AGE
kube-system   coredns-fb8b8dccf-qhm47   1/1     Running   0          23s
kube-system   coredns-fb8b8dccf-rc56v   1/1     Running   0          23s
kube-system   kube-proxy-ksw7d          1/1     Running   0          23s
kube-system   weave-net-dqz4p           2/2     Running   0          23s
```

BenTheElder:
[...]

PR #490, opened 2019-05-05
remove generated base image sources

### LINKED ISSUE #488: remove base-image sources from kind binary
**What should be cleaned up or changed**: we should remove the code generation / baked in sources for `kind build base-image`.

**Why is this needed**: These are unnecessary bloat in retrospect, if you wish to build the base image you're going to want to be able to tweak the sources, we did not wind up exposing any other configuration options for this build related to the contents.

We can remove this, and instead by default autodetect the kind sources location the same as we do for kubernetes, we already have an explicit option to select the sources.

/assign

cc @aojea @neolit123 @tao12345666333 @fabriziopandini

BenTheElder:
fixes #488 
EDIT: also #485, we can't find a good lower bound for docker for desktop because it appears to be lower than the settings offer, but we can stop suggesting that it is 4 core / 4 cpu unless you are building kubernetes.


pkg/build/base/base.go
```diff
@@ -79,22 +80,20 @@ func (c *BuildContext) Build() (err error) {
...
-	// if SourceDir is unset, use the baked in sources
+	// if SourceDir is unset then try to autodetect source dir
 	buildDir := tmpDir
 	if c.sourceDir == "" {
-		// populate with image sources
-		err = sources.RestoreAssets(buildDir, "images/base")
+		pkg, err := build.Default.Import("sigs.k8s.io/kind", build.Default.GOPATH, build.FindOnly)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to locate sources: %v", err)
```
tao12345666333: maybe using errors instead of fmt.

BenTheElder: done, also updated the docs

BenTheElder:
updated the docs, includes minor updates related to https://github.com/kubernetes-sigs/kind/issues/485 as well as the changes here

Issue #493, opened 2019-05-06
Errors while building and running kind


mkumatag:
**What happened**:
Hitting with the following error:

**ERROR1:** Error while building base-image:

```
INFO[09:03:55] Building in kind-build-d0e9e8e8-853e-4095-9c0d-d82d67689c2a
DEBU[09:03:55] Running: /usr/bin/docker [docker exec kind-build-d0e9e8e8-853e-4095-9c0d-d82d67689c2a mkdir /kind/]
DEBU[09:03:55] Running: /usr/bin/docker [docker exec kind-build-d0e9e8e8-853e-4095-9c0d-d82d67689c2a rsync -r /build/bits/ /kind/]
DEBU[09:03:55] Running: /usr/bin/docker [docker exec kind-build-d0e9e8e8-853e-4095-9c0d-d82d67689c2a /bin/sh -c curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add -]
E: gnupg, gnupg2 and gnupg1 do not seem to be installed, but one of them is required for this operation
ERRO[09:03:56] Adding Kubernetes apt key failed! exit status 255
ERRO[09:03:56] Image build Failed! Failed to install Kubernetes: exit status 255
DEBU[09:03:56] Running: /usr/bin/docker [docker rm -f -v kind-build-d0e9e8e8-853e-4095-9c0d-d82d67689c2a]
Error: error building node image: exit status 255
root@ip9-114-192-229:~/kind_ws/src/sigs.k8s.io/kind#
```

**ERROR2:** Error while running create cluster:

```
./kind create cluster --loglevel debug --image node:latest


docker run -d -t --privileged --security-opt seccomp=unconfined --tmpfs /tmp --tmpfs /run -v /lib/modules:/lib/modules:ro --hostname kind-control-plane --name kind-control-plane --label io.k8s.sigs.kind.cluster=kind --label io.k8s.sigs.kind.role=control-plane --expose 35230 -p 127.0.0.1:35230:6443 node:latest

# docker logs 096465b64473
Initializing machine ID from random generator.
mount: /sys/class/dmi/id/product_name: mount point does not exist.
root@ip9-114-192-229:~/kind_ws/src/sigs.k8s.io/kind#
```

aojea:
@mkumatag the base image changed recently #461 , please fetch the latest sources and try to build again the base image.

mkumatag:
[...]

PR #494, opened 2019-05-06
Fix errors

mkumatag:
Fixes: https://github.com/kubernetes-sigs/kind/issues/493

Code reused from - https://github.com/kubernetes-sigs/kubeadm-dind-cluster/blob/master/image/rundocker#L80:L83

### DIFF
--- a/images/base/entrypoint
+++ b/images/base/entrypoint
@@ -61,8 +61,10 @@ fix_machine_id() {
 fix_product_name() {
   # this is a small fix to hide the underlying hardware and fix issue #426
   # https://github.com/kubernetes-sigs/kind/issues/426
-  echo "kind" > /kind/product_name
-  mount -o ro,bind /kind/product_name /sys/class/dmi/id/product_name
+  if [[ $(cat /sys/class/dmi/id/product_name) =~ Google ]]; then
+    echo "kind" > /kind/product_name
+    mount -o ro,bind /kind/product_name /sys/class/dmi/id/product_name
+  fi
 }
 
 # run pre-init fixups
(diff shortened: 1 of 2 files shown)


images/base/entrypoint
```diff
@@ -61,8 +61,10 @@ fix_machine_id() {
 fix_product_name() {
   # this is a small fix to hide the underlying hardware and fix issue #426
   # https://github.com/kubernetes-sigs/kind/issues/426
-  echo "kind" > /kind/product_name
-  mount -o ro,bind /kind/product_name /sys/class/dmi/id/product_name
+  if [[ $(cat /sys/class/dmi/id/product_name) =~ Google ]]; then
```
BenTheElder: We should just check if the file exists, I'd prefer not to branch at all in the entry point for a consistent environment.

BenTheElder: It might even make sense to ensure this directory and fake the whole thing in all environments, but that's a bigger & trickier PR.

Ideally we want kind to act very much the same everywhere.

BenTheElder:
```suggestion
  if [[ ! -f /sys/class/dmi/id/product_name ]]; then
```

mkumatag:
wow... one click on suggestion button helped me merge @BenTheElder changes, nice GitHub feature. ptal latest changes.

images/base/entrypoint
```diff
@@ -61,8 +61,10 @@ fix_machine_id() {
 fix_product_name() {
   # this is a small fix to hide the underlying hardware and fix issue #426
   # https://github.com/kubernetes-sigs/kind/issues/426
-  echo "kind" > /kind/product_name
-  mount -o ro,bind /kind/product_name /sys/class/dmi/id/product_name
+  if [[ -f /sys/class/dmi/id/product_name ]]; then
+    echo "kind" > /kind/product_name
+    mount -o ro,bind /kind/product_name /sys/class/dmi/id/product_name
```
BenTheElder: Note to self: consider always faking the entire DMI decode VFS

PR #496, opened 2019-05-06
bump base to v20190506-d0ac573

BenTheElder:
picks up #494


aojea:
how the base images are built? do the merges kick some pipeline or something?

BenTheElder:
I run `hack/build/push-base.sh`. Pipelines will come later, probably GCB (more secure than prow, we have resources for it). I looked a bit this weekend, but _normally_ this image hasn't been changing as much. The node-image is the one I really want in a pipeline eventually.

PR #497, opened 2019-05-06
collect containerd logs

BenTheElder:
whoops!


BenTheElder:
@aojea this may help debug 😉

PR #498, opened 2019-05-07
[WIP]: kindnet

BenTheElder:
this is pretty rough ATM, throwing it up for visibility. cc @aojea 

the first commit is purely lifting kindnet from https://github.com/aojea/kindnet and cannot merge like that without fixing the accreditation.  

the remaining commits involve rewriting it to use a single container that writes out the templated cni config, fixing up deps, etc.


BenTheElder:
cc also @neolit123 

this works locally now :-)

it's not really ready to go in yet, but I wanted to show where it's at.

cmd/kindnetd/main.go
```diff
@@ -0,0 +1,198 @@
...
+			// we don't need to add routes, but we might need to update
+			// the cni config
+			if nodeIP == hostIP {
+				fmt.Printf("handling current node\n")
+				// compute the current cni config inputs
+				cniData := computeCNIConfigInputs(node)
+				// if it changed, write out the new config
+				if cniData != lastCNIData {
+					fmt.Printf("Writing CNI data\n")
+					if err := writeCNIConfig(cniData); err != nil {
```
BenTheElder: this is the biggest change, we write this on the fly instead of using a busybox container + configmap

cmd/kindnetd/main.go
```diff
@@ -0,0 +1,198 @@
...
+
+func computeCNIConfigInputs(node corev1.Node) cniConfigInputs {
+	podCIDR := node.Spec.PodCIDR
+	return cniConfigInputs{
+		PodCIDR: podCIDR,
+	}
+}
+
+func writeCNIConfig(data cniConfigInputs) error {
+	return fs.AtomicWriteFile("/etc/cni/net.d/10-kindnet.conflist", func(w io.Writer) error {
```
BenTheElder: this doesn't actually do what I wanted in retrospect: this will still create a file in the cni directory, instead we want to mimic what configmaps do: 
- write somewhere else and then create / update a symlink

this is to avoid spurious CNI reload, containerd (and I think cri-o) do an fsnotify watch on the cni config

go.mod
```diff
@@ -9,34 +11,46 @@ require (
...
+	golang.org/x/sys v0.0.0-20190312061237-fead79001313 // indirect
+	golang.org/x/text v0.3.1-0.20181227161524-e6919f6577db // indirect
+	golang.org/x/time v0.0.0-20190308202827-9d24e82272b4 // indirect
 	gopkg.in/airbrake/gobrake.v2 v2.0.9 // indirect
 	gopkg.in/gemnasium/logrus-airbrake-hook.v2 v2.1.2 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
-	k8s.io/api v0.0.0-20181026145037-6e4b5aa967ee // indirect
-	k8s.io/apimachinery v0.0.0-20181126191516-4a9a8137c0a1
-	k8s.io/client-go v7.0.0+incompatible
+	k8s.io/api v0.0.0-20190222213804-5cb15d344471
```
BenTheElder: I had to patch up the API versions to get things to build.

cmd/kindnetd/main.go
aojea: Should we bump the file number? s/10/90/ per example, this will allow other CNIs to take control without any additional changes

images/kindnetd/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+FROM golang:1.12 AS builder
+ENV GO111MODULE=on
+
+WORKDIR /go/src/kindnet
+COPY . .
+RUN go get -d -v ./...
+RUN GOOS=linux CGO_ENABLED=0 go build -o /go/bin/kindnet ./cmd/kindnetd
+
+# STEP 2: Build small image
+FROM scratch
```
aojea: is not the trend to move to distroless?

BenTheElder:
left this branch for reference but significantly redid this

PR #500, opened 2019-05-07
kindnet

BenTheElder:
Introduce a new minimal CNI setup derived from https://github.com/aojea/kindnet.

/cc @aojea


cmd/kindnetd/main.go
```diff
@@ -0,0 +1,238 @@
...
+		// first get this node's IP
+		nodeIP := internalIP(node)
+		fmt.Printf("Handling node with IP: %s\n", nodeIP)
+
+		// This is our node. We don't need to add routes, but we might need to
+		// update the cni config.
+		if nodeIP == hostIP {
+			fmt.Printf("handling current node\n")
+			// compute the current cni config inputs
+			if err := cniConfigWriter.Write(
```
aojea: are we continue writing the file or only if something has changed_

BenTheElder: only if something changed, the writer handles this 😅

BenTheElder:
heh `misspell` is failing on go.sum 🤷‍♂

BenTheElder:
```console
I0507 08:55:30.980] ✓ Waiting ≤ 1m0s for control-plane = Ready ⏳
I0507 08:55:30.980] • Ready after 4s 💚
```
The CLI wasn't designed for node readiness to be this quick :+)

BenTheElder:
Feels like we triggered a new race condition with tests using the e2e PSP... 🤔
/retest

aojea:
seems that the tests that are failing 9/1009 all have the same error

> test/e2e/framework/framework.go:141
> Failed to create PSP e2e-test-privileged-psp
> Expected error:
>     <*errors.StatusError | 0xc421d98750>: {
>         ErrStatus: {
>             TypeMeta: {Kind: "", APIVersion: ""},
>             ListMeta: {SelfLink: "", ResourceVersion: "", Continue: ""},
>             Status: "Failure",
>             Message: "podsecuritypolicies.extensions \"e2e-test-privileged-psp\" already exists",
>             Reason: "AlreadyExists",
>             Details: {
>                 Name: "e2e-test-privileged-psp",
>                 Group: "extensions",
>                 Kind: "podsecuritypolicies",
>                 UID: "",
>                 Causes: nil,
>                 RetryAfterSeconds: 0,
>             },
>             Code: 409,
>         },
>     }
>     podsecuritypolicies.extensions "e2e-test-privileged-psp" already exists
> not to have occurred
>

aojea:
now DNS tests are failing, is too fast now that we are hitting races or is too buggy that makes the test fail?

pkg/cluster/internal/create/actions/config/config.go
```diff
@@ -79,6 +79,7 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 			ControlPlaneEndpoint: controlPlaneEndpoint,
 			APIBindPort:          kubeadm.APIServerPort,
 			Token:                kubeadm.Token,
+			PodSubnet:            "10.244.0.0/16", // TODO: make configurable
```
aojea: right, otherwise others CNI like Calico has to be configured to match that subnet in order to work

cmd/kindnetd/main.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Kubernetes Authors.
+Copyright 2016 The Kubernetes Authors.
```
neolit123: this seems to shift the year back to 2016.

BenTheElder: oops, thanks! this commit started as copying the changes from the last PR 😉

BenTheElder:
so the < 1.13 clusters are now triggering the lack of this patch in the tests https://github.com/kubernetes/kubernetes/commit/0bef5e4325e657ca9370c36138cc485dbe4b2614

we can get it cherry picked to 1.12, but 1.11 is closed for business (probably time for us to stop testing it ...)

not sure on the 1.13+ failures yet.

BenTheElder:
filed https://github.com/kubernetes/kubernetes/pull/77555 to cherry pick to 1.12 for that particular issue.

BenTheElder:
Cool, 1.11 and 1.12 are the test bug, verify is the spell check bug, and the rest work now :-)

neolit123:
> error: go.sum:103:56: "YUO" is a misspelling of "YOU"

looks like we need to stop checking some files for spelling.

pkg/build/node/const.go
```diff
@@ -138,7 +138,10 @@ data:
       "name": "kindOfBridge",
       "plugins": [
         {
-          "type": "ptp",
+          "type": "bridge",
+          "ipMasq": true,
```
aojea: we shouldnt set ipmasq to true
>pods on a node can communicate with all pods on all nodes without NAT

https://kubernetes.io/docs/concepts/cluster-administration/networking/

neolit123:
sent fix here: https://github.com/kubernetes-sigs/kind/pull/501

BenTheElder:
[looking to see that 1.13+ pass again, if so will continue with the current approach and clean up the PR]

pkg/build/node/const.go
BenTheElder: pretty sure this is a normal thing to do https://kubernetes.io/docs/tasks/administer-cluster/ip-masq-agent/, will dig more in a bit.

BenTheElder:
[dropping ipMasq definitely made this fail]

BenTheElder:
/retest
1.12 should be patched

BenTheElder:
1.12 works now :-)

aojea:
Just to confirm, ipmasq false affects the DNS tests only? I want to go deeper on that

BenTheElder:
> Just to confirm, ipmasq false affects the DNS tests only? I want to go deeper on that

no, I also observed failures with the aggregated API server in newer kubernetes versions

aojea:
🤔 The jobs that are failing has problems resolving dns names upstream because they are trying to reach the upstream dns with the POD ip
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_kind/500/pull-kind-conformance-parallel-1-14/1125887548770488324/artifacts/logs/kind-control-plane/pods/kube-system_coredns-fb8b8dccf-h7dkq_875e79b1-7117-11e9-8d5c-0242ac110002/coredns/0.log

> 2019-05-07T22:44:39.304561321Z stdout F 2019-05-07T22:44:39.304Z [ERROR] plugin/errors: 2 10-244-1-157.dns-5014.pod.cluster.local.c.k8s-prow-builds.internal. A: dial tcp 10.63.240.10:53: i/o timeout
2019-05-07T22:44:41.464268946Z stdout F 2019-05-07T22:44:41.463Z [ERROR] plugin/errors: 2 dns-test-service.dns-5014.svc.cluster.local.c.k8s-prow-builds.internal. A: read udp 10.244.0.3:42103->10.63.240.1

The ` ipMasq: true` seems to install an iptables chain per pod like this 

```
Chain POSTROUTING (policy ACCEPT)
target     prot opt source               destination
KUBE-POSTROUTING  all  --  anywhere             anywhere             /* kubernetes postrouting rules */
CNI-78729f4f4ad56a33e1ba62cd  all  --  10.244.0.0/24        anywhere             /* name: "kindOfBridge" id: "b3a6e75c6e295effcdd1cd25aa94a01b4b0ecd7a6149ca
3787e9bc840e75f6ef" */
CNI-0b6173815888f3f16b15c305  all  --  10.244.0.0/24        anywhere             /* name: "kindOfBridge" id: "60d06833d2e3d93425efbd1129f404fb9e0c3c88a1296d
b95cdaa58a9d36f673" */
```
That means that pod to pod traffic between nodes is being MASQUERADE, check that is using the podCIDR /24 subnet instead of the podSubnet that's /16 in our case. Also, the number of "duplicate" iptables rules will be equal to the # of pods, maybe is not an issue for us but is not optimal 😅 

If the problem is the pod2external traffic that needs to be masquerade, we can done it withing kindnetd, is just install an iptables rule in all nodes masquerading the pod2 external traffic, I have doubts where the NAT rules should be applied honestly, if at the external interace only or anywhere
`iptables -t nat -A POSTROUTING -o eth0 ! -d 10.244.0.0/16 -j MASQUERADE` 
in this case POD_SUBNET = 10.244.0.0/16

BenTheElder:
> k8s-prow-builds
that's the CI cluster, which we run kind inside of a pod on ... that's suspect.
we should set DNS policy on the kind job pods.

BenTheElder:
will look at this more sometime tomorrow. I suspect we would be fine with those rules, but we should look into it deeper :-)

pkg/build/node/const.go
```diff
@@ -24,225 +24,202 @@ const (
...
+    {
+      "cniVersion": "0.3.1",
+      "name": "kindOfBridge",
+      "plugins": [
+        {
+          "type": "bridge",
+          "ipMasq": true,
+          "isDefaultGateway": true,
+          "hairpinMode": true,
+          "ipam": {
```
aojea: I guess we should add the dataDir option 

> Allocated IP addresses are stored as files in /var/lib/cni/networks/$NETWORK_NAME. The path can be customized with the dataDir option listed above. Environments where IPs are released automatically on reboot (e.g. running containers are not restored) may wish to specify /var/run/cni or another tmpfs mounted directory instead.
https://github.com/containernetworking/plugins/blob/master/plugins/ipam/host-local/README.md

BenTheElder:
we can test that, also we should consider https://kubernetes.io/docs/tasks/administer-cluster/ip-masq-agent/

aojea:
> we can test that, also we should consider https://kubernetes.io/docs/tasks/administer-cluster/ip-masq-agent/

ups, you sent me the link before but I didn't get it 😅  definitively prefer the ip-masq-agent/

BenTheElder:
ok so tests are fixed, focusing on #505 for a tiny bit longer, then coming back to this.

BenTheElder:
rebased -> actually just made it it's own isolated module, we'll probably need to build this infrequently and the dependencies are larger than `kind` proper

aojea:
@BenTheElder do you mind to try this config? 
```
diff --git a/pkg/build/node/const.go b/pkg/build/node/const.go
index 607fed3..58957e3 100644
--- a/pkg/build/node/const.go
+++ b/pkg/build/node/const.go
@@ -139,7 +139,7 @@ data:
       "plugins": [
         {
           "type": "bridge",
-          "ipMasq": true,
+          "ipMasq": false,
           "isDefaultGateway": true,
           "hairpinMode": true,
           "ipam": {
@@ -222,4 +222,57 @@ spec:
           hostPath:
             path: /etc/cni/net.d
 ---
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: ip-masq-agent
+  namespace: kube-system
+  labels:
+    tier: node
+    app: kindnet
+data:
+  config: |-
+    nonMasqueradeCIDRs:
+      - 10.244.0.0/16
+    masqLinkLocal: false
+    resyncInterval: 60s
+
+---
+apiVersion: extensions/v1beta1
+kind: DaemonSet
+metadata:
+  name: ip-masq-agent
+  namespace: kube-system
+spec:
+  template:
+    metadata:
+      labels:
+        tier: node
+        app: kindnet
+        k8s-app: ip-masq-agent
+    spec:
+      hostNetwork: true
+      tolerations:
+      - operator: Exists
+        effect: NoSchedule
+      serviceAccountName: kindnet
+      containers:
+      - name: ip-masq-agent
+        image: gcr.io/google-containers/ip-masq-agent-amd64:v2.0.0
+        securityContext:
+          privileged: true
+        volumeMounts:
+          - name: config
+            mountPath: /etc/config
+      volumes:
+        - name: config
+          configMap:
+            # Note this ConfigMap must be created in the same namespace as the daemon pods - this spec uses kube-system
+            name: ip-masq-agent
+            optional: true
+            items:
+              # The daemon looks for its config in a YAML file at /etc/config/ip-masq-agent
+              - key: config
+                path: ip-masq-agent
+---

```

it has hardcoded the container platform and the cluster CIDR (podSubnet) but we can figure it out later

pkg/cluster/internal/create/actions/config/config.go
aojea: Please, take a look at this approach to see if it makes sense https://github.com/kubernetes-sigs/kind/pull/348/commits/ef8d84bb33a314426e1c2705fad935191ac4b907
It allows the user to define the podSubnet but defaults to a const value

BenTheElder:
@aojea i'm working on a version of this, first we need to fix the ipmasq agent not having a mult-arch image (thankfully they already build for multiple arches, just not as a manifest list yet)

BenTheElder:
NOTE: this is temporarily using my ipmasq image instead of upstream to test it out while we wait for https://github.com/kubernetes-incubator/ip-masq-agent/pull/36

pkg/build/node/const.go
```diff
@@ -217,10 +218,65 @@ spec:
...
+  name: ip-masq-agent
+  namespace: kube-system
+  labels:
+    tier: node
+    app: kindnet
+data:
+  config: |-
+    nonMasqueradeCIDRs:
+      - 10.244.0.0/16
+    masqLinkLocal: false
```
aojea: This value is available to kind and in the kubadm config map, I guess that we should be able to pass it in

pkg/build/node/const.go
```diff
@@ -27,9 +27,10 @@ const (
 The default CNI manifest and images are our own tiny kindnet
 */
 
-var defaultCNIImages = []string{"kindest/kindnetd:0.1.0"}
+var defaultCNIImages = []string{"kindest/kindnetd:0.1.0", "gcr.io/bentheelder-kind-dev/ip-masq-agent:v2.3.0-1-g51408b3"}
```
aojea: https://github.com/kubernetes-incubator/ip-masq-agent/blob/e54f921c3191d303a708d6a9dbf88fc8d14d6c26/cmd/ip-masq-agent/ip-masq-agent.go#L224 , it doesn´t accept IPv6 prefixes

🤔 do you think that the project will accept patches to make it IPv6 compatible?

BenTheElder: if they're low risk I'd assume so, otherwise we can do some fixes ourselves.

pkg/build/node/const.go
BenTheElder: yep, have some ideas for that

BenTheElder:
we will need new images with https://github.com/kubernetes-incubator/ip-masq-agent/pull/38

pkg/build/node/const.go
BenTheElder: spoke to @MrHohn who has been very gracious accepting PRs so far and getting images pushed, we should be good to send fixes for that. I'm sending one right now for that particular issue.

BenTheElder: https://github.com/kubernetes-incubator/ip-masq-agent/pull/39

BenTheElder: updated to use the upstream image

pkg/build/node/const.go
```diff
@@ -31,6 +31,7 @@ var defaultCNIImages = []string{"kindest/kindnetd:0.1.0", "k8s.gcr.io/ip-masq-ag
 
 const defaultCNIManifest = `
 # kindnetd networking manifest
+# would you kindly template this file
```
BenTheElder: cc @amwat

pkg/build/node/const.go
```diff
@@ -153,7 +154,7 @@ data:
             "ranges": [
               [
                 {
-                  "subnet": "{{ .PodCIDR }}"
+                  "subnet": "{{"{{ .PodCIDR }}"}}"
```
BenTheElder: this is kludgey but necessary, this template executes later on and is inside the manifest template :^)

pkg/build/node/const.go
BenTheElder: this is now templated.

pkg/build/node/const.go
```diff
@@ -217,10 +218,65 @@ spec:
         securityContext:
           privileged: false
           capabilities:
-             add: ["NET_ADMIN"]
+            add: ["NET_RAW", "NET_ADMIN"]
```
BenTheElder: we don't need this one on kindnetd

pkg/cluster/internal/create/actions/config/config.go
BenTheElder: going to punt making this configurable to a follow-up PR just because this is so big, but it's relatively trivial (mostly boilerplate) and I'll be able to file it immediately. this PR leaves all the hooks to inject a configured version

pkg/build/node/const.go
BenTheElder: fixed

pkg/build/node/const.go
BenTheElder: we use the ip masq agent currently

pkg/cluster/config/defaults/image.go
```diff
@@ -18,4 +18,4 @@ limitations under the License.
 package defaults
 
 // Image is the default for the Config.Image field, aka the default node image.
-const Image = "kindest/node:v1.14.1"
+const Image = "kindest/node:alpha-v1.14.1@sha256:1fb6e8a3842eed2ba0cb202ae1a70d208ed372c76b4600aa717c77d589f8334a"
```
BenTheElder: need to think more about naming, this is to signal that it requires an unreleased kind version to run, since we introduced the default manifest templating. might need to rethink that mechanism too.

/hold
 
also after some more though, we should be pinning these by sha always.

BenTheElder: Kubernetes 1.14.2 is cut tomorrow. We can put kind 0.3 on this and document the image breakage in the release notes as well as document and encourage pinning image shas.

BenTheElder:
I _think_ this is good to go now. Filed https://github.com/kubernetes-sigs/kind/pull/524 as a follow-up to make the pod subnet configurable.

pkg/build/node/const.go
BenTheElder: done

pkg/cluster/internal/create/actions/config/config.go
BenTheElder: https://github.com/kubernetes-sigs/kind/pull/524 is that follow-up

aojea:
/lgtm
impressive,
any idea on the improvement on image size and time to create the cluster?

pkg/cluster/internal/create/actions/installcni/cni.go
```diff
@@ -47,30 +51,45 @@ func (a *action) Execute(ctx *actions.ActionContext) error {
...
-			"-f", "/kind/manifests/default-cni.yaml",
-		).Run(); err != nil {
-			return errors.Wrap(err, "failed to apply overlay network")
+	manifest := raw.String()
+
+	// TODO: remove this check?
+	// backwards compatibility for mounting your own manifest file to the default
+	// location
+	// NOTE: this is intentionally undocumented, as an internal implementation
+	// detail. Going forward users should disable the defalut CNI and install
```
neolit123: typo in `defalut`

BenTheElder:
> so the TLDR is that since we can't find a good CNI plugin that suites our needs, we wrote our own?
why didn't calico work?

Well we're using the standard bridge and host-local CNI plugins, and the standard ip-masq-agent.

The only thing we're doing ourselves is:
- configuring them
- writing routes between the nodes

The routes part is something like 200 lines or code and could be done with less if we were terse, the rest is the dockerfile etc. and templating the CNI manifest and setting a podSubnet in kubeadm. which we'd also need to do with calico.

The idea is that this is very cheap and simple. kindnetd is only 30mb or so and at runtime it's just reading the nodes and ensuring static routes so it's cheap and fast.

Incidentally this is basically what kops and GCE / GKE are doing, but using the VPC routing and in some places kubenet for the node part (which is deprecated). Compare https://github.com/containerd/cri/blob/master/cluster/gce/cni.template

BenTheElder:
This is also one route to IPV6 compatibility and cross platform support. We will have an easier time  with that sort of thing because how simple it is.

Cons are:
- a new thing to maintain
- doesn't have networkpolicy
- perhaps less similar to bespoke setups? EG my lazy home cluster currently runs weave.
Neutral:
- IPV6 (Some other CNI setups have this)
- Cross platform (some other CNI setups have this, though we can trivially port)
Pros:
- it's very light on resources
- we have full control to patch etc. as we desire
- most of the code is in the plugins and agent which are light and relatively simple
- very similar to cloud setups like kubenet + routes

BenTheElder:
Will look at Kubernetes network tests and do some more calico tests before moving forward. So far this seems like a win, but I am wary of those cons.

BenTheElder:
Side Note: Forgot to mention that Ironically if we need NetworkPolicy for testing we can deploy calico in just-policy mode alongside our own routes etc. (GKE optionally does this in production)

https://docs.projectcalico.org/v3.7/getting-started/kubernetes/installation/other

BenTheElder:
pretty drastic performance differences actually put into metrics written down somewhere this time :^)

Kubernetes Version | Job Time ([Calico](https://github.com/kubernetes-sigs/kind/pull/527)) | Job Time ([kindnetd](https://github.com/kubernetes-sigs/kind/pull/500)) 
---------------------------|-------------------------|-----------------------------
1.12 | [25m17s](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_kind/527/pull-kind-conformance-parallel-1-12/1128395914617032707/) | [19m59s](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_kind/500/pull-kind-conformance-parallel-1-12/1128114688089919490/)
1.13 | [26m25s](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_kind/527/pull-kind-conformance-parallel-1-13/1128395914617032705/) | [19m23s](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_kind/500/pull-kind-conformance-parallel-1-13/1128114688089919489/)
1.14 | [27m39s](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_kind/527/pull-kind-conformance-parallel-1-14/1128395914612838402/) | [21m45s](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_kind/500/pull-kind-conformance-parallel-1-14/1128114688089919488/)
master | [38m40s](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_kind/527/pull-kind-conformance-parallel/1128395914612838400/) | [19m59s](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_kind/500/pull-kind-conformance-parallel/1128114688085725186/)

image size: 1.53 GB (calico) vs 1.33GB (kindnetd + ipmasq-agent) ~= 15% bigger on disk

to be clear: I don't mean to imply that CNI deployments like Calico wouldn't be excellent choices for serious production clusters, kind is a wildly different environment, and for that environment kindnetd seems to be a good route forward. Weave has also served us reasonably well so far, but is more expensive than kindnetd and no signs of ipv6.

BenTheElder:
/test all

there's a far bit of variance, but this PR seems to be consistently faster.

BenTheElder:
/hold cancel
let's :shipit: 

I'm willing to move to calico or similar in the future as need be if we can show the trade offs are worth it. So far I think this is a good solution for now. Some more prior art https://github.com/leblancd/kube-v6

PR #502, opened 2019-05-07
add makefile to build containerized with modules

BenTheElder:
a lot of people may not have new enough go yet, also we can make this more reproducible than running release builds with the go installed on my computer :wink: 

after this I'm hoping perhaps we can remove vendor/ and go all in on modules

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,56 @@
+# Copyright 2019 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -0,0 +1,56 @@
...
+# cleans the cache volume
+clean-cache:
+	docker volume rm $(CACHE_VOLUME)
+
+# builds kind in a container, outputs to $(REPO_ROOT)/_output/bin
+kind: make-cache
+	docker run \
+		--rm \
+		-v $(CACHE_VOLUME):/go \
+		-e GOCACHE=/go/cache \
```
BenTheElder: this makes the build ~ the same amount of time as just `go build` on my machine after the first build, we get the go module and package build cache which make builds very quick.

just using a volume is dramatically simpler than the rsync etc in the main kubernetes repo, this remains a very simple `go build`, we've just wrapped it with a standard container now and persisted the cache to a volume :-)

BenTheElder:
replaces: https://github.com/kubernetes-sigs/kind/pull/183

PR #504, opened 2019-05-08
build improvements

BenTheElder:
- autodetect GOOS and GOARCH when doing the dockerized build
- use the module proxy to speed up initial build (important for CI)
- force module mode

### DIFF
--- a/hack/goarch.sh
+++ b/hack/goarch.sh
@@ -1,0 +1,44 @@
+#!/bin/sh
+# Copyright 2019 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


BenTheElder:
rebased

hack/goarch.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/sh
```
BenTheElder: posix SH for maximum portability, the idea is to just call docker ASAP.

we'll need something else for windows.

hack/goarch.sh
```diff
@@ -0,0 +1,44 @@
...
+
+# fall back to uname
+if which uname >/dev/null 2>&1; then
+  case "$(uname -m)" in
+    x86_64)
+      echo "amd64"
+      exit 0
+    ;;
+    arm*)
+      if [ "$(getconf LONG_BIT)" == "64" ]; then
```
BenTheElder: this will probably need to be more elaborate, but arm users can also just `go build` and for the release we'll cross build anyhow instead of detecting the host.

hack/goarch.sh
```diff
@@ -0,0 +1,44 @@
...
+
+# if we have go, just ask go!
+if which go >/dev/null 2>&1; then
+    go env GOARCH
+    exit $?
+fi
+
+# fall back to uname
+if which uname >/dev/null 2>&1; then
+  case "$(uname -m)" in
```
BenTheElder: TODO also: power PC (need to find what the value is)

PR #505, opened 2019-05-08
use modules, remove vendor

BenTheElder:
- improve build tooling so users can build even if they have old go on the host (just need docker / a sprinkling of make / posix sh)
- update the hack/ scripts to use modules, simplify away vendor/ concerns
  - get golint / go vet to work without `go list ./...` (faster)
- use the makefile for pushing images
- remove vendor/

/hold


BenTheElder:
The motivation for this is getting ahead of the curve as modules become standardized, and avoiding checking in a ton more vendor when we pull in client-go for `kindnetd` (#500).

We will need to update the docs a bit.

BenTheElder:
(verify-deps is failing because it's purpose was validating vendor, and we'e removed it and that script, if this goes in I'll remove that job)

BenTheElder:
related: kinder is already on modules only.

tao12345666333:
<pre>
GOBIN=/tmp/tmp.97BKNno1Oj/bin make -C /go/src/sigs/k8s.io/kind install
W0508 05:06:44.403] make: *** /go/src/sigs/k8s.io/kind: No such file or directory.  Stop.
W0508 05:06:44.404] + cleanup
W0508 05:06:44.404] + kind export logs /workspace/_artifacts/logs
W0508 05:06:44.405] ./../../sigs.k8s.io/kind/hack/ci/e2e.sh: line 29: kind: command not found
</pre>

the path incorrect.

BenTheElder:
I just pushed a correction. this e2e script is non-ideal, waiting for a
teammate to get kubetest2 into shape :+)

*From: *Jintao Zhang <notifications@github.com>
*Date: *Tue, May 7, 2019 at 10:11 PM
*To: *kubernetes-sigs/kind
*Cc: *Benjamin Elder, Mention

GOBIN=/tmp/tmp.97BKNno1Oj/bin make -C /go/src/sigs/k8s.io/kind install

hack/update-generated.sh
```diff
@@ -38,6 +38,16 @@ go build -o "${BINDIR}/conversion-gen" k8s.io/code-generator/cmd/conversion-gen
 go mod vendor
 export GO111MODULE="off"
 
+# fake being in a gopath
```
BenTheElder: this is a little ugly _but_ it makes it possible to checkout kind to any directory and run this script.

hack/ci/e2e.sh
```diff
@@ -46,13 +46,17 @@ install_kind() {
...
+    # module mode ...
     if [[ $(go list sigs.k8s.io/kind) = "sigs.k8s.io/kind" ]]; then
-        env "GOBIN=${TMP_DIR}/bin" go install sigs.k8s.io/kind
+        DIR="${PWD}"
+        cd ./../../sigs.k8s.io/kind
+        make install INSTALL_PATH="${TMP_DIR}/bin"
+        cd "${DIR}"
     else
-        env "GOPATH=${TMP_DIR}" go get -u sigs.k8s.io/kind
+        env "GOPATH=${TMP_DIR}" GO111MODULE="on" go get -u sigs.k8s.io/kind
```
BenTheElder: this is probably not right. TODO

hack/ci/e2e.sh
```diff
@@ -46,13 +46,17 @@ install_kind() {
...
+    # kind requires go modules now
     # if we have a kind checkout, install that to the tmpdir, otherwise go get it
+    # TODO: this script should probably go away anyhow, but this isn't exactly
+    # module mode ...
     if [[ $(go list sigs.k8s.io/kind) = "sigs.k8s.io/kind" ]]; then
-        env "GOBIN=${TMP_DIR}/bin" go install sigs.k8s.io/kind
+        DIR="${PWD}"
+        cd ./../../sigs.k8s.io/kind
+        make install INSTALL_PATH="${TMP_DIR}/bin"
+        cd "${DIR}"
```
BenTheElder: I'm sure this can be done more cleanly, but it appears to work fine, and it's CI specific anyhow.

pkg/cluster/config/default.go
```diff
@@ -14,6 +14,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// this comment makes golint ignore this file, feel free to edit the file.
+// Code generated by not-actually-generated-but-go-away-golint. DO NOT EDIT.
+// https://github.com/kubernetes/code-generator/issues/30
```
BenTheElder: I like this hack :^)

hack/verify-generated.sh
```diff
@@ -60,8 +59,7 @@ main() {
   diff=$(diff -Nupr \
           -x ".git" \
           -x "_output" \
-          -x "vendor/github.com/jteeuwen/go-bindata/testdata" \
-          -x "vendor/github.com/golang/dep/internal/fs/testdata/symlinks" \
+          -x "vendor" \
```
BenTheElder: this one still needs to ignore vendor because hack/update-generated.sh locally populates vendor (because the code-generators currently need to be run in GOPATH)

hack/util/goinstalldir.sh
```diff
@@ -0,0 +1,47 @@
...
+
+# check if GOBIN is set anyhow
+if [ -n "${GOBIN}" ]; then
+  echo "GOBIN"
+  exit 0
+fi
+
+# check if GOPATH is set anyhow
+if [ -n "${GOPATH}" ]; then
+  echo "${GOPATH}/go/bin"
```
BenTheElder: this should be ${GOPATH}/bin
/hold

Makefile
```diff
@@ -20,6 +20,7 @@ REPO_ROOT=$(PWD)
 # autodetect host GOOS and GOARCH by default, even if go is not installed
 GOOS=$(shell hack/util/goos.sh)
 GOARCH=$(shell hack/util/goarch.sh)
+INSTALL_DIR=$(shell hack/util/goinstalldir.sh)
```
BenTheElder: basically this should usually give the current $GOPATH/bin

hack/util/goinstalldir.sh
BenTheElder: I had to catch some Zzzzzs 😁

hack/ci/e2e.sh
BenTheElder: Well yes, and the working for for this script is Kubernetes, to clarify.

hack/update-generated.sh
```diff
@@ -19,32 +19,49 @@ set -o errexit
...
+"${BINDIR}/deepcopy-gen" -i ./pkg/cluster/config/v1alpha2 -O zz_generated.deepcopy --go-header-file hack/boilerplate.go.txt
+"${BINDIR}/defaulter-gen" -i ./pkg/cluster/config/v1alpha2 -O zz_generated.default --go-header-file hack/boilerplate.go.txt
+"${BINDIR}/conversion-gen" -i ./pkg/cluster/config/v1alpha2 -O zz_generated.conversion --go-header-file hack/boilerplate.go.txt
 
-deepcopy-gen -i ./pkg/cluster/config/v1alpha2 -O zz_generated.deepcopy --go-header-file hack/boilerplate.go.txt
-defaulter-gen -i ./pkg/cluster/config/v1alpha2 -O zz_generated.default --go-header-file hack/boilerplate.go.txt
-conversion-gen -i ./pkg/cluster/config/v1alpha2 -O zz_generated.conversion --go-header-file hack/boilerplate.go.txt
+"${BINDIR}/deepcopy-gen" -i ./pkg/cluster/config/v1alpha3 -O zz_generated.deepcopy --go-header-file hack/boilerplate.go.txt
+"${BINDIR}/defaulter-gen" -i ./pkg/cluster/config/v1alpha3 -O zz_generated.default --go-header-file hack/boilerplate.go.txt
+"${BINDIR}/conversion-gen" -i ./pkg/cluster/config/v1alpha3 -O zz_generated.conversion --go-header-file hack/boilerplate.go.txt
```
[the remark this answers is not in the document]
BenTheElder: They work, I've rerun them with no diff. But see the hacks in this script to keep them working and not slow. If you use them with the module env set they become very slow. If you turn that off, then you need proper gopath and sources.

hack/util/goinstalldir.sh
BenTheElder: fixed

hack/update-generated.sh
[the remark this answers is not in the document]
BenTheElder: also with the fake gopath nonsense they work if you clone kind to a random directory.

hack/ci/e2e.sh
BenTheElder: updated, I fixed the makefile's path stuff and then had this use `"$(dirname "${BASH_SOURCE[0]}")"` to locate itself and compute the path.

BenTheElder: should be much cleaner now

BenTheElder:
I think we just need to remove the verify deps job and update the development / install docs. Will do docs in a follow up as this PR is huge :+)

BenTheElder:
presubmits are fixed.

hack/update-generated.sh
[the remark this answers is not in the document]
BenTheElder: [github seems confused on the timing of these comments]

BenTheElder:
/hold cancel
here we go 🔥 

I expect this may break some setups that are installing kind from HEAD, should be fairly easy to patch up though, and I think this is the way forward for go. 😅now that the proxy server is available the time seems ripe

Issue #112, opened 2018-11-14
kind create cluster creates same kubeconfig user name for each cluster


font:
When creating multiple clusters with `kind create cluster --name <name>` the kubeconfig for each cluster specifies the same user name of `kubernetes-admin`. This becomes a problem when trying to use multiple kubeconfigs in `KUBECONFIG` because the user name overlaps with each of the configs. For example:

```bash
$ export KUBECONFIG="$(kind get kubeconfig-path):$(kind get kubeconfig-path --name 2)"
$ kubectl --context=kubernetes-admin@kind-2 get all --all-namespaces
error: the server doesn't have a resource type "all"
```

Instead each auth user name should include the `--name` of the kind cluster when specified in the `kind create cluster --name <name>` command.

BenTheElder:
this seems to be https://github.com/kubernetes/kubeadm/issues/416, looking into options...

BenTheElder:
I [asked in #sig-cluster-lifecycle](https://kubernetes.slack.com/archives/C13J86Z63/p1542777463422000),  so far I don't think there's an answer for this. Following up...

BenTheElder:
@neolit123 @fabriziopandini WDYT?
Afaict we'd have to provision our own user and matching kubeconfig.

marckhouzam:
I've hit this problem previously while using multiple k8s clusters.  Turns out one can use the same username in all your clusters.  What needs to be tweaked in the kubeconfig file is changing the `name` field of the user to something unique and add the `username` field under `users.user` to be `kubernetes-admin`.

I'll try my hand at writing a PR for it.

marckhouzam:
Ah, I see that it is kubeadm that generates the kubeconfig file and may not allow what kind needs for this.  I'll have a look at kubeadm some more.  But modifying the kubeconfig file after it is generated would be an option.

marckhouzam:
Ok, I've got things working.  Here is what the new kubeconfig file looks like for a cluster named `mycluster` in case people want to manually make the changes for themselves:

```
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: REDACTED
    server: https://localhost:58039
  name: mycluster
contexts:
- context:
    cluster: mycluster
    user: kubernetes-admin-mycluster
  name: kubernetes-admin@mycluster
current-context: kubernetes-admin@mycluster
kind: Config
preferences: {}
users:
- name: kubernetes-admin-mycluster
  user:
[...]

PR #512, opened 2019-05-09
Feat/kubeconfig multi cluster

marckhouzam:
Modify kubeconfig for multiple clusters ( #112 )

kubeadm does not currently allow to configure a user reference id,
and instead always uses kubernetes-admin.  This causes a problem
when we create multiple clusters with Kind and want to use
each corresponding kubeconfig file at the same time in KUBECONFIG.

Until kubeadm supports configuring a user reference id, the only
option to fix this for Kind is to modify the kubeconfig file that
kubeadm provides.  As Kind already did this with the server name,
it made sense to take the logic further and also make the user
reference id unique to a cluster.

Three approaches were considered:

1- continue with the current approach of parsing each line of
   the admin.conf kubeconfig file, and make the modifications
   necessary.  After implementing this approach, the solution
   seemed quite brittle as it uses regex but no yaml structure.

2- use the go package yaml.v2 to -fully- parse the yaml of the
   admin.conf kubeconfig file, make the modifications, and then
   output the new yaml kubeconfig file.  This solution requires
   to define a detailed struct of every field contained in the
   original yaml file. Having to map every field in advance is
   brittle as any modification that kubeadm may make to the file
   in the future would require adaptation in Kind.

3- use the go package yaml.v2 to -generically- parse the yaml of
   the admin.conf kubeconfig file, make the modifications, and then
   output the new yaml kubeconfig file.  This solution only
   accesses the yaml fields that are required to be modified.
   Although any future changes from kubeadm to those fields would
   require modifications in Kind, modifications to all other fields
   would not.

This PR implements solution #3 which was felt to be the most
future-proof and least brittle of the three.

### DIFF
--- a/pkg/cluster/internal/create/actions/kubeadminit/init.go
+++ b/pkg/cluster/internal/create/actions/kubeadminit/init.go
@@ -18,12 +18,10 @@ limitations under the License.
 package kubeadminit
 
 import (
-	"bytes"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
-	"regexp"
 	"strings"
 
 	log "github.com/sirupsen/logrus"
@@ -35,6 +33,8 @@ import (
 	"sigs.k8s.io/kind/pkg/cluster/internal/loadbalancer"
 	"sigs.k8s.io/kind/pkg/cluster/nodes"
 	"sigs.k8s.io/kind/pkg/exec"
+
+	yaml "gopkg.in/yaml.v2"
 )
 
 // kubeadmInitAction implements action for executing the kubadm init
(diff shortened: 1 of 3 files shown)


pkg/cluster/internal/create/actions/kubeadminit/init.go
```diff
@@ -114,33 +115,81 @@ func (a *action) Execute(ctx *actions.ActionContext) error {
...
+	}
+
+	// Use the new user reference id in the users section
+	// which is in the "users[0].name" field.
+	// In the same loop, add the 'username' field in the "users[0].user" section.
+	user := config["users"].([]interface{})[0].(map[interface{}]interface{})
+	for k := range user {
+		if k == "name" {
+			user[k] = newUserName
+		} else if k == "user" {
```
tao12345666333: Is this for backward compatibility? The current config looks like this:

```
(MoeLove) ➜  kind git:(using-network) ✗ kubectl config view 
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: DATA+OMITTED
    server: https://localhost:45965
  name: moe
contexts:
- context:
    cluster: moe
    user: kubernetes-admin
  name: kubernetes-admin@moe
current-context: kubernetes-admin@moe
kind: Config
preferences: {}
users:
- name: kubernetes-admin
  user:
    client-certificate-data: REDACTED
    client-key-data: REDACTED

```

BenTheElder:
Thanks for the PR!

/ok-to-test

I wonder if the KUBECONFIG format is fully stable by now, this change seems a bit brittle still.

marckhouzam:
@BenTheElder Your comment made me realize I may be able to make these kubeconfig modifications by calling `kubectl config`. That should be less brittle.

I'll try it out and report back.

pkg/cluster/internal/create/actions/kubeadminit/init.go
marckhouzam: Which part?

As I understand it, the field `contexts[0].context.user` refers to a `user` defined in this file in the `users[]` section.  So, this PR makes the `user` unique and changes it in the two sections: `contexts[0].context.user` and `users[].name`.

Finally, the username within the cluster is taken by default to be `users[].name` but since we changed it, we need to tell what the real username is.  That is why we must insert the field `username`.

There is not backwards-compatibility issue to worry about, as far as I can tell, although I am no expert in kubeconfig definitions.
However, the real username in the cluster is still `kubernetes-admin`, which used to be the

marckhouzam:
@neolit123 Thanks for the information.  I will look into the different options and re-work this PR.

neolit123:
np, i guess part of my point that this should be already doable without changes in kind.

marckhouzam:
I've did a new PR #520 to use client-go to modify the server in the kubeconfig file, as suggested by @neolit123 

I've also written the modification needed to make the user unique using client-go, but I haven't posted it.  I'll try @neolit123 suggestion about using kubeadm directly for this and see if I can get that to work nicely.

Ultimately, I think it would be nice if kubeadm would automatically generate the userid with the cluster name suffix, as this problem is not limited to Kind but to any time multiple clusters are created using kubeadm.

BenTheElder:
This is done now as part of #850. Thanks again for the PR 🙃

PR #516, opened 2019-05-09
Added instruction for installing Kind tool manually.

sohankunkerkar:
@BenTheElder As discussed, created this PR to add instruction for installing Kind tool manually in case other method fails.
/assign @BenTheElder

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,9 +29,10 @@ You can install the latest bleeding edge kind code with `GO111MODULE="on" go get
 
 **NOTE**: please use the latest go to do this, ideally go 1.12.5 or greater.
 
-This will put `kind` in `$(go env GOPATH)/bin`. You may need to add that directory to your `$PATH` as
-shown [here](https://golang.org/doc/code.html#GOPATH) if you encounter the error
-`kind: command not found` after installation.
+This will put `kind` in `$(go env GOPATH)/bin`. If you encounter the error
+`kind: command not found` after installation then you may need to either add that directory to your `$PATH` as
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -29,9 +29,10 @@ You can install the latest bleeding edge kind code with `GO111MODULE="on" go get
 
 **NOTE**: please use the latest go to do this, ideally go 1.12.5 or greater.
 
-This will put `kind` in `$(go env GOPATH)/bin`. You may need to add that directory to your `$PATH` as
-shown [here](https://golang.org/doc/code.html#GOPATH) if you encounter the error
-`kind: command not found` after installation.
+This will put `kind` in `$(go env GOPATH)/bin`. If you encounter the error
+`kind: command not found` after installation then you may need to either add that directory to your `$PATH` as
+shown [here](https://golang.org/doc/code.html#GOPATH) or do manual installation by cloning the repo and run 
+`go install .` from the repository.
```
BenTheElder: `go install .` will put it in the same path, if the command above fails we should probably suggest cloning and then `make install` (which also currently uses that path though)

BenTheElder: (recommending `make install` because docker is the only dependency, so if the host go is not the right version it will still work)

sohankunkerkar: yeah, that makes sense. Let me change the instruction.

sohankunkerkar:
> thanks for the PR!
> 
> a suggestion, and can we also update site/content/ ? there's some markdown in there with the same instructions as the README 😅

Sure thing!

PR #518, opened 2019-05-10
WIP test Kindnetd only

aojea:
this is a test only, valid one is https://github.com/kubernetes-sigs/kind/pull/500


pkg/build/node/const.go
```diff
@@ -222,4 +223,57 @@ spec:
...
+metadata:
+  name: ip-masq-agent
+  namespace: kube-system
+  labels:
+    tier: node
+    app: kindnet
+data:
+  config: |-
+    nonMasqueradeCIDRs:
+      - 10.244.0.0/16
```
aojea: probably we should template this
```
kubectl -n kube-system get configmap kubeadm-config -o yaml | grep podSubnet
      podSubnet: 10.244.0.0/16
```

pkg/cluster/internal/create/actions/config/config.go
```diff
@@ -79,6 +79,7 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 			ControlPlaneEndpoint: controlPlaneEndpoint,
 			APIBindPort:          kubeadm.APIServerPort,
 			Token:                kubeadm.Token,
+			PodSubnet:            "10.244.0.0/16", // TODO: make configurable
```
aojea: Please, take a look at this approach to see if it makes sense https://github.com/kubernetes-sigs/kind/pull/348/commits/ef8d84bb33a314426e1c2705fad935191ac4b907

PR #520, opened 2019-05-12
Modify kubeconfig by using client-go, not regex

marckhouzam:
This commit uses k8s.io/client-go to modify the kubeconfig
file to use a localhost server.

Signed-off-by: Marc Khouzam <marc.khouzam@ville.montreal.qc.ca>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/google/gofuzz v0.0.0-20170612174753-24818f796faf
 	github.com/google/uuid v1.1.1
 	github.com/googleapis/gnostic v0.2.0 // indirect
+	github.com/imdario/mergo v0.3.7 // indirect
 	github.com/inconshreveable/mousetrap v1.0.0 // indirect
 	github.com/json-iterator/go v1.1.5 // indirect
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
@@ -23,12 +24,14 @@ require (
 	github.com/spf13/cobra v0.0.3
 	github.com/spf13/pflag v1.0.2 // indirect
 	golang.org/x/crypto v0.0.0-20180910181607-0e37d006457b
+	golang.org/x/oauth2 v0.0.0-20190402181905-9f3314589c9a // indirect
+	golang.org/x/time v0.0.0-20190308202827-9d24e82272b4 // indirect
 	gopkg.in/airbrake/gobrake.v2 v2.0.9 // indirect
 	gopkg.in/gemnasium/logrus-airbrake-hook.v2 v2.1.2 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	k8s.io/api v0.0.0-20181026145037-6e4b5aa967ee // indirect
 	k8s.io/apimachinery v0.0.0-20181126191516-4a9a8137c0a1
-	k8s.io/client-go v7.0.0+incompatible
+	k8s.io/client-go v10.0.0+incompatible
 	k8s.io/klog v0.1.0 // indirect
 	k8s.io/kube-openapi v0.0.0-20181025202442-3a9b63ab1e39 // indirect
 	sigs.k8s.io/kustomize v2.0.1+incompatible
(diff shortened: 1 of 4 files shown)


marckhouzam:
@BenTheElder  I made the assumption that Kind was targeting a single kubernetes version, but now I'm thinking that is not correct.

If Kind supports multiple versions of Kubernetes, then I'm not sure it is possible to use client-go, as I believe we need to choose which version of client-go to use based on the version of Kubernetes.  What do you think?

aojea:
reading the [Versioning chapter of the README](https://github.com/kubernetes/client-go/blob/master/README.md ) they are following semver so  I think that we should use latest stable tag? in this case is v1.11.0
What do you think?

marckhouzam:
@aojea you're right that v1.11.0 would be better.  I'm just not sure if all versions of kubernetes have the same data structure to read the kubeconfig file. So if Kind prepared a cluster for, say, k8s v1.10, will client-go v1.11 still work?

I'm out of my depth here...

BenTheElder:
kind indeed supports multiple versions of kubernetes. out of the box there is a default version, but kind's initial support was (and is) actually for developing kubernetes, where we may be building fairly arbitrary (v1.11+) versions of kubernetes and then running them with kind.

neolit123:
> I'm just not sure if all versions of kubernetes have the same data structure to read the kubeconfig file

the client-go Config type is unlikely to change anytime soon. if it does, v1 will be supported for a long period of time. thus using any recent version of `client-go/tools/clientcmd` should be fine.

go.mod
```diff
@@ -23,12 +24,14 @@ require (
...
 	golang.org/x/crypto v0.0.0-20180910181607-0e37d006457b
+	golang.org/x/oauth2 v0.0.0-20190402181905-9f3314589c9a // indirect
+	golang.org/x/time v0.0.0-20190308202827-9d24e82272b4 // indirect
 	gopkg.in/airbrake/gobrake.v2 v2.0.9 // indirect
 	gopkg.in/gemnasium/logrus-airbrake-hook.v2 v2.1.2 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	k8s.io/api v0.0.0-20181026145037-6e4b5aa967ee // indirect
 	k8s.io/apimachinery v0.0.0-20181126191516-4a9a8137c0a1
-	k8s.io/client-go v7.0.0+incompatible
+	k8s.io/client-go v10.0.0+incompatible
```
munnerz: I'd expect that `k8s.io/apimachinery`, `k8s.io/api` and potentially even `k8s.io/klog` are bumped in the `go.sum` file after this change? It seems like `go mod tidy` (or `hack/update-vendor.sh`) needs running 😄

pkg/cluster/internal/create/actions/kubeadminit/init.go
```diff
@@ -35,6 +30,8 @@ import (
 	"sigs.k8s.io/kind/pkg/cluster/internal/loadbalancer"
 	"sigs.k8s.io/kind/pkg/cluster/nodes"
 	"sigs.k8s.io/kind/pkg/exec"
+
+	"k8s.io/client-go/tools/clientcmd"
```
munnerz: nit: external imports should exist in the 2nd block (after stdlib imports but before imports within this module)

pkg/cluster/internal/create/actions/kubeadminit/init.go
```diff
@@ -114,43 +112,26 @@ func (a *action) Execute(ctx *actions.ActionContext) error {
...
 // this should only be called on a control plane node
-// While copyng to the host machine the control plane address
+// While copying to the host machine the control plane address
 // is replaced with local host and the control plane port with
 // a randomly generated port reserved during node creation.
-func writeKubeConfig(n *nodes.Node, dest string, hostPort int32) error {
+func writeKubeConfig(n *nodes.Node, dest string, hostPort int32, clusterName string) error {
 	cmd := n.Command("cat", "/etc/kubernetes/admin.conf")
-	lines, err := exec.CombinedOutputLines(cmd)
+	buff, err := exec.Output(cmd)
```
munnerz: (I'm not too familiar here, so please excuse my question if it's a stupid one!) why the switch from CombinedOutputLines to Output?

BenTheElder:
thanks again for the PR, we've removed client-go as a direct or indirect dependency for the kind binary which makes it easier to import in the future and the binary much smaller.

I'm going to look at doing something similar without client-go but without regex as part of #850

marckhouzam:
Sorry for losing track of this.
I'll be checking out #850.

PR #524, opened 2019-05-14
Configurable podSubnet

BenTheElder:
Adds to #500, only commits from `add podSubnet to config` and forward are new.

Adds `Cluster.network.podSubnet` config field, and respects it.

### DIFF
--- a/pkg/cluster/config/types.go
+++ b/pkg/cluster/config/types.go
@@ -96,4 +96,7 @@ type Networking struct {
 	//
 	// Defaults to 127.0.0.1
 	APIServerAddress string
+	// PodSubnet is the CIDR used for pod IPs
+	// kind will select a default if unspecified
+	PodSubnet string
 }
--- a/pkg/cluster/config/validate_test.go
+++ b/pkg/cluster/config/validate_test.go
@@ -22,6 +22,84 @@ import (
 	"sigs.k8s.io/kind/pkg/util"
 )
 
+func TestClusterValidate(t *testing.T) {
+	cases := []struct {
+		Name         string
+		Cluster      Cluster
+		ExpectErrors int
+	}{
+		{
+			Name: "Defaulted",
+			Cluster: func() Cluster {
+				c := Cluster{}
+				SetDefaults_Cluster(&c)
+				return c
+			}(),
+		},
+		{
+			Name: "bogus podSubnet",
+			Cluster: func() Cluster {
+				c := Cluster{}
+				SetDefaults_Cluster(&c)
+				c.Networking.PodSubnet = "aa"
+				return c
+			}(),
+			ExpectErrors: 1,
+		},
+		{
+			Name: "missing control-plane",
+			Cluster: func() Cluster {
+				c := Cluster{}
+				SetDefaults_Cluster(&c)
+				c.Nodes = []Node{}
+				return c
+			}(),
+			ExpectErrors: 1,
+		},
+		{
+			Name: "bogus node",
+			Cluster: func() Cluster {
+				c := Cluster{}
+				SetDefaults_Cluster(&c)
+				n, n2 := Node{}, Node{}
+				SetDefaults_Node(&n)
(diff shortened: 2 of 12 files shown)


pkg/cluster/config/types.go
```diff
@@ -96,4 +96,7 @@ type Networking struct {
 	//
 	// Defaults to 127.0.0.1
 	APIServerAddress string
+	// PodSubnet is the CIDR used for pod IPs
+	// kind will select a default if unspecified
+	PodSubnet string
```
BenTheElder: TODO: we should validate this by checking `net.ParseCIDR`

aojea: I think that kubedm does it
https://github.com/kubernetes/kubernetes/blob/59cedcf5415c1419d03a443d2a27600e1d3ab8fc/cmd/kubeadm/app/apis/kubeadm/validation/validation.go#L390

BenTheElder: yeah they're doing a bit more. we're validating that it's at least a subnet for now :-)

pkg/cluster/config/validate_test.go
```diff
@@ -22,6 +22,84 @@ import (
 	"sigs.k8s.io/kind/pkg/util"
 )
 
+func TestClusterValidate(t *testing.T) {
```
aojea: I'm curious, what job run these tests?

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -106,6 +108,8 @@ controllerManagerExtraArgs:
   enable-hostpath-provisioner: "true"
 nodeRegistration:
   criSocket: "/run/containerd/containerd.sock"
+networking:
+  podSubnet: {{ .PodSubnet }}
```
aojea: I had issues when using IPv6 addresses, we should quote it
`podSubnet: "{{ .PodSubnet }}"`

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -140,6 +144,8 @@ apiServerExtraVolumes:
 apiServerCertSANs: [localhost]
 controllerManagerExtraArgs:
   enable-hostpath-provisioner: "true"
+networking:
+  podSubnet: {{ .PodSubnet }}
```
aojea: I had issues when using IPv6 addresses, we should quote it
`podSubnet: "{{ .PodSubnet }}"`

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -202,6 +208,8 @@ apiServer:
 controllerManager:
   extraArgs:
     enable-hostpath-provisioner: "true"
+networking:
+  podSubnet: {{ .PodSubnet }}
```
aojea: I had issues when using IPv6 addresses, we should quote it
`podSubnet: "{{ .PodSubnet }}"`

pkg/cluster/config/validate_test.go
BenTheElder: nothing! :this-is-fine: :^)

I think build used to? need to fix that

pkg/cluster/config/types.go
BenTheElder: TODO is done

pkg/cluster/internal/kubeadm/config.go
BenTheElder: done

pkg/cluster/internal/kubeadm/config.go
BenTheElder: done

pkg/cluster/internal/kubeadm/config.go
BenTheElder: done

pkg/cluster/config/validate_test.go
BenTheElder: added a commit to make the build job do that for now

PR #526, opened 2019-05-14
Write down specific command to use for using a configuration file when creating a cluster

fiunchinho:


aojea:
thanks
/ok-to-test
/assign @BenTheElder

PR #527, opened 2019-05-14
[WIP] What about calico

BenTheElder:
cc @aojea @neolit123 
/hold


BenTheElder:
this is the existing kindnetd work which includes the templating etc, but then swapping in calico 3.7 (latest) modified to use our templating.

it will be mildly more annoying to do ipv6 but we can make it work. letting this soak in the CI tests and going to test the image sizes etc. locally..

aojea:
jobs failing with these errors 
```calico-node-2cp7k kind-worker2       Running       [{Type:Initialized Status:True LastProbeTime:0001-01-01 00:00:00 +0000 UTC LastTransitionTime:2019-05-14 17:02:22 +0000 UTC Reason: Message:} {Type:Ready Status:False LastProbeTime:0001-01-01 00:00:00 +0000 UTC LastTransitionTime:2019-05-14 17:11:34 +0000 UTC Reason:ContainersNotReady Message:containers with unready status: [calico-node]} {Type:ContainersReady Status:False LastProbeTime:0001-01-01 00:00:00 +0000 UTC LastTransitionTime:2019-05-14 17:11:34 +0000 UTC Reason:ContainersNotReady Message:containers with unready status: [calico-node]} {Type:PodScheduled Status:True LastProbeTime:0001-01-01 00:00:00 +0000 UTC LastTransitionTime:2019-05-14 17:02:13 +0000 UTC Reason: Message:}]```

File: pkg/build/node/const.go [hunk 1]
```diff
@@ -24,225 +24,771 @@ const (
...
+            # Set Felix endpoint to host default action to ACCEPT.
+            - name: FELIX_DEFAULTENDPOINTTOHOSTACTION
+              value: "ACCEPT"
+            # Disable IPv6 on Kubernetes.
+            - name: FELIX_IPV6SUPPORT
+              value: "false"
+            # Set Felix logging to "info"
+            - name: FELIX_LOGSEVERITYSCREEN
+              value: "info"
+            - name: FELIX_HEALTHENABLED
```
aojea: Try adding this
```
- name: FELIX_IGNORELOOSERPF
  value: true
```

aojea:
The calico node has some FATAL errors
```
2019-05-14T17:14:27.508596964Z stdout F 2019-05-14 17:14:27.504 [INFO][1403] int_dataplane.go 504: Registering to report health.
2019-05-14T17:14:27.508605906Z stdout F 2019-05-14 17:14:27.504 [FATAL][1403] int_dataplane.go 980: Kernel's RPF check is set to 'loose'.  This would allow endpoints to spoof their IP address.  Calico requires net.ipv4.conf.all.rp_filter to be set to 0 or 1. If you require loose RPF and you are not concerned about spoofing, this check can be disabled by setting the IgnoreLooseRPF configuration parameter to 'true'.
```
Let´s try to add this option FELIX_IGNORELOOSERPF

pkg/build/node/const.go
```diff
@@ -24,225 +24,771 @@ const (
...
+      - update
+  # Calico stores some configuration information on the node.
+  - apiGroups: [""]
+    resources:
+      - nodes
+    verbs:
+      - get
+      - list
+      - watch
+  # These permissions are only requried for upgrade from v2.6, and can
```
aojea: s/requried/required/ beceause it fails the pull_verify job

BenTheElder: heh, upstream bug :^)

pkg/build/node/const.go
BenTheElder: done

BenTheElder: xref: https://docs.projectcalico.org/v3.5/reference/felix/configuration

BenTheElder:
Tests are taking considerably longer than before. we're over 28 minutes and counting on the kubernetes master branch job. usually these are no more than 22 minutes (and often 18-19)

BenTheElder:
/test all
wildly slow ... seeing if it's a fluke

Issue #528, opened 2019-05-15
Umbrella Issue: Code Quality Improvements


BenTheElder:
**What should be cleaned up or changed**:
- [x] enable staticcheck and fix errors https://staticcheck.io/docs/
- [x] factor out all exec / docker interaction into interfaces for mocking and implementing against podman etc.
  -  [x] eliminate the caching mechanism on node handles, instead provide a caching wrapper for this docker interface, and only use it by default in the CLI (library users should need to opt into caching since we don't know how long lived their processes are and we probably won't have good invalidation)
- [x] unit tests, as the code grows and matures, we should write more of them! [WIP]
- [x] [shellcheck](https://www.shellcheck.net/) our bash. it's not going anywhere for the little dev scripts, and we run a tiny bit in the node entrypoint now. we should ensure this is high quality. we can create a simpler variation on my kubernetes/kubernetes shellcheck script.

**Why is this needed**:
To ensure kind's code improves in quality and stays that way.

/priority important-longterm
/assign

Will work on this more after the 0.3 release. Plan to go ahead and get some smaller pieces in while some tests soak.

aojea:
@BenTheElder should we add some e2e smoke tests for features like the load balancer and port mapping?

BenTheElder:
@aojea I HA mode?

Eventually yes, but lower priority, to my knowledge nothing depends on that yet. (And shouldn't!)

We also should have smoke tests with the default image.

nzoueidi:
I might help with the two last points :)

alisondy:
another one for the list potentially?
Hadolint our docker files see [here](https://github.com/hadolint/hadolint)

BenTheElder:
#584 and #591 added a shellcheck linter and eliminated failures (except the one lint we turn off regarding command -v vs which)

BenTheElder:
Good question.

It's pretty wildly unusual to have a docker daemon running but no docker client binary so we're not trying to avoid depending on the binary, however we are trying to let kind work with other runtimes (podman).

(Also the docker client library
 is pretty terrible, kubelet is littered with comments about trying to figure out how to use it correctly. Users will have a much better time debugging when they can see the CLI calls by adjusting the log level)

glerchundi:
Sounds reasonable but there could be a common use case where you are
running kind inside a container that has access to a docker daemon for CI
purposes.

Just putting my 2 cents.

Great job by the way 👍🏻

BenTheElder:
[...]

PR #529, opened 2019-05-15
add staticcheck to verify

BenTheElder:
... and fix something it found in the fuzzer tests! (second commit)

part of #528


hack/verify-staticcheck.sh
```diff
@@ -0,0 +1,39 @@
...
+set -o pipefail
+
+# cd to the repo root
+REPO_ROOT=$(git rev-parse --show-toplevel)
+cd "${REPO_ROOT}"
+
+# enable modules and the proxy cache
+export GO111MODULE="on"
+GOPROXY="${GOPROXY:-https://proxy.golang.org}"
+export GOPROXY
```
tao12345666333: can use  `export GOPROXY="${GOPROXY:-https://proxy.golang.org}"` directly

BenTheElder: :man_shrugging: probably. but for now I'd prefer to be consistent with the other scripts

BenTheElder:
corrected the copyright and shebang

Issue #205, opened 2019-01-09
support for custom CNIs


BenTheElder:
this could potentially even be the default.

more generally we should look into suporting multiple other CNI
while low-prio, it is something we should _eventually_ be doing, for sure

/priority important-longterm
/kind feature

mauilion:
https://github.com/kubernetes-sigs/kind/issues/278

mauilion:
Regard #278

BenTheElder:
After https://github.com/kubernetes-sigs/kind/pull/331 it will be pretty straightforward to load an alternative CNI with a custom image at least, we still need to do #278 as well which should be pretty straightforward, will put that in the next minor milestone after 0.2 tentatively.

aojea:
I think that one simple solution to this problem can be adding a new field `CNI` under the `networking` struct in the new v1alpha3 API #340 
This CNI field should be a string pointing to a remote or local manifest, if this field is empty we continue using current implementation. 

```apiVersion: kind.sigs.k8s.io/v1alpha3
kind: Cluster
networking:
  CNI: https://raw.githubusercontent.com/coreos/flannel/master/Documentation/kube-flannel.yml
```
For air-gapped environment, we can deliver manifests for each plugin that uses a local registry #110 avoiding to have to recreate continuously the node-image

neolit123:
> CNI: https://raw.githubusercontent.com/coreos/flannel/master/Documentation/kube-flannel.yml`

or make the CNI field accept a mount path too...
something to note is that some plugins have more than one YAML file.

> For air-gapped environment, we can deliver manifests for each plugin that uses a local registry #110 avoiding to have to recreate continuously the node-image or to ship the images and load them with kind load

we don't have to ship the images as long as the custom CNI process is document in kind.
also a custom registry seems like an overkill for this.

also, we have to keep in mind that custom CNI falls under advanced testing use cases and making this easy to use from kind shouldn't be a primary goal IMO. e.g. "kind does support custom CNI, but you need to do these steps"..

aojea:
@neolit123 thanks, the mount path works with the current version using the `extraMounts` option and overwriting the `/kind/manifests/default-cni.yaml` file

```
# this config file contains all config fields with comments
kind: Config
apiVersion: kind.sigs.k8s.io/v1alpha2
# 3 control plane node and 3 workers
nodes:
# the control plane node config
- role: control-plane
  replicas: 1
  extraMounts:
[...]

PR #530, opened 2019-05-15
support disabling default cni

BenTheElder:
fixes https://github.com/kubernetes-sigs/kind/issues/205 (though perhaps we need to follow up with docs :upside_down_face: )

With this feature, users can skip the CNI installation and instead install their own CNI using the usual mechanisms (`kubectl apply -f my-cni-manifest.yaml ...`). This is an advanced use case and not strongly supported, but should be available. It seems a number of projects have found installing / testing their own CNI useful, and this makes it easier to do without poking at kind's internals :wink:

### DIFF
--- a/pkg/cluster/internal/create/create.go
+++ b/pkg/cluster/internal/create/create.go
@@ -86,8 +86,16 @@ func Cluster(ctx *context.Context, cfg *config.Cluster, opts *Options) error {
 	}
 	if opts.SetupKubernetes {
 		actionsToRun = append(actionsToRun,
-			kubeadminit.NewAction(),                   // run kubeadm init
-			installcni.NewAction(),                    // install CNI
+			kubeadminit.NewAction(), // run kubeadm init
+		)
+		// this step might be skipped, but is next after init
+		if !cfg.Networking.DisableDefaultCNI {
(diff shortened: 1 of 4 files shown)


pkg/cluster/internal/create/create.go
```diff
@@ -86,8 +86,16 @@ func Cluster(ctx *context.Context, cfg *config.Cluster, opts *Options) error {
...
 	if opts.SetupKubernetes {
 		actionsToRun = append(actionsToRun,
-			kubeadminit.NewAction(),                   // run kubeadm init
-			installcni.NewAction(),                    // install CNI
+			kubeadminit.NewAction(), // run kubeadm init
+		)
+		// this step might be skipped, but is next after init
+		if !cfg.Networking.DisableDefaultCNI {
+			actionsToRun = append(actionsToRun,
+				installcni.NewAction(), // install CNI
```
BenTheElder: we could alternatively guard against this inside the action, I'm not sure this detail is important currently though as both are wholly internal to kind and should have roughly the same effect, but that _might_ be a better refactor in the future.

BenTheElder:
cc @mauilion @aojea :-)

PR #533, opened 2019-05-15
use fmt.Println instead of println

BenTheElder:
println uses stderr which is surprising and unintended. supersedes https://github.com/kubernetes-sigs/kind/pull/203


pkg/build/kube/bazelbuildbits.go
```diff
@@ -136,7 +136,7 @@ func fixOldImageTags(path, arch string) error {
 	archSuffix := "-" + arch
 	repositoryFixer := func(repository string) string {
 		if !strings.HasSuffix(repository, archSuffix) {
-			println("fixed: " + repository + " -> " + repository + archSuffix)
+			fmt.Println("fixed: " + repository + " -> " + repository + archSuffix)
```
BenTheElder: we may actually want to redo this entirely, but that only happens in < 1.12 with bazel, and i'll leave that for another PR

PR #535, opened 2019-05-15
patch update-generated.sh

BenTheElder:
- tidy before vendor to minimize
- quote variables properly


BenTheElder:
this is low priority but related to #528

PR #537, opened 2019-05-15
normalize .gitignore

BenTheElder:
unimportant, but this causes my editor to correctly not-highlight ignored directories and makes it more consistent.


BenTheElder:
good idea, will come back to this

BenTheElder:
On second thought /shrug on the binary. Can add that another time if we need it.
/hold cancel

PR #538, opened 2019-05-15
upgrade ip-masq-agent to v2.4.1

BenTheElder:
should have some fixes, in particular we need this because of upstream patches I sent to CIDR validation


aojea:
/lgtm

go go ipv6 go :)

PR #540, opened 2019-05-16
Build cleanup

BenTheElder:
I want to land this before 0.3 -- `make install` is not the right default target yet, since without go installed it will still basically install to `$GOPATH/bin`, we need time to figure out what the behavior should be then.

Instead:
- suggest `make build`
- the make build prints steps and the target output, eg a cached build (less output):
```console
$ make build
+ Ensuring build cache volume exists
docker volume create kind-build-cache
kind-build-cache
+ Ensuring build output directory exists
mkdir -p /usr/local/google/home/bentheelder/go/src/sigs.k8s.io/kind/bin
+ Building kind binary
docker run \
        --rm \
        -v kind-build-cache:/go \
        -e GOCACHE=/go/cache \
        -v /usr/local/google/home/bentheelder/go/src/sigs.k8s.io/kind/bin:/out \
        -v /usr/local/google/home/bentheelder/go/src/sigs.k8s.io/kind:/src/kind \
        -w /src/kind \
        -e GO111MODULE=on \
        -e GOPROXY=https://proxy.golang.org \
        -e CGO_ENABLED=0 \
        -e GOOS=linux \
        -e GOARCH=amd64 \
        --user 484827:89939 \
        golang:1.12.5 \
        go build -v -o /out/kind .
+ Built kind binary to /usr/local/google/home/bentheelder/go/src/sigs.k8s.io/kind/bin/kind
```

----

Also other fixes for the release:
- make some make variables overrideable properly
- update cross.sh (used when releasing) to use the reproducible build
- update the release instructions to use v notation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,22 +43,27 @@ all: build
 
 # creates the cache volume
 make-cache:
+	@echo + Ensuring build cache volume exists
 	docker volume create $(CACHE_VOLUME)
 
 # cleans the cache volume
 clean-cache:
+	@echo + Removing build cache volume
 	docker volume rm $(CACHE_VOLUME)
 
 # creates the output directory
 out-dir:
+	@echo + Ensuring build output directory exists
 	mkdir -p $(OUT_DIR)
 
 # cleans the output directory
 clean-output:
+	@echo + Removing build output directory
 	rm -rf $(OUT_DIR)/
 
 # builds kind in a container, outputs to $(OUT_DIR)
 kind: make-cache out-dir
+	@echo + Building kind binary
 	docker run \
 		--rm \
 		-v $(CACHE_VOLUME):/go \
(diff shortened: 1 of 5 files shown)


BenTheElder:
/hold
tossing on some more fixes for releasing

Makefile
```diff
@@ -18,18 +18,20 @@
 # settings
 REPO_ROOT:=${CURDIR}
 # autodetect host GOOS and GOARCH by default, even if go is not installed
-GOOS=$(shell hack/util/goos.sh)
-GOARCH=$(shell hack/util/goarch.sh)
+GOOS?=$(shell hack/util/goos.sh)
+GOARCH?=$(shell hack/util/goarch.sh)
```
BenTheElder: https://www.gnu.org/software/make/manual/html_node/Setting.html#Setting

site/content/docs/user/quick-start.md
```diff
@@ -23,7 +23,7 @@ This will put `kind` in `$(go env GOPATH)/bin`. You may need to add that directo
 shown [here](https://golang.org/doc/code.html#GOPATH) if you encounter the error
 `kind: command not found` after installation.
 
-Without installing go, kind can be built reproducibly with docker using `make install`.
+Without installing go, kind can be built reproducibly with docker using `make build`.
```
aojea: Do we need to change L18 too_

BenTheElder: yes, thank you.

PR #545, opened 2019-05-17
kindnet 0.2.0

aojea:
Implementing the ipmasq-agent functionality inside kindnet , based on Ben´s suggestion, reducing the number of images needed for the cni and simplifying the CNI configuration, it also adds ipv6 support to the NO-MASQ functionality. We only need to NO-MASQ the traffic between pods, and we have that information, no need to get the podSubnet/Cluster-CIDR.

Improved performance switching from the `bridge` CNI plugin to `ptp`


BenTheElder:
(previously suggested this as an option that may actually be simpler and cheaper)

aojea:
🤔  the CNIconfig writer is not working 
```
              "subnet": "{{ .PodCIDR }}"
```

images/kindnetd/Dockerfile
```diff
@@ -27,6 +27,6 @@ COPY . .
 RUN go build -o /go/bin/kindnetd .
 
 # STEP 2: Build small image
-FROM scratch
+FROM gcr.io/google-containers/debian-iptables-amd64:v11.0.2
```
aojea: this should be `debian-iptables-${GOARCH}:v11.0.2`?

it gives me an error using it

aojea:
/assign @BenTheElder 
/hold cancel

it works with IPv6 too 
```
A POSTROUTING -m comment --comment "kubernetes postrouting rules" -j KUBE-POSTROUTING
-A POSTROUTING -m comment --comment "kind-masq-agent: ensure nat POSTROUTING directs all non-LOCAL destination traffic to our custom KIND-MASQ-AGENT chaT
-A KIND-MASQ-AGENT -d fd00:10:244::/80 -m comment --comment "kind-masq-agent: local traffic is not subject to MASQUERADE" -j RETURN
-A KIND-MASQ-AGENT -d fd00:10:244:0:1::/80 -m comment --comment "kind-masq-agent: local traffic is not subject to MASQUERADE" -j RETURN
-A KIND-MASQ-AGENT -m comment --comment "ip-masq-agent: outbound traffic is subject to MASQUERADE (must be last in chain)" -j MASQUERADE
-A KUBE-MARK-DROP -j MARK --set-xmark 0x8000/0x8000
```

pkg/build/node/const.go
```diff
@@ -27,7 +27,7 @@ const (
 The default CNI manifest and images are our own tiny kindnet
 */
 
-var defaultCNIImages = []string{"kindest/kindnetd:0.1.0", "k8s.gcr.io/ip-masq-agent:v2.4.1"}
+var defaultCNIImages = []string{"aojea/kindnetd:0.2.0"}
```
aojea: let me know when should I replace by `kindest`

pkg/build/node/const.go
```diff
@@ -186,7 +186,7 @@ spec:
       serviceAccountName: kindnet
       containers:
       - name: kindnet-cni
-        image: kindest/kindnetd:0.1.0
+        image: aojea/kindnetd:0.2.0
```
aojea: let me know when should I replace by `kindest`

images/kindnetd/Dockerfile
BenTheElder: We have to pass through GOARCH as an ARG both before and after FROM, otherwise it won't be available (more details in the docker docs for this, on mobile or would link)

pkg/build/node/cni.go
```diff
@@ -186,7 +185,7 @@ spec:
       serviceAccountName: kindnet
       containers:
       - name: kindnet-cni
-        image: kindest/kindnetd:0.1.0
+        image: aojea/kindnetd:0.2.0
```
aojea: @BenTheElder we have to create our own image before merging

pkg/build/node/cni.go
```diff
@@ -27,11 +27,10 @@ const (
 The default CNI manifest and images are our own tiny kindnet
 */
 
-var defaultCNIImages = []string{"kindest/kindnetd:0.1.0", "k8s.gcr.io/ip-masq-agent:v2.4.1"}
+var defaultCNIImages = []string{"aojea/kindnetd:0.2.0"}
```
aojea: Point to our own image

aojea:
@BenTheElder we should check this soon, it also avoids some random issues  with the bridge plugin (sometimes the bridge is not created correctly)

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/568 caused a need for rebase

pkg/cluster/internal/create/actions/installcni/cni.go
```diff
@@ -51,43 +47,11 @@ func (a *action) Execute(ctx *actions.ActionContext) error {
 		return err
 	}
 
-	// read the manifest from the node
-	var raw bytes.Buffer
```
BenTheElder: removing this breaks compatibility with certain kind images for pretty much zero gain

aojea: got it, never thought about it 😅

images/kindnetd/Dockerfile
```diff
@@ -27,6 +27,6 @@ COPY . .
 RUN go build -o /go/bin/kindnetd .
 
 # STEP 2: Build small image
-FROM scratch
+FROM gcr.io/google-containers/debian-iptables:v11.0.2
```
BenTheElder: we need to use the arch specific image and cross build :sweat_smile:

aojea: I´ll need some guidance here, I couldn´t find info or some example on how to do it

aojea: @BenTheElder done

aojea:
@BenTheElder we need the new kindest/kindnetd:0.2.0 images in order to test it

cmd/kindnetd/go.mod
```diff
@@ -22,6 +23,7 @@ require (
 	k8s.io/api v0.0.0-20190409021203-6e4e0e4f393b
 	k8s.io/apimachinery v0.0.0-20190404173353-6a84e37a896d
 	k8s.io/client-go v11.0.0+incompatible
-	k8s.io/utils v0.0.0-20190506122338-8fab8cb257d5 // indirect
+	k8s.io/kubernetes v1.14.2
```
BenTheElder: here we go! :D :arrows_clockwise:

pkg/build/node/cni.go
```diff
@@ -214,135 +213,10 @@ spec:
...
-  name: ip-masq-agent
-  namespace: kube-system
-  labels:
-    tier: node
-    app: ip-masq-agent
-    k8s-app: ip-masq-agent
-data:
-  config: |-
-    nonMasqueradeCIDRs:
-      - {{ .PodSubnet }}
```
BenTheElder: to me this is a better seperation of concerns, but we can address that in a future change

cmd/kindnetd/main.go
```diff
@@ -127,12 +174,54 @@ func makeNodesReconciler(cniConfigWriter *CNIConfigWriter, hostIP string) func(*
 		}
 
 		fmt.Printf("Handling node with IP: %s\n", nodeIP)
+		fmt.Printf("Node %v has CIDR %s \n", node.Name, podCIDR)
+		kindnetd.config.PodCIDRs[podCIDR] = nodeIP
```
BenTheElder: this is not really a reconcile and I don't like this separation of concerns, but functionally this _works_

BenTheElder:
/retest
images pushed

BenTheElder:
/hold cancel
/lgtm
/approve
I want to refactor this code heavily, but functionally this is the right move

pkg/build/node/cni.go
aojea: well, per k8s network design the pod2pod traffic must not be Masquerade.
We are obtaining other nodes pod subnet to install a route to them ... and we have to route and not masquerade that traffic ... however, if we are able to get the whole PodSubnet it will allow us to separate them

Issue #507, opened 2019-05-08
Document different hacks to expose ports


aojea:
**What would you like to be documented**:

We should document the different methods to forward ports so users can access cluster services in all the supported OS (Mac, Linux, Windows)

**Why is this needed**:

Supporting port forwards is very demanded #459 #99  and there are several hacks that can be used.
Aggregating and documenting the different options can give users different choices and help the `kind` community to understand better the use cases.

BenTheElder:
[also just noting that we intend to support something first class, these are options that can potentially be enabled with built-in support]

aojea:
I think that we had some kind of agreement that ideally we should have a `kind` load balancer provider that allows users to reach services with Virtual IPs independently of the OS

michael-burt:
Is there a standard way to simply port-forward into a pod? Is there documentation about how to accomplish this?

aojea:
@michael-burt please check the draft document and provide your feedback https://github.com/kubernetes-sigs/kind/pull/546 in the PR

BenTheElder:
`kubectl port-forward` is the most standard way applicable to any kubernetes setup not just kind. but we'd like something a little more robust for kind.

on linux you can also reach the node IPs directly and go from there (EG with a nodePort)

BenTheElder:
not sure how much we need this now that we have extra ports via https://github.com/kubernetes-sigs/kind/issues/99#issuecomment-505662768

BenTheElder:
(which is documented)

aojea:
Indeed

PR #546, opened 2019-05-20
[WIP] how to access services inside a cluster

aojea:
Explain how to access services and applications inside a `kind` cluster

Ref: #507

### DIFF
--- a/site/content/docs/user/exposing-applications.md
+++ b/site/content/docs/user/exposing-applications.md
@@ -135,11 +135,16 @@ Events:                   <none>
 ```
 
 
-### Nginx ingress controller
+### Ingress
 
 > Ingress is actually NOT a type of service. Instead, it sits in front of multiple services and act as a “smart router” or entrypoint into your cluster.
 
-TODO
+You can install your ingress controller, i.e. the [NGINX Ingress Controller], and expose it using a NodePort or a Cluster IP. 
+
+### LoadBalancer
+
+> Exposes the service externally using a cloud provider’s or custom load balancer. 
+
 
 ## Accesing the application
 
@@ -172,7 +177,18 @@ https://sosedoff.com/2018/04/25/expose-docker-ports.html
 
 https://andrewwilkinson.wordpress.com/2017/11/22/exposing-docker-ports-after-the-fact/
 
+In our example we can connect to the exposed nodePort in one of the cluster nodes:
 
+```sh
+docker run --rm -d -p80:8080 --name=proxy alpine/socat TCP4-LISTEN:8080,fork,reuseaddr TCP4:172.17.0.4:30039
+```
+
+and access the service in our host:
+
+```sh
+$ curl localhost
+Hello Kubernetes!
+```
 
 
 ### Socks Proxy
@@ -186,7 +202,7 @@ SOCKS proxy can be used with any web browser, just have to configure your applic
 Create a container with a SOCKS proxy and forward the port
 
 ```sh
-docker run --rm -d --privileged --name socks5 -p 1080:1080 aojea/socks5
+docker run --rm -d --cap-add=NET_ADMIN --name socks5 -p 1080:1080 aojea/socks5
 ```
 
 Accessing the NodePort service, for that we need only the IP of one of the cluster nodes:
@@ -211,10 +227,11 @@ curl --socks5 localhost:1080  10.96.59.230:8080
 Hello Kubernetes!
 ```
 
+You can use other browser like Google Chrome or Mozilla Firefox too.
 
 ### tun2socks
 
-This is the most complex option but with a better UX, it consists in creating a tun interface in the host and route all traffic that belongs to the k8s cluster through it.
+This is the most complex option but with the best UX, it consists in creating a tun interface in the host and route all traffic that belongs to the k8s cluster through it.
 
 Internally it uses the SOCKS proxy server that we previously deployed in the docker network and forwarded to our host.
 
@@ -241,3 +258,4 @@ https://github.com/FlowerWrong/tun2socks
 [kubernetes exposing services]: https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types
 [Docker for Mac networking]: https://docs.docker.com/docker-for-mac/networking/
 [Docker for Windows networking]: https://docs.docker.com/docker-for-windows/networking/
+[NGINX Ingress Controller]: https://kubernetes.github.io/ingress-nginx/deploy/
--- a/site/content/docs/user/exposing-applications.md
+++ b/site/content/docs/user/exposing-applications.md
@@ -1,0 +1,243 @@
+---
+title: "Accesing applications inside the cluster"
+menu:
+  main:
+    parent: "user"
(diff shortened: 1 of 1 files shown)


site/content/docs/user/exposing-applications.md
```diff
@@ -0,0 +1,242 @@
...
+kind-worker2         Ready    <none>   5m59s   v1.14.1   172.17.0.3    <none>        Ubuntu Disco Dingo (development branch)   4.9.125-linuxkit   containerd://1.2.6-0ubuntu1
+```
+
+### Kubectl port-forward
+
+This is the most simple one and well documented
+
+https://kubernetes.io/docs/tasks/access-application-cluster/port-forward-access-application-cluster/
+
+### Port forwarding with socat
```
aojea: This can be socat or our nginx load balancer, the trick here is to keep adding ports to the same container, i.e.:

Expose localhost 8080 on nodeA:9090
docker run -p 8080 node_proxy "command to proxy to nodeA9090"

if we want to add a new port we recreate the container with the new data
docker run -p 8080 -p <new_port> node_proxy "command to proxy to nodeA9090" "command to proxy <newport>"

site/content/docs/user/exposing-applications.md
```diff
@@ -0,0 +1,242 @@
...
+
+And now we can access the application using the the Service IP and Service port:
+
+```sh
+curl --socks5 localhost:1080  10.96.59.230:8080
+Hello Kubernetes!
+```
+
+
+### tun2socks
```
aojea: I think that this is the best solution and along the lines we discussed, the host has to create the routes to the service subnet through the tun interface, and the tun2socks functionality will forward it to the SOCKS server inside the docker network, so the user won't have to care about anything, he will be able to access the services using their service IPs.

site/content/docs/user/exposing-applications.md
```diff
@@ -0,0 +1,243 @@
...
+> A SOCKS server is a general purpose proxy server that establishes a TCP connection to another server on behalf of a client, then routes all the traffic back and forth between the client and the server. It works for any kind of network protocol on any port
+
+Instead of having to recreate the port mapping constantly we can create a socks proxy inside the docker network and use it to connect to the exposed applications.
+
+SOCKS proxy can be used with any web browser, just have to configure your application to use it and use your service IP and Port target directly.
+
+Create a container with a SOCKS proxy and forward the port
+
+```sh
+docker run --rm -d --privileged --name socks5 -p 1080:1080 aojea/socks5
```
tao12345666333: Probably don't need such a big privilege? May give it NET_ADMIN permission. Using --cap-add NET_ADMIN

(But I need to test it later.)

tao12345666333: > Probably don't need such a big privilege? May give it NET_ADMIN permission. Using --cap-add NET_ADMIN
> 
> (But I need to test it later.)

 Tested, as I said we can just give it `NET_ADMIN` permission only.  :smile:

site/content/docs/user/exposing-applications.md
```diff
@@ -0,0 +1,243 @@
...
+ ✓ Ensuring node image (kindest/node:v1.14.1) 🖼
+ ✓ Preparing nodes 📦📦📦
+ ✓ Creating kubeadm config 📜
+ ✓ Starting control-plane 🕹️
+ ✓ Installing CNI 🔌
+ ✓ Installing StorageClass 💾
+ ✓ Joining worker nodes 🚜
+Cluster creation complete. You can now use the cluster with:
+
+export KUBECONFIG="$(kind get kubeconfig-path --name="kind")"
```
Ilyes512: `--name="kind"` is not needed as it's the default. If you want to be more explicit then it should be added to `kind create cluster --config config-ipv4-ci.yaml` as well to be consistent :)

aojea: this is a copy paste from the output of the `kind` command 😅

site/content/docs/user/exposing-applications.md
```diff
@@ -0,0 +1,261 @@
...
+```  
+
+And now we can access the application using the the Service IP and Service port:
+
+```sh
+curl --socks5 localhost:1080  10.96.59.230:8080
+Hello Kubernetes!
+```
+
+You can use other browser like Google Chrome or Mozilla Firefox too.
```
csantanapr: maybe state that the browser would need to be configured with socks proxy, and maybe links to a doc for each browser mentioned that explains the details.

site/content/docs/user/exposing-applications.md
```diff
@@ -0,0 +1,261 @@
...
+
+NAME                 STATUS   ROLES    AGE     VERSION   INTERNAL-IP   EXTERNAL-IP   OS-IMAGE                                  KERNEL-VERSION     CONTAINER-RUNTIME
+kind-control-plane   Ready    master   7m1s    v1.14.1   172.17.0.4    <none>        Ubuntu Disco Dingo (development branch)   4.9.125-linuxkit   containerd://1.2.6-0ubuntu1
+kind-worker          Ready    <none>   5m58s   v1.14.1   172.17.0.2    <none>        Ubuntu Disco Dingo (development branch)   4.9.125-linuxkit   containerd://1.2.6-0ubuntu1
+kind-worker2         Ready    <none>   5m59s   v1.14.1   172.17.0.3    <none>        Ubuntu Disco Dingo (development branch)   4.9.125-linuxkit   containerd://1.2.6-0ubuntu1
+```
+
+### Kubectl port-forward
+
+This is the most simple one and well documented
```
csantanapr: This is the one that I use most of the time maybe worth putting an example, some folks think this is only for pod. But portforward can do a deployment and service also.

For most demos and tutorials I recommend to folks to use a service.
```
kubectl port-forward service/myservice 8080:8080
```
Then to use `locahost:8080` on their browser of if they need a hostname for routing like knative or istio then use a domain lile `xip.io` or `nip.io` using lo `127.0.0.1`
```
curl http://127.1.0.0.1.nip.io
open http://127.1.0.0.1.xip.io
```

anapsix:
Any Ingress Controller could be used to expose services running in _kind cluster_
- in a case of a single-node cluster, launching container with published port takes care of ingress controller scenario, with ingress using pod with `containerPort`, or service with `nodePort`
- in case of multi-node cluster, ingress controller could be deployed as a single replica, mirroring single-node cluster setup, which should be enough to handle local development.
- for running ingress controller as daemonset in multi-node cluster environment, each node similarly to single-node cluster would have to publish ports, but a unique host port per cluster node
  plus some load-balancer container instance doing round-robin balancing to individual instances of ingress, serving as ingress gateway

Allowing to specify a published port binding per worker node in kind cluster init configuration would allow any user to implement preferred single instance ingress controller

BenTheElder:
> Allowing to specify a published port binding per worker node in kind cluster init configuration would allow any user to implement preferred single instance ingress controller

this is in progress

aojea:
we have a `kind` way to expose the ports, will hold until https://github.com/kubernetes-sigs/kind/pull/637 is merged

PR #560, opened 2019-05-25
deal with systemd MAC address assignment

ytinirt:
Signed-off-by: ZHAO Yao <zhao.yao@h3c.com>

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -121,6 +121,8 @@ COPY [ "entrypoint", "/usr/local/bin/" ]
 # NOTE: this is *only* for documentation, the entrypoint is overridden by the node image
 ENTRYPOINT [ "/usr/local/bin/entrypoint", "/sbin/init" ]
 
-# TODO(bentheelder): deal with systemd MAC address assignment
+# deal with systemd MAC address assignment
 # https://github.com/systemd/systemd/issues/3374#issuecomment-288882355
 # https://github.com/systemd/systemd/issues/3374#issuecomment-339258483
+RUN cp /usr/lib/systemd/network/99-default.link /etc/systemd/network/ \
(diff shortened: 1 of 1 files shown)


images/base/Dockerfile
```diff
@@ -121,6 +121,8 @@ COPY [ "entrypoint", "/usr/local/bin/" ]
 # NOTE: this is *only* for documentation, the entrypoint is overridden by the node image
 ENTRYPOINT [ "/usr/local/bin/entrypoint", "/sbin/init" ]
 
-# TODO(bentheelder): deal with systemd MAC address assignment
+# deal with systemd MAC address assignment
 # https://github.com/systemd/systemd/issues/3374#issuecomment-288882355
 # https://github.com/systemd/systemd/issues/3374#issuecomment-339258483
+RUN cp /usr/lib/systemd/network/99-default.link /etc/systemd/network/ \
```
aojea: Is this file guarantee to be there in newer versions?

images/base/Dockerfile
```diff
@@ -121,6 +121,8 @@ COPY [ "entrypoint", "/usr/local/bin/" ]
 # NOTE: this is *only* for documentation, the entrypoint is overridden by the node image
 ENTRYPOINT [ "/usr/local/bin/entrypoint", "/sbin/init" ]
 
-# TODO(bentheelder): deal with systemd MAC address assignment
+# deal with systemd MAC address assignment
```
aojea: @BenTheElder can you give more context on this issue? I don´t remember related failures

BenTheElder: it's just a known issue with some virtual devices and systemd that I noted for the future, I'm not sure that this is in fact an issue we've had an problem with yet. so far as far as I can tell it is not.

BenTheElder: it might be equally valid to just remove this comment

BenTheElder: we went that route, thanks again!
https://github.com/kubernetes-sigs/kind/pull/590
/close

PR #564, opened 2019-05-26
Allow make build to use proxies

### LINKED ISSUE #551: Make behind a proxy
**What happened**:
Although http_proxy, https_proxy and no_proxy env var are configured, `make build` fails witha lot of:
go: golang.org/x/net@v0.0.0-20190514140710-3ec191127204: Get https://proxy.golang.org/golang.org/x/net/@v/v0.0.0-20190514140710-3ec191127204.info: dial tcp: lookup proxy.golang.org on x.x.x.x:53: no such host

Make success

Run `make build` behind a proxy (docker daemon must also be correctly configured)

Adding 
``` -e http_proxy=$(http_proxy) \``` 
to `Makefile` in `docker run` makes the build pass.

**Environment:**

- kind version: (use `kind version`): 0.4.0-alpha
- Kubernetes version: (use `kubectl version`): not yet installed
- Docker version: (use `docker info`): 18.09.6
- OS (e.g. from `/etc/os-release`): CentOS Linux release 7.6.1810 (Core)

### PR DESCRIPTION
**Pull request by aojea**
`make build` uses docker to build `kind`.
On environments that use proxy we need to pass this variables to the container so it can use it to fetch the dependencies.

Fixes #551

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,6 +78,9 @@ kind: make-cache out-dir
 		-e CGO_ENABLED=0 \
 		-e GOOS=$(GOOS) \
 		-e GOARCH=$(GOARCH) \
+		-e HTTP_PROXY=$(HTTP_PROXY) \
+		-e HTTPS_PROXY=$(HTTPS_PROXY) \
+		-e NO_PROXY=$(NO_PROXY) \
 		--user $(UID):$(GID) \
 		$(GO_IMAGE) \
 		go build -v -o /out/$(KIND_BINARY_NAME) .

### CODE REVIEW

**Comment by tao12345666333** [comment]
Agree with your code.  :smile: 

/lgtm

Not sure if this is good or bad, will it bring other problems? https://github.com/kubernetes-sigs/kind/issues/551#issuecomment-495271837  

I usually execute `export http_proxy=socks5://127.0.0.1:1081 https_proxy=socks5://127.0.0.1:1081` to configure the proxy.

**Comment by aojea** [comment]
🤔  I'm only considering uppercase environment variables, should we consider lowercase too?

**Comment by tao12345666333** [comment]
If we support proxy, then obviously uppercase and lowercase should support it.

File: Makefile [hunk 1]
```diff
@@ -78,6 +78,9 @@ kind: make-cache out-dir
 		-e CGO_ENABLED=0 \
 		-e GOOS=$(GOOS) \
 		-e GOARCH=$(GOARCH) \
+		-e HTTP_PROXY=$(HTTP_PROXY) \
```
BenTheElder: this is a start, we can consolidate lowercase to these above later if absolutely necessary, but imho we should try to keep this very simple, I'm not aiming to grow a "build system", just a simple alternative to running go natively to avoid complaints about packaged go versions and give us a more reproducible option for releases :sweat_smile:

PR #572, opened 2019-05-29
Document how to enable feature gates

### LINKED ISSUE #563: Using Feature Gates
How to specify [feature gates](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/) in the cluster created by kind.

**Why is this needed**:
Sometimes I would like to use and test Alpha features with kind, and I don't quite see how to set this up.

aojea:
Describe how to enable different gates in your `kind` cluster patching `kubeadm`

Fixes #563

### DIFF
--- a/site/content/docs/user/quick-start.md
+++ b/site/content/docs/user/quick-start.md
@@ -259,13 +259,13 @@ kubeadmConfigPatches:
     name: config
   apiServer:
     extraArgs:
-      "feture-gates": "FeatureGateName=true"
+      "feature-gates": "FeatureGateName=true"
   scheduler:
     extraArgs:
-      "feture-gates": "FeatureGateName=true"
+      "feature-gates": "FeatureGateName=true"
   controllerManager:
     extraArgs:
-      "feture-gates": "FeatureGateName=true"
+      "feature-gates": "FeatureGateName=true"
 - |
   apiVersion: kubeadm.k8s.io/v1beta1
   kind: InitConfiguration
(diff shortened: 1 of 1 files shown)


site/content/docs/user/quick-start.md
```diff
@@ -242,6 +242,49 @@ nodes:
...
+# patch the generated kubeadm config with some extra settings
+kubeadmConfigPatches:
+- |
+  apiVersion: kubeadm.k8s.io/v1beta1
+  kind: ClusterConfiguration
+  metadata:
+    name: config
+  apiServer:
+    extraArgs:
+      "feture-gates": "FeatureGateName=true"
```
matthyx: Is it really spelled `feture-gates`? I haven't found anywhere in k/k this way of writing **feature**...

aojea: great catch, thanks

site/content/docs/user/quick-start.md
```diff
@@ -242,6 +242,49 @@ nodes:
 - role: worker
 ```
 
+### Enable Feature Gates in Your Cluster
```
BenTheElder: should this be in another page? the "quick start" seems to be covering a lot now :thinking:

aojea: indeed, but I think that we should first thing how do we want to reorganize the docs, seems that they are starting to grow and I think that adding an advanced section with this stuff is not a good solution because we are going to have the same problem after some time.

PR #578, opened 2019-05-31
Support node/cluster as aliase

clyang82:
Support `kind get node` and `kind get cluster`

Signed-off-by: Chun Lin Yang <clyang@cn.ibm.com>

### DIFF
--- a/cmd/kind/get/clusters/clusters.go
+++ b/cmd/kind/get/clusters/clusters.go
@@ -30,9 +30,10 @@ func NewCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Args: cobra.NoArgs,
 		// TODO(bentheelder): more detailed usage
-		Use:   "clusters",
-		Short: "lists existing kind clusters by their name",
-		Long:  "lists existing kind clusters by their name",
+		Use:     "clusters",
+		Aliases: []string{"cluster"},
+		Short:   "lists existing kind clusters by their name",
+		Long:    "lists existing kind clusters by their name",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return runE(cmd, args)
 		},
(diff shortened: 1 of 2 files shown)


cmd/kind/get/clusters/clusters.go
```diff
@@ -30,9 +30,10 @@ func NewCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Args: cobra.NoArgs,
 		// TODO(bentheelder): more detailed usage
-		Use:   "clusters",
-		Short: "lists existing kind clusters by their name",
-		Long:  "lists existing kind clusters by their name",
+		Use:     "clusters",
+		Aliases: []string{"cluster"},
```
neolit123: i'm not a big fan of very similar aliases, but maybe others will be OK with that.

BenTheElder:
my concern with this is that I would expect `kind get node ...` to be able to give me information about a particular node, IE to act like `kubectl get node`, but we don't actually support that. we should probably support that someday, but it's not clear what that would look like for cluster right now, I think that's a future enhancement that needs some more design (perhaps we preserve the cluster config and other inputs in some format...)

`nodes` is pretty clear about obtaining the list.

clyang82:
> my concern with this is that I would expect `kind get node ...` to be able to give me information about a particular node, IE to act like `kubectl get node`, but we don't actually support that. we should probably support that someday, but it's not clear what that would look like for cluster right now, I think that's a future enhancement that needs some more design (perhaps we preserve the cluster config and other inputs in some format...)
> 
> `nodes` (plural) is pretty clear about obtaining the list.

Thanks for sharing your idea. The original idea for this PR is to keep consistent with `kubectl`. for instance: `kubectl get nodes` and `kubectl get node` are no difference. Correct me if I am wrong.

BenTheElder:
> Thanks for sharing your idea. The original idea for this PR is to keep consistent with kubectl. for instance: kubectl get nodes and kubectl get node are no difference. Correct me if I am wrong.

`kubectl get node foo-node` works, which I'd expect if you have such a command. It won't with kind.

PR #579, opened 2019-05-31
Correct the help message for nodes

clyang82:
Signed-off-by: Chun Lin Yang <clyang@cn.ibm.com>

### DIFF
--- a/cmd/kind/get/nodes/nodes.go
+++ b/cmd/kind/get/nodes/nodes.go
@@ -37,8 +37,8 @@ func NewCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Args:  cobra.NoArgs,
 		Use:   "nodes",
-		Short: "lists existing kind nodes by their name",
-		Long:  "lists existing kind nodes by their name",
+		Short: "lists existing nodes for the kind cluster by --name",
+		Long:  "lists existing nodes for the kind cluster by --name",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return runE(flags, cmd, args)
 		},


aojea:
I think that `lists existing kind nodes by their name` reflects better the behavior than `lists existing nodes for the kind cluster by --name`
Anyway, I prefer @BenTheElder or @munnerz to comment since they are native speakers


```
$ kind get nodes --name kind
kind-worker2
kind-control-plane
kind-worker
```

cmd/kind/get/nodes/nodes.go
```diff
@@ -37,8 +37,8 @@ func NewCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Args:  cobra.NoArgs,
 		Use:   "nodes",
-		Short: "lists existing kind nodes by their name",
-		Long:  "lists existing kind nodes by their name",
+		Short: "lists existing nodes for the kind cluster by --name",
+		Long:  "lists existing nodes for the kind cluster by --name",
```
BenTheElder: so these statements are both true, but what the original one is trying to capture is the _output_ format, which is their name(s), I think we want to get that across rather than the flag, which has its own help text.

clyang82: I think there should have some usability issues for the commands listed below:
```
Available Commands:
  clusters        lists existing kind clusters by their name
  kubeconfig-path prints the default kubeconfig path for the kind cluster by --name
  nodes           lists existing kind nodes by their name
```
when I execute `kind get clusters`, the output is 
```
chunlins-mbp:ibmcloud clyang$ kind get clusters
clusterapi
```
That is what I understand list the clusters with their name
But when I execute `kind get nodes`, the output is
```
chunlins-mbp:ibmcloud clyang$ kind get nodes
Error: unknown cluster "kind
```
it seems the `--name` is required.
At least, I do think the description should be similar for `kubeconfig-path` and `nodes`

BenTheElder: `--name` is required for that command because you're listing nodes for a non-default cluster name. it defaults to "kind". This is in the command usage, not the description. `kind get nodes --help`.

> At least, I do think the description should be similar for kubeconfig-path and nodes

 `kubeconfig-path` returns a specific path which uses the `--name` flag to identify the cluster, `nodes` returns the list of nodes and does _not_ take the flag `--name` flag.

clyang82: Here is my output:
```
chunlins-mbp:envoy clyang$ kind get nodes --name clusterapi
clusterapi-control-plane
chunlins-mbp:envoy clyang$ kind get nodes
Error: unknown cluster "kind"
chunlins-mbp:envoy clyang$
```
from your explanation, the `kind get nodes` should list the nodes. Am I correct?

BenTheElder: it gets the nodes for the cluster filtered by --name, which defaults to `kind`. we could change that to not filter on name by default, but that would be a breaking change and it's not clear what the use case would be currently.

BenTheElder:
```
$ kind get nodes --help
lists existing kind nodes by their name

Usage:
  kind get nodes [flags]

Flags:
  -h, --help          help for nodes
      --name string   the cluster context name (default "kind")

Global Flags:
      --loglevel string   logrus log level [panic, fatal, error, warning, info, debug, trace] (default "warning")
```

the help text for `--name` should clarify that part. the text here should clarify what the command does, and _somewhere_ we need to explain what the output format is. this attempts to mention that it lists the nodes by their name, IE that it prints the names of the nodes. `--name` of the cluster is a different concern that I don't think should be in this text. it's not normal to mention flags here, they're in the generated `--help` output with their own text.

clyang82:
Chatted with @BenTheElder offline. the fix is not good enough. Let me close it. Thanks

PR #580, opened 2019-06-01
Add support for kubeadm v1beta2 config

fabriziopandini:
this PR adds support for kubeadm config v1beta2, introduced in v1.15

v1beta2 config introduces only smalls improvements. 
Some of them might be of interest like e.g. the possibility to pass the ignore preflight error settings or support for passing a well-known certificate key when activating the automatic certificate copy feature.

As of today, kind is not using those features.


BenTheElder:
Just to clarify:
- there are no behavioral changes between these API versions?
- this will only take effect when 1.15 is tagged / released (soon!)

neolit123:
> there are no behavioral changes between these API versions?

"A list of changes since v1beta1"
https://godoc.org/k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm/v1beta2

> this will only take effect when 1.15 is tagged / released (soon!)

the change is already viable in terms of `release-1.15` and `master` of k/k, however given kind is decoupled, it's a question whether kind's `master` supports these new upstream changes.

BenTheElder:
> the change is already viable in terms of release-1.15 and master of k/k, however given kind is decoupled, it's a question whether kind's master supports these new upstream changes.

er it won't take effect because it checks <= 1.15, and pre-release versions should show as less

neolit123:
> er it won't take effect because it checks <= 1.15, and pre-release versions should show as less

that's partially what i meant.

>the change is already viable in terms of release-1.15 and master of k/k, however given kind is decoupled, it's a question whether kind's master supports these new upstream changes.

what i meant is that kind waits for a release to enable the latest kubeadm API.

currently kind favors a config that "works", and kubeadm does transparently support the previous version of the config (v1beta1) in 1.15 and it's pre-releases, however kubeadm wants to test v1beta2 in e2e pre-releases too.

`ver.LessThan` doesn't seem like a good option for that, so the question rolls back to:
> given kind is decoupled, it's a question whether kind's master supports these new upstream changes.

fabriziopandini:
@BenTheElder 
@neolit123 already provided many details; if there is still some doubt, I will happily follow up

BenTheElder:
sorry none at all, just clarifying, was a bit deep in some other bugs, let's get this in :-)

PR #581, opened 2019-06-02
use 'command -v' instead of 'which' in bash scripts

zegl:
command -v is a shell builtin, this change is recommended by shellcheck.

This updates #528

### DIFF
--- a/hack/release/create.sh
+++ b/hack/release/create.sh
@@ -31,7 +31,7 @@ cd "${REPO_ROOT}"
 
 # darwin is great
 SED="sed"
-if which gsed &>/dev/null; then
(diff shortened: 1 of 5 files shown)


BenTheElder:
Thanks for the PR, but actually these are not equivalent, https://github.com/koalaman/shellcheck/issues/1162

We are in fact looking for the gnu-sed binary specifically, `command -v` matches against aliases etc which we do _not_ want. In the Kubernetes main repo we've disabled SC2230.

`which` is available on pretty much any modern posix-y system.

zegl:
It seems like you're right, I'll close this PR. :)

PR #582, opened 2019-06-02
release: use correct exit code

zegl:
This bug was found with shellcheck.

This updates #528

### DIFF
--- a/hack/release/create.sh
+++ b/hack/release/create.sh
@@ -23,7 +23,7 @@ set -o pipefail
 
 if [ "$#" -ne 2 ]; then
     echo "Usage: create.sh release-version next-prerelease-version"
-    exit -1
+    exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


hack/release/create.sh
```diff
@@ -23,7 +23,7 @@ set -o pipefail
 
 if [ "$#" -ne 2 ]; then
     echo "Usage: create.sh release-version next-prerelease-version"
-    exit -1
+    exit 1
```
aojea: @BenTheElder  I think that you are the expert here :)

BenTheElder: _technically_ 255 would be the equivalent code on many platforms (the spec only mentions 8 bits and signed -> unsigned -1 -> 255), but in this case we're just setting any non-zero exit code so 1 SGTM :-)

PR #583, opened 2019-06-02
Add a preflight check that asserts docker connectivity

### LINKED ISSUE #554: Improved error message when Docker is not running
**What happened**:

If you run `kind create cluster` when no Docker daemon is running, kind exits with a unclear message:

```
Error: could not list clusters: failed to list nodes: exit status 1
```

A clearer error message, making the experience nicer for the user.

Since the errors are wrapped, and `docker` is executed fairly deep in the codebase, I suggest the following error message:

```
Error: could not list clusters: failed to list nodes, docker might be unavailable: exit status 1
```

1. Don't run a docker daemon
2. Run `kind create cluster` 

**Environment:**

- kind version: (use `kind version`): v0.4.0-alpha (bed5b7ef6c8ceb69606bfa656bd772e0834ff428)

---

I'll submit a PR with a fix for this, if you think that this is a good idea.

zegl:
Assert that the docker daemon is running, and that we're able to connect to it.
This is done by executing 'docker ps', and checking the exit code.

#39
#554

### DIFF
--- a/pkg/preflight/runtime/docker.go
+++ b/pkg/preflight/runtime/docker.go
@@ -14,7 +14,7 @@ func Preflight() error {
 
 	for _, check := range checks {
 		if err := check(); err != nil {
-			return errors.Wrap(err, "Preflight check failed")
+			return errors.Wrap(err, "preflight check failed")
 		}
 	}
 
@@ -25,7 +25,7 @@ func Preflight() error {
 func dockerIsRunning() error {
 	err := exec.Command("docker", "ps").Run()
 	if err != nil {
-		return errors.Wrap(err, "Could not connect to a docker daemon")
+		return errors.Wrap(err, "could not connect to a Docker daemon")
 	}
 	return nil
 }
--- a/pkg/preflight/runtime/docker.go
+++ b/pkg/preflight/runtime/docker.go
@@ -1,0 +1,31 @@
+package runtime
+
+import (
+	"github.com/pkg/errors"
+	"os/exec"
+)
+
+// Preflight runs checks to make sure that the container runtime
+// is working as expected
+func Preflight() error {
+	checks := []func() error{
+		dockerIsRunning,
+	}
+
+	for _, check := range checks {
+		if err := check(); err != nil {
+			return errors.Wrap(err, "Preflight check failed")
+		}
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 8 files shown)


pkg/preflight/runtime/docker.go
```diff
@@ -0,0 +1,31 @@
...
+		if err := check(); err != nil {
+			return errors.Wrap(err, "Preflight check failed")
+		}
+	}
+
+	return nil
+}
+
+// dockerIsRunning asserts that the docker daemon is running and is responsive
+func dockerIsRunning() error {
```
aojea: I think this should go to `pkg/container/docker`

pkg/preflight/runtime/docker.go
```diff
@@ -0,0 +1,31 @@
...
+// Preflight runs checks to make sure that the container runtime
+// is working as expected
+func Preflight() error {
+	checks := []func() error{
+		dockerIsRunning,
+	}
+
+	for _, check := range checks {
+		if err := check(); err != nil {
+			return errors.Wrap(err, "Preflight check failed")
```
neolit123: error strings should start with lowercase.
[1]

pkg/preflight/runtime/docker.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+
+	return nil
+}
+
+// dockerIsRunning asserts that the docker daemon is running and is responsive
+func dockerIsRunning() error {
+	err := exec.Command("docker", "ps").Run()
+	if err != nil {
+		return errors.Wrap(err, "Could not connect to a docker daemon")
```
neolit123: [1]
`docker` should be upper case.

pkg/preflight/runtime/docker.go
zegl: Thanks, I've fixed it. :)

BenTheElder:
Hi, thanks again for the PR!

If you check the current code, you'll see we actually already do a filtered `docker ps` to list the existing clusters, what we really need is to surface this error better instead of making another call.

I'm going to discuss at today's meeting (in 9 minutes!) and update the preflight checks issue as well to go from pre-flight to post-failure, which I think will help avoid piling on lots of slow checks (not that this one is, but presumably we'd add more) in the normal case, and instead diagnose the environment on failure.

BenTheElder:
going another route, thanks

PR #584, opened 2019-06-03
Add shellcheck verification

aojea:
Based on @BenTheElder`s kubernetes script
 https://github.com/kubernetes/kubernetes/blob/master/hack/verify-shellcheck.sh

Partially implements https://github.com/kubernetes-sigs/kind/issues/528


aojea:
seems that the `pull-kind-verify` job doesn't have access to the docker daemon and fails to run the `shellcheck`container. What's the best way moving forward? modify the job to use docker or install `shellcheck`?

neolit123:
> seems that the pull-kind-verify job doesn't have access to the docker daemon and fails to run the shellcheckcontainer. What's the best way moving forward? modify the job to use docker or install shellcheck?

i'd just run the shellcheck Linux-AMD64 official binary download.

aojea:
it can be merged after #581 and #582

hack/verify-shellcheck.sh
```diff
@@ -0,0 +1,233 @@
...
+    echo 'If the above warnings do not make sense, you can exempt this package from shellcheck'
+    echo 'checking by adding it to hack/.shellcheck_failures (if your reviewer is okay with it).'
+    echo
+  } >&2
+  exit 1
+fi
+
+if [[ ${#not_failing[@]} -gt 0 ]]; then
+  {
+    echo "Some packages in hack/.shellcheck_failures are passing shellcheck. Please remove them."
```
BenTheElder: heh we should really update this to say "files" instead of packages in both here and kubernetes

aojea: https://github.com/kubernetes/kubernetes/pull/78666

BenTheElder:
re: the binaries

yep, sorting out what we wanted to do on this was the big blocker for introducing it to kubernetes, this script is already ~optimal in that regard.

I'd like to in the future make this script very, very simple though. the other thing we should do here is not write large amounts of bash or have any failing files. as soon as we get it to zero, I will remove all of the logic related to known failures

Issue #576, opened 2019-05-31
Performance regression in cluster create


amwat:
**What happened**:
Cluster create took ~50 seconds when using image built from k/k master .

```
real	0m49.629s
user	0m0.579s
sys	0m0.399s
```

**What you expected to happen**:
Cluster create should take around ~30 seconds
which it does when using the kindest/node:v1.14.1 image
```
real	0m31.548s
user	0m0.567s
sys	0m0.363s

```
**How to reproduce it (as minimally and precisely as possible)**:
At kubernetes/kubernetes master (currently https://github.com/kubernetes/kubernetes/commit/990695c839a99cbd0e00fd3806f9802e2af7c2e0)

```
kind build node-image --type=bazel
time kind create cluster --image=kindest/node:latest 
```

**Anything else we need to know?**:
This probably is a regression in k/k and not kind itself
Here are the kubelet logs: https://gist.github.com/amwat/41323bbf70f01cf958d70d66b6f6be75
which show a bunch of fatals at the start which weren't there in 1.14.

**Environment:**
- kind version: (use `kind version`): 0.3.0-alpha
- Kubernetes version: (use `kubectl version`): master
- Docker version: (use `docker info`): 18.09.3
- OS (e.g. from `/etc/os-release`): Debian GNU/Linux

BenTheElder:
cc @neolit123 :eyes:

amwat:
Got similar results on kind HEAD (v0.4.0-alpha).
The kubelet fatals only take up about 2 seconds though so there's also something more at play.

aojea:
I think that this is the change that introduced the kubelet `fatals` https://github.com/kubernetes/kubernetes/commit/7cbe2d6c5f4c3b00d482dc7de3bfd1831ff0e861

I guess that the problem is that now systemd waits until the timeout of 10s expires to restart the kubelet

neolit123:
looks like something that should be discussed with sig node.

BenTheElder:
Got distracted, coming back to this now. Seems that startup time is now consistently at least slightly longer than 1.14.2, but also less consistent and sometimes quite a bit longer. Not sure what the root cause is yet.

BenTheElder:
https://github.com/kubernetes/kubernetes/blob/ba8fcafaf8c502a454acd86b728c857932555315/build/debs/kubelet.service#L9

[...]

PR #585, opened 2019-06-03
use our own systemd units and restart kubelet more often

BenTheElder:
xref: #576 

this can shave around 10 seconds off of startup with v1.14.2

for v1.15+ we still need to track down issues in kubernetes, but this does still generally make things faster, sub 30s startup for a single node is now pretty normal with this patch

we'll want to get new images with this at some point, I'll probably wait for more of the kindnetd work to land though


pkg/build/node/systemd.go
```diff
@@ -0,0 +1,49 @@
...
+Description=kubelet: The Kubernetes Node Agent
+Documentation=http://kubernetes.io/docs/
+
+[Service]
+ExecStart=/usr/bin/kubelet
+Restart=always
+StartLimitInterval=0
+# NOTE: kind deviates from upstream here with a lower RestartSec
+# 100ms is actually the systemd default, we'll use that for now
+RestartSec=100ms
```
BenTheElder: https://www.freedesktop.org/software/systemd/man/systemd.service.html

BenTheElder:
... these changes override what the existing builds do so I updated the PR to:
- drop the `apt` "build" (we'd been talking about this for a while anyhow, and it doesn't fit with this)
- stop handling systemd units and related files in the "bits" implementations

aojea:
> ... these changes override what the existing builds do so I updated the PR to:
> 
> * drop the `apt` "build" (we'd been talking about this for a while anyhow, and it doesn't fit with this)
> * stop handling systemd units and related files in the "bits" implementations

IIRC apt build was used by some specific architecture, I think that we should provide the tar option ASAP for people that don't want/need to compile k8s from source

tao12345666333:
> I think that we should provide the tar option ASAP for people that don't want/need to compile k8s from source

+1

aojea:
heheh, some of the tests run now in 18.30 min, usually they were running in about 20 mins

BenTheElder:
> IIRC apt build was used by some specific architecture, I think that we should provide the tar option ASAP for people that don't want/need to compile k8s from source

we already provide images from all of the current releases, the apt build literally only allowed installing the current release. there's on reason to build that

we should however provide the tar option, but that's somewhere in the mix with all the other changes we need to get in :sweat_smile: 

> heheh, some of the tests run now in 18.30 min, usually they were running in about 20 mins

we'll see how stable that is, if you look it's affected quite a bit by other things (like, prow should not be seeing heavy I/O throttling, which was causing a lot of flakes, because Katharine and I allocated bigger disks in the build cluster :-))

PR #596, opened 2019-06-06
Minor tooling improvements

BenTheElder:
- updated staticcheck to latest
- moved misspell to be tracked in the tools module instead of temporarily cloned, simplified this script as a result
- eliminated paths in shellcheck script that don't use pre-installed binary or docker (developing kind requires docker, using the host binary if the version matches is a small optimization)
- dramatically simplified shellcheck script now that we don't need to support tracking known failures
- only support using docker to shellcheck


BenTheElder:
/retest
[found a bug in the job config]

BenTheElder:
/retest
[more job config bugs]

Issue #588, opened 2019-06-04
create HA cluster is flaky


neolit123:
**What happened**:
i started seeing odd failures in the kind-master and -1.14 kubeadm jobs:
https://k8s-testgrid.appspot.com/sig-cluster-lifecycle-kubeadm#kubeadm-kind-master
https://k8s-testgrid.appspot.com/sig-cluster-lifecycle-kubeadm#kubeadm-kind-1.14

after switching to this HA config:
```
# a cluster with 3 control-planes and 3 workers
kind: Cluster
apiVersion: kind.sigs.k8s.io/v1alpha3
nodes:
- role: control-plane
- role: control-plane
- role: control-plane
- role: worker
- role: worker
- role: worker
```

```
I0604 19:15:09.075770     760 join.go:480] [preflight] Retrieving KubeConfig objects
[preflight] Reading configuration from the cluster...
[preflight] FYI: You can look at this config file with 'kubectl -n kube-system get cm kubeadm-config -oyaml'
I0604 19:15:10.310249     760 round_trippers.go:438] GET https://172.17.0.2:6443/api/v1/namespaces/kube-system/configmaps/kubeadm-config 401 Unauthorized in 1233 milliseconds
error execution phase preflight: unable to fetch the kubeadm-config ConfigMap: failed to get config map: Unauthorized 
 ✗ Joining more control-plane nodes 🎮
DEBU[22:15:10] Running: /usr/bin/docker [docker ps -q -a --no-trunc --filter label=io.k8s.sigs.kind.cluster --format {{.Names}}\t{{.Label "io.k8s.sigs.kind.cluster"}} --filter label=io.k8s.sigs.kind.cluster=kind] 
$KUBECONFIG is still set to use /home/lubo-it/.kube/kind-config-kind even though that file has been deleted, remember to unset it
DEBU[22:15:10] Running: /usr/bin/docker [docker rm -f -v kind-control-plane2 kind-control-plane kind-control-plane3 kind-worker kind-worker3 kind-worker2 kind-external-load-balancer] 
⠈⠁ Joining more control-plane nodes 🎮 Error: failed to create cluster: failed to join a control plane node with kubeadm: exit status 1
```

**What you expected to happen**:
no errors.

**How to reproduce it (as minimally and precisely as possible)**:
```
cd kind-src-path
GO111MODULE=on go build
# install the kind binary to PATH
cd kubernetes-src-path
kind build node-image --kube-root=$(pwd)
[...]

PR #598, opened 2019-06-07
Wait for control plane node to be ready after joining the cluster

aojea:
It can happen that the control plane node is not completely ready after joining the cluster,
If one worker node tries to join against a control plane node that's not ready, it fails to join thus the cluster creation fails.
This is a workaround to wait until the control node is ready after it joins the cluster before joining new nodes.

Fixes: https://github.com/kubernetes-sigs/kind/issues/588

### DIFF
--- a/pkg/cluster/internal/create/actions/kubeadmjoin/join.go
+++ b/pkg/cluster/internal/create/actions/kubeadmjoin/join.go
@@ -23,6 +23,7 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
@@ -209,6 +210,16 @@ func runKubeadmJoinControlPlane(
 	log.Debug(strings.Join(lines, "\n"))
 	if err != nil {
 		return errors.Wrap(err, "failed to join a control plane node with kubeadm")
+	}
+
+	// Wait for the node to be Ready
+	// TODO: remove once https://github.com/kubernetes-sigs/kind/issues/588 is fixed
+	// kubeadm join should guarantee that the cluster is ready
+	startTime := time.Now()
+	// TODO: currently hardcoded to 10s
+	isReady := nodes.WaitForReady(node, startTime.Add(10*time.Second))
+	if !isReady {
+		return errors.Errorf("Timed out waiting for control plane node %s to be ready after join", node.String())
 	}
 
 	return nil


pkg/cluster/internal/create/actions/kubeadmjoin/join.go
```diff
@@ -211,6 +212,16 @@ func runKubeadmJoinControlPlane(
 		return errors.Wrap(err, "failed to join a control plane node with kubeadm")
 	}
 
+	// Wait for the node to be Ready
+	// TODO: remove once https://github.com/kubernetes-sigs/kind/issues/588 is fixed
+	// kubeadm join should guarantee that the cluster is ready
```
aojea: @fabriziopandini @neolit123 I don't know if this is true 😅 , should `kubeadm join`, guarantee that?

pkg/cluster/internal/create/actions/kubeadmjoin/join.go
```diff
@@ -211,6 +212,16 @@ func runKubeadmJoinControlPlane(
...
+	startTime := time.Now()
+	// TODO: currently hardcoded to 10s
+	isReady := nodes.WaitForReady(node, startTime.Add(10*time.Second))
```
thomastaylor312: Super new to the code so I could be completely off, but isn't there a [`WaitForReady`](https://godoc.org/sigs.k8s.io/kind/pkg/cluster/create#WaitForReady)  option already that could be used here?

aojea: yeah, indeed this function `nodes.WaitForReady` is used by the one you mentioned.
In this case we are looking for a specific node so I think that's better to use this one,

pkg/cluster/internal/create/actions/kubeadmjoin/join.go
fabriziopandini: @aojea see my comments on the issue https://github.com/kubernetes-sigs/kind/issues/588#issuecomment-499939113

In a nutshell Kubeadm is not responsible, but it is the kubelet. Additional, it seems also that the API server does not detected properly when the etcd instance is ready

neolit123: > kubeadm join should guarantee that the cluster is ready

well kubeadm can look at pod and endpoint state, but the cluster as a whole - a bit tricky.
kinder currently waits for these pods + the node ready status, as @fabriziopandini  mentioned:
https://github.com/kubernetes/kubeadm/blob/62556834c87e34004ac84c17b2f2c68b5c4f3b22/kinder/pkg/actions/waiter.go#L32-L44

given HA join consistently does not fail using kind 0.2.0 as seen here:
https://k8s-testgrid.appspot.com/sig-cluster-lifecycle-kubeadm#kubeadm-kind-master
i'm trying to get to the bottom of the problem instead - i.e. finding a change in kind that helped the problem surface.

aojea: kind is much faster with the latest versions, the fact that adding delays solve the problem or at least reduce them makes me think that´s tightly related to that

BenTheElder:
Code looks fine FWIW but I'm not convinced we've fully root-caused the issue we're attempting to address yet, and I'm not a fan of adding arbitrary busy waits.

aojea:
Definitely not the right approach

PR #601, opened 2019-06-11
Cleanup hack scripts.

amwat:
Move hack and verify scripts under respective subdirectories.

/cc @BenTheElder


hack/verify-all.sh
```diff
@@ -34,37 +34,37 @@ fi
```
amwat: Didn't move verify-all.sh yet since the prow presubmit still points to this.

hack/update/update-all.sh
```diff
@@ -20,6 +20,6 @@ set -o pipefail
 
 REPO_ROOT=$(git rev-parse --show-toplevel)
 
-"${REPO_ROOT}"/hack/update-deps.sh
-"${REPO_ROOT}"/hack/update-generated.sh
-"${REPO_ROOT}"/hack/update-gofmt.sh
+"${REPO_ROOT}"/hack/update/update-deps.sh
```
BenTheElder: maybe we should drop the `update-` 🙃

hack/verify-all.sh
BenTheElder: we can make this alias the actual script and then update the job, come back and remove the alias

Issue #109, opened 2018-11-14
consider using klog for logging


BenTheElder:
I think https://github.com/kubernetes/klog is the way forward for the ecosystem, and after some experimenting I think structured logging makes the most sense for server / syslog type logs, not debugging a CLI.

I don't love glog/klogs's line formatting, but it seems like the way forward to integrate in the ecosystem.

cc @munnerz

BenTheElder:
@alejandrox1 FYI you should not reference issues from commits, because github will reference them across forks. so when someone merges your PR to another fork github will try to close it here :grimacing: 

The UX for those references is a bit wonky and has been a problem across the project.

alejandrox1:
@BenTheElder :+1: won't happen again. 
Should I amend the commit on #115 ?

BenTheElder:
That would be great 🙂

Also don't worry about it though, just a note. We have a few issues in k8s that get spammed & closed with this very badly 😅

BenTheElder:
https://github.com/kubernetes/klog/tree/master/klogr needs investigating

neolit123:
i'm mostly -1 on structured logging for kind at this point.
due to the scope of the project, i doubt it will face the need for structured logging.

also, the current logr usage in e.g. C-API seems a bit underwhelming and i don't see obvious benefits.

BenTheElder:
we can use it without stuctured logging
mainly my concern is improving the library use case. I have also been thinking about your point about logging to a file though... :thinking:

neolit123:
> we can use it without stuctured logging
mainly my concern is improving the library use case

but there might be no benefits from using logr if we are not using it with structures.

> I have also been thinking about your point about logging to a file though

i was talking about Ilya from Weave about this recently in a PR review as k/k has a similar problem in a lot of places.

front end and back end should have a single point of connection in terms of logging.
this is the point where the front end decides to which streams the backend is going to write to.

we can consider the consumers of the kind library to be a "frontend".

also the go stdlib has cool features like:
https://golang.org/pkg/io/#MultiWriter
which allows you to pipe to a standard stream and a file at the same time, so consumers can do that.

BenTheElder:
realistically though they will want control over EG how much logging, /shrug
most ecosystem projects are klog/glog including kubeadm, cluster-api, etc. right?

The benefit is seamlessly integrating with their stacks. we don't depend heavily on logging anyhow.

neolit123:
> realistically though they will want control over EG how much logging

[...]

PR #603, opened 2019-06-11
refactor kindnetd

BenTheElder:
this breaks things up a little bit and does some misc cleanup

future steps:
- refactor to actually be a proper controller (not super important right now, but a better plan long term...)
- tag routes with kindnetd as the source so we can cleanup after deleted nodes (low priority, but we should consider this future direction for robustness when developing it...)


pkg/build/node/cni.go
```diff
@@ -185,21 +186,23 @@ spec:
...
-          - name: POD_IP
-            valueFrom:
-              fieldRef:
-                fieldPath: status.podIP
-          - name: CNI_CONFIG_TEMPLATE
-            valueFrom:
-              configMapKeyRef:
-                name: kindnet-cfg
-                key: cni-conf-template.json
+        - name: HOST_IP
```
aojea: is this well indented? seem indentation has changed

BenTheElder: this normalized the indentation, see EG the volumeMounts below

BenTheElder:
next up going to refactor the build, iterating on building the image is very slow right now, but should be easy to fix 😅

cmd/kindnetd/main.go
```diff
@@ -94,29 +62,28 @@ func main() {
...
+	// enforce ip masquerade rules
 	ipv6 := false
+	// TODO: dual stack...?
 	if strings.Contains(hostIP, ":") {
 		ipv6 = true
 	}
-	// used to create routes and iptables rules
-	kindnetd := NewKindnet(cniConfigWriter, ipv6, kindnetConfig)
+	masqAgent, _ := NewIPMasqAgent(ipv6, []string{os.Getenv("POD_SUBNET")})
+	go masqAgent.SyncRulesForever(time.Second * 60)
```
aojea: what happens if there is an error with the masqAgent?

BenTheElder: right now it won't handle it. we could panic or log an error.

I think in this case we should panic just until a (near) future refactor adds proper logging

BenTheElder: updated to panic for now with a TODO to log pervasively.

xref: #109 I think it's about time to consider merging with the ecosystem on that front...

BenTheElder: thanks for catching this.

cmd/kindnetd/masq.go
```diff
@@ -0,0 +1,142 @@
...
+	for {
+		if err := ma.SyncRules(); err != nil {
+			return err
+		}
+		time.Sleep(interval)
+	}
+}
+
+// name of nat chain for iptables masquerade rules
+const masqChainName = "KIND-MASQ-AGENT"
```
BenTheElder: this is now fixed instead of a flag (the flag came from the original masq agent)

cmd/kindnetd/masq.go
```diff
@@ -0,0 +1,142 @@
...
+		noMasqueradeCIDRs: noMasqueradeCIDRs,
+	}, nil
+}
+
+// IPMasqAgent is based on https://github.com/kubernetes-incubator/ip-masq-agent
+// but collapsed into kindnetd and made ipv6 aware in an opionated and simplified
+// fashion
+type IPMasqAgent struct {
+	iptables          utiliptables.Interface
+	masqChain         utiliptables.Chain
```
BenTheElder: this chain is now in the agent instead of a global

aojea:
outstanding 🥇

PR #606, opened 2019-06-12
Concurrency: Timing out, moving on

aojea:
To avoid situations that the command `export logs` hangs waiting for the `tar` command to exits, 
we time out on the concurrent functions to be safe.
It also adds a debug message to the untar function and runs it concurrently to benefit of the new timeout protection (Note: the logging is not ordered 😅  )

### DIFF
--- a/pkg/cluster/logs/logs.go
+++ b/pkg/cluster/logs/logs.go
@@ -54,6 +54,22 @@ func Collect(nodes []nodes.Node, dir string) error {
 			return execToPath(cmd, path)
 		}
 	}
+	// helper to run a cmd and untar the output to a path
+	execUntarToPath := func(cmd exec.Cmd, path string) error {
+		return exec.RunWithStdoutReader(cmd, func(outReader io.Reader) error {
+			if err := untar(outReader, path); err != nil {
+				return errors.Wrapf(err, "Untarring %q: %v", path, err)
+			}
+			return nil
+		})
+
+	}
+	execUntarToPathFn := func(cmd exec.Cmd, path string) func() error {
+		return func() error {
+			return execUntarToPath(cmd, path)
+		}
+	}
+
 	// construct a slice of methods to collect logs
 	fns := []func() error{
 		// TODO(bentheelder): record the kind version here as well
@@ -67,16 +83,6 @@ func Collect(nodes []nodes.Node, dir string) error {
 	for _, n := range nodes {
 		node := n // https://golang.org/doc/faq#closures_and_goroutines
 		name := node.String()
-		// grab all logs under /var/log (pods and containers)
-		cmd := node.Command("tar", "--hard-dereference", "-C", "/var/log", "-chf", "-", ".")
-
-		exec.RunWithStdoutReader(cmd, func(outReader io.Reader) error {
-			if err := untar(outReader, filepath.Join(dir, name)); err != nil {
-				return errors.Wrapf(err, "Untarring %q: %v", name, err)
-			}
-			return nil
-		})
-
 		fns = append(fns, func() error {
 			return concurrent.Coalesce(
 				// record info about the node container
@@ -105,6 +111,10 @@ func Collect(nodes []nodes.Node, dir string) error {
 					node.Command("journalctl", "--no-pager", "-u", "containerd.service"),
 					filepath.Join(name, "containerd.log"),
 				),
+				execUntarToPathFn(
+					node.Command("tar", "--hard-dereference", "-C", "/var/log", "-chf", "-", "."),
+					prefixedPath(name),
+				),
 			)
 		})
 	}
@@ -130,6 +140,7 @@ func untar(r io.Reader, dir string) (err error) {
 
 		rel := filepath.FromSlash(f.Name)
 		abs := filepath.Join(dir, rel)
+		log.Debugf("Untarring file %v to %v", f.Name, dir)
 
 		switch f.Typeflag {
 		case tar.TypeReg:
--- a/pkg/concurrent/concurrent.go
+++ b/pkg/concurrent/concurrent.go
@@ -17,10 +17,14 @@ limitations under the License.
 package concurrent
 
 import (
+	"errors"
 	"sync"
+	"time"
 
 	"sigs.k8s.io/kind/pkg/util"
 )
+
+const timeout = 60 * time.Second
 
 // UntilError runs all funcs in separate goroutines, returning the
 // first non-nil error returned from funcs, or nil if all funcs return nil
@@ -33,8 +37,13 @@ func UntilError(funcs []func() error) error {
 		}()
 	}
 	for i := 0; i < len(funcs); i++ {
-		if err := <-errCh; err != nil {
-			return err
+		select {
+		case err := <-errCh:
+			if err != nil {
+				return err
+			}
+		case <-time.After(timeout):
+			return errors.New("UntilError: Timeout waiting for concurrent function")
 		}
 	}
 	return nil
@@ -52,7 +61,9 @@ func Coalesce(fns ...func() error) error {
 			ch <- f()
 		}(fn)
 	}
-	wg.Wait()
+	if waitTimeout(&wg, timeout) {
+		return errors.New("Coalesce: Timeout waiting for wait group")
+	}
 	close(ch)
 	// collect up and return errors
 	errs := []error{}
@@ -67,4 +78,20 @@ func Coalesce(fns ...func() error) error {
 		return errs[0]
 	}
 	return nil
+}
+
+// waitTimeout waits for the waitgroup for the specified max timeout.
+// Returns true if waiting timed out.
+func waitTimeout(wg *sync.WaitGroup, timeout time.Duration) bool {
+	c := make(chan struct{})
+	go func() {
+		defer close(c)
+		wg.Wait()
+	}()
+	select {
+	case <-c:
+		return false // completed normally
+	case <-time.After(timeout):
+		return true // timed out
+	}
 }


aojea:
/assign @neolit123 
/assign @BenTheElder 

It doesn't solve everything but I think it will not hang forever waiting for the tar command

pkg/cluster/logs/logs.go
```diff
@@ -105,6 +111,10 @@ func Collect(nodes []nodes.Node, dir string) error {
 					node.Command("journalctl", "--no-pager", "-u", "containerd.service"),
 					filepath.Join(name, "containerd.log"),
 				),
+				execUntarToPathFn(
+					node.Command("tar", "--hard-dereference", "-C", "/var/log", "-chf", "-", "."),
```
neolit123: is it possible to see output of which file is being tarred? i saw some examples online but it didn't seem a matter of adding a tar flag only.

aojea: adding a -v to the tar command should output that info, but IIRC all the output is redirected to the reader so it can untar it, also I think that there were problems if we allocate a tty to run the command, 
```
DEBU[15:02:00] Running: /usr/bin/docker [docker exec --privileged kind-control-plane tar --hard-dereference -C /var/log -zchf - .]
```
Ben has a much better understanding, it would be great to know his opinion

neolit123: ok, just wanted to point out that there are way to many reports on the web of the tar command hanging for a variety of reasons e.g. cannot read file.

aojea:
Profile of current `export logs` 
[profile001.pdf](https://github.com/kubernetes-sigs/kind/files/3281473/profile001.pdf)

pkg/cluster/logs/logs.go
aojea: that´s why I moved it to their own thread, ... should we add a timeout to the export logs and close if exceeded?

neolit123:
```
     --HANG
           hang for SECS seconds (default 3600)
```
not sure if it the actual timeout flag or something else.
http://linuxcommand.org/lc3_man_pages/tar1.html

also not sure if all versions have it.

pkg/exec/exec.go
```diff
@@ -101,11 +101,11 @@ func RunWithStdoutReader(cmd Cmd, readerFunc func(io.Reader) error) error {
 		return err
 	}
 	defer pw.Close()
-	defer pr.Close()
 	cmd.SetStdout(pw)
 
 	errChan := make(chan error, 1)
 	go func() {
+		defer pr.Close()
```
aojea: no luck with this :(

aojea:
This is the stack trace of a hanging process
[stack.txt](https://github.com/kubernetes-sigs/kind/files/3281840/stack.txt)

pkg/cluster/logs/logs.go
aojea: from the stack trace that I´ve uploaded in another comment and looking at one hanging process the tar command seems to exit, can it be possible that is not closing the pipe?

pkg/exec/exec.go
```diff
@@ -101,19 +101,19 @@ func RunWithStdoutReader(cmd Cmd, readerFunc func(io.Reader) error) error {
...
 		errChan <- readerFunc(pr)
 	}()
 
 	err = cmd.Run()
-	err2 := <-errChan
 	if err != nil {
 		return err
 	}
+	err2 := <-errChan
```
aojea: @neolit123 @BenTheElder I think that this was the deadlock, if the tar command failed it was waiting for the readerFunc to finish

pkg/exec/exec.go
aojea: it should be here or just in the previous place?

pkg/exec/exec.go
neolit123: i would leave it to Ben to comment on the concurrency of exec.

BenTheElder: The reader func should not hang indefinitely, we can fix that instead.

aojea: I thought that the reader func only finish when the pipe is closed

BenTheElder: er actually right, the pipe will be closed and this will exit with just _this_ change. the other changes are unnecessary

pkg/cluster/logs/logs.go
```diff
@@ -113,9 +125,13 @@ func Collect(nodes []nodes.Node, dir string) error {
 	return concurrent.Coalesce(fns...)
 }
 
-// untar reads the tar file from r and writes it into dir.
-func untar(r io.Reader, dir string) (err error) {
-	tr := tar.NewReader(r)
+// untargz reads the tar compressed file from r and writes it into dir.
+func untargz(r io.Reader, dir string) (err error) {
```
BenTheElder: is this actually more performant? splitting this PR up

pkg/exec/exec.go
aojea: but the other changes are cool 😎 , aren´t they?

pkg/cluster/logs/logs.go
aojea: it reduces the amount of bytes going through the pipe, the performance is the same but I think we are less exposed to the docker exec thing

pkg/exec/exec.go
```diff
@@ -100,20 +100,20 @@ func RunWithStdoutReader(cmd Cmd, readerFunc func(io.Reader) error) error {
 	if err != nil {
 		return err
 	}
-	defer pw.Close()
-	defer pr.Close()
 	cmd.SetStdout(pw)
+	defer pr.Close()
+	defer pw.Close()
```
BenTheElder: this change is bad, we should defer these immediately after checking the creation, if the call above panics for some reason and then the caller recovers, we won't close these.

pkg/exec/exec.go
BenTheElder: maybe? https://github.com/kubernetes-sigs/kind/pull/606/files#r293113238 should not happen

pkg/cluster/logs/logs.go
BenTheElder: it does reduce the amount of bytes but I doubt the peformance is actually the same given we have to both compress and decompress, there's going to be a tradeoff in the resources used for that and the resources used to push bytes over the exec stream

pkg/exec/exec.go
aojea: just messed it up trying to recover the original,

aojea:
[targz.pprof.pdf](https://github.com/kubernetes-sigs/kind/files/3283188/targz.pprof.pdf)
[taronly.pprof.pdf](https://github.com/kubernetes-sigs/kind/files/3283211/taronly.pprof.pdf)

That shows the trade-off perfectly 😄

pkg/cluster/logs/logs.go
aojea: check last comments with a comparison of both

pkg/exec/exec.go
aojea: what do you, mean the compress one or the concurrent? I´ve split the changes  in different commits so we can get rid of all of one of them

BenTheElder:
I split out the immediate bug fix from the other changes after working on
this locally.

The PPROF of kind doesn't show the end to end picture, we have resource
usage in the node containers and in the docker daemon. The timings are also
interesting, and the size of the logs etc will change the numbers.

aojea:
keeping only the PR to untar concurrently

pkg/concurrent/concurrent.go
```diff
@@ -17,11 +17,15 @@ limitations under the License.
...
 
 import (
+	"errors"
 	"sync"
+	"time"
 
 	"sigs.k8s.io/kind/pkg/util"
 )
 
+const timeout = 60 * time.Second
```
aojea: Should we start with a higher value?

aojea:
@BenTheElder it's easy to hang `kind` exporting the logs if the `untar`  function hangs
```
diff --git a/pkg/cluster/logs/logs.go b/pkg/cluster/logs/logs.go
index 54029e2..884050f 100644
--- a/pkg/cluster/logs/logs.go
+++ b/pkg/cluster/logs/logs.go
@@ -21,6 +21,7 @@ import (
        "io"
        "os"
        "path/filepath"
+       "time"

        "github.com/pkg/errors"
        log "github.com/sirupsen/logrus"
@@ -125,6 +126,7 @@ func Collect(nodes []nodes.Node, dir string) error {

 // untar reads the tar file from r and writes it into dir.
 func untar(r io.Reader, dir string) (err error) {
+       time.Sleep(1000 * time.Second)
        tr := tar.NewReader(r)
        for {
                f, err := tr.Next()
```

I think that is safer adding timeouts to all the concurrent functions

pkg/concurrent/concurrent.go
neolit123: it will mask the problem and throw the error after a minute.
were we able to figure out why the hangs happen?

is `log.Debugf("Untarring file %v to %v", f.Name, dir)` consistently showing the same file?

aojea: 1. The problem will always be masked, I mean, the only difference with the current situation is that it won´t hang forever, only for the time we want it to hang. 
It will fail with an error saying that it timed out in one concurrent function and we will need to track it down using the logs like we are doing now. I think that´s a good safeguard at this moment, simple and effective. If we continue to hit this kind of problems I think we should start to use other options like `context`, but seems to overkill at this point, however, I´m not an expert in this topic.

2. It prints the name of the file it´s untarring, i.e.
```
DEBU[14:25:38] Untarring file ./pods/kube-system_kindnet-6244g_65fcc7c7-6eab-4024-8323-ec90df9d3582/kindnet-cni/0.log to /tmp/230018879/kind-control-plane
DEBU[14:25:38] Untarring file ./containers/busybox_default_busybox-e31c112c91afda6d8e51cb412100e2a555e97dfe6c4cd8d3550a052d6cea0c5c.log to /tmp/230018879/kind-worker2
DEBU[14:25:38] Untarring file ./containers/nginx-deployment-65bbbdfcd-v6m4b_default_nginx-a77eb9aa92d867c4bfee267ed2bcb822bee022f233420740097cde22412cd601.log to /tmp/230018879/kind-worker2
```

neolit123: > The problem will always be masked

a timeout in general is OK to have, but i'm missing context on the actual problem, is it related to mount points? why does it hang?

aojea: ups, sorry, in this case, [is about the `tar` problem, ](https://github.com/kubernetes-sigs/kind/pull/606#issuecomment-501664744 ) but don't know if it can happen with others functions that we execute concurrently.
 If they hang, the concurrency functions never time out

neolit123: i guess i was trying to understand why the tar operation actually hangs - e.g. is something about the `io.Reader` or about the logic in `archive/tar`.

technically `time.Sleep(1000 * time.Second)` will act as a delay there.

aojea: `time.Sleep(1000 * time.Second)` was an example that I've used to see the effects of the `untar`  function hanging.
If I understood correctly the current issue with tar was fixed  with #608 moving `err2 := <-errChan` after the `if err != nil {	`

```
	go func() {
		errChan <- readerFunc(pr)
	}()		
	err = cmd.Run()		
	err2 := <-errChan	
	if err != nil {		
		return err
	}		
``` 

I think that what happened is that the `cmd.Run()` failed but it never returns because is waiting for `err2 := <-errChan`, that;s waiting for the readerFunc()

pkg/cluster/logs/logs.go
aojea: removing gzip because is not adding any benefit

pkg/concurrent/concurrent.go
BenTheElder: this is known, but also doesn't occur in practical contexts currently.

timeouts are hard to get right. what if there are lots of logs?
ideally we want to detect actual hangs rather than blind timeouts.

aojea: yeah, eventually, the job will timeout too, so we should decide which timeout is more useful
With this approach we will have an error saying `Coalesce: Timeout waiting for wait group"` , we can set a long timeout, 5 mins per example to be safe, just something lower than the prow job timout

aojea:
There weren´t more issues related or at least they weren´t reported, let´s close and open if necessary

Issue #607, opened 2019-06-12
log exporting hangs if a node is not running


BenTheElder:
**What happened**: `kind export logs` hangs indefinitely when run against a cluster with a non-running node container

**What you expected to happen**: kind commands should not hang indefinitely

**How to reproduce it (as minimally and precisely as possible)**: `kind create cluster && docker stop kind-control-plane && kind export logs`

**Anything else we need to know?**: @aojea found the root cause already in RunWithStdoutReader, @joejulian reported the issue originally in the #kind slack channel.

**Environment:**

- kind version: (use `kind version`): 0.3.0
- Kubernetes version: (use `kubectl version`):
- Docker version: (use `docker info`):
- OS (e.g. from `/etc/os-release`):

/assign

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/606/files includes a fix but also many other changes, some unnecessary. sending a hotfix first.

joejulian:
:tada:

PR #608, opened 2019-06-12
prevent hang in log export

BenTheElder:
fixes https://github.com/kubernetes-sigs/kind/issues/607


BenTheElder:
NOTE: this bug was found / patched in #606, but along with some other changes that need evaluation, this change is trivially correct.

PR #609, opened 2019-06-12
small correctness fix

BenTheElder:
handle reader exiting early

we haven't observed this in our current usage, but this is more correct

### DIFF
--- a/pkg/exec/exec.go
+++ b/pkg/exec/exec.go
@@ -107,6 +107,7 @@ func RunWithStdoutReader(cmd Cmd, readerFunc func(io.Reader) error) error {
 	errChan := make(chan error, 1)
 	go func() {
 		errChan <- readerFunc(pr)
+		pr.Close()
 	}()
 
 	err = cmd.Run()


pkg/exec/exec.go
```diff
@@ -107,6 +107,7 @@ func RunWithStdoutReader(cmd Cmd, readerFunc func(io.Reader) error) error {
 	errChan := make(chan error, 1)
 	go func() {
 		errChan <- readerFunc(pr)
```
aojea: just learning with this, please bear with me, why do we have the `defer pr.Close` in L104? can we avoid having this new close by moving the defer just to L109?

BenTheElder: we want to close the reader either way, if we move line 104 here then we won't close it in the case we just fixed in #608 :-)

aojea: then ...  it can happen that we close twice the pr pipe, right? I mean, if we close it here the previous defer will try to close it again but is already closed IIUIC

aojea: nevermind, the function will return an err trying to write to a broken pipe

BenTheElder: Closing twice is safe 😅

PR #613, opened 2019-06-13
introduce logging shim

BenTheElder:
NOTE: this does not actually switch to using it yet, to keep the PR reviewable I'm merely introducing _how_ we can refactor the logging first. The next PR will switch everything over and introduce a compatibility layer in the flags.

To do this I also first refactored the logging / util code to better organize it.

Part of #176 #109 

The basic idea is:
- [PR 1] Have a logging interface (yes, our own) that is a subset of what klog provides. This **only** includes leveled info logging. Why?
  - Only CLI code / main should be able to Panic() / Fatal(), everything else should be returning errors instead. Library packages are banned from exiting the application. This interface excludes these, but the application will have set up the logger and can do this itself separately not via the interface.
   - Having a shim means users do not _have_ to use klog, if they're happy with klog they just need to import and call `sigs.k8s.io/kind/pkg/log/klog.Use()` somewhere, otherwise they can implement the little log interface and pass it to `sigs.k8s.io/kind/pkg/log.SetDefault()`. klog is standard in the kubernetes ecosystem and `V()` logging is nice, but we can be flexible.
  - User facing messages in CLI apps should happen via stdout / stderr with human readable messages, logging should be reserved for debugging. If this were a server application I'd use logrus with JSON logs everywhere and something like stackdriver to search them etc.. but it is **not** a server application and it's unreasonable to expect the output to go through a logging pipeline.
- [PR 2] Map `--loglevel` to `-v` levels temporarily and warn about deprecation `--loglevel` is set.
- [PR 2] Call `sigs.k8s.io/kind/pkg/log/klog.Use()` in the root kind command
- [PR 2] Use the default logger in `sigs.k8s.io/kind/pkg/log.V()` pervasively throughout the `pkg/` and `cmd/`  / replace logrus
- kindnetd isn't meant to be reused and can just use klog directly.

### DIFF
--- a/pkg/env/doc.go
+++ b/pkg/env/doc.go
@@ -1,0 +1,18 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package env contains helpers for environmental details
+package env
--- a/pkg/log/default.go
+++ b/pkg/log/default.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 The Kubernetes Authors.
+Copyright 2019 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -17,22 +17,22 @@ limitations under the License.
 package log
 
 import (
-	"strings"
-
-	"github.com/sirupsen/logrus"
+	"sync"
 )
 
-// LevelsString returns a string representing all log levels
-// this is useful for help text / flag info
-func LevelsString() string {
-	var b strings.Builder
-	b.WriteString("[")
-	for i, level := range logrus.AllLevels {
-		b.WriteString(level.String())
-		if i+1 != len(logrus.AllLevels) {
-			b.WriteString(", ")
-		}
-	}
-	b.WriteString("]")
-	return b.String()
+var std LeveledLogger
+var stdMu sync.Mutex
+
+// SetDefault sets the default logger
+func SetDefault(logger LeveledLogger) {
+	stdMu.Lock()
+	defer stdMu.Unlock()
+	std = logger
+}
+
+// V is leveled logging implemented against the default logger
+func V(level Level) Logger {
+	stdMu.Lock()
+	defer stdMu.Unlock()
+	return std.V(level)
 }
--- a/pkg/log/klog/klog.go
+++ b/pkg/log/klog/klog.go
@@ -1,0 +1,43 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package klog contains logging related functionality
+package klog
+
+import (
+	"k8s.io/klog"
+
+	"sigs.k8s.io/kind/pkg/log"
+)
+
+// Use installs this logger as the default
+func Use() {
+	log.SetDefault(New())
+}
+
+// New returns an opaque implementation of log.LeveledLogger against klog
+func New() log.LeveledLogger {
+	return &leveledLogger{}
+}
+
+// leveledLogger implements log.LeveledLogger against klog
+type leveledLogger struct{}
+
+var _ log.LeveledLogger = &leveledLogger{}
+
+func (l *leveledLogger) V(level log.Level) log.Logger {
+	return klog.V(klog.Level(level))
+}
--- a/pkg/log/logger.go
+++ b/pkg/log/logger.go
@@ -1,0 +1,37 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package log
+
+// Level is a leveled logging level, see https://github.com/kubernetes/klog
+type Level int32
+
+// LeveledLogger is a subset of what https://github.com/kubernetes/klog
+// provides that kind's library code may use. You can (and should) inject your
+// own default logger using SetDefault()
+//
+// See also klog.Use() in our klog package under this one
+type LeveledLogger interface {
+	V(Level) Logger
+}
+
+// Logger is an interface like klog.Verbose
+// see: https://github.com/kubernetes/klog
+type Logger interface {
+	Info(args ...interface{})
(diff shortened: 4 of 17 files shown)


pkg/env/doc.go
```diff
@@ -0,0 +1,18 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package env contains helpers for environmental details
+package env
```
BenTheElder: I needed a better place to put `IsTerminal` and found `pkg/util` which I intend to eventually eliminate (when we overhaul error handling some more, errors are the last thing in that package).

In the future I intend to put more environment detail sniffing in here which will be useful for a few other tasks.

BenTheElder: (we also moved the GOARCH reading code out of pkg/util here)

pkg/log/status/status.go
```diff
@@ -14,18 +14,17 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Package log contains logging related functionality
-package log
+// Package status
+package status
```
BenTheElder: this also didn't really belong in the main log package and was the bulk of this diff.

next up we will have to re-arrange how this interacts with the logger, have a few ideas for that, but it will be easier to work out once we start actually switching the logger...

pkg/log/default.go
```diff
@@ -0,0 +1,38 @@
...
+var stdMu sync.Mutex
+
+// SetDefault sets the default logger
+func SetDefault(logger LeveledLogger) {
+	stdMu.Lock()
+	defer stdMu.Unlock()
+	std = logger
+}
+
+// V is leveld logging implemented against the default logger
```
neolit123: `leveled`

pkg/log/klog/klog.go
```diff
@@ -0,0 +1,43 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package klog contains logging related functionality
+package klog
+
+import (
+	"k8s.io/klog"
```
neolit123: the old PR that i did made /pkg logger agnostic, leaving it to the front end to choose its logger.
are there plans to make /pkg to not import `"k8s.io/klog"`?

BenTheElder: this package should only be imported by `cmd/`, (that's buried in the wall of text in the OP)

BenTheElder: `pkg/` should use `pkg/log` functions against the private `pkg/log.std` which can be set with `pkg/log.SetDefault`

I think we will need to introduce a no-op default logger that is the actual default until `SetDefault` is called

pkg/log/default.go
BenTheElder: fixed, thanks

pkg/log/klog/klog.go
neolit123: > Having a shim means users do not have to use klog, if they're happy with klog they just need to import and call sigs.k8s.io/kind/pkg/log/klog.Use() somewhere, otherwise they can implement the little log interface and pass it to sigs.k8s.io/kind/pkg/log.SetDefault(). klog is standard in the kubernetes ecosystem and V() logging is nice, but we can be flexible.

if kind 's `pkg` imports `klog`, and in a case where a third-party front end uses kind as a library, this:
- could create a conflict if the third-party logger already depends on a different klog version for one reason or another. i think this will force the front-end to shift to the particular klog version that kind uses.
- will result in a couple of loggers to be imported if the third-party front-end already uses a separate logger. the klog import will not be needed in this case.

i would have moved the klog import to `cmd` and document (e.g. in a document called "kind as a library"), how to setup the logger if one wishes to import `kind/pkg`.

BenTheElder: > if kind 's pkg imports klog, and in a case where a third-party front end uses kind as a library, this:
> - could create a conflict if the third-party front end already depends on a different klog version for one reason or another. i think this will force the front-end to shift to the particular klog version that kind uses.

How will this be any different, this is the same go module?

> - will result in a couple of loggers to be imported if the third-party front-end already uses a separate logger. the klog import will not be needed in this case.

No? None of the code they link against will import `pkg/log/klog` unless they do themselves or they import `cmd/`.

> i would have moved the klog import to cmd and document (e.g. in a document called "kind as a library"), how to setup the logger if one wishes to import kind/pkg.

 we need an implementation of the shim against klog somwhere, that's all this is. cmd will be the only one importing `pkg/log/klog` and calling `Use()`

neolit123: > How will this be any different, this is the same go module?

as an example, the cluster-api folk are currently trying to create a kind based deployer where the apimachinery kind uses is not suitable for their needs. they have another version in the go.mod.
and technically the versions are conflicting.

@chuckha has more context.

> No? None of the code they link against will import pkg/log/klog unless they do themselves or they import cmd/.

but klog would still be present in the go.mod, unless i'm mistaken here?

BenTheElder: The reason the klog shim implementation is not in `cmd/` is I actually expect most library users will be using klog/glog because of the Kubernetes ecosystem. They're meant to be able to use this implementation, but they probably should not import cmd/ vs just exec-ing to a kind binary.

Since `pkg/*` doesn't import klog* (ONLY `pkg/log/klog`) you only transitively bring klog into your binary if you import `pkg/log/klog` specifically (or `cmd/`). Otherwise you just get the very-klog-like shim and you'll need to figure out implementing it against some other logger.

Because of our existing dependencies, `klog` is already in `go.mod` and has been. (I think this is due to kustomize, and I have seperate thoughts on fixing that, unrelated to this PR :sweat_smile:)

BenTheElder: To summarize:
- klog will be in your dependencies if you vendor / depend on kind, and _already is prior to this PR_.
- klog will not actually be linked into your binary unless you import the shim package*
- cmd/ will import and link against klog
- logging will look pretty similar to kubeadm

\* except that it is because one of our deps, will fix that later

BenTheElder: > as an example, the cluster-api folk are currently trying to create a kind based deployer where the apimachinery kind uses is not suitable for their needs. they have another version in the go.mod.
> and technically the versions are conflicting.

This is already an issue and already solved by:
https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive

BenTheElder: in the cluster-api `go.mod`:
`replace k8s.io/klog => k8s.io/klog $VERSION_CLUSTER_API_WANTS`

BenTheElder: The shim will make this easier to do safely because it makes `pkg/` (again, not the shim package, but no other pkg/ will import that) depend on a tiny subset of klog's functionality that exists in all known versions. Seperately we need to root out the dependency caused by the patch handling code, which will happen but is a little more involved :sweat_smile:

neolit123: wanted @chuckha to leave a comment about the kind provider use case, but feel free to `hold cancel` to unblock the follow ups.

chuckha:
@neolit123 I have no comment on this code.

Overall I'm super happy with the kind codebase and trust it will continue to be very high quality, please keep up the awesome work!

pkg/log/logger.go
```diff
@@ -0,0 +1,37 @@
...
+	V(Level) Logger
+}
+
+// Logger is an interface like klog.Verbose
+// see: https://github.com/kubernetes/klog
+type Logger interface {
+	Info(args ...interface{})
+	Infoln(args ...interface{})
+	Infof(format string, args ...interface{})
+}
```
munnerz: This looks extremely close to the `github.com/go-logr/logr.Logger & InfoLogger` interfaces, albeit with a few different methods.

Structured logging is obviously not required in kind, but it'd be good if some form of means to support logr here would be possible. From a cursory glance, it looks like it'd be possible for me to implement my own version of kind's `Logger` that wraps around a logr (which in turn would allow any other logging implementation that logr supports to be used by kind 🙄).

If so, great 😄

munnerz:
This looks good to me 😄 I'm looking forward to making use of pluggable log backends! The general implementation plan also sounds sensible :)

/lgtm

pkg/log/logger.go
BenTheElder: yeah, the idea is:
- we won't be doing structured logging for a CLI
- the CLI utils are _banned_ from fatal, error, etc. those are not debug logs. those are user facing messages, and only returning an error should happen from library code, not fatal.

... so here's a subset just for logging debug messages with klog, that could be implemented against something else 😅

BenTheElder:
I have a new idea with submodules and a tiny logging implementation that may work nicer for everyone ... stand by ...

PR #615, opened 2019-06-14
add a workaround to a Docker (aufs) bug

jieyu:
You might see the entrypoint script throw out errors like the following:
```
/usr/bin/mount: Text file busy
```

This is due to this bug (likely in AUFS):
https://github.com/moby/moby/issues/9547

This patch workarounds it by inserting a `sync` in between.

### DIFF
--- a/images/base/entrypoint
+++ b/images/base/entrypoint
@@ -37,6 +37,11 @@ fix_mount() {
   chown root:root /bin/mount
   chmod -s /bin/mount
 
+  # This is a workaround to an AUFS bug that might cause `Text file
+  # busy` on `mount` command below. See more details in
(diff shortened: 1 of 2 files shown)


jieyu:
Just signed CLA.

aojea:
is AUFS going to be supported moving forward? 🤔 
https://docs.docker.com/storage/storagedriver/aufs-driver/#prerequisites

jieyu:
@aojea From docker site, it still shows supported. Some ubuntu distro still has that as the default (e.g., our CI machines). The workaround looks simple enough. I cannot reproduce the issue after the workaround in our CI.

PR #616, opened 2019-06-14
Use config file to join nodes to the cluster

aojea:
This avoids having to write configuration files on runtime, ~however, seems that the control plane nodes can't use it~

### DIFF
--- a/pkg/cluster/internal/create/actions/config/config.go
+++ b/pkg/cluster/internal/create/actions/config/config.go
@@ -25,9 +25,11 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	"sigs.k8s.io/kind/pkg/cluster/config"
+	"sigs.k8s.io/kind/pkg/cluster/constants"
 	"sigs.k8s.io/kind/pkg/cluster/internal/create/actions"
 	"sigs.k8s.io/kind/pkg/cluster/internal/kubeadm"
 	"sigs.k8s.io/kind/pkg/cluster/nodes"
+	"sigs.k8s.io/kind/pkg/concurrent"
 	"sigs.k8s.io/kind/pkg/kustomize"
 )
 
@@ -50,6 +52,7 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 		return err
 	}
 
+	// create the kubeadm init configuration
 	node, err := nodes.BootstrapControlPlaneNode(allNodes)
 	if err != nil {
 		return err
@@ -70,31 +73,69 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 		return err
 	}
 
-	// get kubeadm config content
-	kubeadmConfig, err := getKubeadmConfig(
-		ctx.Config,
-		kubeadm.ConfigData{
-			ClusterName:          ctx.ClusterContext.Name(),
-			KubernetesVersion:    kubeVersion,
-			ControlPlaneEndpoint: controlPlaneEndpoint,
-			APIBindPort:          kubeadm.APIServerPort,
-			APIServerAddress:     ctx.Config.Networking.APIServerAddress,
-			Token:                kubeadm.Token,
-			PodSubnet:            ctx.Config.Networking.PodSubnet,
-		},
-	)
-
-	if err != nil {
-		// TODO(bentheelder): logging here
-		return errors.Wrap(err, "failed to generate kubeadm config content")
+	// if there is no load balancer the bootstrap node address is the control plane endpoint
+	if controlPlaneEndpoint == "" {
+		// get the bootstrap node ip address
+		nodeAddress, err := node.IP()
+		if err != nil {
+			return errors.Wrap(err, "failed to get IP for node")
+		}
+		controlPlaneEndpoint = nodeAddress
 	}
 
-	log.Debug("Using kubeadm config:\n" + kubeadmConfig)
+	// create kubeadm init config
+	fns := []func() error{}
 
-	// copy the config to the node
-	if err := node.WriteFile("/kind/kubeadm.conf", kubeadmConfig); err != nil {
-		// TODO(bentheelder): logging here
-		return errors.Wrap(err, "failed to copy kubeadm config to node")
+	kubeadmClusterConfig := kubeadm.ConfigData{
+		ClusterName:          ctx.ClusterContext.Name(),
+		KubernetesVersion:    kubeVersion,
+		ControlPlaneEndpoint: controlPlaneEndpoint,
+		APIBindPort:          kubeadm.APIServerPort,
+		APIServerAddress:     ctx.Config.Networking.APIServerAddress,
+		Token:                kubeadm.Token,
+		PodSubnet:            ctx.Config.Networking.PodSubnet,
+		ControlPlane:         true,
+	}
+
+	fns = append(fns, func() error {
+		return writeKubeadmConfig(ctx.Config, kubeadmClusterConfig, node)
+	})
+
+	// create the kubeadm join configuration for secondary control plane nodes if any
+	secondaryControlPlanes, err := nodes.SecondaryControlPlaneNodes(allNodes)
+	if err != nil {
+		return err
+	}
+	if len(secondaryControlPlanes) > 0 {
+		kubeadmClusterConfig.ControlPlane = true
+		// create the workers concurrently
+		for _, node := range secondaryControlPlanes {
+			node := node // capture loop variable
+			fns = append(fns, func() error {
+				return writeKubeadmConfig(ctx.Config, kubeadmClusterConfig, &node)
+			})
+		}
+	}
+
+	// then create the kubeadm join config for the worker nodes if any
+	workers, err := nodes.SelectNodesByRole(allNodes, constants.WorkerNodeRoleValue)
+	if err != nil {
+		return err
+	}
+	if len(workers) > 0 {
+		kubeadmClusterConfig.ControlPlane = false
+		// create the workers concurrently
+		for _, node := range workers {
+			node := node // capture loop variable
+			fns = append(fns, func() error {
+				return writeKubeadmConfig(ctx.Config, kubeadmClusterConfig, &node)
+			})
+		}
+	}
+
+	// Create the config in all nodes concurrently
+	if err := concurrent.UntilError(fns); err != nil {
+		return err
 	}
 
 	// mark success
@@ -157,4 +198,32 @@ func setPatchNames(patches []string, jsonPatches []kustomize.PatchJSON6902) ([]s
 		fixedJSONPatches[i] = patch
 	}
 	return fixedPatches, fixedJSONPatches
+}
+
+// writeKubeadmConfig writes the kubeadm configuration in the specified node
+func writeKubeadmConfig(cfg *config.Cluster, data kubeadm.ConfigData, node *nodes.Node) error {
+	// get the node ip address
+	nodeAddress, err := node.IP()
+	if err != nil {
+		return errors.Wrap(err, "failed to get IP for node")
+	}
+
+	data.NodeAddress = nodeAddress
+
+	kubeadmConfig, err := getKubeadmConfig(cfg, data)
+
+	if err != nil {
+		// TODO(bentheelder): logging here
+		return errors.Wrap(err, "failed to generate kubeadm config content")
+	}
+
+	log.Debug("Using kubeadm config:\n" + kubeadmConfig)
+
+	// copy the config to the node
+	if err := node.WriteFile("/kind/kubeadm.conf", kubeadmConfig); err != nil {
+		// TODO(bentheelder): logging here
+		return errors.Wrap(err, "failed to copy kubeadm config to node")
+	}
+
+	return nil
 }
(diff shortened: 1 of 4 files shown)


pkg/container/docker/restart.go
```diff
@@ -0,0 +1,30 @@
...
+package docker
+
+import (
+	"sigs.k8s.io/kind/pkg/exec"
+)
+
+// Restart the named container
+func Restart(containerNameOrID string) error {
+	cmd := exec.Command(
+		"docker", "restart",
```
neolit123: why was this change needed?

edit: looks like it was deleted in the next commit. :)

pkg/cluster/internal/create/actions/kubeadmjoin/join.go
```diff
@@ -227,22 +227,21 @@ func runKubeadmJoin(
 		return err
 	}
 
+	if err = writeKubeadmJoinConfiguration(node, joinAddress); err != nil {
+		return errors.Wrapf(err, "failed to generate kubeadm join configuration")
```
neolit123: `...the kubeadm JoinConfiguration`

pkg/cluster/internal/create/actions/kubeadmjoin/join.go
```diff
@@ -284,3 +283,38 @@ func getJoinAddress(ctx *actions.ActionContext, allNodes []nodes.Node) (string,
 
 	return fmt.Sprintf("%s:%d", controlPlaneIP, kubeadm.APIServerPort), nil
 }
+
+func writeKubeadmJoinConfiguration(node *nodes.Node, joinAddress string) error {
+	// get installed kubernetes version from the node image
+	kubeVersion, err := node.KubeVersion()
+	if err != nil {
+		return errors.Wrap(err, "failed to get kubernetes version from node")
```
neolit123: kubernetes -> Kubernetes

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -110,6 +110,11 @@ controllerManagerExtraArgs:
   enable-hostpath-provisioner: "true"
 nodeRegistration:
   criSocket: "/run/containerd/containerd.sock"
+  kubeletExtraArgs:
+    fail-swap-on: "false"
+    {{ if .APIAdvertiseAddress -}}
+    node-ip: "{{ .APIAdvertiseAddress }}"
```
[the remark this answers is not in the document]
aojea: Ipv6 needs it, otherwise Kubelet picks the ipv4

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -349,9 +288,7 @@ func Config(data ConfigData) (config string, err error) {
 
 	// assume the latest API version, then fallback if the k8s version is too low
 	templateSource := ConfigTemplateBetaV2
-	if ver.LessThan(version.MustParseSemantic("v1.12.0")) {
-		templateSource = ConfigTemplateAlphaV2
```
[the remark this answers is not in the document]
aojea: Yeah, they aren't running, also 1.15 is close to be ready... No strong feelings but we should move on sooner or later :)

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -278,6 +287,20 @@ metadata:
   name: config
 `
 
+// JoinConfigTemplate is the kubadm join config template
+const JoinConfigTemplate = `# config generated by kind
+# no-op entry that exists solely so it can be patched
+apiVersion: kubeadm.k8s.io/{{ .KubedamAPI }}
```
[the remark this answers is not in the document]
aojea: Joins are done with the command line, i took one option but I'm missing lot of context, appreciate any feedback and advice

pkg/cluster/internal/kubeadm/config.go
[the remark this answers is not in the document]
neolit123: SGTM

pkg/cluster/internal/kubeadm/config.go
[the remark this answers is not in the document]
neolit123: understood.
would the kubelet long term always default to a ipv4 from the default interface?

pkg/container/docker/restart.go
aojea: Hehehe

pkg/cluster/internal/kubeadm/config.go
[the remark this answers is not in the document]
neolit123: i think we can pass the same kubeadm.cfg to both init and join (with the multi-yaml doc; `---` separated). i was under the impression we are already doing that.

pkg/cluster/internal/kubeadm/config.go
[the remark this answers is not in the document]
aojea: IIRC the logic implemented in Kubernetes always defaults to the IPv4 address ... don´t think that´s a behavior that can change soon ... will add it to my backlog, it can be something interesting to implement though

pkg/cluster/internal/kubeadm/config.go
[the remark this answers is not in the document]
aojea: I abandoned that idea because I thought that will break the `kubeadm join` and it could simplify the join file creation since it needs fewer parameters, any way you're the expert here

aojea:
seems join configuration for v1.12 needs a different template, it fails with 
```
16:38:43.002425 435 interface.go:244] Interface \"eth0\" has 1 addresses :[172.17.0.3/16].\nI0614 16:38:43.002581 435 interface.go:211] Checking addr 172.17.0.3/16.\nI0614 16:38:43.003448 435 interface.go:218] IP found 172.17.0.3\nI0614 16:38:43.003517 435 interface.go:250] Found valid IPv4 address 172.17.0.3 for interface \"eth0\".\nI0614 16:38:43.003609 435 interface.go:395] Found active IP 172.17.0.3 \n[tlsBootstrapToken: Invalid value: \"\": the bootstrap token is invalid, discovery: Invalid value: \"\": discoveryToken or discoveryFile must be set]"
```

neolit123:
it needs a token:
```
apiVersion: kubeadm.k8s.io/v1alpha3
kind: JoinConfiguration
token: "{{ .Token }}"
discoveryTokenUnsafeSkipCAVerification: true

kind: JoinConfiguration
discovery:
  bootstrapToken:
    token: "{{ .Token }}"
    unsafeSkipCAVerification: true

kind: JoinConfiguration
discovery:
  bootstrapToken:
    token: "{{ .Token }}"
    unsafeSkipCAVerification: true
```

pkg/cluster/internal/create/actions/kubeadmjoin/join.go
```diff
@@ -227,22 +227,21 @@ func runKubeadmJoin(
...
-		// uses a well known token and skipping ca certification for automating TLS bootstrap process
-		"--token", kubeadm.Token,
-		"--discovery-token-unsafe-skip-ca-verification",
 		// preflight errors are expected, in particular for swap being enabled
 		// TODO(bentheelder): limit the set of acceptable errors
 		"--ignore-preflight-errors=all",
 		// increase verbosity for debugging
 		"--v=6",
 		"--cri-socket=/run/containerd/containerd.sock",
+		"--config=/kind/kubeadm-join.conf",
```
neolit123: needs to be done for CP nodes too.

pkg/cluster/internal/create/actions/kubeadmjoin/join.go
```diff
@@ -227,22 +227,21 @@ func runKubeadmJoin(
...
-		joinAddress,
-		// uses a well known token and skipping ca certification for automating TLS bootstrap process
-		"--token", kubeadm.Token,
-		"--discovery-token-unsafe-skip-ca-verification",
 		// preflight errors are expected, in particular for swap being enabled
 		// TODO(bentheelder): limit the set of acceptable errors
 		"--ignore-preflight-errors=all",
 		// increase verbosity for debugging
 		"--v=6",
 		"--cri-socket=/run/containerd/containerd.sock",
```
neolit123: no need to pass if present in config.

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -319,6 +287,24 @@ metadata:
...
+metadata:
+  name: config
+nodeRegistration:
+  criSocket: "/run/containerd/containerd.sock"
+  kubeletExtraArgs:
+    fail-swap-on: "false"
+    node-ip: "{{ .NodeAddress }}"
+discovery:
+  bootstrapToken:
+    apiServerEndpoint: "{{ .JoinAddress }}"
```
neolit123: for v1alpha3 it's different:
https://github.com/kubernetes/kubernetes/blob/release-1.12/cmd/kubeadm/app/apis/kubeadm/v1alpha3/types.go#L299-L302

(thus the recommendation to have separate constants)

for v1beta* this seems OK.

pkg/cluster/internal/create/actions/kubeadmjoin/join.go
```diff
@@ -220,29 +189,17 @@ func runKubeadmJoin(
...
-		// uses a well known token and skipping ca certification for automating TLS bootstrap process
-		"--token", kubeadm.Token,
-		"--discovery-token-unsafe-skip-ca-verification",
+		// the join command uses the config file generated in a well known location
+		"--config", "/kind/kubeadm.conf",
 		// preflight errors are expected, in particular for swap being enabled
 		// TODO(bentheelder): limit the set of acceptable errors
 		"--ignore-preflight-errors=all",
-		// increase verbosity for debugging
+		// increase verbosity for debug
```
aojea: seems I cleaned more than necessary 😓

pkg/cluster/internal/kubeadm/config.go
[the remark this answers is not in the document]
BenTheElder: We only did because one of the tests is broken and officially Kubernetes CI won't support it. We should make best effort to support feasible versions.

pkg/cluster/internal/create/actions/config/config.go
```diff
@@ -62,39 +65,66 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
...
+		PodSubnet:            ctx.Config.Networking.PodSubnet,
+		ControlPlane:         true,
+	}
+
+	fns = append(fns, func() error {
+		return writeKubeadmConfig(ctx.Config, kubeadmClusterConfig, node)
+	})
 
+	// create the kubeadm join configuration for secondary control plane nodes if any
+	secondaryControlPlanes, err := nodes.SecondaryControlPlaneNodes(allNodes)
```
BenTheElder: -1
Each node should get it's own config file but they should come from the same template and inputs except for the node IP.

This needs to be consistent for patching, and so that nodes with different Kubernetes versions work.

aojea: yeah, that´s what I´ve changed from the first version.
I think that now the configs and templates are the same for all nodes, all of them are generated with the `writeKubeadmConfig` function, the only difference between nodes config are the node IP address and the isControlPlane flag

pkg/cluster/internal/kubeadm/config.go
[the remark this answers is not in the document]
aojea: I dropped the 1.11 in a separate commit, let me remove that commit and figure out the code for the v1alpha2 config files

pkg/cluster/internal/create/actions/config/config.go
BenTheElder: thanks! :-)

BenTheElder:
I want to take another pass when I'm a little more awake but this looks right now and like a great improvement, thank you!

BenTheElder:
failing to join workers locally with 1.11, investigating...
/hold

BenTheElder:
v1.11 doesn't have https://github.com/kubernetes/kubernetes/commit/9109e62004cd6956f3c5102755b8d2837dc454c2 :upside_down_face: 
will fix in a follow-up
/hold cancel

pkg/cluster/internal/create/actions/config/config.go
```diff
@@ -62,39 +65,66 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
...
 	}
 
-	log.Debug("Using kubeadm config:\n" + kubeadmConfig)
+	// then create the kubeadm join config for the worker nodes if any
+	workers, err := nodes.SelectNodesByRole(allNodes, constants.WorkerNodeRoleValue)
+	if err != nil {
+		return err
+	}
+	if len(workers) > 0 {
+		kubeadmClusterConfig.ControlPlane = false
```
BenTheElder: this can cause all config templates to be controlplane=false (or rather there's a race)

that's because while the function calls will get a pass by value of this struct, they copy from the same shared instance of the struct and we assign to it here, and the goroutine scheduling is racy.

I'm fixing this in the patch to fix v1.11

BenTheElder:
since this hadn't merged by the time I filed the follow up, I'm just including these commits in https://github.com/kubernetes-sigs/kind/pull/632 :sweat_smile:

BenTheElder:
included in https://github.com/kubernetes-sigs/kind/pull/632

thank you!!

PR #617, opened 2019-06-15
Pull and save Docker images in parallel

praseodym:
On my machine, this change speeds up `kind build node-image --type bazel` by a few seconds:
* Images not in local Docker cache: 36.10s => 29.52s
* Images already in local Docker cache: 23.68s => 20.03s

### DIFF
--- a/pkg/build/node/node.go
+++ b/pkg/build/node/node.go
@@ -33,6 +33,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/version"
 
 	"sigs.k8s.io/kind/pkg/build/kube"
+	"sigs.k8s.io/kind/pkg/concurrent"
 	"sigs.k8s.io/kind/pkg/container/docker"
 	"sigs.k8s.io/kind/pkg/exec"
 	"sigs.k8s.io/kind/pkg/fs"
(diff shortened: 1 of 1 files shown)


pkg/build/node/node.go
```diff
@@ -446,23 +449,36 @@ func (c *BuildContext) prePullImages(dir, containerID string) error {
...
-				return err
-			}
-			// TODO(bentheelder): generate a friendlier name
-			pullName := fmt.Sprintf("%d.tar", i)
-			pullTo := path.Join(imagesDir, pullName)
-			err = docker.Save(image, pullTo)
-			if err != nil {
-				return err
+		i, image := i, image // https://golang.org/doc/faq#closures_and_goroutines
+		g.Go(func() error {
```
aojea: what about using current concurrent functions https://github.com/kubernetes-sigs/kind/blob/master/pkg/concurrent/concurrent.go?

There are some examples in the code of using it in Ben´s commit https://github.com/kubernetes-sigs/kind/commit/110dc82f093cdcb7835a01ba85c1ce714083f434

praseodym: Thanks, didn't know about these. Changed! :slightly_smiling_face:

aojea:
Thanks
/ok-to-test

aojea:
SGTM
we should wait for @BenTheElder review

pkg/build/node/node.go
BenTheElder: this other library is really neat though! will need to revisit this sometime, thanks for the pointer :-)

PR #618, opened 2019-06-17
Add kubectl version issue to known-issues.md

alisondy:
Recently stumbled across an issue in slack,  where mismatched kubectl versions result in a SchemaError when applying yamls.

Updating the known-issues to keep track of this issue , and help other people diagnose their deployments

### DIFF
--- a/site/content/docs/user/known-issues.md
+++ b/site/content/docs/user/known-issues.md
@@ -17,11 +17,45 @@ It may additionally be helpful to:
 - reach out and ask for help in [#kind] on the [kubernetes slack]
 
 ## Contents
+* [Failures involving mismatched kubectl versions](#failures-involving-mismatched-kubectl-version)
 * [Docker on Btrfs](#docker-on-btrfs)
 * [Docker installed with Snap](#docker-installed-with-snap)
 * [Failing to apply overlay network](#failing-to-apply-overlay-network)
 * [Failure to build node image](#failure-to-build-node-image)
 * [Failure for cluster to properly start](#failure-for-cluster-to-properly-start)
+
+## Failures involving mistmatched kubectl versions
+
+`kind` will fail create a cluster, if the version of Kubernetes for the client and server differ.
+
+This is a issue that frequently occurs when running `kind` alongside Docker For Mac.
+
+This problem is related to a bug in [docker on macOS][for-mac#3663]
+
+If you see something like the following error message:
+```bash
+kubectl edit deploy -n kube-system kubernetes-dashboard
+error: SchemaError(io.k8s.api.autoscaling.v2beta1.ExternalMetricStatus): invalid object doesn't have additional properties
+```
+
+You can check your client and server versions by running:
+```bash
+kubectl version
+```
+
+If there is a mismatch between the server and client versions, you should install a newer client version. 
+
+If you are using Mac, you can install kubectl via homebrew by running:
+```bash
+brew install kubernetes-cli
+```
+
+And overwrite the symlinks created by Docker For Mac by running:
+```bash
+brew link --overwrite kubernetes-cli
+```
+
+[for-mac#3663]: https://github.com/docker/for-mac/issues/3663
 
 ## Docker on Btrfs
 


alisondy:
Renamed the branch, hence why double pull

site/content/docs/user/known-issues.md
```diff
@@ -17,12 +17,39 @@ It may additionally be helpful to:
...
 ## Contents
+* [Failures involving mismatched kubectl versions](#failures-involving-mismatched-kubectl-version)
 * [Docker on Btrfs](#docker-on-btrfs)
 * [Docker installed with Snap](#docker-installed-with-snap)
 * [Failing to apply overlay network](#failing-to-apply-overlay-network)
 * [Failure to build node image](#failure-to-build-node-image)
 * [Failure for cluster to properly start](#failure-for-cluster-to-properly-start)
 
+## Failures involving mistmatched kubectl versions
+
```
aojea: I think that we could use an introductory line explaining the context , something along the lines `If you have the following errors using kind:`

alisondy: That sounds like a great idea ✨ will update

alisondy: or would ``If you see something like the following error message:`` be better, as I'm not confident the error looks the absolute same every time, just some similarities

aojea: yeah, much better than mine

site/content/docs/user/known-issues.md
```diff
@@ -17,12 +17,39 @@ It may additionally be helpful to:
...
+kubectl edit deploy -n kube-system kubernetes-dashboard
+error: SchemaError(io.k8s.api.autoscaling.v2beta1.ExternalMetricStatus): invalid object doesn't have additional properties
+```
+
+You can check your client and server versions by running:
+```bash
+kubectl version
+```
+
+As a workaround, you can install kubectl via homebrew by running:
```
aojea: I think we should do this only if there is a mismatch between the client and server versions, thus we should explain it

alisondy: im thinking
```
If the client and server versions returned are different, then the cause of the error mention previously, is due to a mismatch between the versions of kubernetes running on the client and server.
```
It's a really long sentence though, do you have any suggestions on breaking it down?

aojea: What about : if there is a mismatch between the server and client versions, you should install a newer client version. If you are using Mac you can install kubectl via homebrew ...

site/content/docs/user/known-issues.md
```diff
@@ -17,12 +17,47 @@ It may additionally be helpful to:
...
+* [Failures involving mismatched kubectl versions](#failures-involving-mismatched-kubectl-version)
 * [Docker on Btrfs](#docker-on-btrfs)
 * [Docker installed with Snap](#docker-installed-with-snap)
 * [Failing to apply overlay network](#failing-to-apply-overlay-network)
 * [Failure to build node image](#failure-to-build-node-image)
 * [Failure for cluster to properly start](#failure-for-cluster-to-properly-start)
 
+## Failures involving mistmatched kubectl versions
+
+`kind` will fail create a cluster, if the version of Kubernetes for the client and server differ.
```
alisondy: ``used by`` instead of  ``for the``?

site/content/docs/user/known-issues.md
```diff
@@ -17,12 +17,47 @@ It may additionally be helpful to:
...
+If you are using Mac, you can install kubectl via homebrew by running:
+```bash
+brew install kubernetes-cli
+```
+
+And overwrite the symlinks created by Docker For Mac by running:
+```bash
+brew link --overwrite kubernetes-cli
+```
+
```
alisondy: nit: extra whitespace

site/content/docs/user/known-issues.md
BenTheElder: I think for the is right

site/content/docs/user/known-issues.md
BenTheElder: :man_shrugging:

site/content/docs/user/known-issues.md
alisondy: oke

site/content/docs/user/known-issues.md
```diff
@@ -17,12 +17,46 @@ It may additionally be helpful to:
...
 * [Docker installed with Snap](#docker-installed-with-snap)
 * [Failing to apply overlay network](#failing-to-apply-overlay-network)
 * [Failure to build node image](#failure-to-build-node-image)
 * [Failure for cluster to properly start](#failure-for-cluster-to-properly-start)
 
+## Failures involving mistmatched kubectl versions
+
+`kind` will fail create a cluster, if the version of Kubernetes for the client and server differ.
+
+This is a issue that frequently occurs when running `kind` alongiside Docker For Mac.
```
BenTheElder: s/alongiside/alongside/
```suggestion
This is a issue that frequently occurs when running `kind` alongside Docker For Mac.
```

aojea:
/lgtm
Thanks

Issue #623, opened 2019-06-17
kind should warn on cluster names that are likely to long to succeed


BenTheElder:
**What would you like to be added**: kind should detect `--name` inputs that are too long (https://github.com/kubernetes-sigs/kind/issues/619) and warn the user that this will likely fail.

**Why is this needed**: To make the root cause easier to identify. These _may_ succeed depending on the host, docker version etc. so instead of trying to set a hard limit I think we can just warn on likely limits for now.

BenTheElder:
fixed in https://github.com/kubernetes-sigs/kind/pull/624

PR #624, opened 2019-06-17
Warn if cluster name is too long.

amwat:
ref: https://github.com/kubernetes-sigs/kind/issues/623

### DIFF
--- a/pkg/cluster/internal/create/create.go
+++ b/pkg/cluster/internal/create/create.go
@@ -41,6 +41,12 @@ import (
 	"sigs.k8s.io/kind/pkg/cluster/internal/create/actions/waitforready"
 )
 
+const (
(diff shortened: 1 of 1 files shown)


pkg/cluster/internal/create/create.go
```diff
@@ -57,6 +63,11 @@ func Cluster(ctx *context.Context, cfg *config.Cluster, opts *Options) error {
 	// may be constructed in memory rather than from disk)
 	encoding.Scheme.Default(cfg)
 
+	// warn if cluster name might typically be too long
+	if len(ctx.Name()) > clusterNameMax {
+		log.Warnf("cluster name %q is too long, might not work properly on some systems", ctx.Name())
```
BenTheElder: nit: s/is/is probably/
we make no promises :sweat_smile: 
maybe also s/might/this might/

amwat: done

PR #627, opened 2019-06-18
upgrade to go 1.12.6

BenTheElder:
Mostly these aren't super notable for us I think, but we should keep up anyhow. This is now out for a week https://github.com/golang/go/issues?q=milestone%3AGo1.12.6


BenTheElder:
Merging manually due to netlify issue blocking tide. Will investigate later.

Issue #525, opened 2019-05-14
Don't proxy docker network automatically when using proxies


aojea:
**What would you like to be added**:

Add the docker network automatically to the NO_PROXY variable.

**Why is this needed**:

In environments where proxies are needed is easy to forget to add the docker network where the `kind` cluster runs to the NO_PROXY environment variable, and this causes a failure trying to create the cluster.

We added the docker network to the NO_PROXY environment in containerd using the entrypoint script, however seems that `kubeadm init` and maybe other components use the environment variables.

I've tried different options to avoid to use the `docker -e` option without any luck.

```
configure_proxy() {
  # Don't use proxy for the local network
  LOCAL_NETWORK=$(ip route list dev eth0 proto kernel | awk '{ print $1 }')

  cat <<EOF > /etc/profile.d/kind-noproxy.sh
export NO_PROXY=${NO_PROXY:-localhost},${LOCAL_NETWORK}"
EOF

  mkdir -p /etc/systemd/system.conf.d/
  cat <<EOF >/etc/systemd/system.conf.d/kind-proxy.conf
[Manager]
DefaultEnvironment="HTTP_PROXY=${HTTP_PROXY:-}" "HTTPS_PROXY=${HTTPS_PROXY:-}" "NO_PROXY=${NO_PROXY:-localhost},${LOCAL_NETWORK}"
EOF

  mkdir -p /etc/systemd/system/containerd.service.d/
  cat <<EOF >/etc/systemd/system/containerd.service.d/http-proxy.conf
[Service]
Environment="HTTP_PROXY=${HTTP_PROXY:-}"
Environment="HTTPS_PROXY=${HTTPS_PROXY:-}"
Environment="NO_PROXY=${NO_PROXY:-localhost},${LOCAL_NETWORK}"
EOF

}

**Comment by BenTheElder** [comment]
I think we need to go ahead and get 0.3 out ASAP (just going to let a little more testing happen after kindnetd), but we definitely need to fix this problem, this is harming a lot of users unnecessarily.

I will prioritize this and other networking issues for 0.4

**Comment by aojea** [comment]
We should wait for https://github.com/kubernetes-sigs/kind/pull/484, once it's merged is just adding a new method to get the docker network address of the cluster.
I guess that the code should be something like this 

```
diff --git a/pkg/cluster/nodes/create.go b/pkg/cluster/nodes/create.go
index 84a9811..f68f0d5 100644
--- a/pkg/cluster/nodes/create.go
[...]

PR #628, opened 2019-06-19
Automatically add the docker network to no_proxy

amwat:
fixes https://github.com/kubernetes-sigs/kind/issues/525

Compute the host docker network subnets and plumb through the NO_PROXY env.

### DIFF
--- a/pkg/cluster/nodes/node.go
+++ b/pkg/cluster/nodes/node.go
@@ -33,6 +33,14 @@ import (
 	"sigs.k8s.io/kind/pkg/exec"
 )
 
+const (
+	// Docker default bridge network is named "bridge" (https://docs.docker.com/network/bridge/#use-the-default-bridge-network)
+	defaultNetwork = "bridge"
+	httpProxy      = "HTTP_PROXY"
+	httpsProxy     = "HTTPS_PROXY"
+	noProxy        = "NO_PROXY"
+)
+
 // Node represents a handle to a kind node
 // This struct must be created by one of: CreateControlPlane
 // It should not be manually instantiated
@@ -257,7 +265,7 @@ type proxyDetails struct {
 // getProxyDetails returns a struct with the host environment proxy settings
 // that should be passed to the nodes
 func getProxyDetails() proxyDetails {
-	var proxyEnvs = []string{"HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY"}
+	var proxyEnvs = []string{httpProxy, httpsProxy, noProxy}
 	var val string
 	var details proxyDetails
 	details.Envs = make(map[string]string)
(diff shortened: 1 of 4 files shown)


tao12345666333:
/lgtm

Thanks.

pkg/cluster/nodes/node.go
```diff
@@ -33,6 +33,14 @@ import (
 	"sigs.k8s.io/kind/pkg/exec"
 )
 
+const (
+	// Docker default bridge network is named "bridge" (https://docs.docker.com/network/bridge/#use-the-default-bridge-network)
+	defaultNetwork = "bridge"
```
aojea: should it be a constant in the `docker` pkg?

pkg/cluster/nodes/node.go
```diff
@@ -273,5 +281,23 @@ func getProxyDetails() proxyDetails {
 			}
 		}
 	}
+
+	// Specifically add the docker network subnets to NO_PROXY
+	subnets, err := getSubnets(defaultNetwork)
+	if err == nil {
```
aojea: is ok to ignore the error? just curiosity

pkg/cluster/nodes/node.go
```diff
@@ -273,5 +281,23 @@ func getProxyDetails() proxyDetails {
...
+		details.Envs[noProxy] = strings.Join(append(subnets, details.Envs[noProxy]), ",")
+		details.Envs[strings.ToLower(noProxy)] = strings.Join(append(subnets, details.Envs[strings.ToLower(noProxy)]), ",")
```
aojea: we are only passing HTTP_PROXY and HTTPS_PROXY, maybe we should pass the lower caps of those too like you are doing here with no_proxy

BenTheElder: this is a good point, but would be reasonable to tackle in a different PR.

seperately I'd like to decouple this tooling more later, and I think HTTPS_PROXY will require a bit more finesse :-)

pkg/cluster/nodes/node.go
BenTheElder: probably yes, but this is also mostly a stub to make #484 viable if we go that route so I'm ok to deal with it later.

the docker <-> node coupling needs a lot more work soon, I'm sketching that out in a branch ... 😅

pkg/container/docker/network.go
```diff
@@ -0,0 +1,16 @@
+package docker
+
+import (
+	"strings"
+
+	"sigs.k8s.io/kind/pkg/exec"
+)
+
+// NetworkInspect displays detailed information on one or more networks
+func NetworkInspect(networkNames []string, format string) ([]string, error) {
```
BenTheElder: IMHO their should be a higher level abstraction that dealing with docker format strings outside this package, but also fine to iterate on this.

pkg/cluster/nodes/node.go
amwat: I previously had it log the error, but decided to get rid of it since there's not really anything to do if we fail to get the subnets.

BenTheElder: it should probably fail?

amwat: not really? everything else would work just fine for users that aren't using a proxy.

BenTheElder: if we didn't detect any proxy info we can elide even querying this instead

pkg/cluster/nodes/node.go
amwat: On second though, we seem to be prioritizing the all-caps envs over the lower cased one on line 278 and also masking them to be all caps finally, was that intentional?

BenTheElder: on the host the user can control this if necessary and we set caps to be higher priority, but on the nodes we should maybe(?) supply the values in both keys since this isn't fully standardized

on the host side, the user can always explicitly set the caps ones when calling kind.

amwat: done

aojea: seems curl only works with lowercase , there are few of stack overflows posts about this. I dont have it clear though

BenTheElder:
CI (prow) is flaky today and we have a release to get going...

PR #629, opened 2019-06-19
refactor build tooling

BenTheElder:
- move containerized build command out of the makefile to POSIX sh
  - this pattern is now somewhat re-usable
  - it's now shellchecked
  - the `docker run ...` flags have some comments documenting why we set them
- move kindnetd image build to use this tooling (it's cached now!)

this could probably use more work, but getting the `kindnetd` build cached will help iterating on this and this dramatically simplifies the makefile, which I'd generally prefer.


hack/build/go_container.sh
```diff
@@ -0,0 +1,89 @@
...
+}
+
+# runs $@ in a go container with caching etc. and the repo mount to /src
+run_in_go_container() {
+  # get user id and group id so we can run the container
+  _UID=$(id -u)
+  _GID=$(id -g)
+  # run in the container
+  docker run \
+    `# ensure the container is removed on exit` \
```
BenTheElder: Shout-out to @amwat for this cool trick with the backticks for online comments, a $() will not work, apparently they have different parsing rules...

Issue #626, opened 2019-06-18
Kind fails to creates pods when IPV6 is disabled in the host


i3oc9i:
**What happened**:

Kind fails to creates pods when IPV6 is disabled in the UBUNTU 18.04  kernel of the host system.

```
GRUB_CMDLINE_LINUX="ipv6.disable=1"
```
`kubectl -n kube-system describe pod/coredns-...` reports the following event.
```
 Warning  FailedCreatePodSandBox  53m                 kubelet, kind-control-plane  Failed create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "4353e3f9fddbe3c1ac60d98576157e037560cb14dc99c76e6458ed9882ebb96f": failed to add route '{:: 00000000000000000000000000000000} via <nil> dev eth0': operation not supported
```
**HACK** Removing  the ` {"dst": "::/0"}` route request in the `/etc/cni/net.d/10-kindnet.conflist` file solves the issue.

```
...
        "routes": [
          {"dst": "0.0.0.0/0"},
          {"dst": "::/0"}
        ],
...
```

**What you expected to happen**

When IPV6 is not available in the host system, then  `kind create cluster` should manage CNI configuration accordingly. 

Actually the CNI configuration is hardcoded in ` kind/pkg/build/node/cni.go` file.

**How to reproduce it (as minimally and precisely as possible)**:

Disable IPV6 support before to create a cluster.

**Anything else we need to know?**:

**WORKAROUND**: After the cluster is created connect into the docker worker node with the following command 
```
docker exec -it kind-control-plane bash
```
and fix the  `/etc/cni/net.d/10-kindnet.conflist` as explained previously.

**Environment:**

- kind version: v.0.3.0

- Kubernetes version: 
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.3", GitCommit:"5e53fd6bc17c0dec8434817e69b04a25d8ae0ff0", GitTreeState:"clean", BuildDate:"2019-06-06T01:44:30Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.2", GitCommit:"66049e3b21efe110454d67df4fa62b08ea79a19b", GitTreeState:"clean", BuildDate:"2019-05-17T00:58:35Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}

[...]

PR #633, opened 2019-06-20
make CNI ipv6 route optional

BenTheElder:
... also move CNI template into kindnetd code to simplify the template, clean up some ipv4/v6 logic in general


BenTheElder:
/cc @aojea 
ref: #626

cmd/kindnetd/main.go
```diff
@@ -71,7 +71,7 @@ func main() {
 	// enforce ip masquerade rules
 	ipv6 := false
 	// TODO: dual stack...?
-	if strings.Contains(hostIP, ":") {
+	if IsIPv6(hostIP) {
```
aojea: and if we replace the if block by `ipv6 := IsIPv6(hostIP)`?

cmd/kindnetd/util.go
```diff
@@ -0,0 +1,28 @@
...
+
+package main
+
+import (
+	"net"
+	"strings"
+)
+
+// IsIPv6 returns true if the address is ipv6
+func IsIPv6(address string) bool {
```
aojea: just found that the net library already has this function, we can avoid this file https://github.com/kubernetes/utils/blob/c55fbcfc754a5b2ec2fbae8fb9dcac36bdba6a12/net/net.go#L118

pkg/build/node/cni.go
```diff
@@ -141,8 +141,7 @@ data:
             "type": "host-local",
             "dataDir": "/run/cni-ipam-state",
             "routes": [
-              {"dst": "0.0.0.0/0"},
-              {"dst": "::/0"}
+              {"dst": "0.0.0.0/0"},` + "{{`{{ if .IPv6 }}\n              {\"dst\": \"::/0\"}{{end}}`}}" + `
```
aojea: I didn´t dig into this but it may possible that we don´t need this field at all since the CNI plugin set the default gw. 
Will investigate and report back

aojea:
if #635 works we don´t need to make it optional

pkg/build/node/cni.go
aojea: it turns out we do need it

aojea:
I have an idea that may work, but need to test it first

pkg/build/node/cni.go
aojea: We should remove this configmap from the installation manifest and embed it into kindnet, this will solve the nesting templating

aojea:
I was talking with @BenTheElder about this and I think that the conclusion was to do in a first iteration:
1. Move the configmap with the CNI configuration from the installation manifest to a constant inside kindnet.
2. Install only the dst route for the configured IP family in kubernetes.

We can review later the need to add multiple routes, we´ll need it for dual stack for sure, but right now kubernetes is only aware of 1 address per pod thus no needed to have both.

cmd/kindnetd/util.go
BenTheElder: yeah, this is also nice because it works for CIDRs or URLs or addresses afaik.

pkg/build/node/cni.go
BenTheElder: yeah, might do that in a different PR

cmd/kindnetd/main.go
BenTheElder: done

pkg/build/node/cni.go
BenTheElder: done

BenTheElder:
rolling this into 0.4 to get the fix in.

BenTheElder:
pretty sure this broke ipv6.

I'm having difficulty debugging locally because ipv6 is broken in other ways on my machine :disappointed_relieved:

cmd/kindnetd/util.go
BenTheElder: it does not work for CIDRS :man_facepalming:

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/650

cmd/kindnetd/util.go
BenTheElder: https://github.com/kubernetes-sigs/kind/pull/650

PR #635, opened 2019-06-20
Don't use routes field in CNI configuration

aojea:
The CNI `routes` parameter is optional https://github.com/containernetworking/cni/blob/spec-v0.3.1/SPEC.md and seems that is redundant with the `gateway` paramater in the `ranges` field https://github.com/containernetworking/plugins/tree/master/plugins/ipam/host-local

Also the ptp plugin seems to install a default route into the pod obtained from the ipam plugin
https://github.com/containernetworking/plugins/blob/master/plugins/main/ptp/ptp.go#L54-L63

### DIFF
--- a/pkg/build/node/cni.go
+++ b/pkg/build/node/cni.go
@@ -140,10 +140,6 @@ data:
           "ipam": {
             "type": "host-local",
             "dataDir": "/run/cni-ipam-state",
-            "routes": [
-              {"dst": "0.0.0.0/0"},
-              {"dst": "::/0"}
-            ],
             "ranges": [
               [
                 {


BenTheElder:
That doesn't appear to work https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_kind/635/pull-kind-conformance-parallel-1-14/1141727294159589378/artifacts/logs/kind-control-plane/pods/kube-system_coredns-fb8b8dccf-k45m9_fc7e690d-936f-11e9-ada0-0242ac110004/coredns/6.log

aojea:
Because I was completely wrong and there is no default route installation, I misread the comment in the ptp plugin code

PR #636, opened 2019-06-21
Initial IPv6 support

BenTheElder:
@aojea's PR (#348) is really close, seeing if we can finish it up before 0.4 :-)

### DIFF
--- a/site/content/docs/user/quick-start.md
+++ b/site/content/docs/user/quick-start.md
@@ -216,6 +216,7 @@ flag:
 kind create cluster --config kind-example-config.yaml
 ```
 
+#### Multi-node clusters
 In particular, many users may be interested in multi-node clusters. A simple
 configuration for this can be achieved with the following config file contents:
 ```yaml
@@ -228,6 +229,7 @@ nodes:
 - role: worker
 ```
 
+#### Control-plane HA
 You can also have a cluster with multiple control-plane nodes:
 ```yaml
 # a cluster with 3 control-plane nodes and 3 workers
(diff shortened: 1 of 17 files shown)


BenTheElder:
the last commit fixed the job (still failing, but only because some of the conformance test ipv6 issues that are known and don't appear to be kind specific)

BenTheElder:
locally tested with single node :wink: 
this is awesome. no major remaining issues. thank you so much @aojea :tada:

BenTheElder:
(self lgtm because my additional commit is a trivial fix, and I already approve all of the remaining code :-))

aojea:
Omg I'm almost crying with the emotion

neolit123:
nice work, guys. :tada:

BenTheElder:
Let's go get a CI job up!

site/content/docs/user/quick-start.md
```diff
@@ -347,3 +363,4 @@ kind, the Kubernetes cluster itself, etc.
 [Kubernetes imagePullPolicy]: https://kubernetes.io/docs/concepts/containers/images/#updating-images
 [Private Registries]: /docs/user/private-registries
 [customize control plane with kubeadm]: https://kubernetes.io/docs/setup/independent/control-plane-flags/
+[docker enable ipv6]: https://docs.docker.com/config/daemon/ipv6/
```
PatrickLang: the instructions on that page don't set `fixed-cidr-v6` so the daemon fails to start

BenTheElder: I remember running into this now... shall we send them a PR? 🙃 
https://github.com/docker/docker.github.io/edit/master/config/daemon/ipv6.md

aojea: They have it documented here https://docs.docker.com/v17.09/engine/userguide/networking/default_network/ipv6/ , don´t know why they didn´t explain it in that link

BenTheElder:
Should we switch links then?

PR #637, opened 2019-06-22
add Port mappings to Node specs

### LINKED ISSUE #99: Support exposing additional ports to be used for NodePort services

fllaca:
Fixes #99: Allows to specify in configuration port mappings for the cluster nodes:

```yaml
kind: Cluster
nodes:
- role: control-plane
  extraPortMappings:
  - containerPort: 80
    hostPort: 80
  extraMounts:
  - containerPath: /mounts
    hostPath: /var/lib/kind/mounts
```
I have tested and used this locally exposing ingress as a daemonset and hostPorts:

### DIFF
--- a/pkg/cluster/config/v1alpha2/zz_generated.conversion.go
+++ b/pkg/cluster/config/v1alpha2/zz_generated.conversion.go
@@ -83,5 +83,6 @@ func autoConvert_config_Node_To_v1alpha2_Node(in *config.Node, out *Node, s conv
 	out.Role = NodeRole(in.Role)
 	out.Image = in.Image
 	out.ExtraMounts = *(*[]cri.Mount)(unsafe.Pointer(&in.ExtraMounts))
+	// WARNING: in.ExtraPortMappings requires manual conversion: does not exist in peer-type
 	return nil
 }
--- a/pkg/cluster/internal/create/nodes.go
+++ b/pkg/cluster/internal/create/nodes.go
@@ -215,7 +215,7 @@ func (d *nodeSpec) Create(clusterLabel string) (node *nodes.Node, err error) {
 	// TODO(bentheelder): decouple from config objects further
 	switch d.Role {
 	case constants.ExternalLoadBalancerNodeRoleValue:
-		node, err = nodes.CreateExternalLoadBalancerNode(d.Name, d.Image, clusterLabel, d.APIServerAddress, d.APIServerPort, d.ExtraPortMappings)
+		node, err = nodes.CreateExternalLoadBalancerNode(d.Name, d.Image, clusterLabel, d.APIServerAddress, d.APIServerPort)
 	case constants.ControlPlaneNodeRoleValue:
 		node, err = nodes.CreateControlPlaneNode(d.Name, d.Image, clusterLabel, d.APIServerAddress, d.APIServerPort, d.ExtraMounts, d.ExtraPortMappings)
 	case constants.WorkerNodeRoleValue:
--- a/pkg/cluster/internal/create/nodes.go
+++ b/pkg/cluster/internal/create/nodes.go
@@ -125,10 +125,11 @@ func createNodeContainers(
 // nodeSpec describes a node to create purely from the container aspect
 // this does not inlude eg starting kubernetes (see actions for that)
 type nodeSpec struct {
-	Name        string
-	Role        string
-	Image       string
-	ExtraMounts []cri.Mount
+	Name              string
+	Role              string
+	Image             string
+	ExtraMounts       []cri.Mount
+	ExtraPortMappings []cri.PortMapping
 	// TODO(bentheelder): replace with a cri.PortMapping when we have that
 	APIServerPort    int32
 	APIServerAddress string
@@ -180,13 +181,14 @@ func nodesToCreate(cfg *config.Cluster, clusterName string) []nodeSpec {
 			apiServerAddress = "127.0.0.1" // only the LB needs to be non-local
 		}
 		desiredNodes = append(desiredNodes, nodeSpec{
-			Name:             nameNode(role),
-			Image:            configNode.Image,
-			Role:             role,
-			ExtraMounts:      configNode.ExtraMounts,
-			APIServerAddress: apiServerAddress,
-			APIServerPort:    apiServerPort,
-			IPv6:             ipv6,
+			Name:              nameNode(role),
+			Image:             configNode.Image,
+			Role:              role,
+			ExtraMounts:       configNode.ExtraMounts,
+			ExtraPortMappings: configNode.ExtraPortMappings,
+			APIServerAddress:  apiServerAddress,
+			APIServerPort:     apiServerPort,
+			IPv6:              ipv6,
 		})
 	}
 
@@ -215,9 +217,9 @@ func (d *nodeSpec) Create(clusterLabel string) (node *nodes.Node, err error) {
 	case constants.ExternalLoadBalancerNodeRoleValue:
 		node, err = nodes.CreateExternalLoadBalancerNode(d.Name, d.Image, clusterLabel, d.APIServerAddress, d.APIServerPort)
 	case constants.ControlPlaneNodeRoleValue:
-		node, err = nodes.CreateControlPlaneNode(d.Name, d.Image, clusterLabel, d.APIServerAddress, d.APIServerPort, d.ExtraMounts)
+		node, err = nodes.CreateControlPlaneNode(d.Name, d.Image, clusterLabel, d.APIServerAddress, d.APIServerPort, d.ExtraMounts, d.ExtraPortMappings)
 	case constants.WorkerNodeRoleValue:
-		node, err = nodes.CreateWorkerNode(d.Name, d.Image, clusterLabel, d.ExtraMounts)
+		node, err = nodes.CreateWorkerNode(d.Name, d.Image, clusterLabel, d.ExtraMounts, d.ExtraPortMappings)
 	default:
 		return nil, errors.Errorf("unknown node role: %s", d.Role)
 	}
--- a/pkg/cluster/internal/create/nodes.go
+++ b/pkg/cluster/internal/create/nodes.go
@@ -215,7 +215,7 @@ func (d *nodeSpec) Create(clusterLabel string) (node *nodes.Node, err error) {
 	// TODO(bentheelder): decouple from config objects further
 	switch d.Role {
 	case constants.ExternalLoadBalancerNodeRoleValue:
-		node, err = nodes.CreateExternalLoadBalancerNode(d.Name, d.Image, clusterLabel, d.APIServerAddress, d.APIServerPort)
+		node, err = nodes.CreateExternalLoadBalancerNode(d.Name, d.Image, clusterLabel, d.APIServerAddress, d.APIServerPort, d.ExtraPortMappings)
 	case constants.ControlPlaneNodeRoleValue:
 		node, err = nodes.CreateControlPlaneNode(d.Name, d.Image, clusterLabel, d.APIServerAddress, d.APIServerPort, d.ExtraMounts, d.ExtraPortMappings)
 	case constants.WorkerNodeRoleValue:
--- a/pkg/cluster/nodes/create.go
+++ b/pkg/cluster/nodes/create.go
@@ -59,12 +59,16 @@ func CreateControlPlaneNode(name, image, clusterLabel, listenAddress string, por
 		port = p
 	}
 
-	portMapping := net.JoinHostPort(listenAddress, fmt.Sprintf("%d", port)) + fmt.Sprintf(":%d", kubeadm.APIServerPort)
+	// add api server port mapping
+	portMappingsWithAPIServer := append(portMappings, cri.PortMapping{
+		ListenAddress: listenAddress,
+		HostPort:      port,
+		ContainerPort: kubeadm.APIServerPort,
+	})
 	node, err = createNode(
-		name, image, clusterLabel, constants.ControlPlaneNodeRoleValue, mounts, portMappings,
+		name, image, clusterLabel, constants.ControlPlaneNodeRoleValue, mounts, portMappingsWithAPIServer,
 		// publish selected port for the API server
 		"--expose", fmt.Sprintf("%d", port),
-		"-p", portMapping,
 	)
 	if err != nil {
 		return node, err
@@ -80,7 +84,7 @@ func CreateControlPlaneNode(name, image, clusterLabel, listenAddress string, por
 
 // CreateExternalLoadBalancerNode creates an external loab balancer node
 // and gets ready for exposing the the API server and the load balancer admin console
-func CreateExternalLoadBalancerNode(name, image, clusterLabel, listenAddress string, port int32) (node *Node, err error) {
+func CreateExternalLoadBalancerNode(name, image, clusterLabel, listenAddress string, port int32, portMappings []cri.PortMapping) (node *Node, err error) {
 	// gets a random host port for control-plane load balancer
 	// gets a random host port for the API server
 	if port == 0 {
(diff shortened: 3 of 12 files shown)


pkg/cluster/internal/create/nodes.go
```diff
@@ -213,11 +215,11 @@ func (d *nodeSpec) Create(clusterLabel string) (node *nodes.Node, err error) {
 	// TODO(bentheelder): decouple from config objects further
 	switch d.Role {
 	case constants.ExternalLoadBalancerNodeRoleValue:
-		node, err = nodes.CreateExternalLoadBalancerNode(d.Name, d.Image, clusterLabel, d.APIServerAddress, d.APIServerPort)
+		node, err = nodes.CreateExternalLoadBalancerNode(d.Name, d.Image, clusterLabel, d.APIServerAddress, d.APIServerPort, d.ExtraPortMappings)
```
aojea: do we need this for the external load balancer?

pkg/cluster/nodes/create.go
```diff
@@ -80,7 +84,7 @@ func CreateControlPlaneNode(name, image, clusterLabel, listenAddress string, por
 
 // CreateExternalLoadBalancerNode creates an external loab balancer node
 // and gets ready for exposing the the API server and the load balancer admin console
-func CreateExternalLoadBalancerNode(name, image, clusterLabel, listenAddress string, port int32) (node *Node, err error) {
+func CreateExternalLoadBalancerNode(name, image, clusterLabel, listenAddress string, port int32, portMappings []cri.PortMapping) (node *Node, err error) {
```
aojea: same comment as above, is there an use case for port mapping in the load balancer node?

pkg/container/docker/critoflags.go
```diff
@@ -71,3 +71,16 @@ func generateMountBindings(mounts ...cri.Mount) []string {
 	}
 	return result
 }
+
+func generatePortMappings(portMappings ...cri.PortMapping) []string {
+	result := make([]string, 0, len(portMappings))
+	for _, pm := range portMappings {
+		publish := fmt.Sprintf("%d:%d", pm.HostPort, pm.ContainerPort)
+		if pm.ListenAddress != "" {
+			publish = fmt.Sprintf("%s:%s", pm.ListenAddress, publish)
```
aojea: ListenAddress can be an IPv6 address and I think that we need to add square brackets in that case,  check if this approach works for you https://github.com/kubernetes-sigs/kind/blob/edbae5fdd1f1670eef0323018ccb18ced552fd29/pkg/cluster/nodes/create.go#L62

aojea:
/ok-to-test
Thanks for the PR
 @BenTheElder , I remember you mention something about tainting the nodes, could you expand on that?
Another concern that I have is about possible port conflicts, maybe we should add some validation ?https://github.com/kubernetes-sigs/kind/blob/master/pkg/cluster/config/validate.go

pkg/container/cri/types.go
```diff
@@ -49,6 +49,20 @@ type Mount struct {
...
+//  containerPort: 80
+//  hostPort: 8000
+//  listenAddress: 127.0.0.1
+type PortMapping struct {
+	// Port within the container.
+	ContainerPort int32 `protobuf:"varint,1,opt,name=container_port,json=containerPort,proto3" json:"containerPort,omitempty"`
+	// Port on the host.
+	HostPort int32 `protobuf:"varint,2,opt,name=host_path,json=hostPort,proto3" json:"hostPort,omitempty"`
+	// TODO: add protocol (tcp/udp) and port-ranges
+	ListenAddress string `protobuf:"bytes,3,opt,name=listenAddress,json=hostPort,proto3" json:"listenAddress,omitempty"`
```
aojea: I don't know if this should be the same address that we are using for the apiserveraddress
https://github.com/kubernetes-sigs/kind/blob/edbae5fdd1f1670eef0323018ccb18ced552fd29/pkg/cluster/config/types.go#L96

It seems unlikely that someone wants to use different addresses for forwarding ports, thoughts?

pkg/cluster/nodes/create.go
fllaca: I added the `ExtraPortMapping` capability to all roles but I think I agree with you, it doesn't make too much sense having this feature in load balancer nodes, I'll revert these changes here.

pkg/container/docker/critoflags.go
fllaca: Fixed! Thanks!

pkg/container/cri/types.go
fllaca: The current behavior is that it will listen on `0.0.0.0` (since this is the behavior of `docker run --publish` if no listenAddress is given). :thinking: Maybe we can default this value to `APIServerAddress` if no ListenAddress is specified, but not sure what is best.

aojea: @BenTheElder can you help here? ^^^

fllaca:
@aojea honestly I don't know why [pull-kind-conformance-parallel](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_kind/637/pull-kind-conformance-parallel/1142764230181654530/). I check the logs [here](https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_kind/637/pull-kind-conformance-parallel/1142764230181654530/) but I'm not able to spot the issue :(

aojea:
@fllaca it´s because of this https://github.com/kubernetes-sigs/kind/issues/639 , 
I´m submitting a PR, don´t worry about that.

By the way, It will be nice if you add an example on how to use it to the docs

fllaca:
Thanks @aojea ! I added this piece of docs: https://github.com/kubernetes-sigs/kind/blob/45c69d6cb66f9a7aec7d1243c42331cafb6cb69a/site/content/docs/user/quick-start.md#mapping-ports-to-the-host-machine

fllaca:
yay! tests passing now! Thanks @aojea

pkg/cluster/config/v1alpha2/zz_generated.conversion.go
```diff
@@ -83,5 +83,6 @@ func autoConvert_config_Node_To_v1alpha2_Node(in *config.Node, out *Node, s conv
 	out.Role = NodeRole(in.Role)
 	out.Image = in.Image
 	out.ExtraMounts = *(*[]cri.Mount)(unsafe.Pointer(&in.ExtraMounts))
+	// WARNING: in.ExtraPortMappings requires manual conversion: does not exist in peer-type
```
BenTheElder: I think we can just drop v1alph2 for 0.4 to nudge things forward, I'm starting to think about v1alpha4 :-)

BenTheElder: (this PR is fine, but this is a good reminder to drop that version)

pkg/container/cri/types.go
BenTheElder: I think it's reasonable to match docker here for now, with this field you've specifically asked to forward a port, whereas the apiServer port kind is doing even if you don't know about that as a user. Since the user may not even know, we're a bit warier about exposing it beyond localhost by default (also that port is definitely hosting something powerful that should be secured).

In a future API revision we can change the defaulting easily since we're still alpha :sweat_smile:

BenTheElder:
NOTE: we expect this to fail currently^ (the conformance tests are broken on ipv6 for everyone still!) but I want to sanity check that the new port mappings don't cause it to fail _more_ of them. (it all looks right but, why not also check :-))

BenTheElder:
> Another concern that I have is about possible port conflicts, maybe we should add some validation ?https://github.com/kubernetes-sigs/kind/blob/master/pkg/cluster/config/validate.go

absolutely, I'm happy to handle that in a follow-up PR though.

BenTheElder:
hmm the coreDNS pods did not come up, that's not expected
/retest

BenTheElder:
/retest
patched ipv6 now I think (unrelated bug) :-)

BenTheElder:
merged via #654 (sorted out the validation first in additional commits, rebased everything up to the latest master, etc.)

BenTheElder:
thank you!

UkiahSmith:
> This can be useful if using NodePort services or daemonsets exposing host ports.

@fllaca [The documentation](https://kind.sigs.k8s.io/docs/user/quick-start/#mapping-ports-to-the-host-machine) states that the `extraPortMappings` can used with NodePort, but I was unable to get this working with a NodPort Service.

Will you clarify the steps taken to have a NodePort Service exposed to the host machine?  My specific use case is Docker on macOS.

Ciantic:
@soliverprofesional yes it solved everything! Thank you. I think `kind` should have an official Hello World, and this could be it. It's simple and reproducible, someone should put this to docs.

Side note, finding a nginx hello world from *any* kubernetes site/docs is difficult. LoadBalancer hello world is the most common but that can't be used with `kind`.

jmara:
Hi,

Side-note - Using only imperative commands it looks like this:
```
# create the deployment
kubectl run nginx --image=nginx --replicas=1

# expose as service nodePort
kubectl expose deploy nginx --port=80 --target-port=80 --type=NodePort

# sadly ... change the nodePort to 30080
kubectl patch svc nginx --type=json -p '[{"op": "replace", "path": "/spec/ports/0/nodePort","value":30080}]'

```

Cheers,
Jan

UkiahSmith:
@soliverprofesional Thank you for your reply, I have it working also with these settings.

```yaml
kind: Cluster
metadata:
  name: kind-dev
nodes:
  - role: control-plane
    extraPortMappings:
      - hostPort: 7070
        containerPort: 30080
  - role: worker
  - role: worker
```

I tested this with one control node and two worker nodes, and with a single node that was control and worker. Both worked.


```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: deployment-thing
  labels:
    tier: web
    area: tool
spec:
  replicas: 4
  selector:
    matchLabels:
      tier: web
      area: tool
  template:
    metadata:
      name: thing-pod
      labels:
        tier: web
        area: tool
    spec:
      containers:
        - name: thing
          image: ukiahsmith/gecho:v0.2.0
          ports:
            - containerPort: 8080

---
apiVersion: v1
kind: Service
metadata:
  name: thing
  labels:
    tier: web
    area: tool
spec:
  type: NodePort
  ports:
    - name: thing
      port: 8080
      targetPort: 8080
      nodePort: 30080
  selector:
    area: tool
```

This launched 4 pods, with 2 pods on each worker node. Requests to `localhost:7070` were load balanced across the 4 pods. 

```bash
$ for i in {1..16}; do curl -s localhost:7070/; echo " " ; done | sort -h
{"hostname":"deployment-thing-86df946984-5vp6q","ips":["127.0.0.1/8","::1/128","10.244.2.2/24","fe80::c034:beff:feb2:a0ec/64"],"version":"v0.2.0"}
{"hostname":"deployment-thing-86df946984-5vp6q","ips":["127.0.0.1/8","::1/128","10.244.2.2/24","fe80::c034:beff:feb2:a0ec/64"],"version":"v0.2.0"}
{"hostname":"deployment-thing-86df946984-5vp6q","ips":["127.0.0.1/8","::1/128","10.244.2.2/24","fe80::c034:beff:feb2:a0ec/64"],"version":"v0.2.0"}
{"hostname":"deployment-thing-86df946984-5vp6q","ips":["127.0.0.1/8","::1/128","10.244.2.2/24","fe80::c034:beff:feb2:a0ec/64"],"version":"v0.2.0"}
{"hostname":"deployment-thing-86df946984-5vp6q","ips":["127.0.0.1/8","::1/128","10.244.2.2/24","fe80::c034:beff:feb2:a0ec/64"],"version":"v0.2.0"}
{"hostname":"deployment-thing-86df946984-5vp6q","ips":["127.0.0.1/8","::1/128","10.244.2.2/24","fe80::c034:beff:feb2:a0ec/64"],"version":"v0.2.0"}
{"hostname":"deployment-thing-86df946984-5vp6q","ips":["127.0.0.1/8","::1/128","10.244.2.2/24","fe80::c034:beff:feb2:a0ec/64"],"version":"v0.2.0"}
{"hostname":"deployment-thing-86df946984-ck8bp","ips":["127.0.0.1/8","::1/128","10.244.1.3/24","fe80::b431:c9ff:fe79:abb6/64"],"version":"v0.2.0"}
{"hostname":"deployment-thing-86df946984-cnhrl","ips":["127.0.0.1/8","::1/128","10.244.2.3/24","fe80::88da:4cff:fedd:956f/64"],"version":"v0.2.0"}
{"hostname":"deployment-thing-86df946984-cnhrl","ips":["127.0.0.1/8","::1/128","10.244.2.3/24","fe80::88da:4cff:fedd:956f/64"],"version":"v0.2.0"}
{"hostname":"deployment-thing-86df946984-cnhrl","ips":["127.0.0.1/8","::1/128","10.244.2.3/24","fe80::88da:4cff:fedd:956f/64"],"version":"v0.2.0"}
{"hostname":"deployment-thing-86df946984-cnhrl","ips":["127.0.0.1/8","::1/128","10.244.2.3/24","fe80::88da:4cff:fedd:956f/64"],"version":"v0.2.0"}
{"hostname":"deployment-thing-86df946984-kmdqs","ips":["127.0.0.1/8","::1/128","10.244.1.2/24","fe80::3018:2dff:fe93:ccec/64"],"version":"v0.2.0"}
{"hostname":"deployment-thing-86df946984-kmdqs","ips":["127.0.0.1/8","::1/128","10.244.1.2/24","fe80::3018:2dff:fe93:ccec/64"],"version":"v0.2.0"}
{"hostname":"deployment-thing-86df946984-kmdqs","ips":["127.0.0.1/8","::1/128","10.244.1.2/24","fe80::3018:2dff:fe93:ccec/64"],"version":"v0.2.0"}
{"hostname":"deployment-thing-86df946984-kmdqs","ips":["127.0.0.1/8","::1/128","10.244.1.2/24","fe80::3018:2dff:fe93:ccec/64"],"version":"v0.2.0"}
```

Everything is being routed through the control plane. Which for my purposes is fine, as long as it acts like a normal cluster.

```
2342cccd6052        kindest/node:v1.15.3   "/usr/local/bin/entr…"   31 minutes ago      Up 31 minutes                                                                       kind-worker
5179cd918b16        kindest/node:v1.15.3   "/usr/local/bin/entr…"   31 minutes ago      Up 31 minutes                                                                       kind-worker2
704fc87fbb90        kindest/node:v1.15.3   "/usr/local/bin/entr…"   31 minutes ago      Up 31 minutes       61999/tcp, 127.0.0.1:61999->6443/tcp, 0.0.0.0:7070->30080/tcp   kind-control-plane
```

galderz:
@fllaca Thanks for adding this! We were having a lot of issues with minikube on travis due to resources, and switching to kind really enabled our testsuite to run reliably (see [infinispan/infinispan-operator#183](https://github.com/infinispan/infinispan-operator/pull/183). 

I think the kind documentation could be improved in this area since the example kind configuration linked in documentation has a lot of baggage that you're unsure if you need (?). 

Eventually I came across [kind-ci/examples](https://github.com/kind-ci/examples) which had a fairly barebones [kind config](https://github.com/kind-ci/examples/blob/master/kind-config.yaml), but even that one has too much for travis IMO. AFAIK, you just need the control-plane and forward a port and that should be enough. That seemed to work for us.

Issue #639, opened 2019-06-23
the hardcoded CNI manifest fails to apply against current k/k master


neolit123:
**What happened**:
in the current k/k master `extensions/v1beta1` is going away for DaemonSets:
https://github.com/kubernetes/kubernetes/pull/70672

this broke applying our hardcoded CNI manifest:
https://github.com/kubernetes-sigs/kind/blob/6110fdc6af652e25c6e9cc3b7eb50e5546141743/pkg/build/node/cni.go#L165
https://k8s-testgrid.appspot.com/sig-testing-kind#conformance,%20master%20(dev)

i can try sending a fix later, though we me might have to template the manifest with the GroupVersion. a major problem is that older versions of kind will fail deploying against the latest k/k master.

unfortunately hardcoding the manifest is a double edge sword.

the kubeadm/kind jobs are also failing on 0.2, but there is no solution there because we cannot backport a patch (no branches in k-sigs/kind). this means we have to move the job to latest kind master.

**What you expected to happen**:
a CNI config to be applied.

**How to reproduce it (as minimally and precisely as possible)**:
https://k8s-testgrid.appspot.com/sig-testing-kind#conformance,%20master%20(dev)

**Anything else we need to know?**:
NO

**Environment:**
CI
latest k/k master
latest k-sigs/kind master

aojea:
We are safe in the sense that the `disableCNI` option is there, I mean, all versions can choose to disable the CNI and add an extra step to set it up.

Ben demonstrates to have a great vision 🥇

aojea:
It seems that we only need to do this change
```

apiVersion: v1/beta2 # For Kubernetes version 1.9 and later, use apps/v1
```

aojea:
/assign @aojea 
This https://github.com/kubernetes-sigs/kind/pull/641 fixes the problem

BenTheElder:
apps/v1 exists in all supported kind versions, we just shouldn't have used extensions the switch should solve this.

this could still bite us someday, will think about this some more...

BenTheElder:
we should also look at:
- getting more tests running under kind (expand scope beyond conformance)
- making kind a default (though non-blocking initially) kubernetes presubmit ... there's been a lot of talk about doing this from SIG Testing / SIG Cluster Lifecycle, we should probably at least look at  running it.

aojea:
I think we should start adding some periodic tests with a bigger scope
+1 to kubernetes presubmit

neolit123:
[...]

PR #641, opened 2019-06-23
Change CNI daemonset api version

aojea:
Fixes: https://github.com/kubernetes-sigs/kind/issues/639

### DIFF
--- a/pkg/build/node/cni.go
+++ b/pkg/build/node/cni.go
@@ -162,7 +162,7 @@ data:
       ]
     }
 ---
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: DaemonSet
 metadata:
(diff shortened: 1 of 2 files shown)


aojea:
/assign @neolit123 
/assign @BenTheElder 

We need this to unblock the gates

pkg/build/node/cni.go
```diff
@@ -172,6 +172,9 @@ metadata:
     app: kindnet
     k8s-app: kindnet
 spec:
+  selector:
```
BenTheElder: aside: this requirement to duplicate the labels exactly here is a really unfortunate aspect of the kubernetes API, should send feedback upstream about this

BenTheElder:
added a commit bumping the base image to include this

PR #644, opened 2019-06-23
[WIP] Come back to haproxy

aojea:
Seems that haproxy is the only load balancer that allows to check the health endpoint from the api servers while doing TCP load balancing.

This PR restores Haproxy as the load balancer for the kind cluster, and leverages the docker commands `create` and `start` to create the load balancer node.

Fixes: #588


aojea:
/assign @neolit123 
/assign @BenTheElder 

I still need to do more testing, but the simple scenario is working in my laptop

pkg/container/docker/run.go
```diff
@@ -62,7 +62,7 @@ func WithMounts(mounts []cri.Mount) RunOpt {
 }
 
 // Run creates a container with "docker run", with some error handling
-func Run(image string, opts ...RunOpt) error {
+func Run(image string, createonly bool, opts ...RunOpt) error {
```
BenTheElder: -1 too boolean args, this is not `docker.Run` anymore

pkg/cluster/internal/loadbalancer/const.go
```diff
@@ -20,7 +20,7 @@ package loadbalancer
 const ControlPlanePort = 6443
 
 // Image defines the loadbalancer image:tag
-const Image = "nginx:1.15.12-alpine"
+const Image = "haproxy:1.9.8-alpine"
```
BenTheElder: we should go back to the previous approach of using our own image with a default config. I have this WIP.

aojea:
Duplicates

PR #645, opened 2019-06-24
switch external loadbalancer back to haproxy, health check

BenTheElder:
... finally got the image build in reasonable shape, cross platform images are still a total mess with docker (there's no command to delete broken manifest! 🙄)

- build and push our own haproxy image (literally the upstream image, but built to be a manifest image again after adding a trivial default config
- use that image, configure the loadbalancer incl healthchecks

/cc @aojea


pkg/cluster/internal/loadbalancer/config.go
```diff
@@ -32,26 +32,33 @@ type ConfigData struct {
 
 // DefaultConfigTemplate is the loadbalancer config template
 const DefaultConfigTemplate = `# generated by kind
```
BenTheElder: this should look familiar, it's probably good enough, but I'm still iterating on it.

BenTheElder:
/hold
want to do more testing etc... putting this up for looking though as it's good enough now.

images/haproxy/push-cross.sh
```diff
@@ -0,0 +1,75 @@
...
+cd ./images/haproxy
+
+IMAGE="${IMAGE:-kindest/haproxy}"
+TAG="${TAG:-2.0.0-alpine}"
+BASE="haproxy:${TAG}"
+
+# tag arch, manifest architecture, variant
+ARCHES=(
+  "amd64,amd64,"
+  "arm32v6,arm,v6"
```
aojea: is this comma a typo or intended? `arm,v6` or `armv6`

BenTheElder: docker manifest is a mess, this is intended. see the code below
this is:
- `arm32v6` when pulling the matching base image
- arm for the `architecture` in the manifest
- v6 for the variant
The comment above the array references this, but it's all kinda manifest internal stuff. I didn't want to inline an explanation of manifests 😅

BenTheElder: these are basically tuple (tagged_architecture,architecture,variant)
then we make another one below with (image_we_build,tagged_architecture,architecture,variant)

tao12345666333: do we need add arm v7?

images/haproxy/push-cross.sh
```diff
@@ -0,0 +1,75 @@
...
+  docker push "${image}"
+  # join image we tagged with arch info for building the manifest later 
+  images+=("${image}")
+  image_infos+=("${image},${arch_info}")
+done
+
+# This option is required for running the docker manifest command
+export DOCKER_CLI_EXPERIMENTAL="enabled"
+
+# create and push the manifest
```
aojea: why not building and creating the manifest?
just curious, I never did this before, but seems too complicated having to use all this `regex`

pkg/cluster/internal/loadbalancer/const.go
```diff
@@ -20,7 +20,7 @@ package loadbalancer
 const ControlPlanePort = 6443
 
 // Image defines the loadbalancer image:tag
-const Image = "nginx:1.15.12-alpine"
+const Image = "kindest/haproxy:2.0.0-alpine"
```
aojea: I remember talking about prepulling the `loadbalancer` image before but don´t remember the conclusion 😅 , should we prepull it?

aojea:
SGTM

just the comments about the manifests because I really don´t understand them and the question about prepulling the image

pkg/cluster/internal/loadbalancer/const.go
BenTheElder: It's not prepulled because it runs on the host directly as a node. It only needs to pull once though.

images/haproxy/push-cross.sh
BenTheElder: Manifests are a meta image pointing to images for different hosts in this context, I'm not sure what you mean by "why not building and creating the manifest?"

images/haproxy/push-cross.sh
BenTheElder: Alpine doesn't support v7 last I checked, bit arm v6 images / binaries work on v7. I'm pretty sure our only "users" so far are on arm64v8

images/haproxy/push-cross.sh
BenTheElder: The regex is actually pretty simple and just to extract parts of the comma-joined tuple.
Each image has some metadata we need. We need to know:
- what we built & tagged the specific image as
- what architecture it was for
- what variant, if any it was for
- when building: what to pull it as (which isn't always the same as arch + variant ...)
When we build the "meta image" (manifest list image) we have to supply most of that info for each specific image we built.

BenTheElder: Kubernetes has some extremely similar logic for all of the cross platform images.

At some point I'll DRY this out to be nicer with kindnetd but for now this should be suitable and we only need to build if / when we want to upgrade haproxy anyhow.

pkg/cluster/internal/loadbalancer/config.go
```diff
@@ -32,26 +32,33 @@ type ConfigData struct {
...
+  {{ if .IPv6 -}}
+  bind :::{{ .ControlPlanePort }};
+  {{- end }}
+  default_backend kube-apiservers
 
-# minimal events entry to make nginx happy
-events {}
+backend kube-apiservers
+  option httpchk GET /healthz
+  # TODO: we should be verifying (!)
```
BenTheElder: This aspect is super not ideal but could be solved later.
Ideally where possible kind should behave like production. There's no real reason we need to be more insecure here other than :effort: and logistics around the certs. We should be able to do SSL backends with verified certs.

BenTheElder: This is the status quo so it shouldn't block the PR, but ...

images/haproxy/push-cross.sh
tao12345666333:
```
 {                                                                                                              
    "mediaType": "application/vnd.docker.distribution.manifest.v2+json",                                         
    "size": 946,
    "digest": "sha256:35d94f72942e907a56cc201a700e8f2378a204f40b641c55969fabbd259d1ca7",
    "platform": {                                                                                                
      "architecture": "arm",                                                                                     
      "os": "linux",                                                                                             
      "variant": "v7"                                                                                            
    }                                                                                                            
  },       
```
it has arm v7 manifest.

images/haproxy/push-cross.sh
aojea: yeah, that makes sense.
For "why not building and creating the manifest? I meant to do everything in the same `for` loop, seems that the variables you are extracting are available at build time

pkg/cluster/internal/loadbalancer/config.go
aojea: this is an interesting topic. If we move that way we can use other load balancers, currently, we are load balancing at TCP level commonly named SSL Passtrough, and haproxy is the only one  (that I know) that allows to do health checking on a HTTPS endpoint.

If we are able to get the certificates to the load balancer we can start doing load balancer at L7

images/haproxy/push-cross.sh
BenTheElder: Oh! Because if an image build fails we leak the manifest. It only removed the manifest with the -p/--purge on pushing the manifest. We don't want to increase the chance of having a lingering manifest. You have to literally find and remove the files on disk in what I think is an undocumented directory.

BenTheElder: So we only build the manifest after all the images are good to go.

aojea: Got it , thanks

images/haproxy/push-cross.sh
BenTheElder: I confirmed that all the other kubernetes/kubernetes images only have v6 and v8, we can revisit later if that changes.

BenTheElder:
that build script is obviously not ideal and duplicates effort from kindnetd, but I'll revisit it later, it functions well and we don't need to use it often.

I've tested locally and it seems to behave well.

PR #646, opened 2019-06-24
WIP Move CNI configuration into kindnetd

aojea:
Just an example to use in #633


BenTheElder:
thanks
/close
will rework #633 and address comments, we should get that into 0.4

PR #649, opened 2019-06-24
Fix creating bindir

mistydemeo:
Right now, if the `INSTALL_DIR` doesn't exist, `install` will error out with an error message. This could happen if, for example, a package manager like Homebrew was trying to install the `kind` binary to a binary path that won't exist at the start of the installation process. I've added a call to `install -d` beforehand to ensure it gets created.

I've also made the `install` tool configurable; this is pretty common for people who want to specify an alternate binary, for example to switch between BSD and GNU bintools.


mistydemeo:
Thanks!

PR #653, opened 2019-06-25
Remove leftover on CoreDNS IPv6 config

aojea:
The rewrite config was used on my CircleCI setup to fake the google.com service so some tests don´t fail

https://github.com/aojea/kubernetes/blob/circleci_master/.circleci/fake-google-com.yaml

### DIFF
--- a/hack/ci/e2e.sh
+++ b/hack/ci/e2e.sh
@@ -144,7 +144,6 @@ data:
     .:53 {
         errors
         health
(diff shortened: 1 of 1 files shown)


aojea:
/hold
shouldn´t affect the ipv6 tests at atll, but better be cautious

aojea:
/hold cancel

only 7 tests failing ... and counting

Issue #658, opened 2019-06-26
Question: will kind delete unused images?


WeihanLi:
Will kind remove unused images? If not, can I delete the image manually?

**What should be cleaned up or changed**:

unused docker images when the kubernetes does not need that image.

for example, deploy with `revisionHistoryLimit: 0`, when new image used, the before image should be deleted

**Why is this needed**:

if not clean these unused docker images, the disk resource needed will increase crazily

BenTheElder:
If you delete a kind cluster all resources on the host are released including disks space used by the nodes.

WeihanLi:
I wanna delete those unused images without delete the kind cluster, I tried to `docker exec kind-control-plane docker images`, but got an error like  unknown command docker

``` bash
OCI runtime exec failed: exec failed: container_linux.go:348: starting container process caused "exec: \"docker\": executable file not found in $PATH": unknown
```

BenTheElder:
> I wanna delete those unused images without delete the kind cluster, I tried to docker exec kind-control-plane docker images, but got an error like unknown command docker

Ah. So imageGC is disabled by default with kind but you could turn it on with a kubeadm config patch.
The reason this is off is to avoid GCing the core images under disk pressure and because we don't know how much disk you intend to keep free on the host. 

For a production cluster with a dedicated VM / machine this would be a terrible idea, but for kind we expect that you'll run this on your workstation and it will see how much space is free there, which may be some much lower percentage than the eviction threshold for a production setup.

We will probably ease this at some point to some non-zero but small threshold (?) ... it needs thought and experimentation / data.

Now as to why `docker exec kind-control-plane docker images` didn't work, confusingly around kind `v0.3.0` we switched to using CRI (contianerd) inside the nodes for a few reasons, so instead you need `docker exec kind-control-plane crictl inspecti`. We have it already configured so you don't need to specify the CRI endpoint.

More docs for crictl here 😅 https://github.com/kubernetes-sigs/cri-tools/blob/master/docs/crictl.md

WeihanLi:
OK, Thanks @BenTheElder

BenTheElder:
Is this an issue you're actively encountering or a hypothetical? image content storage is somewhat deduped but...

Ideally using kind you never _need_ to `docker exec` yourself, but it is of course available for debugging etc. 😅

WeihanLi:
[...]

PR #663, opened 2019-06-26
doc: configuring kubelet garbage collection

tao12345666333:
xref: #658

### DIFF
--- a/site/content/docs/user/quick-start.md
+++ b/site/content/docs/user/quick-start.md
@@ -389,6 +389,30 @@ The structure of the logs will look more or less like this:
 The logs contain information about the Docker host, the containers running 
 kind, the Kubernetes cluster itself, etc.
 
+### Configuring kubelet Garbage Collection in Your Cluster
+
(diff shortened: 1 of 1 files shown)


site/content/docs/user/quick-start.md
```diff
@@ -363,6 +363,30 @@ The structure of the logs will look more or less like this:
 The logs contain information about the Docker host, the containers running 
 kind, the Kubernetes cluster itself, etc.
 
+### Configuring kubelet Garbage Collection in Your Cluster
```
BenTheElder: I can see documenting this being useful, but this is pretty far into "quick-start has way too much in it"

tao12345666333: yes. I have the same opinion as @aojea . https://github.com/kubernetes-sigs/kind/pull/572#discussion_r290599547

> we should first thing how do we want to reorganize the docs, seems that they are starting to grow and I think that adding an advanced section with this stuff is not a good solution because we are going to have the same problem after some time.

tao12345666333: in addition to this PR, I see that we have some documentation on how to use kind in CI or some extra advanced features, which may need to be reorganized.

BenTheElder: kind in CI is now in the resources doc

I also don't think this is safe to enable in kind. it can evict the core images.

BenTheElder:
so I forgot to follow up here but I've followed up more with SIG-node about this.

currently kubernetes has a gap in this -- airgapped images that are sideloaded (like kind's) are image-gc unsafe.

we should provide a different option, I will think on it some more, but this is too much of a footgun for users.

Issue #662, opened 2019-06-26
`kind` should be able to start the control plane when `/dev/kmsg` is not available


hoegaarden:
**What happened**:

kind is not able to start the control plane (with the provided node images) when `/dev/kmsg` is not available in the node container.

kind then fails with:
```
 ✗ Starting control-plane 🕹️
Error: failed to create cluster: failed to init node with kubeadm: exit status 1
```

Checking on the node, I can see the following error:

```
# docker exec -ti kind-control-plane sh -c 'journalctl -u kubelet | grep -m1 -C3 kmsg'
Jun 26 08:18:59 kind-control-plane kubelet[375565]: I0626 08:18:59.522248  375565 kuberuntime_manager.go:205] Container runtime containerd initialized, version: 1.2.6-0ubuntu1, apiVersion: v1alpha2
Jun 26 08:18:59 kind-control-plane kubelet[375565]: I0626 08:18:59.525786  375565 server.go:1083] Started kubelet
Jun 26 08:18:59 kind-control-plane kubelet[375565]: I0626 08:18:59.526254  375565 server.go:144] Starting to listen on 0.0.0.0:10250
Jun 26 08:18:59 kind-control-plane kubelet[375565]: F0626 08:18:59.532066  375565 kubelet.go:1405] Failed to start OOM watcher open /dev/kmsg: no such file or directory
Jun 26 08:18:59 kind-control-plane kubelet[375565]: I0626 08:18:59.535478  375565 server.go:350] Adding debug handlers to kubelet server.
Jun 26 08:18:59 kind-control-plane systemd[1]: kubelet.service: Main process exited, code=exited, status=255/EXCEPTION
Jun 26 08:18:59 kind-control-plane systemd[1]: kubelet.service: Failed with result 'exit-code'.
```

For the node images I bake myself, I have the following [dirty hack] in place:
- I create the image with `kind build node-image ...`
- I add another layer which adds a systemd file which sets a `ExecStartPre` for the kubelet which just does something like `[ -e /dev/kmsg ] || ln -s /dev/console /dev/kmsg`

After that `kind create cluster ...` successfully starts the control plane.

[dirty hack]: https://github.com/pivotal-k8s/kind-on-c/blob/4f84e4ac28fa4e2e337190494aeaba89a55a94fe/entrypoint.sh#L197-L215

**What you expected to happen**:

kind should be able to bring up the control plane, at least when `/dev/console` is available.

[...]

PR #664, opened 2019-06-26
Link `/dev/kmsg` to `/dev/console`

hoegaarden:
... in the base image's entrypoint.

For environments where `/dev/kmsg` is not available, the kubelet fails
to start because the OOM parser cannot open that file, users see the
following error:

```
Jun 26 08:18:59 kind-control-plane kubelet[375565]: F0626 08:18:59.532066  375565 kubelet.go:1405] Failed to start OOM watcher open /dev/kmsg: no such file or directory
```

We therefore link `/dev/kmsg` to `/dev/console` if the former is missing
but the later is available.

Issue: https://github.com/kubernetes-sigs/kind/issues/662

### DIFF
--- a/images/base/entrypoint
+++ b/images/base/entrypoint
@@ -59,10 +59,22 @@ fix_product_name() {
   fi
 }
 
+link_kmsg() {
+  # In environments where /dev/kmsg is not available, the kubelet (1.15+) won't
+  # start because it cannot open /dev/kmsg when starting the kmsgparser in the
+  # OOM parser.
+  # To support those environments, we link /dev/kmsg to /dev/console.
+  # https://github.com/kubernetes-sigs/kind/issues/662
+  if [[ ! -e /dev/kmsg ]] && [[ -e /dev/console ]]; then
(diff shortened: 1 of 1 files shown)


pkg/build/node/systemd.go
```diff
@@ -23,6 +23,7 @@ Description=kubelet: The Kubernetes Node Agent
 Documentation=http://kubernetes.io/docs/
 
 [Service]
+ExecPreStart=/bin/sh -c '[ -e /dev/console ] && [ ! -e /dev/kmsg ] && ln -s /dev/console /dev/kmsg'
```
BenTheElder: what about doing this in the entrypoint in the base image?

BenTheElder: we have a pattern for this sort of workaround here: https://github.com/kubernetes-sigs/kind/blob/master/images/base/entrypoint

hoegaarden: Updated, moved that to the base image's entrypoint. /ptal

Issue #666, opened 2019-06-27
Failed to create cluster from the kind-example-config.yaml


LiliC:
**What happened**:
```
kind create cluster --config multinode.yaml
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.15.0) 🖼
 ✓ Preparing nodes 📦📦📦📦
 ✗ Creating kubeadm config 📜
Error: failed to create cluster: failed to generate kubeadm config content: failed to find an object with kubeadm.k8s.io_v1beta1_ClusterConfiguration|config to apply the patch
```

```
cat multinode.yaml
# this config file contains all config fields with comments
kind: Cluster
apiVersion: kind.sigs.k8s.io/v1alpha3
# patch the generated kubeadm config with some extra settings
kubeadmConfigPatches:
- |
  apiVersion: kubeadm.k8s.io/v1beta1
  kind: ClusterConfiguration
  metadata:
    name: config
  networking:
    serviceSubnet: 10.0.0.0/16
# patch it further using a JSON 6902 patch
kubeadmConfigPatchesJson6902:
- group: kubeadm.k8s.io
  version: v1beta1
  kind: ClusterConfiguration
  patch: |
    - op: add
      path: /apiServer/certSANs/-
      value: my-hostname
# 1 control plane node and 3 workers
nodes:
# the control plane node config
- role: control-plane
# the three workers
- role: worker
- role: worker
- role: worker
```

**What you expected to happen**:

Have a multi node cluster up and running created with kind. 

**How to reproduce it (as minimally and precisely as possible)**:

Use the above example which is suggested in the [docs](https://kind.sigs.k8s.io/docs/user/quick-start#configuring-your-kind-cluster).

**Anything else we need to know?**:

I noticed this issue came up before, not sure if tests already exist for this, but if not maybe there could be a test for the examples?

I removed the extra config patches and the cluster was created correctly.

**Environment:**

- kind version: (use `kind version`):
`v0.4.0`
- Kubernetes version: (use `kubectl version`):
```
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.0", GitCommit:"641856db18352033a0d96dbc99153fa3b27298e5", GitTreeState:"clean", BuildDate:"2019-03-26T00:04:52Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"darwin/amd64"}
[...]

PR #667, opened 2019-06-27
site/.../user/*: Update kubeadm API version

LiliC:
Update kubeadm APIVersion to kubeadm.k8s.io/v1beta2 in examples.

Closes https://github.com/kubernetes-sigs/kind/issues/666

### DIFF
--- a/site/content/docs/user/kind-example-config.yaml
+++ b/site/content/docs/user/kind-example-config.yaml
@@ -4,7 +4,7 @@ apiVersion: kind.sigs.k8s.io/v1alpha3
 # patch the generated kubeadm config with some extra settings
 kubeadmConfigPatches:
 - |
-  apiVersion: kubeadm.k8s.io/v1beta1
+  apiVersion: kubeadm.k8s.io/v1beta2
   kind: ClusterConfiguration
   metadata:
     name: config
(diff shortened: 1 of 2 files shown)


tao12345666333:
Thanks.

/ok-to-test

aojea:
I think that this only applies to master and k8s versions > 1.15, should we add a note mentioning it? The relationship between kubeadm API version and k8s versions is described in https://github.com/kubernetes-sigs/kind/blob/48f2b327bf69df8a0991ac2ccd13f6d1a949e216/pkg/cluster/internal/kubeadm/config.go#L452

LiliC:
> I think that this only applies to master and k8s versions > 1.15, should we add a note mentioning it? 

By default the 1.15 cluster is created, so not sure it's needed. But if yes where do we want to document that?

aojea:
No worries, just realized about this ... But maybe is not very common, anyway this lgtm, that can be done later if needed

BenTheElder:
the example should absolutely work with the latest default kind.

it is worth knowing that the patches do need to match currently.

looking at options to change how this works as this is something of a UX thorn.

I would not actually use this config either, it merely shows what fields are possible, generally setting all of them should not be required, we might want to clarify that somehow :sweat_smile:

Issue #668, opened 2019-06-28
/hack/ci/e2e.sh should support build new base image for arch other than x86


ZhengZhenyu:
**What would you like to be added**:
Allow /hack/ci/e2e.sh to check arch and build new base image if arch is not
x86.

**Why is this needed**:
The default base image is built for x86 arch, which makes it impossible to
use this script directly for testing on other arch, adding this to allow
build a new base image from scrach will make the arch scope wider.

BenTheElder:
So the scripts in hack/ are not really meant to be reused and are not stable, I think this is probably reasonable but it might be better to copy this and maintain a version that does..

I'm pretty wary that we will change or remove this and break your use case. This script is meant to be a stopgap until test-infra's tools improve (kubetest / kubetest2) and have better integration with kind.

ZhengZhenyu:
@BenTheElder Ah, thanks I was putting up a patch on this, as you know, I have been working on KinD on ARM for quite a while, I did have some findings now, and I will reported latter. I have refactored the test cases in openlab CI, and now we are using this script directly for test, since our system uses zuul which uses ansible, for this kind of script, we have to save it to somewhere and download it and we cannot find a suitable place for it. And besides, using this directly will be able to make the tests on ARM and amd be the same, which seems better? And I can report if there is something broken because of changes on this file and during the meantime, I will think of better choices.

aojea:
I thought that we were building node and base images for most common architectures : thinking

BenTheElder:
>  have refactored the test cases in openlab CI, and now we are using this script directly for test, since our system uses zuul which uses ansible, for this kind of script, we have to save it to somewhere and download it and we cannot find a suitable place for it. And besides, using this directly will be able to make the tests on ARM and amd be the same, which seems better? And I can report if there is something broken because of changes on this file and during the meantime, I will think of better choices.

The issue is that we will stop using this at some point it was never supposed to be a permanent feature, just something used to bootstrap the project. And if users need some complex script that is too difficult to replicate to use kind then we are failing. :thinking: 

---

> I thought that we were building node and base images for most common architecture thinking

not node and base, they're much more painful to do this for and much slower. we will in the future.

BenTheElder:
I am going to replace this script shortly with a much better tool, I'll be sure to include this.

BenTheElder:
[...]

PR #669, opened 2019-06-28
Make e2e.sh build new base image if arch is not amd64

ZhengZhenyu:
As the default base image was built on amd64 platform, makes it
unable to run on other platforms, this change make the script
detect arch and build a new base image from scratch if arch is not
amd64.

Related-Bug: kubernetes-sigs/kind#668

### DIFF
--- a/hack/ci/e2e.sh
+++ b/hack/ci/e2e.sh
@@ -64,7 +64,7 @@ build() {
         /usr/local/bin/create_bazel_cache_rcs.sh || true
     fi
 
-    ARCH=$(dpkg --print-architecture)
+    ARCH=$(go env GOARCH)
     
     if [[ "${ARCH}" == "amd64" ]]; then
         # build the node image w/ kubernetes
--- a/hack/ci/e2e.sh
+++ b/hack/ci/e2e.sh
@@ -64,11 +64,20 @@ build() {
         /usr/local/bin/create_bazel_cache_rcs.sh || true
     fi
 
-    # build the node image w/ kubernetes
-    # TODO(bentheelder): remove the kube-root flag after we make kind try to
-    # find this in a go module compatible way
-    kind build node-image --type=bazel \
-        --kube-root="$(go env GOPATH)/src/k8s.io/kubernetes"
+    ARCH=$(dpkg --print-architecture)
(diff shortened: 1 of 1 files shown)


hack/ci/e2e.sh
```diff
@@ -64,11 +64,20 @@ build() {
         /usr/local/bin/create_bazel_cache_rcs.sh || true
     fi
 
-    # build the node image w/ kubernetes
-    # TODO(bentheelder): remove the kube-root flag after we make kind try to
-    # find this in a go module compatible way
-    kind build node-image --type=bazel \
-        --kube-root="$(go env GOPATH)/src/k8s.io/kubernetes"
+    ARCH=$(dpkg --print-architecture)
```
aojea: Should we use go to detect the architecture? Dpkg sounds very specific

ZhengZhenyu: Sure, I just copied this from the build base image docker file, I can change it

BenTheElder:
/ok-to-test
will need CLA to be sorted out, the Kubernetes orgs require this :grimacing:

hack/ci/e2e.sh
```diff
@@ -64,11 +64,20 @@ build() {
         /usr/local/bin/create_bazel_cache_rcs.sh || true
     fi
 
-    # build the node image w/ kubernetes
-    # TODO(bentheelder): remove the kube-root flag after we make kind try to
-    # find this in a go module compatible way
-    kind build node-image --type=bazel \
-        --kube-root="$(go env GOPATH)/src/k8s.io/kubernetes"
+    ARCH=$(go env GOARCH)
```
BenTheElder: we need to not fail if this fails, this script doesn't otherwise depend on go on the host.

BenTheElder: there's a script for this that the makefile uses

aojea: My fault, I suggested it because I thought go was the preferred way.

I think this is the file https://github.com/kubernetes-sigs/kind/blob/master/hack/build/goarch.sh

aojea:
/close

we should take another approach for multi-arch testing

Issue #680, opened 2019-07-03
Kind load command doesn't check whether the image digest is different from existing?


TommyLike:
**What happened**:

When the original image has been updated, after re-executing command `kind load docker-image xxxxxx` the images used in container keep unchanged.

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment:**

- kind version: 0.4.0:
- Kubernetes version: (use `kubectl version`):
- Docker version: (use `docker info`):
- OS (e.g. from `/etc/os-release`):

asitaru:
This is also happening to us after #382

aojea:
Upps , my fault , nasty bug , sorry
/assign

BenTheElder:
Ah this is a nasty bug :(
We should be checking and comparing the SHA, but it looks like we aren't.
/priority critical-urgent

BenTheElder:
this _should_ be fixed now and in the next release, thanks @aojea

BenTheElder:
this is still incorrect, finally got a chance to look at this.
/assign
/lifecycle active

PR #682, opened 2019-07-03
Load docker image if ID is different

aojea:
We should load the image on the cluster nodes if the ID is different, checking only by name is not enough

Fixes: #680

### DIFF
--- a/cmd/kind/load/docker-image/docker-image.go
+++ b/cmd/kind/load/docker-image/docker-image.go
@@ -72,14 +72,10 @@ func NewCommand() *cobra.Command {
 func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
 	imageName := args[0]
 	// Check that the image exists locally and gets its ID, if not return error
-	lines, err := docker.ImageInspect(imageName, "{{ .Id }}")
+	imageID, err := docker.ImageID(imageName)
 	if err != nil {
 		return errors.Errorf("Image: %q not present locally", imageName)
 	}
-	if len(lines) != 1 {
-		return errors.Errorf("Docker image ID should only be one line, got %d lines", len(lines))
-	}
-	imageID := lines[0]
 	// Check if the cluster name exists
 	known, err := cluster.IsKnown(flags.Name)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/kind/load/docker-image/docker-image.go
```diff
@@ -71,11 +71,15 @@ func NewCommand() *cobra.Command {
 
 func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
 	imageName := args[0]
-	// Check that the image exists locally, if not return error
-	_, err := docker.ImageInspect(imageName)
+	// Check that the image exists locally and gets its ID, if not return error
+	lines, err := docker.ImageInspect(imageName, "{{ .Id }}")
```
BenTheElder: we should really break this out into a higher level abstraction, random packages should not be passing docker format strings and then parsing the output.

can we have `docker.ImageId(imageName)` instead?

aojea: sure

BenTheElder: Thanks, it's not a huge deal now, but we will have to clean all of this up later, and I'd like to stop repeating this pattern as it was a mistake :^)

BenTheElder:
> pkg/container/docker/image.go:35:6: func ImageId should be ImageID
> cmd/kind/load/docker-image/docker-image.go:75:2: var imageId should be imageID

./shrug .. better appease golint :^)

cmd/kind/load/docker-image/docker-image.go
```diff
@@ -116,10 +116,10 @@ func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
 	// pick only the nodes that don't have the image
 	selectedNodes := []clusternodes.Node{}
 	for _, node := range candidateNodes {
-		_, err := node.ImageInspect(imageName)
+		_, err := node.ImageInspect(imageID)
```
BenTheElder: we should probably clean up this API too, but that can be a follow-up

BenTheElder:
bypassing broken master branch tests (upstream patch is pending)

PR #684, opened 2019-07-03
Add working CoreDNS config for IPv6 clusters

aojea:
CoreDNS needs a different configuration to run in the CI because of:
1. CI doesn´t provide IPv6 connectivity, thus we can´t use upstream DNS servers and CoreDNS fail to work. 
2. CI adds to the resolv.conf the following domains to `search`: `c.k8s-prow-builds.internal google.internal`. CoreDNS should handle those domains and answer with `NXDOMAIN` instead of `SERVFAIL` in order that the pod continuous querying the DNS server.

Fixes: https://github.com/kubernetes/kubernetes/issues/70248

### DIFF
--- a/hack/ci/e2e.sh
+++ b/hack/ci/e2e.sh
@@ -134,8 +134,25 @@ run_tests() {
     export KUBECONFIG
 
     if [[ "${IP_FAMILY:-ipv4}" == "ipv6" ]]; then
-        # Create the CoreDNS config for offline and IPv6 clusters
+        # IPv6 clusters need some CoreDNS changes in order to work in k8s CI:
+        # 1. k8s CI doesn´t offer IPv6 connectivity, so CoreDNS should be configured
+        # to work in an offline environment:
         # https://github.com/coredns/coredns/issues/2494#issuecomment-457215452
+        # 2. k8s CI adds following domains to resolv.conf search field :
+        # c.k8s-prow-builds.internal google.internal.
+        # CoreDNS should handle those domains and answer with NXDOMAIN instead of SERVFAIL
+        # otherwise pods stops trying to resolve the domain.
+        # The difference against the default CoreDNS config in k8s 1.15 is:
+        # <         kubernetes cluster.local in-addr.arpa ip6.arpa {
+        # ---
+        # >         kubernetes cluster.local internal in-addr.arpa ip6.arpa {
+        # 9,10d9
+        # <            upstream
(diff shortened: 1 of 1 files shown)


hack/ci/e2e.sh
```diff
@@ -142,13 +142,18 @@ apiVersion: v1
...
+        log
         errors
         health
         kubernetes cluster.local in-addr.arpa ip6.arpa {
            pods insecure
+           fallthrough in-addr.arpa ip6.arpa
+           ttl 30
         }
         prometheus :9153
+        forward . /etc/resolv.conf
```
neolit123: hm, so this will forward everything using the host resolv.conf:
https://github.com/coredns/coredns/tree/master/plugin/forward#examples

why is it needed for ipv6?

aojea: this is for debugging the CI failures, I just took the original CoreDNS configuration and removed the `upstream` field, will try to change only one thing at a time to try to understand why the DNS tests are failing in the CI and working on my host
```
apiVersion: v1
data:
  Corefile: |
    .:53 {
        errors
        health
        kubernetes cluster.local in-addr.arpa ip6.arpa {
           pods insecure
           upstream
           fallthrough in-addr.arpa ip6.arpa
           ttl 30
        }
        prometheus :9153
        forward . /etc/resolv.conf
        cache 30
        loop
        reload
        loadbalance
    }
kind: ConfigMap
metadata:
  creationTimestamp: "2019-07-03T16:30:31Z"
  name: coredns
  namespace: kube-system
```

neolit123:
it would be actually great if we start adding some annotations for those failing sig-network/dns tests, instead of `timed out waiting for the condition`.

aojea:
hehe, I´ve found the problem

aojea:
@neolit123 @BenTheElder IPv6 is green 😄

neolit123:
nice!

hack/ci/e2e.sh
```diff
@@ -144,7 +146,7 @@ data:
     .:53 {
         errors
         health
-        kubernetes cluster.local in-addr.arpa ip6.arpa {
+        kubernetes cluster.local internal in-addr.arpa ip6.arpa {
```
BenTheElder: can we comment concretely on what the diff is vs the default config?

aojea: done

PR #685, opened 2019-07-04
[WIP] enable concurrent join of both worker and control-plane nodes

neolit123:
(first commit fixes some typos)

concurrent join of CP nodes should be now supported in k/k master for kubeadm.
no test signal in kinder yet though.

this change works for me, but oddly, i don't see much improvements in performance by using `time kind create cluster...` with a `3CPx3W` cluster. it's roughly 10 seconds reduced from 2 minutes 40 seconds without the change, but the numbers vary a lot for me.

i'm noticing that once `kind create cluster` finishes most pods (e.g 24 of 26) are already running for this 6 node setup, which should not be the case - most of the pods should be still creating once the kind binary exists, unless this is desired?

do we have sync points or "wait for node/pods ready" logic in kind?

cc @BenTheElder @aojea @fabriziopandini

### DIFF
--- a/pkg/internal/cluster/create/actions/kubeadmjoin/join.go
+++ b/pkg/internal/cluster/create/actions/kubeadmjoin/join.go
@@ -35,10 +35,10 @@ import (
 )
 
 // Action implements action for creating the kubeadm join
-// and deployng it on the bootrap control-plane node.
+// and deploying it on the bootrap control-plane node.
 type Action struct{}
 
-// NewAction returns a new action for creating the kubeadm jion
+// NewAction returns a new action for creating the kubeadm join
 func NewAction() actions.Action {
 	return &Action{}
 }
@@ -121,7 +121,7 @@ func joinWorkers(
 	return nil
 }
 
-// runKubeadmJoinControlPlane executes kubadm join --control-plane command
+// runKubeadmJoinControlPlane executes kubeadm join --control-plane command
 func runKubeadmJoinControlPlane(
 	allNodes []nodes.Node,
 	node *nodes.Node,
--- a/pkg/internal/cluster/create/actions/kubeadmjoin/join.go
+++ b/pkg/internal/cluster/create/actions/kubeadmjoin/join.go
@@ -55,21 +55,14 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 	if err != nil {
 		return err
 	}
-	if len(secondaryControlPlanes) > 0 {
-		if err := joinSecondaryControlPlanes(
-			ctx, allNodes, secondaryControlPlanes,
-		); err != nil {
-			return err
-		}
-	}
 
 	// then join worker nodes if any
 	workers, err := nodes.SelectNodesByRole(allNodes, constants.WorkerNodeRoleValue)
 	if err != nil {
 		return err
 	}
-	if len(workers) > 0 {
-		if err := joinWorkers(ctx, workers); err != nil {
+	if len(secondaryControlPlanes) > 0 || len(workers) > 0 {
+		if err := joinNodes(ctx, allNodes, secondaryControlPlanes, workers); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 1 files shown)


pkg/cluster/internal/create/actions/kubeadmjoin/join.go
```diff
@@ -55,58 +55,46 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
...
 	workers []nodes.Node,
 ) error {
-	ctx.Status.Start("Joining worker nodes 🚜")
+	startMsg := "Joining more "
+	if len(secondaryControlPlanes) > 0 && len(workers) > 0 {
+		startMsg += "control-plane and worker nodes 🎮🚜"
+	} else if len(secondaryControlPlanes) > 0 && len(workers) == 0 {
+		startMsg += "control-plane nodes 🎮"
+	} else {
+		startMsg += "worker nodes 🚜"
```
neolit123: let me know if the messages / icons need tweaks.

aojea:
I´m curious what´s going to be the job results on versions <1.14

> do we have sync points or "wait for node/pods ready" logic in kind?

I think that we have the `wait` flag:
```
cmd/kind/create/cluster/createcluster.go:       cmd.Flags().DurationVar(&flags.Wait, "wait", time.Duration(0), "Wait for control plane node to be ready
```

I use for CI environments `kubectl wait -n kube-system --timeout=360s --for condition=Ready -l  -l k8s-app=kube-dns pods`

neolit123:
> I´m curious what´s going to be the job results on versions <1.14

we technically have to enable this feature in kind only for kubeadm versions >= 1.15.

> I think that we have the wait flag:

yep, but unless i'm missing something it if set to 0 it does not wait.
if set to > 0 it errors out if the cluster actions don't finish that time.

neolit123:
@aojea 

> pull-kind-conformance-parallel-1-14 — Job succeeded.

in my tests if etcd member addition magically aligns, things work fine without the latest changes in k/k.
xref:
- https://github.com/kubernetes/kubernetes/pull/79677
- https://gist.github.com/neolit123/4ecd657175f17df0a8207e7ffb2c0ec4

neolit123:
but i find it quite odd that all 12, 13 and 14 passed.
hm, is the join really concurrent even?

aojea:
hehehe, is too late for me, I just realized that the jobs in the CI only have one control plane 🙃 😫

BenTheElder:
[will come back to this, keeping low priority due to WIP..., @ me if you need something sooner.]

neolit123:
rebased the PR.

when doing the concurrent join it now takes <2minutes, for a 3x3 cluster, however i got it to flake 1 time out of 5 (or so), but i couldn't catch the debug output and real cause.

probably it needs better timeouts/retries for etcd.
will test this more.

pkg/internal/cluster/create/actions/kubeadmjoin/join.go
```diff
@@ -55,58 +55,45 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
...
+	} else if len(secondaryControlPlanes) > 0 && len(workers) == 0 {
+		startMsg += "control-plane nodes 🎮"
+	} else {
+		startMsg += "worker nodes 🚜"
+	}
+	ctx.Status.Start(startMsg)
 	defer ctx.Status.End(false)
 
-	// TODO(bentheelder): it's too bad we can't do this concurrently
-	// (this is not safe currently)
```
BenTheElder: this is only safe on recent versions, right?

neolit123: at least this change is needed, which was added in 1.15.1
https://github.com/kubernetes/kubernetes/commit/bc74ac3eaad642bc748c459b5776701f50c18094#diff-5e4c5bba67c635568de70e0c424882c6

neolit123:
> when doing the concurrent join it now takes <2minutes, for a 3x3 cluster, however i got it to flake 1 time out of 5 (or so), but i couldn't catch the debug output and the real cause.

i only got to flake 1/10 times. still no logs.

pkg/internal/cluster/create/actions/kubeadmjoin/join.go
BenTheElder: we should probably ship this as being concurrent only for >= 1.15.1 then ?

neolit123: yes, i guess there has to be a version check and the old way of serial join has to be supported too.

neolit123:
let's get back to this once the required changes in etcd land.
also the PR needs [branching per kubeadm version](https://github.com/kubernetes-sigs/kind/pull/685#discussion_r320540503).

/close

PR #686, opened 2019-07-05
Fix HTTPS_PROXY and pass through proxy vars if set

electrofelix:
Build script was looking for "${HTTPS_PROX:+}" instead of
"${HTTPS_PROXY:+}" to use resulting in failure to build from behind
proxies.

To prevent typos from occurring it is easier to simply ask Docker to
pass through environment variables that are set by using `-e VAR`
instead of `-e VAR=VALUE`.

### DIFF
--- a/hack/build/go_container.sh
+++ b/hack/build/go_container.sh
@@ -73,9 +73,9 @@ run_in_go_container() {
       -e GOOS="${GOOS}" \
       -e GOARCH="${GOARCH}" \
     `# pass through proxy settings from the host` \
-      -e HTTP_PROXY="${HTTP_PROXY:+}" \
-      -e HTTPS_PROXY="${HTTPS_PROX:+}" \
-      -e NO_PROXY="${NO_PROXY:+}" \
+      -e HTTP_PROXY \
+      -e HTTPS_PROXY \
+      -e NO_PROXY \
     `# run as if the host user for consistent file permissions` \
       --user "${_UID}:${_GID}" \
     `# use the golang image for the container` \


aojea:
@electrofelix good catch, but I think that we should keep the format `-e HTTP_PROXY="${HTTP_PROXY:+}"` 
What happens if the variable is unset and we use only `-e HTTP_PROXY` ? Also, I think that `shellcheck` will fail with an `unbound variable` error

electrofelix:
@aojea Docker supports two formats for passing environment variables to the run command, see https://docs.docker.com/engine/reference/commandline/run/#set-environment-variables--e---env---env-file

The first is the same as what was currently used:
```
-e VAR=VALUE
```

The second is arguably what the current script was duplicating:
```
-e VAR
```

This form just tells docker the name of an environment variable that may or may not be exported within the current user environment. If it is set and exported, docker will add it to the container runtime environment using the value from the users environment. If it is not set, it simply be ignored for the launched container environment.

In fact the second format could probably be used for `GOOS`, `GOARCH`, & `GOPROXY` as you only need the format `VAR=${VAR1:+}` when the name of the variable in the local environment doesn't match the name of the environment variable you wish to set for the container runtime.

It seems the script was trying to workaround something that can be handled directly by the docker client.

aojea:
@electrofelix thanks for the explanation, anyway @BenTheElder is the expert here, let´s see what´s his opinion :)

/ok-to-test

hack/build/go_container.sh
```diff
@@ -73,9 +73,9 @@ run_in_go_container() {
       -e GOOS="${GOOS}" \
       -e GOARCH="${GOARCH}" \
     `# pass through proxy settings from the host` \
-      -e HTTP_PROXY="${HTTP_PROXY:+}" \
-      -e HTTPS_PROXY="${HTTPS_PROX:+}" \
```
BenTheElder: Nice catch on this typo!
We still need a proxied test-env to try these things out in...

Issue #688, opened 2019-07-05
Containerd is not working behind a proxy


aojea:
**What happened**:

Containerd is not able to pull images if is working behind a proxy

**What you expected to happen**:

containerd should work behind a proxy

**How to reproduce it (as minimally and precisely as possible)**:

https://kubernetes.slack.com/archives/CEKK1KTN2/p1562335808184900

**Anything else we need to know?**:

The problem  is that we removed `/etc/systemd/system/containerd.service.d/http-proxy.conf`

when we merged this https://github.com/kubernetes-sigs/kind/pull/628 

**Environment:**

- kind version: (use `kind version`): > 0.4.0
- Kubernetes version: (use `kubectl version`):
- Docker version: (use `docker info`):
- OS (e.g. from `/etc/os-release`):

aojea:
as a workaround 
```
[Service]    
Environment="HTTP_PROXY=${HTTP_PROXY:-}"    
Environment="HTTPS_PROXY=${HTTPS_PROXY:-}"    
Environment="NO_PROXY=${NO_PROXY:-localhost},${LOCAL_NETWORK}"    
EOF
```

@BenTheElder should we recover the `images/base/entrypoint` ?

aojea:
Fixed by #694

PR #690, opened 2019-07-05
Revert "Remove entrypoint configuration workaround."

aojea:
Containerd needs to set the proxy variables using systemd configuration files

This reverts commit 5f43b701879bbf5519dacb2ae5546caede937e33.

Fixes: #688

### DIFF
--- a/images/base/entrypoint
+++ b/images/base/entrypoint
@@ -18,6 +18,19 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+configure_proxy() {
+  # Don't use proxy for the local network
+  LOCAL_NETWORK=$(ip route list dev eth0 proto kernel | awk '{ print $1 }')
+
+  mkdir -p /etc/systemd/system/containerd.service.d/
+  cat <<EOF >/etc/systemd/system/containerd.service.d/http-proxy.conf
+[Service]
+Environment="HTTP_PROXY=${HTTP_PROXY:-}"
+Environment="HTTPS_PROXY=${HTTPS_PROXY:-}"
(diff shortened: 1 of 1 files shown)


images/base/entrypoint
```diff
@@ -18,6 +18,19 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+configure_proxy() {
+  # Don't use proxy for the local network
+  LOCAL_NETWORK=$(ip route list dev eth0 proto kernel | awk '{ print $1 }')
```
BenTheElder: we don't need this

images/base/entrypoint
```diff
@@ -18,6 +18,19 @@ set -o errexit
...
+
+  mkdir -p /etc/systemd/system/containerd.service.d/
+  cat <<EOF >/etc/systemd/system/containerd.service.d/http-proxy.conf
```
BenTheElder: we should really be configuring more than just containerd, working on an alternate fix

images/base/entrypoint
aojea: don´t we? I think we have to add the docker network to NO_PROXY

images/base/entrypoint
aojea: agree

aojea:
Ben is working in a better fix

images/base/entrypoint
BenTheElder: Kind computes this and adds it

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/694

PR #691, opened 2019-07-06
WIP Add service LoadBalancer to kindnetd

aojea:
I just stumbled onto this https://github.com/elsonrodriguez/minikube-lb-patch/blob/master/main.go
and seems a good addition to kind, especially for CI testing

### DIFF
--- a/cmd/kindnetd/main.go
+++ b/cmd/kindnetd/main.go
@@ -94,6 +94,29 @@ func main() {
 			panic(err.Error())
 		}
 
+		// Gets the Services information from API
+		// TODO: use a proper controller instead
+		svc, err := clientset.CoreV1().Services("").List(metav1.ListOptions{})
+		if err != nil {
+			panic(err.Error())
+		}
+
+		// add the Cluster IP to the LoadBalancer
+		for _, svc := range svc.Items {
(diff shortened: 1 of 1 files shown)


tao12345666333:
SGTM 👍

aojea:
@BenTheElder I think that this can be useful but don't know what will be the right place to add it.
I have mixed feelings adding it to kind, maybe as an addon? kindLB?

BenTheElder:
I'm not sure how useful this is tbh. So you just set the LB IP to the in cluster IP?

aojea:
I´m assuming that you are using a test client from the same docker network, using the cluster IP you can reach the services, can't you?

BenTheElder:
No? clusterIP == one of the service IPs IE one of the VIPs by kube-proxy, not the node IPs.

You can only reach these from within the cluster with a typical cluster configuration.

aojea:
you are right, we should have a node ip in linux or localhost in mac-windows

Once I have time will iterate over this, I still think it can be useful for testing applications that use LBs

aojea:
Don't think kindnet should implement this, it can be very easily added as an addon

BenTheElder:
The real problem is that you can't just set it to localhost because localhost is local to each network namespace.

To do this right you have to do something like add an IP on a host interface and bind a listener to that.

BenTheElder:
But also yes, kindnetd might not be the right place for this.

PR #692, opened 2019-07-07
Add dual stack support

aojea:
Add dual stack support to KIND, it also needs to add dual-stack support to KINDNET
Depends on:

- [x] https://github.com/kubernetes/kubernetes/pull/79033
- [x] https://github.com/kubernetes/kubernetes/pull/79386
- [x] https://github.com/kubernetes/kubernetes/pull/82462
- [x] https://github.com/kubernetes/kubernetes/pull/82473
- [x] https://github.com/kubernetes/kubernetes/pull/79993
- [x] https://github.com/kubernetes/kubernetes/pull/78801
- [x] https://github.com/kubernetes/kubernetes/pull/83123

New dependency for kindnet so we don't need annotations on the nodes:
- [x] https://github.com/kubernetes/enhancements/pull/1665


aojea:
cc: @vllry @khenidak @Arvinderpal @aramase

pkg/cluster/config/default.go
```diff
@@ -58,6 +58,9 @@ func SetDefaults_Cluster(obj *Cluster) {
 		if obj.Networking.IPFamily == "ipv6" {
 			obj.Networking.PodSubnet = "fd00:10:244::/64"
 		}
+		if obj.Networking.IPFamily == "ipv4-ipv6" {
+			obj.Networking.PodSubnet = "10.244.0.0/16,fd00:10:244::/64"
```
Arvinderpal: `fd00:10:244::/64` do you want to make this a /16 because currently the controller manager will default to /24 per node even for IPv6. This is a known problem.

pkg/cluster/config/v1alpha3/default.go
```diff
@@ -58,6 +58,9 @@ func SetDefaults_Cluster(obj *Cluster) {
 		if obj.Networking.IPFamily == "ipv6" {
 			obj.Networking.PodSubnet = "fd00:10:244::/64"
 		}
+		if obj.Networking.IPFamily == "ipv4-ipv6" {
+			obj.Networking.PodSubnet = "10.244.0.0/16,fd00:10:244::/64"
```
Arvinderpal: same here. make it `fd00::/16`

Arvinderpal:
@aojea how are you handling DNS/NAT64 in kind?

aojea:
> @aojea how are you handling DNS/NAT64 in kind?

not handling it, should we? conformance tests are passing without it, just modified CoreDNS to work without contacting upstream servers

Arvinderpal:
> > @aojea how are you handling DNS/NAT64 in kind?
> 
> not handling it, should we? conformance tests are passing without it, just modified CoreDNS to work without contacting upstream servers

Ok, but I suspect e2e networking tests like the following that [test external connectivity](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/network/networking.go#L55) from pods won't pass. In that case, we'll either need to use GUA addresses or some sort of NATing with ULA addresses.

aojea:
Full e2e working is the next milestone, we were able to have conformance working yesterday ... One step at s time :sweat_smile:

aojea:
hehe kindnet works with dual - stack :)

BenTheElder:
still waiting on upstream, right?

aojea:
> still waiting on upstream, right?

yeah, https://github.com/kubernetes-sigs/kind/pull/692#issue-295089375

I couldn't attend yesterday's sig-net meeting (🎃 👻 ),
 I've seen in the minutes that they've talked about how to handle node.status.addresses, but don't know the outcome. This is something we can use in kindnet so we don't have to label the nodes with the internal ip addresses.

Also, kube-proxy iptables PR is not merged, there is some discussion about an iptables rule that currently is managed by kubelet (kubelet is not dual stack, it only install rules in one ip family)

@thockin @khenidak you are going to talk in kubecon about k8s dual-stack, 
do you think that these pending issues will be solved by then so we can add it to kind?

aojea:
/cc @BenTheElder 

still some PR missing in k/k https://github.com/kubernetes-sigs/kind/pull/692#issue-295089375

but I'd like to have it working for kubecon, at least people can try it

aojea:
/test pull-kind-verify
:thinking: 

> go: finding github.com/BurntSushi/toml v0.3.1
> level=error msg="Timeout exceeded: try increase it by passing --timeout option"
> verifying generated ...

images/kindnetd/go.mod
```diff
@@ -4,26 +4,20 @@ go 1.13
...
-	k8s.io/apimachinery v0.0.0-20190404173353-6a84e37a896d
-	k8s.io/client-go v11.0.0+incompatible
-	k8s.io/klog v0.3.0
-	k8s.io/utils v0.0.0-20190506122338-8fab8cb257d5
-	sigs.k8s.io/yaml v1.1.0 // indirect
+	k8s.io/api v0.0.0-20190820101039-d651a1528133
+	k8s.io/apimachinery v0.0.0-20190823012420-8ca64af22337
+	k8s.io/client-go v0.0.0-20190620085101-78d2af792bab
+	k8s.io/klog v0.4.0
+	k8s.io/utils v0.0.0-20190809000727-6c36bc71fc4a
```
BenTheElder: what is all of this about?

BenTheElder:
> still some PR missing in k/k #692 (comment)
> but I'd like to have it working for kubecon, at least people can try it

if this is not ready yet, why can't they install from the PR?

this does not look ready / stable yet. I'm also concerned that the current kubeadm configmap approach is going to introduce new flakes.

images/kindnetd/go.mod
aojea: I just did `go mod tidy` :man_shrugging:

aojea:
> > still some PR missing in k/k #692 (comment)
> > but I'd like to have it working for kubecon, at least people can try it
> 
> if this is not ready yet, why can't they install from the PR?

bad worded, I've tried to mean that the PR should be working with master :sweat_smile: 

> 
> this does not look ready / stable yet. 

yeah, totally agree :man_shrugging: 

> I'm also concerned that the current kubeadm configmap approach is going to introduce new flakes.

it defaults to environment variables https://github.com/kubernetes-sigs/kind/pull/692/files#diff-9c552002efcf7083d6acacaf1ce9ae62R74

currently is the only way to get the different CIDRs configured

This is just a prototype for playing with the feature, I think that is far for merge into kind but is nice to have for people that want to play with the feature
/hold

pkg/apis/config/v1alpha4/default.go
```diff
@@ -62,6 +65,9 @@ func SetDefaultsCluster(obj *Cluster) {
 		if obj.Networking.IPFamily == "ipv6" {
 			obj.Networking.ServiceSubnet = "fd00:10:96::/112"
 		}
+		if obj.Networking.IPFamily == "DualStack" {
+			obj.Networking.ServiceSubnet = "10.96.0.0/12,fd00:10:96::/112"
```
song-jiang: @aojea I am currently working on Calico dual stack support.  Your `dualstack` branch works great. The only issue I came across is ServiceSubnet `fd00:10:96::/112` is overlapping with PodSubnet `fd00:10:244::/16`. This causes host accessing service ip failed to hit MASQ rule and the remote host saw service ip as source address. Using ServiceSubnet `fa00:10:96::/112` get this issue fixed.

aojea: yeah, good catch, this was an oversight on my part.
This was because of a limitation that was fixed now with this https://github.com/kubernetes/kubernetes/pull/79993

aojea: it should be working now, I've updated the PR

song-jiang: Hi @aojea , with your latest code, and kubernetes 1.17.0 node image, I hit an issue creating cluster

aojea: the API version has changed, I'm deploying it know with this
```
kind: Cluster
networking:
  ipFamily: DualStack
nodes:
- role: control-plane
- role: worker
- role: worker
kubeadmConfigPatches:
- |
  kind: ClusterConfiguration
  metadata:
    name: config
  featureGates:
    IPv6DualStack: true
```

song-jiang: Thanks! Sure, I will let you know the result today.

song-jiang: @aojea With API version update, I can create dual stack cluster correctly. Dual stack pod IP/service IP all works fine for me. ( I am using lastest Calico ). However, I can only see ipv4 address been picked up for node resources. The kind node got both ipv4 and ipv6 address assigned to eth0. Do you know what possible goes wrong?

aojea: @song-jiang thanks for confirming, appreciate your help.
The kubelet only reports one ip family without a cloud-controller, that's a known issue 
https://github.com/kubernetes/kubernetes/issues/42125
it's not clear what direction is going to take kubernetes regarding this, currently seems that the only solution is with an external cloud-controller or patching manually 🤷‍♂

song-jiang: @aeojea Good to know. Thanks!

aojea:
/hold cancel
and here we go
kindnet dual-stack is backward compatible, tested here https://github.com/kubernetes-sigs/kind/pull/828
/assign @BenTheElder

aojea:
/hold
I have to test different scenarios with dual stack locally

aojea:
/hold cancel
smoke tests work locally
ready to go

hack/ci/e2e-k8s.sh
```diff
@@ -80,6 +80,22 @@ nodes:
 - role: worker
 - role: worker
 EOF
+
+  if [ "${IP_FAMILY:-ipv4}" = "DualStack" ]; then
```
BenTheElder: we should probably not duplicate defaulting this

aojea:
/retest
fixed a bug in kindnet in the maskagent

aojea:
/hold
this has to be greeen https://github.com/kubernetes-sigs/kind/pull/1255

aojea:
/hold
we have to improve the UX
https://github.com/kubernetes-sigs/kind/pull/1263/files#r368252116

aojea:
Also, we shouldn't merge until the node IPs are dual stack or we have a better method than using annotations to get them

aojea:
ok, this is now up to date with latest kind versions, cc @danwinship

pkg/apis/config/v1alpha4/types.go
```diff
@@ -184,6 +184,8 @@ const (
 	IPv4Family ClusterIPFamily = "ipv4"
 	// IPv6Family sets ClusterIPFamily to ipv6
 	IPv6Family ClusterIPFamily = "ipv6"
+	// DualStackFamily sets ClusterIPFamily to DualStack
+	DualStackFamily ClusterIPFamily = "DualStack"
```
BenTheElder, line 188: hmm, yet `ipv6` is lowercase

pkg/cluster/internal/providers/docker/provision.go
```diff
@@ -122,7 +122,7 @@ func createContainer(args []string) error {
 }
 
 func clusterIsIPv6(cfg *config.Cluster) bool {
-	return cfg.Networking.IPFamily == "ipv6"
+	return cfg.Networking.IPFamily == "ipv6" || cfg.Networking.IPFamily == "DualStack"
```
BenTheElder, line 125: we should use the constants from config

site/content/docs/user/quick-start.md
```diff
@@ -377,6 +377,32 @@ networking:
...
+featureGates:
+  IPv6DualStack: true
+kubeadmConfigPatches:
+- |
+  apiVersion: kubeadm.k8s.io/v1beta2
+  kind: ClusterConfiguration
+  metadata:
+    name: config
+  featureGates:
+    IPv6DualStack: true
```
BenTheElder, line 403: drop this patch

site/content/docs/user/quick-start.md
```diff
@@ -377,6 +377,32 @@ networking:
...
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+networking:
+  ipFamily: DualStack
+nodes:
+- role: control-plane
+- role: worker
+- role: worker
+featureGates:
+  IPv6DualStack: true
```
BenTheElder, line 395: we should flip this feature gate automatically on kubernetes versions where it's gated and kind's networking.ipFamily has been set to dualstack

BenTheElder: (i.e. the user should not have to specify this field)

BenTheElder: there's an example of this for 1.11 and provisioning inside the kubeadm config package

images/kindnetd/cmd/kindnetd/main.go
```diff
@@ -127,43 +174,76 @@ func main() {
...
+		if nodeIPs.Has(hostIP) {
+			klog.Info("handling current node\n")
 			// compute the current cni config inputs
 			if err := cniConfig.Write(
 				ComputeCNIConfigInputs(node),
 			); err != nil {
 				return err
 			}
+			if ipFamily == DualStackFamily {
+				// update external IPs node annotations until #42125 is fixed
```
BenTheElder, line 195: surely we should be setting the standard field instead of a custom annotation.

pkg/apis/config/v1alpha4/types.go
aojea: we can always import the core k/ks APIs constants?

pkg/cluster/internal/providers/docker/provision.go
aojea: indeed, tech debt accumulating :shushing_face:

site/content/docs/user/quick-start.md
aojea: we are passing a bool with IPv6 , not the ipFamily, but we can add it before that (I guess)

pkg/internal/apis/config/default.go
```diff
@@ -41,31 +41,39 @@ func SetDefaultsCluster(obj *Cluster) {
 		SetDefaultsNode(a)
 	}
 	if obj.Networking.IPFamily == "" {
-		obj.Networking.IPFamily = "ipv4"
+		obj.Networking.IPFamily = IPv4Family
+	} else if obj.Networking.IPFamily == DualStackFamily {
+		obj.FeatureGates["IPv6DualStack"] = true
```
aojea, line 46: is ok defaulting it here?

aojea:
> W0428 22:33:02.289822     215 configset.go:202] WARNING: kubeadm cannot validate component configs for API groups [kubelet.config.k8s.io kubeproxy.config.k8s.io]

kubeadm has to enable its feature flag 🤔

neolit123:
> > W0428 22:33:02.289822     215 configset.go:202] WARNING: kubeadm cannot validate component configs for API groups [kubelet.config.k8s.io kubeproxy.config.k8s.io]
> 
> kubeadm has to enable its feature flag

no we adding this warning when we stopped importing the internal types of the components in question. we no longer validate if users feed e.g. a bogus kube-proxy config and users need to debug on component level (logs).

BenTheElder:
coming back to this after v0.8.0 is sorted. back on that once https://github.com/kubernetes/kubernetes/issues/90568 is resolved

site/content/docs/user/quick-start.md
aojea: kubeadm featuregates follow a different path, still needed

aojea: but I think we can remove it

pkg/cluster/internal/providers/docker/provision.go
```diff
@@ -134,7 +134,7 @@ func createContainer(args []string) error {
...
```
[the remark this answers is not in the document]
aojea: docker nodes does not have ipv6 enabed by default, so in dual-stack or ipv6 we have to enable
```
	// enable IPv6 if necessary
	if clusterIsIPv6(cfg) {
		args = append(args, "--sysctl=net.ipv6.conf.all.disable_ipv6=0", "--sysctl=net.ipv6.conf.all.forwarding=1")
	}
```

images/kindnetd/cmd/kindnetd/subnets.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// getNoMasqueradeSubnets tries to get the POD networks subnets to not Masquerade them
+// It returns an array of strings with the Cluster CIDR subnets
+// It can only obtain the POD subnet parameter from one place for consistency
+// The order is:
+// 1. POD_SUBNET environment variables
+// 2. Pod subnet value in kubeadm configmap
+// 3. Cluster CIDR value in kube-proxy configmap
+func getNoMasqueradeSubnets(clientset *kubernetes.Clientset) ([]string, []string) {
```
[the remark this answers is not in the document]
aojea: we run a goroutine with a kind of `IP Masquerade`agent in kindnet, 
https://kubernetes.io/docs/tasks/administer-cluster/ip-masq-agent/
so we don´t masquerade the traffic between pods, but it can receive more CIDR.
However, you are right, since this is only returning the cluster subnets

images/kindnetd/cmd/kindnetd/main.go
```diff
@@ -134,43 +181,76 @@ func main() {
...
+				// TODO: https://github.com/kubernetes/kubernetes/issues/42125
+				// in KIND we know that the outer interface is eth0 to find the external ips
+				// but we can be smarter and find the interface based on the HOST_IP
+				hostExternalIPs := getHostExternalIPs(hostIP)
```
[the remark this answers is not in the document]
aojea: yep

images/kindnetd/cmd/kindnetd/main.go
```diff
@@ -134,43 +181,76 @@ func main() {
...
+		if nodeIPv4 != "" && len(podCIDRsv4) > 0 {
+			if err := syncRoute(nodeIPv4, podCIDRsv4); err != nil {
+				return err
+			}
+		}
+		if nodeIPv6 != "" && len(podCIDRsv6) > 0 {
+			if err := syncRoute(nodeIPv6, podCIDRsv6); err != nil {
+				return err
+			}
+		}
```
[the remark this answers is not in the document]
aojea: we are not handling most errors in kindnetd, to try to keep it as simple as possible, it just simple keep retrying
We can take this approach because we have control of the network scenario with docker/podman, so if any node does not have the corresponding IP family address is mostly a docker/podman bug and more things will break at the deployment before kindnet.

pkg/cluster/internal/providers/docker/provision.go
```diff
@@ -134,7 +134,7 @@ func createContainer(args []string) error {
 }
 
 func clusterIsIPv6(cfg *config.Cluster) bool {
-	return cfg.Networking.IPFamily == "ipv6"
```
aojea, line 137: Dan was right, this is setting the API to listen in ::1 in a dual stack cluster, I have to fix it

BenTheElder:
Any updates on the dual stack issue post 1.19?
actually: let's discuss on the issue. leaving a note here anyhow to show @fejta-bot activity on the PR

images/kindnetd/cmd/kindnetd/main.go
```diff
@@ -134,27 +179,17 @@ func main() {
...
-
+		// first get this node's IPs
+		// we don't support more than one IP address per IP family for simplification
+		nodeIPs := internalIPs(node)
+		klog.Infof("Handling node with IPs: %v\n", nodeIPs)
 		// This is our node. We don't need to add routes, but we might need to
-		// update the cni config.
-		if nodeIP == hostIP {
-			klog.Infof("handling current node\n")
+		// update the cni config and "annotate" our external IPs
```
aojea, line 190: this is no longer needed, just forget to update the comment, but we no longer need to annotate IPs
This is going ready to merge soon :smile: @BenTheElder

pkg/internal/apis/config/default.go
BenTheElder: this isn't going to work if featureGates has gates set but dualstack isn't explicitly set.
also we need to set this gate only for certain k8s versions.

we should do the defaulting further down the stack, and it should be *defaulting*, if the user explicitly sets it to false then 🤷

pkg/build/nodeimage/const_cni.go
```diff
@@ -46,6 +46,12 @@ rules:
     verbs:
       - list
       - watch
+  - apiGroups:
```
BenTheElder, line 49: we don't need to be able to get _all_ configmaps..?

pkg/apis/config/v1alpha4/types.go
BenTheElder: without importing them, I think we should consider matching them anyhow.
we should not actually import them.

images/kindnetd/go.mod
```diff
@@ -1,6 +1,6 @@
 module sigs.k8s.io/kind/images/kindnetd
 
-go 1.13
+go 1.14
```
BenTheElder, line 3: this seems arbitrary and wrong.

images/kindnetd/cmd/kindnetd/subnets.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	// TODO: we have other fallback options for completeness including:
+	// - kube-apiserver flags
+	// - component config
+	// - defaults
+	return nil, nil
+}
+
+// splitCIDRs given a slice of strings with CIDRS it returns 2 slice of strings per IP family
```
BenTheElder, line 105: we should clarify that it's v4, v6

images/kindnetd/Dockerfile
```diff
@@ -14,10 +14,10 @@
 
 # first stage build kindnetd binary
 # NOTE: tentatively follow upstream kubernetes go version based on k8s in go.mod
-FROM golang:1.13
+FROM golang:1.15
```
BenTheElder, line 17: this should only be bumped if the k8s.io imports match and are using 1.15

images/kindnetd/cmd/kindnetd/cni.go
```diff
@@ -33,22 +33,33 @@ import (
...
 // ComputeCNIConfigInputs computes the template inputs for CNIConfigWriter
 func ComputeCNIConfigInputs(node corev1.Node) CNIConfigInputs {
-	podCIDR := node.Spec.PodCIDR
-	defaultRoute := "0.0.0.0/0"
 
-	if net.IsIPv6CIDRString(podCIDR) {
-		defaultRoute = "::/0"
+	defaultRoutes := []string{"0.0.0.0/0", "::/0"}
+	// check if is a DualStack cluster
+	if len(node.Spec.PodCIDRs) > 1 {
```
BenTheElder, line 46: this will be zero on older clusters? or some client side magic.

images/kindnetd/cmd/kindnetd/subnets.go
```diff
@@ -0,0 +1,116 @@
...
+func getClusterSubnets(clientset *kubernetes.Clientset) ([]string, []string) {
+	// check for environment variables (legacy)
+	podSubnetEnv := os.Getenv("POD_SUBNET")
+	if podSubnetEnv != "" {
+		podSubnetEnv = strings.TrimSpace(podSubnetEnv)
+		return splitCIDRs(strings.Split(podSubnetEnv, ","))
+	}
+
+	// try getting from kubeadm configmap
+	kubeadmSubnets := newKubeSubnets(kubeadmClusterCIDRRegex, "kubeadm-config", "kube-system")
```
BenTheElder, line 82: this method seems like a pretty bad API.
we should fetch the configmap exactly once and then inspect it for CIDR / CIDRs as a following step.

images/kindnetd/cmd/kindnetd/subnets.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func (ks *kubeSubnets) Get(clientset *kubernetes.Clientset) (string, error) {
+	cidrRegexp := regexp.MustCompile(ks.regex)
+	configMap, err := clientset.CoreV1().ConfigMaps(ks.namespace).Get(context.TODO(), ks.configmap, metav1.GetOptions{})
+	if err != nil {
+		println("ERROR: " + err.Error())
+		return "", err
+	}
+	for _, data := range configMap.Data {
```
BenTheElder, line 56: code from here down should just be a function of it's own.

images/kindnetd/cmd/kindnetd/subnets.go
```diff
@@ -0,0 +1,116 @@
...
+func newKubeSubnets(regex, configmap, namespace string) *kubeSubnets {
+	return &kubeSubnets{
+		regex:     regex,
+		configmap: configmap,
+		namespace: namespace,
+	}
+}
+
+func (ks *kubeSubnets) Get(clientset *kubernetes.Clientset) (string, error) {
+	cidrRegexp := regexp.MustCompile(ks.regex)
```
BenTheElder, line 50: code from here to next comment should just be a function.
also it should have retries with backoff.

images/kindnetd/cmd/kindnetd/subnets.go
```diff
@@ -0,0 +1,116 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/utils/net"
+)
+
+const (
+	kubeadmClusterCIDRRegex   = `\s+podSubnet: (.*)\n`
+	kubeproxyClusterCIDRRegex = `\s+clusterCIDR: (.*)\n`
+)
+
+type kubeSubnets struct {
```
BenTheElder, line 35: this type is superfluous

images/kindnetd/cmd/kindnetd/subnets.go
aojea: yep

images/kindnetd/go.mod
aojea: yep

pkg/cluster/internal/providers/docker/provision.go
BenTheElder: and podman?

images/kindnetd/cmd/kindnetd/main.go
BenTheElder: is it? commeent still mentions annotation

images/kindnetd/cmd/kindnetd/cni.go
aojea: since PodCIDRs is a []string , my understanding is that client-go will set the nil value if it does not exist in the server.
You know that @liggitt will illustrate us :smile:

images/kindnetd/Dockerfile
aojea:
```
	k8s.io/api v0.19.2
	k8s.io/apimachinery v0.19.2
	k8s.io/client-go v0.19.2
```

pkg/build/nodeimage/const_cni.go
aojea: done

pkg/cluster/internal/providers/docker/provision.go
aojea: ?? in which sense?

pkg/internal/apis/config/default.go
aojea: yeah, is bad shortcut

images/kindnetd/cmd/kindnetd/subnets.go
aojea: I honestly don't remember why I did it this way, but does not make sense

pkg/cluster/internal/providers/docker/provision.go
aojea: ahh, why I didn't change in podman yo mean

aojea: checked again, is ::1 in HA, I don't think that should be a problem right now, we can iterate later

pkg/internal/apis/config/default.go
aojea: > this isn't going to work if featureGates has gates set but dualstack isn't explicitly set.

why? this is a kubeadm feature-gate, not a kubernetes feature gate.

> need to set this gate only for certain k8s versions.

I assume that kubeadm will fail if the gate doesn't exist

aojea: actually, I think that this is correct, is setting the kubeadm feature gate to true if the IPFamily == DualStack.
If the user sets it to false, he is doing something contradictory, because he is telling kind to deploy a dual-stack cluster and configuring the gate to not do it.
I think that the problem here is the semantics of defaulting, but my logic is if we do this for podSubnets
```
	// default the pod CIDR
	if obj.Networking.PodSubnet == "" {
		obj.Networking.PodSubnet = "10.244.0.0/16"
		if obj.Networking.IPFamily == "ipv6" {
			// node-mask cidr default is /64 so we need a larger subnet, we use /56 following best practices
			// xref: https://www.ripe.net/publications/docs/ripe-690#4--size-of-end-user-prefix-assignment---48---56-or-something-else-
			obj.Networking.PodSubnet = "fd00:10:244::/56"
		}
	}
```

we should be able to do the same for dual-stack IPFamily, just the difference is that it implease setting the kubeadm feature-gate

aojea: scratch it, I'm more trying to find an excuse, this does not make sense here

aojea:
/test all
wasn't using the last kindnet image

pkg/internal/apis/config/validate.go
```diff
@@ -49,13 +50,15 @@ func (c *Cluster) Validate() error {
 		}
 	}
 
+	isDualStack := c.Networking.IPFamily == DualStackFamily
```
aojea, line 53: should we validate against the kubernetes version?

aojea:
@BenTheElder we should revisit this now the 1.21 will have dualstack enabled by default

pkg/internal/apis/config/validate.go
BenTheElder: config doesn't know the kubernetes version, so no.

we can check later, when we get nodes.

go.mod
```diff
@@ -14,5 +14,6 @@ require (
 	golang.org/x/sys v0.0.0-20210124154548-22da62e12c0c // indirect
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
 	k8s.io/apimachinery v0.20.2
+	k8s.io/utils v0.0.0-20210305010621-2afb4311ab10
```
BenTheElder, line 17: this is at least not a staging repo right? but also no versioned tags?

BenTheElder: we appear to only use two small methods. we should consider just copying those into an `internal` package.

pkg/apis/config/v1alpha4/types.go
BenTheElder: does this match?

pkg/build/nodeimage/const_cni.go
```diff
@@ -95,7 +95,7 @@ spec:
       serviceAccountName: kindnet
       containers:
       - name: kindnet-cni
-        image: kindest/kindnetd:v20210220-5b7e6d01
+        image: aojea/kindnetd:0.6.0
```
BenTheElder, line 98: we should continue to use date-sha and you should start pushing to the real registry (please ensure fully rebased before pushing though)

BenTheElder: this is the main blocker for merge I think. the other stuff is more reasonable to iterate in follow ups

BenTheElder:
generally lgtm. left comments.
in particular need to update the image, and I'm still dubious on the constant values.
we can iterate on the dependencies angle. I think we should cease importing k8s.io/* for trivial things, but we can handle that seperately.

pkg/apis/config/v1alpha4/types.go
aojea: there is not "dualstack" family in kubernetes, 
```
// IPFamily represents the IP Family (IPv4 or IPv6). This type is used
// to express the family of an IP expressed by a type (e.g. service.spec.ipFamilies).
type IPFamily string

const (
	// IPv4Protocol indicates that this IP is IPv4 protocol
	IPv4Protocol IPFamily = "IPv4"
	// IPv6Protocol indicates that this IP is IPv6 protocol
	IPv6Protocol IPFamily = "IPv6"
)

// IPFamilyPolicyType represents the dual-stack-ness requested or required by a Service
type IPFamilyPolicyType string

const (
	// IPFamilyPolicySingleStack indicates that this service is required to have a single IPFamily.
	// The IPFamily assigned is based on the default IPFamily used by the cluster
	// or as identified by service.spec.ipFamilies field
	IPFamilyPolicySingleStack IPFamilyPolicyType = "SingleStack"
	// IPFamilyPolicyPreferDualStack indicates that this service prefers dual-stack when
	// the cluster is configured for dual-stack. If the cluster is not configured
	// for dual-stack the service will be assigned a single IPFamily. If the IPFamily is not
	// set in service.spec.ipFamilies then the service will be assigned the default IPFamily
	// configured on the cluster
	IPFamilyPolicyPreferDualStack IPFamilyPolicyType = "PreferDualStack"
	// IPFamilyPolicyRequireDualStack indicates that this service requires dual-stack. Using
	// IPFamilyPolicyRequireDualStack on a single stack cluster will result in validation errors. The
	// IPFamilies (and their order) assigned  to this service is based on service.spec.ipFamilies. If
	// service.spec.ipFamilies was not provided then it will be assigned according to how they are
	// configured on the cluster. If service.spec.ipFamilies has only one entry then the alternative
	// IPFamily will be added by apiserver
	IPFamilyPolicyRequireDualStack IPFamilyPolicyType = "RequireDualStack"
)
```

aojea: our "dualstack" is different in the sense that here it means "configure my cluster to actually have dualstack services, ips, ..."
in upstream dualstack depends on how the user configure the cluster.

pkg/internal/apis/config/validate.go
aojea: kubeadm will fail then

aojea:
/test pull-kind-conformance-parallel-dual-stack-ipv4-ipv6


> generally lgtm. left comments.
> in particular need to update the image, and I'm still dubious on the constant values.
> we can iterate on the dependencies angle. I think we should cease importing k8s.io/* for trivial things, but we can handle that seperately.

modified constants and removed all the dependencies on k8s.io/utils

aojea:
we need this for the dualstack CI job 
https://github.com/kubernetes/test-infra/pull/21537

once it is in kind I need to remember to change all the kubernetes jobs

BenTheElder:
Antonio and I discussed this a bit and looked around quite a bit, and also discussed with the azure folks (since AKS supports this though without any high level config for it).

... There really isn't any standard way to refer to this. Kubernetes has a bunch of distinct concepts with different names surrounding dualstack and the broader networking community seems to use a variety of formats for referring to this.

https://kubernetes.io/docs/concepts/services-networking/dual-stack/

`dual` seems fine enough. I didn't intend for this to be a major sticking point, just a chance to find the least surprising / most obvious value for users before shipping the API, but that doesn't really seem to exist.

BenTheElder:
/ pull-kind-conformance-parallel-dual-stack-ipv4-ipv6

pkg/internal/apis/config/validate.go
BenTheElder: we can still check prior to kubeadm, but after creating node containers. this can be in a follow-up.

site/content/docs/user/quick-start.md
```diff
@@ -370,6 +370,22 @@ networking:
...
+You can run dual stack clusters using `kind`, on kubernetes versions +1.21
+but first you need to [enable ipv6 in your docker daemon][docker enable ipv6].
+
+```yaml
+# a dual-stack cluster
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+networking:
+  ipFamily: dual
+nodes:
```
BenTheElder, line 383: todo (follow-up): just show the minimal config relevant to get a dualstack cluster.

Also consider moving this entirely to the the configuration page, as we have a better configuration guide over there and this page is way too disorganized and lengthy already.

BenTheElder:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/692/pull-kind-conformance-parallel-dual-stack-ipv4-ipv6/1375693467413385216

🎉

BenTheElder:
/lgtm
/approve
remaining nits can be addressed in follow up. past time to ship this.

for anyone following along, this has been helped along by a number of other PRs like https://github.com/kubernetes-sigs/kind/pull/2043

PR #700, opened 2019-07-11
upgrade to go 1.12.7, cleanup build scripts a bit

BenTheElder:
- upgrade to go 1.12.7 (latest stable, some bug fixes)
- inline goos.sh and goarch.sh, simplify the logic, remove these scripts
- add more comments to everything


hack/build/go_container.sh
```diff
@@ -18,49 +18,81 @@
 set -o nounset
 set -o errexit
 
-# get and go to the repo root
-REPO_ROOT="${REPO_ROOT:-$(git rev-parse --show-toplevel)}"
+# get the repo root for defaulting OUT_DIR and SOURCE_DIR
+# we assume the repo root is two levels up from this script
+REPO_ROOT="${REPO_ROOT:-$(cd "$(dirname "$0")/../.." && pwd)}"
```
tao12345666333: IMO it is clearer to use git rev-parse --show-toplevel

BenTheElder: that doesn't work with a release zip archive though, but either way we don't use this code path as we always set it when calling this script. Instead we typically get the makefile's directory using make.

BenTheElder: I do usually prefer the git approach if I can reasonably expect the tool only needs to function at development time, but this is more portable and this tool is a user facing implementation detail of the portable build.

tao12345666333: I agree with you. 
From this perspective, we can merge it.

/lgtm

tao12345666333: I agree with you. 
From this perspective, we can merge it.

/lgtm

Issue #698, opened 2019-07-09
Increase number of e2e tests run


aojea:
**What would you like to be added**:

Run more e2e tests with kind

**Why is this needed**:

We are running only conformance tests, that´s about 210/4414

BenTheElder:
❤️ We definitely need this.
/priority important-longterm

To be clear: We're looking to do this mainly for testing Kubernetes proper, not necessarily the kind repo.

We also definitely aren't interested in all 4000 or so tests, many of those are provider (GCE, AWS) specific or arguably don't make sense at all (`[Feature:Dashboard]`) but certainly there are many tests beyond conformance that would be useful for testing Kubernetes, which will also give us some more CI signal...

BenTheElder:
Some work towards getting kind to run roughly what `pull-kubernetes-e2e-gce` runs:
- https://github.com/kubernetes/kubernetes/pull/80044
- https://github.com/kubernetes/kubernetes/pull/80329

aojea:
those patches were merged

BenTheElder:
/assign
/lifecycle active
More than 2x now

PR #701, opened 2019-07-11
[Only for testing] Run k/k PR presubmit tests with kind

aojea:
It adds an environment variable to the script that runs the e2e tests to allow to choose between
conformance tests or the e2e tests that run in a k/k  presubmit job (skipping the ones that have known issues with kind)

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-testing/e2e-tests.md

xref: #698

### DIFF
--- a/hack/ci/e2e.sh
+++ b/hack/ci/e2e.sh
@@ -221,6 +221,11 @@ EOF
         # [sig-autoscaling] ReplicationController light Should scale from 1 pod to 2 pods
         # It reuires metrics addon, ref: https://github.com/kubernetes/kubernetes/pull/79954
         SKIP="ReplicationController.light.Should.scale.from.1.pod.to.2.pods|${SKIP}"
+        if [[ "${IP_FAMILY:-ipv4}" == "ipv6" ]]; then
+            # TODO(aojea): The ipv6 job times out and have 520 failures at that time
+            # lets skip storage at this moment
+            SKIP="\\[sig-storage\\]|${SKIP}"
+        fi
     else
         SKIP="${SKIP:-}"
         FOCUS="${FOCUS:-"\\[Conformance\\]"}"
--- a/hack/ci/e2e.sh
+++ b/hack/ci/e2e.sh
@@ -187,6 +187,7 @@ EOF
         | grep -cv "node-role.kubernetes.io/master" \
     )"
 
+    KUBERNETES_PRESUBMIT_JOB=true
     # ginkgo regexes
     if [[ "${KUBERNETES_PRESUBMIT_JOB:-}" = true ]]; then
         # k/k presubmit jobs regex
--- a/hack/ci/e2e.sh
+++ b/hack/ci/e2e.sh
@@ -191,8 +191,8 @@ EOF
     # ginkgo regexes
     if [[ "${KUBERNETES_PRESUBMIT_JOB:-}" = true ]]; then
         # k/k presubmit jobs regex
-        SKIP="${SKIP:-"\\[Slow\\]|\\[Disruptive\\]|\\[Flaky\\]|\\[Feature:.+\\]"}"
-        FOCUS="${FOCUS:-""}"
+        SKIP="${SKIP:-"\\[Disruptive\\]|\\[Flaky\\]|\\[Feature:.+\\]"}"
+        FOCUS="${FOCUS:-"\\[Slow\\]"}"
         # skip tests that are not able to rung with kind
         # [sig-node] RuntimeClass should run a Pod requesting a RuntimeClass with a configured handler [NodeFeature:RuntimeHandler]
         # https://github.com/kubernetes/kubernetes/issues/80098
--- a/hack/ci/e2e.sh
+++ b/hack/ci/e2e.sh
@@ -221,11 +221,6 @@ EOF
         # [sig-autoscaling] ReplicationController light Should scale from 1 pod to 2 pods
         # It reuires metrics addon, ref: https://github.com/kubernetes/kubernetes/pull/79954
         SKIP="ReplicationController.light.Should.scale.from.1.pod.to.2.pods|${SKIP}"
-        if [[ "${IP_FAMILY:-ipv4}" == "ipv6" ]]; then
-            # TODO(aojea): The ipv6 job times out and have 520 failures at that time
-            # lets skip storage at this moment
-            SKIP="\\[sig-storage\\]|${SKIP}"
-        fi
     else
         SKIP="${SKIP:-}"
         FOCUS="${FOCUS:-"\\[Conformance\\]"}"
--- a/hack/ci/e2e.sh
+++ b/hack/ci/e2e.sh
@@ -188,8 +188,43 @@ EOF
     )"
 
     # ginkgo regexes
-    SKIP="${SKIP:-}"
-    FOCUS="${FOCUS:-"\\[Conformance\\]"}"
+    if [[ "${KUBERNETES_PRESUBMIT_JOB:-}" = true ]]; then
+        # k/k presubmit jobs regex
+        SKIP="${SKIP:-"\\[Slow\\]|\\[Disruptive\\]|\\[Flaky\\]|\\[Feature:.+\\]"}"
+        FOCUS="${FOCUS:-""}"
+        # skip tests that are not able to rung with kind
+        # [sig-node] RuntimeClass should run a Pod requesting a RuntimeClass with a configured handler [NodeFeature:RuntimeHandler]
+        # https://github.com/kubernetes/kubernetes/issues/80098
+        SKIP="RuntimeClass.should.run.a.Pod.requesting.a.RuntimeClass.with.a.configured.handler|${SKIP}"
+        # [sig-storage] In-tree Volumes [Driver: nfs]
+        # [sig-storage] PersistentVolumes NFS
+        # TODO: kind nodes can't mount nfs volumes: install apt install -y nfs-common in the base image, kernel version, ...??
+        SKIP="In-tree.Volumes.\\[Driver:.nfs\\]|PersistentVolumes.NFS|${SKIP}"
+        # [sig-auth] PodSecurityPolicy
+        # We are not enabling the security admission controler in kind
+        # However, if enabled, it breaks kindnet because it needs apparmor in the host
+        SKIP="PodSecurityPolicy|${SKIP}"
+        # [sig-network] Network should set TCP CLOSE_WAIT timeout
+        # [sig-node] Mount propagation should propagate mounts to the host
+        # Use ssh to login to the kind nodes TODO: change to docker exec commands, install ssh server, ...
+        SKIP="Network.should.set.TCP.CLOSE_WAIT.timeout|${SKIP}"
+        SKIP="Mount.propagation.should.propagate.mounts.to.the.host|${SKIP}"
+        # [sig-cli] Kubectl client [k8s.io] Simple pod should support exec through an HTTP proxy 
+        # Needs to use the internal API server address, it can be passed as a kubetest arg
+        SKIP="Simple.pod.should.support.exec.through.an.HTTP.proxy|${SKIP}"
+        # [sig-storage] In-tree Volumes [Driver: local][LocalVolumeType: block] [Testpattern: Pre-provisioned PV (default fs)]
+        # subPath should support existing directories when readOnly specified in the volumeSource
+        # requires mounting a loop device inside a container, ref: https://serverfault.com/q/701384
+        SKIP="subPath.should.support.existing.directories.when.readOnly.specified.in.the.volumeSource|${SKIP}"
+        # [sig-autoscaling] [HPA] Horizontal pod autoscaling (scale resource: CPU)
+        # [sig-autoscaling] ReplicationController light Should scale from 1 pod to 2 pods
+        # It reuires metrics addon, ref: https://github.com/kubernetes/kubernetes/pull/79954
+        SKIP="ReplicationController.light.Should.scale.from.1.pod.to.2.pods|${SKIP}"
+    else
+        SKIP="${SKIP:-}"
+        FOCUS="${FOCUS:-"\\[Conformance\\]"}"
+    fi
+
     # if we set PARALLEL=true, skip serial tests set --ginkgo-parallel
     PARALLEL="${PARALLEL:-false}"
     if [[ "${PARALLEL}" == "true" ]]; then


aojea:
hahaha , that was fast

> I0711 11:52:35.410] Ran 0 of 2162 Specs in 1.881 seconds
I0711 11:52:35.415] SUCCESS! -- 0 Passed | 0 Failed | 0 Pending | 2162 Skipped

aojea:
1.14 run took 35m54s
Pass 467/3586
Fail 185/3586 
Skip 2934/3586

----

Failure analysis

   1 [k8s.io]
   3 [sig-auth]
   2 [sig-autoscaling]
   1 [sig-cli]
   1 [sig-network]
 176 [sig-storage]
   1 [sig-ui]

cc: @BenTheElder

BenTheElder:
that dashboard test should not be running, filed https://github.com/kubernetes/kubernetes/pull/80044

BenTheElder:
(dashboard test change is in)

hack/ci/e2e.sh
```diff
@@ -187,9 +187,29 @@ EOF
...
+    # [sig-storage] PersistentVolumes-local
+    # fails because sh: 1: sudo: not found, TODO check if we can detect sudo test/e2e/storage/utils/local.go:
+    SKIP="PersistentVolumes-local|${SKIP}"
+    # [sig-auth] PodSecurityPolicy
+    SKIP="PodSecurityPolicy|${SKIP}"
+    # [sig-network] Network should set TCP CLOSE_WAIT timeout
+    # Use ssh to login to the kind nodes TODO: change to docker exec commands, install ssh server, ...
+    SKIP="Network should set TCP CLOSE_WAIT timeout|${SKIP}"
+    # [sig-cli] Kubectl client [k8s.io] Simple pod should support exec through an HTTP proxy 
+    SKIP="Simple pod should support exec through an HTTP proxy|${SKIP}"
```
aojea: 🤷‍♂ I'm not able to aggregate all the tests to create the SKIP regex
/help @BenTheElder @neolit123

BenTheElder: IIRC with ginkgo you cannot use spaces and need to use . Instead.

aojea: I can use a regex like that with spaces if I execute the e2e binary from the command line, will try the dots , thanks

BenTheElder:
so at least some of the failing tests are using SSH to the node in the e2e code but not checking if this is available first.

a) they should be checking (will PR that)

b) we should be able to make the SSHing to run commands more flexible in the e2e framework so you can point it at docker exec instead. SSH seems to be used only to exec looking so far, not any port forwarding or anything like that.

BenTheElder:
some tests also _require_ that the "master" node be named `.*master` in order to do things like grab metrics ... :man_facepalming: 

we could maybe introduce some secret compatibility mode for this, but I don't really like the idea that to use kind for it's default purpose you will need to do this. I also don't like the spurious breaking change of changing our node naming scheme or using that name anyhow ... :|

I'm not sure how critical these are, and we're discussing this upstream (see eg #testing-commons) ...

other things:
tests that use nsenter + sudo (!)

aojea:
> other things:
> tests that use nsenter + sudo (!)

Those were introduced with https://github.com/kubernetes/kubernetes/pull/73525

is sudo really necessary?

hack/ci/e2e.sh
```diff
@@ -187,9 +187,30 @@ EOF
         | grep -cv "node-role.kubernetes.io/master" \
     )"
 
-    # ginkgo regexes
-    SKIP="${SKIP:-}"
-    FOCUS="${FOCUS:-"\\[Conformance\\]"}"
+    # ginkgo regexes, use CI k/k presubmit regex by default
+    SKIP="${SKIP:-"\\[Slow\\]|\\[Disruptive\\]|\\[Flaky\\]|\\[Feature:.+\\]"}"
+    FOCUS="${FOCUS:-""}"
```
BenTheElder: /hold
explicitly holding this because conformance testing jobs at minimum should explicitly set focus / no skip and ignore any choices here.

this PR is useful to explore the tests, but we'll need to implement a bit differently to avoid suddenly changing what all of the various jobs using this script do.

aojea: sure, I just was trying to get the right regex and identify the problems

hack/ci/e2e.sh
```diff
@@ -187,9 +187,30 @@ EOF
...
+
+    # skip known issues
+    # [sig-node] RuntimeClass should run a Pod requesting a RuntimeClass with a configured handler [NodeFeature:RuntimeHandler]
+    # https://github.com/kubernetes/kubernetes/issues/80098
+    SKIP="RuntimeClass.should.run.a.Pod.requesting.a.RuntimeClass.with.a.configured.handler|${SKIP}"
+    # [sig-storage] In-tree Volumes [Driver: nfs]
+    # [sig-storage] PersistentVolumes NFS
+    # TODO: kind nodes can't mount nfs volumes: install apt install -y nfs-common in the base image, ...??
+    SKIP="In-tree.Volumes.\\[Driver:.nfs\\]|PersistentVolumes.NFS|${SKIP}"
```
aojea: seems NFS can´t export an overlay fs https://lwn.net/Articles/743275/

aojea: available since kernel 4.16 https://www.phoronix.com/scan.php?page=news_item&px=OverlayFS-NFS-Export-Linux-4.16

BenTheElder: it may be a bit before our hosts have this, let's put a TODO to try this when the host kernels are new enough. pretty sure our ubuntu GKE nodes are ~4.15

BenTheElder:
three failures that have not been skipped in master:
- HPA (we don't have the metrics API, open discussion on that currently)
- Mount Propagation: this uses SSH. Can we port it to their host-exec pod implementation upstream? Or do we need to implement ""SSH"", need to look at both
- not sure on the statefulset identity test

/retest

hack/ci/e2e.sh
```diff
@@ -187,9 +187,30 @@ EOF
...
+    # [sig-storage] In-tree Volumes [Driver: local]
+    # fails because sh: 1: sudo: not found, TODO check if we can detect sudo test/e2e/storage/utils/local.go:
+    SKIP="PersistentVolumes-local|In-tree.Volumes.\\[Driver:.local\\]|${SKIP}"
+    # [sig-auth] PodSecurityPolicy
+    SKIP="PodSecurityPolicy|${SKIP}"
+    # [sig-network] Network should set TCP CLOSE_WAIT timeout
+    # Use ssh to login to the kind nodes TODO: change to docker exec commands, install ssh server, ...
+    SKIP="Network.should.set.TCP.CLOSE_WAIT.timeout|${SKIP}"
+    # [sig-cli] Kubectl client [k8s.io] Simple pod should support exec through an HTTP proxy 
+    SKIP="Simple.pod.should.support.exec.through.an.HTTP.proxy|${SKIP}"
```
BenTheElder: is this one using SSH?

hack/ci/e2e.sh
```diff
@@ -202,7 +223,7 @@ EOF
     fi
 
     # add ginkgo args
-    KUBETEST_ARGS="${KUBETEST_ARGS} --test_args=\"--ginkgo.focus=${FOCUS} --ginkgo.skip=${SKIP} --report-dir=${ARTIFACTS} --disable-log-dump=true --num-nodes=${NUM_NODES}\""
+    KUBETEST_ARGS="${KUBETEST_ARGS} --test_args=\"--ginkgo.focus=${FOCUS} --ginkgo.skip=${SKIP} --container-runtime=containerd --report-dir=${ARTIFACTS} --disable-log-dump=true --num-nodes=${NUM_NODES}\""
```
BenTheElder: this coupling is :(

aojea: this was because of https://github.com/kubernetes/kubernetes/issues/80098 , but reading the issue seems that is not the problem

BenTheElder: ah yes, the problem is the test-handler, what happens if we remove this flag?

hack/ci/e2e.sh
aojea: nope, it seems is not able to use the proxy 🤷‍♂ 
> Jul 12 04:47:56.459: INFO: Missing expected log result on proxy server for https_proxy. Expected: "Accepting CONNECT to 127.0.0.1:35545", got ""

hack/ci/e2e.sh
aojea: it doesn´t affect the other tests

hack/ci/e2e.sh
BenTheElder: is that because it's tying to use the "External" address for the api server?

BenTheElder: ... that's a bit of a problem for kind :thinking:

aojea: good catch, it´s because is using the api server endpoint on localhost.
Passing the host as parameter to the e2e tests with the node API server address works
` bazel-bin/test/e2e/e2e.test -ginkgo.focus="Simple pod should support exec through an HTTP proxy" -host https://172.17.0.5:6443`

BenTheElder: hmmm I wonder if this works on macbook ... the usual suspect :P

BenTheElder:
[BTW I am locally testing some of this currently]

hack/ci/e2e.sh
aojea: I'm running them on Linux

hack/ci/e2e.sh
```diff
@@ -187,9 +187,32 @@ EOF
...
+    SKIP="RuntimeClass.should.run.a.Pod.requesting.a.RuntimeClass.with.a.configured.handler|${SKIP}"
+    # [sig-storage] In-tree Volumes [Driver: nfs]
+    # [sig-storage] PersistentVolumes NFS
+    # TODO: kind nodes can't mount nfs volumes: install apt install -y nfs-common in the base image, kernel version, ...??
+    SKIP="In-tree.Volumes.\\[Driver:.nfs\\]|PersistentVolumes.NFS|${SKIP}"
+    # [sig-storage] PersistentVolumes-local
+    # [sig-storage] In-tree Volumes [Driver: local]
+    # fails because sh: 1: sudo: not found 
+    # Fix: https://github.com/kubernetes/kubernetes/pull/80135
+    SKIP="PersistentVolumes-local|In-tree.Volumes.\\[Driver:.local\\]|${SKIP}"
```
BenTheElder: https://github.com/kubernetes/kubernetes/pull/80329 is in for removal of sudo

hack/ci/e2e.sh
```diff
@@ -187,9 +187,32 @@ EOF
...
+    # [sig-auth] PodSecurityPolicy
```
BenTheElder: (noting that this is because we don't have PSP working properly in kind)

aojea: I remember some conversations about that in the slack channel, IIRC you can´t have swap enable.
Is this something we can have working in current CI?

BenTheElder: CI doesn't have swap enabled because it is also kubernetes. swap relates to memory limits though, I think this is a container "nesting" issue.

aojea: we didn´t enable the admission controller, in order to do that we need
```
diff --git a/pkg/cluster/internal/kubeadm/config.go b/pkg/cluster/internal/kubeadm/config.go
index eb87dfb..15e0f2f 100644
--- a/pkg/cluster/internal/kubeadm/config.go
+++ b/pkg/cluster/internal/kubeadm/config.go
@@ -99,6 +99,9 @@ controlPlaneEndpoint: "{{ .ControlPlaneEndpoint }}"
 api:
   advertiseAddress: "{{ .NodeAddress }}"
   bindPort: {{.APIBindPort}}
+# enable PodSecurityPolicy admission controller
+apiServerExtraArgs:
+  enable-admission-plugins: PodSecurityPolicy
 # we need nsswitch.conf so we use /etc/hosts
 # https://github.com/kubernetes/kubernetes/issues/69195
 apiServerExtraVolumes:
@@ -165,6 +168,9 @@ controlPlaneEndpoint: "{{ .ControlPlaneEndpoint }}"
 networking:
   podSubnet: "{{ .PodSubnet }}"
   serviceSubnet: "{{ .ServiceSubnet }}"
+# enable PodSecurityPolicy admission controller
+apiServerExtraArgs:
+  enable-admission-plugins: PodSecurityPolicy
 # we need nsswitch.conf so we use /etc/hosts
 # https://github.com/kubernetes/kubernetes/issues/69195
 apiServerExtraVolumes:
@@ -259,6 +265,8 @@ controlPlaneEndpoint: "{{ .ControlPlaneEndpoint }}"
 apiServer:
+  extraArgs:
+    enable-admission-plugins:  PodSecurityPolicy,LimitRanger,ResourceQuota,AlwaysPullImages,DefaultStorageClass
 controllerManager:
   extraArgs:
@@ -356,6 +364,8 @@ controlPlaneEndpoint: "{{ .ControlPlaneEndpoint }}"
 apiServer:
+  extraArgs:
+    enable-admission-plugins:  PodSecurityPolicy,LimitRanger,ResourceQuota,AlwaysPullImages,DefaultStorageClass
 controllerManager:
   extraArgs:
```
Problem is that host without apparmor don´t allow to create pods with apparmor annotations, and this breaks `kindnet`

BenTheElder: perhaps we should adjust the kindnetd spec

aojea: but if we enable it, ir can break a lot of users that use manifests with the apparmor and seccomp annotations

BenTheElder: that's orthogonal to adjusting the kindnetd spec so that it _Could_ be enabled

aojea: @BenTheElder check the last commit 1d0c656f1a9a8d791421d1446c52ab7980aff0f8

aojea: Let me check tomorrow if it works locally and will submit a new PR with the change

hack/ci/e2e.sh
BenTheElder: right, but this is a reminder that we should make sure not to merge it, as that will break the conformance CI :sweat_smile:

aojea:
@BenTheElder what do you think about having an env variable to switch ginkgo regexs?

aojea:
@BenTheElder right now the SKIP exceptions are right for master, however these tests are flaky, may it be due to resources constrains?
> [k8s.io] Pods should support pod readiness gates [NodeFeature:PodReadinessGate] expand_more	
> [sig-apps] StatefulSet [k8s.io] Basic StatefulSet functionality [StatefulSetBasic] should provide basic identity expand_more

hack/ci/e2e.sh
```diff
@@ -187,9 +187,50 @@ EOF
...
+        SKIP="${SKIP:-"\\[Slow\\]|\\[Disruptive\\]|\\[Flaky\\]|\\[Feature:.+\\]"}"
+        FOCUS="${FOCUS:-""}"
+        # skip tests that are not able to rung with kind
+        # [sig-node] RuntimeClass should run a Pod requesting a RuntimeClass with a configured handler [NodeFeature:RuntimeHandler]
+        # https://github.com/kubernetes/kubernetes/issues/80098
+        SKIP="RuntimeClass.should.run.a.Pod.requesting.a.RuntimeClass.with.a.configured.handler|${SKIP}"
+        # [sig-storage] In-tree Volumes [Driver: nfs]
+        # [sig-storage] PersistentVolumes NFS
+        # TODO: kind nodes can't mount nfs volumes: install apt install -y nfs-common in the base image, kernel version, ...??
+        SKIP="In-tree.Volumes.\\[Driver:.nfs\\]|PersistentVolumes.NFS|${SKIP}"
```
aojea: @BenTheElder this works for me in a kernel 5.2.2, the fix was in version 4.16.
Is it possible to run that kernel version in the CI?
We'll need to install nfs-common in the nodes too.

hack/ci/e2e.sh
```diff
@@ -187,9 +187,50 @@ EOF
...
+        # [sig-auth] PodSecurityPolicy
+        # We are not enabling the security admission controler in kind
+        # However, if enabled, it breaks kindnet because it needs apparmor in the host
+        SKIP="PodSecurityPolicy|${SKIP}"
```
aojea: need to retest with this https://github.com/kubernetes-sigs/kind/pull/725

aojea:
/test pull-kind-conformance-parallel

the failing tests in the CI are passing in my local desktop 🤔

aojea:
ok, current regex are Green, some flakiness but has to be due to environment constrains

aojea:
Slow tests only 10 failures with ipv4 and 169 pass

```
[sig-network] Services should be able to switch session affinity for LoadBalancer service with ESIPP off [Slow] [DisabledForLargeClusters] expand_more	20m38s
[sig-storage] Dynamic Provisioning DynamicProvisioner External should let an external dynamic provisioner create and delete persistent volumes [Slow] expand_more	5m47s
[sig-network] Services should remove load balancer cleanup finalizer when service is deleted [Slow] expand_more	20m7s
[sig-storage] In-tree Volumes [Driver: local][LocalVolumeType: block] [Testpattern: Pre-provisioned PV (default fs)] subPath should support creating multiple subpath from same volumes [Slow] expand_more	21s
[sig-network] Services should have session affinity work for LoadBalancer service with ESIPP off [Slow] [DisabledForLargeClusters] expand_more	20m38s
[sig-storage] Dynamic Provisioning DynamicProvisioner [Slow] should provision storage with different parameters expand_more	8s
[sig-network] Services should have session affinity work for LoadBalancer service with ESIPP on [Slow] [DisabledForLargeClusters] expand_more	20m28s
[sig-network] Services should be able to switch session affinity for LoadBalancer service with ESIPP on [Slow] [DisabledForLargeClusters] expand_more	20m25s
[sig-storage] PersistentVolumes-local [Volume type: blockfswithoutformat] Set fsGroup for local volume should set same fsGroup for two pods simultaneously [Slow] expand_more	16s
[sig-autoscaling] [HPA] Horizontal pod autoscaling (scale resource: CPU) [sig-autoscaling] ReplicationController light Should scale from 2 pods to 1 pod [Slow] expand_more
```

hack/ci/e2e.sh
BenTheElder: no, the nodes are not going to be that new anytime soon. pretty sure they are ubuntu LTS or COS so

hack/ci/e2e.sh
```diff
@@ -187,9 +187,45 @@ EOF
         | grep -cv "node-role.kubernetes.io/master" \
     )"
 
+    KUBERNETES_PRESUBMIT_JOB=true
     # ginkgo regexes
-    SKIP="${SKIP:-}"
-    FOCUS="${FOCUS:-"\\[Conformance\\]"}"
+    if [[ "${KUBERNETES_PRESUBMIT_JOB:-}" = true ]]; then
+        # k/k presubmit jobs regex
+        SKIP="${SKIP:-"\\[Disruptive\\]|\\[Flaky\\]|\\[Feature:.+\\]"}"
```
BenTheElder: we need to collect this regex up and set it on the prowjob
as nice as it is to just checkout this repo and run `PRESUBMIT=true e2e.sh` or similar, we need what the skip is to be per-kubernetes branch going forward, as we may have to fix the _tests_ and the skip / focus will diverge.

this PR has been _super_ helpful to discover which tests though

BenTheElder: (and we should keep experimenting)

aojea: Do you mean passing the SKIP var as an ENV variable in the prow job?

BenTheElder:
https://github.com/kubernetes/test-infra/pull/14524

BenTheElder:
xref: https://github.com/kubernetes/test-infra/pull/14530

hack/ci/e2e.sh
```diff
@@ -187,9 +187,45 @@ EOF
...
+        FOCUS="${FOCUS:-"\\[Slow\\]"}"
+        # skip tests that are not able to rung with kind
```
neolit123: rung -> run

aojea: this was just only for testing, the important one is the job that Ben is setting up in test-infra

BenTheElder: Yeah, this is one of the ways we determined the regex (though not all of it, it doesn't match what were actually using for now), we opted to manage them with the jobs instead to better handle different Kubernetes branches etc.

The regex will change as we manage to improve the tests upstream.

BenTheElder: (in addition to kind changes that would affect all branches)

PR #704, opened 2019-07-12
README.md: we also need set $KUBECONFIG env

anjiawei1991:
we need set  $KUBECONFIG enviroment after created the cluster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 kind is a tool for running local Kubernetes clusters using Docker container "nodes".  
 kind is primarily designed for testing Kubernetes 1.11+, initially targeting the [conformance tests].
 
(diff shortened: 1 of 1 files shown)


anjiawei1991:
I signed it

anjiawei1991:
I signed it

BenTheElder:
Hi, thanks for the PR.

These instructions are output in the relevant format for the host OS on each platform when you run the `kind create cluster` command. I don't think we should duplicate them here (especially as we need multiple versions depending on your OS / shell) 😅

Technically to use the cluster you may also need to install a client like kubectl but this part depends on your use case for any cluster and isn't included here. 

The actual user guide also covers this in better detail https://kind.sigs.k8s.io/docs/user/quick-start/#interacting-with-your-cluster

tao12345666333:
Thanks for your work.  :+1: 
ButI agree with @BenTheElder 
In this content of README.md, we only need to include the most basic `kind` command itself.

Issue #659, opened 2019-06-26
[probably already fixed] cmder: stderr can end up in stdout even if using cmd.SetStdout()


neolit123:
**What happened**:
running a command in a node-image container that has klog stderr output and using SetStdout on the command can result in the stderr output present in the io.Writer/buffer set for stdout.

**How to reproduce it (as minimally and precisely as possible)**:

kubeadm needs to be 1.14.x

```
var etcdVersionBuf bytes.Buffer
cmder := docker.ContainerCmder(containerName)
cmd := cmder.Command("/bin/sh", "-c", "kubeadm config images list | grep etcd")
cmd.SetStdout(&etcdVersionBuf)
cmd.Run()
```
if the kubeadm command outputs a klog warning it will end up in `etcdVersionBuf`.
contents of `etcdVersionBuf`:
```
I0626 02:54:47.931193    3673 version.go:240] remote version
  is much newer: v1.15.0; falling back to: stable-1.14
k8s.gcr.io/etcd:3.3.10
```

current fix is to do this:
```
cmd := cmder.Command("/bin/sh", "-c", "kubeadm config images list 2> /dev/null | grep etcd")
```

i don't see issues in the kind code, so maybe a Go thing? please close if this is expected.
locally if i do:
```
kubeadm config images list | grep etcd > test.log
```
`test.log` only contains `k8s.gcr.io/etcd:3.3.10`

**Anything else we need to know?**:

xref https://github.com/kubernetes/kubeadm/pull/1636

**Environment:**
kinder

- kind version: (use `kind version`):
- Kubernetes version: (use `kubectl version`): ci/latest-1.14
- Docker version: (use `docker info`): 18.06.3-ce
- OS (e.g. from `/etc/os-release`):
```
NAME="Ubuntu"
VERSION="17.10 (Artful Aardvark)"
```

BenTheElder:
is it possible it's using some old kind version? I can't find any logical errors in the current code, stdout and stderr are managed distinctly.

BenTheElder:
I think this bug was (mostly) fixed 2 months ago, it seems the pseudo tty from `docker exec -t` includes both stdout and stderr. we will only choose this if the cmder stdout or stderr are wired to a terminal currently. we should probably not support that and never allocate one.

BenTheElder:
specifically: https://github.com/kubernetes-sigs/kind/commit/97d0968a8e505382088587afeb6f8814a9fb5e6f

[...]

PR #709, opened 2019-07-16
Image load refactor

BenTheElder:
/hold
I need to do some cleanup, but this is the general approach.

- stop copying in images like every other file, instead have bits provide a separate list of image files
- load images directly from the list of image files, no unnecessary copies / staging / rsync / ...
- replace the bash one-liner to load images with a minimal internal API to prepare the container for image loading, load, and cleanup
- leverage loading each image explicitly from the kind code to stream images through a function that rewrites their repositories (tags) as necessary to add / remove `-${arch}`
- remove tag rewriting logic from the bazel build, as we've now centralized it

besides some simple cleanup, the node build desperately needs a refactor. I'm holding that until after the logger is switched out so we can overhaul the output while we move the logic around ...

### DIFF
--- a/pkg/build/node/import.go
+++ b/pkg/build/node/import.go
@@ -1,0 +1,59 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package node
+
+import (
+	"sigs.k8s.io/kind/pkg/exec"
+)
+
+type imageImporter interface {
+	Prepare() error
+	LoadCommand() exec.Cmd
+	End() error
+}
+
+type containerdImporter struct {
+	containerCmder exec.Cmder
+}
+
+func newContainerdImporter(containerCmder exec.Cmder) imageImporter {
+	return &containerdImporter{
+		containerCmder: containerCmder,
+	}
+}
+
+var _ imageImporter = &containerdImporter{}
+
+func (c *containerdImporter) Prepare() error {
+	if err := c.containerCmder.Command(
+		"bash", "-c", "nohup containerd > /dev/null 2>&1 &",
+	).Run(); err != nil {
+		return err
+	}
+	// TODO(bentheelder): some healthcheck?
+	return nil
+}
+
+func (c *containerdImporter) LoadCommand() exec.Cmd {
+	return c.containerCmder.Command(
+		"ctr", "--namespace=k8s.io", "images", "import", "--no-unpack", "-",
+	)
+}
+
+func (c *containerdImporter) End() error {
+	return c.containerCmder.Command("pkill", "containerd").Run()
+}
--- a/pkg/build/node/node.go
+++ b/pkg/build/node/node.go
@@ -19,9 +19,9 @@ package node
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"path"
-	"regexp"
 	"strings"
 	"time"
 
@@ -178,21 +178,15 @@ func (c *BuildContext) populateBits(buildDir string) error {
 	return nil
 }
 
-// matches image tarballs kind will sideload
-var imageRegex = regexp.MustCompile(`images/[^/]+\.tar`)
-
 // returns a set of image tags that will be sideloaded
 func (c *BuildContext) getBuiltImages() (sets.String, error) {
-	bitPaths := c.bits.Paths()
 	images := sets.NewString()
-	for src, dest := range bitPaths {
-		if imageRegex.MatchString(dest) {
-			tags, err := docker.GetArchiveTags(src)
-			if err != nil {
-				return nil, err
-			}
-			images.Insert(tags...)
+	for _, path := range c.bits.ImagePaths() {
+		tags, err := docker.GetArchiveTags(path)
+		if err != nil {
+			return nil, err
 		}
+		images.Insert(tags...)
 	}
 	return images, nil
 }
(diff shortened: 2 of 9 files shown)


pkg/build/node/node.go
```diff
@@ -496,19 +497,88 @@ func (c *BuildContext) prePullImages(dir, containerID string) error {
...
-	)); err != nil {
-		log.Errorf("Image build Failed! Failed to load images into containerd %v", err)
+	// setup image importer
+	importer := newContainerdImporter(cmder)
+	if err := importer.Prepare(); err != nil {
+		log.Errorf("Image build Failed! Failed to prepare containerd to load images %v", err)
+		return err
+	}
+
+	// TODO: return this error?
```
BenTheElder: I'm not sure how much we care, and this is hardly the messiest thing in this file :/

pkg/container/docker/exec.go
```diff
@@ -68,11 +67,11 @@ func (c *containerCmd) Run() error {
 		)
 	}
 	// if the command is hooked up to the processes's output we want a tty
-	if log.IsTerminal(c.stderr) || log.IsTerminal(c.stdout) {
+	/*if log.IsTerminal(c.stderr) || log.IsTerminal(c.stdout) {
```
BenTheElder: this section be outright deleted, not commented out.

pkg/exec/exec.go
```diff
@@ -120,3 +120,30 @@ func RunWithStdoutReader(cmd Cmd, readerFunc func(io.Reader) error) error {
...
+	err = cmd.Run()
+	if err != nil {
+		return err
+	}
+	err2 := <-errChan
+	if err2 != nil {
+		return err2
+	}
+	return nil
+}
```
BenTheElder: This function is basically RunWithStdoutReader but Stdin instead (see above), I dislike that these don't compose but that's a hypothetical use-case at the moment.

pkg/build/node/node.go
```diff
@@ -496,19 +497,88 @@ func (c *BuildContext) prePullImages(dir, containerID string) error {
...
+	// run all image loading concurrently until one fails or all succeed
+	if err := concurrent.UntilError(loadFns); err != nil {
+		log.Errorf("Image build Failed! Failed to load images %v", err)
 		return err
 	}
 
 	return nil
 }
 
+func repositoryCorrectorForVersion(kubeVersion *version.Version) func(string) string {
```
BenTheElder: this logic should be leveraged to determine which tags we built as well when deciding what to pull.

BenTheElder: done.

pkg/container/docker/exec.go
BenTheElder: done

pkg/build/node/import.go
```diff
@@ -0,0 +1,61 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
BenTheElder: should be 2019

pkg/build/node/import.go
```diff
@@ -0,0 +1,61 @@
...
+	return &containerdImporter{
+		containerCmder: containerCmder,
+	}
+}
+
+var _ imageImporter = &containerdImporter{}
+
+func (c *containerdImporter) Prepare() error {
+	if err := c.containerCmder.Command(
+		"bash", "-c", "nohup containerd > /dev/null 2>&1 &",
```
BenTheElder: we could probably rewrite this to just need sh but i'm not sure we care right now

pkg/build/node/import.go
```diff
@@ -0,0 +1,61 @@
...
+	).SetStdout(os.Stdout).Run(); err != nil {
```
BenTheElder: this SetStdout should be removed

pkg/build/node/import.go
BenTheElder: NOTE: we are writing this to /dev/null for now so it doesn't leave any stray files in the built image, but we should consider debuggability here in the future. Ideally viewing this isn't necessary.

pkg/container/docker/exec.go
```diff
@@ -67,12 +66,6 @@ func (c *containerCmd) Run() error {
 			"-i", // interactive so we can supply input
 		)
 	}
-	// if the command is hooked up to the processes's output we want a tty
-	if log.IsTerminal(c.stderr) || log.IsTerminal(c.stdout) {
-		args = append(args,
-			"-t",
-		)
-	}
```
BenTheElder: minor unrelated bug, resolved in the course of working on this.

pkg/container/docker/image.go
```diff
@@ -17,10 +17,52 @@ limitations under the License.
...
+// resolve the implicit :latest
+func SplitImage(image string) (registry, tag string, err error) {
+	// we are looking for ':' and '@'
+	firstColon := strings.IndexByte(image, 58)
+	firstAt := strings.IndexByte(image, 64)
+
+	// there should be a registry before the tag, and @/: should not be the last
+	// character, these cases are assumed not to exist by the rest of the code
+	if firstColon == 0 || firstAt == 0 || firstColon+1 == len(image) || firstAt+1 == len(image) {
+		return "", "", fmt.Errorf("unexpected image: %q", image)
```
BenTheElder: will do a sweep and convert to pkg/errors later. would rather focus on getting the fix in and I don't think this error will be difficult to locate.

pkg/build/node/import.go
BenTheElder: done.

pkg/build/node/import.go
BenTheElder: done, and one in pkill below

BenTheElder:
/hold cancel
FYI @hoegaarden

BenTheElder:
going to use my super powers to unbreak things, will address any feedback in follow ups.

pkg/build/node/node.go
```diff
@@ -401,14 +395,24 @@ func (c *BuildContext) prePullImages(dir, containerID string) error {
...
 	}
-	if ver.LessThan(version.MustParseSemantic("v1.12.0")) {
-		archSuffix := fmt.Sprintf("-%s:", c.arch)
-		for _, image := range builtImages.List() {
-			if !strings.Contains(image, archSuffix) {
-				builtImages.Insert(strings.Replace(image, ":", archSuffix, 1))
-			}
+
+	// get image tag fixing function for this version
+	fixRepository := repositoryCorrectorForVersion(ver)
```
neolit123: :+1:

pkg/build/node/node.go
```diff
@@ -496,19 +500,85 @@ func (c *BuildContext) prePullImages(dir, containerID string) error {
...
+			if err != nil {
+				return err
+			}
+			defer f.Close()
+			//return importer.LoadCommand().SetStdout(os.Stdout).SetStderr(os.Stderr).SetStdin(f).Run()
+			// we will rewrite / correct the tags as we load the image
+			if err := exec.RunWithStdinWriter(importer.LoadCommand().SetStdout(os.Stdout).SetStderr(os.Stdout), func(w io.Writer) error {
+				return docker.EditArchiveRepositories(f, w, fixRepository)
+				// _, err := io.Copy(w, f)
+				// return err
```
neolit123: left over comments.

pkg/container/docker/exec.go
neolit123: can you please elaborate?
AFAIK this was the fix for https://github.com/kubernetes-sigs/kind/issues/659

pkg/container/docker/image_test.go
```diff
@@ -0,0 +1,130 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package docker
+
+import "testing"
+
+func TestSplitImage(t *testing.T) {
```
neolit123: unit tests are nice.

pkg/build/node/node.go
BenTheElder: Yes left over testing. Will clean it up. Thanks.

pkg/container/docker/exec.go
BenTheElder: There's no time we need this flag. The fix for that issue was to set it less often, this sets it never.

BenTheElder: Elaborating: this flag allocates a pseudo tty which is nice for interactively working with a container from a shell but a) we don't do that b) it allocates the tty to both stderr and stdout c) it changes behavior around exiting (because the tty needs to be closed)

PR #715, opened 2019-07-17
add omitempty tag to Nodes and Networking attributes

bonifaido:
both of the attributes are optional when reading back a marshaled version


BenTheElder:
That shouldn't happen, both should have defaulting?

BenTheElder:
We're waiting for an upstream fix for that failure to merge 😅

bonifaido:
Ah thanks, didn't know that :)

BenTheElder:
It actually just merged 😁, not sure if it snuck in before this retest though. I think just after.

PR #722, opened 2019-07-19
Build arm32 kind and ctr variants

munnerz:
Raspberry Pi's by default run in 32-bit mode, for {reasons}.

This means that when building a node image on a raspberry pi, currently it errors with:

```
curl -fSL -o /usr/local/bin/ctr       "https://storage.googleapis.com/bentheelder-kind-dev/containerd/linux/${ARCH}/ctr"     && chmod +x /usr/local/bin/ctr     && echo "done installing packages"' returned a non-zero code: 22
```

This is because there is no `arm` variant of containerd published to the bucket.

This PR adds `arm` as a build target for `containerd`, as well as for the `kind` binary itself (to make peoples lives easier 😄).

As we don't publish multi arch *node images* currently, users will still be required to build their own. But this at least edges us closer towards it being *possible* 😄.

~I've not actually tested the `kind build base-image` command after running this, but I *have* run the `ctr` build script which succeeds.~

EDIT: with a few more changes, it's now possible to build the base image for arm32. More work is required to get the node image to build 😄 

/cc @BenTheElder

### DIFF
--- a/hack/release/build/cross.sh
+++ b/hack/release/build/cross.sh
@@ -52,6 +52,7 @@ build() {
 # TODO(bentheelder): support more platforms
 echo "Building in parallel for:"
 build "linux" "amd64" & \
+build "linux" "arm" & \
 build "linux" "arm64" & \
 build "linux" "ppc64le" & \
 build "darwin" "amd64" & \
--- a/hack/release/build/ctr/run.sh
+++ b/hack/release/build/ctr/run.sh
@@ -53,6 +53,7 @@ REVISION=$(git rev-parse HEAD)$(if ! git diff --no-ext-diff --quiet --exit-code;
 
 # submit a build for each arch
 GOARCHES=(
+  "arm"
   "arm64"
   "amd64"
   "ppc64le"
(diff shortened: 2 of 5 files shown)


munnerz:
An update on this effort: I actually have got the node image to at least begin building. Due to Bazel not publishing `arm` builds (and myself not having a suitable cross compiler set up), I've opted for the Docker based build.

This has required me to produce my own kube-cross image, which *also* required some additional changes to that Dockerfile (namely to reduce the amount of stuff built, and also to use a correct architecture protoc).

I'll be sure to update here with the result of this build.. in the meantime, enjoy this image of my homebrew raspberry pi 4 fan.

Image of a Raspberry Pi with a huge fan pointed at it

BenTheElder:
oops, sorry about the bentheelder storage bit, we should check if the upstream packages already contain my `ctr` flag (`--no-unpack`)

BenTheElder:
we should add arm 32 to https://github.com/kubernetes-sigs/kind/tree/master/hack/release/build/ctr

BenTheElder:
running the ctr build now (From https://github.com/kubernetes-sigs/kind/pull/723 but w/e)

BenTheElder:
this will be super nice :-)

PR #725, opened 2019-07-19
Add an option to enable PodSecurityPolicy

aojea:
To enable PodSecurityPolicy we have to configure it in kubeadm.
Main problem is that if apparmor or seccomp annotations are detected and those features are not available in the host, those pods with annotations fail to start (i.e. kindnet)

### DIFF
--- a/pkg/build/node/cni.go
+++ b/pkg/build/node/cni.go
@@ -37,11 +37,6 @@ apiVersion: policy/v1beta1
 kind: PodSecurityPolicy
 metadata:
   name: kindnet
-  annotations:
-    seccomp.security.alpha.kubernetes.io/allowedProfileNames: docker/default
-    seccomp.security.alpha.kubernetes.io/defaultProfileName: docker/default
-    apparmor.security.beta.kubernetes.io/allowedProfileNames: runtime/default
-    apparmor.security.beta.kubernetes.io/defaultProfileName: runtime/default
 spec:
   privileged: false
   volumes:
--- a/pkg/apis/config/v1alpha3/types.go
+++ b/pkg/apis/config/v1alpha3/types.go
@@ -49,6 +49,9 @@ type Cluster struct {
 	// KubeadmConfigPatchesJSON6902 are applied to the generated kubeadm config
 	// as patchesJson6902 to `kustomize build`
 	KubeadmConfigPatchesJSON6902 []PatchJSON6902 `json:"kubeadmConfigPatchesJson6902,omitempty"`
+
+	// EnablePodSecurityPolicy enable the admission controller
+	EnablePodSecurityPolicy bool `json:"enablePodSecurityPolicy,omitempty"`
 }
 
 // Node contains settings for a node in the `kind` Cluster.
--- a/pkg/internal/apis/config/types.go
+++ b/pkg/internal/apis/config/types.go
@@ -49,6 +49,9 @@ type Cluster struct {
 	// KubeadmConfigPatchesJSON6902 are applied to the generated kubeadm config
 	// as patchesJson6902 to `kustomize build`
 	KubeadmConfigPatchesJSON6902 []PatchJSON6902
+
+	// EnablePodSecurityPolicy enable the admission controller
+	EnablePodSecurityPolicy bool
 }
 
 // Node contains settings for a node in the `kind` Cluster.
--- a/pkg/internal/apis/config/v1alpha3/zz_generated.conversion.go
+++ b/pkg/internal/apis/config/v1alpha3/zz_generated.conversion.go
@@ -87,6 +87,7 @@ func autoConvert_v1alpha3_Cluster_To_config_Cluster(in *v1alpha3.Cluster, out *c
 	}
 	out.KubeadmConfigPatches = *(*[]string)(unsafe.Pointer(&in.KubeadmConfigPatches))
 	out.KubeadmConfigPatchesJSON6902 = *(*[]config.PatchJSON6902)(unsafe.Pointer(&in.KubeadmConfigPatchesJSON6902))
+	out.EnablePodSecurityPolicy = in.EnablePodSecurityPolicy
 	return nil
 }
 
@@ -102,6 +103,7 @@ func autoConvert_config_Cluster_To_v1alpha3_Cluster(in *config.Cluster, out *v1a
 	}
 	out.KubeadmConfigPatches = *(*[]string)(unsafe.Pointer(&in.KubeadmConfigPatches))
 	out.KubeadmConfigPatchesJSON6902 = *(*[]v1alpha3.PatchJSON6902)(unsafe.Pointer(&in.KubeadmConfigPatchesJSON6902))
+	out.EnablePodSecurityPolicy = in.EnablePodSecurityPolicy
 	return nil
 }
 
--- a/pkg/internal/cluster/create/actions/config/config.go
+++ b/pkg/internal/cluster/create/actions/config/config.go
@@ -92,6 +92,7 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 		ServiceSubnet:        ctx.Config.Networking.ServiceSubnet,
 		ControlPlane:         true,
 		IPv6:                 ctx.Config.Networking.IPFamily == "ipv6",
+		PodSecurityPolicy:    ctx.Config.EnablePodSecurityPolicy,
 	}
 
 	fns = append(fns, func() error {
--- a/pkg/internal/cluster/create/actions/installpsp/psp.go
+++ b/pkg/internal/cluster/create/actions/installpsp/psp.go
@@ -1,0 +1,133 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package installpsp implements the an action to isntall a default
+// Pod Security Policy
+package installpsp
+
+import (
+	"strings"
+
+	"github.com/pkg/errors"
+
+	"sigs.k8s.io/kind/pkg/cluster/nodes"
+	"sigs.k8s.io/kind/pkg/internal/cluster/create/actions"
+)
+
+type action struct{}
+
+// NewAction returns a new action for installing storage
+func NewAction() actions.Action {
+	return &action{}
+}
+
+// Execute runs the action
+func (a *action) Execute(ctx *actions.ActionContext) error {
+	ctx.Status.Start("Installing Default PodSecurityPolicy  💾")
+	defer ctx.Status.End(false)
+
+	allNodes, err := ctx.Nodes()
+	if err != nil {
+		return err
+	}
+
+	// get the target node for this task
+	node, err := nodes.BootstrapControlPlaneNode(allNodes)
+	if err != nil {
+		return err
+	}
+
+	// add the default storage class
+	if err := addDefaultPodSecurityPolicy(node); err != nil {
+		return errors.Wrap(err, "failed to add default Pod Security Policy")
+	}
+
+	// mark success
+	ctx.Status.End(true)
+	return nil
+}
+
+// a default pod security policy unrestricted
+// we need this for e2es (PodSecurityPolicy)
+const defaultPodSecurityPolicyManifest = `# least restricted policy
+---
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: psp-privileged
+  annotations:
+    seccomp.security.alpha.kubernetes.io/allowedProfileNames: '*'
+spec:
+  privileged: true
+  allowPrivilegeEscalation: true
+  allowedCapabilities:
+  - '*'
+  volumes:
+  - '*'
+  hostNetwork: true
+  hostPorts:
+  - min: 0
+    max: 65535
+  hostIPC: true
+  hostPID: true
+  runAsUser:
+    rule: 'RunAsAny'
+  seLinux:
+    rule: 'RunAsAny'
+  supplementalGroups:
+    rule: 'RunAsAny'
+  fsGroup:
+    rule: 'RunAsAny'
+---
+# Cluster role which grants access to the default pod security policy
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: psp-privileged
+rules:
+- apiGroups:
+  - policy
+  resourceNames:
+  - psp-privileged
+  resources:
+  - podsecuritypolicies
+  verbs:
+  - use
+---
+# Cluster role binding for default pod security policy granting all authenticated users access
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: psp-privileged
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: psp-privileged
+subjects:
+- apiGroup: rbac.authorization.k8s.io
+  kind: Group
+  name: system:authenticated
+`
+
+func addDefaultPodSecurityPolicy(controlPlane *nodes.Node) error {
+	in := strings.NewReader(defaultPodSecurityPolicyManifest)
+	cmd := controlPlane.Command(
+		"kubectl",
+		"--kubeconfig=/etc/kubernetes/admin.conf", "apply", "-f", "-",
+	)
+	cmd.SetStdin(in)
+	return cmd.Run()
+}
--- a/pkg/internal/cluster/create/create.go
+++ b/pkg/internal/cluster/create/create.go
@@ -34,6 +34,7 @@ import (
 
 	configaction "sigs.k8s.io/kind/pkg/internal/cluster/create/actions/config"
 	"sigs.k8s.io/kind/pkg/internal/cluster/create/actions/installcni"
+	"sigs.k8s.io/kind/pkg/internal/cluster/create/actions/installpsp"
 	"sigs.k8s.io/kind/pkg/internal/cluster/create/actions/installstorage"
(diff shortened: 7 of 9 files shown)


pkg/build/node/cni.go
```diff
@@ -97,6 +92,12 @@ rules:
     verbs:
       - list
       - watch
+  - apiGroups:
+      - ""
+    resources:
+      - pods
```
BenTheElder: ??

aojea: unrelated , is a leftover that I forget to clean because when I tried to test I used an image with the bug that doesn´t show the pod_IP as an environment variable and I was thinking that was related to the PSP.

aojea:
/cc @mauilion 

seems we are having this issue too https://github.com/kubernetes/kubernetes/issues/70952

aojea:
I´ve added an option in the API that allows to configure the cluster with the admission controller enabled. However, per @mauilion slack message we should enable at least a non restrictive default policy, where should we add these policies? is a step inside kind or should we done once the cluster has been deployed?
xref: https://github.com/kubernetes/kubernetes/blob/33541bdd34c32ad1dbef20a1282d7917b300d952/cluster/gce/gci/configure-helper.sh#L2195
https://github.com/kubernetes/kubernetes/tree/master/cluster/gce/addons/podsecuritypolicies

BenTheElder:
I'm not sure this should be an option, is there any reason enabling it with a non-restrictive default policy would be a problem?

aojea:
> I'm not sure this should be an option, is there any reason enabling it with a non-restrictive default policy would be a problem?

I wanted to mean that, I do not have a good understanding of PSP, I was trying to ask how can I add a non-restrictive default policy.

I was assuming that is done using a manifest, if that's the case, my question is if the manifest should be embedded inside kind ala CNI or we document it.

aojea:
@neolit123 @BenTheElder is failing because is not able to spawn the kube-proxy pod 🤷‍♂ 

> ul 24 11:28:52 kind-control-plane kubelet[374]: E0724 11:28:52.282001     374 remote_image.go:113] PullImage "k8s.gcr.io/kube-proxy:v1.16.0-alpha.1.191_da3daf2e8a99bb" from image service failed: rpc error: code = Unknown desc = failed to resolve image "k8s.gcr.io/kube-proxy:v1.16.0-alpha.1.191_da3daf2e8a99bb": no available registry endpoint: k8s.gcr.io/kube-proxy:v1.16.0-alpha.1.191_da3daf2e8a99bb not found
Jul 24 11:28:52 kind-control-plane kubelet[374]: E0724 11:28:52.282846     374 kuberuntime_image.go:50] Pull image "k8s.gcr.io/kube-proxy:v1.16.0-alpha.1.191_da3daf2e8a99bb" failed: rpc error: code = Unknown desc = failed to resolve image "k8s.gcr.io/kube-proxy:v1.16.0-alpha.1.191_da3daf2e8a99bb": no available registry endpoint: k8s.gcr.io/kube-proxy:v1.16.0-alpha.1.191_da3daf2e8a99bb not found

however, the image is available in the nodes `k8s.gcr.io/kube-proxy:v1.16.0-alpha.1.191_da3daf2e8a99bb`

I tried to add the k8s.gcr.io endpoint mirror to containerd without success, do you have any idea what can be happening? it has to be related to PSP, but can´t see the problem

pkg/cluster/internal/create/actions/installpsp/psp.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	// mark success
+	ctx.Status.End(true)
+	return nil
+}
+
+// a default pod security policy unrestricted
+// we need this for e2es (PodSecurityPolicy)
+const defaultPodSecurityPolicyManifest = `# least restricted policy
```
neolit123: `# a non-restrictive PodSecurityPolicy`

aojea:
🤔 it seems like a bug, it seems that the kubelet is not sending the `k8s.io` namespace info to containerd

pkg/cluster/internal/create/actions/installpsp/psp.go
```diff
@@ -0,0 +1,133 @@
...
+  runAsUser:
+    rule: 'RunAsAny'
+  seLinux:
+    rule: 'RunAsAny'
+  supplementalGroups:
+    rule: 'RunAsAny'
+  fsGroup:
+    rule: 'RunAsAny'
+---
+# Cluster role which grants access to the default pod security policy
```
neolit123: `# a ClusterRole which grants access to the default PodSecurityPolicy`

pkg/cluster/internal/create/actions/installpsp/psp.go
```diff
@@ -0,0 +1,133 @@
...
+- apiGroups:
+  - policy
+  resourceNames:
+  - psp-privileged
+  resources:
+  - podsecuritypolicies
+  verbs:
+  - use
+---
+# Cluster role binding for default pod security policy granting all authenticated users access
```
neolit123: `# a ClusterRoleBinding for the default PodSecurityPolicy granting all authenticated users access`

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -99,6 +101,11 @@ controlPlaneEndpoint: "{{ .ControlPlaneEndpoint }}"
 api:
   advertiseAddress: "{{ .NodeAddress }}"
   bindPort: {{.APIBindPort}}
+{{ if .PodSecurityPolicy -}}
+# enable PodSecurityPolicy admission controller
+apiServerExtraArgs:
+  enable-admission-plugins: PodSecurityPolicy
```
neolit123: for the newer versions, this will stomp the NodeRestriction controller entry, but it's not really required for the kind use case.

pkg/cluster/config/types.go
```diff
@@ -50,6 +50,9 @@ type Cluster struct {
 	// KubeadmConfigPatchesJSON6902 are applied to the generated kubeadm config
 	// as patchesJson6902 to `kustomize build`
 	KubeadmConfigPatchesJSON6902 []kustomize.PatchJSON6902
+
+	// EnablePodSecurityPolicy enable the admission controller
+	EnablePodSecurityPolicy bool
```
neolit123: seems like something _most_ of the users will not care about?

pkg/cluster/internal/create/actions/installpsp/psp.go
```diff
@@ -0,0 +1,133 @@
...
+		return errors.Wrap(err, "failed to add default Pod Security Policy")
+	}
+
+	// mark success
+	ctx.Status.End(true)
+	return nil
+}
+
+// a default pod security policy unrestricted
+// we need this for e2es (PodSecurityPolicy)
```
neolit123: i'd change the comment to:

> a default, non-restrictive PodSecurityPolicy required for e2e testing

pkg/cluster/internal/create/actions/installpsp/psp.go
```diff
@@ -0,0 +1,133 @@
...
+const defaultPodSecurityPolicyManifest = `# least restricted policy
+---
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: psp-privileged
+  annotations:
+    seccomp.security.alpha.kubernetes.io/allowedProfileNames: '*'
```
neolit123: does it make sense to add the `apparmor... *` here too?
do we have it running on the latest kind node image (Ubuntu)?

aojea:
> 
> i'm curious what e2e tests do we have in the suite for PSP? i'm partially missing the point of having a policy that is pretty much a NO-OP for the purpose of passing an e2e test. 🤔

This is the default security policy, the tests are setting new policies IIUIC
https://github.com/kubernetes/kubernetes/blob/master/test/e2e/auth/pod_security_policy.go

However, I don't have much idea about PSP, the problem of restricting it too much is that the CNI, CoreDNS and other addons are not able to spawn.

pkg/build/node/cni.go
```diff
@@ -37,11 +37,6 @@ apiVersion: policy/v1beta1
 kind: PodSecurityPolicy
 metadata:
   name: kindnet
-  annotations:
-    seccomp.security.alpha.kubernetes.io/allowedProfileNames: docker/default
-    seccomp.security.alpha.kubernetes.io/defaultProfileName: docker/default
-    apparmor.security.beta.kubernetes.io/allowedProfileNames: runtime/default
-    apparmor.security.beta.kubernetes.io/defaultProfileName: runtime/default
```
neolit123: we seem to have another PSP here for kindnet.
with the addition of the permissive global PSP i don't think this will be needed.

neolit123:
> This is the default security policy, the tests are setting new policies IIUIC
https://github.com/kubernetes/kubernetes/blob/master/test/e2e/auth/pod_security_policy.go

> However, I don't have much idea about PSP, the problem of restricting it too much is that the CNI, CoreDNS and other addons are not able to spawn.

getting ordering and precedence of the present PSP is tricky, see:
https://github.com/kubernetes/kubernetes/pull/77792#discussion_r283529408
https://kubernetes.io/docs/concepts/policy/pod-security-policy/#policy-order

pod_security_policy.go has a restrictive and permissive policies.

kind enabling the default permissive policy will conflate with the test of the permissive PSP, but that should not be a problem. what i don't know will happen is if you have a default non-mutating permissive policy and then you apply the restrictive policy.

i could be wrong but it seems to me the Pod will be permitted by the non-mutating permissive policy and the test will fail. i would deffer to @mauilion on this.

pkg/cluster/internal/create/actions/installpsp/psp.go
aojea: the problem I had with apparmor is that if you add it and the host doesn't have apparmor it blocks the pod.

neolit123: ok, understood.

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -356,6 +372,10 @@ controlPlaneEndpoint: "{{ .ControlPlaneEndpoint }}"
 # to the cluster after rewriting the kubeconfig to point to localhost
 apiServer:
   certSANs: [localhost, "{{.APIServerAddress}}"]
+  {{ if .PodSecurityPolicy -}}
+  extraArgs:
+    enable-admission-plugins:  PodSecurityPolicy,LimitRanger,ResourceQuota,AlwaysPullImages,DefaultStorageClass
```
BenTheElder: Uh, you think AlwaysPullImages might be why it wasn't using the preloaded images? 🙃

aojea: 🤦‍♂

pkg/build/node/cni.go
aojea: if users change the global PSP maybe this will break, I will keep it, only removing those entries because of the apparmor problems mentioned in other comment

aojea:
/hold 

test CI with PSP enabled

aojea:
/hold cancel
CI is green with PSP enabled, so we can boot a cluster with an admission controller enabled but the same functionality that they have now. 
Users can modify and tweak PSP as they wish, but we don´t break/block anything :)

pkg/build/node/cni.go
BenTheElder: interestingly calico for example does not seem to ship with a PSP

aojea: I've copied pasted from flannel IIRC

pkg/cluster/internal/create/actions/installpsp/psp.go
BenTheElder: this part concerns me a bit

pkg/cluster/internal/kubeadm/config.go
BenTheElder: hmm that's not great though ....
really we should consider reworking the config to use patches internally. (and possibly making patches less annoying ...)

BenTheElder:
> i'm curious what e2e tests do we have in the suite for PSP? i'm partially missing the point of having a policy that is pretty much a NO-OP for the purpose of passing an e2e test. 🤔

there's ~ 1 kubernetes presubmit test for this, which makes me both want to enable this feature since the test seems reasonable enough, and wary of enabling it and breaking users just to get one test.

pkg/cluster/internal/kubeadm/config.go
aojea: I think that´s a great idea, I'm always afraid of breaking things when modifying this file and adding new fields

pkg/cluster/internal/create/actions/installpsp/psp.go
BenTheElder: do we know how widespread this is in EG helm charts?

aojea: I was digging more into this:
* AKS approach https://github.com/Azure/aks-engine/blob/e492894dfd9bd6668d778c524084254971dd6d88/parts/k8s/addons/kubernetesmasteraddons-pod-security-policy.yaml
* Kops https://github.com/kubernetes/kops/blob/56ccfac26d5a85b72ef9bee7e9c613cd01e749d9/upup/models/cloudup/resources/addons/podsecuritypolicy.addons.k8s.io/k8s-1.10.yaml.template
* kubespray https://github.com/kubernetes-sigs/kubespray/blob/master/roles/kubernetes-apps/cluster_roles/templates/psp.yml.j2
* GCE https://github.com/kubernetes/kubernetes/tree/master/cluster/gce/addons/podsecuritypolicies

aojea:
should we have this into account https://github.com/kubernetes/kubernetes/issues/75238#issuecomment-520251270?

neolit123:
> should we have this into account kubernetes/kubernetes#75238 (comment)?

hm, how does IPVS relate to this PR?
from what i've seen in a number of user reports, IPVS mode is not there yet and enabling it by default should not be advised IMO.

neolit123:
with respect to enabling PSP in kind,
i think adding `EnablePodSecurityPolicy` makes sense only if the PSP can be customizable or if the user can provide custom PSP manifests.

the alternative is the add a guide how to enable PSP after the initial kind cluster is already up and use the same steps in e2e tests. the UX around that will not be great, but that's an artifact of PSP not having a great UX.

aojea:
> the alternative is the add a guide how to enable PSP after the initial kind cluster is already up and use the same steps in e2e tests. the UX around that will not be great, but that's an artifact of PSP not having a great UX.

is that possible? I thought that you have to enable the admission controller when the cluster is being created, 
```
{{ if .PodSecurityPolicy -}}
# enable PodSecurityPolicy admission controller
apiServerExtraArgs:
  enable-admission-plugins: PodSecurityPolicy
{{- end }}
```

I think that will be a better option too

neolit123:
see:
https://kubernetes.io/docs/concepts/policy/pod-security-policy/#enabling-pod-security-policies

you need to apply the policies and only then enable the controller.
there is an unfortunate chicken and egg scenario here, but it does work.

> enable-admission-plugins: PodSecurityPolicy

as mentioned a couple of times, we should preserve NodeRestriction:
> enable-admission-plugins: PodSecurityPolicy,NodeRestriction

aojea:
> you need to apply the policies and only then enable the controller.
> there is an unfortunate chicken and egg scenario here, but it does work.

Based on https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#how-do-i-turn-on-an-admission-controller I don't have clear how can we enable the controller once the cluster is deployed. However, seems that it can be done editing the kubeadm config and `kubeadm upgrade` based on this comment https://github.com/kubernetes/kubeadm/issues/762#issuecomment-385202158

I'm more than happy to change this PR removing the code and adding a document only, if you can show me the best way to enable the admission controller once the cluster is running. :)

neolit123:
> kubeadm upgrade based on this comment kubernetes/kubeadm#762 (comment)

using `kubeadm upgrade` should not be used for modifying static pods unless an actual upgrade is performed.

> I'm more than happy to change this PR removing the code and adding a document only, if you can show me the best way to enable the admission controller once the cluster is running. :)

as mentioned the PR alternatively can be extended by allowing the user to feed custom policy instead of always using a default.

the current recommended UX is:
- create a cluster
- apply a PSP to a running cluster
- go on all CP nodes and patch /etc/kubernetes/manifests/kube-apiserver.yaml to enable the admission controller and restart the api-servers.

what the above VOD showcases is the alternative (not recommended):
- create a cluster
- go on all CP nodes and patch /etc/kubernetes/manifests/kube-apiserver.yaml to enable the admission controller and restart the api-servers.
- apply a PSP to a running cluster
- unfortunately this might require manual interaction, to nudge the replication controller to recreate pods that the PSP previously blocked!

aojea:
> * go on all CP nodes and patch /etc/kubernetes/manifests/kube-apiserver.yaml to enable the admission controller and restart the api-servers

I got it now, this was the part I was missing :sweat_smile: 
If Ben or anyone else doesn't have any problem with this approach I will move forward with it

BenTheElder:
IMO it would be better if kubeadm provided the option to enable PSP w/ a default policy, bootstrapping has some precedence issues.

I haven't pushed on this one further due to it being non portable to hosts, and kind probably not being the best environment to test these.

BenTheElder:
PSP seems to be legacy and possibly not going GA, let's see how that conversation goes and circle back.

schnatterer:
Is there a source stating PSP as legacy? I've read [this comment](https://github.com/kubernetes-sigs/kind/issues/973#issuecomment-552205319) - are there any further more concrete sources?
Asking because we're planning on introducing PSPs don't want to waste a lot of effort on something soon to be deprecated.

aojea:
> Is there a source stating PSP as legacy? I've read [this comment](https://github.com/kubernetes-sigs/kind/issues/973#issuecomment-552205319) - are there any further more concrete sources?
> Asking because we're planning on introducing PSPs and don't want to waste a lot of efforts on something soon to be deprecated.

@schnatterer I think this is the main place where it is being discussed
https://github.com/kubernetes/enhancements/issues/5

BenTheElder:
It's unclear that it will be deprecated "soon" but it definitely does not
appear to be on GA track and the last word on it was leaning deprecated. I
don't think it can easily be removed and for production workloads it might
even be a good idea to go ahead and use, but for KIND it is problematic and
it doesn't look worth investing in that context right now.

Something like OPA may be more powerful but PSP is probably more widely
available on production cluster configurations.
I would ask SIG auth and cluster lifecycle in one of their communication
channels if you're looking for a more official answer, otherwise you can
see some context in
https://github.com/kubernetes/enhancements/issues/5#issuecomment-503672863

schnatterer:
Thanks @BenTheElder and @aojea for elaborating!

PR #727, opened 2019-07-21
Pass ipv6 argument to kubetest

aojea:
Depends on https://github.com/kubernetes/kubernetes/pull/80398

If the option to specify the IP family is enabled in the e2e testing framework, we should add it when testing IPv6 clusters


neolit123:
added a couple of quick questions to the framework PR.

aojea:
https://github.com/kubernetes/kubernetes/pull/80398#issuecomment-513954426

PR #730, opened 2019-07-23
Add an example of using extraMounts

rafax:
Modify user guide to add a section about using extraMounts.

### DIFF
--- a/site/content/docs/user/quick-start.md
+++ b/site/content/docs/user/quick-start.md
@@ -352,7 +352,8 @@ nodes:
 - role: control-plane
   extraMounts:
   - containerPath: /foo/path/on-the/node
-     hostPath: /path/on-the/host
+    hostPath: /path/on-the/host
+    readOnly: true
 ```
 
 ### Configure kind to use a proxy
--- a/site/content/docs/user/quick-start.md
+++ b/site/content/docs/user/quick-start.md
@@ -341,6 +341,20 @@ nodes:
 - role: worker
 ```
 
+#### Mounting files from host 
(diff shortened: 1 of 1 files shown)


rafax:
I updated my CNCF account to make sure CLA is correctly detected.

site/content/docs/user/quick-start.md
```diff
@@ -341,6 +341,20 @@ nodes:
...
+
+```yaml
+kind: Cluster
+apiVersion: kind.sigs.k8s.io/v1alpha3
+# mount a path from the host to the control-plane node
+nodes:
+- role: control-plane
+  extraMounts:
+  - containerPath: /foo/path/on-the/node
+     hostPath: /path/on-the/host
```
aojea: I think this should have the same indentation as `containerPath`

site/content/docs/user/quick-start.md
```diff
@@ -341,6 +341,20 @@ nodes:
...
+```
```
aojea: I would add below `hostPath`this line `readOnly: true`. It can produce undesirable results if itś writable by default

rafax: Added

site/content/docs/user/quick-start.md
rafax: Fixed, my bad.

BenTheElder:
thanks for the PR!

rather than make the quick start even longer, I've stubbed in a new page covering configuration specifically, while doing this I happened to cover extraMounts.

Let's move configuration details there

Issue #732, opened 2019-07-23
Update client-go to v12.0.0


shaneutt:
**What should be cleaned up or changed**:

Currently if you want to include `kind` as a library in another tool it will hold your `client-go` version at `v11.0.0` by default. The latest release for `client-go` is now [v12.0.0](https://github.com/kubernetes/client-go/releases/tag/v12.0.0) and it would be helpful to update kind to this version.

**Why is this needed**:

This makes it a bit easier to integrate with `kind` as a library.

BenTheElder:
Relatedly please know that the library form is significantly less stable, @munnerz and I have been working on a plan to clear that up, expect future versions of the library to change quite a bit. 

We export much more than we should.

shaneutt:
Thanks for the heads up @BenTheElder, currently I only have a lightweight integration. I will be cautious about going deeper with that integration. :+1:

BenTheElder:
We're still working on the library interface but it's actively coming along 😅

I beleive https://github.com/kubernetes-sigs/kind/pull/744 removed client-go from being tracked here.

shaneutt:
When trying current master it seems that #744 didn't stop client-go from getting pulled in at `v11.0.0+incompatible` when kind is used as a library.

In case others run into this I've just been using a simple workaround in my `go.mod` with the `replace` verb e.g.:

```
replace k8s.io/client-go v11.0.0+incompatible => k8s.io/client-go v0.0.0-20190819141724-e14f31a72a77
```

BenTheElder:
thanks!

I admit i've not imported a recent version of kind into another project yet, I'm currently working on refining the library surface and usability more.

I think the replace is probably the right solution long term, it looks like we'll want to actually start using client go for https://github.com/kubernetes-sigs/kind/issues/850 in the future possibly...

PR #733, opened 2019-07-23
Update client-go to v12.0.0

shaneutt:
Resolves https://github.com/kubernetes-sigs/kind/issues/732


aojea:
@BenTheElder it seems there are some PR updating go dependencies, should we aggregate all of them in one?

BenTheElder:
so I did some digging and actually we're only even importing client-go in the main module in order to get the homedir package which is pretty silly, we can just mimic that behavior and drop that dependency.

/assign

will tackle after https://github.com/kubernetes-sigs/kind/pull/737

BenTheElder:
client-go is now only an indirect dependency by way of kustomize and in such a way that it is not tracked in go.mod after https://github.com/kubernetes-sigs/kind/pull/744

I beleive that should replace this, please rebase and reopen if not.

Thanks for the PR!

BenTheElder:
client-go is now only an indirect dependency by way of kustomize and in such a way that it is not tracked in go.mod after https://github.com/kubernetes-sigs/kind/pull/744

I beleive that should replace this, please rebase and reopen if not.

Thanks for the PR!

PR #737, opened 2019-07-25
make internal API version internal

BenTheElder:
- make our internal API (= config types) version _actually_ internal
  - this includes conversions, scheme, etc.. consumers must use an external API version
  - update `cluster.Create(...)` options to make only using external API versions / types easier and improve defaulting
  - move code that depends on these types (in order to compile....) side benefit is the public go package surface is more accurate to what we semver (but incomplete, we will need to do more work here)
- update the public API to follow kubernetes patterns better (move to pkg/apis/config)

NOTE: We are also now close to enforcing that `cmd/` only use public / external types and methods that are suitable for library usage. We won't give ourselves special access to runtime / internal types either.


BenTheElder:
I am pretty unhappy with how large this change is, but I had to move quite a bit just to get everything to compile sanely after moving the internal API to internal.

xref: #737 (this is most of the work referenced there)

BenTheElder:
fixed verify :-)

aojea:
SGTM, I let @fabriziopandini and @neolit123 decide since they are heavy users of the API with `kinder`

neolit123:
> SGTM, I let @fabriziopandini and @neolit123 decide since they are heavy users of the API with kinder

kinder is a user of the internal config of kind, and i agree it should not be.
i think it will not cost much for that to be avoided and the only loss is validation.

WRT to validation, as thockin pointed out on a mailing list / issue thread somewhere a couple of days ago, validation should have been on the public side and is a flaw in api-machinery.

BenTheElder:
> SGTM, I let @fabriziopandini and @neolit123 decide since they are heavy users of the API with kinder

We can refactor as necessary later and kinder can third_party packages etc.

The internal API types should not be exported.

BenTheElder:
> WRT to validation, as thockin pointed out on a mailing list / issue thread somewhere a couple of days ago, validation should have been on the public side and is a flaw in api-machinery

Yeah, I've been watching this.

What we could maybe do for kind is create a public call to perform Validation specifically.

That said, Validation _is_ performed and returned on the only call to create already, the kind CLI must handle this the same way.

neolit123:
smarterclayton during the code-organization meeting today summarized the usage Go "internal" package names as "being a jerk". i don't agree 100%, but still had a LOL.

BenTheElder:
heh

IMHO if you can afford to have many repos what you should do is publish reusable modules in various repos with their own versioning and import them into your project. We aren't realistically going to create tons of repos for packages we write for kind though nor do we want the overhead of managing these correctly. A mature project with its own org can do this, but they will pay the penalty to maintain those packages for reuse.

If you _don't_ use internal packages, then you either can't make breaking changes, or you're not versioning your module in a consumable way, IMHO not following semver (once you hit stable, 0.X is arguably different) is being a bigger jerk to your users by lying to them...

Since the code is open source, it can always be copied out if someone _really_ wants to reuse a snippet for some reason, but acknowledging the lack of guarantees.

Go itself even uses some `.*/internal/.*` packages.

BenTheElder:
to be clear: the intention is to implement `kind` the binary using public packages and bring those to the same semver as the CLI (and drive both towards being more stable), and enforce the ability to refactoring E.G. to support ignite without breaking the CLI or consumers trying to test with kind.

internal keeps us honest. we're locking `cmd/kind` out of depending on implementation details versus intentional APIs.

PR #738, opened 2019-07-26
fix image sideloading during node build

BenTheElder:
/priority critical-urgent

thankfully I don't think this is in 0.4.0, but at HEAD we are not sideloading images we didn't build (!)

TODO: clean up this package

also: this now only warns if it can't pull an image. it will **fail** only when it can't save the image (IE it will fall back to a local already existing copy if there is one), this is useful for offline development or not-yet-published kindnetd versions etc.

### DIFF
--- a/pkg/build/node/node.go
+++ b/pkg/build/node/node.go
@@ -460,7 +460,7 @@ func (c *BuildContext) prePullImages(dir, containerID string) error {
 				fmt.Printf("Pulling: %s\n", image)
 				err := docker.Pull(image, 2)
 				if err != nil {
-					return err
+					log.Warnf("Failed to pull %s with error: %v", image, err)
 				}
 				// TODO(bentheelder): generate a friendlier name
 				pullName := fmt.Sprintf("%d.tar", i)
@@ -469,7 +469,7 @@ func (c *BuildContext) prePullImages(dir, containerID string) error {
 				if err != nil {
 					return err
 				}
-				pulledImages <- fmt.Sprintf("/build/bits/images/%s", pullName)
+				pulledImages <- pullTo
 			}
 			return nil
 		})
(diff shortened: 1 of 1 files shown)


pkg/build/node/node.go
```diff
@@ -134,10 +134,10 @@ func NewBuildContext(options ...Option) (ctx *BuildContext, err error) {
 func (c *BuildContext) Build() (err error) {
 	// ensure kubernetes build is up to date first
 	log.Infof("Starting to build Kubernetes")
-	if err = c.bits.Build(); err != nil {
+	/*if err = c.bits.Build(); err != nil {
```
BenTheElder: remove

BenTheElder: done

BenTheElder:
we need to bump the conformance image job up to pick up this fix

pkg/build/node/node.go
```diff
@@ -469,7 +469,7 @@ func (c *BuildContext) prePullImages(dir, containerID string) error {
 				if err != nil {
 					return err
 				}
-				pulledImages <- fmt.Sprintf("/build/bits/images/%s", pullName)
+				pulledImages <- pullTo
```
aojea: what is this `pullTo` doing? 😅

pkg/build/node/node.go
```diff
@@ -460,7 +460,7 @@ func (c *BuildContext) prePullImages(dir, containerID string) error {
 				fmt.Printf("Pulling: %s\n", image)
 				err := docker.Pull(image, 2)
 				if err != nil {
-					return err
+					log.Warnf("Failed to pull %s with error: %v", image, err)
```
aojea: if we are not able to pull one of the required images we don´t fail?
 will not fail later trying to setup the cluster?

aojea:
/lgtm
sorry for the dumb questions

Issue #710, opened 2019-07-16
WSLv1: No DNS between pods or to the internet


wilmardo:
**What happened**:

There is no DNS within the cluster, pods cannot resolve services within the cluster or access the internet on hostnames.

**What you expected to happen**:

DNS working to the internet and to services within the cluster

**How to reproduce it (as minimally and precisely as possible)**:

I am running kind in WSL (v1) setup to run containers in Docker for Desktop installed within Win10. Docker for Desktop exposes the daemon over tcp and the DOCKER_HOST env var within WSL is set to `tcp://127.0.0.1:2375`.

I used the following configuration
```yaml
kind: Cluster
apiVersion: kind.sigs.k8s.io/v1alpha3
nodes:
- role: control-plane
- role: worker
  extraPortMappings:
  # Home Assistant
  - containerPort: 31123
    hostPort: 31123
    listenAddress: 127.0.0.1
  # Mosquitto
  - containerPort: 31883
    hostPort: 31883
    listenAddress: 127.0.0.1
  # Traefik
  - containerPort: 30080
    hostPort: 30080
    listenAddress: 127.0.0.1
  # Traefik
  - containerPort: 30443
    hostPort: 30443
    listenAddress: 127.0.0.1
```

**Anything else we need to know?**:

I am experiencing a very similar situation on my home arm64 cluster so this might not be related to kind and more to kubernetes. I am not to the bottom of this issue yet but the symptoms (no DNS) are the same.

I updated the CoreDNS configmap the following to see the logs and I updated the forward to 9.9.9.9 to ensure it isn't a loopback issue in /etc/resolv.conf ([link](https://coredns.io/plugins/loop/#troubleshooting-loops-in-kubernetes-clusters)):

```yaml
apiVersion: v1
data:
  Corefile: |
    .:53 {
        log
        errors
        health
        kubernetes cluster.local in-addr.arpa ip6.arpa {
           pods insecure
           upstream
           fallthrough in-addr.arpa ip6.arpa
           ttl 30
        }
        prometheus :9153
        forward . 9.9.9.9
        cache 30
        loop
        reload
        loadbalance
    }
kind: ConfigMap
metadata:
  creationTimestamp: "2019-07-16T13:00:43Z"
  name: coredns
  namespace: kube-system
  resourceVersion: "1234"
  selfLink: /api/v1/namespaces/kube-system/configmaps/coredns
  uid: 862c0677-19d5-48ee-8647-1b962737e2c9
``` 

[...]

PR #739, opened 2019-07-26
don't masquerade service subnet

BenTheElder:
this is incomplete. see https://github.com/kubernetes/kubernetes/issues/80243#issuecomment-515303818 for context

basically:
- we need to be not masquerading the service IP range (!)
- add the ability to get this from an environment variable (TODO: make this more generally flexible and figure out something with the templating ... this may be tricky without breaking existing images :/)
- add the ability in kindnetd to just obtain the CIDRs from the kubeadm config map at runtime
- don't masquerade any of these CIDRs

TODO:
- We can get the CIDRs also from the api-server flags (services CIDR) and from kube-proxy (clusterCIDR) component config. This version would actually be super neat and pretty ideal as kindnetd could be deployed to clusters without any configuration :D (xref https://github.com/kubernetes/kubernetes/issues/46508) ... if we do this we can just drop the templating. we can tighten the RBAC to just have get to these configmaps.


BenTheElder:
/hold
this is still hacky and I have not pushed the images for the moment, this is the rough approach though.

imho the getting configmap seems to be supportable trivially back to kubernetes 1.11 for kubeadm at least. it's a bit weird to configure kubeadm and then have kindnetd re-obtain that config at runtime, but it just drops in and works without needing to plumb it.

if we went all in on the runtime configmap hijinks, then we can remove the templating entirely and just make kindnetd a simple manifest (and one that could probably be deployed to other toy clusters, not that this is important...)

BenTheElder:
will follow up in the morning with cleanup and a look at kube-proxy's configmap and how we'd get the kube-apiserver `service-cluster-ip-range` flag in kindnetd at runtime as a fallback.

cmd/kindnetd/subnets.go
```diff
@@ -0,0 +1,104 @@
...
+
+	return s.All()
+}
+
+
+
+
+const (
+	kubeadmServiceCIDRRegex = `\s*serviceSubnet: (.*)\n`
+	kubeadmClusterCIDRRegex = `\s*podSubnet: (.*)\n`
```
BenTheElder: yes regex again. simple regexes work against all kubeadm configs that I am aware of. the field names and field types never changed so far, but the types etc. have moved around a bit. I took a quick look at 1.11 ... 1.15

we'll of course need to improve this for dualstack, but this is hardly the only thing that will need fixing

cmd/kindnetd/subnets.go
```diff
@@ -0,0 +1,104 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+type subnets struct {
+	pods sets.String
+	services sets.String
+}
```
BenTheElder: this is not really a baked abstraction, but I think we will need at least one CIDR for each of these categories while detecting them.

the rest of the code just needs the whole list, or handles different CIDRs (like the per-node pod CIDRs)

wilmardo:
Also related is the original issue: #710

pkg/build/node/cni.go
```diff
@@ -97,6 +97,12 @@ rules:
     verbs:
       - list
       - watch
+  - apiGroups:
```
aojea: just unrelated to this change but can you remove the apparmor profiles?
```
    apparmor.security.beta.kubernetes.io/allowedProfileNames: runtime/default
    apparmor.security.beta.kubernetes.io/defaultProfileName: runtime/default	    apparmor.security.beta.kubernetes.io/defaultProfileName: runtime/default
```

aojea:
/hold
If you don't mind I'd like to go deeper into this. There are iptables rules for services and want to be sure we are not going to affect them.

aojea:
I created an environment with @wilmardo´s config 1 control-plane and 1 worker

If I dump the iptables rules we can observe the following chains (for ref on how iptables process the rules https://stuffphilwrites.com/2014/09/iptables-processing-flowchart/)

```
-A PREROUTING -m comment --comment "kubernetes service portals" -j KUBE-SERVICES
-A OUTPUT -m comment --comment "kubernetes service portals" -j KUBE-SERVICES
-A POSTROUTING -m comment --comment "kubernetes postrouting rules" -j KUBE-POSTROUTING
-A POSTROUTING -m comment --comment "kind-masq-agent: ensure nat POSTROUTING directs all non-LOCAL destination traffic to our custom KIND-MASQ-AGENT chain" -m addrtype ! --dst-type LOCAL -j KIND-MASQ-
AGENT
```

one the pod sends the UDP request to the DNS server 10.96.0.10 and it goes to PREROUTING that should match
`-A KUBE-SERVICES -d 10.96.0.10/32 -p udp -m comment --comment "kube-system/kube-dns:dns cluster IP" -m udp --dport 53 -j KUBE-SVC-TCOU7JCQXEZGVUNU`

and this goes to
`-A KUBE-SVC-TCOU7JCQXEZGVUNU -m statistic --mode random --probability 0.50000000000 -j KUBE-SEP-YIL6JZP7A3QYXJU2`

and 
```
-A KUBE-SEP-YIL6JZP7A3QYXJU2 -s 10.244.0.2/32 -j KUBE-MARK-MASQ
-A KUBE-SEP-YIL6JZP7A3QYXJU2 -p udp -m udp -j DNAT --to-destination 10.244.0.2:53
````

this will mean that the packet that reaches the POSTROUTING rule will not be directed to the 10.96.0.10 service ip, it will be using 10.244.0.2 as we see in the last rule,
If I didn´t overlook anything I think that we don´t need to NAT the service subnet, actually, if that were the case I don´t know how the e2e tests are passing and this is not happening to more people.

File: cmd/kindnetd/subnets.go [hunk 4]
```diff
@@ -0,0 +1,104 @@
...
+	podSubnetEnv := os.Getenv("POD_SUBNET")
+	if podSubnetEnv != "" {
+		s.pods.Insert(podSubnetEnv)
+	}
+	serviceSubnetEnv := os.Getenv("SERVICE_SUBNET")
+	if serviceSubnetEnv != "" {
+		s.services.Insert(serviceSubnetEnv)
+	}
+
+	// try getting from kubeadm config
```
aojea: I don´t know if it´s better to fall back only if we didn´t get it from the env variable, don´t like the idea of having multiple sources of truth for the subnets

BenTheElder: agree

cmd/kindnetd/subnets.go
```diff
@@ -0,0 +1,104 @@
...
+
+package main
+
+import (
+	"os"
+	"regexp"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/apimachinery/pkg/util/sets"
```
aojea: we should use this https://github.com/kubernetes/utils/blob/master/net/ipnet.go  instead

cmd/kindnetd/subnets.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (s *subnets) All() []string {
+	return s.pods.Union(s.services).List()
+}
+
+
+func getNoMasqueradeSubnets(clientset *kubernetes.Clientset) []string {
+	s := newSubnets()
+
+	// check for environment variables (legacy)
```
aojea: maybe we should start considering dual stack, this function it will accept 1 cidr or multiple in a comma separated list https://github.com/khenidak/kubernetes/blob/0fbfa755d3960eaa0ab3ea38453a934d2188e7f8/cmd/cloud-controller-manager/app/core.go#L144

cmd/kindnetd/subnets.go
aojea: regex are cool :)

aojea:
/hold cancel
after taking with Ben I agree with masquerading the services subnet,

BenTheElder:
/hold
:^)
revisiting how this is implemented exactly

aojea:
we should close these, I moved the commit with the important bits to my dual-stack kindnet PR keeping the author

BenTheElder:
> we should close these, I moved the commit with the important bits to my dual-stack kindnet PR keeping the author

your dual stack PR is a lot more involved and unlikely to merge quickly with the remaining dual stack issues, I'd like to get an updated version of just:
- appropriate masquerade
- reading from kubeadm config instead of templateing
and land that for 0.6 images, because that has some compatibility implications and is a smaller change

cmd/kindnetd/subnets.go
```diff
@@ -0,0 +1,104 @@
...
+	podSubnetEnv := os.Getenv("POD_SUBNET")
```
BenTheElder: we should just ignore the env.

pkg/build/node/cni.go
BenTheElder: we probably shouldn't? people are still trying to get psp to work

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/990

PR #740, opened 2019-07-27
Tune haproxy configuration

aojea:
Current haproxy configuration declared the session closed after 50 sec if no packet was received from the client or from the server.
This caused issues, per example, when trying to tail logs with 

```kubectl logs -f nginx-554b9c67f9-r9ls6
...
127.0.0.1 - - [26/Jul/2019:21:20:28 +0000] "GET /favicon.ico HTTP/1.1" 404 555 "http://localhost:8000/" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/75.0.3770.142 Safari/537.36" "-"
F0726 22:21:48.325924    1750 helpers.go:114] error: unexpected EOF
```

For long lived connections the user has to define the appropriate values to his environment, we can safely define 1h in kind, is unlikely that we can DOS the server with the number of open connections.
```
  timeout client 1h
  timeout server 1h
```

However, in other environments, and specifically in kind, there are more things that can force a disconnection. Keep in mind that in kind, kubectl connects to the exposed port on localhost, so our TCP connections can be closed by the conntrack timeouts of the host. We can set the `option tcpka`,  enabling socket-level TCP keep-alives, that makes the system regularly send packets
to the other end of the connection, leaving it active. The configurations parameters of the TCP keepalives depends on the operations systems, and these keepalives doesn´t prevent the haproxy session to be expired.

Ref: https://github.com/kubernetes/kubeadm/issues/1685


pkg/internal/cluster/loadbalancer/config.go
```diff
@@ -33,18 +33,21 @@ type ConfigData struct {
 // DefaultConfigTemplate is the loadbalancer config template
 const DefaultConfigTemplate = `# generated by kind
 global
+  stats socket /var/run/haproxy.sock mode 600 level admin
```
aojea: this is useful for debugging, installint socat or netcat in the haproxy container, you can check the sessions status per example

````
/ # echo "show sess" | socat  /var/run/haproxy.sock stdio
0x5655011ce8a0: proto=tcpv4 src=172.17.0.1:35988 fe=control-plane be=kube-apiservers srv=ha-control-plane ts=00 age=14m54s calls=3 rate=0 cpu=0 lat=0 rq[f=848202h,i=0,an=00h,rx=45m11s,wx=,ax=] rp[f=80048202h,i=0,an=00h,rx=45m11s,wx=,ax=] s0=[8,200008h,fd=32,ex=] s1=[8,200018h,fd=33,ex=] exp=45m5s
0x5655011cd880: proto=unix_stream src=unix:1 fe=GLOBAL be=<NONE> srv=<none> ts=00 age=0s calls=2 rate=2 cpu=0 lat=0 rq[f=c4c220h,i=0,an=00h,rx=,wx=,ax=] rp[f=80008002h,i=0,an=00h,rx=,wx=,ax=] s0=[8,280008h,fd=34,ex=] s1=[8,204018h,fd=-1,ex=] exp=2m
```

aojea:
cc @ereslibre

pkg/internal/cluster/loadbalancer/config.go
```diff
@@ -33,18 +33,21 @@ type ConfigData struct {
...
+  stats timeout 2m
   log /dev/log local0
   log /dev/log local1 notice
   daemon
 
 defaults
   log global
   mode tcp
+  option tcpka
```
ereslibre: I'm still seeing the same behavior with `clitcpka` and `srvtcpka` (equivalent to `tcpka`), without tweaking timeouts.

aojea: the `tcpka` avoids that the intermediate systems closes the connection (in our case iptables since we are doing docker port forwarding).
The `kubectl` commands will always die after the inactivity timeout we configure in haproxy if there is no packet between the server and the client.
To avoid this, kubectl should keep the connection open

aojea: 🤔 seems that `TCP keep alive doesn't kick in until the connection has been idle for 2 hours.` by default

pkg/internal/cluster/loadbalancer/config.go
BenTheElder: SGTM

pkg/internal/cluster/loadbalancer/config.go
```diff
@@ -33,18 +33,21 @@ type ConfigData struct {
...
   option dontlognull
   # TODO: tune these
   timeout connect 5000
-  timeout client 50000
```
BenTheElder: what unit where these in?

aojea: milliseconds https://cbonte.github.io/haproxy-dconv/1.7/configuration.html#4-timeout%20server

BenTheElder: So 50s to an hour.

Seems arbitrarily large, how do other setups compare (eg kubespray)?

BenTheElder:
[I've just caught up on the linked issue]

pkg/internal/cluster/loadbalancer/config.go
aojea: is to be safe, maybe exaggerated, but don't think that anybody is going to exhaust the host with open tcp connections with kind :)

BenTheElder: fair I supposed, I'm not sure this is the correct or right fix, but I'm OK to try it, mostly seems harmless

BenTheElder: mild concern that we should not do non-production ready configs where avoidable.

I feel like the actual bug here is in whatever is talking to the API server without any data flowing or any reconnects, I'm not a fan of masking a bug there.

BenTheElder:
> Current haproxy configuration declared the session closed after 50 sec if no packet was received from the client or from the server.


when is this happening? what client? what use case?

pkg/internal/cluster/loadbalancer/config.go
aojea: is hard to say if it is a bug or a limitation, the software does what's supposed to do, just doesn`t implement keepalives to avoid the intermediate devices closes the connection, all of the involved components can be blamed, is a common TCP problem with long lived connections :man_shrugging:

aojea:
> when is this happening? what client? what use case?
Using those commands
https://github.com/kubernetes-sigs/kind/pull/740#pullrequestreview-267483222

haproxy opens 2 sockets and forward traffic between them, if haproxy doesn't see any traffic it expires the session. The OS TCP keepalives are useless because they work at Kernel level (ending in each of the sockets open by haproxy) and haproxy doesn't see them, hence the connection timeout is not refreshed.

pkg/internal/cluster/loadbalancer/config.go
ereslibre: @BenTheElder I am going to finish [this patch](https://github.com/kubernetes/kubernetes/compare/master...ereslibre:ping-on-long-running-commands) and open a PR. It covers `kubectl exec -it`, `kubectl logs -f` and `kubectl get -w`. The idea is to ping the other end on a regular basis.

BenTheElder: thanks @ereslibre! + 💯

BenTheElder:
> haproxy opens 2 sockets and forward traffic between them, if haproxy doesn't see any traffic it expires the session. The OS TCP keepalives are useless because they work at Kernel level (ending in each of the sockets open by haproxy) and haproxy doesn't see them, hence the connection timeout is not refreshed.


that doesn't really answer the questions:
- which client? 
  - answer: kubectl, any kubernetes client ...
- what use case?
  - answer: `kubectl logs`, apparently when there's no output there's no packets

https://github.com/kubernetes-sigs/kind/pull/740#discussion_r310918484 sounds like the right solution to me, this is an upstream issue that should be fixed there. 

we wouldn't want people using this fix on a real cluster, and there's nothing special about kind in this respect.

aojea:
any other tools or applications that uses haproxy and not send traffic in 50 sec will have the same problem, however I agree with you it should be solved at the app level

BenTheElder:
imho that's the kind of bug you want to find testing your app :^)

pkg/internal/cluster/loadbalancer/config.go
ereslibre: The patch is almost ready. I will review it tomorrow, checking that I am not breaking anything, rereading for minor changes and I think I should be good to open a PR and ping all stakeholders.

ereslibre: Side note, when testing manually I go to the extreme and use this patch in kind: https://github.com/kubernetes-sigs/kind/compare/master...ereslibre:low-timeout, it's much more convenient to force the issue to ensure that pings are working as intended.

aojea:
we can reopen in necessary, seems there is no needed

PR #743, opened 2019-07-31
more refactoring

BenTheElder:
- start eliminating the catch-all util package in favor of scoped utility packages
- move unstable build/kube package (build/node should be preferred)
- move patches to the config types instead of the kustomize package, this makes it possible to change the patch types

/cc @amwat


aojea:
this is progress :)

PR #744, opened 2019-07-31
reduce client-go dependency, upgrade kustomize

BenTheElder:
- stop depending on client-go directly for a single simple method (`HomeDir`)
- upgrade kustomize to the latest release, majorly reducing the versions we need in go.mod

NOTE: we still transitively depend on many more packages than are listed go.mod, however we should be less dependent on the particular versions which should make kind easier to use as a library (in theory)


pkg/internal/util/kustomize/kustomize.go
```diff
@@ -91,9 +95,12 @@ func Build(resources, patches []string, patchesJSON6902 []config.PatchJSON6902)
...
 	// now we can build the kustomization
 	var out bytes.Buffer
-	f := k8sdeps.NewFactory()
-	cmd := build.NewCmdBuild(&out, memFS, f.ResmapF, f.TransformerF)
-	cmd.SetArgs([]string{fakeDir})
+	uf := kunstruct.NewKunstructuredFactoryImpl()
+	pf := transformer.NewFactoryImpl()
+	rf := resmap.NewFactory(resource.NewFactory(uf), pf)
+	v := validator.NewKustValidator()
+ 	cmd := build.NewCmdBuild(&out, memFS, v, rf, pf)
```
BenTheElder: https://github.com/kubernetes-sigs/kustomize/blob/3d2e956b1912c8b41b74b139f197d89e8f3ad56f/pkg/commands/commands.go#L38-L45

pkg/internal/util/env/homedir.go
```diff
@@ -0,0 +1,96 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package env
+
+// NOTE this is from client-go. Rather than pull in client-go for this one
+// standalone method, we have an (unmodified) fork here.
+// https://github.com/kubernetes/client-go/blob/6d7018244d72350e2e8c4a19ccdbe4c8083a9143/util/homedir/homedir.go
```
BenTheElder: https://github.com/kubernetes/client-go/blob/6d7018244d72350e2e8c4a19ccdbe4c8083a9143/util/homedir/homedir.go

aojea:
heheh nice trick

PR #749, opened 2019-08-02
iterate on the roadmap

BenTheElder:
capturing some points from discussing this in various forums since it was last updated, updating items that are complete, etc.

I think we'll definitely need another pass over this, but it's at least more accurate than the previous version

cc @spiffxp

### DIFF
--- a/site/content/docs/contributing/1.0-roadmap.md
+++ b/site/content/docs/contributing/1.0-roadmap.md
@@ -15,25 +15,43 @@ as a project.
 
 High level goals for kind include:
 
-- [ ] Reach [1.0] status / 
-stability by the end of Q1 2019 / the Kubernetes 1.14 release cycle
-- [x] Support multi-node clusters - [#117]
-- [ ] Support offline / air-gapped clusters
-  - [ ] pre-loaded / offline CNI - [#200]
-- [ ] Support mounting host directories - [#62]
-- [ ] Improve Windows support
-  - [x] add Windows binaries to releases - [#155]
-  - [ ] improve instructions for KUBECONFIG in particular
-- [ ] Improve documentation - In Progress ...
-  - [ ] expand examples of using kind
-  - [ ] cover known issues, debugging, work-arounds, etc.
-  - [X] create a documentation site - [#268]
-- [ ] Improving logging and debuggability
-  - [ ] enable audit-logging
-- [ ] Support non-amd64 architectures (namely ARM) - [#166]
-- [ ] Automated publishing of Kubernetes release based kind "node" images - [#197]
+## Beta Requirements
 
-**Non**-Goals include:
+To reach "beta" [grade][deprecation-policy] kind needs to at minimum:
+
+- [x] Improve documentation (Though this will eternally be "In Progress" !)
+  - [X] create a documentation site - [#268]
+  - [ ] expand examples of using kind
+  - [x] cover known issues, debugging, work-arounds, etc.
+- [x] Reliably [pass][kind-conformance-dashboard] the Kubernetes [conformance tests]
+ - [x] [Certify][certified] Conformance
+- [x] Support multi-node clusters - [#117]
+- [x] Support offline / air-gapped clusters
+  - [x] pre-loaded / offline CNI - [#200]
+- [x] Support mounting host directories - [#62]
+- [x] Improve Windows support
+  - [x] add Windows binaries to releases - [#155]
+  - [x] improve instructions for KUBECONFIG in particular
+- [ ] Support usage as a properly versioned, supported, and documented library. This includes following semver without every release being a major / breaking change to the API (which must be extensible without breakage), ensuring the CLI only uses a suitable public library surface, documentation and examples for the library, versioned types for public APIs (E.G. config format), etc.
+- [ ] should be possible to troubleshoot kind without needing to modify kind or use external debugging tools
+  - [ ] consistent logging (what is logged, when should it be logged, what levels are used)
+  - [ ] errors have appropriate context
+for managing clusters in test harnesses
+- [ ] move API types / labels from `*.k8s.io` into [`*.x-k8s.io`][api-review-voluntary]
+- [x] Support all currently [upstream-supported Kubernetes versions][version-skew-policy]
+
+## GA Requirements
+
+To reach "GA" [grade][deprecation-policy] kind needs to at minimum:
+
+- [ ] Support non-AMD64 architectures (namely ARM) - [#166]
+- [ ] Automated publishing of Kubernetes release based kind "node" images - [#197]
+- [ ] Support for runtimes other than docker/default including podman, ignite etc.
+- [ ] Enable audit-logging
+- [ ] First class support for skewed node (Kubernetes) versions
+- ... TBD, more will be added here ...
+
+## Non-Goals
 
 - Supporting every possible Kubernetes configuration
   - In order to best support offline / hermetic clusters, we will likely not
@@ -50,9 +68,8 @@ Longer Term goals include:
 
 - Enabling a suitable local storage provider for testing applications that need
 persistent storage
-- Improving "kind as a library"
 
-Misc:
+## Misc:
 
 - [x] setup a regular Zoom meeting for the project [#244]
 - [x] achieve certified Kubernetes conformance [#245]
@@ -60,6 +77,12 @@ Misc:
 Other goals / tasks not listed here can be found both in [the 1.0 project] and 
 more generally triaged for rough-priority in the [GitHub issues].
 
+[kind-conformance-dashboard]: https://testgrid.k8s.io/conformance-kind
+[version-skew-policy]: https://kubernetes.io/docs/setup/release/version-skew-policy/
+[deprecation-policy]: https://kubernetes.io/docs/reference/using-api/deprecation-policy/
+[certified]: https://github.com/cncf/k8s-conformance/pull/451#issuecomment-457663982
+[conformance tests]: https://github.com/cncf/k8s-conformance
+[api-review-voluntary]: https://github.com/kubernetes/community/blob/master/sig-architecture/api-review-process.md#voluntary
 [1.0]: https://github.com/kubernetes-sigs/kind/projects/1
 [the 1.0 project]: https://github.com/kubernetes-sigs/kind/projects/1
 [GitHub issues]: https://github.com/kubernetes-sigs/kind/issues


site/content/docs/contributing/1.0-roadmap.md
```diff
@@ -15,25 +15,41 @@ as a project.
...
 
-- [ ] Reach [1.0] status / 
-stability by the end of Q1 2019 / the Kubernetes 1.14 release cycle
+## Beta Requirements
+
+To reach "beta" [grade][deprecation-policy] kind needs to at minimum:
+
+- [x] Improve documentation (Though this will eternally be "In Progress" !)
+  - [X] create a documentation site - [#268]
+  - [ ] expand examples of using kind
```
aojea: I think that @mauilion , and our community was very active in this regard, i.e.
https://github.com/charlesakalugwu/kind-examples
https://github.com/kind-ci/examples

site/content/docs/contributing/1.0-roadmap.md
```diff
@@ -15,25 +15,41 @@ as a project.
...
+- [ ] move API types / labels from `*.k8s.io` into [`*.x-k8s.io`][api-review-voluntary]
+- [x] Support all currently [upstream-supported Kubernetes versions][version-skew-policy]
+
+## GA Requirements
+
+To reach "GA" [grade][deprecation-policy] kind needs to at minimum:
+
+- [ ] Support non-AMD64 architectures (namely ARM) - [#166]
 - [ ] Automated publishing of Kubernetes release based kind "node" images - [#197]
+- [ ] Support for runtimes other than docker/default including podman, ignite etc.
```
aojea: I would reword this since this the definition is very open,  something along the lines "Implement an interface to support other runtimes than docker/default like podman, ignite etc."

site/content/docs/contributing/1.0-roadmap.md
```diff
@@ -15,25 +15,41 @@ as a project.
...
+- [ ] Enable audit-logging
+- [ ] First class support for skewed node (Kubernetes) versions
+- ... TBD, more will be added here ...
```
aojea: This is redundant with L76 
https://github.com/kubernetes-sigs/kind/pull/749/files#diff-7a7e6ffc98134bc2043eafd337187bc3R76

site/content/docs/contributing/1.0-roadmap.md
BenTheElder: open is fine. we'll iterate.

site/content/docs/contributing/1.0-roadmap.md
```diff
@@ -15,25 +15,41 @@ as a project.
...
-  - [ ] enable audit-logging
-- [ ] Support non-amd64 architectures (namely ARM) - [#166]
+  - [x] improve instructions for KUBECONFIG in particular
+- [ ] Support usage as a properly versioned, supported, and documented library 
+  - [ ] Improving logging and debuggability 
+for managing clusters in test harnesses
+- [ ] move API types / labels from `*.k8s.io` into [`*.x-k8s.io`][api-review-voluntary]
+- [x] Support all currently [upstream-supported Kubernetes versions][version-skew-policy]
+
+## GA Requirements
```
aojea: I will add a target of code coverage, at least 70% per example

site/content/docs/contributing/1.0-roadmap.md
BenTheElder: It's important to note here that this is not the full GA requirements.

site/content/docs/contributing/1.0-roadmap.md
BenTheElder: yes, but we need to cover some of them in our docs, especially as searching the internet for this can be hard :-)

aojea: well, that´s a good challenge for any SEO specialist, with a project named `kind` :)

aojea:
SGTM , just some food for thought :)

site/content/docs/contributing/1.0-roadmap.md
BenTheElder: Yeah, not sure what we're shooting for there, I would say 100% but I'd rather have a more informed idea first 😓 

The GA requirements in particular are incomplete which is why I added the last bullet 😅

aojea: Nah, my experience says that above 70% is fine, is mostly just to avoid aggressive refactoring and regressions.
We can reword as "implement gating based on code coverage"

BenTheElder: this could be a beta requirement even, it needs more thought :-)

e2e coverage may be more important than unit even though we want unit, it's not obvious that unit test coverage will actually prevent eg incorrectly creating the containers.

site/content/docs/contributing/1.0-roadmap.md
```diff
@@ -15,25 +15,41 @@ as a project.
...
+
+To reach "GA" [grade][deprecation-policy] kind needs to at minimum:
+
+- [ ] Support non-AMD64 architectures (namely ARM) - [#166]
 - [ ] Automated publishing of Kubernetes release based kind "node" images - [#197]
+- [ ] Support for runtimes other than docker/default including podman, ignite etc.
+- [ ] Enable audit-logging
```
tao12345666333: I want to know the specific meaning here, does it mean https://kubernetes.io/docs/tasks/debug-application-cluster/audit/ or does Kind's own audit log?

BenTheElder: Kubernetes audit logging. Also this was already on the roadmap 🙃

site/content/docs/contributing/1.0-roadmap.md
aojea: I´m a big fan of the Test pyramid concept https://martinfowler.com/articles/practical-test-pyramid.html, every testing layer has its importance. I think unit tests will be important once we start to add integrations like podman, ignite, and the public API moves to beta

site/content/docs/contributing/1.0-roadmap.md
```diff
@@ -15,25 +15,41 @@ as a project.
...
-  - [ ] improve instructions for KUBECONFIG in particular
-- [ ] Improve documentation - In Progress ...
-  - [ ] expand examples of using kind
-  - [ ] cover known issues, debugging, work-arounds, etc.
-  - [X] create a documentation site - [#268]
-- [ ] Improving logging and debuggability
-  - [ ] enable audit-logging
-- [ ] Support non-amd64 architectures (namely ARM) - [#166]
+  - [x] improve instructions for KUBECONFIG in particular
+- [ ] Support usage as a properly versioned, supported, and documented library
```
neolit123: @BenTheElder @aojea 

something i wanted to recommend (without logging a ticket) is to not employ the kube-proxy pattern where an alpha API version grows with new features being added without bumping the API version.

in kubeadm we switched away from that somewhere after i joined, yet there were still cases when someone just approved changes on the old API. this is a bad pattern and those who import the kind public types would expect them to not change without a version bump. there are other areas of k8s that are guilty of this - e.g. some `v1` areas that don't want to introduce a `v2alpha1`.

this however means that new features have to be accumulated in the new API version that is not released yet.
e.g. 0.5.0 -> v1alpha4.

site/content/docs/contributing/1.0-roadmap.md
```diff
@@ -15,25 +15,41 @@ as a project.
...
-  - [ ] cover known issues, debugging, work-arounds, etc.
-  - [X] create a documentation site - [#268]
-- [ ] Improving logging and debuggability
-  - [ ] enable audit-logging
-- [ ] Support non-amd64 architectures (namely ARM) - [#166]
+  - [x] improve instructions for KUBECONFIG in particular
+- [ ] Support usage as a properly versioned, supported, and documented library 
+  - [ ] Improving logging and debuggability 
+for managing clusters in test harnesses
+- [ ] move API types / labels from `*.k8s.io` into [`*.x-k8s.io`][api-review-voluntary]
```
neolit123: @BenTheElder 
have you experimented with converters related to this change? TMK api-machinery does not support converting between types in a different Group. one way to solve this might be to keep types registered under both Groups, until the old API (`*.k8s.io) is deprecated.

the cluster API will do this (for alpha2 or 3) too but they use conversion web hooks, so they can tolerate such a change. given the challenges i see, i don't think we want to do that for projects like kops and kubeadm that have more mature APIs and user base.

i know that CertManager also uses `k8.io` (cc @munnerz) and SIG arch were giving it as an example of an external project that uses the "reserved" k8s.io domain.

BenTheElder: For kops and kubeadm I'm not sure, for kind we're early enough that it can even just be a breaking change if it comes down to it.

We have a lot of options given that these are not CRDs and the conversion only happens internally to Create(), shouldn't be a real problem.

I'm also increasingly convinced that for our use case the API machinery is more trouble than it's worth, we could implement conversion etc without it easily and already did once in a previous version of the code. Would probably keep deepcopy and skip the rest.

For the docker labels we will temporarily handle both but mark new things with the new one.

site/content/docs/contributing/1.0-roadmap.md
BenTheElder: Even go itself adds new fields to structs I thought, they just must have sane defaults and the type accepted everywhere.

Will consider though.

site/content/docs/contributing/1.0-roadmap.md
BenTheElder: I don't think the unit tests will give use useful coverage of those backends. It will give us some coverage of the layers above. But in practice most of our bugs have been in the lower layers

site/content/docs/contributing/1.0-roadmap.md
neolit123: > I'm also increasingly convinced that for our use case the API machinery is more trouble than it's worth, we could implement conversion etc without it easily and already did once in a previous version of the cod

the folks at WG Component Standard are also partially convinced at this point that API machinery is not really well suited for component configs WRT to validation and conversation.

yet the the WG is also driven by SIG api-machinery, so not sure how things would evolve there.

site/content/docs/contributing/1.0-roadmap.md
BenTheElder: Kubernetes has some very extensive rules for this fwiw, we do our best to follow those.

site/content/docs/contributing/1.0-roadmap.md
spiffxp: Please don't over-specify here, especially when we're not yet in beta.

A testing plan yes, that includes some level of unit and e2e sure, but the details of it can be spelled out after kind reaches beta.

site/content/docs/contributing/1.0-roadmap.md
spiffxp: Enable by default? Or support enablement of audit logging as an option?

site/content/docs/contributing/1.0-roadmap.md
spiffxp: I read this as migration from `k8s.io` to `x-k8s.io` for kind's API types only, nothing about attempting to convert or support migration of anyone else's types.  Given the intended lifetime of kind clusters, it's not clear to me how much time we want to invest in automated migration vs. guides or tooling for one-time changes.

I agree we should do this while in alpha when we're allowed to make clean breaks like this, and think beta is the appropriate boundary to stop using `k8s.io`

site/content/docs/contributing/1.0-roadmap.md
```diff
@@ -15,25 +15,41 @@ as a project.
...
-- [ ] Improve documentation - In Progress ...
-  - [ ] expand examples of using kind
-  - [ ] cover known issues, debugging, work-arounds, etc.
-  - [X] create a documentation site - [#268]
-- [ ] Improving logging and debuggability
-  - [ ] enable audit-logging
-- [ ] Support non-amd64 architectures (namely ARM) - [#166]
+  - [x] improve instructions for KUBECONFIG in particular
+- [ ] Support usage as a properly versioned, supported, and documented library 
+  - [ ] Improving logging and debuggability
```
spiffxp: Requesting a change here: this is too vague.

What specific improvements need to be made to logging?  What features need to be implemented to make kind appropriately debuggable?  

Or maybe looked at from another angle, what problems are really hard to troubleshoot right now, that we would want to be able to troubleshoot before reaching beta?

site/content/docs/contributing/1.0-roadmap.md
spiffxp: I'm not clear what the ask is here, but I can't tell that it requires any changes to this line.

To me, I think use as a library means we're going to have some specific use cases documented, and those are going to be documented and tested.

We should make sure that we spell out what "properly versioned" means as part of these use cases.

site/content/docs/contributing/1.0-roadmap.md
BenTheElder: we need to enable good support for it, whether that means by default or a guide or a toggle remains to be determined

site/content/docs/contributing/1.0-roadmap.md
BenTheElder: er right, a clean break may have to be what we do.

I think we're all in agreement that we need to make this move, and it should happen for beta, @neolit123 was bringing up some technical aspects of how this might be done, and how this is a problem for many projects now.

I think we have a number of options to achieve this with greater ease in kind, up to and including just dropping support for the old API group if we have to. That's a discussion best had elsewhere though :sweat_smile:

site/content/docs/contributing/1.0-roadmap.md
BenTheElder: Fair:
- it's wildly inconsistent currently
- it's difficult to consume as a library
- errors and other output generally lack enough context to determine what failed.

There are a number of specific fixes I'm looking at here, I'd like to keep this _somewhat_ vague as to how we solve them, but it should be possible to debug kind without needing to modify it or use some external debugging tools, currently that's not always the case due to the poor logging / errors.

site/content/docs/contributing/1.0-roadmap.md
BenTheElder: I think we need a bit more work into determining what "properly versioned" is, but roughly it should be possible to follow semver without every release being a major version, and we should not violate semver.

That means things like the public API types being stable from a compatibility perspective, and the non-public API types not being part of the library. Changes to the library methods need to be non-breaking, which means moves like ensuring cluster creation takes an extensible set of options so we can add new ones with defaulting without breaking the existing options usage.

Supported is mostly redundantly emphasizing this, IE not breaking people and having a public interface.

documented of course is self explanatory, we have poor / nonexistent godocs.

site/content/docs/contributing/1.0-roadmap.md
spiffxp: What about something like this? Trying to think from perspective of: if someone wanted to help you get to beta, what could they help out with
```suggestion
- [ ] should be possible to troubleshoot kind without needing to modify kind or use external debugging tools
  - [ ] consistent logging (what is logged, when should it be logged, what levels are used)
  - [ ] errors have appropriate context
```

site/content/docs/contributing/1.0-roadmap.md
BenTheElder: implemented https://kind.sigs.k8s.io/docs/user/resources/ to link out to external resources, but we will want to organize more first-party resources as well

site/content/docs/contributing/1.0-roadmap.md
```diff
@@ -52,14 +68,20 @@ Longer Term goals include:
 persistent storage
 - Improving "kind as a library"
```
amwat: This contradicts the beta requirements now. remove?

site/content/docs/contributing/1.0-roadmap.md
BenTheElder: [working on this one]

site/content/docs/contributing/1.0-roadmap.md
BenTheElder: ack, done

BenTheElder:
addressed suggestions / nits, accepted one of @spiffxp's, squashed.

BenTheElder:
Prow is still failing to trigger tests. This isn't a code change and the netlify output looks correct https://deploy-preview-749--k8s-kind.netlify.com/docs/contributing/1.0-roadmap/

PR #751, opened 2019-08-04
Add .travis.yml

krichter722:
Easy test of build instructions which allows to detect that their not working (now and in the future after they've been fixed).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,13 @@
+language: sh
+
+services:
+- docker
+
+env:
+  matrix:
+   - BUILD_IMG=ubuntu:16.04
+   - BUILD_IMG=ubuntu:18.04
+   - BUILD_IMG=ubuntu:19.04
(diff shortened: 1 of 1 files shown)


krichter722:
I signed the CLA

BenTheElder:
Thank you, but we already have build CI, every PR builds kind and runs the conformance tests against all supported Kubernetes branches.

You need a supported version of go. https://golang.org/doc/devel/release.html#policy

I'll update the docs to be clear about this.

In the meantime please use the makefile as an alternative that doesn't require a supported go version.

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/753 will clarify the go requirements

BenTheElder:
we've clarified this now, hope that helps!
go 1.11+ has modules support, which we need for our dependencies etc. (eg kustomize which is a v3 module)

thanks again though!

Issue #747, opened 2019-07-31
clarify go version requirements


sb1975:
# go get sigs.k8s.io/kind
package sigs.k8s.io/kustomize/v3/k8sdeps/kunstruct: cannot find package "sigs.k8s.io/kustomize/v3/k8sdeps/kunstruct" in any of:
        /usr/lib/golang/src/sigs.k8s.io/kustomize/v3/k8sdeps/kunstruct (from $GOROOT)
        /root/go/src/sigs.k8s.io/kustomize/v3/k8sdeps/kunstruct (from $GOPATH)
package sigs.k8s.io/kustomize/v3/k8sdeps/transformer: cannot find package "sigs.k8s.io/kustomize/v3/k8sdeps/transformer" in any of:
        /usr/lib/golang/src/sigs.k8s.io/kustomize/v3/k8sdeps/transformer (from $GOROOT)
        /root/go/src/sigs.k8s.io/kustomize/v3/k8sdeps/transformer (from $GOPATH)
package sigs.k8s.io/kustomize/v3/k8sdeps/validator: cannot find package "sigs.k8s.io/kustomize/v3/k8sdeps/validator" in any of:
        /usr/lib/golang/src/sigs.k8s.io/kustomize/v3/k8sdeps/validator (from $GOROOT)
        /root/go/src/sigs.k8s.io/kustomize/v3/k8sdeps/validator (from $GOPATH)
package sigs.k8s.io/kustomize/v3/pkg/commands/build: cannot find package "sigs.k8s.io/kustomize/v3/pkg/commands/build" in any of:
        /usr/lib/golang/src/sigs.k8s.io/kustomize/v3/pkg/commands/build (from $GOROOT)
        /root/go/src/sigs.k8s.io/kustomize/v3/pkg/commands/build (from $GOPATH)
package sigs.k8s.io/kustomize/v3/pkg/fs: cannot find package "sigs.k8s.io/kustomize/v3/pkg/fs" in any of:
        /usr/lib/golang/src/sigs.k8s.io/kustomize/v3/pkg/fs (from $GOROOT)
        /root/go/src/sigs.k8s.io/kustomize/v3/pkg/fs (from $GOPATH)
package sigs.k8s.io/kustomize/v3/pkg/resmap: cannot find package "sigs.k8s.io/kustomize/v3/pkg/resmap" in any of:
        /usr/lib/golang/src/sigs.k8s.io/kustomize/v3/pkg/resmap (from $GOROOT)
[...]

PR #753, opened 2019-08-05
clarify go support, call out strongly to the docs site

BenTheElder:
fixes: #747, replaces #751

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,12 +4,12 @@
 <img alt="Build Status" src="https://prow.k8s.io/badge.svg?jobs=ci-kind-build">
 </a> <a href="https://godoc.org/sigs.k8s.io/kind"><img src="https://godoc.org/sigs.k8s.io/kind?status.svg"></a> <a href="https://goreportcard.com/report/sigs.k8s.io/kind"><img alt="Go Report Card" src="https://goreportcard.com/badge/sigs.k8s.io/kind" /></a></p>
 
-### [View The Documentation](https://kind.sigs.k8s.io)
+# Please see [Our Documentation](https://kind.sigs.k8s.io/docs/user/quick-start/) for more in-depth installation etc.
 
 kind is a tool for running local Kubernetes clusters using Docker container "nodes".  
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -4,12 +4,12 @@
 <img alt="Build Status" src="https://prow.k8s.io/badge.svg?jobs=ci-kind-build">
 </a> <a href="https://godoc.org/sigs.k8s.io/kind"><img src="https://godoc.org/sigs.k8s.io/kind?status.svg"></a> <a href="https://goreportcard.com/report/sigs.k8s.io/kind"><img alt="Go Report Card" src="https://goreportcard.com/badge/sigs.k8s.io/kind" /></a></p>
 
-### [View The Documentation](https://kind.sigs.k8s.io)
+# Please see [Our Documentation](https://kind.sigs.k8s.io/docs/user/quick-start/) for more in-depth installation etc.
```
BenTheElder: - users looking for docs from the should really start at this page, the root page of the site is a landing page nearly 1:1 with the readme
- this should be bigger, ideally many potential users will click this, however we should still have a readme because people don't like having to go to another site :-)

README.md
```diff
@@ -4,12 +4,12 @@
...
 
 kind is a tool for running local Kubernetes clusters using Docker container "nodes".  
 kind is primarily designed for testing Kubernetes 1.11+, initially targeting the [conformance tests].
 
-If you have [go] and [docker] installed `GO111MODULE="on" go get sigs.k8s.io/kind@v0.4.0 && kind create cluster` is all you need!
+If you have [go] ([1.11+][go-supported]) and [docker] installed `GO111MODULE="on" go get sigs.k8s.io/kind@v0.4.0 && kind create cluster` is all you need!
```
BenTheElder: this is _another_ place we have to keep updated, I'm tempted to remove this shoutout entirely in the future once we have the podman etc support landed in favor of just linking to the install guide in the quick start...

for now we should keep this clear though, and we're pointing out _why_ (we follow the upstream support policy for now of the 2 latest go minor versions, in the future since we have the containerized build it might be reasonable to just focus on a specific version and tell people to use that...)

PR #754, opened 2019-08-05
tidy up youtube embeds

BenTheElder:
- use the built in hugo shortcode instead of inlining the html
- add a tiny bit of CSS to ensure videos expand to the width of the context box

/cc @jeefy

### DIFF
--- a/site/assets/css/inline.css
+++ b/site/assets/css/inline.css
@@ -72,6 +72,7 @@ body {
   border-bottom: 1px solid rgba(0, 0, 0, 0.1);
   display: flex;
   justify-content: space-between;
+  z-index: 99999;
 }
 .sidebar-collapsed #navbar{
   width: 100%;
@@ -195,6 +196,21 @@ h1:hover a, h2:hover a, h3:hover a, h4:hover a {
 }
 #sidebar ul ul {
   padding-left: 1.25em;
+}
+
+/* style video embeds to be full-width in content */
+.video-wrapper {
+  position: relative;
+  width: 100%;
+  height: 0;
+  padding-bottom: 56.25%;
+}
+.video-wrapper iframe {
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
 }
 
 /* mobile */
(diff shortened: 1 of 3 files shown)


site/config.toml
```diff
@@ -53,3 +53,9 @@ baseName = "_redirects"
 
 [outputs]
 home = ["HTML", "REDIRECTS"]
+
+# privacy settings
+[privacy]
+  [privacy.youtube]
+    # enable the cookie-less youtube in built-in hugo shortcode
+    privacyEnhanced = true
```
BenTheElder: this ensures we get the same cookieless content as before, the new HTML produced by the shortcode is like:
```html
<div class="video-wrapper">
  <iframe src="//www.youtube-nocookie.com/embed/8KtmevMFfxA" allowfullscreen="" title="YouTube Video"></iframe>
</div>
```

site/content/docs/user/resources.md
```diff
@@ -18,21 +18,21 @@ The authors have given the following talks relating to kind:
 
 At KubeCon EU 2019 we spoke about [kind and testing your Kubernetes Applications][testing-k8s-apps-with-kind].
 
-<iframe width="560" height="315" src="https://www.youtube-nocookie.com/embed/8KtmevMFfxA" frameborder="0" allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
+{{< youtube id="8KtmevMFfxA" class="video-wrapper" >}}
```
BenTheElder: yields this HTML
```html
<div class="video-wrapper">
  <iframe src="//www.youtube-nocookie.com/embed/8KtmevMFfxA" allowfullscreen="" title="YouTube Video"></iframe>
</div>
```

site/assets/css/inline.css
```diff
@@ -72,6 +72,7 @@ body {
   border-bottom: 1px solid rgba(0, 0, 0, 0.1);
   display: flex;
   justify-content: space-between;
+  z-index: 99999;
```
BenTheElder: z-index added so the navbar stays above the youtube videos now that we have content ~ab~using `position: absolute` :-)

BenTheElder:
rendered: https://deploy-preview-754--k8s-kind.netlify.com/docs/user/resources/

Issue #594, opened 2019-06-05
add credentials programattically docs are out of date


BenTheElder:
https://kind.sigs.k8s.io/docs/user/private-registries/#add-credentials-programmatically is currently out of date, specifically it leverages `docker login` on the nodes which is not going to work.

roughly the same sample can be used, but the login needs to be on the host, or something else equivalent.

BenTheElder:
This would be easy to fix. The `docker login` generated credentials just need to be generated by another means or on the host. The upstream kubernetes docs have a guide for this I believe.

/good-first-issue

paranoidaditya:
@BenTheElder Should we point users at https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/ or maintain a copy of instructions?

BenTheElder:
sorry I missed this, yes, I think we should probably point to that and maybe just note ourselves how you could copy the credentials to the node and ensure kubelet picks them up instead of using them natively in kubernetes

WalkerGriggs:
Saw the 'help wanted' / 'good first issue' label and thought I'd pitch in. Changes in script tested with personal, private GCR. 

Aside from updating the snippet to generate credentials on the host machine, is there anything else worth mentioning?

BenTheElder:
> Saw the 'help wanted' / 'good first issue' label and thought I'd pitch in. Changes in script tested with personal, private GCR.

Thank you!

> Aside from updating the snippet to generate credentials on the host machine, is there anything else worth mentioning?

Just anything you needed to know / you think would be helpful here. I'm not currently using this but I know some users trip over it and I'd like to make sure we keep this up to date / helpful 😅

BenTheElder:
https://github.com/kubernetes-sigs/kind/compare/master...WalkerGriggs:private-registry-docs looks great, please do send a PR!

PR #756, opened 2019-08-06
Updates private registry programmatic key addition docs

WalkerGriggs:
Simple doc update which pulls `docker login` off the nodes and onto the host machine.

See #594

### DIFF
--- a/site/content/docs/user/private-registries.md
+++ b/site/content/docs/user/private-registries.md
@@ -20,6 +20,8 @@ images. If possible, this is the preferable and most portable route.
 See [the upstream kubernetes docs for this][imagePullSecrets],
 kind does not require any special handling to use this.
 
+If you already have the config file local but would still like to use secrets,
+read through kubernetes' docs for [creating a secret from a file][imagePullFileSecrets].
 
 ## Pull to the Host and Side-Load
 
@@ -57,12 +59,16 @@ A credential can be programmatically added to the nodes at runtime.
 
 If you do this then kubelet must be restarted on each node to pick up the new credentials.
 
-An example bash snippet for doing this with with [gcr.io][GCR]:
+An example bash snippet for generating a [gcr.io][GCR] cred file on your host machine:
 
 ```bash
 # login to GCR on all your kind nodes
+
 # KUBECONFIG should point to your kind cluster
 export KUBECONFIG="$(kind get kubeconfig-path --name="kind")"
+
+# move the host config out of the way if it exists
+[ -f $HOME/.docker/config.json ] && mv $HOME/.docker/config.json $HOME/.docker/config.json.host
 
 # https://cloud.google.com/container-registry/docs/advanced-authentication#access_token
 gcloud auth print-access-token | docker login -u oauth2accesstoken --password-stdin https://gcr.io
@@ -76,12 +82,16 @@ for node in $(kubectl get nodes -oname); do
     # restart kubelet to pick up the config
     docker exec ${node_name} systemctl restart kubelet.service
 done
+
+# delete the temporary config
+rm $HOME/.docker/config.json
+
+# move the original, host config back if it exists
+[-f $HOME/.docker/config.json.host] && mv $HOME/.docker/config.json.host $HOME/.docker/config.json
 ```
 
-Move preexisting configs or clean up the new config on the host machine as needed.
-
-
 [imagePullSecrets]: https://kubernetes.io/docs/concepts/containers/images/#specifying-imagepullsecrets-on-a-pod
+[imagePullFileSecrets]: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#registry-secret-existing-credentials
 [loading an image]: /docs/user/quick-start/#loading-an-image-into-your-cluster
 [using a private registry]: https://kubernetes.io/docs/concepts/containers/images/#using-a-private-registry
 [GCR]: https://cloud.google.com/container-registry/
--- a/site/content/docs/user/private-registries.md
+++ b/site/content/docs/user/private-registries.md
@@ -63,18 +63,23 @@ An example bash snippet for doing this with with [gcr.io][GCR]:
 # login to GCR on all your kind nodes
 # KUBECONFIG should point to your kind cluster
 export KUBECONFIG="$(kind get kubeconfig-path --name="kind")"
+
+# https://cloud.google.com/container-registry/docs/advanced-authentication#access_token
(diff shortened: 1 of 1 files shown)


WalkerGriggs:
Signed the CLA before submitting the PR; perhaps it hasn't propagated?

site/content/docs/user/private-registries.md
```diff
@@ -63,19 +63,24 @@ An example bash snippet for doing this with with [gcr.io][GCR]:
...
-    docker exec ${node_name} cp /root/.docker/config.json /var/lib/kubelet/config.json
+    docker cp $HOME/.docker/config.json ${node_name}:/var/lib/kubelet/config.json
     # restart kubelet to pick up the config
     docker exec ${node_name} systemctl restart kubelet.service
 done
 ```
 
+Move preexisting configs or clean up the new config on the host machine as needed.
+
+
```
neolit123: one extra new line can be removed.

site/content/docs/user/private-registries.md
```diff
@@ -63,19 +63,24 @@ An example bash snippet for doing this with with [gcr.io][GCR]:
 # login to GCR on all your kind nodes
 # KUBECONFIG should point to your kind cluster
 export KUBECONFIG="$(kind get kubeconfig-path --name="kind")"
+
+# https://cloud.google.com/container-registry/docs/advanced-authentication#access_token
+gcloud auth print-access-token | docker login -u oauth2accesstoken --password-stdin https://gcr.io
```
neolit123: this however can be disruptive to the host config.json, in case someone has manually set up their host already.
i think the old approach is better in that regard.

neolit123:
> WalkerGriggs needs CLA

/check-cla

some unknown/related errors happen from time to time, and the only good option is to contact the CNCF help desk.

site/content/docs/user/private-registries.md
WalkerGriggs: As @chemikadze brought up in https://github.com/kubernetes-sigs/kind/commit/6b796849caec75dba1699a02d38cb67a0596d637, the old method will no longer work with containerd.

See https://github.com/kubernetes-sigs/kind/issues/594#issuecomment-507824471. More specifically, @BenTheElder's comments about a potential solution

> The docker login generated credentials just need to be generated by another means or on the host.

We can script around the disruption of the host config? (ie. move the old config out of the way and move it back later)

WalkerGriggs:
> if you have your CLA email added to your github account, then contacting the help desk is needed.

Can do

site/content/docs/user/private-registries.md
neolit123: > the old method will no longer work with containerd.

looks like i've missed the existing comments.

> We can script around the disruption of the host config? (ie. move the old config out of the way and move it back later)

backup on the existing configuration on the host, seems good to me.

WalkerGriggs: Sounds like a plan. I'll get those changes made soon :tm:

site/content/docs/user/private-registries.md
WalkerGriggs: 👍

BenTheElder:
Sorry about the CLA issues :/
/ok-to-test

BenTheElder:
I wonder how hard it would be to find a middle ground between this approach and http://docs.heptio.com/content/private-registries/pr-gcr.html

IE kubectl / Kubernetes must eventually be using this special registry secret similar to the on-disk config format, I wonder if we can generate that ourselves without docker login 🤔

site/content/docs/user/private-registries.md
chemikadze: Access tokens seem to expire in a matter of hours, does it make sense to use serviceaccount-based auth in the example?

chemikadze:
@BenTheElder do you mean generation of config file as part of `kind create cluster`?

Implementation in kubectl is not exported, however looks rather trivial:  https://github.com/kubernetes/kubernetes/blob/bdde11a664632e10b426ae1cbe684f238b47ef53/staging/src/k8s.io/kubectl/pkg/generate/versioned/secret_for_docker_registry.go#L151

BenTheElder:
I mean in the docs here as a snippet of bash or python or something.

kind could always use kubectl internally if we were doing a real cluster secret, but I think we still want to show creating the global config file for the runtime / kubelet because it's the simplest to work with and we already have the option for using an imagePullSecret documented.

WalkerGriggs:
@BenTheElder I see where you're coming from about global configs for kubelet being straight forward, but I'm going to push back on that just a wee bit.

I think there's some great middle ground with either creating the secret from file -- ie 

```
kubectl create secret generic regcred \
    --from-file=.dockerconfigjson=<path/to/.docker/config.json> \
    --type=kubernetes.io/dockerconfigjson
```

or just outright creating the secret with your creds 
```
kubectl create secret docker-registry regcred \
    --docker-server=<your-registry-server> \
    --docker-username=<your-name> \
    --docker-password=<your-pword> \
    --docker-email=<your-email>
```

and then patching that into your service account.

```
kubectl patch serviceaccount default -p '{"imagePullSecrets": [{"name": "regcred"}]}'
```

That turns a looping bash script which requires backing up / restoring the local config into a 2 command copy & paste. The only added complexity I see is understanding service accounts.

At the risk of listing too many choices and confusing end users, why don't we add both options? I can see a use case for both.

BenTheElder:
> At the risk of listing too many choices and confusing end users, why don't we add both options? I can see a use case for both.

SGTM 👍

WalkerGriggs:
@BenTheElder I noticed you already link upstream for the kubernetes secrets docs, so I added a short blurb about creating secrets from file (also linking upstream) and left it at that.

@neolit123 I added a few lines to the snippet to 'backup'/'restore' the config file on the host machine so it isn't disrupted.

Apparently my CLA is still an issue, so I'll get in touch with the help desk.

WalkerGriggs:
I signed it

site/content/docs/user/private-registries.md
```diff
@@ -20,6 +20,8 @@ images. If possible, this is the preferable and most portable route.
 See [the upstream kubernetes docs for this][imagePullSecrets],
 kind does not require any special handling to use this.
 
+If you already have the config file local but would still like to use secrets,
```
BenTheElder: tiny nit: s/local/locally/?

site/content/docs/user/private-registries.md
BenTheElder: @chemikadze let's do that in a follow up, since this _works_ and the current documentation is broken. I agree though.

PR #757, opened 2019-08-06
docs: ensure the sidebar stays above embedded videos

BenTheElder:
on mobile the sidebar navigation slides under the embedded videos currently because we didn't apply the same fix we applied to the header.


aojea:
/lgtm
at least it renders ok in the chrome mobile browser emulator (didn´t know it exists until now :)

PR #767, opened 2019-08-13
Node image contains only one pulled image

barkbay:
The `node` image contains only one pulled image, this is because each closure uses the same instance of the `image` variable

Not sure how to add a test but I would be happy to do that if you tell me how 😄


BenTheElder:
thanks for this fix :-)

refactoring things to be more (unit) testable is in progress. in general the build code is also a little hairy :disappointed:

PR #770, opened 2019-08-14
[HOLD ON UNTIL K/K MERGE] Bump cni version to v0.8.2

aojea:
xref: https://github.com/kubernetes/kubernetes/pull/78819

PR #773, opened 2019-08-15
use anonymous volume for /var/lib/kubelet

### LINKED ISSUE #771: Pod stuck in Terminating if a configmap is mounted with subPath
**What happened**:

When I create a pod which includes a configmap volume and if the volume mount uses a `subPath` it's not possible delete the Pod, it stays in `Terminating` forever.
This is a weird one and it only seems to happen on Mac, colleagues on Linux were unable to reproduce, so it's possibly a docker issue.

I expected the Pod to be deleted after a delete call.

With the following yaml as `example.yaml`:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: test-terminating
spec:
  containers:
  - command: ["sleep", "10000000000"]
    image: busybox
    name: test
    volumeMounts:
    - mountPath: /xyz
      name: cmap
      subPath: abc
  volumes:
  - configMap:
      name: test-terminating
    name: cmap
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: test-terminating
data:
  abc: "xyz"
```

Run:

```
$ kind create cluster
$ export KUBECONFIG="$(kind get kubeconfig-path --name="kind")"
$ kubectl apply -f example.yaml
$ kubectl wait --for=condition=Ready pod test-terminating
$ kubectl delete pod test-terminating # This will hang forever!!!
```

After a while it's possible to see that the pod is still Terminating:

```
$ kubectl get pod
NAME               READY   STATUS        RESTARTS   AGE
test-terminating   0/1     Terminating   0          28m
```

The problem only seems to happen on MacOS, on Linux I was unable to reproduce, the pod is successfully deleted.

**Environment:**

```
$ kind version
v0.4.0
```
```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.2", GitCommit:"f6278300bebbb750328ac16ee6dd3aa7d3549568", GitTreeState:"clean", BuildDate:"2019-08-05T16:54:35Z", GoVersion:"go1.12.7", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.0", GitCommit:"e8462b5b5dc2584fdcd18e6bcfe9f1e4d970a529", GitTreeState:"clean", BuildDate:"2019-06-25T23:41:27Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}
```
```
$ docker info
Client:
 Debug Mode: false

Server:
 Containers: 1
  Running: 1
  Paused: 0
  Stopped: 0
 Images: 1
 Server Version: 19.03.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Plugins:
  Volume: local
 Swarm: inactive
 Runtimes: runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 894b81a4b802e4eb2a91d1ce216b8817763c29fb
 runc version: 425e105d5a03fabd737a126ad93d62a9eeede87f
 init version: fec3683
 Security Options:
  seccomp
   Profile: default
 Kernel Version: 4.9.184-linuxkit
 Operating System: Docker Desktop
 OSType: linux
 Architecture: x86_64
 CPUs: 2
 Total Memory: 1.952GiB
 Name: docker-desktop
 ID: F4T7:DRSI:2M7J:5NHO:M4NR:GQUU:OKLV:KQQI:7CUQ:6XUE:ONRB:4EFF
 Debug Mode: true
  File Descriptors: 34
  Goroutines: 49
  System Time: 2019-08-14T19:28:34.768985516Z
  EventsListeners: 2
 HTTP Proxy: gateway.docker.internal:3128
 HTTPS Proxy: gateway.docker.internal:3129
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
```
```
$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.14.4
BuildVersion:	18E226
```

BenTheElder:
fixes #771

### DIFF
--- a/pkg/cluster/nodes/create.go
+++ b/pkg/cluster/nodes/create.go
@@ -146,6 +146,10 @@ func createNode(name, image, clusterLabel, role string, mounts []cri.Mount, port
 		"--tmpfs", "/run", // systemd wants a writable /run
 		// some k8s things want /lib/modules
 		"-v", "/lib/modules:/lib/modules:ro",
+		// ensure pods etc. are not on container filesystem
+		// TODO: we could do this in the image instead
+		// However this would leave old images with this issue
+		"-v", "/var/lib/kubelet",
 		"--hostname", name, // make hostname match container name
 		"--name", name, // ... and set the container name
 		// label the node with the cluster ID


BenTheElder:
FYI @cjwagner this should obviate the need for the additional mount in the local prowjob tool.

neolit123:
> fixes #770

is 770 correct here?

BenTheElder:
corrected to 771

pkg/cluster/nodes/create.go
```diff
@@ -146,6 +146,10 @@ func createNode(name, image, clusterLabel, role string, mounts []cri.Mount, port
 		"--tmpfs", "/run", // systemd wants a writable /run
 		// some k8s things want /lib/modules
 		"-v", "/lib/modules:/lib/modules:ro",
+		// ensure pods etc. are not on container filesystem
+		// TODO: we could do this in the image instead
+		// However this would leave old images with this issue
+		"-v", "/var/lib/kubelet",
```
neolit123: hm, this would expose the kind nodes to the host /var/lib/kubelet that may already contain kubelet data, which may result in:
- the kubelet on the node re-using the existing kubelet certificates
- the kubelet on the node stoping kubelet data on the host

is there a link to the original problem?

BenTheElder: no, that's not what this does. this is an anonymous volume, not a bind mount.

BenTheElder: https://success.docker.com/article/different-types-of-volumes

neolit123: my mistake.

BenTheElder:
prow [is broken at the moment](https://kubernetes.slack.com/archives/C09QZ4DQB/p1565910173360700). I've tested this locally and FWIW this is the same thing we do for `/var/lib/containerd`, just via a different means.

pkg/cluster/nodes/create.go
BenTheElder: this flag format does a lot of in-band signaling :sweat_smile: 
soon moving this to higher level logic... going to need to re-work that a bit now :upside_down_face:

PR #776, opened 2019-08-16
compare host and node image ID when loading images

BenTheElder:
fixes https://github.com/kubernetes-sigs/kind/issues/680

### DIFF
--- a/cmd/kind/load/docker-image/docker-image.go
+++ b/cmd/kind/load/docker-image/docker-image.go
@@ -116,8 +116,8 @@ func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
 	// pick only the nodes that don't have the image
 	selectedNodes := []clusternodes.Node{}
 	for _, node := range candidateNodes {
-		_, err := node.ImageInspect(imageID)
-		if err != nil {
+		id, err := node.ImageID(imageName)
+		if err != nil || id != imageID {
 			selectedNodes = append(selectedNodes, node)
 			log.Debugf("Image: %q with ID %q not present on node %q", imageName, imageID, node.String())
 		}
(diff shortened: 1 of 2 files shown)


cmd/kind/load/docker-image/docker-image.go
```diff
@@ -116,8 +116,8 @@ func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
 	// pick only the nodes that don't have the image
 	selectedNodes := []clusternodes.Node{}
 	for _, node := range candidateNodes {
-		_, err := node.ImageInspect(imageID)
```
aojea: @BenTheElder just curious, why this wasn't working? 
`crictl inspectid ImageID` returns an error if the id doesn't exist inside the image

```
# crictl inspecti 241709fdde11b50175f296c6c88c103ceb845dea69f27b350bebfc04c60a3 > /dev/null && echo "present"
present
root@kindv4-worker:/# crictl inspecti 241709fdde11b50175f296c6c88c103ceb845dea69f27b350bebfc04c60a4 > /dev/null && echo "present"
FATA[0000] no such image "241709fdde11b50175f296c6c88c103ceb845dea69f27b350bebfc04c60a4" present
root@kindv4-worker:/# crictl inspecti 241709fdde11b50175f296c6c8 > /dev/null && echo "present"
present
```

BenTheElder: Consider the case that two image names map to the same image.

BenTheElder: You need to check the (name, ID) tuple to ensure that loading would do nothing. 

We could get more clever when we observe that the ID already exists and avoid the save with some hacks but it's probably not worth the complexity and potential brittleness at the moment.

aojea: ahh ok ,got it, I was working under the false assumption that the name -> ID relation was unique

Thanks for the explanation

PR #777, opened 2019-08-19
Allow additional images when building the node image

barkbay:
Hi,

I'm trying to use Kind to run some tests on the Elastic Stack, in order to do that I have to pull several large images in every node container.
I have tried to let the node download them through the network, or load them via the `load` command but none of these solutions is ideal because they all involve a lot of I/O (either on network or on the disk of the host)
I would like to have your thoughts on allowing the user to add some additional images in the node image.
In addition to speeding up the startup of the application containers it would be the only way to add some images because `/var/lib/container` is a `VOLUME` and layers can't be added once the node image has been built.

Thank you !

### DIFF
--- a/cmd/kind/build/nodeimage/nodeimage.go
+++ b/cmd/kind/build/nodeimage/nodeimage.go
@@ -24,11 +24,12 @@ import (
 )
 
 type flagpole struct {
-	Source    string
-	BuildType string
-	Image     string
-	BaseImage string
-	KubeRoot  string
+	Source           string
+	BuildType        string
+	Image            string
+	BaseImage        string
+	AdditionalImages string
+	KubeRoot         string
 }
 
 // NewCommand returns a new cobra.Command for building the node image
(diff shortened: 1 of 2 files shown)


BenTheElder:
> In addition to speeding up the startup of the application containers it would be the only way to add some images because /var/lib/container is a VOLUME and layers can't be added once the node image has been built.

Heh, I'm actually working on changing that today :-)

> I would like to have your thoughts on allowing the user to add some additional images in the node image.

I think we want this feature, I think it should be possible to do without building Kubernetes though.

barkbay:
Nice :)
Let me know if I can help you on this one or on any related issue.

barkbay:
Hi, let me know your thoughts on this feature. No need to hurry _(I understand that you are a little bit overwhelmed :) )_ but it is something I would love to have in our CI.

I can also open a dedicated issue and close this PR if it helps.

Thank you !

BenTheElder:
spring cleaning, will revisit build later this year and come back to this then

PR #778, opened 2019-08-19
Improved installation snippet in readme

afritzler:
Added `uname` usage to detect on which system kind should be installed. Works now for Darwin and Linux based machines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,12 +39,12 @@ Without installing go, kind can be built reproducibly with docker using `make bu
 Stable binaries are also available on the [releases] page. Stable releases are
 generally recommended for CI usage in particular.
 To install, download the binary for your platform from "Assets" and place this
-into your `$PATH`. E.G. for macOS:
+into your `$PATH`:
 
 ```console
-curl -Lo ./kind-darwin-amd64 https://github.com/kubernetes-sigs/kind/releases/download/v0.4.0/kind-darwin-amd64
-chmod +x ./kind-darwin-amd64
-mv ./kind-darwin-amd64 /some-dir-in-your-PATH/kind
+curl -Lo ./kind https://github.com/kubernetes-sigs/kind/releases/download/v0.4.0/kind-$(uname)-amd64
+chmod +x ./kind
+mv ./kind /some-dir-in-your-PATH/kind
 ```
 
 To use kind, you will need to [install docker].  
--- a/images/base/entrypoint
+++ b/images/base/entrypoint
@@ -37,12 +37,9 @@ fix_mount() {
   # and this flag also happens to make /sys rw, amongst other things
   mount -o remount,ro /sys
 
-  echo 'INFO: making mounts shared for "/", "/run", "/var/lib/containerd"'
+  echo 'INFO: making mounts shared'
   # for mount propagation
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -39,12 +39,12 @@ Without installing go, kind can be built reproducibly with docker using `make bu
...
 generally recommended for CI usage in particular.
 To install, download the binary for your platform from "Assets" and place this
-into your `$PATH`. E.G. for macOS:
+into your `$PATH`:
 
 ```console
-curl -Lo ./kind-darwin-amd64 https://github.com/kubernetes-sigs/kind/releases/download/v0.4.0/kind-darwin-amd64
-chmod +x ./kind-darwin-amd64
-mv ./kind-darwin-amd64 /some-dir-in-your-PATH/kind
+curl -Lo ./kind-$(uname)-amd64 https://github.com/kubernetes-sigs/kind/releases/download/v0.4.0/kind-$(uname)-amd64
```
BenTheElder: `uname` prints `Linux` on my machine, we could store a variable with lowercased uname output though, either with `tr` or `awk`

BenTheElder: unless you are saying that github will download this with the wrong case?

we should also probably save this as just `./kind` and let the commands below not need the platform at all.

afritzler: The uppercase download will work on github. Of course the cleaner solution would be using `tr`. I just wanted to keep it simple though.

And yes just using `./kind` is a good idea. Let me amend those changes.

afritzler:
Also removed some trailing spaces :)

README.md
```diff
@@ -6,7 +6,7 @@
 
 # Please see [Our Documentation](https://kind.sigs.k8s.io/docs/user/quick-start/) for more in-depth installation etc.
 
-kind is a tool for running local Kubernetes clusters using Docker container "nodes".
```
BenTheElder: this is intentional, please maintain these. this moves the follow text to a newline in the rendered output.

README.md
```diff
@@ -31,24 +31,24 @@ You can install kind with `GO111MODULE="on" go get sigs.k8s.io/kind@v0.4.0`.
...
 ```console
-curl -Lo ./kind-darwin-amd64 https://github.com/kubernetes-sigs/kind/releases/download/v0.4.0/kind-darwin-amd64
-chmod +x ./kind-darwin-amd64
-mv ./kind-darwin-amd64 /some-dir-in-your-PATH/kind
+curl -Lo ./kind https://github.com/kubernetes-sigs/kind/releases/download/v0.4.0/kind-$(uname)-amd64
+chmod +x ./kind
+mv ./kind /some-dir-in-your-PATH/kind
 ```
 
-To use kind, you will need to [install docker].
```
BenTheElder: same for each of these

BenTheElder:
> Also removed some trailing spaces :)

those were actually intentional, two spaces is a line break :sweat_smile:  
https://github.github.com/gfm/#hard-line-breaks

afritzler:
Oh gosh. Maybe the vscode markdown auto formatting was not such a good idea :yum: 

Ok, I rolled that back now.

BenTheElder:
Thanks!

PR #779, opened 2019-08-19
make /var a runtime volume

BenTheElder:
Per the [Filesystem Hierarchy Standard](https://refspecs.linuxfoundation.org/fhs.shtml) all of `/var` is persistent runtime files. This should not be on the container filesystem.

Moving this to a runtime anonymous volume for _all_ of `/var`:
- allows us to not put any `/var/...` volumes in the base image (for container runtime storage, which must not be stacked)
  - which will allow us to potentially add pre-loaded images to an existing node image
- ensures writes to anything in `/var` goes to a volume instead of the container filesystem

The only downside is that new kind images will require a new version of kind. Images built prior to this should continue to work even with the new version.

While doing this I also normalized on long flags for node creation for clarity and consistency.

### DIFF
--- a/images/base/entrypoint
+++ b/images/base/entrypoint
@@ -37,12 +37,9 @@ fix_mount() {
   # and this flag also happens to make /sys rw, amongst other things
   mount -o remount,ro /sys
 
-  echo 'INFO: making mounts shared for "/", "/run", "/var/lib/containerd"'
+  echo 'INFO: making mounts shared'
   # for mount propagation
-  # TODO(bentheelder): determine which exact mounts we need to do this on
-  mount --make-shared /
-  mount --make-shared /run
-  mount --make-shared /var/lib/containerd
+  mount --make-rshared /
 }
 
 fix_machine_id() {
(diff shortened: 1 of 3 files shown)


BenTheElder:
/hold
doing some debugging locally to make sure everything does what I want :-)

images/base/entrypoint
```diff
@@ -37,12 +37,9 @@ fix_mount() {
...
   mount -o remount,ro /sys
 
-  echo 'INFO: making mounts shared for "/", "/run", "/var/lib/containerd"'
+  echo 'INFO: making mounts shared'
   # for mount propagation
-  # TODO(bentheelder): determine which exact mounts we need to do this on
-  mount --make-shared /
-  mount --make-shared /run
-  mount --make-shared /var/lib/containerd
+  mount --make-rshared /
```
BenTheElder: this is recursive, and from upstream (difficult to find docs for this though...)
we don't know what mounts users will try to propagate, so everything is made shared

we need this change not just because it's more correct, but because /var/lib/containerd is no longer a mount point

BenTheElder:
/retest
(prow failing to create the pod)

PR #780, opened 2019-08-20
update quick-start install docs following readme update

BenTheElder:
xref: #778


BenTheElder:
not a code change, looking to get the release tagged, let's not run tests again for this ...

PR #784, opened 2019-08-20
update install docs to point to latest release

BenTheElder:
... now that the release / binaries are up.


BenTheElder:
trivial docs correctness ...

PR #785, opened 2019-08-21
correct api redirects

BenTheElder:
whoops :man_shrugging:


BenTheElder:
we're also going to lose these when we go to x-k8s.io though :man_shrugging: 

we'll probably wind up adding better docs at some point, the godocs we have now are minimally useful, but that will be more important as this stabilizes. we have config usage docs in the user guide.

BenTheElder:
TBR

PR #789, opened 2019-08-21
Add fish shell detection

Qinusty:
Implement basic shell detection for Linux platforms in the `kind create` command. 

### Additions:
- Implement `GetLinuxShell()` utility function
- Add fish as a custom output whilst leaving bash/zsh under default

### DIFF
--- a/pkg/internal/cluster/create/create.go
+++ b/pkg/internal/cluster/create/create.go
@@ -39,6 +39,7 @@ import (
 	"sigs.k8s.io/kind/pkg/internal/cluster/create/actions/kubeadmjoin"
 	"sigs.k8s.io/kind/pkg/internal/cluster/create/actions/loadbalancer"
 	"sigs.k8s.io/kind/pkg/internal/cluster/create/actions/waitforready"
+	"sigs.k8s.io/kind/pkg/internal/util/env"
 )
 
 const (
@@ -170,8 +171,8 @@ func collectOptions(options ...create.ClusterOption) (*createtypes.ClusterOption
 }
 
 func printUsage(name string) {
-	// TODO: consider shell detection.
 	if runtime.GOOS == "windows" {
+		// TODO: consider shell detection.
 		fmt.Printf(
 			"Cluster creation complete. To setup KUBECONFIG:\n\n"+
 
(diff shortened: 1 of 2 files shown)


pkg/internal/cluster/create/create.go
```diff
@@ -190,12 +191,20 @@ func printUsage(name string) {
 			name,
 		)
 	} else {
+		// Detect shell and usage build prompt appropriately.
+		setKubeConfigPrompt := ""
+		switch shell := env.GetLinuxShell(); shell {
```
neolit123: shouldn't the function be called `UnixShell`, as this branch is `!windows`?
can it cover custom OSX shells too?

neolit123:
thanks for the PR @Qinusty 

please make sure you sign the CLA if this command fails:
/check-cla

neolit123:
if you have signed the CLA already, make sure your CLA email is added to your github account.
if there is a problem with that try contacting the CNCF help-desk.

WRT this PR, i must admit this is the first time i've heard about "fish" shell, and given all the flavors of shells out there, it might be better to not add special handling for all of them?
/assign @BenTheElder

pkg/internal/cluster/create/create.go
```diff
@@ -190,12 +191,20 @@ func printUsage(name string) {
...
+		case "fish":
+			setKubeConfigPrompt = "set -x KUBECONFIG (kind get kubeconfig-path --name=%q)"
```
BenTheElder: I'm a bit dumbfounded that a Linux shell doesn't support POSIX... 🧐
https://pubs.opengroup.org/onlinepubs/009695399/utilities/export.html

pkg/internal/cluster/create/create.go
BenTheElder: also noting that fish appears to be on macOS

pkg/internal/util/env/shell.go
```diff
@@ -0,0 +1,13 @@
+package env
+
+import (
+	"os"
+	"strings"
+)
+
+func GetLinuxShell() string {
+	shell := os.Getenv("SHELL")
```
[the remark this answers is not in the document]
Qinusty: This is currently a best guess, I'm not certain of a guaranteed way to get the users current shell. Scraping current running processes would run into a similar issue if the user was running multiple shells.

pkg/internal/cluster/create/create.go
Qinusty: Yeah it's a bit strange but it appears to be the case. https://fishshell.com/docs/current/tutorial.html#tut_exports

BenTheElder:
#850 landed so we don't need to export this env anymore.

Thanks for the PR though!

PR #790, opened 2019-08-21
Add an option to update kubeconfig

barkbay:
This feature is motivated by the fact that most of our current deployments _(minikube, gke ...)_ are managed through the default kubeconfig file.
It might also be convenient for people that use tools like [kubecxt](https://github.com/ahmetb/kubectx).
With this PR user can add `--set-context` when a cluster is created to import the new context to its kubeconfig.

Note that it does not remove the context when the cluster is deleted, but it can be done in an other PR _(I wanted to keep this one small)_.
Also I'm not sure about the unit test policy, I can add some of them to at least cover the merge with the current kubeconfig _(could also be done in an other PR)_

Let me know your thoughts.

This PR supersedes:
* https://github.com/kubernetes-sigs/kind/pull/512
* https://github.com/kubernetes-sigs/kind/pull/520
* https://github.com/kubernetes-sigs/kind/pull/517

### DIFF
--- a/cmd/kind/create/cluster/createcluster.go
+++ b/cmd/kind/create/cluster/createcluster.go
@@ -31,11 +31,12 @@ import (
 )
 
 type flagpole struct {
-	Name      string
-	Config    string
-	ImageName string
-	Retain    bool
-	Wait      time.Duration
+	Name       string
+	Config     string
+	ImageName  string
+	Retain     bool
+	Wait       time.Duration
+	SetContext bool
 }
 
 // NewCommand returns a new cobra.Command for cluster creation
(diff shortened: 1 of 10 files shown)


BenTheElder:
We should discuss this in an issue I think, there's some design and tradeoffs involved.

The users field is going to collide if multiple kind clusters are created IIRC.
This re-introduces a direct dependency on client-go which may make kind harder to depend on in test harnesses.
The flag name is not quite clear in versus actually happens.
Maybe this should be the new default?

This is definitely a neat feature, but it can _also_ be accomplished with a tiny bit of `kubectl` in the meantime, other than the merging issue which we need to settle on a solution for.

barkbay:
> We should discuss this in an issue I think, there's some design and tradeoffs involved.

I have created https://github.com/kubernetes-sigs/kind/issues/802

> The users field is going to collide if multiple kind clusters are created IIRC.

Yes, this PR takes care of that by adding the cluster name as a suffix.

> This re-introduces a direct dependency on client-go which may make kind harder to depend on in test harnesses.

👍 makes sense

> The flag name is not quite clear in versus actually happens.
> Maybe this should be the new default?

Agreed

Issue #765, opened 2019-08-12
use service account instead of access token for private registry example


BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/756#discussion_r310884414

/cc @chemikadze

WalkerGriggs:
@BenTheElder I might as well take this one on, too, since I started us down this path. 

I added Service Account documentation but left the access token snippet in as well -- gives people the option. Let me know what you think

BenTheElder:
that looks great, thanks!

PR #796, opened 2019-08-21
Adds services account auth example to private registry docs

WalkerGriggs:
Doc update to add example of using a GCP service account to authenticate with a GCR.

See https://github.com/kubernetes-sigs/kind/issues/765 and https://github.com/kubernetes-sigs/kind/pull/756#discussion_r310884414

### DIFF
--- a/site/content/docs/user/private-registries.md
+++ b/site/content/docs/user/private-registries.md
@@ -31,12 +31,12 @@ image(s) and then load them to the nodes you can avoid needing to authenticate
 on the nodes.
 
 
-## Add Credentials to the Nodes
+# Add Credentials to the Nodes
 
 Generally the upstream docs for [using a private registry] apply, with kind
 there are two options for this.
 
-### Mount a Config File to Each Node
+## Mount a Config File to Each Node
 
 If you pre-create a docker config.json containing credential(s) on the host
 you can mount it to each kind node.
@@ -53,7 +53,7 @@ nodes:
     hostPath: /path/to/my/secret.json
 ```
 
-### Add Credentials Programmatically
+### Use an Access Token
 
 A credential can be programmatically added to the nodes at runtime.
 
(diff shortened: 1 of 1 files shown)


site/content/docs/user/private-registries.md
```diff
@@ -90,6 +91,24 @@ rm $HOME/.docker/config.json
 [-f $HOME/.docker/config.json.host] && mv $HOME/.docker/config.json.host $HOME/.docker/config.json
 ```
 
+### ... with Service Accounts
```
neolit123: possibly expand the `...` here?

WalkerGriggs: Perhaps `Add Credentials Programmatically with Access Token` and `Add Credentials Programmatically with Service Account`? A little redundant, but we sadly can't nest further than an h3.

neolit123: ok, i see the nesting now.

IMO it should be:

> ## Add Credentials to the Nodes
> ### Mount a Config File to Each Node
> ### Use an  Access Token
> ### Use a Service Account

site/content/docs/user/private-registries.md
```diff
@@ -90,6 +91,24 @@ rm $HOME/.docker/config.json
...
+
+Access tokens are short lived though, so you may prefer to use a Service Account and keyfile.
+First, either download the key from the console or generate one with gcloud:
+
+```
+gcloud iam service-accounts keys create <output.json> --iam-account <account email>
+```
+
+Then, replace the `gcloud auth...` line from the previous snippet with:
```
neolit123: `from the previous section`
?

site/content/docs/user/private-registries.md
```diff
@@ -90,6 +91,24 @@ rm $HOME/.docker/config.json
 [-f $HOME/.docker/config.json.host] && mv $HOME/.docker/config.json.host $HOME/.docker/config.json
 ```
 
+### ... with Service Accounts
+
+Access tokens are short lived though, so you may prefer to use a Service Account and keyfile.
```
neolit123: > Access tokens are short lived, so you may prefer to use a Service Account and keyfile instead.

BenTheElder:
/ok-to-test
thanks for filing this

site/content/docs/user/private-registries.md
BenTheElder: > ...  but we sadly can't nest further than an h3.

Not that we should but, why can't we?

WalkerGriggs: @BenTheElder I might actually be wrong. Some markdown interpreters don't like going being h3; I think @neolit123 has the right nesting idea.

Will make changes soon :tm:

site/content/docs/user/private-registries.md
WalkerGriggs: Yea, I probably should be more direct and anchor the section

> Then, replace `gcloud auth print-access-token | ...` from the [previous section] (#use-an-access-token) with:

Or maybe you have another way to format the reference without copying the entire snippet?

site/content/docs/user/private-registries.md
BenTheElder: FWIW we use [hugo](https://gohugo.io/) and you can view the rendered version at https://deploy-preview-796--k8s-kind.netlify.com/ but @neolit123's suggestion SGTM

BenTheElder:
/lgtm
/approve
thanks a ton for writing this up :-)

Issue #792, opened 2019-08-21
release process failures


BenTheElder:
The following things are wrong with the release process:

- the cross build script used for releasing does not fail if one of the builds fails :warning: 
- the cross build script is not in CI :warning: 
- the kustomize version we are depending on does not build for windows :warning: 
- the release scripts do not perform a clean before building :warning: 

I've uploaded cleanly built binaries for the other platforms, but Windows is still broken, and these need fixing.

TODO:
- [x] fix the cross build script to fail if any build fails
- [x] add the cross build script to presubmit
- [x] upgrade kustomize to a patched (unfortunately unreleased) version that builds for windows
- [x] double check the v0.5.0 binaries and remove windows
- [ ] cut a new release

BenTheElder:
see also: https://github.com/kubernetes-sigs/kind/issues/791, https://github.com/kubernetes-sigs/kustomize/issues/1212, https://github.com/kubernetes-sigs/kustomize/pull/1374

sudermanjr:
Are these release issues also related to this error message I got when using one of the images from the v0.5.0 release?

```
Creating cluster "test-infra" ...
 ✓ Ensuring node image (kindest/node:v1.14.6) 🖼
ERRO[11:55:40] Unable to find image 'kindest/node:v1.14.6@sha256:4d53072ec035d82a70e9d88972d0eb63846706a892438238b6b54a4531c33a9d' locally
ERRO[11:55:40] docker: Error response from daemon: manifest for kindest/node@sha256:4d53072ec035d82a70e9d88972d0eb63846706a892438238b6b54a4531c33a9d not found.
```

If not, I will open another issue.

BenTheElder:
No that is not related.

BenTheElder:
regarding the failure to exit non-zero on build failure: https://ss64.com/bash/wait.html

> If n is not given, all currently active child processes are waited for, and the return status is zero.

:man_facepalming:

monopole:
releasing kustomize for windows at https://github.com/kubernetes-sigs/kustomize/releases is blocked by https://github.com/kubernetes-sigs/kustomize/issues/834.  We need a build and unit test run in the pre-commit.

BenTheElder:
@monopole what about just a cross build? We don't need windows machines or unit tests to ensure build works.

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/794 should:
- make the cross build included in presubmit
[...]

PR #797, opened 2019-08-21
cleanup cross.sh

BenTheElder:
remove redundant logic

### DIFF
--- a/hack/release/build/cross.sh
+++ b/hack/release/build/cross.sh
@@ -23,22 +23,6 @@ set -o pipefail
 REPO_ROOT=$(git rev-parse --show-toplevel)
 cd "${REPO_ROOT}"
 
-OUT="${REPO_ROOT}/bin"
-mkdir -p "${OUT}"
-
-CLEAN="false"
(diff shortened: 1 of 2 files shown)


BenTheElder:
next we should fix the waiting for jobs https://github.com/kubernetes-sigs/kind/issues/792#issuecomment-523647231

hack/release/build/cross.sh
```diff
@@ -23,22 +23,6 @@ set -o pipefail
 REPO_ROOT=$(git rev-parse --show-toplevel)
 cd "${REPO_ROOT}"
 
-OUT="${REPO_ROOT}/bin"
-mkdir -p "${OUT}"
```
BenTheElder: `make build` handles this

hack/release/build/cross.sh
```diff
@@ -23,22 +23,6 @@ set -o pipefail
...
-
-CLEAN="false"
```
BenTheElder: make clean handles all of this, and more

PR #798, opened 2019-08-21
ensure we get a combined status from the cross build

BenTheElder:
GNU `parallel` is another option but may not be installed.
`xargs` is widely installed and returns a combined status.

xref: https://github.com/kubernetes-sigs/kind/issues/792#issuecomment-523647231

### DIFF
--- a/hack/release/build/cross.sh
+++ b/hack/release/build/cross.sh
@@ -22,7 +22,7 @@ REPO_ROOT=$(git rev-parse --show-toplevel)
 cd "${REPO_ROOT}"
 
 # controls the number of concurrent builds
-PRALLELISM=${PRALLELISM:-6}
+PARALLELISM=${PARALLELISM:-6}
 
 echo "Building in parallel for:"
 # What we do here:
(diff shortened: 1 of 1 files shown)


BenTheElder:
/hold
verifying macOS

hack/release/build/cross.sh
```diff
@@ -15,31 +15,35 @@
...
-build "windows" "amd64" & \
-# we want each pid to be an argument
-# shellcheck disable=SC2046
-wait $(jobs -p)
+# What we do here:
+# - use xargs to build in parallel (-P) while collecting a combined exit code
+# - use cat to supply the individual args to xargs (one line each)
+# - use env -S to split the line into environment variables and execute
+# - ... the build
+# NOTE: the binary name needs to in single quotes so we delay evaluating
```
neolit123: `to be in single`

hack/release/build/cross.sh
```diff
@@ -15,31 +15,35 @@
...
+# GOOS / GOARCH
+if xargs -0 -n1 -P "${PRALLELISM}" -I{} \
+    bash -c 'eval $0; make build KIND_BINARY_NAME=kind-${GOOS}-${GOARCH}' {}; then
+    echo "Cross build passed!" 1>&2
+else
+    echo "Cross build failed!" 1>&2
+    exit 1
+fi < <(cat <<EOF | tr '\n' '\0'
```
BenTheElder: macOS doesn't have `xargs -d` but mac and linux both have `-0`, and otherwise decide on their own what the delimeter is (which may otherwise be the spaces)

BenTheElder:
/hold cancel
should pass lints now, works on mac and linux

hack/release/build/cross.sh
BenTheElder: done

BenTheElder:
cc @amwat @aojea 
you two will like this bash :wink:

hack/release/build/cross.sh
```diff
@@ -15,31 +15,36 @@
...
-    GOOS="${1}"
-    GOARCH="${2}"
-    export GOOS
-    export GOARCH
-    KIND_BINARY_NAME="kind-${GOOS}-${GOARCH}"
-    export KIND_BINARY_NAME
-    make build
-}
+# controls the number of concurrent builds
+PRALLELISM=${PRALLELISM:-6}
```
amwat: nit: s/PRALLELISM/PARALLELISM

aojea:
Hehe nice, bash parallel programming :)

PR #799, opened 2019-08-22
[TEST][WIP] Update CoreDNS config to 1.5

aojea:
Add the ready plugin, otherwise, the readiness probe doesn't allow the pod to run

### DIFF
--- a/hack/ci/e2e.sh
+++ b/hack/ci/e2e.sh
@@ -145,6 +145,7 @@ metadata:
   namespace: kube-system
 ---
 EOF
+    kubectl -n kube-system delete pods -l k8s-app=kube-dns
   fi
 
   # ginkgo regexes
--- a/hack/ci/e2e.sh
+++ b/hack/ci/e2e.sh
@@ -128,6 +128,7 @@ apiVersion: v1
 data:
   Corefile: |
     .:53 {
+        ready
         errors
         health
         kubernetes cluster.local internal in-addr.arpa ip6.arpa {


aojea:
I'm starting to suspect that we are facing different issues with IPv6
```
Aug 22 11:58:22.746: INFO: kindnet-965h5 started at 2019-08-22 11:44:58 +0000 UTC (0+1 container statuses recorded)
Aug 22 11:58:22.746: INFO: 	Container kindnet-cni ready: true, restart count 0
Aug 22 11:58:22.746: INFO: coredns-d4b9d4d8b-n6884 started at 2019-08-22 11:45:14 +0000 UTC (0+1 container statuses recorded)
Aug 22 11:58:22.746: INFO: 	Container coredns ready: false, restart count 0
```

in this case CoreDNS 1.5.0 is not `ready`, however I added the config 🤷‍♂

aojea:
ok, one issue solved, `CoreDNS-1.5.0` doesn't work after reloading the configuration, you have to recreate the pods and it works. The tests that are failing now is because of the test not picking the `framework.TestContext.IPFamily == ipv6` variable

@neolit123 the changelog of CoreDNS-1.5.2 mentions issues with `ready` and `reload`, is any specific reason to point to 1.5.0 instead of 1.5.2?

BenTheElder:
We added the restart in https://github.com/kubernetes-sigs/kind/pull/803, which also adds the `ready` plugin by way of not patching it out 😅

hack/ci/e2e.sh
```diff
@@ -144,6 +145,7 @@ metadata:
   namespace: kube-system
 ---
 EOF
+    kubectl -n kube-system delete pods -l k8s-app=kube-dns
```
BenTheElder: I stole this line for https://github.com/kubernetes-sigs/kind/pull/803 :^)

BenTheElder:
thanks for this, assuming we can close now 😅

aojea:
fixed by #803

PR #804, opened 2019-08-22
upgrade hugo

BenTheElder:
to 0.57.2


site/Makefile
```diff
@@ -1,3 +1,5 @@
+# TODO: use a docker image to serve a development copy using
```
BenTheElder: I'm going to file a help wanted issue for this rather than do it now, this should be a good new contributor project :-)

BenTheElder:
seems kubernetes/website is also still on 0.53 ...

PR #805, opened 2019-08-22
improve repo tooling

BenTheElder:
- perform `set` on one line in all of the scripts, it's shorter and just as clear
- use `BASH_SOURCE[0]` + `pwd -P` trick instead of `git ...` to make scripts work from tarballs etc. instead of just from git clones
- move boilerplate file out of top level hack/
- upgrade to go 1.13rc1 to prepare for the pending go 1.13 release
- simplify `go_container.sh` and move it up to `hack/` since we want people to use this
- replace all linters with `golangci-lint`
- fix lint errors


BenTheElder:
/retest
prow flake

hack/verify/lint.sh
```diff
@@ -13,26 +13,19 @@
...
-go build -o "${BINDIR}/staticcheck" honnef.co/go/tools/cmd/staticcheck
-# go back to the root
-cd "${REPO_ROOT}"
-
-"${BINDIR}/staticcheck" ./pkg/... ./cmd/... .
+# run golangci-lint
+GO111MODULE=on bin/golangci-lint \
+  --enable=golint --enable=vet --enable=gofmt \
+  --enable=misspell \
+  run ./pkg/... ./cmd/... .
```
BenTheElder: this includes staticcheck and other lints by default, we're explicitly enabling a few that are not on by default:
- standard go lints (golint, vet, gofmt)
- misspell

hack/tools/go.mod
```diff
@@ -3,12 +3,31 @@ module sigs.k8s.io/kind/hack/tools
...
-	golang.org/x/lint v0.0.0-20190409202823-959b441ac422
+	golang.org/x/crypto v0.0.0-20190510104115-cbcb75029529 // indirect
 	golang.org/x/tools v0.0.0-20190606174628-0139d5756a7d // indirect
-	honnef.co/go/tools v0.0.0-20190605142022-0a11fc526260
 	k8s.io/code-generator v0.0.0-20190311093542-50b561225d70
 	k8s.io/gengo v0.0.0-20190327210449-e17681d19d3a // indirect
 	k8s.io/klog v0.3.0 // indirect
 )
+
+// deal with golangci-lint being broken
```
BenTheElder: https://github.com/golangci/golangci-lint/issues/595#issuecomment-524166777

hack/verify/lint.sh
BenTheElder: iterative (hot) run:
```
[bentheelder@bentheelder-macbookpro:~/go/src/sigs.k8s.io/kind·2019-08-22T21:38:14-0700·shell-nits@245ca10]
$ time hack/verify/lint.sh 

real    0m7.842s
user    0m23.441s
sys     0m4.258s
```

much faster than running these individually.

I'm exploring the additional lints we can enable, but may leave that for a follow up.

hack/go_container.sh
```diff
@@ -0,0 +1,93 @@
...
+      *) echo "Unknown host architecture! '$(uname -m)'" exit 2 ;;
+    esac
+  fi
+
+  export GOOS GOARCH
+}
+
+# runs $@ in a go container with caching etc.
+run_in_go_container() {
+  # run in the container
```
BenTheElder: this comment is a bit redundant 🙃 ...

hack/tools/go.mod
BenTheElder: basically: go 1.13 is stricter about the timestamps matching the source when using a pseudo-version, the module proxy is equally strict now even in 1.12 / 1.11. golangci-lint needs to update their go.mod upstream, there are PRs open for this.

hack/verify/all.sh
```diff
@@ -24,39 +24,9 @@ res=0
 
 # run all verify scripts, optionally skipping any of them
 
-if [[ "${VERIFY_SHELLCHECK:-true}" == "true" ]]; then
-  echo "verifying shellcheck ..."
-  hack/verify/shellcheck.sh || res=1
-  cd "${REPO_ROOT}"
-fi
-
-if [[ "${VERIFY_SPELLING:-true}" == "true" ]]; then
```
BenTheElder: these are all in lint.sh now, by way of golangci-lint

pkg/internal/apis/config/encoding/scheme.go
```diff
@@ -52,14 +52,13 @@ func AddToScheme(scheme *runtime.Scheme) {
 }
 
 // V1Alpha3ToInternal converts to the internal API version
-func V1Alpha3ToInternal(cluster *v1alpha3.Cluster) *config.Cluster {
+func V1Alpha3ToInternal(cluster *v1alpha3.Cluster) (*config.Cluster, error) {
 	// apply defaults
 	Scheme.Default(cluster)
 	// the convert
 	out := &config.Cluster{}
 	// TODO: error handling??
```
BenTheElder: stale TODO

hack/go_container.sh
```diff
@@ -0,0 +1,93 @@
...
+# ============================ SCRIPT SETTINGS =================================
+# get the repo root for defaulting OUT_DIR and SOURCE_DIR
+REPO_ROOT="${REPO_ROOT:-$(cd "$(dirname "$0")/.." && pwd)}"
+# output directory, will be mounted to /out, defaults to /bin in REPO_ROOT
+OUT_DIR="${OUT_DIR:-${REPO_ROOT}/bin}"
+# source directory, will be mounted to /src, defaults to current directory
+SOURCE_DIR="${SOURCE_DIR:-$(pwd -P)}"
+# default to disabling CGO for easier reproducible builds and cross compilation
+export CGO_ENABLED="${CGO_ENABLED:-0}"
+# the container image, by default a recent official golang image
```
aojea: when do you say recent do you want to do this smarter?
https://gist.github.com/n8henrie/1043443463a4a511acf98aaa4f8f0f69#file-get_golang-sh-L11

BenTheElder: we could use `golang:latest` but that might get a bit surprising ... probably best to periodically manually update it.

hack/release/create.sh
```diff
@@ -17,18 +17,18 @@
 # builds binaries between the commits
 # Use like: create.sh <release-version> <next-prerelease-version>
 # EG: create.sh v0.3.0 v0.4.0-alpha
-set -o nounset
-set -o errexit
-set -o pipefail
+set -o errexit -o nounset -o pipefail
 
+# cd to the repo root
```
aojea: why do you move this before checking?
if there are no arguments and the next `if` fails we don't need to cd to the repo root, right?

BenTheElder: for consistency with the other scripts. it doesn't matter but it's easier to spot check these between the scripts (EG if we tweak the `REPO_ROOT` logic)

hack/go_container.sh
BenTheElder: (also `golang:latest` actually works currently, but it's not reproducible...)

hack/verify/lint.sh
aojea: that's progress

BenTheElder: it's more like a minute on a cold run, but that's not bad considering how much static analysis we're getting. this already caught some bugs we hadn't before, just with the default checks.

pkg/internal/apis/config/encoding/scheme.go
BenTheElder: done

hack/go_container.sh
BenTheElder: fixed

hack/verify/shellcheck.sh
```diff
@@ -1,58 +0,0 @@
-#!/usr/bin/env bash
```
BenTheElder, line 1: this was unintentional, I think...

PR #806, opened 2019-08-23
Remove CoreDNS hack because of k/kubeadm#1745

aojea:
CoreDNS should be able to reload the configuration and work automatically, no need to delete the pods

Once CoreDNS 1.6.2 is merged we should verify it fixes the issue https://github.com/kubernetes/kubeadm/issues/1745

### DIFF
--- a/hack/ci/e2e.sh
+++ b/hack/ci/e2e.sh
@@ -130,8 +130,6 @@ run_tests() {
     echo "Patched CoreDNS config:"
     echo "${fixed_coredns}"
     kubectl apply -f - <<< "${fixed_coredns}"
-    # restart the pods
-    kubectl -n kube-system delete pods -l k8s-app=kube-dns
   fi
 
   # ginkgo regexes


aojea:
https://github.com/kubernetes/kubernetes/pull/81789

aojea:
This should work with the new CoreDNS version
#82127

aojea:
ipv6 job is working, we are ready to go

aojea:
ping @BenTheElder

hack/ci/e2e.sh
```diff
@@ -130,8 +130,6 @@ run_tests() {
     echo "Patched CoreDNS config:"
     echo "${fixed_coredns}"
     kubectl apply -f - <<< "${fixed_coredns}"
-    # restart the pods
-    kubectl -n kube-system delete pods -l k8s-app=kube-dns
```
BenTheElder: just to confirm: we only ever needed this for unreleased versions of kubernetes, right? (IE, 1.16 development commits)

aojea: exactly, this was a workaround for the 1.5.0 CoreDNS bump. AFAIK it was temporary because 1.16 released with 1.6.2

PR #809, opened 2019-08-23
use go_container.sh more consistently

BenTheElder:
so we get consistent formatting, linting, builds, etc.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,7 @@
 module sigs.k8s.io/kind
 
+go 1.13
+
 require (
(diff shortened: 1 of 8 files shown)


go.mod
```diff
@@ -1,5 +1,7 @@
 module sigs.k8s.io/kind
 
+go 1.13
```
BenTheElder: note: it automatically added something like this. this statement is just a suggestion to go for errors when the build fails, it is NOT a requirement or anything.

BenTheElder:
/retest
Prow pod pending timeout ....

PR #811, opened 2019-08-24
enable more lints

BenTheElder:
... and fix them


BenTheElder:
minor lint cleanup mostly, these should discourage potential future bugs though. in particular we are lucky that none of the things caught by scopelint were actual bugs _this time_, it would have caught the last image preloading bug.

this is not enabled for kindnetd yet.

PR #812, opened 2019-08-24
Lint kindnetd

BenTheElder:


cmd/kindnetd/cni.go
```diff
@@ -114,8 +114,8 @@ func (c *CNIConfigWriter) Write(inputs CNIConfigInputs) error {
 		os.Remove(f.Name())
 		return err
 	}
-	f.Sync()
-	f.Close()
+	_ = f.Sync()
+	_ = f.Close()
```
BenTheElder: on the fence about these (look at the context), but at least `_ =` is trivial to grep later. error overhaul comes after we revive the log shim work post 0.5 ...

hack/verify/lint.sh
```diff
@@ -23,6 +23,8 @@ cd "${REPO_ROOT}"
 # build golangci-lint
 SOURCE_DIR="${REPO_ROOT}/hack/tools" hack/go_container.sh \
   go build -o /out/golangci-lint github.com/golangci/golangci-lint/cmd/golangci-lint
+SOURCE_DIR="${REPO_ROOT}/hack/tools" GOOS="linux" hack/go_container.sh \
```
BenTheElder: we specifically need a linux binary regardless of the host to do kindnetd

tao12345666333: Do we need to add judgment here?
Although it doesn't matter, it just takes a little more space for `golangci-lint-linux` and repeats it when the OS is Linux.

PR #803, opened 2019-08-22
patch coreDNS config instead of replacing it outright

BenTheElder:
this should handle updates to the default config and multiple versions better

PR #813, opened 2019-08-24
Add internal domain to CoreDNS for IPv6

aojea:
Seems current sed regex is not adding the internal domain to the CoreDNS configmap, thus the DNS tests that run in prow with IPv6 fail

xref: https://github.com/kubernetes-sigs/kind/pull/803/files#r317355315

### DIFF
--- a/hack/ci/e2e.sh
+++ b/hack/ci/e2e.sh
@@ -122,7 +122,7 @@ run_tests() {
     local fixed_coredns
     fixed_coredns=$(
       sed \
-        -e 's/^.*kubernetes cluster\.local/\& internal/'
+        -e 's/^.*kubernetes cluster\.local/\& internal/' \
         -e '/^.*upstream$/d' \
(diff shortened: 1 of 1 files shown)


aojea:
--

hack/ci/e2e.sh
```diff
@@ -122,7 +122,7 @@ run_tests() {
     local fixed_coredns
     fixed_coredns=$(
       sed \
-        -e 's/^(\s)+.*cluster\.local.*$/\0kubernetes cluster.local internal in-addr.arpa ip6.arpa/' \
```
BenTheElder: Internal is right here...?

aojea: Is after cluster.local because prow uses that domain, if doesn't resolve, because we don't have upstream DNS servers, it answer no exist and the pod stops trying

BenTheElder: No I mean, the existing replace includes internal 🤷‍♂️

PR #815, opened 2019-08-26
allow to add extraArgs for containers

jannishuebl:
Hi, 

We need to give static IPs for the control node, with docker this is only possible when adding a container to a docker network, like already metioned in #273. 

My current solution is allowing extra args in the config for a node, which are passed to the docker run command. Of couse the solution discussed in #273 for static ips is better, but adding the possibility to give extraArgs still makes sense to me. It would give users more flexibilty. 

Since I needed this quickly I did an implementation, I am not used to program in go-lang, but for us this works well. But commends are welcome.

Cheers,

Jannis


jannishuebl:
ontributor License Agreement (CLA). agreed

aojea:
Just curiosity @jannishuebl , can you expand on your use case that requires static IP addresses?

PR #828, opened 2019-08-30
[Only for testing][part of #692] Kindnet 0.6.0 with dual stack support

aojea:
Some nice improvements:
* Dual-Stack support
* Automatically fetching the Cluster-CIDR, no more need to template the manifest in kind:
   1. First uses the POD_SUBNET env variable for backwards compatibility
   2. Get the podSubnet from the kubeadm configmap
   3. Get the cluster-cidr from the kube-proxy configmap
* Use of node annotations to discover multiple node IP addresses


aojea:
/hold
testing only, too much spaguetti :)

aojea:

there were only 2 failures, check if they are flaky

aojea:
hehe it seems to work 
ping @BenTheElder

aojea:
/hold cancel
/retest
ready for review

cmd/kindnetd/go.mod
```diff
@@ -3,27 +3,40 @@ module sigs.k8s.io/kind/cmd/kindnetd
 go 1.12
 
 require (
```
BenTheElder: why is this change so large? have you go mod tidy?

pkg/internal/cluster/create/nodes.go
```diff
@@ -165,7 +165,7 @@ func nodesToCreate(cfg *config.Cluster, clusterName string) []nodeSpec {
 	isHA := controlPlanes > 1
 	// obtain IP family
 	ipv6 := false
-	if cfg.Networking.IPFamily == "ipv6" {
+	if cfg.Networking.IPFamily == "ipv6" || cfg.Networking.IPFamily == "dual-stack" {
```
BenTheElder: the family isn't ipv4ipv6?

aojea: Indeed it did but was thinking that  dual-stack has better UX as a parameter, also is the way the feature is called in kubernetes
 I think ipv4ipv6 can be misleading, anyway, I'm totally open to any suggestion

cmd/kindnetd/go.mod
aojea: err, nope , still catching up with go modules :) let me try to redo it

pkg/internal/cluster/create/nodes.go
BenTheElder: we should be consistent with upstream.

khenidak: The constants upstream are "IPv4" and "IPv6" for `service.Spec.IPFamily`. We don't carry `dualstack` anywhere as a configuration value/type value. So it should be `IPv4` or `IPv6` or `dualstack`. I don't think i have seen `ipv4ipv6` reference anywhere else. it is always `dualstack ipv4/ipv6` or just `dualstack`.

BenTheElder: thanks!

hmm, so this is the IP family "at the cluster scope", `dualstack` could work, no hyphens? hyphen? :+)

cmd/kindnetd/main.go
```diff
@@ -100,43 +151,69 @@ func main() {
...
 			if err := cniConfig.Write(
 				ComputeCNIConfigInputs(node),
 			); err != nil {
 				return err
 			}
+			// update internal IPs node annotations until #42125 is fixed
+			// TODO: https://github.com/kubernetes/kubernetes/issues/4A2125
+			// in kind we can hardcode the outer interface but we can be smarter
+			// and find the interface matching the HOST_IP
+			node.Annotations["io.k8s.sigs.kind.kindnet/addresses"] = getNodeExternalIPs(hostIP)
```
BenTheElder: this is kinda gross :/

aojea: I know, but ... what are the alternatives to https://github.com/kubernetes/kubernetes/issues/42125?
using kubelet with the controller=external?

BenTheElder: yeah I think so, I'm exploring that now.

cmd/kindnetd/go.mod
aojea: can you take a look if is it right now?

cmd/kindnetd/main.go
aojea: Using CRDs?

aojea:
@BenTheElder bear in mind this is the kindnet part only, the full dual stack PR is here https://github.com/kubernetes-sigs/kind/pull/692

pkg/internal/cluster/create/nodes.go
thockin: API constant style would be `DualStack`.  We don't have an explict constant for that because we're not offering it for Services yet.

aojea: this was moved to the full dualstack PR with the name `DualStack`, I will use this PR only for kindnet

aojea:
@BenTheElder this new kindnet version is backwards compatible, all tests are passing, I'll rebase the dualstack PR on top of this

aojea:
/close
superseded by https://github.com/kubernetes-sigs/kind/pull/1255

PR #832, opened 2019-09-02
[WIP] Adding more detail to the v1alpha3 config godocs

alejandrox1:
This adds more details to the available fields and settings in the
kind.sigs.k8s.io/v1alpha3 config.

Signed-off-by: alejandrox1 <alarcj137@gmail.com>

### DIFF
--- a/pkg/apis/config/v1alpha3/doc.go
+++ b/pkg/apis/config/v1alpha3/doc.go
@@ -19,4 +19,70 @@ limitations under the License.
 //
 // +k8s:deepcopy-gen=package
 // +k8s:defaulter-gen=TypeMeta
+//
+// Package v1alpha3 defines the v1alpha version of the kind configuration file format.
+//
+// Basics
+//
+// The preferred way to configure kind is to pass an YAML configuration file with the --config option.
+//
+// kind supports the following configuration types:
+//
+//     apiVersion: kind.sigs.k8s.io/v1alpha3
+//     kind: Cluster
+//
+// The Cluster configuration type should be used to configure kind-cluster settings,
+// including settings for:
+//
+// - KubeadmConfigPatches are applied to the generated kubeadm config as
+// strategic merge patches to `kustomize build` internally
+// https://github.com/kubernetes/community/blob/master/contributors/devel/strategic-merge-patch.md
+// This should be an inline yaml blob-string.
+//
+// - KubeadmConfigPatchesJSON6902 are applied to the generated kubeadm config
+// as patchesJson6902 to `kustomize build`.
+//
+// - Networking contains cluster wide network settings.
+//
+// - Nodes contains the list of nodes defined in the kind cluster
+// If unset this will default to a single control-plane node.
+// Note that if more than one control plane is specified, an external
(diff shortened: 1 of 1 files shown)


alejandrox1:
Godoc does not build for branches but if anyone is interested in reviewing the changes locally you can try by running:
```
$ godoc -http=localhost:6060
```

and going to http://localhost:6060/pkg/sigs.k8s.io/kind/pkg/apis/config/v1alpha3/

pkg/apis/config/v1alpha3/doc.go
```diff
@@ -19,4 +19,70 @@ limitations under the License.
...
+// Basics
+//
+// The preferred way to configure kind is to pass an YAML configuration file with the --config option.
+//
+// kind supports the following configuration types:
+//
+//     apiVersion: kind.sigs.k8s.io/v1alpha3
+//     kind: Cluster
+//
+// The Cluster configuration type should be used to configure kind-cluster settings,
```
neolit123: kind-cluster -> kind cluster

pkg/apis/config/v1alpha3/doc.go
```diff
@@ -19,4 +19,70 @@ limitations under the License.
...
+//
+// - Networking contains cluster wide network settings.
+//
+// - Nodes contains the list of nodes defined in the kind cluster
+// If unset this will default to a single control-plane node.
+// Note that if more than one control plane is specified, an external
+// control plane load balancer will be provisioned implicitly.
+//
+// Here is a fully populated example of a single YAML file containing multiple
+// configuration types to be used during a `kind create cluster` run.
```
neolit123: should the "kind-example-config" from here, link to this new godocs page now?
https://kind.sigs.k8s.io/docs/user/quick-start/#configuring-your-kind-cluster

alejandrox1: Ah this is actually something that I wanted to run by you all 1) do we want a godocs page that goes into detail about the kind config or 2) add all this info to https://kind.sigs.k8s.io/docs/ ?

Option 1 is the "standard", most projects (that I've seen) that have a config link to their godocs for full details and there you can browse between the different versions. I.e., https://godoc.org/k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm#pkg-subdirectories

Option 2, we could put all these docs in the site itself, may be even dedicate an entire page for the config itself and give multiple config examples.

What do you think?
/cc @BenTheElder @munnerz @tao12345666333 @aojea

alejandrox1: ping @BenTheElder what do you think? is godoc worth it or should we stick to the official docs for this?

BenTheElder: I think we should follow up on changing the link, but imo getting godoc or something similar that covers the full type spec would be best long term

tao12345666333: IMO, kind is both a cli tool and a library, so maybe we should using option 1.

BenTheElder:
Doing a configuration docs page now, and moving on to newer API versions

Thank you for your effort on this

PR #834, opened 2019-09-03
move status out of log package

BenTheElder:
breaking out a very small bit of the logging work, clearing things out of the exported log package that don't particularly belong there before adding the shiny new logging ...

### DIFF
--- a/cmd/kind/kind.go
+++ b/cmd/kind/kind.go
@@ -94,17 +94,14 @@ func Run() error {
 
 // Main wraps Run and sets the log formatter
 func Main() {
-	// let's explicitly set stdout
-	log.SetOutput(os.Stdout)
+	// TODO(bentheelder): replace this with a shimmable logger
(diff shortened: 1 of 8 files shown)


cmd/kind/kind.go
```diff
@@ -94,17 +94,14 @@ func Run() error {
 
 // Main wraps Run and sets the log formatter
 func Main() {
+	// TODO(bentheelder): replace this with a shimmable logger
 	// let's explicitly set stdout
-	log.SetOutput(os.Stdout)
+	log.SetOutput(os.Stderr)
```
neolit123: possibly, modify the above comment `set stdout` or keep as is?

BenTheElder: that should say stderr for sure, using stdout was a bug but not related to the rest of this PR ...

BenTheElder: fixed

PR #835, opened 2019-09-04
use providerless build in CI

BenTheElder:
... and cleanup e2e.sh junk in the process, to avoid growing that script even more, and fix some typos

### DIFF
--- a/hack/ci/e2e.sh
+++ b/hack/ci/e2e.sh
@@ -23,19 +23,14 @@ REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)"
 
 # our exit handler (trap)
 cleanup() {
-  # always attempt to dump logs
   kind "export" logs "${ARTIFACTS}/logs" || true
   # KIND_IS_UP is true once we: kind create
   if [[ "${KIND_IS_UP:-}" = true ]]; then
     kind delete cluster || true
   fi
-  # clean up e2e.test symlink
   rm -f _output/bin/e2e.test || true
-  # remove our tempdir
-  # NOTE: this needs to be last, or it will prevent kind delete
-  if [[ -n "${TMP_DIR:-}" ]]; then
-    rm -rf "${TMP_DIR}"
-  fi
+  # remove our tempdir, this needs to be last, or it will prevent kind delete
+  [[ -n "${TMP_DIR:-}" ]] && rm -rf "${TMP_DIR:?}"
 }
 
 # install kind to a tempdir GOPATH from this script's kind checkout
@@ -45,8 +40,8 @@ install_kind() {
   export PATH="${TMP_DIR}/bin:${PATH}"
 }
 
-# build kubernetes / node image, e2e binaries
-build() {
+# build kubernetes / node image, e2e binaries, with bazel
+build_with_bazel() {
   # possibly enable bazel build caching before building kubernetes
   if [[ "${BAZEL_REMOTE_CACHE_ENABLED:-false}" == "true" ]]; then
     create_bazel_cache_rcs.sh || true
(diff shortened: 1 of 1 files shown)


hack/ci/e2e.sh
```diff
@@ -23,19 +23,14 @@ REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)"
 
 # our exit handler (trap)
 cleanup() {
-  # always attempt to dump logs
```
BenTheElder: superfluous, we're stating what the code does, which is obvious :man_shrugging:

hack/ci/e2e.sh
```diff
@@ -23,19 +23,14 @@ REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)"
...
   kind "export" logs "${ARTIFACTS}/logs" || true
   # KIND_IS_UP is true once we: kind create
   if [[ "${KIND_IS_UP:-}" = true ]]; then
     kind delete cluster || true
   fi
-  # clean up e2e.test symlink
```
BenTheElder: it's not a symlink, and this is obvious

hack/ci/e2e.sh
```diff
@@ -23,19 +23,14 @@ REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)"
...
   rm -f _output/bin/e2e.test || true
-  # remove our tempdir
-  # NOTE: this needs to be last, or it will prevent kind delete
-  if [[ -n "${TMP_DIR:-}" ]]; then
-    rm -rf "${TMP_DIR}"
-  fi
+  # remove our tempdir, this needs to be last, or it will prevent kind delete
+  [[ -n "${TMP_DIR:-}" ]] && rm -rf "${TMP_DIR:?}"
```
BenTheElder: this is equivalent but shorter, except it also now explodes if we get to `rm ...` and `TMP_DIR` is not set

hack/ci/e2e.sh
```diff
@@ -45,49 +40,49 @@ install_kind() {
...
   # build the node image w/ kubernetes
   kind build node-image --type=bazel --kube-root="$(go env GOPATH)/src/k8s.io/kubernetes"
-
   # make sure we have e2e requirements
-  #make all WHAT="cmd/kubectl test/e2e/e2e.test vendor/github.com/onsi/ginkgo/ginkgo"
   bazel build //cmd/kubectl //test/e2e:e2e.test //vendor/github.com/onsi/ginkgo/ginkgo
 
   # ensure the e2e script will find our binaries ...
   # https://github.com/kubernetes/kubernetes/issues/68306
+  # TODO: remove this, it was fixed in 1.13+
```
BenTheElder: when 1.16 lands we should consider how we want to keep testing kind with old kubernetes versions, our long tail is getting longer. I'd love to support users as far back as possible, but at some point it's probably going to get ridiculous :thinking:

hack/ci/e2e.sh
```diff
@@ -45,49 +40,49 @@ install_kind() {
...
   cat <<EOF > "${ARTIFACTS}/kind-config.yaml"
-# config for 1 control plane node and 2 workers
-# necessary for conformance
+# config for 1 control plane node and 2 workers (necessary for conformance)
 kind: Cluster
 apiVersion: kind.sigs.k8s.io/v1alpha3
 networking:
   ipFamily: ${IP_FAMILY:-ipv4}
 nodes:
-# the control plane node
```
BenTheElder: pointless comment

hack/ci/e2e.sh
```diff
@@ -153,29 +148,33 @@ run_tests() {
     -o=jsonpath='{range .items[*]}{.metadata.name}{"\t"}{.spec.taints}{"\n"}{end}' \
     | grep -cv "node-role.kubernetes.io/master" )"
 
-  # setting this env prevents ginkg e2e from trying to run provider setup
```
BenTheElder: typo fix

hack/ci/e2e.sh
```diff
@@ -153,29 +148,33 @@ run_tests() {
...
+  # setting this env prevents ginkgo e2e from trying to run provider setup
   export KUBERNETES_CONFORMANCE_TEST="y"
-  # run the tests
```
BenTheElder: pointless comment, if you don't know what this script is, it's probably best to go take a look at it, as it has plenty of surprising properties...

hack/ci/e2e.sh
```diff
@@ -153,29 +148,33 @@ run_tests() {
...
   ./hack/ginkgo-e2e.sh \
     '--provider=skeleton' "--num-nodes=${NUM_NODES}" \
     "--ginkgo.focus=${FOCUS}" "--ginkgo.skip=${SKIP}" \
     "--report-dir=${ARTIFACTS}" '--disable-log-dump=true'
 }
 
-# setup kind, build kubernetes, create a cluster, run the e2es
-main() {
```
BenTheElder: pointless main, as this isn't re-entrant, and we don't support `curl | sh` for this script.

BenTheElder:
after this PR it should be possible to build like before, but the default will now just be with docker / make, and will have the providerless tag. we want to get something like this into kubernetes blocking presubmits if possible.

hack/ci/e2e.sh
tao12345666333: Maybe we have to provide some extra support for some versions of EOL.

BenTheElder: we have some code in kind that deals with < 1.12 behavior, it's not so bad, but we can't continue to run presubmits for arbitrarily many kubernetes versions, at some point it's going to be prohibitive.

we might try something like a manual smoke test before releasing, or some limited periodic CI or ...

BenTheElder:
/test all
slower than I wanted

aojea:
> after this PR it should be possible to build like before, but the default will now just be with docker / make, and will have the providerless tag. we want to get something like this into kubernetes blocking presubmits if possible.

docker builds takes around 6 mins, is not slower?

BenTheElder:
... yes, see the comments above and testing the impact in CI.

However, it's also useful to have CI covering the make build and in particular that build tag, which will otherwise inevitably break on us soon when code freeze lifts.


We probably want _some_ jobs with this. The question is which and if it should be default.


Most users are probably not using the bazel mode.

BenTheElder:
Bazel in CI is only slightly faster for this mode on a good run though, it shows much bigger differences with the remote cache building everything vs just what kind builds.
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_kind/835/pull-kind-conformance-parallel-1-15/1170595011432026117/

matthyx:
Oh, man, I hate bash...
/lgtm

BenTheElder:
/retest
prow flake

BenTheElder:
moved to https://github.com/kubernetes-sigs/kind/pull/898

PR #839, opened 2019-09-05
Make CLI-output more rich

takmatsu:
Current `kind get clusters` command outputs simply the list of clusters.

This patch makes a little more rich like this.
```
kind get clusters
NAME    RUNNING   VERSION
kind    1/1       v1.17.0-alpha.0.147+9a7dace10f630c-dirty
multi   2/4       v1.15.0
```


BenTheElder:
this is pretty cool but I don't think we want to depend on k8s.io/kubernetes for a tool that tests k8s.io/kubernetes 🙃

takmatsu:
I see, removed that dependency.

pkg/cluster/nodes/node.go
```diff
@@ -238,6 +238,23 @@ func (n *Node) Role() (role string, err error) {
 	return role, nil
 }
 
+// IsRunning returns the status of the node
+func (n *Node) IsRunning() (status bool, err error) {
+	// retrive the IP address of the node using docker inspect
```
aojea: is retrieving the status not the IP, right?

cmd/kind/get/clusters/clusters.go
```diff
@@ -41,12 +45,29 @@ func NewCommand() *cobra.Command {
...
-	for _, cluster := range clusters {
-		fmt.Println(cluster.Name())
+	printer := util.GetNewTabWriter(os.Stdout)
+	fmt.Fprintln(printer, strings.Join([]string{"NAME", "RUNNING", "VERSION"}, "\t"))
+	for cluster, nodeList := range clusters {
+		size := strconv.Itoa(len(nodeList))
+		running := 0
+		version := ""
+		for _, n := range nodeList {
+			status, _ := n.IsRunning()
```
aojea: I don't know if we shouldn't show the nodes that belongs to the cluster but are not running

aojea:
I don't know if it would be better adding this as a new flag, same that kubectl does with the `-o wide` option. 
Also, I don't know if we should filter by status, is not the first time that doing some testing the nodes doesn't come up

pkg/cluster/nodes/node.go
takmatsu: That's right! Thanks.

cmd/kind/get/clusters/clusters.go
takmatsu: I think it should be shown because containers are down unexpectedly, but currently we cannot know if each cluster's all containers are up easily. That's one of the reason of this PR.

aojea: Ups, sorry, my bad, I didn't get it until now, you show the total and how many are running, that's great

BenTheElder:
so a couple problems:
- clusters do not necessarily have a single version
- we need to not break users that are piping this output
- we should be consistent across commands

I'd like to see an issue outlining the plans for these first.

thanks for the PR though!

PR #841, opened 2019-09-06
create and switch to log shim and minimal logger

BenTheElder:
This will make it easier to import kind into test harnesses (with some more work...)

The logging interface is a subset of the k8s.io/klog API, but without depending on k8s.io/klog at all.

By default, the packages will use a no-op logger that does nothing, so importing them will not spam the console ...

The CLI sets enables use of a simple logger that writes to stderr (and supports the CLI spinner magic).

Library users can either not have random logs (stick to noop) or wire up their logger against the klog-subset interfaces.

Additionally:
- deprecates `--loglevel [logrus level string]`
- introduces `-v  --verbose [info log verbosity level AKA debug noise level]`
- introduces `-q --quiet` AKA silence stderr / log output (though we need to fix a few places to respect this / not write to stderr via other means)

Major TODO: this currently uses a global logger instance, similar to how were using logrus's default global log instance. this is isolated to a temporary package for the purposes of keeping this PR simpler. In the future we may (will?) replace this package with explicitly passing down the logger as we improve the API surface more.


BenTheElder:
replaces #613 #614

BenTheElder:
most of the diff is mechanically changing usage of the previous log package to the new one. the other commits are smaller.

this logger is _super simple_ but already functional enough in some limited testing. I expect to refine it in follow ups before 0.6 comes out.

another important follow up will be managing error reporting better rebased on this stack, as part of that I'm working to get stack traces reported from _any_ errors that bubble all the way up.

BenTheElder:
this also drops logrus as a dependency

pkg/log/std.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// Errorf wraps Errorf on the standard logger
+func Errorf(format string, args ...interface{}) {
+	Get().Errorf(format, args...)
+}
+
+// V wraps V on the standard logger
+func V(level Level) InfoLogger {
+	return Get().V(level)
```
munnerz: teeny tiny nit that doesn't really need to be addressed: instead of `Get`, we could just use `std` directly here and avoid acquiring a lock for every call to the logger... seems quite wasteful (i.e. `logs.V(5).Info()` will acquire two locks just to call the func, and internally klog itself already has some buffering/locking stuff as well).

For the most part, it's unlikely this will change "in flight". IDK, tough call to make 😄 it's a shame we can't somehow denote the field as final or immutable easily. Perhaps we *can* by only allowing Set to be called once?

BenTheElder: this got refactored in a following commit, though it is functionally equivilant.

I agree, but I think the future is going to be plumbing through the logger in some form anyhow, and I suspect locking this mutex is not really noticeable in the scheme of kind's performance, we're largely waiting on docker / kubernetes.

pkg/globals/logger.go
```diff
@@ -0,0 +1,60 @@
...
+	SetLogger(&logger.Default{
+		Verbosity: verbosity,
+		Writer:    os.Stderr,
+	})
+}
+
+// GetLogger returns the standard logger used by this package
+func GetLogger() log.Logger {
+	globalLoggerMu.Lock()
+	defer globalLoggerMu.Unlock()
```
munnerz: Hm, now I'm getting confused.. there seems to be ~3 structures each with their own locks in this logging chain?

What's the `logs` package for (with its `Get` and `Set` functions)?

munnerz: d'oh.. just seen that this commit removes that stuff 🤦‍♂

cmd/kind/kind.go
```diff
@@ -76,14 +90,13 @@ func NewCommand() *cobra.Command {
 }
 
 func runE(flags *Flags) error {
-	level := defaultLevel
-	parsed, err := log.ParseLevel(flags.LogLevel)
-	if err != nil {
-		log.Warnf("Invalid log level '%s', defaulting to '%s'", flags.LogLevel, level)
+	// TODO: handle --loglevel which previously supported:
+	// panic, fatal, error, warning, info, debug, trace
```
munnerz: Not for or against this, but when is this TODO going to be done given we intend to remove this flag anyway? Seems like it either happens now, or doesn't happen (because by the time it comes to implementing it, we'll have removed the flag anyway)

cmd/kind/kind.go
```diff
@@ -76,14 +90,13 @@ func NewCommand() *cobra.Command {
...
+	if flags.Quiet {
+		globals.SetLogger(log.NoopLogger{})
```
munnerz: If I run `kind get kubeconfig-path -q`, will it suppress all output *except for* the KUBECONFIG path?

pkg/globals/logger.go
BenTheElder: the log package should now just be the types
this one has a mutex to protect the global

there's also mutexes:
- inside the default logger for synchronizing writing (one log message at a time, so they aren't interleaved)
- inside the spinner stuff to synchronize starting / stopping it safely and flushing the console

cmd/kind/load/docker-image/docker-image.go
```diff
@@ -119,7 +119,7 @@ func runE(flags *flagpole, args []string) error {
 		id, err := node.ImageID(imageName)
 		if err != nil || id != imageID {
 			selectedNodes = append(selectedNodes, node)
-			log.Debugf("Image: %q with ID %q not present on node %q", imageName, imageID, node.String())
+			globals.GetLogger().V(0).Infof("Image: %q with ID %q not present on node %q", imageName, imageID, node.String())
```
munnerz: Do we need `V(0)` here? The klog specification says that `klog.Info` and `klog.V(0).Info` are semantically equivalent afaik?

cmd/kind/kind.go
BenTheElder: that's the plan

cmd/kind/load/docker-image/docker-image.go
munnerz: Also, do we want to consider maybe creating an internal wrapper package or something so we can just call `log.Info` instead of `globals.GetLogger().Info` every time? It could quite simply wrap the default configured logger 😄

cmd/kind/kind.go
BenTheElder: i think we probably need at least one release where the flag is still accepted but either just triggers a warning or triggers a warning and guesses at setting a roughly equivalent verbosity (like, if you were setting DEBUG, maybe v3, if you were setting TRACE, maybe maximum verbosity)

then the _next_ release would drop it entirely.

otherwise we risk surprise breaking a bunch of CI. OTOH we _could_ do that since we're in alpha, but it seems maybe better to put it in the release notes and print a warning about migrating?

this is TODO because I wasn't sure what the mapping should be yet.

cmd/kind/load/docker-image/docker-image.go
BenTheElder: our interface is slightly simplified by only having Info on V()

BenTheElder: yeah probably? but also we can grep for this and be like "bad globals, bad!" :P

BenTheElder: in a lot of these places we could have also stored the return of `GetLogger()` but this was a pretty mechanical port (read: lots of sed)

cmd/kind/kind.go
BenTheElder: FWIW:
```
$ kind get kubeconfig-path -q
/Users/bentheelder/.kube/kind-config-kind
```

cmd/kind/kind.go
BenTheElder: I think the main thing we want to not break immediately is all the CI setups enabling higher levels to get more debug info.

If we want to leave support in the next release then we should:

Detect if the flag was set, if so:
- print a deprecation warning
- if -v was NOT set, map trace to max int32, debug to 3, and anything else just to 0

But I can also see an argument for just breaking it and moving on, we already don't guarantee that arbitrary kind commits are ready for usage, and the release notes will contain a breaking change notice for this in the next release...

munnerz: I like the idea of not breaking the CLI for a release or so, but IMO just writing "WARNING: THE OPTION YOU SPECIFIED ACTUALLY DOESN'T DO WHAT YOU EXPECT ANYMORE, TRY USING -v".

If we don't implement the mapping logic in this PR, I think we're unlikely to implement it before we just remove the flag entirely

BenTheElder: I'll add a stab at this in the morning. I don't think it will be much work, IIRC it's possible in pflag to tell if a flag was specified or not as well 😅

BenTheElder: done

pkg/internal/util/cli/status.go
```diff
@@ -34,24 +34,28 @@ type Status struct {
...
 
-// StatusForLogger returns a status object for the logger l,
+// StatusForLogger returns a new status object for the logger l,
 // if l is the kind default logger it will inject a wrapped writer into it
 // and if we've already attached one it will return the previous status
 func StatusForLogger(l log.Logger) *Status {
 	s := &Status{
 		logger: l,
 	}
 	if v, ok := l.(*logger.Default); ok {
```
BenTheElder: the CLI should probably explicitly setup a spinner capable logger so this logger can otherwise be used without this stuff. I'll make the logger nicer in a future commit to keep this PR from being massive

PR #846, opened 2019-09-07
WIP Use memory to store etcd data

aojea:
The /var/lib folder is to suppose to store only the data needed by programs as they run, due to the nature of containers we shouldn't need it once the containers are stopped.
This adds a boost on performance because the software using that folder, specially etcd, use memory directly instead of disks for storage

### DIFF
--- a/pkg/cluster/nodes/create.go
+++ b/pkg/cluster/nodes/create.go
@@ -145,6 +145,7 @@ func createNode(name, image, clusterLabel, role string, mounts []cri.Mount, port
 		// runtime temporary storage
 		"--tmpfs", "/tmp", // various things depend on working /tmp
 		"--tmpfs", "/run", // systemd wants a writable /run
+		"--tmpfs", "/var/lib/etcd", // etcd needs fast storage
 		// runtime persistent storage
 		// this ensures that E.G. pods, logs etc. are not on the container
 		// filesystem, which is not only better for performance, but allows


aojea:
🤔 is this related to this change?

> W0907 21:51:12.421] 	[WARNING ImagePull]: failed to pull image k8s.gcr.io/kube-controller-manager:v1.13.11-beta.0.18_fccd9fb754965b: output: time="2019-09-07T21:47:02Z" level=fatal msg="pulling image failed: rpc error: code = Unknown desc = failed to resolve image \"k8s.gcr.io/kube-controller-manager:v1.13.11-beta.0.18_fccd9fb754965b\": no available registry endpoint: k8s.gcr.io/kube-controller-manager:v1.13.11-beta.0.18_fccd9fb754965b not found"
W0907 21:51:12.421] , error: exit status 1

pkg/cluster/nodes/create.go
```diff
@@ -145,6 +145,9 @@ func createNode(name, image, clusterLabel, role string, mounts []cri.Mount, port
 		// runtime temporary storage
 		"--tmpfs", "/tmp", // various things depend on working /tmp
 		"--tmpfs", "/run", // systemd wants a writable /run
+		// FHS State information. Persistent data modified by programs as they run,
+		// e.g., databases, packaging system metadata, etc.
+		"--tmpfs", "/var/lib",
```
BenTheElder: You shouldn't make all of var a tmpfs.

BenTheElder: This will consume inordinate memory for one.

aojea: I see the mistake now 😓

aojea:
there is no significant difference in the CI, will try to document it, I think that can be done only with mounts

PR #853, opened 2019-09-11
Fix quickstart releases link

manics:
Fixes the `[releases]` link on https://kind.sigs.k8s.io/docs/user/quick-start/

### DIFF
--- a/site/content/docs/user/quick-start.md
+++ b/site/content/docs/user/quick-start.md
@@ -26,7 +26,7 @@ shown [here](https://golang.org/doc/code.html#GOPATH) if you encounter the error
 
 Without installing go, kind can be built reproducibly with docker using `make build`.
 
-Stable binaries are also available on the [releases](https://github.com/kubernetes-sigs/kind/releases) page.
+Stable binaries are also available on the [releases] page.
 Stable releases are generally recommended for CI usage in particular.
 To install, download the binary for your platform from "Assets" and place this
 into your `$PATH`. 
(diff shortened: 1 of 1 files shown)


site/content/docs/user/quick-start.md
```diff
@@ -26,8 +26,8 @@ shown [here](https://golang.org/doc/code.html#GOPATH) if you encounter the error
 
 Without installing go, kind can be built reproducibly with docker using `make build`.
 
-Stable binaries are also available on the [releases] page. Stable releases are
-generally recommended for CI usage in particular.
+Stable binaries are also available on the [releases](https://github.com/kubernetes-sigs/kind/releases) page.
```
BenTheElder: can we define this at the end of the file instead please?

BenTheElder:
/approve
/hold
one request

site/content/docs/user/quick-start.md
manics: Could you clarify what you mean please- do you want me to move the whole paragraph to the end?

aojea: Check how the other links are defined , the URL is at the end of the file

manics: @BenTheElder @aojea Thanks, now fixed. There's an in-line URL in the paragraph above (`[here](https://golang.org/doc/code.html#GOPATH)`) which confused me

BenTheElder: ah, we should move that one too, but in another PR. thanks!

BenTheElder:
/lgtm
/hold cancel
thank you!

PR #855, opened 2019-09-11
cleanup error packages

BenTheElder:
We have multiple packages for dealing with errors, consolidate them into one drop-in helper package.

Also begin some minor shimming over `github.com/pkg/errors` given that the API and implementation are excellent but EOL.
https://github.com/pkg/errors#roadmap


BenTheElder:
importantly it makes it possible to use the utilities to get the stacktrace etc. out of errors from consuming test harnesses

cmd/kind/build/baseimage/baseimage.go
```diff
@@ -17,8 +17,8 @@ limitations under the License.
 package baseimage
 
 import (
-	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	"sigs.k8s.io/kind/pkg/errors"
```
BenTheElder: nit: should be grouped with the packages below

BenTheElder: fixed

PR #857, opened 2019-09-12
Add instructions for using bash completions

Wetrain:
This adds some instructions for enabling bash completions when using kind for a Bash shell.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,23 @@ To use kind, you will need to [install docker].
 Once you have docker running you can create a cluster with `kind create cluster`  
 To delete your cluster use `kind delete cluster`
 
+### Auto completions
+
+To enable Bash completion for Kind commands, you can use the bash-completions project:
+
+```console
+sudo apt | dnf | pacman install bash-completion
+kind completion bash | sudo tee /etc/bash_completion.d/kind
+```
+
+Then reload your .bashrc:
+
+```console
+source ~/.bashrc
+```
+
+You should now be able to <tab> complete for kind options!
+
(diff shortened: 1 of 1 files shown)


BenTheElder:
CLA still doesn't appear to be passing.

Wetrain:
Hi sorry seems my email was wrong, i've corrected this now

Wetrain:
Any update on getting this merged? :D

BenTheElder:
I'd like to keep the readme high level and focused on the few quick things necessary to get an idea of what kind is, roughly how you'd use it, and why you might want to, as you can see we have a strong callout to the full docs at the top.

Ideally this would be a new guide under the user content, with zsh instructions.

README.md
```diff
@@ -51,6 +51,23 @@ To use kind, you will need to [install docker].
 Once you have docker running you can create a cluster with `kind create cluster`  
 To delete your cluster use `kind delete cluster`
 
+### Auto completions
+
+To enable Bash completion for Kind commands, you can use the bash-completions project:
+
+```console
+sudo apt | dnf | pacman install bash-completion
```
BenTheElder: users are going to cut and paste this and run into it not being a real command, so we should split it out and instead point to the upstream install docs.

Wetrain:
@BenTheElder Ok so this should be moved to the site repository with some instructions for each distro?

BenTheElder:
yes please, a new page under the user guide.

sorry for the delay -- missed this but going back through the tracker now.

BenTheElder:
let's file an updated PR to do this as a guide page under the user docs.

PR #875, opened 2019-09-27
Upgrade base image

BenTheElder:
- upgrade to ubuntu 19.10 (pre-release) which should include containerd 1.2.9
- upgrade crictl to v1.16.1
- upgrade CNI to v0.8.2
- bump the base image

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -19,7 +19,7 @@
 # https://developers.redhat.com/blog/2014/05/05/running-systemd-within-docker-container/
 # https://developers.redhat.com/blog/2016/09/13/running-systemd-in-a-non-privileged-container/
 
-ARG BASE_IMAGE="ubuntu:19.04"
+ARG BASE_IMAGE="ubuntu:19.10"
 FROM ${BASE_IMAGE}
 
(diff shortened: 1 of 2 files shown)


aojea:
/lgtm

What's the size difference by the way?

BenTheElder:
it's slightly larger, but I'm more concerned with getting the fixes at the moment.

tao12345666333:
/lgtm

And CNI v0.8.2 has not merged into upstream https://github.com/kubernetes/kubernetes/pull/78819 (just for ref

BenTheElder:
/retest
prow flake

aojea:
> /lgtm
> 
> And CNI v0.8.2 has not merged into upstream [kubernetes/kubernetes#78819](https://github.com/kubernetes/kubernetes/pull/78819) (just for ref

there is an email thread discussing the CNI dependency https://groups.google.com/d/msg/kubernetes-sig-network/hlwUInyCDHI/A-j72fATAgAJ

Short history first is that is not a kubernetes dependency and it depends on the CNI plugin to choose the version and the binary, I think that for kindnet we should stay with the latest gratest

tao12345666333:
Thanks @aojea :+1:

PR #878, opened 2019-09-27
remove custom ctr binary from base image

BenTheElder:
the upstream binaries have been released with the patch we wanted in ctr, this actually reduces the base image size slightly.


aojea:
if github UI is right that patch is only in containerd 1.3.0 https://github.com/containerd/containerd/pull/3259/commits/cb7c780af2394ab08d5d8a3932ca7437074ae179 , and we are on 1.2.9

BenTheElder:
:disappointed: I thought this was in one of the 1.2.x releases previously and we're now on the latest 1.2.x :sigh:

BenTheElder:
let's get on 1.3 anyhow

PR #879, opened 2019-09-27
the grand node refactor

BenTheElder:
This is **NOT** the final state I'd like to get node management into, but it is in a functional enough state.

High Level:
- there's a `Provider` interface (internal) that handles: Provisioning (machine-like node creation based on config, NOT kubeadm init etc), Listing Clusters, Listing Nodes, Deleting nodes ...
  - externally surfaced via the `pkg/cluster` package which is a tad higher level
- there's a `Node` interface (public) that allows executing commands, getting the role etc.
- there's a `docker` implementation of these interfaces
- most operations now use these interfaces (not quite all yet!)
- minor streamlining of bootstrap 
  - admin config copied to all control plane nodes so we don't need to keep locating the bootstrap node
  - minor unification of constants and code reorganization
- lots of dead code removal
- generally moving things away from the `docker` package (except image build) so we can use mocks and non-docker implementations

Unfortunately every minor change towards this pulled the thread on everything else that depends on it.

I'd like to do a lot more work streamlining the code speeding up boot, but this PR is already huge so I'm going to try to finish the key functional changes related to the node interfaces first.


BenTheElder:
The commit history is terrible (sorry, like I said decoupling these changes was hard because we inherently had high coupling for these packages) and there is still one place we directly use docker to manipulate nodes (copying out the certs) that needs some more thought before replacing.

which bits are and aren't public are also not quite right yet, but in the end state it should be easy to provision:
- clusters (on a particular provider)
- bare nodes (pre-kubeadm etc.)
and to:
- manipulate nodes
- list clusters
- delete clusters
- ...

while keeping the exact details well contained (eg ideally only the single docker provider package is involved in execing docker during `kind create cluster`, everything else deals with a node provider interface or a node interface, which are higher level and much more mockable...)

BenTheElder:
/assign @aojea @amwat 

I'll do some more cleanup, but this is roughly it. most of the code is just moved either:
- behind an interface
- to a method that takes an interface instead of concretely being on the one-true-node-type

BenTheElder:
there are some other functional changes that should follow, like the loadbalancer node should NOT have the same docker flags as the worker nodes, but I tried to avoid many functional changes in this refactor given the scope is already large

cmd/kind/get/kubeconfig/kubeconfig.go
```diff
@@ -26,7 +26,7 @@ import (
 	"sigs.k8s.io/kind/pkg/errors"
 
 	"sigs.k8s.io/kind/pkg/cluster"
-	clusternodes "sigs.k8s.io/kind/pkg/cluster/nodes"
+	clusternodes "sigs.k8s.io/kind/pkg/internal/cluster/nodes"
```
neolit123: yeah...this will break kinder, so we have to "third_party" it.
cc @fabriziopandini

pkg/internal/cluster/providers/provider.go
```diff
@@ -0,0 +1,40 @@
...
+package providers
+
+import (
+	"sigs.k8s.io/kind/pkg/cluster/nodes"
+
+	"sigs.k8s.io/kind/pkg/internal/apis/config"
+)
+
+// Provider represents a provider of cluster / node infrastructure
+type Provider interface {
```
neolit123: nice

cmd/kind/get/kubeconfig/kubeconfig.go
BenTheElder: see the later commits

pkg/internal/cluster/create/actions/config/config.go
```diff
@@ -95,24 +81,17 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 		IPv6:                 ctx.Config.Networking.IPFamily == "ipv6",
 	}
 
-	fns = append(fns, func() error {
-		return writeKubeadmConfig(ctx.Config, configData, node)
-	})
-
-	// create the kubeadm join configuration for secondary control plane nodes if any
-	secondaryControlPlanes, err := nodeutils.SecondaryControlPlaneNodes(allNodes)
+	// create the kubeadm join configuration for ontrol plane nodes
```
neolit123: `ontrol` -> `control`

cmd/kind/get/kubeconfig/kubeconfig.go
BenTheElder: we still support:
- provisioning a cluster 
- getting some node details and execing against it

however the API itself breaks, and some of the utilities for interacting with nodes may become public or private over time (to reflect how stable they are)

each of those are much smaller though, the bulk of the complicated code here (provisioning "machines", supporting the node interface to exec etc.) will stay usable via public APIs, but I think those need some more reshaping (cluster.Context is weird now and needs a slight rethink, but should provide the same functionality ish)

BenTheElder: again, this PR will _not_ be the final state for that as it's just way too big of a change, `cmd/` still imports the old docker package for example, I will ~eliminate that but ensure that the same level of power is available through a more deliberate API.

neolit123: ok, i saw the later commits.
understood.

BenTheElder: I should probably squash this PR, the commits are wildly non-meaningful, I spent a bit of time refactoring the history but didn't come up with a clean version with small commits

pkg/internal/cluster/providers/provider.go
BenTheElder: once everything is retargeted on this I'm going to start using a mock one in a few places and then podman, ignite etc. :D

between this and the node interface, almost all of the rest of the code is much higher level and no longer needs to think in terms of `docker exec ...` etc. and depends on a much narrower set of methods.

the docker provider has quite a lot of internal code that used to be spread over several packages.

the reusable bits I've been moving into a shared library though (like getting an open port on the host) to ease implementing podman etc.

pkg/internal/cluster/create/actions/config/config.go
BenTheElder: fixed

pkg/internal/cluster/providers/docker/provider.go
```diff
@@ -0,0 +1,162 @@
...
+type Provider struct{}
+
+// Provision is part of the providers.Provider interface
+func (p *Provider) Provision(status *cli.Status, cluster string, cfg *config.Cluster) (err error) {
+	// TODO: validate cfg
+	// ensure node images are pulled before actually provisioning
+	ensureNodeImages(status, cfg)
+
+	// actually provision the cluster
+	// TODO: strings.Repeat("📦", len(desiredNodes))
```
amwat: I would prefer `📦 x n` :)

pkg/internal/cluster/providers/docker/provider.go
```diff
@@ -0,0 +1,162 @@
...
+		// filter for nodes with the cluster label
+		"--filter", "label="+constants.ClusterLabelKey,
+		// format to include the cluster name
+		"--format", fmt.Sprintf(`{{.Label "%s"}}`, constants.ClusterLabelKey),
+	)
+	lines, err := exec.OutputLines(cmd)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to list clusters")
+	}
+	return lines, nil
```
amwat: won't this repeat the cluster name for every node in multi node clusters? Probably dedup this in a sets.String

pkg/cluster/clusters.go
```diff
@@ -17,21 +17,13 @@ limitations under the License.
...
-		return nil, errors.Wrap(err, "could not list clusters")
-	}
-	clusters := []Context{}
-	for name := range n {
-		clusters = append(clusters, *NewContext(name))
-	}
-	return clusters, nil
+// TODO: refactor this to not assume a particular provider
+func List() ([]string, error) {
+	return docker.NewProvider().ListClusters()
```
amwat: providers.New("docker") ?
although there is only one for now so fine to cleanup later.

BenTheElder: Yeah, there's a TODO to refactor how this package interacts with them before we add more.

BenTheElder: Just keeping the diff down a little

pkg/internal/cluster/providers/docker/provider.go
BenTheElder: Ack. This is a bug.

BenTheElder: fixed

pkg/internal/cluster/create/actions/kubeadmjoin/join.go
```diff
@@ -167,12 +170,13 @@ func runKubeadmJoinControlPlane(
 		containerPath := path.Join("/etc/kubernetes/pki", fileName)
 		// set the path of the certificate into the tmp area on the host
 		tmpPath := filepath.Join(tmpDir, fileName)
+		// TODO(bentheelder): refactor to copy between nodes instead
 		// copies from bootstrap control plane node to tmp area
-		if err := controlPlaneHandle.CopyFrom(containerPath, tmpPath); err != nil {
+		if err := docker.CopyFrom(controlPlaneHandle.String(), containerPath, tmpPath); err != nil {
```
BenTheElder: this is a major TODO, however it still works for now.

this and the call below must be replaced before we can implement another provider.

this is the last thorny change. I have a plan for it, but I want to stop increasing the size of this PR and iterate since it is still _functionally correct_ for now, I know it took a while to read (thank you @amwat @neolit123 🙏 )

BenTheElder:
Merging and iterating. Will try to keep the next one smaller :-)
/hold cancel

PR #880, opened 2019-09-28
Enable IPv6 on nodes

aojea:
We have to enable ipv6 inside the nodes

https://docs.docker.com/engine/reference/commandline/run/#configure-namespaced-kernel-parameters-sysctls-at-runtime

### DIFF
--- a/pkg/internal/cluster/providers/docker/provision.go
+++ b/pkg/internal/cluster/providers/docker/provision.go
@@ -58,6 +58,8 @@ func provision(cluster string, cfg *config.Cluster) error {
 		apiServerAddress = "127.0.0.1" // only the LB needs to be non-local
 	}
 
+	isIPv6 := cfg.Networking.IPFamily == "ipv6"
+
 	// plan node creation
 	createNodeFuncs := []func() error{}
 	for _, configNode := range cfg.Nodes {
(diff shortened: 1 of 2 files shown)


aojea:
/assign @BenTheElder 

don't know if in the long term is better to run the commands on the nodes or use the runtime options (need to check how it works for other runtimes)

pkg/internal/cluster/providers/docker/provision.go
```diff
@@ -48,6 +48,7 @@ func provision(cluster string, cfg *config.Cluster) error {
 		}
 	}
 	isHA := controlPlanes > 1
+	isIPv6 := cfg.Networking.IPFamily == "ipv6"
```
tao12345666333: Is there a need to add a `strings.ToUpper` method here?   Allow users to use `IPv6`/`ipv6` etc.

pkg/internal/cluster/providers/docker/provision.go
```diff
@@ -56,6 +57,10 @@ func provision(cluster string, cfg *config.Cluster) error {
 	if isHA {
 		apiServerPort = 0              // replaced with a random port
 		apiServerAddress = "127.0.0.1" // only the LB needs to be non-local
+		if isIPv6 {
+			apiServerAddress = "::1" // only the LB needs to be non-local
```
BenTheElder: Have you tested this? We were not doing this before as far as I could tell.

BenTheElder:
/lgtm
/approve
Not a fan of the extra bool parameter but it is internal. 
The listen addr I think is correct but is different behavior versus before and should be tested offline and followed up.
We'll wind up reimplementing the exec way for other providers but it's trivial and this should be more efficient.

pkg/internal/cluster/providers/docker/provision.go
BenTheElder: No, prefer handling this in the parsing layer.

We do validation and exact strings currently

pkg/internal/cluster/providers/docker/provision.go
aojea: will test it and report back

PR #881, opened 2019-09-28
README: Added usage with kubectl

albertvaka:
It's kind of basic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ kind bootstraps each "node" with [kubeadm][kubeadm]. For more details see [the d
 
 **NOTE**: kind is still a work in progress, see the [1.0 roadmap].
 
-## Installation and usage
+## Installation
 
 You can install kind with `GO111MODULE="on" go get sigs.k8s.io/kind@v0.5.1`.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -47,6 +47,8 @@ chmod +x ./kind
 mv ./kind /some-dir-in-your-PATH/kind
 ```
 
+## Usage
+
 To use kind, you will need to [install docker].
```
neolit123: this sentence seems more relevant to the installation section.

albertvaka: Fixed!

README.md
```diff
@@ -59,6 +62,12 @@ To create a cluster from Kubernetes source:
 Multi-node clusters and other advanced features may be configured with a config
 file, for more usage see [the docs][user guide] or run `kind [command] --help`
 
+To make `kubectl` use your local kind cluster, use:
```
BenTheElder: We already print a per-platform relevant usage when you run kind create cluster. I'd rather not duplicate that here. For example this version is now lacking windows.

albertvaka: I probably missed that when I first created a cluster... But you have a point, closing this then!

BenTheElder:
We should prefer keeping minimal details in the readme just to illustrate: what it is / why it exists, and how simple it is to install and run, otherwise the docs site should cover just about everything else.

Duplicating content makes it less likely to stay up to date.

README.md
BenTheElder: Thanks anyhow though, I appreciate documentation efforts especially :-)

PR #885, opened 2019-09-30
Update to Kustomize 3.2.0

jbrette:
This PR does three things:

- Update to kustomize 3.2.0
- Solve the bug related to "patches:". That field was deprecated in kustomize 2.x. User were supposed to patchesStrategicMerge instead.
- Use the types.kustomization coming from kustomize which will ensure that the structure of the generated kustomization.yaml is aligned with the version of kustomize picked by the kind team.


aojea:
it will be good to test if patching works before merging,  I guess we are not testing patching in the CI

neolit123:
> it will be good to test if patching works before merging, I guess we are not testing patching in the CI

isn't kind applying patches to the kubeadm config as part of CI?

jbrette:
I was trying to figure out how it pass testing, because [patches: used here](https://github.com/kubernetes-sigs/kind/blob/master/pkg/internal/util/kustomize/kustomize.go#L65) has been deprecated for a long time still accepted in [2.0.3](https://github.com/kubernetes-sigs/kustomize/blob/v2.0.3/pkg/types/kustomization.go#L162), and repurposed in 3.1.x. The users were supposed to use patchesStrategicMerge.

The issues are:
1. There is no real unit test in kind for the usage of "patches:" field by kustomize.

2. There are some tests in kubeadm/kinder but kubeadm/kinder/go.mod still points at kind 0.4.0 which in turns uses kustomize 2.0.3 (Which explains which kubeadm/kinder unit tests are still working). [Check here](https://github.com/kubernetes-sigs/kind/blob/v0.4.0/go.mod#L25)

3. kubernetes/cmd/kubeadm is not using kind....but is still using kustomize [2.0.3](https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/util/kustomize/kustomize.go#L33) and is using [patchStrategicMerges:](https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/util/kustomize/kustomize_test.go#L109)

Ideally there some needs from more unit tests in the kind, and also a need to bump kubeadm/kinder to/go.mod to a newer version of kind.

aojea:
I guess that at least we should try something like this in the config file

```
  kubeadmConfigPatches:
  - |
    kind: ClusterConfiguration
    metadata:
      name: config
    networking:
      serviceSubnet: 10.0.0.0/16
  # patch it further using a JSON 6902 patch
  kubeadmConfigPatchesJson6902:
  - group: kubeadm.k8s.io
    version: v1beta1
    kind: ClusterConfiguration
    patch: |
      - op: add
        path: /apiServer/certSANs/-
        value: my-hostname
```

BenTheElder:
/lgtm
/approve
will file follow-up with nits and tests. I'm working on tests today anyhow.

PR #892, opened 2019-09-30
get the unit test ball rolling

BenTheElder:
- tidy up kustomize package
- add **tiny** assert package (rather than keep re-writing this, or pulling in another dep)
- stub in unit testing kutomize package (more test cases to come later)
- add a hermetic target to invoke unit tests `make test` (runs in the go build container)
  - export coverage profile


pkg/internal/util/kustomize/kutomize_test.go
```diff
@@ -0,0 +1,216 @@
...
+func TestBuild(t *testing.T) {
+	type testCase struct{
+		Name string
+		Resources []string
+		Patches []string
+		PatchesJSON6902 []config.PatchJSON6902
+		ExpectError bool
+		ExpectOutput string
+	}
+	cases := []testCase{
```
BenTheElder: these are certainly not sufficient, however some is better than none.

next up I'm going to start tracking our unit test coverage in CI, but I wanted to have a package to test it out with, so that will be this package.

if everything works we'll see that the coverage is subpar, and then after a follow-up it will go up.

Makefile
```diff
@@ -57,4 +57,8 @@ clean-output:
 # standard cleanup target
 clean: clean-output clean-cache
 
-.PHONY: all kind build install clean-cache clean-output clean
+# unit tests (hermetic)
+test:
+	GOOS=linux hack/go_container.sh go test ./...
```
BenTheElder: TODO: we'll need to add something that dumps the cover profile and invoke gopherage (pinned in hack/tools/go.mod possibly? might even want another module for that as test-infra is a mess to import...)

BenTheElder: dumping the cover profile now, just need to have CI copy it to artifacts and render it.

pkg/internal/util/kustomize/kutomize_test.go
```diff
@@ -0,0 +1,216 @@
...
+			Patches:     []string{"b o g u s"},
+			ExpectError: true,
+		},
+	}
+	for _, tc := range cases {
+		tc := tc // capture test case
+		t.Run(tc.Name, func(t *testing.T) {
+			t.Parallel()
+			out, err := Build(tc.Resources, tc.Patches, tc.PatchesJSON6902)
+			assert.ExpectError(t, tc.ExpectError, err)
```
neolit123: ExpectError contains Errorf, but should it be Fatalf, so that the test exits right away?
in this case it's fine as `out` is a string that is later compared.

perhaps ExpectError should have an optional `bool fatal` flag.

BenTheElder: I didn't need either of those behaviors currently, if we need it later we can adapt it, otherwise YAGNI

BenTheElder: continuing on is often preferable to produce more information about other things that do not check out.

neolit123: ok, sgtm.

neolit123:
LGTM on this set of commits.
/lgtm
/hold

BenTheElder:
/hold cancel
will follow up with more after I get CI running and rendering these (will require test-infra work)

BenTheElder:
we have junit and coverage in CI now, though the coverage results are currently slightly misleading (packages with no tests are ignored)

will follow through more soon, switching gears to e2e fixups for a bit

PR #893, opened 2019-10-01
get junit output in tests

BenTheElder:
- use gotestsum to get junit output (the same as what we chose for k/k)
- handle copying to artifacts so it shows up in CI
- filter out `zz_generated.*` in code coverage
- dedupe unit testing in presubmit


BenTheElder:
need to update the test-infra job...

BenTheElder:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kind-unit/1178845969123381256 :tada:

aojea:
impressive

PR #894, opened 2019-10-01
add CRI "test-handler"

BenTheElder:
... so RuntimeClass tests can pass on kind

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -80,10 +80,19 @@ RUN clean-install \
 COPY 10-limits.conf 10-restart.conf /etc/systemd/system/containerd.service.d/
 
 # debug containerd version and create default config
-# additionally, disable some plugins we don't use / support
+# additionally:
+# - disable some plugins we don't use / support
+# - setup the "test-handler" used by kubernetes runtime class testing
 RUN containerd --version \
     && mkdir -p /etc/containerd \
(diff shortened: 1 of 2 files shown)


BenTheElder:
xref:
https://github.com/kubernetes/test-infra/blob/bd64e9a6a8c376e25b124810024b256318d6bb1f/jobs/e2e_node/containerd/config.toml#L19-L21
```toml
# Runtime handler used for runtime class test.
[plugins.cri.containerd.runtimes.test-handler]
  runtime_type = "io.containerd.runtime.v1.linux"
```

images/base/Dockerfile
```diff
@@ -80,10 +80,19 @@ RUN clean-install \
...
 # debug containerd version and create default config
-# additionally, disable some plugins we don't use / support
+# additionally:
+# - disable some plugins we don't use / support
+# - setup the "test-handler" used by kubernetes runtime class testing
 RUN containerd --version \
     && mkdir -p /etc/containerd \
-    && echo 'disabled_plugins = ["aufs", "btrfs", "zfs"]' > /etc/containerd/config.toml \
+    && ( \
+        echo "# disable plugins we don't / can't support" \
```
BenTheElder: bah, this doesn't work right in a dockerfile

BenTheElder: https://github.com/kubernetes-sigs/kind/pull/895

aojea: :-)

PR #897, opened 2019-10-01
updates rp_filter to 1

### LINKED ISSUE #891: installing calico requires a change to net.ipv4.conf.all.rp_filter
**What happened**:
When deploying calico against kind I used the following kind config:
```
kind: Cluster
networking:
  disableDefaultCNI: True
nodes:
- role: control-plane
- role: worker
- role: worker
kubeadmConfigPatches:
- |
  kind: ClusterConfiguration
  metadata:
    name: config
  networking:
    serviceSubnet: "10.96.0.1/12"
    podSubnet: "192.168.0.0/16"
```
I then apply the latest calico manifest from: https://docs.projectcalico.org/latest/getting-started/kubernetes/installation/calico

This results in a crashlooping calico-node pod on each host with the following presented in the log:
```
2019-09-30 18:38:28.452 [FATAL][42] int_dataplane.go 1037: Kernel's RPF check is set to 'loose'.  This would allow endpoints to spoof their IP address.  Calico requires net.ipv4.conf.all.rp_filter to be set to 0 or 1. If you require loose RPF and you are not concerned about spoofing, this check can be disabled by setting the IgnoreLooseRPF configuration parameter to 'true'.
```
This can be worked around by running the following:
```
kind get nodes --name=kind   | xargs -n1 -I {} docker exec {} sysctl -w net.ipv4.conf.all.rp_filter=0
``` 
adjust the --name argument to the name of your cluster or leave it off for the "default" kind cluster.

I then looked into when this value was being set. 

In the standard bring up this is the configured value:
```
docker exec -ti kind-control-plane  sysctl -a | grep all.rp_filter
net.ipv4.conf.all.rp_filter = 2
```
which it appears is being set by 
```
/etc/sysctl.d/10-network-security.conf:net.ipv4.conf.default.rp_filter=2
/etc/sysctl.d/10-network-security.conf:net.ipv4.conf.all.rp_filter=2
```
This was in turn changed to 2 with this issue: 
https://bugs.launchpad.net/ubuntu/+source/procps/+bug/1814262

in my other findings I found that:
the base image that we use ubuntu:19.04 has it set:
```
11:27 $ docker run -it ubuntu:19.04 sysctl -a | grep all.rp_filter
net.ipv4.conf.all.rp_filter = 1
```
and the base-image freshly built:
```
11:32 $ docker run -it  --tmpfs /tmp --tmpfs /run  --privileged --entrypoint /bin/bash mauilion/base 
root@914dc973cf59:/# sysctl -a | grep rp_filter
net.ipv4.conf.all.rp_filter = 1
root@914dc973cf59:/# exit
```
altho! the security file is present at this time!. 

```
root@2f587629b72c:/# cat /etc/sysctl.d/10-network-security.conf 

# Turn on Source Address Verification in all interfaces to
# prevent some spoofing attacks.
net.ipv4.conf.default.rp_filter=2
net.ipv4.conf.all.rp_filter=2
```
I think what's happening is that the sysctl is being honored when we start up the "real" node-image and that is what's causing the problem for calico.

that rp_filter would be set to 0 or 1 as it is set by default in 19.03

This is true of most of the recent base images.

In my opinion it's safe to set all.rp_filter to a value of 1 explicitly. 

**Environment:**

- kind version: (use `kind version`): 0.5.1

mauilion:
This fixes #891 and changes the default rpfilter to 1 from 2 enabling folks to install calico without requiring any change to the node.

### DIFF
--- a/images/base/10-network-security.conf
+++ b/images/base/10-network-security.conf
@@ -1,0 +1,4 @@
+# Turn on Source Address Verification in all interfaces to
+# prevent some spoofing attacks.
+net.ipv4.conf.default.rp_filter=1
+net.ipv4.conf.all.rp_filter=1
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -76,9 +76,11 @@ RUN clean-install \
     && chmod +x /usr/local/bin/ctr \
     && echo "done installing packages"
 
+# override the rp_filter settings to enable calico cni to "just work"
+COPY 10-network-security.conf /etc/sysctl.d/
+
 # add overrides to containerd
 COPY 10-limits.conf 10-restart.conf /etc/systemd/system/containerd.service.d/
-
 # debug containerd version and create default config
 # additionally:
 # - disable some plugins we don't use / support
--- a/pkg/build/node/node.go
+++ b/pkg/build/node/node.go
@@ -44,7 +44,7 @@ import (
 const DefaultImage = "kindest/node:latest"
 
 // DefaultBaseImage is the default base image used
-const DefaultBaseImage = "kindest/base:v20190930-9e021be@sha256:80af283eb8c335a8a3a2bc65d12befc4f97e0aed48ea3847c320d51ffac20bf2"
+const DefaultBaseImage = "kindest/base:v20191001-fed614d9@sha256:43273bb2d3d8f1fd680879e216a176900073b2bbfd0297ec84048685e0e3fa00"
 
 // DefaultMode is the default kubernetes build mode for the built image
 // see pkg/build/kube.Bits


BenTheElder:
prow bans commits with "fixes" in them because github's lovely feature that causes forks / other repos to close issues mentioned by commits in them

BenTheElder:
it's fine to put this comment in the PR body, but it can't be in the commit

BenTheElder:
otherwise though, looks good to me, thanks!
/lgtm

images/base/10-network-security.conf
```diff
@@ -0,0 +1,4 @@
+# Turn on Source Address Verification in all interfaces to
+# prevent some spoofing attacks.
+net.ipv4.conf.default.rp_filter=1
+net.ipv4.conf.all.rp_filter=1
```
BenTheElder: noting that this switches us from 2 = loose to 1 = strict

BenTheElder:
thanks! will update the base image in a follow up
/lgtm
/approve

images/base/10-network-security.conf
mauilion: yeah the other option is no validation and since we really only have the one interface for docker containers 1 is reasonable. :)

BenTheElder:
/hold
pushing the base image

PR #902, opened 2019-10-01
preserve permissions on upload

BenTheElder:
this will make it easier to script using these (the permissions can be preserved on download)


BenTheElder:
https://cloud.google.com/storage/docs/gsutil/commands/cp#options see `-P`

PR #905, opened 2019-10-02
Update example to use 0.0.0.0 instead of 127.0.0.1

mauilion:
minor fix in example a number of folks have tried to reuse this example
and it leads to unexpected behavior for folks who don't understand what
binding to localhost does.


BenTheElder:
/lgtm
/approve
Thanks!

PR #910, opened 2019-10-02
enables relative path for extraMounts

mauilion:
This enables a user to define a relative path for volumes and files that
are specified in extraMounts.

with this change a configuration like the following will determine the
path to adv-audit.yaml as relative to where `kind` is run. If you
specify a fully qualified path that will also be honored.

This is implemented using: https://golang.org/pkg/path/filepath/#Abs

```
kind: Cluster
networking:
nodes:
- role: control-plane
  extraMounts:
  - containerPath: /etc/kubernetes/policies/adv-audit.yaml
    hostPath: adv-audit.yaml
    readOnly: true
    type: File
- role: worker
```

Use Case:

When you have files that are checked in as part of a github repository
it's helpful to use relative paths so that the kind config will work
regardless of where the files are checked out :)

### DIFF
--- a/pkg/internal/apis/config/validate.go
+++ b/pkg/internal/apis/config/validate.go
@@ -18,6 +18,8 @@ package config
 
 import (
 	"net"
+	"os"
+	"path/filepath"
 
 	"sigs.k8s.io/kind/pkg/errors"
 )
@@ -101,10 +103,18 @@ func (n *Node) Validate() error {
 		}
 	}
 
+	// validate HostPath
+	for i, mapping := range n.ExtraMounts {
+		absHostPath, err := validateHostPath(mapping.HostPath)
+		if err != nil {
+			errs = append(errs, errors.Wrap(err, "Invalid hostPath"))
+		}
+		n.ExtraMounts[i].HostPath = absHostPath
+	}
+
 	if len(errs) > 0 {
 		return errors.NewAggregate(errs)
 	}
-
 	return nil
 }
 
@@ -113,4 +123,15 @@ func validatePort(port int32) error {
 		return errors.Errorf("invalid port number: %d", port)
 	}
 	return nil
+}
+
+func validateHostPath(HostPath string) (string, error) {
+	absHostPath, err := filepath.Abs(HostPath)
+	if err != nil {
+		return "", errors.Errorf("unable to determine AbsolutePath for hostPath:", HostPath)
+	}
+	if _, err := os.Stat(absHostPath); os.IsNotExist(err) {
+		return "", errors.Errorf("hostPath doesn't exist!: %s", (absHostPath))
+	}
+	return absHostPath, nil
 }
--- a/site/content/docs/user/quick-start.md
+++ b/site/content/docs/user/quick-start.md
@@ -287,6 +287,44 @@ This can be useful if using `NodePort` services or daemonsets exposing host port
 
 Note: binding the `listenAddress` to `127.0.0.1` may affect your ability to access the service.
 
+#### Mapping extra files or volumes into a node
+You can map extraMounts into a node. This is useful for propogating config files for features like the apiserver audit policy.
+```yaml
+kind: Cluster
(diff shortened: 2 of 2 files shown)


mauilion:
/assign @BenTheElder
I can squash these if you want but typically I go with 1 commit per "thing" 
In this case 1 for functionality change and another to add some docs. 
Thanks!

pkg/internal/cluster/providers/docker/provision.go
```diff
@@ -242,7 +244,11 @@ func getSubnets(networkName string) ([]string, error) {
 func generateMountBindings(mounts ...cri.Mount) []string {
 	args := make([]string, 0, len(mounts))
 	for _, m := range mounts {
-		bind := fmt.Sprintf("%s:%s", m.HostPath, m.ContainerPath)
+		absHostPath, err := filepath.Abs(m.HostPath)
+		if err != nil {
+			globals.GetLogger().Errorf("Couldn't determine absolute path for extraMounts", err)
```
BenTheElder: let's do this at a layer above and return the error, unless this is expected.

mauilion: it's not expected. In this case we proceed merrily to create a directory anyway. I can change it propogate the error up I guess.

BenTheElder: er rephrasing: this is a common thing we need to resolve anyhow, this should be done outside of the provider.

mauilion: Are you referring to mutating hostpath? Or error handling?

mauilion: Are you referring to mutating hostpath? Or error handling?

pkg/internal/apis/config/validate.go
```diff
@@ -101,10 +103,18 @@ func (n *Node) Validate() error {
 		}
 	}
 
+	// validate HostPath
```
BenTheElder: I would not expect `Validate()` to mutate any fields or depend on the host.

can we take this same logic and move it to some other method?

pkg/internal/cluster/providers/docker/provision.go
mauilion: Alright I moved this to validate.go as that seems the easiest. It will now error out if you specify a hostPath that doesn't exist :+1:

pkg/internal/apis/config/validate.go
mauilion: maybe default or check? what would you like to see here?

BenTheElder: i'm not sure what to name it offhand but, something that implies we're modifying it

BenTheElder: doing it inline in cluster.Create() would also probably be fine since it's relatively simple.

mauilion: ok I will look at this later in the week. Thanks for the feedback!

aojea:
one question, just curiosity, can we embed environment variables in the config?
This can solve this problem ala docker is doing $PWD/relative-path/

mauilion:
@aojea looks like nope I haven't tried relaxing the validation but not sure where that would apply anyway. I think the right thing is to determine the abs path based on where the kind command is run


```
17:39 $ kind create cluster --config config 
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.15.3) 🖼
ERRO[17:40:06] docker: Error response from daemon: create ${PWD}/adv-audit.yaml: "${PWD}/adv-audit.yaml" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path. 
ERRO[17:40:06] See 'docker run --help'.                     
 ✗ Preparing nodes 📦📦
ERRO[17:40:06] docker run error: exit status 125            
Error: failed to create cluster: docker run error: exit status 125
```

PR #911, opened 2019-10-03
Update apimachinery and enable strict decoding

### LINKED ISSUE #907: use strict config parsing
When I pass a configuration file to kind, e.g., to define extra mounts for the containers, I would like some confirmation that it read and applied this configuration.

This is not output in the debug or trace-level logs.

**Why is this needed**:
The non-strict unmarshalling  ignores configuration malformed in a specific way (e.g. `extraMounts` field has a typo). So it would be nice to realize this without having to resort to inspecting the container after it's up.

BenTheElder:
fixes #907


BenTheElder:
see also: https://github.com/kubernetes-sigs/kind/pull/912

neolit123:
commented here: https://github.com/kubernetes-sigs/kind/pull/912#issuecomment-538682325

BenTheElder:
I think this route is unsafe, because we're stuck at client-go v11 or else we risk breaking go get. cc @munnerz

munnerz:
How come we’re locked to client-go v11? With a replace directive, when building *this* repo, we can force whatever version we want to be used.

Consumers of this package may find themselves pinned to v11, but that’s an issue they will run into regardless of our decision here - anyone depending (or indirectly depending) on client-go will ultimately need a replace directive.

It may be that it’s not acceptable to expect those users to do this - in which case, +1 to not accepting this PR. But if that truly is the case, I think we should try and expedite resolving this upstream, even if that means (1) deleting the old v11 tags or (2) publishing more major versions

🤷‍♀️

Totally see where you're coming from though - but this really really shouldn't be an issue for k8s.io/* consumers if they are depending on staging repos only. Has there been any discussion on resolving the issues with v11 existing, aside from some of the proposed larger changes to staging repos and go mod? cc @nikhita @sttts

BenTheElder:
> How come we’re locked to client-go v11? With a replace directive, when building this repo, we can force whatever version we want to be used.

consumers of this repo `go get` it

munnerz:
Right, but isn’t this repos go mod then honoured? Unless they ‘go get’ from within another module

munnerz:
(not trying to tout go’s great behaviour here - just trying to assert a shared understanding)

BenTheElder:
> Right, but isn’t this repos go mod then honoured? Unless they ‘go get’ from within another module

not afaik. see also: https://github.com/kubernetes/test-infra/issues/14712

---

I don't think supporting modules for the staging repos is going to be easily doable, because it will be a breaking change. I spoke to `@liggitt` about this more recently.

BenTheElder:
in an empty newly created tempdir:
`GO111MODULE=on go1.13 get -v k8s.io/test-infra/kubetest` fails due to ignoring the replace (And getting v11 something instead from the require)

https://github.com/kubernetes/test-infra/blob/dfc1766ef0704616730fdbf591295b957d11dca4/go.mod#L81

https://github.com/kubernetes/test-infra/blob/dfc1766ef0704616730fdbf591295b957d11dca4/go.mod#L12

I opted to go the other route for other reasons, but this problem does still apply and goes beyond kind.

`replace` is neat, but only seems to work with `go build` and `go install` (possibly others, but not any forms of `go get` as far as I can tell)

PR #912, opened 2019-10-03
roll our own strict config loading

BenTheElder:
/hold

this is an interesting possible alternative to https://github.com/kubernetes-sigs/kind/pull/911

it's less code, simpler, doesn't use `unsafe`, doesn't fail lints ....


BenTheElder:
/hold cancel

reasonably happy with this now, might experiment with other strict parsers later... (namely: https://godoc.org/gopkg.in/yaml.v3)

aojea:
What are the pros and cons of moving away from apimachinery?

BenTheElder:
> What are the pros and cons of moving away from apimachinery?

pros:
- less code
- less dependency on generated code (easier to develop, and the generators need work... especially to work well with go modules)
- apimachinery has unsafe code (conversions trying to be simple to generate and fast), we can eliminate that
- loading config files without the kubernetes api machinery is a perfectly normal task to do in any language without large dependencies and generate code ...
- we stop leaking internal types via apimachinery's runtime type system
- we're already writing most of eg the defaulting, conversions by hand anyhow. any non-trivial changes to the types requires this already.
- developers do not need to learn the apimachinery libraries / generators to work on kind / contribute to the config
- we're able to do things like stricter parsing with tight control over it and without pulling in super new kubernetes dependencies. go modules users are somewhat stuck on client-go v11 / kubernetes 1.14...

cons:
- experienced kubernetes developers know the api machinery
- you need to be thorough with eg writing conversions. this is mitigated by tests.
- this diff could be smaller

The api machinery makes sense for large sets of kubernetes APIs. It's overkill for loading a config file and doing some defaulting / converting between sets of types.

DeepCopy is always purely mechanical though, so it's kinda nice to keep for those, and has a very simple API.

aojea:
then, it's clear the path here :smile:

neolit123:
the picture around component-config vs usage of api-machinery is not very clear, yet component-config is still leaning towards continuing to use api-machinery for config management.

the alternatives introduces by this approach are fine and would work.

my personal vote would be to continue using api-machinery, because without it kind will deviate from the plan for the ecosystem. yet in the end of the day the maintainers know best what is best for the project.

BenTheElder:
this gives considerably better errors.
```
$ kind create cluster --config=$HOME/kind.yaml --name=aa
Creating cluster "aa" ...
ERROR: failed to create cluster: unable to decode config: yaml: unmarshal errors:
  line 12: field aa not found in type v1alpha3.Node
  line 13: field cc not found in type v1alpha3.Node
```

BenTheElder:
> my personal vote would be to continue using api-machinery, because without it kind will deviate from the plan for the ecosystem. yet in the end of the day the maintainers know best what is best for the project.

Totally understand and sympathize with that, but the ecosystem has it's own tech-debt (EG switching to yaml.v3 has  been "fun" I gather and not looking super feasible soon...) that kind does not need to be held back by, and I think many of the components have different needs. `kind` is not a component, it's a user facing CLI tool.

kind is a good place to try things and report back, it's OK to be a little different. We can always switch back at a later date if the trade-off changes.

With a little more cleanup this:
- is less code (yay!)
- less dependency on code generation (code-gen = bad contributor experience)
- gives better errors (main reason I'm interested)

BenTheElder:
/hold
forgot to finish converting everything to have `yaml` tags.

BenTheElder:
(confirming https://status.docker.com/pages/533c6539221ae15e3f000031 impact, see https://github.com/kubernetes-sigs/kind/pull/951 for temporary work around)

aojea:
how is the workflow of adding a new api field or adding a new api version?

BenTheElder:
> how is the workflow of adding a new api field or adding a new api version?

unchanged.

pkg/apis/config/v1alpha3/types.go
```diff
@@ -17,38 +17,40 @@ limitations under the License.
...
-	// TypeMeta representing the type of the object and its API schema version.
-	metav1.TypeMeta `json:",inline"`
+	TypeMeta `yaml:",inline" json:",inline"`
 
 	// Nodes contains the list of nodes defined in the `kind` Cluster
 	// If unset this will default to a single control-plane node
 	// Note that if more than one control plane is specified, an external
 	// control plane load balancer will be provisioned implicitly
-	Nodes []Node `json:"nodes,omitempty"`
+	Nodes []Node `yaml:"nodes,omitempty" json:"nodes,omitempty"`
```
neolit123: sorry for not having a look at this PR again.

i think this dropped support for user provided JSON kind config files?
which is not very critical but diverges from the "component config" / api-machinery trend.

BenTheElder: - this also has json tags? (reread the diff?)
- we never supported JSON config files. I don't intend to. yaml is a superset of json.

pkg/internal/apis/config/encoding/load.go
```diff
@@ -0,0 +1,104 @@
...
+
+	// load the raw contents
+	raw, err := readAll(path)
+	if err != nil {
+		return nil, err
+	}
+
+	// get kind & apiVersion
+	tm := typeMeta{}
+	if err := yaml.Unmarshal(raw, &tm); err != nil {
```
neolit123: hm, i've heard that v3 enables strict by default for the Unmarshal function.
https://github.com/kubernetes-sigs/yaml/pull/26#issuecomment-510687743

BenTheElder: it doesn't. we parse a config file in CI.

pkg/apis/config/v1alpha3/types.go
BenTheElder: Kubernetes needs a seperate JSON parser for other reasons (because it gives better errors?). yaml.v3 also gives better errors https://github.com/kubernetes/kubernetes/issues/78946

pkg/internal/apis/config/encoding/load.go
BenTheElder: https://github.com/go-yaml/yaml/blob/fc94e3f7165204dd261b8d39d39a1e1773b931c9/yaml.go#L89

BenTheElder: (I took a pass through the actual yaml.v3 code before switching)

pkg/apis/config/v1alpha3/types.go
neolit123: missed the extra tags in the diff.

pkg/internal/apis/config/encoding/load.go
neolit123: i see, perhaps the linked commend was misleading.

BenTheElder: also the new Node object's `Decode()` does not preserve strict being enabled :disappointed: 

AFAICT implementing the secret "obsolete" (v2 api) custom `YamlUnmarshal` method works fine though.

PR #914, opened 2019-10-03
tidy containerd config, make runtimeclass tests work

BenTheElder:
writing out the default config is pretty redundant, I think it's also interfering with the overrides we added. let's stop doing that.

also, we need to set some values for the e2e tests to make runtimeclass tests work with containerd (docker/dockershim is the assumed default)


BenTheElder:
/cc @aojea

https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/914/pull-kind-e2e-kubernetes/1179633562131894274
(this run had an extra commit to tweak SKIP temporarily to not skip these)

PR #919, opened 2019-10-04
Add tests for provider/common

aojea:
Add testing for the sigs.k8s.io/kind/pkg/internal/cluster/providers/provider/common package

```
Running tool: /usr/local/go/bin/go test -timeout 30s sigs.k8s.io/kind/pkg/internal/cluster/providers/provider/common -coverprofile=/tmp/vscode-gozIao1h/go-code-cover

ok  	sigs.k8s.io/kind/pkg/internal/cluster/providers/provider/common	0.003s	coverage: 97.1% of statements
Success: Tests passed.
```

### DIFF
--- a/pkg/internal/cluster/providers/docker/provision.go
+++ b/pkg/internal/cluster/providers/docker/provision.go
@@ -209,7 +209,8 @@ func runArgsForLoadBalancer(cfg *config.Cluster, name string, args []string) ([]
 }
 
 func getProxyEnv(cfg *config.Cluster) (map[string]string, error) {
-	envs := common.GetProxyEnvs(cfg)
+	getEnv := &common.OsEnv{}
+	envs := common.GetProxyEnvs(cfg, getEnv)
 	// Specifically add the docker network subnets to NO_PROXY if we are using a proxy
 	if len(envs) > 0 {
 		// Docker default bridge network is named "bridge" (https://docs.docker.com/network/bridge/#use-the-default-bridge-network)
--- a/pkg/internal/cluster/providers/provider/common/env.go
+++ b/pkg/internal/cluster/providers/provider/common/env.go
@@ -1,0 +1,46 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"os"
+)
+
+// Getenver is the interface we use to mock out the env for easier testing. OS env
+// vars can't be as easily tested since internally it uses sync.Once.
+type Getenver interface {
+	Getenv(string) string
+}
+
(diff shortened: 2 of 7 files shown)


pkg/internal/cluster/providers/provider/common/proxy_test.go
```diff
@@ -0,0 +1,73 @@
...
+				return &c
+			}(),
+			envs: map[string]string{"HTTP_PROXY": "5.5.5.5"},
+			want: map[string]string{"HTTP_PROXY": "5.5.5.5", "http_proxy": "5.5.5.5", "NO_PROXY": ",10.0.0.0/24,12.0.0.0/24", "no_proxy": ",10.0.0.0/24,12.0.0.0/24"},
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			os.Clearenv()
```
BenTheElder: we should avoid mutating global state, I can write a follow-up to fix this though.

aojea: I'm happy to fix it if you just guide me

BenTheElder: we should make `GetProxyEnvs()` a thin wrapper around a method with an injected dependency.

aojea: PTAL, I've used this approach
https://github.com/kubernetes/kubernetes/blob/master/cluster/images/conformance/go-runner/env.go

pkg/internal/cluster/providers/docker/provision.go
```diff
@@ -209,7 +209,8 @@ func runArgsForLoadBalancer(cfg *config.Cluster, name string, args []string) ([]
 }
 
 func getProxyEnv(cfg *config.Cluster) (map[string]string, error) {
-	envs := common.GetProxyEnvs(cfg)
+	getEnv := &common.OsEnv{}
+	envs := common.GetProxyEnvs(cfg, getEnv)
```
BenTheElder: the caller shouldn't be concerned with this.

pkg/internal/cluster/providers/provider/common/proxy.go
```diff
@@ -34,12 +33,12 @@ const (
 
 // GetProxyEnvs returns a map of proxy environment variables to their values
 // If proxy settings are set, NO_PROXY is modified to include the cluster subnets
-func GetProxyEnvs(cfg *config.Cluster) map[string]string {
+func GetProxyEnvs(cfg *config.Cluster, env Getenver) map[string]string {
```
BenTheElder: the public API doesn't need changing, just make the public a wrapper around a private method that passes in the real getenv

pkg/internal/cluster/providers/provider/common/env.go
```diff
@@ -0,0 +1,46 @@
...
+package common
+
+import (
+	"os"
+)
+
+// Getenver is the interface we use to mock out the env for easier testing. OS env
+// vars can't be as easily tested since internally it uses sync.Once.
+type Getenver interface {
+	Getenv(string) string
```
BenTheElder: this is overkill, we can just pass in a `func(string)(string)`

pkg/internal/cluster/providers/provider/common/proxy.go
BenTheElder: I implemented this in https://github.com/kubernetes-sigs/kind/pull/969

BenTheElder:
rebased commits merged in https://github.com/kubernetes-sigs/kind/pull/969
thanks again!

PR #931, opened 2019-10-07
reach parity with other kubernetes e2e testing on retrying tests

BenTheElder:
other kubernetes e2e presubmits are retrying tests... we should match them. though we are considering disabling this broadly (at which time we should also disable this).
https://github.com/kubernetes/test-infra/blob/608557bbe4342da9895a0762cbdbf8a508ee59e9/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml#L5-L6

~for now: only enabling this if we are not focused on conformance tests (which is a weak check, but sufficient to detect newer jobs that are running specific tests beyond conformance AKA `pull-kubernetes-e2e-kind`).~

EDIT: enabling it if not otherwise specified instead.

FYI @liggitt

### DIFF
--- a/hack/ci/e2e-k8s.sh
+++ b/hack/ci/e2e-k8s.sh
@@ -140,11 +140,15 @@ run_tests() {
   fi
 
   # setting this env prevents ginkgo e2e from trying to run provider setup
-  export KUBERNETES_CONFORMANCE_TEST="y"
+  export KUBERNETES_CONFORMANCE_TEST='y'
   # setting these is required to make RuntimeClass tests work ...
   export KUBE_CONTAINER_RUNTIME=remote
   export KUBE_CONTAINER_RUNTIME_ENDPOINT=unix:///run/containerd/containerd.sock
   export KUBE_CONTAINER_RUNTIME_NAME=containerd
+  # unless we specify otherwise, tolerate flakes :/
+  # TODO: remove this when kubernetes e2e testing no longer tolerates flakes ...
+  # https://github.com/kubernetes/test-infra/blob/608557bbe4342da9895a0762cbdbf8a508ee59e9/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml#L5-L6
+  export GINKGO_TOLERATE_FLAKES="${GINKGO_TOLERATE_FLAKES:-y}"
   ./hack/ginkgo-e2e.sh \
     '--provider=skeleton' "--num-nodes=${NUM_NODES}" \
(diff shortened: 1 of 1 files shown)


hack/ci/e2e-k8s.sh
```diff
@@ -140,11 +140,17 @@ run_tests() {
...
+  export KUBERNETES_CONFORMANCE_TEST='y'
   # setting these is required to make RuntimeClass tests work ...
   export KUBE_CONTAINER_RUNTIME=remote
   export KUBE_CONTAINER_RUNTIME_ENDPOINT=unix:///run/containerd/containerd.sock
   export KUBE_CONTAINER_RUNTIME_NAME=containerd
+  # unless we are conformance testing, tolerate flakes :/
+  # TODO: remove this when kubernetes e2e testing no longer tolerates flakes ...
+  # https://github.com/kubernetes/test-infra/blob/608557bbe4342da9895a0762cbdbf8a508ee59e9/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml#L5-L6
+  if [ "${FOCUS}" -ne '\[Conformance\]' ]; then
+    export GINKGO_TOLERATE_FLAKES='y'
```
liggitt: don't override if already set?

liggitt: (I'm not sure how this is called, so that might not make sense)

BenTheElder: that SGTM, let's simplify it to that and then I can update a few jobs to explictly say n

BenTheElder: it's called here https://github.com/kubernetes/test-infra/blob/53631782335c7d322c4c4626460c20dade44a3de/config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml#L243-L253

(kind repo postsubmit uploads a copy of the now tested and merged latest kind binary & e2e script)

in the kind repo we call it with a layer above that builds a shiny new kind binary first and uses local copies of the script instead.

BenTheElder: (and yes I'm going to move to a wg-k8s-infra bucket O(soon))

BenTheElder: updated.

justaugustus:
@liggitt @BenTheElder -- can we open up an issue in... some repo to track?
I fear this will get lost if we don't.

BenTheElder:
well there was https://github.com/kubernetes/test-infra/issues/2372, I'm pretty sure @dims also had an issue tracking this already

BenTheElder:
https://github.com/kubernetes/kubernetes/issues/68091

PR #932, opened 2019-10-08
use containerd V2 runtime

BenTheElder:
- use an actual file for containerd config to make it easier to develop
- use the v2 runtime xref: https://github.com/containerd/cri/issues/1075


BenTheElder:
this puts us in line with other kubernetes containerd based CI and nets us less memory / process usage (one v2 shim process per pod, instead of per container)

aojea:
Yeah, looking forward to seeing the performance improvement
Better files that bash chaining command +1
/Lgtm

PR #939, opened 2019-10-09
RFC - Attaching cluster nodes to complex upstream networks

neiljerram:
Hi - this is a collection of hackery that I've developed in order to model a Kubernetes cluster that is split across racks, and with redundant planes of connectivity between all the nodes.

- Splitting across racks means we need to be able to specify a Docker network to attach each node to, instead of assuming the default 'bridge'.

- Redundant planes of connectivity means:
  - We want to be able to attach each node to multiple specified Docker networks.
  - We want to configure a loopback address for each node (i.e. an address that is not tied to a particular network) and arrange for those addresses to be used as source and destination IPs by Kubernetes, instead of interface-specific addresses.
  - As part of the previous point, we need to add routing for the loopback addresses, after the node containers have been created and before kubeadm starts running.

Here's an example config with the new fields I've added support for:
```
kind: Cluster
networking:
  disableDefaultCNI: true
nodes:
- role: control-plane
  networks: [ra1, ra2]
  loopback: 172.31.10.3
  routes:
  - "172.31.10.0/24 src 172.31.10.3 nexthop dev eth0 nexthop dev eth1"
  - "172.31.20.0/24 src 172.31.10.3 nexthop via 172.31.11.1 nexthop via 172.31.12.1"
- role: worker
  networks: [ra1, ra2]
  loopback: 172.31.10.4
  routes:
  - "172.31.10.0/24 src 172.31.10.4 nexthop dev eth0 nexthop dev eth1"
  - "172.31.20.0/24 src 172.31.10.4 nexthop via 172.31.11.1 nexthop via 172.31.12.1"
- role: worker
  networks: [rb1, rb2]
  loopback: 172.31.20.3
  routes:
  - "172.31.20.0/24 src 172.31.20.3 nexthop dev eth0 nexthop dev eth1"
  - "172.31.10.0/24 src 172.31.20.3 nexthop via 172.31.21.1 nexthop via 172.31.22.1"
- role: worker
  networks: [rb1, rb2]
  loopback: 172.31.20.4
  routes:
  - "172.31.20.0/24 src 172.31.20.4 nexthop dev eth0 nexthop dev eth1"
  - "172.31.10.0/24 src 172.31.20.4 nexthop via 172.31.21.1 nexthop via 172.31.22.1"
```

I'm sure I haven't done this all right first time, but I guess that enhancements like these will be of wider interest than just for me.  Hence I'm posting this as an RFC to ask if it would make sense to upstream this kind of support, and if so to get your advice on how to put into a potentially mergeable form.

### DIFF
--- a/pkg/apis/config/v1alpha3/types.go
+++ b/pkg/apis/config/v1alpha3/types.go
@@ -81,6 +81,8 @@ type Node struct {
 
 	// Loopback IP.
 	Loopback string `json:"loopback,omitempty"`
+
+	Routes []string `json:"routes,omitempty"`
 }
 
 // NodeRole defines possible role for nodes in a Kubernetes cluster managed by `kind`
--- a/pkg/apis/config/v1alpha3/types.go
+++ b/pkg/apis/config/v1alpha3/types.go
@@ -75,6 +75,9 @@ type Node struct {
 	// ExtraPortMappings describes additional port mappings for the node container
 	// binded to a host Port
 	ExtraPortMappings []cri.PortMapping `json:"extraPortMappings,omitempty"`
+
+	// Docker networks to attach to.  Defaults to the Docker default ("bridge").
+	Networks []string `json:"networks,omitempty"`
 }
 
 // NodeRole defines possible role for nodes in a Kubernetes cluster managed by `kind`
--- a/pkg/apis/config/v1alpha3/types.go
+++ b/pkg/apis/config/v1alpha3/types.go
@@ -78,6 +78,9 @@ type Node struct {
 
 	// Docker networks to attach to.  Defaults to the Docker default ("bridge").
 	Networks []string `json:"networks,omitempty"`
+
+	// Loopback IP.
+	Loopback string `json:"loopback,omitempty"`
 }
 
 // NodeRole defines possible role for nodes in a Kubernetes cluster managed by `kind`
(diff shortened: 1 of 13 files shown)


BenTheElder:
> RFC - Attaching cluster nodes to complex upstream networks #939

usually an RFC comes before sending an implementation 🙃

/hold


- loopback is generally not a configurable IP?
- writing routes inline seems brittle and very specific to your use case
- configurable docker network does not work safely yet because the default network has different behavior, please see the issues in the issue tracker regarding this

aojea:
hi @neiljerram , I think that all this configuration can be done with a script once kind finish the deployment, configuring the routes and loopback interfaces on nodes. 
Also, it's adding too many dependencies on docker network that will be bad if we want to support more providers, 
I have several examples of creating complex topologies with containers in this repo https://github.com/aojea/frr-lab , using tools like [koko](https://github.com/redhat-nfvpe/koko), you can use something similar to script your complex topologies and kind to deploy the different clusters

BenTheElder:
I think https://github.com/kubernetes-sigs/kind/pull/939#issuecomment-540186911 is the advice to go with here.

Thanks for the PR, but let's discuss this in an issue first if you want to follow up. I think the end goal can be done without these changes and the complexity will not help the existing use cases.

neiljerram:
Thanks @BenTheElder .  I'm not sure you are right about the end goal being achievable without any KIND changes, but I'm happy for this PR to be closed for now, and will come back to this when I have time, hopefully soon.

PR #940, opened 2019-10-09
use normal containerd version strings, update containerd with latest patches

BenTheElder:
containerd had a minor patch to release 1.3 since our last update.

however, I'm most interested in managing this using normal upstream version strings that we can compare to upstream.

basically these are `git describe`. we patched this in the kind-ci/containerd-nightly CI infrastructure.

i've bumped the base / node image to pick this up.

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -74,18 +74,15 @@ RUN clean-install \
 COPY 10-network-security.conf /etc/sysctl.d/
 
 # Install containerd and runc binaries from kind-ci/containerd-nightlies repository
-# The repository contains latest stable releases and nightlies using golang's pseudo-version
-# CONTAINERD_VERSION=0.0.0-yyyymmddhhmmss-commitid
-ARG CONTAINERD_VERSION="0.0.0-20191007143651-efd38f48"
-ARG CONTAINERD_BASE_URL="https://github.com/kind-ci/containerd-nightlies/releases/download/"
+# The repository contains latest stable releases and nightlies built for multiple architectures
+ARG CONTAINERD_VERSION="v1.3.0-7-g0b43a311"
+ARG CONTAINERD_BASE_URL="https://github.com/kind-ci/containerd-nightlies/releases/download"
 RUN export ARCH=$(dpkg --print-architecture | sed 's/ppc64el/ppc64le/' | sed 's/armhf/arm/') \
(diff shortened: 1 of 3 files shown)


images/base/Dockerfile
```diff
@@ -74,18 +74,15 @@ RUN clean-install \
 COPY 10-network-security.conf /etc/sysctl.d/
 
 # Install containerd and runc binaries from kind-ci/containerd-nightlies repository
-# The repository contains latest stable releases and nightlies using golang's pseudo-version
-# CONTAINERD_VERSION=0.0.0-yyyymmddhhmmss-commitid
-ARG CONTAINERD_VERSION="0.0.0-20191007143651-efd38f48"
-ARG CONTAINERD_BASE_URL="https://github.com/kind-ci/containerd-nightlies/releases/download/"
+# The repository contains latest stable releases and nightlies built for multiple architectures
+ARG CONTAINERD_VERSION="v1.3.0-7-g0b43a311"
```
neolit123: should the `v` be part of the version given it's replaced in all references bellow?

BenTheElder: `v` is part of the version, however for :reasons: it's not in the download URL currently :roll_eyes:

BenTheElder: this is also true of the official releases

neolit123: i see, so at some point it might become part of the URLs.

BenTheElder: i'm not really sure what makes more sense. multiple formats are used, but this is the format used for VERSION in the build tooling, in the CLI, and in the tags. if they become a go module then the v will become more present.

https://github.com/containerd/containerd/releases/download/v1.3.0/containerd-1.3.0.linux-amd64.tar.gz

BenTheElder: versus ours:
https://github.com/kind-ci/containerd-nightlies/releases/download/containerd-1.3.0-7-g0b43a311/containerd-1.3.0-7-g0b43a311.linux-amd64.tar.gz

(which are still a work in progress.. we're actively improving that infra eg https://github.com/kind-ci/containerd-nightlies/pull/8)

neolit123: i think i prefer to have the `v` everywhere. :thinking:

BenTheElder: +1 cc @aojea

BenTheElder:
:/ ipv6 is definitely not as stable.
/retest

aojea:
I see the job failing lately, is because of the host or because of kubernetes?

PR #943, opened 2019-10-12
don't interrupt kind export logs on failures

### LINKED ISSUE #942: kind export logs sometimes doesn't export logs on all nodes on failure
**What happened**:

When the tar command fails to export the logs in one node it seems that doesn't try to continue exporting the logs of the other nodes

Obtain the logs on all nodes

There is several occurrences in the CI 

https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-kind-conformance-parallel-ipv6/1182869489591521285

**Environment:**

aojea:
If the tar command detects that some files change after archiving them, it exit with status 1. The problem is that we are running the tar command at the beginning of the export process and this interrupts the whole log exporting process.

Per Ben suggestion https://github.com/kubernetes-sigs/kind/pull/943#issuecomment-541312609 we store the errors and only report them once we finish collecting the logs 

Fixes #942

### DIFF
--- a/pkg/internal/cluster/logs/logs.go
+++ b/pkg/internal/cluster/logs/logs.go
@@ -65,19 +65,23 @@ func Collect(nodes []nodes.Node, dir string) error {
 		),
 	}
 	// add a log collection method for each node
+	// collect up and return errors
+	errs := []error{}
 	for _, n := range nodes {
 		node := n // https://golang.org/doc/faq#closures_and_goroutines
 		name := node.String()
 		// grab all logs under /var/log (pods and containers)
+		// TODO: don't fail if exit status is 1. It only means that some files were changed while being archived
+		// http://man7.org/linux/man-pages/man1/tar.1.html
 		cmd := node.Command("tar", "--hard-dereference", "-C", "/var/log", "-chf", "-", ".")
 
 		if err := exec.RunWithStdoutReader(cmd, func(outReader io.Reader) error {
 			if err := untar(outReader, filepath.Join(dir, name)); err != nil {
-				return errors.Wrapf(err, "Untarring %q: %v", name, err)
+				errs = append(errs, errors.Wrapf(err, "Untarring %q: %v", name, err))
 			}
 			return nil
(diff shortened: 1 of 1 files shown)


BenTheElder:
/priority important-soon
Logs are best effort and this does not functionally break CI

BenTheElder:
We should probably continue on error instead (but record it).
Doing this concurrently was deemed a flaky overload previously IIRC. The tar moves the most data and frequently fails at the moment.

aojea:
> We should probably continue on error instead (but record it).
> Doing this concurrently was deemed a flaky overload previously IIRC. The tar moves the most data and frequently fails at the moment.

 The tar error is because some files change and tar exit with status 1 , I think that the best thing is to not fail on that error

BenTheElder:
> The tar error is because some files change and tar exit with status 1 , I think that the best thing is to not fail on that error

this seems like the more important issue, I spent some time looking into it and wrote https://github.com/kubernetes-sigs/kind/pull/946 while doing so

pkg/internal/cluster/logs/logs.go
```diff
@@ -65,19 +65,23 @@ func Collect(nodes []nodes.Node, dir string) error {
...
 		name := node.String()
 		// grab all logs under /var/log (pods and containers)
+		// TODO: don't fail if exit status is 1. It only means that some files were changed while being archived
+		// http://man7.org/linux/man-pages/man1/tar.1.html
 		cmd := node.Command("tar", "--hard-dereference", "-C", "/var/log", "-chf", "-", ".")
 
 		if err := exec.RunWithStdoutReader(cmd, func(outReader io.Reader) error {
 			if err := untar(outReader, filepath.Join(dir, name)); err != nil {
-				return errors.Wrapf(err, "Untarring %q: %v", name, err)
+				errs = append(errs, errors.Wrapf(err, "Untarring %q: %v", name, err))
```
BenTheElder: this is wrong and unsafe. take a look at `RunWithStdoutReader`

PR #945, opened 2019-10-12
[WIP][DO NOT MERGE] Profile CI execution time

aojea:
We can time each phase of the CI just looking at the logs, but this simplifies the process since we don't have to substract timestamps

### DIFF
--- a/hack/ci/e2e-k8s.sh
+++ b/hack/ci/e2e-k8s.sh
@@ -169,10 +169,11 @@ main() {
   BUILD_TYPE="${BUILD_TYPE:-bazel}"
 
   # build kubernetes
+  echo "INFO: Building kubernetes"
   if [ "${BUILD_TYPE:-}" = "bazel" ]; then
-    build_with_bazel
+    time build_with_bazel
   else
-    build
+    time build
   fi
 
   # in CI attempt to release some memory after building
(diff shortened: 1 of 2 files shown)


aojea:
we can decide to merge or not, I just used for convenience to avoid dealing with the timestamps 😅

hack/ci/e2e-k8s.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
```
BenTheElder: This was intentional. Please see the commit.

aojea: > we can decide to merge or not, I just used for convenience to avoid dealing with the timestamps 😅
https://github.com/kubernetes-sigs/kind/pull/945#issuecomment-541334267
:)

BenTheElder: You should add a [WIP] to the title to block merge

aojea: yeah,my bad, sorry

aojea:
|   |  1.12 |1.14  |  1.16 |  master |
|---|---|---|---|---|
| Install kind  | 1m2.721s  |   |   |   |
|  Build Kubernetes | 4m30.060s  |   |   |   |
|  Create Cluster |1m33.768s|   |   |   |
|  Run e2e | 7m57.611s  |   |   |   |

BenTheElder:
> ... where should we focus? ...

I'm not worried about the current runtime --I've already profiled it pretty extensively.

You are unlikely to get these noticeably faster through changes to kind. 

> ... build kubernetes? using tarballs can help? ...

Kubernetes presubmit cannot use tarballs anyhow, and kind repo presubmit needs to mirror Kubernetes presubmit so we do not break it.

The build is already heavily optimized and easily competitive or better than other CI jobs.

BenTheElder:
(also not that in kubernetes presubmit, kind is installed from a prebuilt binary, which we obviously cannot do here)

aojea:
/close

cool

PR #946, opened 2019-10-12
aggregate log export errors, ignore changed files

BenTheElder:
fixes https://github.com/kubernetes-sigs/kind/issues/942

/cc @aojea 
I spent some time digging into this, and I think this is a more complete fix.

### DIFF
--- a/pkg/internal/cluster/logs/logs.go
+++ b/pkg/internal/cluster/logs/logs.go
@@ -69,7 +69,15 @@ func Collect(nodes []nodes.Node, dir string) error {
 		node := n // https://golang.org/doc/faq#closures_and_goroutines
 		name := node.String()
 		// grab all logs under /var/log (pods and containers)
-		cmd := node.Command("tar", "--hard-dereference", "-C", "/var/log", "-chf", "-", ".")
+		cmd := node.Command(
+			"sh", "-c",
+			// Tar will exit 1 if a file changed during the archival.
+			// We don't care about this, so we're invoking it in a shell
+			// And masking out 1 as a return value.
+			// Fatal errors will return exit code 2.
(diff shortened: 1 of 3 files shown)


BenTheElder:
/hold 
need to fix up some more errors handling... in other parts of the code

pkg/internal/cluster/logs/logs.go
```diff
@@ -64,20 +64,30 @@ func Collect(nodes []nodes.Node, dir string) error {
...
 		// grab all logs under /var/log (pods and containers)
-		cmd := node.Command("tar", "--hard-dereference", "-C", "/var/log", "-chf", "-", ".")
+		cmd := node.Command(
+			"sh", "-c",
+			// Tar will exit 1 if a file changed during the archival.
+			// We don't care about this, so we're invoking it in a shell
+			// And masking out 1 as a return value.
+			// Fatal errors will return exit code 2.
+			// http://man7.org/linux/man-pages/man1/tar.1.html#RETURN_VALUE
+			`tar --hard-dereference -C /var/log -chf - . || (r=$?; [ $r -eq 1 ] || exit $r)`,
```
BenTheElder: I would really prefer to not invoke shell, but this is not doing much and there actually doesn't seem to be another good solution.

there is no flag that disables this behavior, and we want to preserve other exit codes:
- from tar that are not equal to 1
- any exit code from `docker exec ...`

pkg/internal/cluster/logs/logs.go
```diff
@@ -113,7 +123,8 @@ func Collect(nodes []nodes.Node, dir string) error {
 	}
 
 	// run and collect up all errors
-	return concurrent.Coalesce(fns...)
```
aojea: ?

aojea:
/lgtm

Thanks

Issue #944, opened 2019-10-12
kindnet panics sometimes on ip6tables-restore and iptables-restore


aojea:
**What happened**:

> 2019-10-12T07:48:20.101394314Z stderr F panic: exit status 1 (ip6tables-restore: line 5 failed
> 2019-10-12T07:48:20.101469169Z stderr F )
> 2019-10-12T07:48:20.101476874Z stderr F 
> 2019-10-12T07:48:20.101483906Z stderr F goroutine 22 [running]:
> 2019-10-12T07:48:20.101491264Z stderr F main.main.func1(0xc000243400)
> 2019-10-12T07:48:20.101498175Z stderr F 	/src/main.go:76 +0x67
> 2019-10-12T07:48:20.101504071Z stderr F created by main.main
> 2019-10-12T07:48:20.101509918Z stderr F 	/src/main.go:73 +0x318
> 

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_kind/943/pull-kind-conformance-parallel-ipv6/1182919191846457348/artifacts/logs/kind-worker2/pods/kube-system_kindnet-g248m_e664ffe1-a561-4897-8952-487625db3a1d/kindnet-cni/2.log

**Anything else we need to know?**:

Ben mentioned several times that ipv6 tests are flakier than the ipv4 ones. This can be some of the reasons

**Environment:**

- kind version: (use `kind version`):
- Kubernetes version: (use `kubectl version`):
- Docker version: (use `docker info`):
- OS (e.g. from `/etc/os-release`):

aojea:
The lines are exactly the same as the ones created 1 minute before, however, this time fail to commit them

```
2019-10-12T07:48:19.495239187Z stdout F Handling iptables: *nat
2019-10-12T07:47:17.992170687Z stdout F Handling iptables: *nat

2019-10-12T07:48:19.495245083Z stdout F :KIND-MASQ-AGENT - [0:0]
2019-10-12T07:47:17.992178034Z stdout F :KIND-MASQ-AGENT - [0:0]

2019-10-12T07:48:19.495251727Z stdout F -A KIND-MASQ-AGENT -m comment --comment "kind-masq-agent: local traffic is not subject to MASQUERADE" -d fd00:10:244::/64 -j RETURN
[...]

PR #947, opened 2019-10-12
Fix Kindnet ip-masq-agent panics

aojea:
The IP masq agent avoids masquerading the pod to pod traffic, inserting some iptables rules in the POSTROUTING chain.

It turns out that sometimes it fails to write the iptables rules and panics (it uses iptables-restore and fails trying to commit), causing some flakiness in the CI due to the time needed for kindnet to recover (the pod has to be restarted)

Since the agent is running periodically and the rules only need to be inserted once,  we shouldn't panic and just log the error and try later. If there is a problem with iptables, the cluster won't work correctly because kubelet and kube-proxy will fail.

This also refactors the masq-agent using github.com/coreos/go-iptables/iptables, removing the dependency on k8s/pkg/utils/iptables, that is much simpler and doesn't use `iptables-restore`, that makes it less exposed to problems with the iptables locking mechanism

Fixes https://github.com/kubernetes-sigs/kind/issues/944


images/kindnetd/cmd/kindnetd/main.go
```diff
@@ -70,12 +70,8 @@ func main() {
 	// enforce ip masquerade rules
 	// TODO: dual stack...?
 	masqAgent := NewIPMasqAgent(net.IsIPv6String(hostIP), []string{os.Getenv("POD_SUBNET")})
-	go func() {
-		// TODO: use logging and continue retrying instead...
-		if err := masqAgent.SyncRulesForever(time.Second * 60); err != nil {
-			panic(err)
-		}
-	}()
+	// TODO: handle from the main control loop once this moves to a proper controller
```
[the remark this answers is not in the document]
BenTheElder: controllers. don't have a loop (at least, not an explicit one you write)

images/kindnetd/cmd/kindnetd/main.go
```diff
@@ -70,12 +70,8 @@ func main() {
...
+	go masqAgent.SyncRulesForever(time.Second * 60)
```
BenTheElder: this is _technically_ fine but is better to still be wrapped in a `func() { }()`.

the reason for this is that arguments are evaluated immediately, which can be a major footgun in the future, it's probably best to go ahead and enclose this in it's own scope now.

images/kindnetd/cmd/kindnetd/main.go
```diff
@@ -70,12 +70,8 @@ func main() {
 	// enforce ip masquerade rules
 	// TODO: dual stack...?
 	masqAgent := NewIPMasqAgent(net.IsIPv6String(hostIP), []string{os.Getenv("POD_SUBNET")})
-	go func() {
-		// TODO: use logging and continue retrying instead...
```
BenTheElder: using actual logging still applies.

images/kindnetd/cmd/kindnetd/masq.go
```diff
@@ -56,10 +56,10 @@ type IPMasqAgent struct {
 }
 
 // SyncRulesForever syncs ip masquerade rules forever
-func (ma *IPMasqAgent) SyncRulesForever(interval time.Duration) error {
+func (ma *IPMasqAgent) SyncRulesForever(interval time.Duration) {
 	for {
 		if err := ma.SyncRules(); err != nil {
-			return err
+			fmt.Printf("kind-masq-agent: error syncing rules %v chain, retrying ...", err)
```
BenTheElder: this should write a newline

images/kindnetd/cmd/kindnetd/main.go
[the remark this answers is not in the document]
aojea: I need to create one to fully understand the concepts :/

images/kindnetd/go-mod-k8s.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
```
aojea: This script comes from https://github.com/kubernetes/kubernetes/issues/79384#issuecomment-521493597

images/kindnetd/go.mod
```diff
@@ -3,27 +3,64 @@ module sigs.k8s.io/kind/images/kindnetd
...
-	gopkg.in/yaml.v2 v2.2.2 // indirect
-	k8s.io/api v0.0.0-20190409021203-6e4e0e4f393b
-	k8s.io/apimachinery v0.0.0-20190404173353-6a84e37a896d
-	k8s.io/client-go v11.0.0+incompatible
-	k8s.io/kubernetes v1.14.2
-	k8s.io/utils v0.0.0-20190506122338-8fab8cb257d5
-	sigs.k8s.io/yaml v1.1.0 // indirect
+	k8s.io/api v0.0.0
+	k8s.io/apimachinery v0.0.0
+	k8s.io/client-go v0.0.0
```
aojea: I really need your help here Ben, is this the best way to use kubernetes as a module?
we only need these

```
cmd/kindnetd/cni.go:    corev1 "k8s.io/api/core/v1"
cmd/kindnetd/cni.go:    "k8s.io/utils/net"
cmd/kindnetd/main.go:   corev1 "k8s.io/api/core/v1"
cmd/kindnetd/main.go:   metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
cmd/kindnetd/main.go:   "k8s.io/client-go/kubernetes"
cmd/kindnetd/main.go:   "k8s.io/client-go/rest"
cmd/kindnetd/main.go:   "k8s.io/utils/net"
cmd/kindnetd/masq.go:   utildbus "k8s.io/kubernetes/pkg/util/dbus"
cmd/kindnetd/masq.go:   utiliptables "k8s.io/kubernetes/pkg/util/iptables"
cmd/kindnetd/masq.go:   utilexec "k8s.io/utils/exec"
```

aojea:
/hold 

I'm thinking.the masq.agent can be much more simpler

images/kindnetd/go-mod-k8s.sh
BenTheElder: We already have this working, you just set the version in requirements and it will be replaced with a pseudo version

Why are we changing the version anyhow?

images/kindnetd/go.mod
BenTheElder: Not like this ...
Will follow up Monday but ... Why are we even touching this? This is totally unrelated to not panicing.

aojea: the `pkg/util/iptables` has changed in the last releases, I was especially interested in the `dbus` removal, but I just found it didn't make it to the 1.16 and is only available in 1.17

https://github.com/kubernetes/kubernetes/pull/81517

I'm experimenting with several approaches, basically trying to understand if we can simplify the ipmasq code.

images/kindnetd/go-mod-k8s.sh
aojea: https://github.com/kubernetes-sigs/kind/pull/947/commits/c34ac87fb5d6928d4a5a20e93c7f8d8e8ae84cfc#r334287177

aojea:
We only need to append one rule at the end of the POSTROUTING table, instead of the current

```
-A POSTROUTING -m comment --comment "kubernetes postrouting rules" -j KUBE-POSTROUTING
-A POSTROUTING -m comment --comment "kind-masq-agent: ensure nat POSTROUTING directs all non-LOCAL destination traffic to our custom KIND-MASQ-AGENT chain" -m addrtype ! --dst-type LOCAL -j KIND-MASQ-AGENT
-A KIND-MASQ-AGENT -d fd00:10:244::/64 -m comment --comment "kind-masq-agent: local traffic is not subject to MASQUERADE" -j RETURN
-A KIND-MASQ-AGENT -m comment --comment "ip-masq-agent: outbound traffic is subject to MASQUERADE (must be last in chain)" -j MASQUERADE
```

we can use 

```
-A POSTROUTING -m comment --comment "kubernetes postrouting rules" -j KUBE-POSTROUTING
-A KIND-MASQ-AGENT ! -d fd00:10:244::/64 -o eth0 -m comment --comment "kind-masq-agent: local traffic is not subject to MASQUERADE" -j MASQUERADE
```

images/kindnetd/go.mod
aojea: we can use go-iptables and not rely on kubernetes for this

images/kindnetd/go-mod-k8s.sh
aojea: using go-iptables we no longer need to use kubernetes as a module

aojea:
All green in the CI testing with my image aojea/kindnet:0.5.1
/assign @BenTheElder 
PTAL I think that is important to remove the dependency on the kubernetes iptables utils

images/kindnetd/cmd/kindnetd/masq.go
```diff
@@ -17,49 +17,42 @@ limitations under the License.
...
-	dbus := utildbus.New()
-	protocol := utiliptables.ProtocolIpv4
+	protocol := iptables.ProtocolIPv4
 	if ipv6 {
-		protocol = utiliptables.ProtocolIpv6
+		protocol = iptables.ProtocolIPv6
 	}
-	masqChain := utiliptables.Chain(masqChainName)
-	iptables := utiliptables.New(execer, dbus, protocol)
+	ipt, _ := iptables.NewWithProtocol(protocol)
```
BenTheElder: Do not ignore this error. This method actually returns errors.

images/kindnetd/cmd/kindnetd/main.go
```diff
@@ -70,11 +70,9 @@ func main() {
 	// enforce ip masquerade rules
 	// TODO: dual stack...?
 	masqAgent := NewIPMasqAgent(net.IsIPv6String(hostIP), []string{os.Getenv("POD_SUBNET")})
+	// TODO: handle the errors and logging here, temporally the agent will run forever
```
BenTheElder: Temporally?
Also we are not handling errors here? but the logging point applies to the sync forever code.

images/kindnetd/cmd/kindnetd/masq.go
```diff
@@ -17,49 +17,42 @@ limitations under the License.
...
+		iptables:          ipt,
+		masqChain:         masqChainName,
 		noMasqueradeCIDRs: noMasqueradeCIDRs,
 	}
 }
 
 // IPMasqAgent is based on https://github.com/kubernetes-incubator/ip-masq-agent
 // but collapsed into kindnetd and made ipv6 aware in an opionated and simplified
-// fashion
+// fashion using "github.com/coreos/go-iptables" instead of k8s/pkg/util/iptables
```
BenTheElder: Looks actively enough maintained

aojea: is being used by the cni plugins, we have an indirect dependency right now anyway

images/kindnetd/cmd/kindnetd/masq.go
aojea: it returns errors if it doesn't find the `iptables`binary or is not able to find the version
https://github.com/coreos/go-iptables/blob/master/iptables/iptables.go#L99
this is running inside the `debian-iptables` image that should guarantee that

images/kindnetd/cmd/kindnetd/main.go
aojea: I was trying to understand what happens if the goroutine dies (somehow) or if it is constantly failing, shouldn't we handle it from here?

images/kindnetd/cmd/kindnetd/masq.go
BenTheElder: Ignoring this is very poor practice because it may return errors for new reason in the future, and we don't know when the version requirements will change, etc.

images/kindnetd/cmd/kindnetd/main.go
BenTheElder: Goroutines dont just die.
If it was constantly failing that should be handled inside the method, possibly with a panic.

images/kindnetd/cmd/kindnetd/masq.go
aojea: Good point

aojea:
@BenTheElder PTAL, I think I addresed all the comments

BenTheElder:
pushing

BenTheElder:
[rebasing before updating base image]

BenTheElder:
thank you!

Issue #110, opened 2018-11-14
Support insecure-registries for container runtime running inside of kind container


font:
The host that is running `kind` to set up kind clusters may want to create container images to be pulled by the container runtime (docker/containerd daemons) running inside of the `kind-<name>-control-plane` containers e.g. `kind-1-control-plane`. To simplify this, it would be great to have a way to easily configure the container runtime running inside the kind containers with `insecure-registries` in order to pull images from the host's insecure registry. This would simplify the local registry setup on the host to not require TLS.

For now, I have used the following workaround:

1. Run local insecure registry on host:
```bash
docker run -d -p 5000:5000 --restart=always --name registry registry:2
```
2. Create `daemon.json` inside of the kind container via docker exec:
```bash
docker exec kind-1-control-plane bash -c "cat <<EOF > /etc/docker/daemon.json
{
    "insecure-registries": ["172.17.0.1:5000"]
}
EOF
```
3. Send `SIGHUP` to docker daemon in kind container in order to reload config:

```bash
docker exec kind-1-control-plane bash -c 'kill -s SIGHUP $(pgrep dockerd)'
```
This works for now and then any container image to be pulled needs to be specified like so:
```bash
docker pull 172.17.0.1:5000/<imagename>:<tag>
```

BenTheElder:
SGTM, looks like both [cri-o](https://github.com/kubernetes-sigs/cri-o/blob/master/docs/crio.conf.5.md#crioimage-table) and [containerd](https://github.com/containerd/cri/blob/master/docs/registry.md) support this as well so if we want to use those inside the container in the future this can still be supported.

We can add a config option to specify a list of insecure registries and write it through to the daemon config before we start the daemon.

One thought though, specifically for the case of using a registry running on the host where kind is running, probably we can avoid the user needing to know what IP kind will see the host as, otherwise this config will be brittle / non-portable.

https://dev.to/bufferings/access-host-from-a-docker-container-4099 looks like an option for that. perhaps we can have config like:
```yaml
insecureLocalRegistryPort: 5000
```

and then images can be at `host.docker.internal:5000/foo-image` ?

cc @munnerz 

This also seems related to #28

padiazg:
Please, take in account also that there is the possibility of using a private registry with self signed certificates, and to use this you need also put the corresponding CA certificate in place. 
Ex: `/etc/docker/certs.d/<registry-ip>/ca.crt`

[...]

PR #953, opened 2019-10-16
update containerd config

BenTheElder:
- use non-deprecated config options for default CRI runtime (see https://github.com/containerd/containerd/releases/tag/v1.3.0)
- enable config imports (added in v1.3.0) which allow creating partial config file drop-ins that are merged with the main config file xref https://github.com/kubernetes-sigs/kind/issues/110

### DIFF
--- a/images/base/containerd-config.toml
+++ b/images/base/containerd-config.toml
@@ -2,10 +2,11 @@
 disabled_plugins = ["aufs", "btrfs", "zfs"]
 
 # set default runtime handler to v2, which has a per-pod shim
-[plugins.cri.containerd.default_runtime]
-  runtime_type="io.containerd.runc.v2"
+[plugins."io.containerd.grpc.v1.cri".containerd]
+  default_runtime_name="runc"
+[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
+  runtime_type = "io.containerd.runc.v2"
 
-# Setup a runtime with the magic name ("test-handler") used for Kubernetes
-# runtime class tests ...
+# Setup a runtime with the magic name ("test-handler") for runtime class tests
 [plugins.cri.containerd.runtimes.test-handler]
   runtime_type = "io.containerd.runc.v2"
--- a/images/base/containerd-config.toml
+++ b/images/base/containerd-config.toml
@@ -1,6 +1,10 @@
 # disable plugins we don't / can't support
 disabled_plugins = ["aufs", "btrfs", "zfs"]
 
+# import any additional config files written by kind, and any user written
+# config files (/etc/containerd/user-config/*.toml)
+imports = ["/kind/containerd/*.toml", "/etc/containerd/user/*.toml"]
+
 # set default runtime handler to v2, which has a per-pod shim
 [plugins."io.containerd.grpc.v1.cri".containerd]
   default_runtime_name="runc"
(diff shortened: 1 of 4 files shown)


images/base/containerd-config.toml
```diff
@@ -2,10 +2,15 @@
 disabled_plugins = ["aufs", "btrfs", "zfs"]
 
 # set default runtime handler to v2, which has a per-pod shim
-[plugins.cri.containerd.default_runtime]
-  runtime_type="io.containerd.runc.v2"
+[plugins."io.containerd.grpc.v1.cri".containerd]
+  default_runtime_name="runc"
+[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
+  runtime_type = "io.containerd.runc.v2"
```
BenTheElder: xref: https://github.com/containerd/cri/blob/815f3e727de84cda4c74ac4324d6469d20df94f7/cluster/gce/configure.sh#L180-L183

BenTheElder:
xref: https://github.com/containerd/cri/issues/1076, https://github.com/containerd/containerd/pull/3574

images/base/containerd-config.toml
```diff
@@ -2,10 +2,15 @@
...
 
-# Setup a runtime with the magic name ("test-handler") used for Kubernetes
-# runtime class tests ...
+# Setup a runtime with the magic name ("test-handler") for runtime class tests
 [plugins.cri.containerd.runtimes.test-handler]
   runtime_type = "io.containerd.runc.v2"
+
+# import any additional config files written by kind, and any user written
+# config files (/etc/containerd/user-config/*.toml)
+imports = ["/kind/containerd/*.toml", "/etc/containerd/user/*.toml"]
```
BenTheElder: i think maybe this needs to move up? so far this part isn't working, the rest of the config is definitely being loaded though

/hold

images/base/containerd-config.toml
aojea: :thinking: in the long term, how can we be sure we don't deviate from the official cri?

BenTheElder: huh?

images/base/containerd-config.toml
BenTheElder: testing more pending https://github.com/kubernetes-sigs/kind/pull/954

aojea: yeah, the examples show it always at the top of the file

BenTheElder: I did this locally but haven't gotten a registry mirror override working yet.

I need to brush up on toml, but it seems:
- indent is just for humans, not syntax
- any top level keys need to be before any tables

BenTheElder: pushed the moved up re-do. still need to test more with dropins. working on it... will resume in the morning (have wg-k8s-infra meeting in ~7 hours ...)

images/base/containerd-config.toml
aojea: you are getting the configs models from a bash script in the containerd/cri repo, I was wondering how we can track any changes there, it can happen that they change things and we deviate, thus we'll be testing different things.

BenTheElder: No, this is documented elsewhere. I'm just referencing that in this regard we're the same as another user, we have no reason not to be.

BenTheElder: This is both in the release notes and in the config docs upstream. We're just opting into the v2 (pod shim)

BenTheElder:
https://github.com/containerd/containerd/issues/3776 was the issue, moved to https://github.com/kubernetes-sigs/kind/pull/1007

Issue #821, opened 2019-08-28
Issue while creating kind cluster with workers


harshavardhanc:
I'm trying to create cluster in local(Arch linux) with single master node it is working fine, but if I try to create a cluster with workers getting the below error.

```
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.15.0) 🖼 
ERRO[19:19:20] 40e341e7e12ebf6301f6ad6eab0927d16689032987a6cb796af209c28df1c12b 
ERRO[19:19:20] docker: Error response from daemon: driver failed programming external connectivity on endpoint kind-worker (b3ad65eb44b2852397f568c3f074129f65c8039e4bc66a298f69d6dfae03970e):  (iptables failed: iptables --wait -t nat -A DOCKER -p tcp -d 0/0 --dport 32787 -j DNAT --to-destination 172.17.0.4:0 ! -i docker0: iptables v1.8.3 (nf_tables): Port `0' not valid 
ERRO[19:19:20]                                              
ERRO[19:19:20] Try `iptables -h' or 'iptables --help' for more information. 
ERRO[19:19:20]  (exit status 2)).                           
 ✗ Preparing nodes 📦📦
ERRO[19:19:20] docker run error: exit status 125            
Error: failed to create cluster: docker run error: exit status 125
```
Below is my config file to create a cluster.
```
kind: Cluster
apiVersion: kind.sigs.k8s.io/v1alpha3
nodes:
- role: control-plane
- role: worker
  extraPortMappings:
  - containerPort: 3000
    hostPort: 3000
  - containerPort: 3443
    hostPort: 3443
  - role: worker
```

Please help me in fixing this issue.

BenTheElder:
see https://github.com/kubernetes-sigs/kind/issues/821 for improvements to catching this error.

the corrected config would be:
```
kind: Cluster
apiVersion: kind.sigs.k8s.io/v1alpha3
nodes:
- role: control-plane
- role: worker
  extraPortMappings:
  - containerPort: 3000
    hostPort: 3000
  - containerPort: 3443
    hostPort: 3443
- role: worker
```

PR #954, opened 2019-10-16
fix extraPortMappings / extraMounts and add more tests

BenTheElder:
- fix yaml unmarshal for types with custom unmarshal
- add test data to cover these

should be caught by CI next time. will spend more time on covering this in the future...

### DIFF
--- a/pkg/container/cri/yaml.go
+++ b/pkg/container/cri/yaml.go
@@ -28,17 +28,24 @@ Custom YAML (de)serialization for these types
 // UnmarshalYAML implements custom decoding YAML
 // https://godoc.org/gopkg.in/yaml.v3
 func (m *Mount) UnmarshalYAML(unmarshal func(interface{}) error) error {
-	type Alias Mount
-	aux := &struct {
-		Propagation string `yaml:"propagation"`
-		*Alias
-	}{
-		Alias: (*Alias)(m),
+	// this is basically Mount, except Propagation is a string for further parsing
+	type MountYaml struct {
+		ContainerPath  string `yaml:"containerPath,omitempty"`
+		HostPath       string `yaml:"hostPath,omitempty"`
+		Readonly       bool   `yaml:"readOnly,omitempty"`
+		SelinuxRelabel bool   `yaml:"selinuxRelabel,omitempty"`
+		Propagation    string `yaml:"propagation,omitempty"`
 	}
+	aux := MountYaml{}
 	if err := unmarshal(&aux); err != nil {
 		return err
 	}
-	// if unset, will fallback to the default (0)
+	// copy over normal fields
+	m.ContainerPath = aux.ContainerPath
+	m.HostPath = aux.HostPath
+	m.Readonly = aux.Readonly
+	m.SelinuxRelabel = aux.SelinuxRelabel
+	// handle special field
 	if aux.Propagation != "" {
 		val, ok := MountPropagationNameToValue[aux.Propagation]
 		if !ok {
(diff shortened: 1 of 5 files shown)


pkg/container/cri/yaml.go
```diff
@@ -28,17 +28,24 @@ Custom YAML (de)serialization for these types
 // UnmarshalYAML implements custom decoding YAML
 // https://godoc.org/gopkg.in/yaml.v3
 func (m *Mount) UnmarshalYAML(unmarshal func(interface{}) error) error {
-	type Alias Mount
-	aux := &struct {
-		Propagation string `yaml:"propagation"`
-		*Alias
-	}{
```
BenTheElder: sadly we have to give up this trick, yaml.v3 disallows duplicate keys.

pkg/container/cri/yaml.go
```diff
@@ -28,17 +28,24 @@ Custom YAML (de)serialization for these types
...
-		Alias: (*Alias)(m),
+	// this is basically Mount, except Propogation is a string for further parsing
+	type MountYaml struct {
```
BenTheElder: doing this gives pretty good errors though (type names show up in errors regarding keys)

pkg/container/cri/yaml.go
aojea: well, in the other said it forces us to be explicit :)

pkg/internal/apis/config/encoding/testdata/v1alpha3/invalid-bad-indent.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - containerPath: /foo
+    hostPath: /bar
+    readOnly: true
+    selinuxRelabel: false
+    propagation: Bidirectional
+  extraPortMappings:
+  - containerPort: 8080
+    hostPort: 8080
+    protocol: UDP
+  - role: worker
```
BenTheElder: we had an issue similar to this config, https://github.com/kubernetes-sigs/kind/issues/821

BenTheElder: this is now an error :-)

pkg/internal/apis/config/encoding/testdata/v1alpha3/valid-many-fields.yaml
```diff
@@ -0,0 +1,17 @@
+kind: Cluster
+apiVersion: kind.sigs.k8s.io/v1alpha3
+networking:
+  ipFamily: ipv6
```
BenTheElder: doesn't cover _every_ field, however the mounts / port mappings are the finicky ones, will come back and add more test cases later.

BenTheElder:
[gofmt-ed]

pkg/container/cri/yaml.go
BenTheElder: yeah. this was a clever and concise trick though, I had fun discovering this :-)

pkg/container/cri/yaml.go
```diff
@@ -28,17 +28,24 @@ Custom YAML (de)serialization for these types
 // UnmarshalYAML implements custom decoding YAML
 // https://godoc.org/gopkg.in/yaml.v3
 func (m *Mount) UnmarshalYAML(unmarshal func(interface{}) error) error {
-	type Alias Mount
-	aux := &struct {
-		Propagation string `yaml:"propagation"`
-		*Alias
-	}{
-		Alias: (*Alias)(m),
+	// this is basically Mount, except Propogation is a string for further parsing
```
aojea: nit typo Propogation -> Propagation

pkg/internal/apis/config/encoding/testdata/v1alpha3/valid-many-fields.yaml
aojea: I suggest creating an issue with help-wanted for these things, the same way you did the shellcheck meta issue in k/k

pkg/container/cri/yaml.go
BenTheElder: CI caught this too :P

BenTheElder: done

aojea:
/test pull-kind-conformance-parallel-1-16
failure unrelated to the PR

BenTheElder:
prow failure specifically, no logs

pkg/internal/apis/config/encoding/testdata/v1alpha3/valid-many-fields.yaml
aojea: also, some fields like `ipFamily` are covered by validation, not perfect, but at least the bad config is not going through

BenTheElder:
ignoring ipv6 flake

PR #959, opened 2019-10-17
docs: few clarifications for Windows

### LINKED ISSUE #837: README.md: Getting Started on Windows
Getting started on Windows

**Why is this needed**:
The current "Windows" download (in the "Assets" section of the release notes) is a file without an extension.  This leads to confusion on what you are supposed to do with it.

Additionally, the main page says to run `curl -Lo ./kind https://github.com/kubernetes-sigs/kind/releases/download/v0.5.1/kind-$(uname)-amd64` and place it in the path.  But running that on Windows (powershell) gives the error :

> uname : The term 'uname' is not recognized as the name of a cmdlet, function, script file, or operable program.

Addionally the next instruction uses `chmod`, which also errors out on Windows.

Some instructions similar to this would have helped me:
 * Go to https://github.com/kubernetes-sigs/kind/releases
 * Find the "► Assets" dropdown and then the Windows download.
 * Renamed the downloaded file to `kind.exe`
 * Place the downloaded file in your PATH
 * open PowerShell/Command Prompt and run `kind`.  You should get the help output  for kind.

PatrickLang:
This is a few quick updates to make the Windows experience a bit smoother.

1. Update the WSL2 doc to skip updating the kernel. All builds since July 2019 have had a compatible kernel already
2. Resolves #837 by clarifying download steps in README.md to match the quick start.
3. Should fix the word "powershell" appearing in code section. If the fix doesn't work, I'll need to make another update

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,10 +41,19 @@ generally recommended for CI usage in particular.
 To install, download the binary for your platform from "Assets" and place this
 into your `$PATH`:
 
+On Mac & Linux:
+
 ```console
 curl -Lo ./kind https://github.com/kubernetes-sigs/kind/releases/download/v0.5.1/kind-$(uname)-amd64
 chmod +x ./kind
 mv ./kind /some-dir-in-your-PATH/kind
+```
+
+On Windows:
+
+```powershell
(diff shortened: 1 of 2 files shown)


BenTheElder:
/approve
thank you!

PR #963, opened 2019-10-18
remove bazel platforms flag since cross compiling doesn't work anyhow

BenTheElder:
/cc @cblecker


BenTheElder:
definitely built and tests are up and running across the board

PR #964, opened 2019-10-19
use json-patch

### LINKED ISSUE #962: do unstructured patch with json-patch library directly
**What should be cleaned up or changed**: rewrite our kubeadm patching logic to use https://github.com/evanphx/json-patch directly.

**Why is this needed**: 
- kustomize is > 3/4 of our binary size 
  - macOS binary as of 7d8bbb4 is 34M
  - macOS binary as of 7d8bbb4 + a trivial patch to no-op patching is 8.3M (!)
- kustomize brings in heavy dependencies (in terms of managing a `go get`-able / embeddable tool) like client-go.
  - In fact most of our transitive dependencies are due to kustomize. 
  - The latest kustomize go.mod even includes things like golangci-lint

Most of these things are perfectly reasonable for kustomize to do as a tool, however we're not seeing the benefits of them in kind specifically because the types were are patching are totally unstructured anyhow without type definitions.

A tool like kubeadm on the other hand sees the strategic-merge benefits because it patches objects that are core k8s api types.

In addition, we can loosen the restrictions around matching patches. It will be easier to write trivial kubeadm config patches if you don't _have_ to specify the apiVersion, many of the fields are the same anyhow.

We can also get rid of the in memory filesystem and windows work arounds, and hopefully generally have a simpler implementation of patching.

/assign

BenTheElder:
fixes #962


BenTheElder:
TODO:
- [ ] more test cases
- [ ] verify new test cases against previous impl
- [ ] cleanup error handling
- [ ] add comments

go.mod
```diff
@@ -4,13 +4,16 @@ go 1.13
...
 	github.com/pkg/errors v0.8.1
 	github.com/spf13/cobra v0.0.3
+	github.com/spf13/pflag v1.0.5 // indirect
 	golang.org/x/crypto v0.0.0-20190611184440-5c40567a22f8
+	gopkg.in/yaml.v2 v2.2.4 // indirect
 	gopkg.in/yaml.v3 v3.0.0-20191010095647-fc94e3f71652
-	k8s.io/api v0.0.0-20190409021203-6e4e0e4f393b // indirect
 	k8s.io/apimachinery v0.0.0-20190404173353-6a84e37a896d
-	sigs.k8s.io/kustomize/v3 v3.2.0
+	k8s.io/klog v1.0.0 // indirect
```
BenTheElder: guessing apimachinery is still pulling this in :|

go.mod
```diff
@@ -4,13 +4,16 @@ go 1.13
 
 require (
 	github.com/alessio/shellescape v0.0.0-20190409004728-b115ca0f9053
+	github.com/evanphx/json-patch v4.5.0+incompatible
 	github.com/google/uuid v1.1.1
+	github.com/inconshreveable/mousetrap v1.0.0 // indirect
```
BenTheElder: cobra pulls this in

go.mod
```diff
@@ -4,13 +4,16 @@ go 1.13
...
 	github.com/pkg/errors v0.8.1
 	github.com/spf13/cobra v0.0.3
+	github.com/spf13/pflag v1.0.5 // indirect
```
BenTheElder: also cobra

go.mod
```diff
@@ -4,13 +4,16 @@ go 1.13
...
 	golang.org/x/crypto v0.0.0-20190611184440-5c40567a22f8
+	gopkg.in/yaml.v2 v2.2.4 // indirect
```
BenTheElder: sigs.k8s.io/yaml, eventually we may de-dupe this if that library gets on .v3 but it's not looking likely soon

BenTheElder:
```console
$ du -h bin/*
9.1M    bin/kind-darwin-amd64
9.2M    bin/kind-linux-amd64
8.1M    bin/kind-linux-arm
8.9M    bin/kind-linux-arm64
9.1M    bin/kind-linux-ppc64le
9.1M    bin/kind-windows-amd64
```

pkg/internal/util/patch/resource.go
```diff
@@ -0,0 +1,118 @@
...
+		return true, errors.WithStack(err)
+	}
+	r.json = patched
+	return true, nil
+}
+
+func (r resource) matches(o matchInfo) bool {
+	m := &r.matchInfo
+	// we require kind to match, but if the patch does not specify
+	// APIVersion we ignore it (eg to allow trivial patches across kubeadm versions)
```
aojea: what are the tradeoffs of this decision?

BenTheElder: It makes it easier to write kind configs with patches.
I don't see any obvious downside to our use case, it's up to the user to be more specific if they want.

The bigger change is that we don't require all patches to match something, but again, that's up to the user, and actually enables writing a config that patches all versions...

This is more relaxed behavior, and easier to work with.

BenTheElder: Kubernetes types are identified by (`kind`, `apiVersion`), the `apiVersion` is just that, the API Version. Many versions are near identical. Different kinds are not expected to be.

For our specific use case, I'm not aware of any useful patch you could write that would apply to all objects. However, it would be useful to patch identical fields across kubeadm API versions.

BenTheElder:
`go get sigs.k8s.io/kind` (without modules enabled) also works after this :-)

PR #969, opened 2019-10-20
provider/common tests

BenTheElder:
this is PR #919 + cleanup


BenTheElder:
rebased, and increased coverage to 100%, also fixed some behavior slightly while inspecting odd existing test case values in GetProxyEnvs

aojea:
/lgtm

thanks for the fixes  😅

pkg/internal/cluster/providers/provider/common/proxy.go
```diff
@@ -48,7 +52,11 @@ func GetProxyEnvs(cfg *config.Cluster) map[string]string {
 	}
 	// Specifically add the cluster subnets to NO_PROXY if we are using a proxy
 	if len(envs) > 0 {
-		noProxy := strings.Join([]string{envs[NOProxy], cfg.Networking.ServiceSubnet, cfg.Networking.PodSubnet}, ",")
```
BenTheElder: previously we would do something like `,10.0.0.0/24,12.0.0.0/24` if `NO_PROXY` was not found which technically seems to have worked (?) but at best seems strange and at worse seems likely to break some application, given how under-defined this environment variable "API" is.

aojea: I remember vaguely to have verified it ... but don't remember how 😅

Issue #971, opened 2019-10-20
panic in exec


BenTheElder:
https://github.com/kubernetes/kubernetes/pull/84122#issuecomment-544215170

/priority critical-urgent
/assign

BenTheElder:
https://github.com/kubernetes-sigs/kind/blob/7dd62cd264beb7850d367a736bf5c7b5d4af015e/pkg/exec/local.go#L75-L76

🤦‍♂ 🤦‍♂

BenTheElder:
Go's standard library has:
```
// CombinedOutput runs the command and returns its combined standard
// output and standard error.
func (c *Cmd) CombinedOutput() ([]byte, error) {
	if c.Stdout != nil {
		return nil, errors.New("exec: Stdout already set")
	}
	if c.Stderr != nil {
		return nil, errors.New("exec: Stderr already set")
	}
	var b bytes.Buffer
	c.Stdout = &b
	c.Stderr = &b
	err := c.Run()
	return b.Bytes(), err
}
```
however, presumably this is a bit safer than a pair of MultiWriters.

BenTheElder:
on the plus side, digging into what's going on here ... i think there's a chance fixing this fixes #949

BenTheElder:
So if `cmd.Stdout` and `cmd.Stderr` are identical, then Go will set up the actual fd for stdout and then set stderr to the same fd.

https://golang.org/src/os/exec/exec.go
```go
func (c *Cmd) stdout() (f *os.File, err error) {
	return c.writerDescriptor(c.Stdout)
}

func (c *Cmd) stderr() (f *os.File, err error) {
	if c.Stderr != nil && interfaceEqual(c.Stderr, c.Stdout) {
		return c.childFiles[1], nil
	}
	return c.writerDescriptor(c.Stderr)
}
```

This is why `CombinedOuput` in the stdlib is safe.

When we set up an `io.MultiWriter` they are _not_ the same, and get different fds.

In that case, we need to handle the writer receiving concurrent writes.

BenTheElder:
Filed https://github.com/kubernetes-sigs/kind/pull/972 with more details and a fix.

BenTheElder:
This is a little tricky to intentionally reproduce, I believe #972 should prevent it, but we should keep our eyes peeled.

PR #972, opened 2019-10-20
synchronize exec combined output buffer when necessary

BenTheElder:
When we execute things in kind we capture the output in case of an error and provide this when debugging. This is very handy, but had a nasty bug, xref: #971.

When `cmd.Stdout == cmd.Stderr`, go will set up a single fd passed to the underyling process,  and single `io.Copy` goroutine will be used for that fd. This makes it safe to share a buffer between `cmd.Stdout` and `cmd.Stderr` IFF `cmd.Stdout == cmd.Stderr`.

When this is not the case, we need to synchronize writes to the combined output buffer.

See: 
- https://go-review.googlesource.com/c/go/+/42052/ (`os/exec: document that non-comparable writers may race`)
- https://github.com/golang/go/blob/72275c0dea7887597e3e445676674ba6af702f06/src/os/exec/exec.go#L216-L223
- https://github.com/golang/go/blob/72275c0dea7887597e3e445676674ba6af702f06/src/os/exec/exec.go#L282-L287

fixes: #971

### DIFF
--- a/pkg/exec/local.go
+++ b/pkg/exec/local.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"io"
 	osexec "os/exec"
+	"sync"
 
 	"sigs.k8s.io/kind/pkg/errors"
 	"sigs.k8s.io/kind/pkg/globals"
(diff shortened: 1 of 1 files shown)


BenTheElder:
IPv6 e2e flakes

aojea:
nice catch and thanks to @liggitt for keep finding occurrences
/lgtm

the ipv6 flakes show errors with some application holding the lock, I'm trying to investigate more

BenTheElder:
> the ipv6 flakes show errors with some application holding the lock, I'm trying to investigate more

[iptables, on the node, not the lock we introduced here]

Issue #976, opened 2019-10-20
[Flaky Test] kind-ipv6-master-parallel


aojea:
It was initially reported here https://github.com/kubernetes/kubernetes/issues/83903#issuecomment-542867717

but we really should know why these iptables errors are not showing up on the ipv4 jobs, ~is it possible that the different CI hardware is not fast enough to release the lock?~

>Another flaky with iptables errors, this time wasn't ip6tables-restore , is the new kindnet that uses go-iptables

```
2019-10-16T18:20:33.983136705Z stdout F Node kind-worker2 has CIDR fd00:10:244:0:1::/80 
2019-10-16T18:20:39.810650256Z stderr F panic: Can't synchronize rules after 3 attempts: running [/sbin/ip6tables -t nat -C KIND-MASQ-AGENT -j MASQUERADE -m comment --comment kind-masq-agent: outbound traffic is subject to MASQUERADE (must be last in chain) --wait]: exit status 4: ip6tables: Resource temporarily unavailable.
2019-10-16T18:20:39.810729436Z stderr F 
2019-10-16T18:20:39.810738416Z stderr F 
2019-10-16T18:20:39.810745057Z stderr F goroutine 20 [running]:
2019-10-16T18:20:39.810754027Z stderr F main.main.func1(0xc00029e150)
2019-10-16T18:20:39.810761796Z stderr F 	/src/cmd/kindnetd/main.go:79 +0x93
2019-10-16T18:20:39.810768298Z stderr F created by main.main
2019-10-16T18:20:39.810789908Z stderr F 	/src/cmd/kindnetd/main.go:77 +0x32a
Job link https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-kind-conformance-parallel-ipv6/1184528903125864450
```

Log https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-kind-conformance-parallel-ipv6/1184528903125864450/artifacts/logs/kind-worker/pods/kube-system_kindnet-7c82z_7a2d7628-c682-42b2-bdeb-b3a60420969c/kindnet-cni/0.log

aojea:
Jobs to track
* https://k8s-testgrid.appspot.com/sig-release-master-blocking#kind-ipv6-master-parallel&width=5
* https://k8s-testgrid.appspot.com/sig-testing-kind#kind%20(IPv6),%20master%20(dev)%20%5Bnon-serial%5D&width=5

[...]

PR #977, opened 2019-10-21
mount xtables lock file and lib/modules in kindnetd

BenTheElder:
fixes: #976 

iptables commands with `--wait` (as used by kube-proxy and kindnetd and...) use a file based lock which is by default at `/run/xtables.lock`, we need to mount this.

We should also be mounting `/lib/modules`

See: https://git.netfilter.org/iptables/tree/configure.ac?id=55a7558bb2c86e650809610e976e9d5192fe4e7e#n72

### DIFF
--- a/pkg/build/node/cni.go
+++ b/pkg/build/node/cni.go
@@ -51,6 +51,8 @@ spec:
     - hostPath
   allowedHostPaths:
     - pathPrefix: "/etc/cni/net.d"
+    - pathPrefix: "/run"
+    - pathPrefix: "/lib"
   readOnlyRootFilesystem: false
   # Users and groups
   runAsUser:
@@ -159,6 +161,12 @@ spec:
         volumeMounts:
         - name: cni-cfg
           mountPath: /etc/cni/net.d
+        - name: xtables-lock
+          mountPath: /run/xtables.lock
+          readOnly: false
+        - name: lib-modules
+          mountPath: /lib/modules
+          readOnly: true
         resources:
           requests:
             cpu: "100m"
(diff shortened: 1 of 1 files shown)


BenTheElder:
(also kube-proxy mounts these)

aojea:
/hold
kindnetd is suing hostNetwork, AFAIK he has access to those files

```
root@kind-worker:/# crictl ps
CONTAINER           IMAGE               CREATED             STATE               NAME                ATTEMPT             POD ID
8b3fdeadeb7e9       5f29e9888973c       10 hours ago        Running             kube-proxy          0                   c07df53f2f416
3f988106dad9f       67a7e17226181       10 hours ago        Running             kindnet-cni         0                   e644b2c76ee57
root@kind-worker:/# crictl exec -it 3f988106dad9f sh
# ps axf
sh: 1: ps: not found
# ps
sh: 2: ps: not found
# ls /
bin  boot  dev  etc  home  lib  lib64  media  mnt  opt  proc  root  run  sbin  srv  sys  tmp  usr  var
# ls /run 
lock  secrets  utmp  xtables.lock
```

BenTheElder:
> kindnetd is using hostNetwork: true, AFAIK it has access to those files

hostNetwork has nothing to do with those?

the fact that it has _a_ `/run/xtables.lock` does not mean it has the _same_ xtables.lock from the host.

BenTheElder:
note that:
- kube-proxy (kubeadm and kube-up gce)
- weave
- calico
- ...

all have these mounts.

aojea:
/hold cancel
/lgtm

Thanks for the lessons by the way

BenTheElder:
of course!
[we discussed more in slack, TLDR `hostNetwork` affects the network namespace, but this is in the filesystem]

🤞 hopefully this finishes deflaking kindnetd!

PR #978, opened 2019-10-21
Allow specifying network name to use in docker provisioner

keymone:


aojea:
/hold
duplicate of https://github.com/kubernetes-sigs/kind/pull/484

Thanks for your PR, but this has some issues on the way that docker handles user defined bridges t

tao12345666333:
Thanks for your work.

As #484 comments you can find more detail.

BenTheElder:
this won't work very well without a more complete design, in particular because the default network has different behavior. let's figure out the details in the related issues and then PR an implementation once we have that agreed and workable.

Issue #980, opened 2019-10-21
duplicated key 'name' in struct v1alpha3.PatchJSON6902


ysz:
**What happened**:

```
$ ./kind create cluster --wait 5m --config site/content/docs/user/kind-example-config.yaml
Creating cluster "kind" ...
panic: duplicated key 'name' in struct v1alpha3.PatchJSON6902 [recovered]
	panic: duplicated key 'name' in struct v1alpha3.PatchJSON6902

goroutine 1 [running]:
gopkg.in/yaml%2ev3.handleErr(0xc0000e58e0)
	/Users/ysz/go/pkg/mod/gopkg.in/yaml.v3@v3.0.0-20191010095647-fc94e3f71652/yaml.go:276 +0x9a
panic(0x137e6a0, 0xc00004f1b0)
```

**What you expected to happen**:

Create a cluster 

**How to reproduce it (as minimally and precisely as possible)**:

```
go build
./kind create cluster --wait 5m --config site/content/docs/user/kind-example-config.yaml
```

**Environment:**

- kind version: (use `kind version`): master

BenTheElder:
We should also clarify: the example config is a straw man showing all of the fields, it _should_ work but it's not actually meant to be a useful config.

ysz:
@BenTheElder Thank you for clarifying. I'm reading Kubernetes Up and Running and want to quickly spin up a multi node cluster to follow examples. :-)

BenTheElder:
fixed in #981, thanks again!
clarifying config slightly in #983

PR #983, opened 2019-10-21
update example config to clarify slightly

BenTheElder:
- clarify that this is not really a config you'd want to use, it's just to show some fields
- don't clobber the serviceSubnet now that we use higher level config for that, patch something else instead.


BenTheElder:
xref: #980

PR #984, opened 2019-10-21
docs: Adding troubleshooting section for WSL2

PatrickLang:
Just linking workarounds for a few known issues. I'm not sure if/when they'll be fixed since both are related to systemd and there doesn't seem to be a motion to enable that on WSL2 yet.

### DIFF
--- a/site/content/docs/user/using-wsl2.md
+++ b/site/content/docs/user/using-wsl2.md
@@ -66,6 +66,48 @@ Once your Windows Insider machine is ready, you need to do a few more steps to s
 
 Now, move on to the [Quick Start](/docs/user/quick-start) to set up your cluster with kind.
 
+## Troubleshooting
+
+### Failures with `kind cluster create`
+
+#### Cannot find cgroup mount destination: unknown
+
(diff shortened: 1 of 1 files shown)


site/content/docs/user/using-wsl2.md
```diff
@@ -66,6 +66,48 @@ Once your Windows Insider machine is ready, you need to do a few more steps to s
 
 Now, move on to the [Quick Start](/docs/user/quick-start) to set up your cluster with kind.
 
+## Troubleshooting
+
+### Failures with `kind cluster create`
```
aojea: this renders a bit weird to me, is this on purpose using the code marks on the title?

BenTheElder: https://deploy-preview-984--k8s-kind.netlify.com/docs/user/using-wsl2/#troubleshooting

BenTheElder:
/lgtm
/approve
thank you!

site/content/docs/user/using-wsl2.md
aojea: :man_facepalming: 
I always forget I can see the preview on that link
sorry

PR #985, opened 2019-10-21
test kind config with patches

aojea:
we should try to add unit tests for each bug to cover regressions


BenTheElder:
> test kind config with patches #985

we actually already test patching with these type of patches, however we don't use namespace as it's deprecated

BenTheElder:
(this is why I didn't rush to write a PR like this :-))

aojea:
we have an awesome community, detected the bug in master and patched it 😍

Issue #986, opened 2019-10-21
Support an integrated test workflow


BenTheElder:
IE: replace the prototype in `hack/ci/e2e-k8s.sh` with a built in test command.

This will make it even easier to replicate kind's Kubernetes PR test results, and allow us to better test and support this.

kind should remain trivially scriptable for supported use csaes, but for testing Kubernetes specifically we should provide a robust tool that does this right out of the box, seeing as this is the founding purpose of the project.

More details (shared with kubernetes-dev@googlegroups.com and kubernetes-sig-testing@googlegroups.com) https://docs.google.com/document/d/16WdI8VTDmNt-TlFGzW1tJaLD7NM1r6ppn0k4GXQJNHU

/assign

BenTheElder:
added git version https://github.com/kubernetes-sigs/kind/pull/988
/lifecycle active

aojea:
@BenTheElder is this for this milestone?

BenTheElder:
I keep the milestone up to date. This is not a blocker for the milestone It would be nice to have and might land before then if I'm blocked on other things and finish submitting it.

aojea:
I'd love to have a `kind e2e.test --name kind --focus  \[sig-network\].*Conformance"` flag
maybe we can add the e2e.test to the node images when we are building the node image, what do you think?

BenTheElder:
did some more design and asked some key power users of this more about their thoughts directly ... bumping to the next milestone though because it's not quite as pressing as unblocking PV work

BenTheElder:
status update: writing down some fresh code for v0.8.0 MVP, will PR

BenTheElder:
reviving this, rebasing on some other test changes

BenTheElder:
i really wanted this, but there's too many important user fixes that need to ship. v0.9.0 we will start using this and collecting e2e coverage

BenTheElder:
# Updated Proposal

**NOTE**: Some of this is already covered by the document above, however I'm rehashing a current proposal here.

## 1. We should not attempt to support testing anything other than Kubernetes core, to keep the project maintainable and focus on things clearly in-scope for the project.

Users can implement and indeed already have implemened custom test workflows using the other kind commands (see for example the `hack/ci/e2e-k8s.sh` that we have today) or using the kind Go pkg APIs.

[...]

PR #988, opened 2019-10-21
include commit in pre-release builds

BenTheElder:
xref: #986

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,9 @@ INSTALL?=install
 # make install will place binaries here
 # the default path attempts to mimic go install
 INSTALL_DIR?=$(shell hack/build/goinstalldir.sh)
+# record the source commit in the binary
+COMMIT?=$(shell git rev-parse HEAD 2>/dev/null)
+LD_FLAGS:=-X sigs.k8s.io/kind/cmd/kind/version.GitCommit=$(COMMIT)
 # the output binary name, overridden when cross compiling
 KIND_BINARY_NAME?=kind
 
@@ -36,7 +39,7 @@ all: build
 
 # builds kind in a container, outputs to $(OUT_DIR)
 kind:
-	hack/go_container.sh go build -v -o /out/$(KIND_BINARY_NAME)
+	hack/go_container.sh go build -v -o /out/$(KIND_BINARY_NAME) -ldflags "$(LD_FLAGS)"
 
 # alias for building kind
 build: kind
--- a/cmd/kind/version/version.go
+++ b/cmd/kind/version/version.go
@@ -23,19 +23,41 @@ import (
 	"github.com/spf13/cobra"
 )
 
-// Version is the kind CLI version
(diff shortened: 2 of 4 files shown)


BenTheElder:
later we'll also use this in kubernetes CI to record the kind commit used to test.

BenTheElder:
/cc @amwat 
```
[bentheelder@bentheelder:~/go/src/sigs.k8s.io/kind·2019-10-21T15:37:22-0700·include-commit@dfbd929e]
$ make install
hack/go_container.sh go build -v -o /out/kind -ldflags "-X sigs.k8s.io/kind/cmd/kind/version.GitCommit=dfbd929e5ff1e665b07ff9bd46d966d281c51b4c"
install -d /usr/local/google/home/bentheelder/go/bin
install /usr/local/google/home/bentheelder/go/src/sigs.k8s.io/kind/bin/kind /usr/local/google/home/bentheelder/go/bin/kind

[bentheelder@bentheelder:~/go/src/sigs.k8s.io/kind·2019-10-21T15:45:20-0700·include-commit@dfbd929e]
$ kind version
v0.6.0-alpha+dfbd929e5ff1e665b07ff9bd46d966d281c51b4c
```

cmd/kind/version/version.go
```diff
@@ -23,19 +23,41 @@ import (
...
-const Version = "v0.6.0-alpha"
+// Version returns the kind CLI Semantic Version
+func Version() string {
+	v := VersionCore
+	// add pre-release version info if we have it
+	if VersionPreRelease != "" {
+		v += "-" + VersionPreRelease
+		// if commit was set, add the + <build>
+		// we only do this for pre-release versions
+		if GitCommit != "" {
```
BenTheElder: to consider in a follow-up: we might want to instead (or additionally) have the release scripts to `export COMMIT=""` to make it easier to reproducibly build the binaries from a source tarball. or instructions for doing `COMMIT=<release-commit> make build`

BenTheElder: I think it is probably better to include the commit, and then if someone wants to reproduce the binaries without git they can set `COMMIT=` (the makefile should already support this)

Makefile
```diff
@@ -28,6 +28,9 @@ INSTALL?=install
 # make install will place binaries here
 # the default path attempts to mimic go install
 INSTALL_DIR?=$(shell hack/build/goinstalldir.sh)
+# record the source commit in the binary
+COMMIT?=$(shell git rev-parse HEAD 2>/dev/null)
```
aojea: Why not the short commit i

aojea: It's easier to read imho

BenTheElder: The short version is just this truncated. We can truncate at runtime when formatting if we want, but this is the correct unambiguous commit ID.

BenTheElder: https://github.com/kubernetes-sigs/kind/pull/995

PR #990, opened 2019-10-22
kindnetd: read subnets from kubeadm config

BenTheElder:
revamped #739


BenTheElder:
NOTE @aojea  this version is slightly different, we're being slightly more robust to obtaining and processing the configmap.

BenTheElder:
local testing
```console
$ kubectl -n=kube-system logs kindnet-psflh 
hostIP = 192.168.9.3
podIP = 192.168.9.3
Syncing Masquerade rules with noMasqueradeCIDRs: [10.244.0.0/16 10.96.0.0/12]
handling current node
```

images/kindnetd/cmd/kindnetd/masq.go
```diff
@@ -58,7 +60,9 @@ type IPMasqAgent struct {
 func (ma *IPMasqAgent) SyncRulesForever(interval time.Duration) error {
 	errs := 0
 	for {
-		if err := ma.SyncRules(); err != nil {
+		// get subnets each time (so we eventually reconcile)
+		noMasqueradeCIDRs := getNoMasqueradeSubnets(ma.clientset)
```
aojea: we should check that this variable is not empty.
should we try to sync rules if is empty?

images/kindnetd/cmd/kindnetd/subnets.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// TODO: we have other fallback options for completeness including:
+	// - kube-apiserver flags
+	// - component config
+	// - defaults
+	return s.All()
+}
+
+const (
+	kubeadmServiceCIDRRegex = `(?m)^\s*serviceSubnet: (.*)\n`
```
aojea: @neolit123 are these fields `serviceSubnet` and `podSubnet` stable in kubeadm, are they following the API policies? also, you know if there were changes? if there were changes we should keep the previous environment variable configuration for backwards compatibility

images/kindnetd/cmd/kindnetd/subnets.go
```diff
@@ -0,0 +1,102 @@
...
+
+func getKubeadmSubnets(clientset *kubernetes.Clientset) *subnets {
+	s := newSubnets()
+	configMap, err := clientset.CoreV1().ConfigMaps("kube-system").Get("kubeadm-config", metav1.GetOptions{})
+	if err != nil {
+		fmt.Println("error getting kubeadm config: " + err.Error())
+		return s
+	}
+	for _, data := range configMap.Data {
+		// ensure data is yaml formatted and does not contain comments
```
aojea: dumb question, shouldn't this be guaranteed by client-go and kubeadm?

aojea: or is for double-checking and for removing the comments?

aojea: without the yamlify we should be able to work only with strings

pkg/build/node/cni.go
```diff
@@ -27,11 +27,10 @@ const (
 The default CNI manifest and images are our own tiny kindnet
 */
 
-var defaultCNIImages = []string{"kindest/kindnetd:0.5.1"}
+var defaultCNIImages = []string{"kindest/kindnetd:0.6.0"}
 
 const defaultCNIManifest = `
 # kindnetd networking manifest
-# would you kindly template this file
```
aojea: great :)

images/kindnetd/cmd/kindnetd/subnets.go
```diff
@@ -0,0 +1,102 @@
...
+	return s.pods.Union(s.services).List()
+}
+
+func getNoMasqueradeSubnets(clientset *kubernetes.Clientset) []string {
+	s := newSubnets()
+
+	// try getting from kubeadm config
+	kubeadmSubnets := getKubeadmSubnets(clientset)
+	s.pods = s.pods.Union(kubeadmSubnets.pods)
+	s.services = s.services.Union(kubeadmSubnets.services)
```
aojea: IIRC there is no need to insert the services subnet in the no-masquerade chain,  Tim Hockin  explained here https://github.com/kubernetes/kubernetes/issues/79866#issuecomment-519218155 that pod to service traffic is always masquerade 

All traffic directed to services inside a node is processed by this chain:
`
-A PREROUTING -m comment --comment "kubernetes service portals" -j KUBE-SERVICES 
`
and if it is directed to one service it sets a mark
`-A KUBE-MARK-MASQ -j MARK --set-xmark 0x4000/0x4000`
so this traffic is *always* masquerade
`-A KUBE-POSTROUTING -m comment --comment "kubernetes service traffic requiring SNAT" -m mark --mark 0x4000/0x4000 -j MASQUERADE `

I think that this no-masq rule will only affect the traffic from pods to services ips that doesn't exists

images/kindnetd/cmd/kindnetd/subnets.go
BenTheElder: This is for completeness. What if the user edits the configmap?
Casting to byte slices is nearly free.

aojea: good point, I imagine that other things will be affected but at least we are more or less safe

images/kindnetd/cmd/kindnetd/subnets.go
BenTheElder: Kubeadm config is beta.
Even if they weren't stae, we'd just make a new version of kindnetd when they change.

> if there were changes we should keep the previous environment variable configuration for backwards compatibility

This is for comparability with the kind binary only, when templating. If we are in fact only doing podSubnet it makes more sense, but even then we don't need it.

images/kindnetd/cmd/kindnetd/subnets.go
BenTheElder: Sure, this is a carry over from that PR. I can remove looking at services in the morning. I also want to make some improvements to error handling in the route agent.

images/kindnetd/cmd/kindnetd/subnets.go
BenTheElder: (also it wasn't free before, but is in current go to the point the stdlib takes advantage of this to be simpler now, yamlify isn't free but the data is small and this makes us more robust to changes. It also ensures we know how the yaml is formatted, so we can match on a field being a line)

images/kindnetd/cmd/kindnetd/masq.go
BenTheElder: For the reconcile pattern: yes. If there are no desired subnets to not masquerade, we should sync the rules.

BenTheElder: Imagine for testing I want to disable this. Right now I could eg remove the kubeadm configmap.

images/kindnetd/cmd/kindnetd/subnets.go
neolit123: > also, you know if there were changes?

kubeadm has been exposing these fields since at least 1.9 / early-v1alpha1.

BenTheElder:
discussed some more with @aojea, given that we are only handling podsubnet and we already have a robust templating mechanism for that (with less moving parts), I'm inclined to cancel the entire idea for simplicity.

I don't think we'll need other fields from kubeadm config going forward.

I will file another PR with some different fixes I found while hacking on this.

Issue #949, opened 2019-10-14
infrequent `docker exec .... signal: broken pipe` during kubeadm init | kubeadm join


BenTheElder:
so far I've not once seen this locally, but we have this occasionally in CI, maybe a few times a day with many, many runs.

/assign
/lifecycle active

## Examples

https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/83914/pull-kubernetes-e2e-kind/1183813161560576002

https://github.com/kubernetes-sigs/kind/issues/928#issuecomment-541954010

https://github.com/kubernetes/kubernetes/issues/83903

BenTheElder:
per the stack trace, this is definitely failing in `LocalCmd.Run()`, so we're seeing a `signal: broken pipe` from `docker exec ...`.

currently I've exclusively found these with `kubeadm init`.

however in this example: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-kind-conformance-ipv6/1182717569350504448

you can also see `ERROR: command "docker exec --privileged kind-control-plane tar --hard-dereference -C /var/log -chf - ." failed with error: write |1: broken pipe` from `kind export logs`

everything else I can find involves `kubeadm init`

BenTheElder:
another example https://github.com/kubernetes/kubernetes/pull/83956#issuecomment-542277813

BenTheElder:
our docker is fairly old, https://github.com/kubernetes/test-infra/pull/14784

BenTheElder:
Still trying to get us on newer docker. Couple of pending test-infra PRs.

BenTheElder:
not much progress today -- infra wg this morning and then mitigating, debugging, and dealing with https://github.com/kubernetes/test-infra/pull/14812

now have https://github.com/kubernetes/test-infra/pull/14820 to see about using newer docker in slightly streamlined kind CI specific image experimentally.

BenTheElder:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-kind-conformance-parallel/1184482589986000898

here's an example with `kubeadm join`
> W1016 15:02:01.370] ERROR: failed to create cluster: failed to join node with kubeadm: command "docker exec --privileged kind-worker kubeadm join --config /kind/kubeadm.conf --ignore-preflight-errors=all --v=6" failed with error: signal: broken pipe

BenTheElder:
[...]

PR #992, opened 2019-10-22
[WIP] log untar errors

BenTheElder:
xref: https://github.com/kubernetes-sigs/kind/issues/949#issuecomment-544786355

### DIFF
--- a/pkg/internal/cluster/logs/logs.go
+++ b/pkg/internal/cluster/logs/logs.go
@@ -80,7 +80,6 @@ func Collect(nodes []nodes.Node, dir string) error {
 			// http://man7.org/linux/man-pages/man1/tar.1.html#RETURN_VALUE
 			`tar --hard-dereference -C /var/log -chf - . || (r=$?; [ $r -eq 1 ] || exit $r)`,
 		)
-
 		if err := exec.RunWithStdoutReader(cmd, func(outReader io.Reader) error {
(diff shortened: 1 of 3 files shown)


pkg/internal/cluster/logs/logs.go
```diff
@@ -130,6 +129,11 @@ func Collect(nodes []nodes.Node, dir string) error {
 // untar reads the tar file from r and writes it into dir.
 func untar(r io.Reader, dir string) (err error) {
 	tr := tar.NewReader(r)
+	defer func() {
```
aojea: is this really needed? is is possible this func returns without error?

BenTheElder: The WIP is to prevent merge. This PR is just being used to obtain debugging information versus trying to trigger this failure mode locally because CI will run many jobs at once for me :-)

BenTheElder: (also that's why it checks if err != nil)

BenTheElder:
/test all
of course now it won't happen 🙃

BenTheElder:
aha https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_kind/992/pull-kind-conformance-parallel-ipv6/1187907454936027145/

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/1014

aojea:
Why is the pipe closed?

PR #993, opened 2019-10-22
entrypoint: fix cgroups mounts

jieyu:
For each cgroup subsystem, Docker does a bind mount from the current
cgroup to the root of the cgroup subsystem. For instance:
  `/sys/fs/cgroup/memory/docker/<cid> -> /sys/fs/cgroup/memory`

This will confuse Kubelet and cadvisor and will dump the following error
messages in kubelet log:

```
summary_sys_containers.go:47] Failed to get system container stats for
".../kubelet.service"
```

This is because `/proc/<pid>/cgroup` is not affected by the bind mount.
The following is a workaround to recreate the original cgroup
environment by doing another bind mount for each subsystem.

### DIFF
--- a/images/base/entrypoint
+++ b/images/base/entrypoint
@@ -42,6 +42,31 @@ fix_mount() {
   mount --make-rshared /
 }
 
+fix_cgroup() {
+  echo 'INFO: fix cgroup mounts for all subsystems'
+  # For each cgroup subsystem, Docker does a bind mount from the current
+  # cgroup to the root of the cgroup subsystem. For instance:
+  #   /sys/fs/cgroup/memory/docker/<cid> -> /sys/fs/cgroup/memory
+  #
+  # This will confuse Kubelet and cadvisor and will dump the following error
+  # messages in kubelet log:
+  #   `summary_sys_containers.go:47] Failed to get system container stats for ".../kubelet.service"`
+  #
+  # This is because `/proc/<pid>/cgroup` is not affected by the bind mount.
+  # The following is a workaround to recreate the original cgroup
+  # environment by doing another bind mount for each subsystem.
+  MOUNT_TABLE=$(cat /proc/self/mountinfo)
+  DOCKER_CGROUP_MOUNTS=$(echo "${MOUNT_TABLE}" | grep /sys/fs/cgroup | grep docker)
+  DOCKER_CGROUP=$(echo "${DOCKER_CGROUP_MOUNTS}" | head -n 1 | cut -d' ' -f 4)
+  CGROUP_SUBSYSTEMS=$(echo "${DOCKER_CGROUP_MOUNTS}" | cut -d' ' -f 5)
+
+  echo "${CGROUP_SUBSYSTEMS}" |
+  while IFS= read -r SUBSYSTEM; do
+    mkdir -p "${SUBSYSTEM}${DOCKER_CGROUP}"
+    mount --bind "${SUBSYSTEM}" "${SUBSYSTEM}${DOCKER_CGROUP}"
+  done
+}
+
 fix_machine_id() {
   # Deletes the machine-id embedded in the node image and generates a new one.
   # This is necessary because both kubelet and other components like weave net
@@ -90,6 +115,7 @@ EOF
 # run pre-init fixups
 fix_kmsg
 fix_mount
+fix_cgroup
 fix_machine_id
 fix_product_name
 configure_proxy


images/base/entrypoint
```diff
@@ -42,6 +42,31 @@ fix_mount() {
...
 
+fix_cgroup() {
+  echo 'INFO: fix cgroup mounts for all subsystems'
+  # For each cgroup subsystem, Docker does a bind mount from the current
+  # cgroup to the root of the cgroup subsystem. For instance:
+  #   /sys/fs/cgroup/memory/docker/<cid> -> /sys/fs/cgroup/memory
+  #
+  # This will confuse Kubelet and cadvisor and will dump the following error
+  # messages in kubelet log:
+  #   `summary_sys_containers.go:47] Failed to get system container stats for ".../kubelet.service"`
```
BenTheElder: note that these are just cadvisor warnings, they should't affect anything?

images/base/entrypoint
```diff
@@ -42,6 +42,31 @@ fix_mount() {
...
+  #
+  # This is because `/proc/<pid>/cgroup` is not affected by the bind mount.
+  # The following is a workaround to recreate the original cgroup
+  # environment by doing another bind mount for each subsystem.
+  MOUNT_TABLE=$(cat /proc/self/mountinfo)
+  DOCKER_CGROUP_MOUNTS=$(echo "${MOUNT_TABLE}" | grep /sys/fs/cgroup | grep docker)
```
BenTheElder: we should be resilient to not running in docker, we're nearing support for eg podman.

images/base/entrypoint
jieyu: yes, they don't affect anything, but still annoying and spamming the kubelet logs.

images/base/entrypoint
jieyu: Hum, ok, i don't know that. I haven't tested podman... what do you think about adding some check (e.g., checking .dockerenv) and only run this for docker?

aojea: do we have to do this from the entrypoint? 
can we do this from https://github.com/kubernetes-sigs/kind/blob/master/pkg/internal/cluster/providers/docker/provision.go#L147?

the docker cgroup folder is based in the container id, right?

images/base/entrypoint
aojea: agree that removing the noise from the logs is nice, especially those kubelet ones

prometherion:
Thanks for your contribution @jieyu : actually, I had an issue with _KinD_ not enabling me to stop and restart a cluster.

1. `kind create cluster --name foo`
2. waiting for every component is up and running, finally `docker stop foo-control-plane`
3. docker start `foo-control-plane`
4. `foo-control-plane` container crashes with the following output on logs

```
Set hostname to <foo-control-plane>.
Failed to attach 1 to compat systemd cgroup /docker/4ee96a312380827704e110faea5e19b1369c41a416465b853a5577c12fbbb434/init.scope: No such file or directory
Failed to open pin file: No such file or directory
Failed to allocate manager object: No such file or directory
[!!!!!!] Failed to allocate manager object.
Exiting PID 1...
```

I just built my own node image with your `entrypoint` script updated and my issue is gone... so many thanks!

images/base/entrypoint
tao12345666333: Agree with @aojea . Moving these logics to provides/docker can be clearer.

aojea:
Based on this https://docs.docker.com/config/containers/runmetrics/ maybe is enough only mounting 3 folders after the container is created
https://github.com/kubernetes-sigs/kind/blob/85a46de118167a88a8e69d814ec5f5a7b8093ab0/pkg/internal/cluster/providers/docker/provision.go#L97

@BenTheElder ^^?

images/base/entrypoint
BenTheElder: it should happen before we start systemd, entrypoint is appropriate.
this is roughly the right approach, but we can make the bash slightly more clever.

I'll look into this more today, I had to handle some other pressing issues yesterday.

prometherion: Do you know if using the `cgroup-driver=systemd` as _kubeadm_ extra options could fix the issue?

[Doc](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#configure-cgroup-driver-used-by-kubelet-on-control-plane-node) here

BenTheElder: no, the driver should match the container runtime. it does.

https://kubernetes.io/docs/setup/production-environment/container-runtimes/#systemd

BenTheElder: kubelet and the CRI need to agree on this, they do. IIRC we tried using the systemd driver but it did not help, we can try this again, but it needs to be both kubelet and containerd.

BenTheElder: I think we can deal with podman later, however I've made some minor tweaks in additional commit to the bash

BenTheElder:
@prometherion I can confirm the restart issue. Currently that has not been supported for other reasons (search the issue tracker...) but we'd like to long term.

https://github.com/kubernetes-sigs/kind/pull/1013 rebases this, adds some minor fixups, and updates the images.

BenTheElder:
Merged (in rebased form with extra commits) in https://github.com/kubernetes-sigs/kind/pull/1013

Thanks!

prometherion:
@BenTheElder well, actually with the cgroup hotfix in the entrypoint script, the restart issue is gone... os it has been indirectly supported! :)

PR #995, opened 2019-10-22
truncate commit in version

BenTheElder:
xref discussion in https://github.com/kubernetes-sigs/kind/pull/988#discussion_r337336824

### DIFF
--- a/cmd/kind/version/version.go
+++ b/cmd/kind/version/version.go
@@ -35,7 +35,8 @@ func Version() string {
 		// if commit was set, add the + <build>
 		// we only do this for pre-release versions
 		if GitCommit != "" {
-			v += "+" + GitCommit
+			// NOTE: use 14 character short hash, like Kubernetes
+			v += "+" + truncate(GitCommit, 14)
 		}
 	}
 	return v
@@ -77,4 +78,11 @@ func NewCommand() *cobra.Command {
 		},
 	}
 	return cmd
+}
+
+func truncate(s string, maxLen int) string {
+	if len(s) < maxLen {
+		return s
+	}
+	return s[:maxLen]
 }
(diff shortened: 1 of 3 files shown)


go.sum
```diff
@@ -17,6 +17,7 @@ golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2 h1:VklqNMn3ovrHsnt90Pveol
 golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2/go.mod h1:djNgcEr1/C05ACkg1iLfiJU5Ep61QUkGW8qpdssI0+w=
 golang.org/x/crypto v0.0.0-20190611184440-5c40567a22f8 h1:1wopBVtVdWnn03fZelqdXTqk7U7zPQCb+T4rbU9ZEoU=
 golang.org/x/crypto v0.0.0-20190611184440-5c40567a22f8/go.mod h1:yigFU9vqHzYiE8UmvKecakEJjdnWj3jj499lnFckfCI=
+golang.org/x/net v0.0.0-20190404232315-eb5bcb51f2a3 h1:0GoQqolDA55aaLxZyTzK/Y2ePZzZTUrRacwib7cNsYQ=
```
BenTheElder: not sure why this changed? will see if CI verify checks complain

cmd/kind/version/version.go
```diff
@@ -35,7 +35,8 @@ func Version() string {
 		// if commit was set, add the + <build>
 		// we only do this for pre-release versions
 		if GitCommit != "" {
-			v += "+" + GitCommit
+			// NOTE: use 14 character short hash, like Kubernetes
+			v += "+" + truncate(GitCommit, 14)
```
BenTheElder: Git says 12 characters is suficcient for the linux kernel, but kubernetes uses 14 and this still seems pretty short
https://git-scm.com/book/en/v2/Git-Tools-Revision-Selection
https://github.com/kubernetes/kubernetes/blob/6bcfd75dd662625b2f92c41cc19fb0fd1cc5aa32/hack/lib/version.sh#L69

```
$ kind version
kind v0.6.0-alpha+cb1e6727f4d507 go1.13.3 linux/amd64
```
/shrug
we can always revisit and truncate more or less, this is trivial to change now.

go.sum
neolit123: LGTM, but not sure about this change too.

BenTheElder: CI seems fine with it, and it's just one hash so /shrug
I'll follow up with taking another look at ensuring verify tidies

neolit123: /lgtm

BenTheElder: https://github.com/kubernetes-sigs/kind/issues/996

PR #1000, opened 2019-10-22
Use klog for kindnetd #991

### LINKED ISSUE #991: use klog for kindnetd
**What should be cleaned up or changed**:

`images/kindnetd` should stop using `fmt` for logging. we should use `k8s.io/klog` instead.

**Why is this needed**:

to provide better logs consistently with other kubernetes system components

u5surf:
Fix #991 
I changed the logging tools `k8s.io/klog` instead of `fmt ` in `images/kindnetd`


BenTheElder:
/lgtm
/approve
working on some unrelated kindnetd fixes, will create a new revision in a follow up.

thanks for this!

PR #1003, opened 2019-10-23
Revamped base image

### LINKED ISSUE #1001: make necessary `/etc/kubernetes/manifests` in workers to avoid unnecessary error logs
**What happened**:

In workers, as the folder of `/etc/kubernetes/manifests` doesn't exist by default to cater for static pods, `kubelet` will keep throwing errors complaining `path does not exist, ignoring`.

```
# journalctl -f
...
Oct 23 07:35:58 kind-cluster-worker kubelet[508]: E1023 07:35:58.547619     508 file_linux.go:61] Unable to read config path "/etc/kubernetes/manifests": path does not exist, ignoring
...
```

The folder should be created by default to avoid such annoying warnings prompting up in `journalctl` logs.

Spin up a `kind` cluster and run:

```
$ docker exec --privileged -i kind-cluster-worker journalctl -f
```

Simply create that folder can stop such warnings:

```
docker exec --privileged -i kind-cluster-worker mkdir -p /etc/kubernetes/manifests
```

**Environment:**

- kind version: (use `kind version`): v0.5.1
- OS (e.g. from `/etc/os-release`): MacBook Pro

BenTheElder:
- fixes #1001 
- minimizes number of layers
  - makes the location of files we copy in extremely apparent
- updates comments
- simplifies logic
- updates base and node image


.gitignore
```diff
@@ -1,10 +1,10 @@
...
-bin/
-_output/
-_artifacts/
+/bin/
+/_output/
+/_artifacts/
 
 # used for the code generators only
-vendor/
+/vendor/
```
BenTheElder: > If there is a separator at the beginning or middle (or both) of the pattern, then the pattern is relative to the directory level of the particular .gitignore file itself. Otherwise the pattern may also match at any level below the .gitignore level.

https://git-scm.com/docs/gitignore

BenTheElder: (we were matching `images/base/files/usr/local/bin/`)

aojea: :man_shrugging:

images/base/Dockerfile
```diff
@@ -12,51 +12,53 @@
...
+ARG CONTAINERD_VERSION="v1.3.0-7-g0b43a311"
+# Configure CNI binaries from upstream
+ARG CNI_VERSION="v0.8.2"
+# Configure crictl binary from upstream
+ARG CRICTL_VERSION="v1.16.1"
 
-COPY clean-install /usr/local/bin/clean-install
-RUN chmod +x /usr/local/bin/clean-install
+# copy in static files (configs, scripts)
+COPY files/ /
```
aojea: This is a bit risky but simpler

BenTheElder: what is risky about it? 🙃

images/base/Dockerfile
```diff
@@ -12,51 +12,53 @@
...
+# The base image already has: ssh, apt, snapd, but we need to install more packages.
+# Packages installed are broken down into (each on a line):
 # - packages needed to run services (systemd)
-# - CRI (containerd)
 # - packages needed for kubernetes components
 # - misc packages kind uses itself
-# Then we cleanup (removing unwanted systemd services)
-# Then we disable kmsg in journald (these log entries would be confusing)
+# After installing packages we cleanup by:
+# - removing unwanted systemd services)
```
aojea: nit , parenthesis at the end

images/base/Dockerfile
```diff
@@ -12,51 +12,53 @@
...
 
-# Get dependencies
-# The base image already has: ssh, apt, snapd
-# This is broken down into (each on a line):
+# Install dependencies, first from apt, then from release tarballs.
+# NOTE: we use one RUN to minimize layers.
+#
+# First we must ensure that our util scripts are executable.
+#
+# The base image already has: ssh, apt, snapd, but we need to install more packages.
```
aojea: just curious now that you mention, should we add a TODO to try to reduce more the image size removing more packages?

images/base/Dockerfile
aojea: drop or delete a file that gets into the image accidentally

images/base/Dockerfile
BenTheElder: docker image layers don't work that way. a tombstone is added to the next layer if you delete something in a new layer.

aojea:
/lgtm
all comments are  nits

BenTheElder:
ref: https://github.com/kubernetes/kubernetes/pull/84212

aojea:
:thinking: 
```
[k8s.io] Security Context when creating a pod in the host PID namespace should show its pid in the host PID namespace [LinuxOnly] [NodeFeature:HostAccess] expand_more	19s
[sig-node] RuntimeClass should run a Pod requesting a RuntimeClass with a configured handler [NodeFeature:RuntimeHandler] expand_more	5m3s
[sig-node] RuntimeClass should run a Pod requesting a RuntimeClass with a configured handler [NodeFeature:RuntimeHandler] expand_more
```

BenTheElder:
also runtimeclass flakes, looking

BenTheElder:
/hold 
looks like we have an issue with containerd config maybe for the runtimeclass tests

BenTheElder:
> images/base/containerd-config.toml → images/base/files/etc/containerd/containerd-config.toml

should be:
> images/base/containerd-config.toml → images/base/files/etc/containerd/config.toml

BenTheElder:
/hold cancel
should be fixed, those tests failed because the containerd config file had the wrong name

images/base/Dockerfile
```diff
@@ -68,61 +70,34 @@ RUN clean-install \
...
-# TODO(bentheelder): doc why / what here
-ARG CNI_VERSION="v0.8.2"
-ARG CNI_BASE_URL="https://github.com/containernetworking/plugins/releases/download/"
-RUN export ARCH=$(dpkg --print-architecture | sed 's/ppc64el/ppc64le/' | sed 's/armhf/arm/') \
+    && containerd --version \
+    && systemctl enable containerd \
+ && echo "Installing crictl ..." \
+    && curl -fSL "https://github.com/kubernetes-sigs/cri-tools/releases/download/${CRICTL_VERSION}/crictl-${CRICTL_VERSION}-linux-${ARCH}.tar.gz" | tar xzC /usr/local/bin \
+ && echo "Installing CNI binaries ..." \
+    && export ARCH=$(dpkg --print-architecture | sed 's/ppc64el/ppc64le/' | sed 's/armhf/arm/') \
```
aojea: I think we don't need this duplicated line

aojea:
/hold

tests failing because of
>nlosetup: /tmp/local-driver-47eee502-026b-4205-9537-2d59c6050a67/file: failed to set up loop device: No such file or directory\ncommand terminated with exit code 1\n\nerror:\nexit status 1

images/base/Dockerfile
BenTheElder: sure, it's not harmful though

images/base/Dockerfile
BenTheElder: "drop or delete" ?
how are files getting into the image accidentally?

images/base/Dockerfile
aojea: I was thinking that `--squash` was more mature

BenTheElder: squash is experimental in the daemon with no sign of that changing. experimental is problem because that means it is not on by default in the daemon, so we can't use it.

BenTheElder:
that's a flake I think. the image does contain losetup, dd, and mkdir...
/hold cancel

BenTheElder:
will address nits in a follow up, none of them affect the image functionally, and I'd rather get the bug closed with improvements.

images/base/Dockerfile
aojea: I mean, if you add or remove one file in that folder `files/` accidentally, it will be included in the image and probably you won't notice. 
It can happen, it's difficult that something like that happens, but is it possible since we are adding the folder recursively and copying it to the root folder

images/base/Dockerfile
BenTheElder: i'd actually rather keep this, because it's used for this install, and costs very little to have.

images/base/Dockerfile
BenTheElder: it should not be difficult to notice. We should be reviewing changes to this directory regardless.

BenTheElder: adding or removing will show up in git

PR #1005, opened 2019-10-23
add some trivial retries to kindnetd

BenTheElder:
I'm exploring options for monitoring these better (nodes are a bit of a fun case...) but in the meantime let's not panic when we fail to list the nodes once ...


BenTheElder:
prow flake
/retest

PR #1012, opened 2019-10-25
Fix base image info in documentation.

dasmfm:
Base image was changed to ubuntu:19.10, so documentation should also be updated.


BenTheElder:
So this was the original design of it, but most of it is still up to date.

Let's just drop the version IE say `ubuntu`. It should also have a "the" before the image name.

PR #1014, opened 2019-10-26
use rsync to archive logs

### LINKED ISSUE #999: periodic errors exporting logs
we're seeing errors tarring up log files for exporting the logs
/assign 
/priority important-soon
/lifecycle active

xref previously: https://github.com/kubernetes-sigs/kind/issues/949

BenTheElder:
attempting to fix #999 

rsync should be more resilient to files changing, so let's stage the logs to a tempdir with rsync before we copy them out with tar.

I suspect we're getting bogus tar output due to the files changing, we know we've hit the file change throwing errors and are currently masking them out ...

this costs a little more, but it's not dramatic and I hope it will be more robust.

### DIFF
--- a/pkg/exec/helpers.go
+++ b/pkg/exec/helpers.go
@@ -110,10 +110,11 @@ func RunWithStdoutReader(cmd Cmd, readerFunc func(io.Reader) error) error {
 	}()
 
 	err = cmd.Run()
+	pw.Close()
+	err2 := <-errChan
 	if err != nil {
 		return err
 	}
-	err2 := <-errChan
 	if err2 != nil {
 		return err2
 	}
@@ -137,10 +138,11 @@ func RunWithStdinWriter(cmd Cmd, writerFunc func(io.Writer) error) error {
 	}()
 
 	err = cmd.Run()
+	pr.Close()
+	err2 := <-errChan
 	if err != nil {
 		return err
 	}
-	err2 := <-errChan
 	if err2 != nil {
 		return err2
 	}
--- a/pkg/internal/cluster/logs/logs.go
+++ b/pkg/internal/cluster/logs/logs.go
@@ -18,9 +18,12 @@ package logs
 
 import (
 	"archive/tar"
+	"fmt"
 	"io"
 	"os"
+	"path"
 	"path/filepath"
+	"strings"
 
 	"sigs.k8s.io/kind/pkg/cluster/nodes"
 	"sigs.k8s.io/kind/pkg/errors"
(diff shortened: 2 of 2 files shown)


BenTheElder:
/hold cancel

This seems to be well behaved so far.

pkg/internal/cluster/logs/logs.go
```diff
@@ -127,6 +113,46 @@ func Collect(nodes []nodes.Node, dir string) error {
...
+// dumpDir dumps the dir nodeDir on the node to the dir hostDir on the host
+func dumpDir(node nodes.Node, nodeDir, hostDir string) (err error) {
+	// make tempdir to rsync nodeDir into (rsync handles taking a snapshot better)
+	tmp, err := mktemp(node)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if rerr := node.Command("rm", "-rf", tmp).Run(); rerr != nil && err == nil {
+			err = rerr
```
aojea: if there is an error here, it will be passed to the calling function?
 reading https://golang.org/ref/spec#Defer_statements, `If the deferred function has any return values, they are discarded when the function completes`

pkg/exec/helpers.go
```diff
@@ -110,10 +110,11 @@ func RunWithStdoutReader(cmd Cmd, readerFunc func(io.Reader) error) error {
 	}()
 
 	err = cmd.Run()
+	pw.Close()
```
aojea: this is smart, if we close it here we can't write thus we can't break the pipe, right?

pkg/exec/helpers.go
```diff
@@ -110,10 +110,11 @@ func RunWithStdoutReader(cmd Cmd, readerFunc func(io.Reader) error) error {
...
+	err2 := <-errChan
```
aojea: why moving it here? I guess it can block the func if `err!=nil`, can't it?

pkg/exec/helpers.go
BenTheElder: Er.
So this is an os.Pipe, which would make more sense if we didn't have the multiwriter thing underneath etc...

Anyhow, it should get closed when the process exits if it were a real pipe, go would have passed it directly as an fd and closed this process's duplicate immediately after starting the process, retaining only the end this process uses.

We should close it here to ensure the reader can't hang on a still open pipe waiting for more.

pkg/internal/cluster/logs/logs.go
BenTheElder: Yes, but only if we didn't already have an error

pkg/exec/helpers.go
BenTheElder: Yes, we should be waiting. We didn't wait before because it could hang. That _should_ be noticed and caught because it means a bug. In this case the pipe still being open.

This was the original impl (wait for both) and then we had a quick fix the prevent a hang by moving it 😛

Compare to os/exec.

PR #1015, opened 2019-10-28
Mount ssl certs folder into go build container

matthyx:
This might be needed behind a corporate proxy that intercepts HTTPS traffic with self-generated certificates to verify the root CA.

### DIFF
--- a/hack/go_container.sh
+++ b/hack/go_container.sh
@@ -82,6 +82,8 @@ run_in_go_container() {
       -e GO111MODULE -e GOPROXY -e CGO_ENABLED -e GOOS -e GOARCH \
     `# pass through proxy settings` \
       -e HTTP_PROXY -e HTTPS_PROXY -e NO_PROXY \
+    `# use host certificates` \
+      -v "/etc/ssl/certs:/etc/ssl/certs" \
     `# run the image with the args passed to this script` \
       "${GOIMAGE}" "$@"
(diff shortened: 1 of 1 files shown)


hack/go_container.sh
```diff
@@ -82,6 +82,8 @@ run_in_go_container() {
       -e GO111MODULE -e GOPROXY -e CGO_ENABLED -e GOOS -e GOARCH \
     `# pass through proxy settings` \
       -e HTTP_PROXY -e HTTPS_PROXY -e NO_PROXY \
+    `# use host certificates` \
+      -v "/etc/ssl/certs:/etc/ssl/certs" \
```
aojea: I think this is OS-specific, not all distros have the  certs in the same folder IIRC
Also, I think  the mount should be read-only to avoid possible issues

aojea: maybe adding it as an env variable? CERT_DIR?

BenTheElder: the OS specific ones may not be bind mountable anyhow (eg on mac)

It might be best to use a custom `GOIMAGE` instead. I'm not sure if there's a "best practice" for this though.

matthyx:
> I'm not sure if there's a "best practice" for this though.

Yeah agreed, I think for just 0.1% of the case it's not worth the complexity...

Issue #1017, opened 2019-10-28
Unable to update the default storage class on Kind


reshdebug:
**What you expected to happen**:
Not able update  default storage class. Expected to set, is-default-class":"false" for standard storage class

**How to reproduce it (as minimally and precisely as possible)**:

```
$ kubectl patch storageclass standard -p '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"false"}}}'
storageclass.storage.k8s.io/standard patched

```

```
$ kubectl describe sc standard
Name:            standard
IsDefaultClass:  Yes
Annotations:     kubectl.kubernetes.io/last-applied-configuration={"apiVersion":"storage.k8s.io/v1","kind":"StorageClass","metadata":{"annotations":{"storageclass.beta.kubernetes.io/is-default-class":"true"},"labels":{"addonmanager.kubernetes.io/mode":"EnsureExists"},"name":"standard"},"provisioner":"kubernetes.io/host-path"}
,storageclass.beta.kubernetes.io/is-default-class=false
Provisioner:           kubernetes.io/host-path
Parameters:            <none>
AllowVolumeExpansion:  <unset>
MountOptions:          <none>
ReclaimPolicy:         Delete
VolumeBindingMode:     Immediate
Events:                <none>


```
```
$ kubectl get sc
NAME                 PROVISIONER                            AGE
standard (default)   kubernetes.io/host-path                3h14
```

**Anything else we need to know?**: 
Did try to refer existing bug #66222, no luck!

**Environment:** 

- kind version: (use `kind version`): v0.5.1
- Kubernetes version: (use `kubectl version`):v1.15.3
- Docker version: (use `docker info`): Docker version 18.09.3
- OS (e.g. from `/etc/os-release`): Debian GNU/Linux 10

BenTheElder:
I'm pretty sure you need to create another storage class first.
kind does nothing special here beyond installing one.

BenTheElder:
by 66222I assume you mean https://github.com/kubernetes/kubernetes/issues/66222#issuecomment-449121999, which leads to this change, https://github.com/kubernetes/kubernetes/pull/66235/files

kind is already using EnsureExists, but it also doesn't deploy the addon manager.

reshdebug:
Did try creating new storage class and updated the Standard storage class but same behavior. In fact, I have 2 default storage class:

`kubectl get sc
NAME                  PROVISIONER                            AGE
[...]

PR #1018, opened 2019-10-28
update storageclass annotation to GA

BenTheElder:
fixes https://github.com/kubernetes-sigs/kind/issues/1017

also removes a pointless annotation from upstream

### DIFF
--- a/pkg/internal/cluster/create/actions/installstorage/storage.go
+++ b/pkg/internal/cluster/create/actions/installstorage/storage.go
@@ -71,7 +71,7 @@ metadata:
   namespace: kube-system
   name: standard
   annotations:
-    storageclass.beta.kubernetes.io/is-default-class: "true"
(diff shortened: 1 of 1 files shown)


aojea:
is this backward compatible?

aojea:
/lgtm
1.13 
https://github.com/kubernetes/kubernetes/pull/68345/commits/f167413661347e399cc7a2b2b89910a5a744833d

BenTheElder:
not 1.13, 1.13 is when the addons were updated, the actual annotation was https://github.com/kubernetes/kubernetes/commit/3beee8da90c19ef4ee99367c9823ba48ecd36a06 which was in v1.7 ... long before kind

BenTheElder:
we just cargo culted the entire storage class from the addons which hadn't been updated because the beta annotation is still supported.

PR #1023, opened 2019-10-29
remove docker and tempdir from bootstrapping

BenTheElder:
this is a small but critical step to finally decoupling the codebase to allow multiple providers

this removes the last point in cluster creation that uses the docker package instead of a provider


BenTheElder:
I've tested this locally.

BenTheElder:
removed unused parameter

Issue #734, opened 2019-07-23
Internal APIs should not be exposed


BenTheElder:
**What happened**:

- The internal API version is both public and expected to be used when using kind as a library. Neither of these should be true.
- Basically all of our library code is unstable and importable. We're running into the same problem as Kubernetes here where people are depending on packages that are not intended for public consumption (silly example: the CLI spinner). We need to clarify this so we can actually follow semver with our public APIs as we move to beta.

**What you expected to happen**:

- Public APIs should be intended for consumption and moving towards stability.

**Anything else we need to know?**:  @munnerz and I have been discussing what to do about this for a while. Recent uptick in references to kind as a library suggests we should actually fix this.

/assign
/priority important-soon

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/737 was a big fix for this

BenTheElder:
Next up: The public types package should contain _all_ of the types, no importing types from other packages. This is hostile to the library user (can't see the API types when looking at the docs without hopping to another file just to read one type), and couples the public API versions to types controlled by non-API packages.

BenTheElder:
patch types were moved into the config types spec.

BenTheElder:
"CRI"-ish types still need to be moved inline to the config versions. Waiting till after I land the bigger container interaction refactor...

BenTheElder:
will handle in refactors after https://github.com/kubernetes-sigs/kind/pull/1023 lands

BenTheElder:
CRI part is done in #1023 
remaining major public API fixup is in #1025

BenTheElder:
this is more or less done

PR #1024, opened 2019-10-29
move mount / portmapping into config types

BenTheElder:
part of #734 
this will also make the documentation easier, as all the config types will be in the config ...

this is a quick port, we'll want to go back over this and cleanup the comments possibly

we also probably want to consider only using yaml and consider just using the on-disk format (IE strings for the enums) instead of the integers we currently use.

those can be handled at a later date though.


pkg/apis/config/v1alpha3/json.go
```diff
@@ -0,0 +1,109 @@
+/*
```
BenTheElder: this is a copy of pkg/container/cri/json.go

pkg/apis/config/v1alpha3/yaml.go
```diff
@@ -0,0 +1,86 @@
+/*
```
BenTheElder: this is a copy of pkg/container/cri/yaml.go

pkg/internal/apis/config/types.go
```diff
@@ -130,3 +126,110 @@ type PatchJSON6902 struct {
 	// Patch should contain the contents of the json patch as a string
 	Patch string
 }
+
+// Mount specifies a host volume to mount into a container.
```
BenTheElder: these are copied from pkg/container/cri/types.go, but the struct tags have been removed (for the internal API version)

pkg/apis/config/v1alpha3/types.go
```diff
@@ -140,3 +136,115 @@ type PatchJSON6902 struct {
 	// Patch should contain the contents of the json patch as a string
 	Patch string `yaml:"patch" json:"patch"`
 }
+
+/*
```
BenTheElder: these are copied from pkg/container/cri/types.go

BenTheElder:
I've tested this locally. It also contains no functional changes other than the trivial conversion, the rest is just moving / copying types.

neolit123:
LGTM, should `pkg/container/cri` be removed at this point?

BenTheElder:
that will be next, the problem is we still use this and the docker package directly in build/, that PR may  be a bit bigger

BenTheElder:
er pkg/build/node uses pkg/container/docker and pkg/container/cri.Mount

I'm preparing a bigger PR to clean that up and make providers viable.

PR #1025, opened 2019-10-30
Go API cleanup

BenTheElder:
This cleans up a lot of the public API (notably not pkg/build)

For `pkg/cluster`:
- rename `Context` to `Provider`
  - more accurately reflects the updated usage
  - leaves room to use standard library context in the future
  - do NOT take a cluster name when creating a provider, instead accept it as the first parameter to `Create`, `Delete`, etc.
    - internal code still has a "Context" with the name. We can rethink that in a future PR.
  - add Options pattern to `NewProvider` leaving room to select which backend etc. in the future
- move `cluster.List()` onto `Provider`
- drop `cluster.IsKnown` which is redundant to `Provider.ListNodes(name)`

For `pkg/container/...`
- move the two tiny docker methods we use in `cmd/kind/load/docker-image/` inline
- move `pkg/container/...` under the remaining user, `pkg/build/node/internal/...`
  - we'll want to clean up node image build in a future PR. thinking about how best to approach that

For `pkg/util/concurrent`
- fold into the new `pkg/errors` which generally has error management code, these are mostly about handling errors when running things concurrently.
  - this reduces the number of public packages we / users need to think about, but does NOT reduce the power available to `cmd/` / users.


BenTheElder:
fixes: #734

BenTheElder:
The first and second commits are most important, providing a better mapping onto providers.

These are both relevant to #154, in particular:

- `List()` no longer magically always using the docker provider, moving that up to the `NewProvider` layer where it will be easy to change in a future PR 
- moving the hacky remnants of the public docker specific packages in favor of the `Node` and `Provider` APIs which are not docker specific

PR #1026, opened 2019-10-30
enable relative mounts in docker provider

BenTheElder:
fixes #910 

This is in the docker provider since it's a docker limitation that absolute paths are required.

This does not validate, because docker will create the mount location if it does not exist, it also means the config is consistently validated across hosts and more easily tested.

I also tweaked the error wrapping.

cc @mauilion @amwat


BenTheElder:
/hold cancel
confirmed that `filepath.Abs` internally does a `filepath.IsAbs` pre-check internally that ensures that we won't fail on existing abs paths due to some error probing the filesystem.

Issue #850, opened 2019-09-11
RFC: CLI UX Improvements


BenTheElder:
Recently we landed some changes to revamp the logging / debugging experience and to ensure that status messages can be suppressed (`-q` / `--quiet`). However, I think kind still has plenty to improve on this front.

In particular I propose the following changes to improve the usability of `kind`, some of which may be slightly breaking:

## Drop `export KUBECONFIG=$(kind get kubeconfig-path)`

The `kind create cluster` command should accept a `--kubeconfig` flag and the environment variable `KUBECONFIG` to determine which file to use in for the export cluster kubeconfig. We should merge the generated cluster config into this file and set the default context to point to the newly created cluster.

### Existing Feature Requests
-  https://github.com/kubernetes-sigs/kind/issues/802 "Update kubeconfig when creating cluster"
- https://github.com/kubernetes-sigs/kind/issues/113 "Add option to specify kubeconfig when creating kind clusters")
- https://github.com/kubernetes-sigs/kind/issues/112 "kind create cluster creates same kubeconfig user name for each cluster"

### PROS:
- Users are already used to this with `kops`, `minikube`, `gcloud` (GKE), ...
- Less effort to use (`kind create cluster` is now the entire creation flow)
- Easier to switch shells. 
- Works out of the box on every OS.

### CONS:
- Need to be more explicit for multicluster testing. HOWEVER if you set `KUBECONFIG` _before_ `kind create cluster`
- Slight complexity increase in kind
- client-go direct dependency. 
  - We already have an indirect dependency via kustomize
  - I think this can be mitigated by documenting `replace` using in `go.mod` for library consumers, and avoiding depending on newer client-go features. Worst case we re-implement or fork the bits we need
- This is a breaking change
  - This will be simple to update though:
     - Just export `KUBECONFIG` with your preferred path _before_ calling `kind create cluster` instead of after. If you mimic the old path behavior this will work with old and new kind versions. Some users already hardcoded these paths in various places.
     - we can leave `kind get kubeconfig-path` for a migration period but accept the same inputs and point to the path we'd choose in `kind create cluster`
  - `v0.6` will already more or less break the CLI slightly to implement the revamped logging

## Support `KIND_NAME` environment variable

We should allow re-using entire kind scripts without the `--name` flag for each command by also accepting a `KIND_NAME` environment variable to specify the cluster name if the `--name` flag is not specified.

### Existing Feature Requests

[...]

Issue #113, opened 2018-11-14
Add option to specify kubeconfig when creating kind clusters


font:
The `kind create cluster` command creates a `~/.kube/kind-config-<name>` kubeconfig file to access that kind cluster. When working with multiple clusters, you end up with multiple of these files. Could we add an option to the `kind create cluster` command (e.g. `--kubeconfig`) so that the user can provide a file to use for all the kubeconfigs such as `~/.kube/config`. I've also filed #112, which would need to be resolved first.

BenTheElder:
We've also debated just writing to this in the first place like most deployers, cc @Katharine @munnerz @justinsb 

We can definitely add a `--kubeconfig` as well
/assign

font:
> We've also debated just writing to this in the first place like most deployers

Yeah I think that would be consistent with most other tools. Then the `--kubeconfig` option could be used to implement the existing default behavior i.e. separate `kind-config-<name>` files.

BenTheElder:
Yeah, I just dislike that this seems racy (writing to the shared file) but it seems to be done by like, every other deployer in existence, so I guess that's OK.

We'd also have to consider if deleting the cluster should remove those then, since before we were going to just implement deleting the file.

BenTheElder:
https://github.com/kubernetes-sigs/kind/issues/112 seems less trivial than initially expected, so far I think users have ultimately been pretty fine with how the kubeconfig works and it's been reliable.

BenTheElder:
https://github.com/kubernetes-sigs/kind/issues/850

aojea:
/close

fixed by #1029

PR #1029, opened 2019-10-30
kubeconfig overhaul

BenTheElder:
Implements most of #850, also #112, #113

This is smaller than it looks, a lot of the lines are boilerplate license etc. because I broke up the implementation into small portions and tried to comment it pretty strongly

There are a few additional not strictly necessary but nice to have features like ensuring that the formatting matches client-go generated kubeconfigs.

This does:
- reading in kubeconfig in a structured way with support for (de)serializing arbitrary ignored fields
  - modifying server field in a structured way instead of regex
- ensuring kind entries are unique (`kind-${cluster_name}` instead of the ~static kubeadm names) #112 
- selecting the kubeconfig file to use in the same way as `kubectl` #113 
- merging in kubeconfig on `kind create cluster`
- removing kind entries from kubeconfig in `kind delete cluster`
- drops `kind get kubeconfig-path` as it's no longer relevant
- fixes `kind get kubeconfig` to not depend on any files on the host

This is a mildly breaking change. We'll want a simple guide in the release notes for migration.

If you want kind to use a different config file, you can either set the `--kubeconfig` option to `kind create cluster` / `kind delete cluster` or you can export `KUBECONFIG` prior to using kind.

This is basically the same as before, just that you decide the path and you do it _before_ `kind create cluster` instead of after.


BenTheElder:
important reference for expected behavior: https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands

in addition, from inspecting code (mainly client-go directly) and reading some other sources
- for the case of `KUBECONFIG` being a list of paths, empty paths are discarded, duplicate paths are discarded, THEN the first existing one OR the last file is used
- path+".lock" should be used to 'lock' modifications, we've lifted the exact little lockfile creation / deletion routines from client-go
- merging clusters into kubeconfig universally sets the current context (kops, gcloud, etc. ...)
- removing clusters from kubeconfig universally unsets the current context if the context refers to the cluster being removed

pkg/internal/util/kubeconfig/remove.go
```diff
@@ -0,0 +1,110 @@
...
+	"sigs.k8s.io/kind/pkg/errors"
+)
+
+// RemoveKIND removes the kind cluster kindClusterName from the KUBECONFIG
+// files at configPaths
+func RemoveKIND(kindClusterName string, explicitPath string) error {
+	configPaths := paths(explicitPath, os.Getenv)
+
+	// lock all the files for modifying
+	sort.Strings(configPaths)
```
amwat: This sort seems unnecessary?

pkg/internal/util/kubeconfig/remove.go
```diff
@@ -0,0 +1,110 @@
...
+	for _, configPath := range configPaths {
+		if err := lockFile(configPath); err != nil {
+			return errors.Wrap(err, "failed to lock config file")
+		}
+		defer func(configPath string) {
```
amwat: we should probably just process each file individually instead of locking all files together. 
locak -> remove -> unlock.

pkg/internal/util/kubeconfig/merge.go
```diff
@@ -0,0 +1,100 @@
...
+	// verify assumptions about kubeadm / kind kubeconfigs
+	if err := checkKubeadmExpectations(kind); err != nil {
+		return err
+	}
+
+	// insert or append cluster entry
+	shouldAppend := true
+	for i := range existing.Clusters {
+		if existing.Clusters[i].Name == kind.Clusters[0].Name {
+			existing.Clusters[i] = kind.Clusters[0]
```
amwat: do we want to warn that we are overwriting an existing entry ?

BenTheElder: I don't think so, though maybe we should explicitly overwrite based on the cluster key though

pkg/internal/cluster/kubeconfig/kubeconfig.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	b, err := kubeconfig.Encode(cfg)
+	if err != nil {
+		return "", err
+	}
+	return string(b), err
+}
+
+func Context(kindClusterName string) string {
+	return kubeconfig.KINDClusterKey(kindClusterName)
```
aojea: is this passing the linter with all CAPS?

pkg/internal/util/kubeconfig/remove.go
BenTheElder: yeah this is a carry over from client go. there's some hand waving about avoiding deadlocks.

https://github.com/kubernetes/kubernetes/blame/48bc59a997a3490398d0d234d4f60d3121db9b74/staging/src/k8s.io/client-go/tools/clientcmd/config.go#L161

pkg/internal/cluster/kubeconfig/kubeconfig.go
BenTheElder: yes, acronyms are same case

BenTheElder: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/1029/pull-kind-verify/1189674919814762498

pkg/internal/util/kubeconfig/encode_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+package kubeconfig
+
+import (
+	"testing"
+
+	"sigs.k8s.io/kind/pkg/internal/util/assert"
+)
+
+func TestEncodeRoundtrip(t *testing.T) {
```
aojea: can we add a test failing?

pkg/internal/util/kubeconfig/remove.go
BenTheElder: I haven't found a way to see the history behind this yet, as it predates staging.

pkg/internal/util/kubeconfig/encode_test.go
BenTheElder: not trivially, it only fails if it cannot marshal the type so ...

BenTheElder: FYI we have 41.5% statements covered in this package, the rest are mostly error cases that aren't really triggerable like this, or file system interactions (and somewhat unit tests then, but I will probably still add some)

pkg/internal/util/kubeconfig/remove.go
BenTheElder: done

pkg/internal/util/kubeconfig/remove.go
BenTheElder: removed

pkg/internal/util/kubeconfig/merge.go
BenTheElder: OTOH this method is internal and the second config is expected to be a kind config

BenTheElder: (and internally documented as such)

pkg/internal/cluster/create/create.go
```diff
@@ -177,41 +177,3 @@ func collectOptions(options ...create.ClusterOption) (*createtypes.ClusterOption
 
 	return opts, nil
 }
-
-func printUsage(name string) {
```
aojea: are we missing this message?

pkg/internal/util/kubeconfig/write.go
```diff
@@ -0,0 +1,45 @@
...
+// it will ensure the directories in the path if necessary
+func write(cfg *Config, configPath string) error {
+	encoded, err := Encode(cfg)
+	if err != nil {
+		return err
+	}
+	// NOTE: 0755 / 0600 are to match client-go
+	dir := filepath.Dir(configPath)
+	if _, err := os.Stat(dir); os.IsNotExist(err) {
+		if err = os.MkdirAll(dir, 0755); err != nil {
```
aojea: this creates a folder, do we clean it if fails or if we no longer need it?

BenTheElder: no:
- it ensures an entire path of directories and we don't know which were created
- we intend to have persistent storage there
- client-go does not do this

pkg/internal/cluster/create/create.go
BenTheElder: not really, though we did move it off of writing via fmt.Printf which was bad.

I'm going to add a small tweak to it to be closer to before. see above in the code

pkg/internal/util/kubeconfig/encode.go
```diff
@@ -0,0 +1,64 @@
...
+func normYaml(y []byte) ([]byte, error) {
+	var unstructured interface{}
+	if err := kubeyaml.Unmarshal(y, &unstructured); err != nil {
+		return nil, err
+	}
+	encoded, err := kubeyaml.Marshal(&unstructured)
+	if err != nil {
+		return nil, err
+	}
+	// special case: don't write anything when empty
```
aojea: maybe we can detect this first?

BenTheElder: not really, the input is fairly arbitrary.

this is a really simple nice to have thing to make the output more like the user would expect. it would be totally valid to not have this entire method.

pkg/internal/cluster/create/create.go
BenTheElder: I added a more detailed output, including --kubeconfig if an explicit --kubeconfig was supplied. see above.

BenTheElder:
brought coverage up to 59.4% of statements for the kubeconfig package by covering `WriteMerged` with writing in tempdirs.

BenTheElder:
ok covered removal. this is now at 64.7% statements covered.
the non-covered statements are largely:
- the homedir implementation copied from client-go which has a large amount of windows specific code. /shrug we won't be modifying this, only keeping in sync with upstream
- error branches that are technically possible but won't really happen (eg failing to marshal a kubeconfig)

BenTheElder:
more specific coverage: 
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/1029/pull-kind-unit/1189704250880430080
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/1029/pull-kind-unit/1189704250880430080/artifacts/filtered.html#file96

pkg/internal/util/kubeconfig/paths.go
```diff
@@ -0,0 +1,93 @@
...
+
+/*
+paths returns the list of paths to be considered for kubeconfig files
+where explicitPath is the value of --kubeconfig
+
+Logic based on kubectl
+
+https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands
+
+- If the --kubeconfig flag is set, then only that file is loaded. The flag may only be set once and no merging takes place.
```
aojea: We should document this later in the quick start guide or the readme

aojea:
/lgtm
this is simply amazing, impressive work

pkg/internal/util/kubeconfig/paths.go
aojea: done https://github.com/kubernetes-sigs/kind/pull/1032

pkg/internal/cluster/create/actions/kubeadminit/init.go
```diff
@@ -135,42 +114,3 @@ func (a *action) Execute(ctx *actions.ActionContext) error {
 	ctx.Status.End(true)
 	return nil
 }
-
-// matches kubeconfig server entry like:
-//    server: https://172.17.0.2:6443
-// which we rewrite to:
-//    server: https://$ADDRESS:$PORT
-var serverAddressRE = regexp.MustCompile(`^(\s+server:) https://.*:\d+$`)
```
maelvls: Since the 'rewrite' doesn't (seem to) happen, the `kind get kubeconfig` prints the container's host and port instead of 127.0.0.1 + the forwarded port. Or am I wrong?

aojea: seems this is failing 
https://github.com/kubernetes-sigs/kind/pull/1029/files#diff-8e81c51b562c668c8c80420dd1917316R81
:/

can you open a bug  @maelvls ?

PR #1038, opened 2019-11-01
doc: Fix minor issue with kernel output path

PatrickLang:
I had a small mistake in the WSL2 kernel build steps. On my setup the same file gets placed in both `arch/x86/boot/bzImage` and `arch/x86_64/boot/bzImage`. Using either worked, but technically `x86_64` is correct so this fixes the doc to use it.


BenTheElder:
/lgtm
/approve
thanks!

PR #1039, opened 2019-11-01
add more test coverage

BenTheElder:
and other minor cleanup. eliminates the last `panic` in `kind`


BenTheElder:
this brings us up to 20% of statements and 30% of files https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/1039/pull-kind-unit/1190123013454958592

BenTheElder:
there's plenty of code that doesn't make much sense to unit test at the moment (spinners? loggers? :P) -- this shores up coverage on a lot of the bits that do.

pkg/internal/util/kubeconfig/paths.go
```diff
@@ -91,3 +92,76 @@ func discardEmptyAndDuplicates(paths []string) []string {
...
+// homeDir returns the home directory for the current user.
+// On Windows:
+// 1. the first of %HOME%, %HOMEDRIVE%%HOMEPATH%, %USERPROFILE% containing a `.kube\config` file is returned.
+// 2. if none of those locations contain a `.kube\config` file, the first of %HOME%, %USERPROFILE%, %HOMEDRIVE%%HOMEPATH% that exists and is writeable is returned.
+// 3. if none of those locations are writeable, the first of %HOME%, %USERPROFILE%, %HOMEDRIVE%%HOMEPATH% that exists is returned.
+// 4. if none of those locations exists, the first of %HOME%, %USERPROFILE%, %HOMEDRIVE%%HOMEPATH% that is set is returned.
+// NOTE this is from client-go. Rather than pull in client-go for this one
+// standalone method, we have a fork here.
+// https://github.com/kubernetes/client-go/blob/6d7018244d72350e2e8c4a19ccdbe4c8083a9143/util/homedir/homedir.go
+// We've modified this to require injecting os.Getenv and runtime.GOOS as a dependencies for testing purposes
```
neolit123: i think there were complains that this logic in client-go is not ideal, but i don't remember the PR/issue or any details.

neolit123: so this seems fine as a fork in kind.

pkg/internal/util/kubeconfig/write_test.go
```diff
@@ -0,0 +1,113 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+
+	"sigs.k8s.io/kind/pkg/internal/util/assert"
+)
+
+func TestWrite(t *testing.T) {
+	t.Parallel()
+	t.Run("non-existent file", testWriteNoExistingFile)
```
neolit123: inline the function here?
i don't prefer lambdas or closures, but it seems to be the testing pattern in go.

pkg/build/node/node.go
```diff
@@ -120,14 +124,24 @@ func NewBuildContext(options ...Option) (ctx *BuildContext, err error) {
...
-	bits, err := kube.NewNamedBits(ctx.mode, ctx.kubeRoot)
+	bits, err := kube.NewNamedBits(ctx.mode, ctx.kubeRoot, ctx.arch)
 	if err != nil {
 		return nil, err
 	}
 	ctx.bits = bits
 	return ctx, nil
 }
 
+func supportedArch(arch string) bool {
```
aojea: why is this change in this PR?
seems unrelated

aojea: ok, I've read the description now, scratch this comment 🙃

pkg/internal/util/kubeconfig/paths.go
BenTheElder: This 100% matches logic wise, we want to produce the same path. We already had this code, it just moved and got more testable.

pkg/build/node/node.go
BenTheElder: I went to test this package and refactored to eliminate the panic and simplify.

pkg/internal/util/kubeconfig/write_test.go
BenTheElder: Shrug, IMO this pattern is nicer for large test methods so you can more easily read what the cases are more quickly 😅
Functionally it's the same so.

pkg/internal/util/kubeconfig/paths.go
BenTheElder: FWIW the client-go logic is now pretty well defined to be $HOME, just with fallbacks on windows. This is locally internal because it's only the "homedir" for the purposes of kubeconfig. Otherwise we'd ask the go stdlib which added this in 1.12 https://golang.org/pkg/os/#UserHomeDir

PR #1041, opened 2019-11-01
Cleanup old / unused code

BenTheElder:
~depends on #1039~ rebased after merge


BenTheElder:
eliminates the CRI package and related code, old serialization logic, etc.
follow up to the provider PR(s).

pkg/build/node/internal/container/docker/run.go
```diff
@@ -17,77 +17,17 @@ limitations under the License.
...
-	}
-	// convert mounts to container run args
-	runArgs := o.RunArgs
-	for _, mount := range o.Mounts {
-		runArgs = append(runArgs, generateMountBindings(mount)...)
-	}
-	for _, portMapping := range o.PortMappings {
-		runArgs = append(runArgs, generatePortMappings(portMapping)...)
-	}
+func Run(image string, runArgs []string, containerArgs []string) error {
```
aojea: can we consolidate runArgs and containerArgs?

BenTheElder: no, they are different things.

container args are args to the container. runArgs are options to docker run itself. they must come before and after the image respectively.

BenTheElder: `docker run <flags> <image> <container args>`

PR #1043, opened 2019-11-01
minor update to base image docs

BenTheElder:
fixes https://github.com/kubernetes-sigs/kind/pull/1012

PR #1044, opened 2019-11-01
carry PR 807

BenTheElder:
we've had multiple attempts at this, so while I personally don't think this is an improvement, carrying it to at least be the same on both and to put one command to a line when we use code blocks

/cc @aojea 

fixes #807

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,13 +48,14 @@ mv ./kind /some-dir-in-your-PATH/kind
 ```
 
 To use kind, you will need to [install docker].  
-Once you have docker running you can create a cluster with 
+Once you have docker running you can create a cluster with:
 
 ```console
 kind create cluster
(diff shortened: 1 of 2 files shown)


BenTheElder:
Hmm netlify isn't showing

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/1043#issuecomment-548809384

aojea:
/lgtm

> we've had multiple attempts at this,

it's a cosmetic change that people seems to like ;)

BenTheElder:
Normally with N=2 I'd not be so easily swayed but Jim is on the docs team so :-)

BenTheElder:
Netlify is still unable to deploy due to upstream issues, but I rendered this locally.

PR #1046, opened 2019-11-02
start to treat commands as a [go] API

BenTheElder:
I aim to get both the go packages used by the commands and the commands themselves to be a stable API, in order to do that we should treat them the same.

Following the pattern from E.G. kubectl I've moved `cmd/kind/...` into `pkg/cmd/kind/...` as a first step.

As a result some initial follow up:
- break out `Main()` and `Run()` into `cmd/kind/app`
- add a main.go to `cmd/kind` to match the root main.go
- start to shuffle `pkg/internal` to more specific internal packages (eg the details of building kubernetes are node image build details) in order to avoid accidental dependencies in `pkg/cmd` which would now be allowed by go to import `pkg/internal/...` as a result of the move
  - while I was working on this, I improved the cluster create options API to not expose internal types in the signature and be more discoverable and explicit

I will write more changes to allow injecting the input / output

/cc @amwat @aojea


pkg/internal/cluster/kubeconfig/kubeconfig.go
```diff
@@ -23,7 +25,9 @@ import (
 	"sigs.k8s.io/kind/pkg/errors"
 	"sigs.k8s.io/kind/pkg/internal/cluster/context"
 
-	"sigs.k8s.io/kind/pkg/internal/util/kubeconfig"
+	// this package has slightly more generic kubeconfig helpers
+	// and minimal dependencies on the rest of kind
+	"sigs.k8s.io/kind/pkg/internal/cluster/kubeconfig/internal/kubeconfig"
```
aojea: too many internals?

BenTheElder: not at all 🙃

BenTheElder: the way go handles this is that for any given directory tree with `internal` as a segment, only packages rooted in the same directory can import it.

IE the package here can only be imported by packages with the prefix `sigs.k8s.io/kind/pkg/internal/cluster/kubeconfig/`

BenTheElder: this allows us to split up the code here into a tiering of unexported methods (the internal package, vs the unexported helpers) which is nicer than one big flat space of all lowercase helper methods.

PR #1048, opened 2019-11-04
Just printing error to standard output

### LINKED ISSUE #1047: cobra errors are not logged
**What happened**:

I tried to load images into my brand new KinD cluster but didn't know how `load docker-images` works, so I started typing and notice I was missing the first argument that has to be the Docker image to load.

Expecting to see the error `name of image is required` printed on _standard output_ but wasn't there.

Just issue `kind load docker-image`: nothing is printed.

Going to prepare little PR: just need to handle it at Cobra level with a simple `fmt.Println`

**Environment:**

- kind version: (use `kind version`): v0.6.0-alpha+0ff0546bc81543
- Kubernetes version: (use `kubectl version`): _not relevant_
- Docker version: (use `docker info`): _not relevant_
- OS (e.g. from `/etc/os-release`): _not relevant_

prometherion:
Closes #1047: printing errors to standard output as suggested also by Cobra [docs](https://github.com/spf13/cobra#create-rootcmd)

```golang
  if err := rootCmd.Execute(); err != nil {
    fmt.Println(err)
    os.Exit(1)
  }
```


BenTheElder:
Errors go to Stderr and pipeable output goes to stdout.

We already print the error?

BenTheElder:
Oh I see, these are errors from cobra ...

We should move logError to Run() rather than printing errors to stdout.

prometherion:
I don't think that's the best place where to put that logic:

```
func Main() {
	if err := Run(); err != nil {
		fmt.Println(err)
		os.Exit(1)
	}
}
```

We got already an `err != nil` there and `logError` is scoped to `root` available in `pkg/cmd/kind`.

Going to update PR with an additional commit in order to use the _logger_

BenTheElder:
It should be in run, Run is the main just without calling os.Exit.

We also need to move the logError from kind's root command. It's smarter about extracting error details and we don't want to log the error twice.

prometherion:
@BenTheElder please take a look to the last commit pushed: found a _typo_ regarding the use of `Args` vs `PreRunE`.

pkg/cmd/kind/load/docker-image/docker-image.go
```diff
@@ -42,15 +42,15 @@ type flagpole struct {
 func NewCommand() *cobra.Command {
 	flags := &flagpole{}
 	cmd := &cobra.Command{
-		Args: func(cmd *cobra.Command, args []string) error {
```
BenTheElder: This should have been valid, cobra uses this semantically to validate arguments

https://github.com/spf13/cobra/blob/77e4d5aecc4d34e58f72e5a1c4a5a13ef55e6f44/args.go#L7

cmd/kind/app/main.go
```diff
@@ -16,6 +16,10 @@ func Main() {
 
 // Run invokes the kind root command, returning the error.
 // See: sigs.k8s.io/kind/pkg/cmd/kind
-func Run() error {
-	return kind.NewCommand().Execute()
+func Run() (err error) {
+	err = kind.NewCommand().Execute()
+	if err != nil {
+		globals.GetLogger().Error(err.Error())
```
BenTheElder: this should still be `logError` moved from the root command

pkg/cmd/kind/load/docker-image/docker-image.go
prometherion: @BenTheElder I tested switching from `PreRunE` to `Args` and I cannot see logs. Give it a try on your own: doesn't work because it looks `Args` is executed &before* the persistent preRun :(

prometherion:
Honesly, I don't understand why we have `Run()` func in `cmd/kind/app/main.go` if it's just used by the `Main` func in the same pkg :)

Looks an useless overhead for me

BenTheElder:
> Honesly, I don't understand why we have Run() func in cmd/kind/app/main.go if it's just used by the Main func in the same pkg :)

This package is importable, and the intention is to grow stable support for import. `Main()` is the overhead because otherwise we'd be duplicating those lines in two files /shrug

prometherion:
> > Honesly, I don't understand why we have Run() func in cmd/kind/app/main.go if it's just used by the Main func in the same pkg :)
> 
> This package is importable, and the intention is to grow stable support for import. `Main()` is the overhead because otherwise we'd be duplicating those lines in two files /shrug

Well, if I want to import KinD I would use `pkg/cmd` instead but... nevermind, my bad I started to point out the focus from the PR :)

BenTheElder:
> Well, if I want to import KinD I would use pkg/cmd instead but... nevermind, my bad I started to point out the focus from the PR :)

you would be missing the error logging that we need to move to run :-)

we will also soon be injecting the dependencies here, but that change will need to rebase on this one :-)

prometherion:
> > Well, if I want to import KinD I would use pkg/cmd instead but... nevermind, my bad I started to point out the focus from the PR :)
> 
> you would be missing the error logging that we need to move to run :-)
> 
> we will also soon be injecting the dependencies here, but that change will need to rebase on this one :-)

If I'm importing the project, I don't want to get logging because I'm already dealing with a Go project, so pointers to errors are enough!

pkg/cmd/kind/load/docker-image/docker-image.go
prometherion: As you can see [here](https://github.com/spf13/cobra/blob/77e4d5aecc4d34e58f72e5a1c4a5a13ef55e6f44/command.go#L802-L804), `Args` validation is performed before `PersistentPreRunE`

BenTheElder: this problem applies to other subcommands, we need a more general solution

BenTheElder:
> If I'm importing the project, I don't want to get logging because I'm already dealing with a Go project, so pointers to errors are enough!

For your usage you case you can import `pkg/cmd/...`, however we flexibly support this at multiple layers, and we do provide injectable logging. Other users want the logging, we have this use case already.

The default unless you import the app will be a do-nothing logger.

I have another change in progress to remove the temporary `globals` package we used to migrate off of the previous log package, making this fully injectable.

pkg/cmd/kind/load/docker-image/docker-image.go
prometherion: Nope, Kind is just using the `Args` func for `load docker-images` and `load image-archive` sub-commands.

The unique exception is the `export logs` but `Args` there is optional:
https://github.com/kubernetes-sigs/kind/blob/aea9a47507085fe7bf89f7f65d31ead92876a42d/pkg/cmd/kind/export/export.go#L29-L30

prometherion:
> I have another change in progress to remove the temporary `globals` package we used to migrate off of the previous log package, making this fully injectable.

I can wait for it in order to rebase my changes since I guess the others are ok (please review also latest comment about review) or let's merge this and then solve the injectable logger.

pkg/cmd/kind/load/docker-image/docker-image.go
BenTheElder: `NoArgs` returns an error when there are unexpected arguments.

Cobra can also error in pflag parsing. We need an approach that can handle these. I think I've found that in https://github.com/kubernetes-sigs/kind/pull/1050

BenTheElder:
We moved error reporting and log setup out of cobra so we also handle flag parse errors and dependency inject the I/O / logger.

Thanks for reporting this and sending the patch.

PR #1050, opened 2019-11-04
Inject IO / logger

BenTheElder:
To make importing easier and decouple these.

Eliminates the `globals` package 🎉 

We'll probably need some follow up to this, but it's a big step.


aojea:
can we squash them before merge?

BenTheElder:
sure, I was mostly aiming to make it easier to read, prow can squash on
merge even.

BenTheElder:
prow will squash it now

aojea:
naive question, should we pass the logger as a pointer to avoid issues with concurrency or is safe this way?

BenTheElder:
the logger is concurrent safe, and interfaces are internally pointers.

BenTheElder:
/hold
CI output doesn't seem right.

aojea:
indeed
```
W1105 08:31:31.676] + NUM_NODES=2
W1105 08:31:31.676] + KIND_CREATE_ATTEMPTED=true
W1105 08:31:31.676] + kind create cluster --image=kindest/node:latest --retain --wait=1m -v=3 --config=/workspace/_artifacts/kind-config.yaml
W1105 08:31:32.199] Creating cluster "kind" ...
I1105 08:32:46.379]  â€¢ Ready after 0s ðŸ’š
W1105 08:32:48.164] + run_tests
W1105 08:32:48.165] + [ ipv4 = ipv6 ]
```

BenTheElder:
the problem with this currently is that it won't pick up the flags before errors with arguments, so `-q` isn't respected. investigating options.

BenTheElder:
we're just also handling `-q` / `--quiet` before we even invoke cobra, this is the only robust solution I found that didn't involve replacing cobra.

help output should work correctly.

BenTheElder:
other than squashing this should be fine, I think I'm going to let CI squash it though so you can review just the diff since last time 😅

BenTheElder:
/retest
looks like a flake

PR #1054, opened 2019-11-06
avoid tempfiles cleanup collision

### LINKED ISSUE #999: periodic errors exporting logs

BenTheElder:
... with exporting logs :upside_down_face: 

fixes #999

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -52,9 +52,13 @@ COPY files/ /
 #
 # Next we download and extract crictl and CNI plugin binaries from upstream.
 #
-# Finally we ensure the /etc/kubernetes/manifests directory exists. Normally
+# Next we ensure the /etc/kubernetes/manifests directory exists. Normally
 # a kubeadm debain / rpm package would ensure that this exists but we install
 # freshly built binaries directly when we build the node image.
+#
+# Finally we adjust tempfiles cleanup to be 1 minute after "boot" instead of 15m
+# This is plenty after we've done initial setup for a node, but before we are
+# likely to try to export logs etc.
 RUN echo "Ensuring scripts are executable ..." \
     && chmod +x /usr/local/bin/clean-install /usr/local/bin/entrypoint \
  && echo "Installing Packages ..." \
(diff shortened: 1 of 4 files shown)


images/base/Dockerfile
```diff
@@ -91,7 +95,9 @@ RUN echo "Ensuring scripts are executable ..." \
     && tar -C /opt/cni/bin -xzf /tmp/cni.tgz \
     && rm -rf /tmp/cni.tgz \
  && echo "Ensuring /etc/kubernetes/manifests" \
-    && mkdir -p /etc/kubernetes/manifests
+    && mkdir -p /etc/kubernetes/manifests \
+ && echo "Adjusting systemd-tmpfiles timer" \
+    && sed -i /usr/lib/systemd/system/systemd-tmpfiles-clean.timer -e 's#OnBootSec=.*#OnBootSec=1min#'
```
BenTheElder: the default is 15m after boot and then every 24 hours.
for now just moving it to 1m after boot which imo makes more sense for kind nodes which generally live on a compressed times cale, and leaving the 24 hours alone.

the real fix is hopefully in the /var/tmp change :upside_down_face:

BenTheElder:
/assign @jeefy 
/uncc @krzyzacy 
[OOO, gracious review offer :-)]

pkg/build/node/node.go
```diff
@@ -43,7 +43,7 @@ import (
 const DefaultImage = "kindest/node:latest"
 
 // DefaultBaseImage is the default base image used
-const DefaultBaseImage = "kindest/base:v20191030-ae10d883@sha256:bccb2ad3686a11ea920b5eaf701b252166affa3c3362e38c549374f1e467ae80"
+const DefaultBaseImage = "kindest/base:v20191105-ee880e9b@sha256:f4c7c8ac0d387a18ba0bd2756b17551102b7322497796e3ad3f366c3e67c01d2"
```
BenTheElder: built with the base image dockerfile tweak above

PR #1057, opened 2019-11-06
bikeshed quiet handling and add tests

BenTheElder:
mostly wanted to write some tests because this is a little strange but also eliminated a useless intermediate variable and just inlined declaring the flag.

we shouldn't change it except maybe the usage which doesn't really matter in `checkQuiet`, and having it as an exported API was kinda 👎

### DIFF
--- a/cmd/kind/app/main.go
+++ b/cmd/kind/app/main.go
@@ -41,8 +41,7 @@ func Main() {
 // See: sigs.k8s.io/kind/pkg/cmd/kind
 func Run(logger log.Logger, streams cmd.IOStreams, args []string) error {
 	// NOTE: we handle the quiet flag here so we can fully silence cobra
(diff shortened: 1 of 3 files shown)


BenTheElder:
~1% more unit test (statement, file) coverage 🙃 
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/1057/pull-kind-unit/1191960663132475394

BenTheElder:
/test pull-kind-unit
(getting updated results, tests are all passing)

BenTheElder:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/1057/pull-kind-unit/1191991374711361537

Issue #1060, opened 2019-11-06
KUBECONFIG management migration for kind v0.6.0+


BenTheElder:
## `KUBECONFIG` management for kind `v0.6.0+` versus previous releases:

If you just want to use kind for local development, you can skip the `export KUBECONFIG="$(kind get kubeconfig-path --name=kind)"` step.

KIND now exports / merges kubeconfigs in the same way that kops, minikube, and other tools do. This means that when switching terminals kind will "just work" without remembering to export the `KUBECONFIG` environment variable in each terminal.

The file selection logic follows `kubectl` and client-go, which is roughly:

> The loading order follows these rules:
>
> 1. If the --kubeconfig flag is set, then only that file is loaded. The flag may only be set once and no merging takes place.
> 2. If $KUBECONFIG environment variable is set, then it is used as a list of paths (normal path delimiting rules for your system). These paths are merged. When a value is modified, it is modified in the file that defines the stanza. When a value is created, it is created in the first file that exists. If no files in the chain exist, then it creates the last file in the list.
> 3. Otherwise, ${HOME}/.kube/config is used and no merging takes place.

source: https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#config

The Kubernetes client credentials will be written by `kind create cluster` to the path determined by these rules, with a context name like `kind-foo` where `foo` is the `--name` of the cluster.

Like other tools, when exporting the credentials kind will set the current context to the new cluster. If you switch the current context again you can specifically reference the kind cluster by using `kubectl` with `--context=kind-foo`.

The instructions output by `kind create cluster` have been updated to reflect this.

## Emulating The Old Behavior

You can emulate the previous behavior by switching from:
```shell
kind create cluster --name=foo
export KUBECONFIG="$(kind get kubeconfig-path --name=foo)"
```
to:
```shell
export KUBECONFIG="${HOME}/.kube/kind-config-foo"
kind create cluster --name=foo
```

This may be useful for some test scripts.

## Additional Notes

- Going forward kind clusters will be identified in kubeconfigs by the `kind-` prefix on entries
- Note that your current context will be set when creating a cluster. If you wish to keep kind seperate from your other clusters, please specify `--kubeconfig` or `KUBECONFIG` to point to another file when calling `kind create cluster`
[...]

PR #1059, opened 2019-11-06
fix cli.Logger.writeBuffer

BenTheElder:
fixes logging empty strings

PR #1061, opened 2019-11-06
Bring back get kubeconfig-path

BenTheElder:
xref: https://github.com/kubernetes-sigs/kind/issues/850 
references: https://github.com/kubernetes-sigs/kind/issues/1060

depends on: https://github.com/kubernetes-sigs/kind/pull/1059

### DIFF
--- a/pkg/cmd/kind/get/kubeconfigpath/kubeconfigpath.go
+++ b/pkg/cmd/kind/get/kubeconfigpath/kubeconfigpath.go
@@ -1,0 +1,69 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package kubeconfigpath implements the `kubeconfig-path` command
+package kubeconfigpath
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"sigs.k8s.io/kind/pkg/cluster"
+	"sigs.k8s.io/kind/pkg/cmd"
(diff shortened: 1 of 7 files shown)


BenTheElder:
```
$ kind get kubeconfig-path
`kind get kubeconfig-path` is deprecated!

KIND will export and merge kubeconfig like kops, minikube, etc.
This command is now unnecessary and will be removed in a future release.

For more info see: https://github.com/kubernetes-sigs/kind/issues/1060
See also the output of `kind create cluster`

/Users/bentheelder/.kube/config
```
(note that only the last line is stdout)


The only problem is that users that are doing `$(kind get kubeconfig-path)` will never see this...

Hopefully they check the release notes, which will include this information as well.

aojea:
maybe we can remove it in 0.7.0 and warn it now, just using 0.6.0 as a deprecation cycle
This will give at least some notice period, what do you think?

BenTheElder:
Right. That's what the comments say roughly I think?
This needs to exist when we release 0.6.0 but it can be removed at some point after.

pkg/cmd/kind/get/kubeconfigpath/kubeconfigpath.go
```diff
@@ -0,0 +1,69 @@
...
+		"the cluster context name",
+	)
+	return cmd
+}
+
+func runE(logger log.Logger, streams cmd.IOStreams) error {
+	logger.Warn("`kind get kubeconfig-path` is deprecated!")
+	logger.Warn("")
+	logger.Warn("KIND will export and merge kubeconfig like kops, minikube, etc.")
+	logger.Warn("This command is now unnecessary and will be removed in a future release.")
```
aojea: I guess we can replace `in a future release`  by `next release`?

BenTheElder: that's confusing for anyone that is using it at head?
this warning doesn't need to be that specific, we will put it in the release notes too.
we can also choose not to remove it in 0.7 for some reason

tao12345666333: > that's confusing for anyone that is using it at head?
> this warning doesn't need to be that specific, we will put it in the release notes too.
> we can also choose not to remove it in 0.7 for some reason

+1 .

BenTheElder:
definitely underestimated https://github.com/search?p=11&q=%22kind+get+kubeconfig-path%22&type=Code before 🙃

dims:
LOL indeed :)

PR #1064, opened 2019-11-06
offline docs

### LINKED ISSUE #345: document offline clusters
xref: https://github.com/kubernetes-sigs/kind/pull/331#issuecomment-466668764

We should also think about how to start breaking up some of these things into seperate advanced guides to keep the quick start focused. unsure on this one, but https://github.com/kubernetes-sigs/kind/issues/303 will be a good one to split out

/kind documentation
/priority important-soon

tao12345666333:
fix #345

### DIFF
--- a/site/content/docs/user/working-offline.md
+++ b/site/content/docs/user/working-offline.md
@@ -16,7 +16,7 @@ You can first prepare the necessary environment according to the [installation d
 
 ## Using a pre-built [node image][node image]
 
-Kind provided some pre-built images,
+KIND provides some pre-built images,
 these images contain everything necessary to create a cluster and can be used in an offline environment.
 
 You can find available image tags on the [releases page][releases page].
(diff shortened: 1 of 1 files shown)


BenTheElder:
rendered: https://deploy-preview-1064--k8s-kind.netlify.com/

tao12345666333:
What I am going to write includes two parts.

1. Use an existing image (this process can also be considered offline when the image exists)

But the Pull image also needs the network, so I am not sure if I need to write it here. According to https://github.com/kubernetes-sigs/kind/pull/331#issuecomment-466674273 it seems meaningful to write it. @BenTheElder  WDYT?

2. Build your own node image

BenTheElder:
> But the Pull image also needs the network, so I am not sure if I need to write it here. According to #331 (comment) it seems meaningful to write it. @BenTheElder WDYT?

the node image must be obtained somehow, the easiest way is to pull it when you have network access.

it could also be transported to a machine by transfering a `docker save` style tarball

tao12345666333:
I didn't describe how to build own node image.
Because I found that the process of building an image is not very suitable for offline use.

@BenTheElder  PTAL. Thanks.

tao12345666333:
https://deploy-preview-1064--k8s-kind.netlify.com/docs/user/working-offline/

tao12345666333:
ping @BenTheElder If you have time, PTAL. Thanks.

BenTheElder:
Image building should be fixed for offline as well but we can document it when that is true.

tao12345666333:
> we should probably mention side loading and if doing HA clusters the loadbalancer image?

ok, I will add it.



> Image building should be fixed for offline as well but we can document it when that is true.

If we build `node-image` with `docker` type (default), we need many other things like `k8s.gcr.io/kube-cross` image and some others for buil kubernetes.

It might be more convenient to use `bazel`, let me try! (I always using `docker` type)

BenTheElder:
> It might be more convenient to use bazel, let me try! (I always using docker type)

bazel is maybe going away upstream

I would just document that you should build *once* online to let the upstream build tools fetch dependencies. this also applies to bazel, which I assure you downloads far more things than the make build.

BenTheElder:
sorry for the ridiculous delay here. everything is crazy right now :-)

tao12345666333:
Be careful :cat: 

Let me update this PR.

tao12345666333:
> bazel is maybe going away upstream

I see related issues https://github.com/kubernetes/kubernetes/issues/88553

tao12345666333:
Add some contents about building node-image and create HA cluster. 

@BenTheElder  PTAL, thanks!

tao12345666333:
ping @BenTheElder  PTAL.  
Thanks. :smile_cat:

BenTheElder:
pong ... 8 days later.
my latency has gone up 😂 :this-is-fine:

BenTheElder:
/lgtm
/approve
we can iterate, this is a wonderful addition, thank you

PR #1067, opened 2019-11-07
Update dependencies

BenTheElder:
for the main kind CLI
will approach kindnetd carefully in a different PR at some point

EDIT:
- switch to go-isatty which is a better scoped for checking if a file is a tty than x/crypto (and has version tags...)
- drop superfluous uuid dependency, just use the timestamp and a random int for naming the node build container


BenTheElder:
updated to cleanup the deps a bit (updated the PR body to describe)

pkg/internal/util/env/term.go
```diff
@@ -20,13 +20,13 @@ import (
 	"io"
 	"os"
 
-	"golang.org/x/crypto/ssh/terminal"
+	isatty "github.com/mattn/go-isatty"
```
tao12345666333: :+1:

BenTheElder:
```
[bentheelder@bentheelder-macbookpro:~/go/src/sigs.k8s.io/kind·2019-11-07T01:40:49-0800·master@f56b2cb]
$ kind version
kind v0.6.0-alpha+f56b2cbad927a5 go1.13.4 darwin/amd64

[bentheelder@bentheelder-macbookpro:~/go/src/sigs.k8s.io/kind·2019-11-07T01:40:51-0800·master@f56b2cb]
$ du -h $(which kind)
8.8M    /Users/bentheelder/go/bin/kind
```

:-)

dropped another half megabyte somewhere in here as a side effect, nice bonus!

PR #1068, opened 2019-11-07
add `kind export kubeconfig`

BenTheElder:
This allows re-obtaining kind kubeconfig similar to `kind get kubeconfig`, but merging it in like `kind create cluster` instead of printing it to stdout.

xref: #850


BenTheElder:
inspired by: https://github.com/kubernetes/kops/blob/master/docs/cli/kops_export_kubecfg.md

BenTheElder:
/retest
Prow lack of job started

PR #1072, opened 2019-11-07
Update tools

BenTheElder:
- update golangci-lint
- drop unused tools (code generators we no longer user)
- update gotestsum
- fix new lint errors
- fix golangci-lint containerization ref: https://github.com/golangci/golangci-lint/issues/734
- add a make target for lint


hack/go_container.sh
```diff
@@ -75,7 +75,7 @@ run_in_go_container() {
     `# docker options: remove container on exit, run as the host user / group` \
       --rm --user "$(id -u):$(id -g)" \
     `# golang caching: mount and use the cache volume` \
-      -v "${CACHE_VOLUME}:/go" -e GOCACHE=/go/cache \
+      -v "${CACHE_VOLUME}:/go" -e GOCACHE=/go/cache -e XDG_CACHE_HOME=/go \
```
BenTheElder: https://github.com/golangci/golangci-lint/issues/734
> ... golangci-lint was changed to use GOLANGCI_LINT_CACHE for its cache directory, or os.UserCacheDir if that is not defined ...

`os.UserCacheDir` will read `XDG_CACHE_HOME` on linux, which is more general than setting the special golangci-lint option so /shrug

tao12345666333: /hold

tao12345666333: I did a quick test locally.

If I set ` -e XDG_CACHE_HOME=/go/cache` all cached data will be 0.7 M smaller than it is now.

tao12345666333: setting| cached directory |all cached data size|
---|--|--
`-e XDG_CACHE_HOME=/go`| /go/cache and /go/golangci-lint|157.9M
`-e XDG_CACHE_HOME=/go/cache`|  /go/cache|157M

BenTheElder: Will update when I'm back at a keyboard to just set XDG_CACHE_HOME, I checked and go should use os.UserCacheDir unless GOCACHE is set.

BenTheElder: setting this now

BenTheElder: ah, though I'm not setting `GOCACHE` which sets the exact path of the go cache.
IMO that's wrong, the overall cache directory `XDG_CACHE_HOME` should contain each cache directory, and it's WAI that golangci-lint has it's own small additional cache.

.7M is less than half a percent here.

tao12345666333: The result is here:
```
(MoeLove) ➜  kind git:(pr1072) docker run --rm -it  -v kind-build-cache:/go golang:1.13.4 bash
root@2f3431bfa012:/go# ls cache
go-build  golangci-lint
root@2f3431bfa012:/go# du -sh cache/
157M    cache/
root@2f3431bfa012:/go# du -sh cache/go-build/
151M    cache/go-build/
root@2f3431bfa012:/go# du -sh cache/golangci-lint/
6.2M    cache/golangci-lint/
```

go does use `os.UserCacheDir` when `GOCACHE` not set.

tao12345666333: Not sure why the above result will appear. Re-test here:
```
MoeLove) ➜  kind git:(pr1072) docker run --rm -it  -v kind-build-cache:/go golang:1.13.4 bash
root@c42b368e75fd:/go# du -sh cache/
157M    cache/
root@c42b368e75fd:/go# ls -al cache/
total 16
drwxr-xr-x.   4 1000 1000 4096 Nov  7 16:36 .
drwxrwxrwx.   6 root root 4096 Nov  7 16:35 ..
drwxr-xr-x. 258 1000 1000 4096 Nov  7 16:35 go-build
drwxr--r--. 258 1000 1000 4096 Nov  7 16:37 golangci-lint
root@c42b368e75fd:/go# du -sh cache/go-build/
152M    cache/go-build/
root@c42b368e75fd:/go# du -sh cache/golangci-lint/
5.9M    cache/golangci-lint/
```

tao12345666333: Now the directory structure is looking right. :+1:

Issue #1070, opened 2019-11-07
v1alpha4


BenTheElder:
In order to minimize the odds of notably breaking releases in the future we should round out v0.6 with a trivial `v1alpha4` config version that at least moves us out of the `*.k8s.io` namespace into `x-k8s.io` for kind's APIs to clarify that they are not official kubernetes APIs, in line with the API review requirements and guidelines.

### I propose the following minimal changes:

- Updated GVK:
```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
```
- Simply use strings (instead of string -> int) for "enums" I.E. for `MountPropagation` and `PortMappingProtocol`.  This will simplify parsing and make the on-disk format match the in-memory format. It's technically more expensive but not in a way that is material for kind ...

### And the following tentative additional feature:

Support for containerd config patches. This is necessary to sanely solve https://github.com/kubernetes-sigs/kind/issues/110. 

Rather than adding these to `v1alpha3` and possibly not working with previous versions of kind (especially now that we strict parse config!) these should be added to a new API version. 

Essentially these should work like kubeadm config patches, but the non-json-6902 format should parse as toml initially instead of yaml, and we don't need document separation or any GVK sniffing / matching so it should be much simpler.

Concretely this means `containerdConfigPatches []string` and `containerdConfigPatchesJSON6902 []string` fields, and some minor-ish work to apply them.

Originally I intended to leverage the built in merging in v1.3.0, but it is quite insufficient / the design does not seem to be intended for this use case (rather, for kata / gVisor runtime drop-ins).

### Additionally ...

We should transform `kind.sigs.k8s.io` docker labels to `kind.x-k8s.io` labels, but in v0.6 we should apply and list BOTH forms so as to continue to detect and manage existing clusters etc.

In the future we can drop the k8s.io labels.

When this apiVersion exists, we should start a new "Configuration" guide / page under /users in the docs to cover all of the features instead of making the quick start too long or relying on the godoc contents. 

We should still link to the godoc contents for completeness.

aojea:
Now that the provider code is properly isolated, should we add provider to the API?
I think that  we should also add the provider bridge too

```// Provider contains kind runtime provider settings
type Provider struct {
	// Provider is the runtime provider, currently it can be docker
	Provider string `yaml:"provider,omitempty"`
	// Bridge is the bridge used by the provider runtime
	Bridge string `yaml:"bridge,omitempty"`
```

[...]

PR #1074, opened 2019-11-07
introduce v1alpha4, move to x-k8s.io

BenTheElder:
This does NOT add any additional fields, however it does:
- move our labels to x-k8s.io (continuing to use the old ones as well, and listing based on the old ones)
- introduce a 1:1 version of v1alpha3 under v1alpha4 and x-k8s.io
  - move from int enums to plain strings for portmapping / mountpropagation in v1alpha4

In follow-ups:
- move kind CI to start using v1alpha4
- add new v1alpha4 feature(s)
- documentation user guide page for the API version prior to v0.6


BenTheElder:
this adds proper conversions as well, the intention is to continue to support v1alpha3 through v0.6

BenTheElder:
xref: https://github.com/kubernetes-sigs/kind/issues/1070

PR #1078, opened 2019-11-08
update link to strategic merge docs

dustinspecker:
Also used a SHA in the URL to prevent the link from breaking in the
future.


BenTheElder:
/lgtm
/approve
/ok-to-test
thanks!

since you're looking at these docs, I should mention that they actually don't use kustomize for v0.6  (kind at HEAD right now and soon to be released) forward and turned out to only be normal merge patch (lacking definitions for the types to do strategic merge)

BenTheElder:
I'll file a follow-on to clarify these with the current state, and finish up 0.6 😅

PR #1081, opened 2019-11-09
containerd TOML patching

BenTheElder:
xref: #1070, #110 

TODO:
- [x] add v1alpha4 config fields that are meant to leverage this
  - [x] wire up patching containerd config

### DIFF
--- a/pkg/internal/cluster/create/actions/config/config.go
+++ b/pkg/internal/cluster/create/actions/config/config.go
@@ -18,6 +18,7 @@ limitations under the License.
 package config
 
 import (
+	"bytes"
 	"strings"
 
 	"sigs.k8s.io/kind/pkg/cluster/constants"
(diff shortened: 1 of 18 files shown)


pkg/internal/util/patch/toml_test.go
```diff
@@ -0,0 +1,198 @@
...
+		{
+			Name: "patch registry",
+			ToPatch: `disabled_plugins = ["restart"]
+[plugins.linux]
+  shim_debug = true
+[plugins.cri.containerd.runtimes.runsc]
+  runtime_type = "io.containerd.runsc.v1"`,
+			Patches: []string{`[plugins.cri.registry.mirrors]
+  [plugins.cri.registry.mirrors."registry:5000"]
+    endpoint = ["http://registry:5000"]`},
```
BenTheElder: sample data from a real end-user work-around in #110 (kubevirt is doing this with `sed` at runtime today)

BenTheElder: (which will be increasingly tricky for them to do correctly as we configure more things in the base config, applying a patch avoids that problem)

pkg/internal/cluster/create/actions/config/config.go
```diff
@@ -119,11 +119,42 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
...
 	}
 
-	// Create the config in all nodes concurrently
+	// Create the kubeadm config in all nodes concurrently
 	if err := errors.UntilErrorConcurrent(fns); err != nil {
 		return err
 	}
 
+	// if we have containerd config, patch all the nodes concurrently
+	if len(ctx.Config.ContainerdConfigPatches) > 0 {
```
amwat: also check for ContainerdConfigPatchesJSON6902 ?

BenTheElder: done, thanks :-)

BenTheElder:
/retest
prow flake

BenTheElder:
fixes https://github.com/kubernetes-sigs/kind/issues/1070

PR #1083, opened 2019-11-10
Provide ability for Worker Kubeadm Config Patches

robertojrojas:
Issue Reference #1084 

Hopefully this is a good approach, but I'm open to any changes.
One thing is, I wonder if it makes more sense to place **`KubeadmConfigPatches`** and **`KubeadmConfigPatchesJSON6902`** into a separate struct and embed.

TODO:
 
- Integration Testing (working on this)

### DIFF
--- a/hack/update/generated.sh
+++ b/hack/update/generated.sh
@@ -45,13 +45,9 @@ cd "${FAKE_REPOPATH}"
 
 # run the generators
 bin/deepcopy-gen -i ./pkg/internal/apis/config/ -O zz_generated.deepcopy --go-header-file hack/tools/boilerplate.go.txt
-<<<<<<< HEAD
-bin/deepcopy-gen -i ./pkg/apis/config/v1alpha4 -O zz_generated.deepcopy --go-header-file hack/tools/boilerplate.go.txt
-=======
 bin/deepcopy-gen -i ./pkg/apis/config/v1alpha3 -O zz_generated.deepcopy --go-header-file hack/tools/boilerplate.go.txt
 bin/deepcopy-gen -i ./pkg/apis/config/v1alpha4 -O zz_generated.deepcopy --go-header-file hack/tools/boilerplate.go.txt
 
->>>>>>> master
 
 # set module mode back, return to repo root and gofmt to ensure we format generated code
 export GO111MODULE="on"
--- a/hack/update/generated.sh
+++ b/hack/update/generated.sh
@@ -45,7 +45,7 @@ cd "${FAKE_REPOPATH}"
 
 # run the generators
 bin/deepcopy-gen -i ./pkg/internal/apis/config/ -O zz_generated.deepcopy --go-header-file hack/tools/boilerplate.go.txt
-bin/deepcopy-gen -i ./pkg/apis/config/v1alpha3 -O zz_generated.deepcopy --go-header-file hack/tools/boilerplate.go.txt
+bin/deepcopy-gen -i ./pkg/apis/config/v1alpha4 -O zz_generated.deepcopy --go-header-file hack/tools/boilerplate.go.txt
 
 # set module mode back, return to repo root and gofmt to ensure we format generated code
 export GO111MODULE="on"
--- a/pkg/apis/config/v1alpha4/types.go
+++ b/pkg/apis/config/v1alpha4/types.go
@@ -38,6 +38,8 @@ type Cluster struct {
 	// This should be an inline yaml blob-string
 	//
 	// https://tools.ietf.org/html/rfc7386
+	//
+	// The cluster-level patches are appied before the node-level patches.
 	KubeadmConfigPatches []string `yaml:"kubeadmConfigPatches,omitempty"`
 
 	// KubeadmConfigPatchesJSON6902 are applied to the generated kubeadm config
@@ -51,6 +53,8 @@ type Cluster struct {
 	// always been a no-op.
 	//
 	// https://tools.ietf.org/html/rfc6902
+	//
+	// The cluster-level patches are appied before the node-level patches.
 	KubeadmConfigPatchesJSON6902 []PatchJSON6902 `yaml:"kubeadmConfigPatchesJson6902,omitempty"`
 }
 
@@ -93,6 +97,9 @@ type Node struct {
 	// This should be an inline yaml blob-string
 	//
 	// https://tools.ietf.org/html/rfc7386
+	//
+	// The node-level patches will be applied after the cluster-level patches
+	// have been applied. (See Cluster.KubeadmConfigPatches)
 	KubeadmConfigPatches []string `yaml:"kubeadmConfigPatches,omitempty"`
 
 	// KubeadmConfigPatchesJSON6902 are applied to the generated kubeadm config
@@ -106,6 +113,9 @@ type Node struct {
 	// always been a no-op.
 	//
 	// https://tools.ietf.org/html/rfc6902
+	//
+	// The node-level patches will be applied after the cluster-level patches
+	// have been applied. (See Cluster.KubeadmConfigPatchesJSON6902)
 	KubeadmConfigPatchesJSON6902 []PatchJSON6902 `yaml:"kubeadmConfigPatchesJson6902,omitempty"`
 }
 
--- a/pkg/apis/config/v1alpha4/types.go
+++ b/pkg/apis/config/v1alpha4/types.go
@@ -85,6 +85,28 @@ type Node struct {
 	// ExtraPortMappings describes additional port mappings for the node container
 	// binded to a host Port
 	ExtraPortMappings []PortMapping `yaml:"extraPortMappings,omitempty"`
+
+	// KubeadmConfigPatches are applied to the generated kubeadm config as
+	// merge patches. The `kind` field must match the target object, and
+	// if `apiVersion` is specified it will only be applied to matching objects.
+	//
+	// This should be an inline yaml blob-string
+	//
+	// https://tools.ietf.org/html/rfc7386
+	KubeadmConfigPatches []string `yaml:"kubeadmConfigPatches,omitempty"`
+
+	// KubeadmConfigPatchesJSON6902 are applied to the generated kubeadm config
+	// as JSON 6902 patches. The `kind` field must match the target object, and
+	// if group or version are specified it will only be objects matching the
+	// apiVersion: group+"/"+version
+	//
+	// Name and Namespace are now ignored, but the fields continue to exist for
+	// backwards compatibility of parsing the config. The name of the generated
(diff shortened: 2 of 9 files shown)


neolit123:
@robertojrojas 
was this discussed in an issue?

in the earlier iterations of the kind config it supported per-node patches, but it then moved to having patches per cluster.

how would the per-node patches interact with the cluster patches?
apply cluster patches to all nodes, then apply per-node patches?

robertojrojas:
@neolit123 My apologies I should have created an issue first. I just created it #1084 
We can continue discussing there.

pkg/internal/cluster/create/actions/config/config.go
```diff
@@ -85,12 +84,28 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 	if err != nil {
 		return err
 	}
+
+	kubeadmConfigPlusPatches := func(node nodes.Node, configData kubeadm.ConfigData) func() error {
+		return func() error {
+			// determine if we're dealing with a worker node
+			// so we can include it's corresponding patches
+			includeWorkerPatches := !configData.ControlPlane
```
BenTheElder: Why are worker nodes being special cased?
Why not just include any node level patches?

pkg/internal/cluster/create/actions/config/config.go
```diff
@@ -153,12 +165,21 @@ func removeMetadata(kustomized string) string {
 	)
 }
 
-func allPatchesFromConfig(cfg *config.Cluster) (patches []string, jsonPatches []config.PatchJSON6902) {
-	return cfg.KubeadmConfigPatches, cfg.KubeadmConfigPatchesJSON6902
+func allPatchesFromConfig(cfg *config.Cluster, includeWorkerPatches bool) (patches []string, jsonPatches []config.PatchJSON6902) {
+	patches, jsonPatches = cfg.KubeadmConfigPatches, cfg.KubeadmConfigPatchesJSON6902
+
+	if includeWorkerPatches {
+		for _, node := range cfg.Nodes {
```
BenTheElder: Why would ever node's patches be applied to all nodes? Why not just the patch for the current node?

pkg/internal/cluster/create/actions/config/config.go
```diff
@@ -121,22 +134,21 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
...
 		return "", err
 	}
-	// fix all the patches to have name metadata matching the generated config
-	patches, jsonPatches := allPatchesFromConfig(cfg)
+
 	// apply patches
-	// TODO(bentheelder): this does not respect per node patches at all
-	// either make patches cluster wide, or change this
-	patched, err := patch.Patch(config, patches, jsonPatches)
+	patched, err := patch.Patch(cf, patches, jsonPatches)
```
BenTheElder: This is going to have surprising behavior in the current implementation. All merge patches will be applied and then all 6902 patches.

As a user I would instead expect all cluster global patches and then all node level patches

hack/update/generated.sh
```diff
@@ -45,7 +45,7 @@ cd "${FAKE_REPOPATH}"
 
 # run the generators
 bin/deepcopy-gen -i ./pkg/internal/apis/config/ -O zz_generated.deepcopy --go-header-file hack/tools/boilerplate.go.txt
-bin/deepcopy-gen -i ./pkg/apis/config/v1alpha3 -O zz_generated.deepcopy --go-header-file hack/tools/boilerplate.go.txt
+bin/deepcopy-gen -i ./pkg/apis/config/v1alpha4 -O zz_generated.deepcopy --go-header-file hack/tools/boilerplate.go.txt
```
BenTheElder: v1alpha3 should not be removed

pkg/apis/config/v1alpha4/types.go
```diff
@@ -85,6 +85,28 @@ type Node struct {
 	// ExtraPortMappings describes additional port mappings for the node container
 	// binded to a host Port
 	ExtraPortMappings []PortMapping `yaml:"extraPortMappings,omitempty"`
+
+	// KubeadmConfigPatches are applied to the generated kubeadm config as
```
BenTheElder: These should document their relationship to the cluster wide patches, IE that they are applied next and are more specific.

pkg/internal/cluster/create/actions/config/config.go
robertojrojas: The idea was to only apply the worker patches to workers and not to the control planes nodes, but this should also merge with the cluster level patches.

pkg/internal/cluster/create/actions/config/config.go
robertojrojas: Yes, you're right. The patches should be applied to the current node. Looks like we would need to send the current node down to this function or the like.

pkg/internal/cluster/create/actions/config/config.go
robertojrojas: Yes, I agree as a user I would expect the same.  I didn't think about it throughly. So, something like this maybe? Of course, I would need to provide patches separately and better code. wdyt?

```Go
clusterPatched, err := patch.Patch(cf, clusterPatches, clusterJSONPatches)
...
allPatched, err := patch.Patch(clusterPatched nodePatches, nodeJSONPatches)
...

```

hack/update/generated.sh
robertojrojas: so both v1alpha3 and v1alpha4 be update and verified, right?

pkg/apis/config/v1alpha4/types.go
robertojrojas: Agreed. At first I thought about struct embedding, but I honestly don't know if it'll make the code and docs more confusing.

hack/update/generated.sh
BenTheElder: right

pkg/apis/config/v1alpha4/types.go
BenTheElder: we just need a comment here about the behavior here (being applied on top of the others)

pkg/internal/cluster/create/actions/config/config.go
BenTheElder: I think this works, I would add that we can skip running the second patch logic in the case that there are no patches, but we should always run the first patch (it also has the behavior of normalizing the output).

```go
patchedConfig, err := patch.Patch(cf, clusterPatches, clusterJSONPatches)
// handle error
if len(nodePatches) > 0 || len(nodeJSONPatches) > 0 {
  patchedConfig, err = patch.Patch(patchedConfig, nodePatches, nodeJSONPatches)
  // handle error
}
```

pkg/apis/config/v1alpha4/types.go
robertojrojas: Cool beans! I'll work on these changes later today. I'll be integration testing this quite a bit as I'm working on an issue for [kube-bench](https://github.com/aquasecurity/kube-bench/issues/439) 😄

BenTheElder: thanks :-)

pkg/internal/cluster/create/actions/config/config.go
robertojrojas: @BenTheElder  I ran into an interesting situation - **config.Node** vs **nodes.Node**.  As you know, the later is what I have access to within the **getKubeadmConfig** function, so how do I match the right **config.Node.'KubeadmConfigPatches'** patch(es) to apply it to the current  **nodes.Node** without having anything to correlate with?  I thought about the Role, but that's is not specific enough. The array indexes from nodes.Node(i) == cfg.Nodes[i] might do it, but is the index reliable in this case?  thoughts?

BenTheElder:
The index of the nodes.Node[] is not reliable.
However the name of the node currently maps to the role and index in the
config.
This is brittle coupling wise and needs to go away in a future version, but
it is stable for now.

We can match the entries based on naming them again back to the named nodes.

robertojrojas:
@BenTheElder I must be missing something because I don't see a **'name'** on the node (nodes.Node type) 
It's an interface and I do see a **String()**, is that what you mean?

BenTheElder:
yes, the nodes.Node interface allows obtaining the actual node name with
String()
https://github.com/kubernetes-sigs/kind/blob/55dd46c27c8af766c5712212b8a759c7e5e208cb/pkg/cluster/nodes/types.go#L28

robertojrojas:
Is this the format of the String() output?
https://github.com/kubernetes-sigs/kind/blob/55dd46c27c8af766c5712212b8a759c7e5e208cb/pkg/internal/cluster/providers/provider/common/namer_test.go#L43

BenTheElder:
yes, if we're going to recompute the node names we should use the namer.

robertojrojas:
yep, I saw that.

robertojrojas:
@BenTheElder Please let me know what you think. In the meantime, I'm going to be testing it.

pkg/internal/cluster/create/actions/config/config.go
```diff
@@ -158,7 +181,7 @@ func allPatchesFromConfig(cfg *config.Cluster) (patches []string, jsonPatches []
 }
 
 // writeKubeadmConfig writes the kubeadm configuration in the specified node
-func writeKubeadmConfig(logger log.Logger, cfg *config.Cluster, data kubeadm.ConfigData, node nodes.Node) error {
+func writeKubeadmConfig(cfg *config.Cluster, kubeadmConfig string, data kubeadm.ConfigData, node nodes.Node) error {
```
BenTheElder: this is wrong, we should not be taking in data and the kubeadm config.
we fix up the data below, but we've already generated the config

robertojrojas: Good catch! It turns out that there were some **"data"** configs that I missed. Fixing...

robertojrojas:
@BenTheElder So far my integration tests are working fine.
 I did noticed the **apiVersion** changed from 
_**`kind.sigs.k8s.io`**/v1alpha3_ 
to
 _**`kind.x-k8s.io`**/v1alpha4_

BenTheElder:
Yes, there's two supported config versions for this release, we'll put this in v1alpha4, we also needed to move domains because of revised guidelines around that.

BenTheElder:
Sorry for the delay, I had to do some test-infra oncall (rotation over now) and sort out some Kubernetes things pre-code freeze, I'll go over this tonight and try to get v0.6.0 rolled up for tomorrow including this 😁

robertojrojas:
Awesome sauce!

BenTheElder:
/lgtm
/approve
Thanks for this!
I have some bikeshed, but we can iterate later :+)

BenTheElder:
Fixes https://github.com/kubernetes-sigs/kind/issues/1084

PR #1089, opened 2019-11-11
handle windows terminals better

BenTheElder:
go-colorable turns out to be really slow AND on modern windows terminals it's not necessary.
https://github.com/microsoft/terminal has pretty much full support for UTF-8, escape codes, etc...

We can detect the modern terminal by checking for `$WT_SESSION`, which will have a guid for modern terminal sessions.  On other platforms we shouldn't need to check, but we can support explicit `TERM=dumb` which is a pattern I found in a few places...

Also on Windows the escape codes for line wrapping seems to behave poorly in the vscode terminal (which is otherwise a modern terminal host) ... they're only necessary for dealing with wrapped lines, so let's just disable them on Windows for now.

### DIFF
--- a/pkg/internal/util/env/term.go
+++ b/pkg/internal/util/env/term.go
@@ -19,6 +19,7 @@ package env
 import (
 	"io"
 	"os"
+	"runtime"
 
 	isatty "github.com/mattn/go-isatty"
 )
@@ -29,4 +30,22 @@ func IsTerminal(w io.Writer) bool {
 		return isatty.IsTerminal(v.Fd())
 	}
 	return false
+}
+
+// IsSmartTerminal returns true if the writer w is a terminal AND
+// we think that the terminal is smart enough to use VT escape codes etc.
+func IsSmartTerminal(w io.Writer) bool {
+	if !IsTerminal(w) {
+		return false
+	}
+	// explicitly dumb terminals are not smart
+	if os.Getenv("TERM") == "dumb" {
+		return false
+	}
+	// On Windows WT_SESSION is set by the modern terminal component.
+	// Older terminals have poor support for UTF-8, VT escape codes, etc.
+	if runtime.GOOS == "windows" && os.Getenv("WT_SESSION") == "" {
+		return false
+	}
+	return true
 }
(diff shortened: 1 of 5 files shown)


BenTheElder:
I'll check some linux terminals tomorrow to verify that they do handle the VT codes we use properly.
The macOS terminal does.

The ANSI codes work everywhere except `cmd.exe`. We could even get the ansi color codes to work there by doing something like this:
```go
func enableANSI(console *os.File) error {
  handle := windows.Handle(console.Fd())
  var originalMode uint32
  if err := windows.GetConsoleMode(handle, &originalMode); err != nil {
    return err
  }
  if err := windows.SetConsoleMode(handle, originalMode|windows.ENABLE_VIRTUAL_TERMINAL_PROCESSING); err != nil {
    return err
  }
  return nil
}
```
But that terminal doesn't handle much well, so we it's not really worth it at the moment.

pkg/internal/util/env/term.go
```diff
@@ -30,3 +31,21 @@ func IsTerminal(w io.Writer) bool {
...
 }
+
+// IsSmartTerminal returns true if the writer w is a terminal AND
+// we think that the terminal is smart enough to use VT escape codes etc.
+func IsSmartTerminal(w io.Writer) bool {
+	if !IsTerminal(w) {
+		return false
+	}
+	// explicitly dumb terminals are not smart
+	if os.Getenv("TERM") == "dumb" {
```
aojea: this is the output with TERM=dumb

```
Creating cluster "kind" ...
 • Ensuring node image (kindest/node:v1.16.2) 🖼  ...
 ✓ Ensuring node image (kindest/node:v1.16.2) 🖼
 • Preparing nodes 📦  ...
 ✓ Preparing nodes 📦
 • Creating kubeadm config 📜  ...
 ✓ Creating kubeadm config 📜
 • Starting control-plane 🕹️  ...
```

aojea:
:thinking: 

> ERROR: failed to create cluster: failed to get kubernetes version from node: failed to get file: command "docker exec --privileged kind-worker2 cat /kind/version" failed with error: exit status 1
> Output:
> Error response from daemon: Container 9287d971d292606c5e902bf6b13a5c2b12ffe189e8f5163a90e4cc1c1f31167d is not running

pkg/internal/util/env/term.go
BenTheElder: Right. No spinner, no vt codes.
This is already the fallback.

BenTheElder:
Unrelated flake. New mode though.

aojea:
/lgtm
I've tested it locally and in a remote ssh session and colors are ok

PR #1092, opened 2019-11-12
Fix line wrapping in terminals.

amwat:
The ansi escape code to erase the current line adds mysterious whitespace when terminal is resized.
Just simple carriage return should be sufficient.


amwat:
This is in GNOME Terminal:

kindbug

BenTheElder:
/lgtm
/approve
The line clear is a nice to have in theory for line rewriting but in practice when we rewrite the line we clobber all characters so..

PR #1093, opened 2019-11-12
Improve section about tags and imagePullPolicy

blacksails:
This is an attempt to make the explaination of tags and
`imagePullPolicy` combinations less confusing.

### DIFF
--- a/site/content/docs/user/quick-start.md
+++ b/site/content/docs/user/quick-start.md
@@ -132,16 +132,21 @@ kind load docker-image my-custom-image:unique-tag
 kubectl apply -f my-manifest-using-my-image:unique-tag
 ```
 
-**Note**: The Kubernetes default pull policy is `IfNotPresent` unless
-the image tag is `:latest` in which case the default policy is `Always`.
-`IfNotPresent` causes the Kubelet to skip pulling an image if it already exists.
-If you want those images loaded into node to work as expected, please:
+**Note**: The Kubernetes default pull policy is `IfNotPresent` unless the image
+tag is `:latest` in which case the default policy is `Always`.  `IfNotPresent`
+causes the Kubelet to skip pulling an image if it already exists. When using
+`kind load` we would like to avoid the `Always` pull policy, so that we can
+instead rely on the images which we inject. If you want images loaded
+into the cluster to work as expected, please:
 
-- don't use a `:latest` tag
+- don't use a `:latest` tag. If you must, then you required to set
+  `imagePullPolicy` to `IfNotPresent` or `Never`.
 
-and / or:
+or
 
-- specify `imagePullPolicy: IfNotPresent` or `imagePullPolicy: Never` on your container(s).
+- use a unique tag whenever your image changes. Then you can leave
+  `imagePullPolicy` unset, which will make it default to `IfNotPresent`. If you
+  want you can also explicitly set it to `IfNotPresent` or `Never`.
 
 See [Kubernetes imagePullPolicy][Kubernetes imagePullPolicy] for more information.
 


blacksails:
I signed it

site/content/docs/user/quick-start.md
```diff
@@ -135,13 +135,12 @@ kubectl apply -f my-manifest-using-my-image:unique-tag
...
 `IfNotPresent` causes the Kubelet to skip pulling an image if it already exists.
-If you want those images loaded into node to work as expected, please:
+If you want those images loaded into node to work as expected, please either:
 
-- don't use a `:latest` tag
-
-and / or:
-
-- specify `imagePullPolicy: IfNotPresent` or `imagePullPolicy: Never` on your container(s).
+- Leave `imagePullPolicy` unset so that it uses the default behavior
```
BenTheElder: The default image pull policy depends on the tag.

site/content/docs/user/quick-start.md
```diff
@@ -135,13 +135,12 @@ kubectl apply -f my-manifest-using-my-image:unique-tag
...
+- If you use a `:latest` tag set `imagePullPolicy` to `Always`
```
BenTheElder: You probably want it to explicitly not be always, always will pull. This explanation is the opposite of how to use `kind load ..` ?

BenTheElder:
This is more or less https://kubernetes.io/docs/concepts/containers/images/#updating-images (which we link to) but flipped for forcing NOT pulling when we've "side-loaded" an image.

site/content/docs/user/quick-start.md
blacksails: Changed it to only say that you can only use the default if using unique tags.

blacksails:
@BenTheElder I think I get it now. I changed the PR to my most recent understanding 😅

site/content/docs/user/quick-start.md
```diff
@@ -135,13 +135,11 @@ kubectl apply -f my-manifest-using-my-image:unique-tag
...
-If you want those images loaded into node to work as expected, please:
+If you want those images loaded into node to work as expected, please either:
 
-- don't use a `:latest` tag
-
-and / or:
-
-- specify `imagePullPolicy: IfNotPresent` or `imagePullPolicy: Never` on your container(s).
+- If you use a `:latest` tag set `imagePullPolicy` to `IfNotPresent` or `Never`
+- If you use a unique tag whenever your image changes either leave
```
BenTheElder: Thanks for the PR, but, I'm concerned that this doesn't discouragd `:latest` and conflates dealing with `:latest` and generally dealing with `imagePullPolicy` which should be two distinct but related topics, and more generally covered in the upstream documentation.

We should maybe breifly mention that you don't want `Always` above this, and link to the upstream docs again there.

The upstream docs recommend against latest, and they're right to do so, we should find a way to continue to suggest that here.

The "either or" should probably still be written along the lines of "please not use a :latest tag, or if you must, then use these settings".

blacksails:
I added a sentence mentioning that we would like to avoid the `Always` pull policy, and made it clear that `:latest` is discouraged.

site/content/docs/user/quick-start.md
```diff
@@ -132,16 +132,21 @@ kind load docker-image my-custom-image:unique-tag
...
 
-- don't use a `:latest` tag
+- don't use a `:latest` tag. If you must, then you required to set
+  `imagePullPolicy` to `IfNotPresent` or `Never`.
 
-and / or:
+or
 
-- specify `imagePullPolicy: IfNotPresent` or `imagePullPolicy: Never` on your container(s).
+- use a unique tag whenever your image changes. Then you can leave
```
BenTheElder: you don't need to use a unique tag though, it just needs to not be `:latest`.

Issue #1094, opened 2019-11-13
support `KIND_CLUSTER_NAME`


BenTheElder:
needs a little more thought but roughly see https://github.com/kubernetes-sigs/kind/issues/850

moving tracking over here, the bulk of #850 (the breaking and most useful parts imho) landed for v0.6.0, this will be an interesting additional feature for v0.7.0

prksu:
@BenTheElder can i do this one?

some question. currently, we have `"kind"` as `DefaultClusterName` should we replace it with env variable? or still used it if KIND_CLUSTER_NAME env not set?

BenTheElder:
Hi, you can but I recommend a "good first issue" or "help wanted" for starting usually. We are not going to merge this before v0.6.0

> some question. currently, we have "kind" as DefaultClusterName should we replace it with env variable? or still used it if KIND_CLUSTER_NAME env not set?

The environment variables should only be used when set and when the flag is not set. Otherwise we should maintain the default. (See #850)

This logic should be de-duplicated in an internal package (something like pkg/cmd/internal/flag).

prksu:
@BenTheElder just created a PR on #1099 

please take a look 😁

PR #1099, opened 2019-11-17
Support KIND_CLUSTER_NAME env

prksu:
This PR allow kind obtain cluster name from `KIND_CLUSTER_NAME` environment variable

ref #1094 

/assign @BenTheElder

### DIFF
--- a/pkg/cmd/kind/create/cluster/createcluster.go
+++ b/pkg/cmd/kind/create/cluster/createcluster.go
@@ -28,6 +28,7 @@ import (
 	"sigs.k8s.io/kind/pkg/cluster"
 	"sigs.k8s.io/kind/pkg/cmd"
 	"sigs.k8s.io/kind/pkg/errors"
+	"sigs.k8s.io/kind/pkg/internal/util/cli"
 	"sigs.k8s.io/kind/pkg/log"
 )
 
@@ -49,10 +50,10 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 		Short: "Creates a local Kubernetes cluster",
 		Long:  "Creates a local Kubernetes cluster using Docker container 'nodes'",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runE(logger, streams, flags)
+			return runE(cmd, logger, streams, flags)
 		},
 	}
-	cmd.Flags().StringVar(&flags.Name, "name", cluster.DefaultName, "cluster context name")
+	cmd.Flags().String("name", "", fmt.Sprintf(`the cluster name. Overrides KIND_CLUSTER_NAME environment variable (default "%s")`, cluster.DefaultName))
 	cmd.Flags().StringVar(&flags.Config, "config", "", "path to a kind config file")
 	cmd.Flags().StringVar(&flags.ImageName, "image", "", "node docker image to use for booting the cluster")
 	cmd.Flags().BoolVar(&flags.Retain, "retain", false, "retain nodes for debugging when cluster creation fails")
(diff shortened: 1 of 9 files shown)


pkg/cmd/kind/create/cluster/createcluster.go
```diff
@@ -49,10 +50,10 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 		Short: "Creates a local Kubernetes cluster",
 		Long:  "Creates a local Kubernetes cluster using Docker container 'nodes'",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runE(logger, streams, flags)
+			return runE(cmd, logger, streams, flags)
 		},
 	}
-	cmd.Flags().StringVar(&flags.Name, "name", cluster.DefaultName, "cluster context name")
+	cmd.Flags().String("name", "", fmt.Sprintf(`the cluster name (default "%s")`, cluster.DefaultName))
```
BenTheElder: we should continue to document the default normally. it's possible to tell if a flag was set specifically.

pkg/cmd/kind/create/cluster/createcluster.go
```diff
@@ -27,6 +27,7 @@ import (
 
 	"sigs.k8s.io/kind/pkg/cluster"
 	"sigs.k8s.io/kind/pkg/cmd"
+	cmdutil "sigs.k8s.io/kind/pkg/cmd/util"
```
BenTheElder: pkg/cmd is already utils for cmd.
however, setting up this flag should also be internal, it's an implementation detail, NOT a public API

prksu: make sense, i'll update it

prksu: addressed

prksu:
@BenTheElder i just updated this PR. PTAL

pkg/cmd/kind/create/cluster/createcluster.go
BenTheElder: this comment still stands

pkg/cmd/kind/create/cluster/createcluster.go
BenTheElder: intended `sigs.k8s.io/kind/pkg/cmd/internal/...` but this is OK, only cmd needs access to this

pkg/cmd/kind/create/cluster/createcluster.go
prksu: i added 
```
the cluster name. Overrides KIND_CLUSTER_NAME environment variable (default "kind")
```
in `usage` argument. is that not enough?

rajalokan:
Hi @BenTheElder , Is this still open ? If this is not being done by anyone, I would like to take this.

BenTheElder:
it's still open. please check on the issue instead of PRs though, this interfers with the lifecycle bot

pkg/internal/util/cli/flag.go
```diff
@@ -0,0 +1,39 @@
...
+
+// GetClusterNameFlags get cluster name from flags and lookup
+// KIND_CLUSTER_NAME env if flags not set. default value is kind
+func GetClusterNameFlags(cmd *cobra.Command) string {
+	nameFromFlags, err := cmd.Flags().GetString("name")
+	if err != nil {
+		log.Fatalf("error accessing flag name for command %s: %v", cmd.Name(), err)
+	}
+
+	if len(nameFromFlags) > 0 {
```
BenTheElder, line 32: for anyone else attempting this "it's possible to tell if a flag was set specifically." => please check that instead of breaking defaulting, you can find more in the cobra / pflag docs.

PR #1100, opened 2019-11-17
improve installation doc

eromano:


README.md
```diff
@@ -42,7 +42,7 @@ On Mac & Linux:
 ```console
 curl -Lo ./kind https://github.com/kubernetes-sigs/kind/releases/download/v0.5.1/kind-$(uname)-amd64
 chmod +x ./kind
-mv ./kind /some-dir-in-your-PATH/kind
+mv ./kind $(go env GOPATH)/bin/kind
```
BenTheElder: If you have go installed you probably won't download a binary?

This won't work for users without go.

eromano: maybe I misunderstood. I got that is a must , have go installed

Issue #1103, opened 2019-11-18
404 url


tanjunchen:
|file_path|url|
|----|----|
|kind/site/content/docs/user/quick-start.md |https://github.com/kubernetes-sigs/kind/tree/master/pkg/build/base/sources|
|kind/site/content/docs/design/initial.md| https://github.com/kubernetes-sigs/kind/tree/master/images/base/entrypoint|

hwdef:
I will fix these

tao12345666333:
#1104  and #1105 

/close

PR #1104, opened 2019-11-18
fix 404 link

hwdef:
Ref: #1103


BenTheElder:
In the future can we just fix both links in one commit? Thanks :-)

hwdef:
@BenTheElder 
I'm not sure how to fix the other link.And I don't think the fix in #1105 is right.

PR #1115, opened 2019-11-20
Added notes on loading images to named clusters

storey247:
I got very confused trying to load images onto named instances of Kind, seemed nowhere documented that I had to specify the name of the cluster i wanted to load the image into.

Have updated the docs as appropriate to hopefully save confusion for people in future

### DIFF
--- a/site/content/docs/user/known-issues.md
+++ b/site/content/docs/user/known-issues.md
@@ -17,19 +17,17 @@ It may additionally be helpful to:
 - reach out and ask for help in [#kind] on the [kubernetes slack]
 
 ## Contents
-- [Known Issues](#known-issues)
-  - [Contents](#contents)
-  - [Failures involving mismatched kubectl versions](#failures-involving-mismatched-kubectl-versions)
-  - [Older Docker Installations](#older-docker-installations)
-  - [Docker on Btrfs](#docker-on-btrfs)
-    - [Docker Installed with Snap](#docker-installed-with-snap)
-  - [Failing to apply overlay network](#failing-to-apply-overlay-network)
-  - [Failure to build node image](#failure-to-build-node-image)
-  - [Failing to properly start cluster](#failing-to-properly-start-cluster)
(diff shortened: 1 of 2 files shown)


storey247:
I have signed the agreement

neolit123:
> storey247 authorized, but 1 other problem

@storey247 please author your git commit with the same email as the one you signed the CLA with.

storey247:
Whoops, yes of course, will push another commit shortly, apologies

storey247:
@neolit123 the CLA issue should now be resolved, apologies for the confusion :)

site/content/docs/user/known-issues.md
```diff
@@ -286,6 +287,19 @@ or try to use `sudo` before your commands (if you get `command not found` please
...
+
+```
+Failed to pull image "docker.io/my-custom-image:tag": rpc error: code = Unknown desc = failed to resolve image "docker.io/library/my-custom-image:tag": no available registry endpoint: pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
+```
+
+If this image has been loaded onto your kind cluster using the command `kind load docker-image my-custom-image` then you have likely not provided the name parameter. 
+
+Re-run the command this time adding the `--name my-cluster-name` param:
+
+`kind load docker-image my-custom-image --name my-cluster-name`
```
neolit123: i wonder if` --name` related issues can be added as a generic item?
the same issue applies to e.g. `kind delete cluster`

storey247: While I agree the `—name` usage doesn’t need a bit more coverage. This particular piece of documentation was written because the inability to pull an image in a pod manifests itself with a cryptic error and isn’t immediately obvious

BenTheElder: agreed, we should probably expand on this entry later to link to the quick-start entry covering `imagePullPolicy` as well which is another frequent issue here.

BenTheElder:
skipping tests, thanks for this!

Issue #1116, opened 2019-11-20
Running kind delete cluster has issue deleting linked containers


christianh814:
**What happened**:

Running `kind delete cluster` gives the following output

```
ERROR: failed to delete cluster: failed to delete nodes: command "docker rm -f -v kind-control-plane2 kind-control-plane,ingress-proxy-80/target,ingress-proxy-443/target kind-external-load-balancer kind-control-plane3 kind-worker2 kind-worker kind-worker3" failed with error: exit status 1
```

**What you expected to happen**:

For the cluster to get deleted without error

**How to reproduce it (as minimally and precisely as possible)**:

1. Create cluster

```
kind create cluster --config=config.yaml
```

The contents of `config.yaml`

```
kind: Cluster
apiVersion: kind.sigs.k8s.io/v1alpha3
networking:
  disableDefaultCNI: True
nodes:
- role: control-plane
- role: control-plane
- role: control-plane
- role: worker
- role: worker
- role: worker
kubeadmConfigPatches:
- |
  apiVersion: kubeadm.k8s.io/v1beta2
  kind: ClusterConfiguration
  metadata:
    name: config
  networking:
    serviceSubnet: "172.30.0.0/16"
    podSubnet: "10.254.0.0/16"
```

2. Run the workaournd to make calico work

```
docker exec -ti kind-control-plane sysctl -w net.ipv4.conf.all.rp_filter=0
docker exec -ti kind-control-plane2 sysctl -w net.ipv4.conf.all.rp_filter=0
docker exec -ti kind-control-plane3 sysctl -w net.ipv4.conf.all.rp_filter=0
docker exec -ti kind-worker sysctl -w net.ipv4.conf.all.rp_filter=0
docker exec -ti kind-worker2 sysctl -w net.ipv4.conf.all.rp_filter=0
docker exec -ti kind-worker3 sysctl -w net.ipv4.conf.all.rp_filter=0
```

3. Install calico

```
curl -so manifests/calico.yaml https://docs.projectcalico.org/v3.8/manifests/calico.yaml
sed -i 's/192\.168/10\.254/g' manifests/calico.yaml
kubectl apply -f manifests/calico.yaml
```

4. Label the workers as such

```
for worker in kind-worker kind-worker{2..3}
do
  kubectl label node ${worker} node-role.kubernetes.io/worker=''
done
```

5. Install nginx using helm

```
kubectl create ns ingress-nginx

[...]

PR #1117, opened 2019-11-21
providers/docker: omit linked containers when listing nodes

neolit123:
When a kind node container is linked to another container
using something like "--link kind-control-plane:target"
The output of commands such as "kind get nodes"
can include the following:

  kind-control-plane,some-container-80/target

Make ListNodes() include only the part before the first ","
as the node name.

/kind bug
/priority important-longterm

fixes https://github.com/kubernetes-sigs/kind/issues/1116

### DIFF
--- a/pkg/internal/cluster/providers/docker/provider.go
+++ b/pkg/internal/cluster/providers/docker/provider.go
@@ -106,8 +106,9 @@ func (p *Provider) ListNodes(cluster string) ([]nodes.Node, error) {
 	}
 	// convert names to node handles
 	ret := make([]nodes.Node, 0, len(lines))
-	for _, name := range lines {
-		ret = append(ret, p.node(name))
+	for _, line := range lines {
+		names := strings.Split(line, ",") // omit linked containers
+		ret = append(ret, p.node(names[0]))
 	}
 	return ret, nil
 }


pkg/internal/cluster/providers/docker/provider.go
```diff
@@ -106,8 +106,9 @@ func (p *Provider) ListNodes(cluster string) ([]nodes.Node, error) {
 	}
 	// convert names to node handles
 	ret := make([]nodes.Node, 0, len(lines))
-	for _, name := range lines {
-		ret = append(ret, p.node(name))
+	for _, line := range lines {
+		names := strings.Split(line, ",") // omit linked containers
```
BenTheElder: is this documented in docker anywhere?

neolit123: couldn't find any details on why these are printed as part of `.Names`.

BenTheElder:
coming back to this, this is a pretty frustrating leaky implementation detail :(
want to be sure we know what the expected contents are and get this right

neolit123:
closing as per the discussion here:
https://github.com/kubernetes-sigs/kind/issues/1116#issuecomment-561409444

/close

BenTheElder:
thanks again,
for future reference, I think the complete fix would involve walking the list of .Names and keeping the one without a `/` in it. The output seems to vary a bit between usage, locally reproducing the issue gave different results...

PR #1119, opened 2019-11-21
fix the script and enhance the doc

brightzheng100:
two small updates:

1. fixed the script as in `v0.6.0`, `kind get nodes` will get nothing, so `--name` must be provided:

```
$ kind version
kind v0.6.0 go1.13.4 darwin/amd64
$ kind get nodes
$ kind get nodes --name my-cluster
my-cluster-worker3
my-cluster-worker2
my-cluster-control-plane
my-cluster-worker
```

2. enhanced a bit on the doc on how to use the registry.

### DIFF
--- a/site/content/docs/user/local-registry.md
+++ b/site/content/docs/user/local-registry.md
@@ -22,6 +22,7 @@ with it enabled.
 set -o errexit
 
 # create registry container unless it already exists 
+CLUSTER_NAME="my-cluster"   # desired cluster name; default is "kind"
 REGISTRY_CONTAINER_NAME='kind-registry'
 REGISTRY_PORT='5000'
 if [ "$(docker inspect -f '{{.State.Running}}' "${REGISTRY_CONTAINER_NAME}")" != 'true' ]; then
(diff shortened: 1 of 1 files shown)


site/content/docs/user/local-registry.md
```diff
@@ -22,6 +22,7 @@ with it enabled.
 set -o errexit
 
 # create registry container unless it already exists 
+CLUSTER_NAME="my-cluster"   # desired cluster name; default is "kind"
```
BenTheElder: maybe just make this `kind`?

BenTheElder: Can punt to a follow up

BenTheElder:
/lgtm
/approve
Thanks!

BenTheElder:
🙄 flakes

site/content/docs/user/local-registry.md
brightzheng100: Sorry you're much faster than I could respond:)
I think having a dummy name like that might draw the right attention that "it's your decision" -- especially with a note there; while naming it `kind` may cause someone think of "is that a keyword"...

Anyway, thanks for merging it!

PR #1120, opened 2019-11-21
Update RBAC to v1, drop PSP

liggitt:
Updates RBAC manifests to v1, drops requirement for beta PSP API to be enabled (clusters wanting to enable PSP will need to add policies to allow default pods)


BenTheElder:
/lgtm
/approve
thanks!

PR #1123, opened 2019-11-23
Set a default config file (.kind.yaml)

pysysops:
I use kind for testing out different things against clusters and different versions. It's also a great way to bring development of kubernetes resources local and be able to clean up the cluster quickly and test configs against different versions. 

One small thing that I often forget is to type `--config .whatever.yaml ` when bringing up a cluster to develop against. This PR adds a default config file path of `.kind.yaml` and will test it's existence before trying to use it. Now I can have a .kind.yaml config at the root of my git projects and bring up a cluster for testing against without the additional `--config` flag.

Any feedback is greatly appreciated.

### DIFF
--- a/pkg/cluster/constants/constants.go
+++ b/pkg/cluster/constants/constants.go
@@ -20,6 +20,9 @@ package constants
 // DefaultClusterName is the default cluster Context name
 const DefaultClusterName = "kind"
 
+// DefaultConfigPath is the default cluster config file path
+const DefaultConfigPath = ".kind.yaml"
+
 /* node role value constants */
 const (
 	// ControlPlaneNodeRoleValue identifies a node that hosts a Kubernetes
--- a/pkg/cluster/provider.go
+++ b/pkg/cluster/provider.go
@@ -35,6 +35,9 @@ import (
 // DefaultName is the default cluster name
 const DefaultName = constants.DefaultClusterName
 
+// DefaultConfig is the default config file path
+const DefaultConfig = constants.DefaultConfigPath
+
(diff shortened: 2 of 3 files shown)


pysysops:
CLA signed

BenTheElder:
This is an interesting idea.

Can you please file an issue to track this feature?
I'd prefer to discuss there before committing to adding a new feature, to seperate discussion of the idea from discussion of the specific implementation.

I'd like to look around at how the ecosystem handles this more before committing to a specific filepath selection logic, and we'd want to think about how to document this and make it non-surprising.

Currently kind generally* does not depend on where it is run

BenTheElder:
(and once we add this, everyone will depend on whatever we choose, and we'll never be able to change it ;-))

BenTheElder:
In your git projects you can also have a small makefile or shell script to run `kind create cluster --config=.kind.yaml --name=my-project` And just run `make kind-cluster` consistently or something like that.

Should we instead do EG config file(s) in `$HOME/.kind/config` or ....

I think the makefile solution would work pretty well right now and is somewhat less magical. 🤔

BenTheElder:
... and thanks for the PR! I neglected to say that earlier 😅

js-timbirkett:
It's an idea to start conversation around. I'm not particularly attached to the idea or it's implementation so any changes, new ideas or canning it altogether are good with me 😄

BenTheElder:
I appreciate the PR and the idea but for now I think we're going to hold off, this definitely seems worth exploring in the future though.

Thanks :-)

PR #1130, opened 2019-11-28
skip restarting the containerd if the systemd is not fully up and running

cofyc:
hi, I use kind (0.6) in CI job (DinD) but it fails with the following error when I try to patch contained configuration.

```
 ✗ Writing configuration 📜 
ERROR: failed to create cluster: failed to restart containerd after patching config: command "docker exec --privileged tidb-operator-worker2 systemctl restart containerd" failed with error: exit status 1

Output:
System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down

Stack Trace: 
	/src/pkg/errors/errors.go:51
	/src/pkg/exec/local.go:116
sigs.k8s.io/kind/pkg/internal/cluster/providers/docker.(*nodeCmd).Run
	/src/pkg/internal/cluster/providers/docker/node.go:130
sigs.k8s.io/kind/pkg/internal/cluster/create/actions/config.(*Action).Execute.func2
	/src/pkg/internal/cluster/create/actions/config/config.go:147
	/src/pkg/errors/concurrent.go:30
runtime.goexit
	/usr/local/go/src/runtime/asm_amd64.s:1357
```

I figured out that at the bootstrap of node container, the PID 1 is not the systemd. Even if it's the systemd, we should wait for the systemd to be fully up and running.  So I patch the kind to make it work stable in CI job.

**Update**: skip restarting the containerd if the systems is not running yet

https://www.freedesktop.org/software/systemd/man/systemctl.html#is-system-running

### DIFF
--- a/pkg/internal/cluster/create/actions/config/config.go
+++ b/pkg/internal/cluster/create/actions/config/config.go
@@ -144,7 +144,8 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 					return errors.Wrap(err, "failed to write patched containerd config")
 				}
 				// restart containerd now that we've re-configured it
-				if err := node.Command("systemctl", "restart", "containerd").Run(); err != nil {
+				// skip if the systemd (also the containerd) is not running
+				if err := node.Command("bash", "-c", `! systemctl is-system-running || systemctl restart containerd`).Run(); err != nil {
 					return errors.Wrap(err, "failed to restart containerd after patching config")
 				}
 				return nil


BenTheElder:
can you tell me more about this environment? we've removed this kind of wait before because it caused issues.

BenTheElder:
actually, in fact we just have a race here, if systemd is not running we don't need to reload

what would be more ideal is if we could tell from the exit code if it was systemd not running yet, in that case we can just skip restarting it

pkg/internal/cluster/create/actions/config/config.go
```diff
@@ -143,6 +144,9 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 				if err := nodeutils.WriteFile(node, containerdConfigPath, patched); err != nil {
 					return errors.Wrap(err, "failed to write patched containerd config")
 				}
+				if err := waitForSystemdFullyUpAndRunning(node, time.Now().Add(5*time.Second)); err != nil {
+					return errors.Wrap(err, "failed to wait for the systemd to be fully up and running")
+				}
 				// restart containerd now that we've re-configured it
 				if err := node.Command("systemctl", "restart", "containerd").Run(); err != nil {
```
BenTheElder: what about making this:
```suggestion
				if err := node.Command("bash", "-c", `! systmctl is-system-running || systemctl restart containerd`).Run(); err != nil {
```

BenTheElder: (or possibly using a check to see if containerd is running, we only need to restart it if it's already running possibly with the unpatched config)

cofyc:
job image is `gcr.io/k8s-testimages/kubekins-e2e:v20191108-9467d02-master`
pod yaml:

```
apiVersion: v1
kind: Pod
spec:
  containers:
  - name: main
    image: gcr.io/k8s-testimages/kubekins-e2e:v20191108-9467d02-master
    command:
    - runner.sh
    - sleep
    - 99d
    # we need privileged mode in order to do docker in docker
    securityContext:
      privileged: true
    env:
    - name: DOCKER_IN_DOCKER_ENABLED
      value: "true"
    resources:
      requests:
        memory: "16000Mi"
        cpu: 8000m
    # kind needs /lib/modules and cgroups from the host
    volumeMounts:
    - mountPath: /lib/modules
      name: modules
      readOnly: true
    - mountPath: /sys/fs/cgroup
      name: cgroup
    # dind expects /var/lib/docker to be volume
    - name: docker-root
      mountPath: /var/lib/docker
    # legacy docker path for cr.io/k8s-testimages/kubekins-e2e
    - name: docker-graph
      mountPath: /docker-graph
  volumes:
  - name: modules
    hostPath:
      path: /lib/modules
      type: Directory
  - name: cgroup
    hostPath:
      path: /sys/fs/cgroup
      type: Directory
  - name: docker-root
    emptyDir: {}
  - name: docker-graph
    emptyDir: {}
```
we run CI job in jenkins with Kubernetes plugin, the Kubernetes version is 1.12.3. Hard to reproduce locally, but it fails in CI job frequently.

pkg/internal/cluster/create/actions/config/config.go
```diff
@@ -143,8 +144,11 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 				if err := nodeutils.WriteFile(node, containerdConfigPath, patched); err != nil {
 					return errors.Wrap(err, "failed to write patched containerd config")
 				}
+				if err := waitForSystemdFullyUpAndRunning(node, time.Now().Add(5*time.Second)); err != nil {
+					return errors.Wrap(err, "failed to wait for the systemd to be fully up and running")
+				}
 				// restart containerd now that we've re-configured it
-				if err := node.Command("systemctl", "restart", "containerd").Run(); err != nil {
+				if err := node.Command("bash", "-c", `! systmctl is-system-running || systemctl restart containerd`).Run(); err != nil {
```
BenTheElder: oops, typoed the first systemctl

we shouldn't need the wait if this is right, though i'm wondering if we should instead structure it like `! containerd-running || restart-containerd`

BenTheElder:
thanks!

we're not using this in CI yet, this race seems not surprising looking at it, thanks for catching this.

definitely think the cheapest and most effective solution is to skip the reload if systemd is not running.

everything else should be fine, the next step we do is run kubeadm which has some reasonable retries etc.
everything else waits until after that.

pkg/internal/cluster/create/actions/config/config.go
cofyc: you're right, I've removed the wait. 

I think `! containerd-running || restart-containerd` should work too.

cofyc:
I've updated the PR and verified that it works in our CI job. Thanks for your suggestion!

BenTheElder:
/lgtm
/approve
thanks!!

Issue #1124, opened 2019-11-25
kindnetd image is not present in the v1.16.3 node image


aojea:
**What happened**:

The kindnetd image is not present on the docker images and the CNI fails

**What you expected to happen**:

Have the CNI kindnetd image inside the nodes images so it works in airgapped environment

**How to reproduce it (as minimally and precisely as possible)**:

Compile the master version a create a cluster `./kind create cluster`

Once it boots check the images inside the container

```
docker exec -it kind-control-plane crictl ps
CONTAINER           IMAGE               CREATED             STATE               NAME                      ATTEMPT             POD ID
cfa147aceaae4       74308f46cf2d3       10 seconds ago      Running             kube-controller-manager   0                   d56de2f644749
92b1fef768cd8       6750aac993260       10 seconds ago      Running             kube-apiserver            0                   2f5bb293fdc41
91f83cf886dff       b2756210eeabf       10 seconds ago      Running             etcd                      0                   3f0fc0e1eab8b
4417271451444       0f7390a0dfb32       11 seconds ago      Running             kube-scheduler            0                   4dc7e7aac03c8
```

kindnetd is not running and the image is not present

```
docker exec -it kind-control-plane crictl images
IMAGE                                TAG                 IMAGE ID            SIZE
k8s.gcr.io/coredns                   1.6.2               bf261d1579144       44.2MB
k8s.gcr.io/etcd                      3.3.15-0            b2756210eeabf       248MB
k8s.gcr.io/kube-apiserver            v1.16.3             6750aac993260       185MB
k8s.gcr.io/kube-controller-manager   v1.16.3             74308f46cf2d3       128MB
k8s.gcr.io/kube-proxy                v1.16.3             2388adcec48cf       103MB
k8s.gcr.io/kube-scheduler            v1.16.3             0f7390a0dfb32       105MB
```

**Anything else we need to know?**:

After pulling the image crictl doesn't show the tag version 🤔, may be a crictl bug???
```
docker exec -it kind-control-plane crictl images
IMAGE                                TAG                 IMAGE ID            SIZE
k8s.gcr.io/kube-controller-manager   v1.16.3             74308f46cf2d3       128MB
k8s.gcr.io/kube-proxy                v1.16.3             2388adcec48cf       103MB
docker.io/kindest/kindnetd           <none>              aa67fec7d7ef7       32.4MB
k8s.gcr.io/coredns                   1.6.2               bf261d1579144       44.2MB
[...]

PR #1132, opened 2019-12-02
CNI portmap plugin fixes and improvements

aojea:
The portmap plugin can use external chains for SNAT

> externalSetMarkChain - string, default nil. If you already have a Masquerade mark chain (e.g. Kubernetes), specify it here. This will use that instead of creating a separate chain. When this is set, markMasqBit must be unspecified.

https://github.com/containernetworking/plugins/tree/v0.8.3/plugins/meta/portmap

v0.8.4 CNI portmap fixes  a bug that was causing flakiness in the k8s CI

It also bumps the base image version with a newer version of iptables user space binaries


pkg/build/node/cni.go
```diff
@@ -27,7 +27,7 @@ const (
 The default CNI manifest and images are our own tiny kindnet
 */
 
-var defaultCNIImages = []string{"kindest/kindnetd:0.5.3@sha256:bc1833b3da442bb639008dd5a62861a0419d3f64b58fce6fb38b749105232555"}
+var defaultCNIImages = []string{"aojea/kindnetd:0.5.4"}
```
neolit123: should a new image be pushed to `kindest/kindnet`?

BenTheElder: yeah, discussed this PR in the kind meeting today, holding it pending #1124 fix and then pushing a small kind release with assorted small fixes after .6

then this, and also upgrading the debian-ipatables base to the new one coming in k8s 1.17 with much newer iptables

aojea:
/hold
https://github.com/kubernetes-sigs/kind/pull/1132#discussion_r352911665

aojea:
/hold cancel
the small release was already done

aojea:
/retest
forget to push the new image for testing

aojea:
/hold
we need to wait for CNI plugins 0.8.4 release

BenTheElder:
- can we downgrade to a version without the bug while we wait for v0.8.4 to be released?

aojea:
> * can we downgrade to a version without the bug while we wait for v0.8.4 to be released?

🤷‍♂ this bug was there since the beginning

BenTheElder:
sorry will need an update, I did some heavy work on deps today which overlapped (but yay updated deps and smaller images!)

please send me a reminder when the CNI plugins is out and we'll try to get this in ASAP to get the flake fix :-)

aojea:
we only need the new portmap plugin, but we have to wait until the CNI folks cut a new release with the fix I did to avoid the concurrency issues that we see in the CI

PR #1144, opened 2019-12-05
upgrade to go 1.13.5

BenTheElder:
The docker image is not available yet, but CI won't pass until it is.

### DIFF
--- a/hack/go_container.sh
+++ b/hack/go_container.sh
@@ -30,7 +30,7 @@ SOURCE_DIR="${SOURCE_DIR:-$(pwd -P)}"
 # default to disabling CGO for easier reproducible builds and cross compilation
 export CGO_ENABLED="${CGO_ENABLED:-0}"
 # the container image, by default a recent official golang image
-GOIMAGE="${GOIMAGE:-golang:1.13.4}"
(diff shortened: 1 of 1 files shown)


tao12345666333:
wait for official docker image https://github.com/docker-library/official-images/pull/7080

BenTheElder:
well even after the PR merges we need to wait for the image to be live, but it's easier to just go ahead and have it a button click away :-)

still tempted to just use our own, it's not much effort to install go into a docker image :P

BenTheElder:
(we should not actually do that, but it is tempting :-))

Nexucis:
maybe it would be worth it to actually automatise the creation of these kind of PR (such as the one I did in the official docker repository) when docker update the golang version

(hello, I'm just a random guys that was passing here :))

BenTheElder:
hey :-)

I thought about that actually, but sometimes we also need to update the docs, and creating these PRs is super low effort, I just did a quick replace in my editor :-)

I think the cost of filing these PRs for the duration of the project will be lower than maintaining automation.

BenTheElder:
(I do also manually check the delta in go versions first!, but I'm looking at those for github.com/kubernetes/kubernetes, kind is just cheaper to update first :sweat_smile: )

PR #1147, opened 2019-12-06
upgrade to containerd 1.3.2

BenTheElder:
this is the latest in the 1.3 release branch

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -23,7 +23,7 @@ FROM ubuntu:19.10
 
 # Configure containerd and runc binaries from kind-ci/containerd-nightlies repository
 # The repository contains latest stable releases and nightlies built for multiple architectures
-ARG CONTAINERD_VERSION="v1.3.0-27-g54658b88"
+ARG CONTAINERD_VERSION="v1.3.2"
 # Configure CNI binaries from upstream
 ARG CNI_VERSION="v0.8.3"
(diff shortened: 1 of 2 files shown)


images/base/Dockerfile
```diff
@@ -23,7 +23,7 @@ FROM ubuntu:19.10
 
 # Configure containerd and runc binaries from kind-ci/containerd-nightlies repository
 # The repository contains latest stable releases and nightlies built for multiple architectures
-ARG CONTAINERD_VERSION="v1.3.0-27-g54658b88"
+ARG CONTAINERD_VERSION="v1.3.2"
```
aojea: We are downloading containerd from our nighlies repository, can check from my mobile if we are releasing this version

BenTheElder: Give me a little credit :-)

BenTheElder: Not only do I actually check the releases when doing this, I bumped the base image to include it, which won't build if the release doesn't exist 🙃

aojea: :man_shrugging: holiday are starting to affect my brain :upside_down_face: 
Sorry

aojea: I was doubting if it was correct because I was missing the short hash appended to the version, I've forgotten that we have 2 jobs in travis to release nightlies, and one is publishing containerd releases without the commit hash appended :sweat_smile:

Issue #1140, opened 2019-12-04
kind load docker-image doesn't include tag when a sha256 is specified


BenTheElder:
`kind load docker image foo:bar@sha256:...` -> the image will only be loaded as the `sha256:...`

This is due to `docker save` intentionally not including the tag metadata when a sha256 is specified.

IMO kind should support this anyhow.

We can inject the metadata parsed from the kind CLI argument into the image before loading it.

/priority important-longterm

BenTheElder:
I think `kind load docker-image` supporting this doesn't make sense after digging into this more.

The sha256 of manifests in particular doesn't appear to be feasible to get both saved and then tagged correctly.

PR #1150, opened 2019-12-07
try to fix https://github.com/kubernetes-sigs/kind/issues/1140

jeffwubj:
inject tag if sha256 is specified

### DIFF
--- a/pkg/cmd/kind/load/docker-image/docker-image.go
+++ b/pkg/cmd/kind/load/docker-image/docker-image.go
@@ -43,6 +43,10 @@ type flagpole struct {
 	Nodes []string
 }
 
+var (
+	shaKey = "@sha256:"
+)
+
 // NewCommand returns a new cobra.Command for loading an image into a cluster
 func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 	flags := &flagpole{}
@@ -120,8 +124,13 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
 
 	// pick only the nodes that don't have the image
 	selectedNodes := []nodes.Node{}
+	tagName := imageName
+	strs := strings.Split(imageName, shaKey)
+	if len(strs) > 0 {
+		tagName = strs[0]
+	}
 	for _, node := range candidateNodes {
-		id, err := nodeutils.ImageID(node, imageName)
+		id, err := nodeutils.ImageID(node, tagName)
 		if err != nil || id != imageID {
 			selectedNodes = append(selectedNodes, node)
 			logger.V(0).Infof("Image: %q with ID %q not present on node %q", imageName, imageID, node.String())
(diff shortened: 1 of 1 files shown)


BenTheElder:
Hi, did you test this?
I did some digging and I'm not sure it's actually possible to make this work correctly in the common case due to how manifests work, I didn't get to finish and update the issue yet though.

jeffwubj:
> Hi, did you test this?
> I did some digging and I'm not sure it's actually possible to make this work correctly in the common case due to how manifests work, I didn't get to finish and update the issue yet though.

Hi,
In my test (on macOS), before modification 
`kind load docker-image docker.io/library/ubuntu:18.04@sha256:017eef0b616011647b269b5c65826e2e2ebddbe5d1f8c1e56b3599fb14fabec8` 
will succeed, but bash into the node, using crictl won't display this image.
after modification, the same command can succeed, and bash into the node, using crictl can display the image
`docker.io/library/ubuntu             18.04               94e814e2efa88       91.4MB`

jeffwubj:
> Hi, did you test this?
> I did some digging and I'm not sure it's actually possible to make this work correctly in the common case due to how manifests work, I didn't get to finish and update the issue yet though.

hi,
feel free to have other solutions, I'm new to kind and want to get familiar with it by fixing ...

I got a test on macOS as follows:
before fix,
`kind load docker-image ubuntu:18.04@sha256:017eef0b616011647b269b5c65826e2e2ebddbe5d1f8c1e56b3599fb14fabec8`
will success but crictl inside node wont show this image
after fix,
`kind load docker-image ubuntu:18.04@sha256:017eef0b616011647b269b5c65826e2e2ebddbe5d1f8c1e56b3599fb14fabec8`
will success and crictl inside node can show the image
`
docker.io/library/ubuntu             18.04               94e814e2efa88       91.4MB
`

pkg/cmd/kind/load/docker-image/docker-image.go
```diff
@@ -166,7 +179,70 @@ func loadImage(imageTarName string, node nodes.Node) error {
...
+	}
+
+	tarFile, err := os.Create(dest)
+	if err != nil {
+		return err
+	}
+	defer tarFile.Close()
+	tw := tar.NewWriter(tarFile)
+	defer tw.Close()
+	saveCmd := exec.Command("docker", "save", image)
```
BenTheElder: this may mean we save the wrong image.

pkg/cmd/kind/load/docker-image/docker-image.go
```diff
@@ -166,7 +179,70 @@ func loadImage(imageTarName string, node nodes.Node) error {
...
+		if err != nil {
+			return err
+		}
+		var content bytes.Buffer
+		if _, err := io.Copy(&content, tr); err != nil {
+			return err
+		}
+		if hdr.Name == "manifest.json" {
+			s := content.String()
+			s = strings.Replace(s, "\"RepoTags\":null", "\"RepoTags\":[\""+tag+"\"]", -1)
```
BenTheElder: when is this actually happening if we're not saving with the sha256?

BenTheElder:
thanks for the PR! but after digging into this more actually closing #1140 as working as intended / wont-fix, this behavior happens upstream for a reason.

PR #1157, opened 2019-12-11
ship a better storage solution

### LINKED ISSUE #1151: Ship a working storageclass

We would like KIND to create a working storageclass by default.

**Why is this needed**:

We build a few Kubernetes operators. We would like to update our documentation so that users use KIND instead of Minikube as a test/dev Kubernetes setup in case they don't have one. Here is an example: https://github.com/stashed/docs/pull/54

One of the blockers for us is that KIND does not come with a working storageclass. We had success with Rancher's local provisioner. This is what we do internally. https://github.com/appscodelabs/e2e-test-setup/blob/kind-v0.6.1/run.sh

Obviously we could point our users to our custom docs for how to setup KIND, but it will be nice it KIND could ship a working storageclass solution.

BenTheElder:
- install a customized rancher.io/local-host-path driver config at image build time (with preloaded images)
- when installing the default storage class, attempt to install from the node instead of from the one in the kind binary first

TODO:
- [x] handle kubernetes < 1.12
  - DONE: for less than 1.12 we will ship the old manifest.
- [x] Push an image

I'll handle those later tonight. I just want to surface progress on this.

Fixes #1151 #118

### DIFF
--- a/pkg/build/node/node.go
+++ b/pkg/build/node/node.go
@@ -392,7 +392,7 @@ func (c *BuildContext) prePullImages(dir, containerID string) error {
 	// later releases use manifest list images
 	// at node boot time we retag our images to handle this where necessary,
 	// so we virtually re-tag them here.
-	ver, err := version.ParseGeneric(rawVersion[0])
+	ver, err := version.ParseSemantic(rawVersion[0])
 	if err != nil {
 		return err
 	}
@@ -431,7 +431,7 @@ func (c *BuildContext) prePullImages(dir, containerID string) error {
 	requiredImages = append(requiredImages, defaultCNIImages...)
 
 	// for v1.12.0+ we support a nicer storage driver
-	if ver.LessThan(version.MustParseGeneric("v1.12.0")) {
+	if ver.LessThan(version.MustParseSemantic("v1.12.0")) {
 		// otherwise, we must use something built in and simpler, which is
 		// also the same as what kind previously used...
 		if err := writeManifest(cmder, legacyDefaultStorage, defaultStorageManifest); err != nil {
--- a/pkg/build/node/node.go
+++ b/pkg/build/node/node.go
@@ -414,18 +414,6 @@ func (c *BuildContext) prePullImages(dir, containerID string) error {
 	builtImages = fixedImages
 	c.logger.V(0).Info("Detected built images: " + strings.Join(builtImages.List(), ", "))
 
-	// write the default CNI manifest
-	if err := writeManifest(cmder, defaultCNIManifestLocation, defaultCNIManifest); err != nil {
-		c.logger.Errorf("Image build Failed! Failed write default CNI Manifest: %v", err)
-		return err
-	}
-
-	// write the default Storage manifest
-	if err := writeManifest(cmder, defaultStorageManifestLocation, defaultStorageManifest); err != nil {
-		c.logger.Errorf("Image build Failed! Failed write default Storage Manifest: %v", err)
-		return err
-	}
-
 	// gets the list of images required by kubeadm
 	requiredImages, err := exec.OutputLines(cmder.Command(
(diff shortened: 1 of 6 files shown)


BenTheElder:
note that this does unfortunately increase the node image size slightly, but having somewhat saner PVs is worth it.

pkg/build/node/node.go
```diff
@@ -447,6 +436,8 @@ func (c *BuildContext) prePullImages(dir, containerID string) error {
 
 	// all builds should isntall the default CNI images currently
 	requiredImages = append(requiredImages, defaultCNIImages...)
+	// all builds should isntall the default storage driver images currently
```
aojea: carrying over the typo 🙃

pkg/build/node/storage.go
```diff
@@ -0,0 +1,147 @@
...
+metadata:
+  name: local-path-config
+  namespace: local-path-storage
+data:
+  config.json: |-
+        {
+                "nodePathMap":[
+                {
+                        "node":"DEFAULT_PATH_FOR_NON_LISTED_NODES",
+                        "paths":["/var/local-path-provisioner"]
```
aojea: /var is a volume if IIRC, does this have any implication?

BenTheElder: Yes, that's why we selected it. It's the appropriate place for persistent data.

BenTheElder: There is a set of notes about the delta above.

pkg/cluster/internal/create/actions/installstorage/storage.go
```diff
@@ -53,7 +55,7 @@ func (a *action) Execute(ctx *actions.ActionContext) error {
 	node := controlPlanes[0] // kind expects at least one always
 
 	// add the default storage class
-	if err := addDefaultStorageClass(node); err != nil {
+	if err := addDefaultStorage(ctx.Logger, node); err != nil {
```
aojea: should we be able to opt-out and not install it as we did with the CNI?
or opt-in and keep the legacy by default?

BenTheElder: Not necessary initially.
You can easily replace the storage class with your own and delete the deployment, unlike CNI.

Let me mark this PR WIP.

BenTheElder: There's no point in opt-in. We ship hermetic images, the images etc. will already be there and just running the deployment is low cost.

And the legacy one is know to have major limitations.

pkg/build/node/node.go
BenTheElder: fixed

BenTheElder:
handled v1.11 node image builds by shipping the old manifest in that case (and not bundling the images)

that's pretty low effort but will let us continue to best effort support v1.11 minimally without much code

BenTheElder:
rebased to upstream and pushed the node image

aojea:
/lgtm 
I couldn't test it though, will be nice to have feedback from users

tamalsaha:
This is great! Looking forward to the next release of Kind.

BenTheElder:
I did a _LOT_ of testing with these. It should work great with the tweaks we've applied for kind :-)

monotek:
Is there a timeline / roadmap / due date for the next kind release?

Very interested in this fix as i hit this with the Elasticsearch (https://github.com/elastic/helm-charts/issues/429) and likely also the Prometheus chart (https://github.com/eclipse/packages/pull/13#issuecomment-573674567).

BenTheElder:
https://github.com/kubernetes-sigs/kind/issues/1250 working on it

nominally we use milestones, but that got a bit out of date over the
holidays.

BenTheElder:
the release is out!

monotek:
Thanks a lot! :-)

PR #1159, opened 2019-12-11
Add user guide for ingress nginx.

amwat:
Create a basic user guide that follows https://kubernetes.github.io/ingress-nginx/deploy/#prerequisite-generic-deployment-command and exposes hostPorts to make it work.


site/content/docs/user/ingress-nginx.md
```diff
@@ -0,0 +1,56 @@
+# Ingress Nginx
```
BenTheElder: we need some front matter metadata at the top

BenTheElder: nit: Nginx -> NGINX, it's same case

site/content/docs/user/ingress-nginx.md
```diff
@@ -0,0 +1,56 @@
...
+
+Ingress Nginx in kind works by exposing ports `80(http)` and `443(https)`
```
BenTheElder: maybe s/works/can be setup/ -- "works" implies this is the only solution or that we bundle a solution.

site/content/docs/user/ingress-nginx.md
```diff
@@ -0,0 +1,56 @@
...
+from the host to the nginx controller using hostPorts.
+
+## Create A Cluster with Ingress Nginx
+
+The following shell script will create a kind cluster deploy
```
BenTheElder: `, deploying` ?

site/content/docs/user/ingress-nginx.md
```diff
@@ -0,0 +1,56 @@
...
+        # schedule it on the control-plane node
+        node-role.kubernetes.io/master: ''
+      tolerations:
+      # tolerate the the control-plane taints
+      - key: node-role.kubernetes.io/master
+        operator: Equal
+        effect: NoSchedule
+EOF
+)"
+```
```
BenTheElder: can we add some basic usage?

we should at least link to the k8s.io ingress docs more

BenTheElder:
WDYT about titling the guide `Ingress` and then at least leaving room to expand it in the future to cover multiple options and general details of dealing with KIND & Ingress?

I'm sure the contour folks would love to pitch in, and we should probably dedupe the content.

site/content/docs/user/ingress.md
```diff
@@ -0,0 +1,144 @@
...
+            serviceName: foo-service
+            servicePort: 5678
+        - path: /bar
+          backend:
+            serviceName: bar-service
+            servicePort: 5678
+---
+EOF
+```
+Now verify that the ingress works
```
BenTheElder: lets add a newline after the code block above before this line

site/content/docs/user/ingress.md
```diff
@@ -0,0 +1,144 @@
...
+```bash
+curl localhost/foo
+curl localhost/bar
+```
```
BenTheElder: maybe follow this with what the expected output is?

site/content/docs/user/ingress.md
```diff
@@ -0,0 +1,144 @@
...
+  main:
+    parent: "user"
+    identifier: "user-ingress"
+    weight: 3
+---
+# Ingress
+
+[Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) in kind can be setup by exposing ports `80(http)` and `443(https)`
+from the host to the ingress controller using hostPorts.
+
```
BenTheElder: Rough suggestion for hierarchy with room for adding other controllers:
```
## Setting Up An Ingress Controller

We can leverage KIND's `extraPortMapping` config option when creating a cluster to
forward ports from the host to an ingress controller running on a node.
The following ingress controllers are known to work:

 - [Ingress NGINX](#ingress-nginx)

### Ingress NGINX

<!--setup details here-->

<!--TODO: subheadings contour etc., add them to the list above as well>

Now you will want to checkout [Using Ingress](#using-ingress)

## Using Ingress
```

site/content/docs/user/ingress.md
```diff
@@ -0,0 +1,144 @@
...
+        ports:
+        - containerPort: 80
+        # Proxy the host port 80 for http
+          hostPort: 80
+        - containerPort: 443
+        # Proxy the host port 443 for https
+          hostPort: 443
+      nodeSelector:
+        # schedule it on the control-plane node
+        node-role.kubernetes.io/master: ''
```
aojea: maybe mention somewhere that this config only works if we have only one control-plane node, but is independent of the number of worker nodes?

amwat: ah yes. good point

BenTheElder: maybe we should do the label thing then /shrug

BenTheElder:
netlify is gone again :(

BenTheElder:
https://deploy-preview-1159--k8s-kind.netlify.com/

site/content/docs/user/ingress.md
```diff
@@ -0,0 +1,157 @@
+---
+title: "Ingress"
+menu:
+  main:
+    parent: "user"
+    identifier: "user-ingress"
+    weight: 3
```
BenTheElder: known issues should have a higher weight

site/content/docs/user/ingress.md
```diff
@@ -0,0 +1,157 @@
...
+Now you will want to checkout [Using Ingress](#using-ingress)
+
+
+## Using Ingress
+
+The following example creates simple http-echo services 
+and an Ingress object to route to these services.
+
+```bash
+cat <<EOF | kubectl apply -f -
```
BenTheElder: this is somewhat unweildy for a cat EOF pipe thing, maybe we should split out the file contents and the apply (and perhaps host the actual contents as a file in static?)

site/content/docs/user/ingress.md
amwat: yeah added labels instead

site/content/docs/user/ingress.md
```diff
@@ -0,0 +1,168 @@
...
+          backend:
+            serviceName: bar-service
+            servicePort: 5678
+---
+```
+
+Apply the contents
+
+```bash
+kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/kind/master/site/content/docs/user/ingress-nginx-example.yaml
```
BenTheElder: if we put this in static/ we can get a URL on the site

maybe static/manifests or something

then https://kind.sigs.k8s.io/manifests/ingress-nginx-example.yaml

later we can template in the site URL

site/content/docs/user/ingress.md
```diff
@@ -0,0 +1,168 @@
...
+
+The following shell script will create a kind cluster, deploy 
+the standard ingress-nginx components, and apply the necessary patches.
+
+```bash 
+#!/bin/sh
+set -o errexit
+
+# Create cluster with hostPorts opened for http and https
+cat <<EOF | kind create cluster --config=-
```
BenTheElder: let's put all of these as files + inline their contents
maybe kind should support config files over HTTPS :thinking:

site/content/docs/user/ingress.md
```diff
@@ -0,0 +1,168 @@
...
+  - containerPort: 443
+    hostPort: 443
+EOF
+
+# Apply the mandatory ingress-nginx components 
+# https://kubernetes.github.io/ingress-nginx/deploy/#prerequisite-generic-deployment-command
+kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/static/mandatory.yaml
+
+# Apply kind specific patches
+kubectl patch deployments -n ingress-nginx nginx-ingress-controller -p "$(cat<<EOF
```
BenTheElder: can kubectl patch take in a URL for the patch?

BenTheElder:
merging and iterating. still would like to see some changes
/lgtm
/approve

site/content/docs/user/ingress.md
```diff
@@ -0,0 +1,167 @@
...
+nodes:
+- role: control-plane
+  kubeadmConfigPatches:
+  - |
+    apiVersion: kubeadm.k8s.io/v1beta2
+    kind: InitConfiguration
+    nodeRegistration:
+      kubeletExtraArgs:
+        node-labels: "ingress-ready=true"
+        authorization-mode: "AlwaysAllow"
```
BenTheElder, line 37: Wait a sec, what's this about? 😅

neolit123: i don't think this is needed for nginx ingress and it's not a recommended practice.

aledbf: I can confirm this is not required or recommended in ingress-nginx

BenTheElder: i suspect this is not needed.

amwat: Yeah, if it's working without it we can probably safely remove this.

PR #1164, opened 2019-12-12
Cleanup Ingress guide

amwat:
Use markdown shortcodes to preview files.

/cc @BenTheElder

PR #1166, opened 2019-12-12
template site baseURL

BenTheElder:
for use when referencing hosted content, this makes it possible to correctly use previews and handle local development properly ...

### DIFF
--- a/site/content/docs/user/ingress.md
+++ b/site/content/docs/user/ingress.md
@@ -156,7 +156,7 @@ spec:
 Apply the contents
 
 ```shell script
-kubectl apply -f https://kind.sigs.k8s.io/manifests/ingress/nginx/example.yaml
+kubectl apply -f {{< absurl "manifests/ingress/nginx/example.yaml" >}}
 ```
 
 Now verify that the ingress works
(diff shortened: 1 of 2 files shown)


BenTheElder:
https://deploy-preview-1166--k8s-kind.netlify.com/docs/user/ingress/#using-ingress

aojea:
it shows 
`https://deploy-preview-1166--k8s-kind.netlify.commanifests/ingress/nginx/example.yaml`

🤔 `commanifests` ???

site/content/docs/user/ingress.md
```diff
@@ -156,7 +156,7 @@ spec:
 Apply the contents
 
 ```shell script
-kubectl apply -f https://kind.sigs.k8s.io/manifests/ingress/nginx/example.yaml
+kubectl apply -f {{< baseurl >}}manifests/ingress/nginx/example.yaml
```
aojea: it's missing one `/` ???

BenTheElder: it's an inconsistency between local dev and netlify build

the correct answer is using the `absURL` method, which I've updated to.

BenTheElder: locally there's a trailing slash, in CI there is not. absURL handles this appropriately.

BenTheElder:
there are also more of these that can be done and we should re-arrange them quickly before people depend on them, but the intent is to just implement this utility for use in https://github.com/kubernetes-sigs/kind/pull/1164, so just using it in one place to verify that it works as intended in CI ... which is fixed now

PR #1171, opened 2019-12-14
flip GINKGO_TOLERATE_FLAKES to no by default

BenTheElder:
https://github.com/kubernetes/test-infra/commit/2aa69a4a3ced429ed37cf7ae9131d74a7bd0e63c#diff-bc99b88e1dd79f8f303ffcf9c8c9e0fcL5

https://github.com/kubernetes/test-infra/pull/15516

### DIFF
--- a/hack/ci/e2e-k8s.sh
+++ b/hack/ci/e2e-k8s.sh
@@ -137,14 +137,10 @@ run_tests() {
 
   # setting this env prevents ginkgo e2e from trying to run provider setup
   export KUBERNETES_CONFORMANCE_TEST='y'
-  # setting these is required to make RuntimeClass tests work ...
+  # setting these is required to make RuntimeClass tests work ... :/
   export KUBE_CONTAINER_RUNTIME=remote
   export KUBE_CONTAINER_RUNTIME_ENDPOINT=unix:///run/containerd/containerd.sock
   export KUBE_CONTAINER_RUNTIME_NAME=containerd
-  # unless we specify otherwise, tolerate flakes :/
-  # TODO: remove this when kubernetes e2e testing no longer tolerates flakes ...
-  # https://github.com/kubernetes/test-infra/blob/608557bbe4342da9895a0762cbdbf8a508ee59e9/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml#L5-L6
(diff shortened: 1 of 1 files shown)


BenTheElder:
cc @aojea (brace for especially bad ipv6 flakes ...)
cc @spiffxp @dims @liggitt (FYI, this was still enabled for kind CI other than conformance testing...)

hack/ci/e2e-k8s.sh
```diff
@@ -137,14 +137,12 @@ run_tests() {
 
   # setting this env prevents ginkgo e2e from trying to run provider setup
   export KUBERNETES_CONFORMANCE_TEST='y'
-  # setting these is required to make RuntimeClass tests work ...
+  # setting these is required to make RuntimeClass tests work ... :/
```
BenTheElder: it is incredibly lame that we still need to do this. we should pull this from the node info or something.
... but that's another problem for another day.

BenTheElder: ... these tests _also_ still require creating a magical `test-handler` runtime ahead of time in your CRI configuration (eg configuring containerd) so :|

something to keep in mind for the three of you working on conformance testing ...

hack/ci/e2e-k8s.sh
```diff
@@ -137,14 +137,12 @@ run_tests() {
...
   export KUBE_CONTAINER_RUNTIME=remote
   export KUBE_CONTAINER_RUNTIME_ENDPOINT=unix:///run/containerd/containerd.sock
   export KUBE_CONTAINER_RUNTIME_NAME=containerd
-  # unless we specify otherwise, tolerate flakes :/
-  # TODO: remove this when kubernetes e2e testing no longer tolerates flakes ...
-  # https://github.com/kubernetes/test-infra/blob/608557bbe4342da9895a0762cbdbf8a508ee59e9/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml#L5-L6
```
BenTheElder: @spiffxp made this change now so :-)

BenTheElder:
/hold cancel
since we already flipped the switch on other jobs

however, I expect this may cause presubmit to klog up in github.com/kubernetes/kubernetes, I'm fairly certain a few test cases do not reliably pass the first time.

BenTheElder:
simplified by simply not setting this now, since the default is already not `y` otherwise.

aojea:
Let the fun start  :-)

Issue #1173, opened 2019-12-17
Cannot start clusters on remote docker servers in 0.6.x without specifying a port


justenwalker:
**What happened**:

Running kind when targeting a docker daemon on a remote host no longer works in 0.6.x when it used to in 0.5.x

Instead, you get an error: 

```
ERROR: failed to create cluster: failed to get port for API server: listen tcp 10.114.35.99:0: bind: cannot assign requested address
```

**What you expected to happen**:

Kind should automatically choose a random port as it has done before.

**How to reproduce it (as minimally and precisely as possible)**:

Given: 
- a remote docker daemon running on 10.114.35.99 (or some other non-local address)
- config.yaml:
```yaml
---
kind: Cluster
apiVersion: kind.sigs.k8s.io/v1alpha3
networking:
  apiServerAddress: 10.114.35.99
```
- Run `kind create cluster -v5 --config config.yaml`

Then: get an error when Preparing Nodes 
```
failed to get port for API server: listen tcp 10.114.35.99:0: bind: cannot assign requested address
ERROR: failed to create cluster: failed to get port for API server: listen tcp 10.114.35.99:0: bind: cannot assign requested address

Stack Trace:
sigs.k8s.io/kind/pkg/errors.Wrap
        /src/pkg/errors/errors.go:39
sigs.k8s.io/kind/pkg/internal/cluster/providers/docker.planCreation.func2
        /src/pkg/internal/cluster/providers/docker/provision.go:84
sigs.k8s.io/kind/pkg/errors.UntilErrorConcurrent.func1
        /src/pkg/errors/concurrent.go:30
runtime.goexit
        /usr/local/go/src/runtime/asm_amd64.s:1357
```

**Anything else we need to know?**:

While reviewing the source, I've found that this only incidentally worked in 0.5.x due to the `getPort` function using a bind-any address instead of specifying the `APIServerAddress` for the bind; Hence, the 0.6.1 bind fails and the 0.5.1 bind succeeds.
  - v0.5.1: https://github.com/kubernetes-sigs/kind/blob/v0.5.1/pkg/cluster/nodes/create.go#L41
  - v0.6.1: https://github.com/kubernetes-sigs/kind/blob/v0.6.1/pkg/internal/cluster/providers/provider/common/getport.go#L34

In both cases, they probe for an available port on the host running the kind binary, which is not the correct host when targeting a remote docker daemon. The 0.5.1 success is misleading, and probably as effective as using `rand`.

[...]

PR #1174, opened 2019-12-17
let docker map to random port for us

BenTheElder:
this fixes https://github.com/kubernetes-sigs/kind/issues/1173


BenTheElder:
satisfied lints I think

mrbobbytables:
will re-lgtm once jobs are done.

justenwalker:
FWIW I also built and ran this locally and it fixes #1173 for me

BenTheElder:
huh weird flake in `pull-kind-e2e-kubernetes` (I don't think related at all)
/retest

aojea:
do we need this code pkg/cluster/internal/providers/provider/common/getport.go after this PR?
or do we want to keep it for future providers?

BenTheElder:
Kept for future providers

PR #1176, opened 2019-12-18
support docker info proxy details

### LINKED ISSUE #1175: proxy info from docker config is not respected
This should be an easy fix, thanks to @sudeshsh for debugging this with me.

We just need to detect and respect https://docs.docker.com/network/proxy/#configure-the-docker-client

passing in env explicitly will still be expected, but docker will automagically plumb through proxy env, just missing the things kind would otherwise append to noProxy.

/assign
/lifecycle active

BenTheElder:
fixes #1175 

I also wound up needing to fix some things along the way to test this properly:
- subpar output from our test helpers (couldn't tell nil map vs empty map)
- bugs versus our interface in cmd helpers
- needed a simple exec fake (will write a nicer one later when I write some more test with it...)

### DIFF
--- a/pkg/cluster/internal/providers/provider/common/proxy.go
+++ b/pkg/cluster/internal/providers/provider/common/proxy.go
@@ -20,6 +20,9 @@ import (
 	"os"
 	"strings"
 
+	"sigs.k8s.io/kind/pkg/errors"
+	"sigs.k8s.io/kind/pkg/exec"
+
 	"sigs.k8s.io/kind/pkg/internal/apis/config"
 )
 
@@ -34,11 +37,11 @@ const (
 
 // GetProxyEnvs returns a map of proxy environment variables to their values
 // If proxy settings are set, NO_PROXY is modified to include the cluster subnets
-func GetProxyEnvs(cfg *config.Cluster) map[string]string {
-	return getProxyEnvs(cfg, os.Getenv)
+func GetProxyEnvs(cfg *config.Cluster) (map[string]string, error) {
+	return getProxyEnvs(cfg, os.Getenv, exec.DefaultCmder)
 }
 
-func getProxyEnvs(cfg *config.Cluster, getEnv func(string) string) map[string]string {
+func getProxyEnvs(cfg *config.Cluster, getEnv func(string) string, cmder exec.Cmder) (map[string]string, error) {
 	envs := make(map[string]string)
 	for _, name := range []string{HTTPProxy, HTTPSProxy, NOProxy} {
 		val := getEnv(name)
@@ -50,6 +53,18 @@ func getProxyEnvs(cfg *config.Cluster, getEnv func(string) string) map[string]st
 			envs[strings.ToLower(name)] = val
 		}
 	}
+
+	// if we didn't detect proxy environment variable settings, probe docker
+	// we prefer the proxy env to allow more explicit per-invocation override
+	// of proxy settings
+	if len(envs) < 1 {
+		e, err := getProxyEnvFromDocker(cmder)
+		if err != nil {
+			return nil, err
+		}
+		envs = e
+	}
+
 	// Specifically add the cluster subnets to NO_PROXY if we are using a proxy
 	if len(envs) > 0 {
 		noProxy := envs[NOProxy]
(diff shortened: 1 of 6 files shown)


pkg/cluster/internal/providers/provider/common/proxy.go
```diff
@@ -50,6 +53,18 @@ func getProxyEnvs(cfg *config.Cluster, getEnv func(string) string) map[string]st
 			envs[strings.ToLower(name)] = val
 		}
 	}
+
+	// if we didn't detect proxy environment variable settings, probe docker
+	// we prefer the proxy env to allow more explicit per-invocation override
+	// of proxy settings
+	if len(envs) < 1 {
+		e, err := getProxyEnvFromDocker(cmder)
```
BenTheElder: TODO: probably re-work this so some providers don't do this? need to think about this more, but that shouldn't be hard and it's academic at the moment.

pkg/cluster/internal/providers/provider/common/proxy.go
```diff
@@ -60,5 +75,35 @@ func getProxyEnvs(cfg *config.Cluster, getEnv func(string) string) map[string]st
...
+
+// obtains proxy information from docker via docker info against cmder
+func getProxyEnvFromDocker(cmder exec.Cmder) (map[string]string, error) {
+	// get raw fields
+	lines, err := exec.OutputLines(
+		cmder.Command(
+			"docker", "info",
+			"--format",
+			// one per line, upper(key)=value
+			"HTTP_PROXY={{.HTTPSProxy}}\nHTTPS_PROXY={{.HTTPSProxy}}\nNO_PROXY={{.NoProxy}}",
```
tao12345666333: find typo it should be  `HTTP_PROXY={{.HTTPProxy}}`

BenTheElder: thanks

BenTheElder:
/hold
sigh. why is this so complicated with docker.
I suppose we may have to identify the config location and parse that too but I'd rather avoid that :|

tao12345666333:
We don't need to cover everything, IMO.

For me personally, due to my network situation, I have to configure a proxy for the docker daemon, but usually this proxy runs on my local `localhost:8081`, and I don't want to proxy this to the cluster created by kind. Because that doesn't make it works well.

BenTheElder:
Probably we should pick up the config.json options instead of the daemon
options then..?

BenTheElder:
Those seem to be purpose built as a way to set the env on containers, but
we want to append to it.

tao12345666333:
Sorry for delay.

As I said just now, because of my network environment (basically the same for most users in China), I need to run a proxy locally. 
I will set the docker daemon use it, this is for the convenience of pulling the image from a registry like gcr.io.

But usually when the container is running, no proxy is needed. Even if I want to use a proxy, it will be configured in `~ /.docker/config`, or passed in through env.



> Probably we should pick up the config.json options instead of the daemon options then..?

For me, yes.

pkg/cluster/internal/providers/provider/common/proxy.go
aojea: can it be an interface `getProxyEnvFromProvider`?

pkg/cluster/internal/providers/provider/common/proxy.go
```diff
@@ -60,5 +75,35 @@ func getProxyEnvs(cfg *config.Cluster, getEnv func(string) string) map[string]st
...
+
+// obtains proxy information from docker via docker info against cmder
+func getProxyEnvFromDocker(cmder exec.Cmder) (map[string]string, error) {
+	// get raw fields
+	lines, err := exec.OutputLines(
+		cmder.Command(
+			"docker", "info",
+			"--format",
+			// one per line, upper(key)=value
+			"HTTP_PROXY={{.HTTPProxy}}\nHTTPS_PROXY={{.HTTPSProxy}}\nNO_PROXY={{.NoProxy}}",
```
aojea: is this format stable?
This is what I have in my local machine
```
$ docker info --format "HTTP_PROXY={{.HTTPProxy}}\nHTTPS_PROXY={{.HTTPSProxy}}\nNO_PROXY={{.NoProxy}}"
HTTP_PROXY=\nHTTPS_PROXY=\nNO_PROXY=
$
```

pkg/cluster/internal/providers/provider/common/proxy.go
```diff
@@ -60,5 +75,35 @@ func getProxyEnvs(cfg *config.Cluster, getEnv func(string) string) map[string]st
...
+		),
+	)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get proxy information from docker")
+	}
+	// parse out environment settings from the format we specified
+	envs := make(map[string]string)
+	for _, line := range lines {
```
aojea: we should check that the number of lines is 3 and the names of the variables `[]string{HTTPProxy, HTTPSProxy, NOProxy}`
 to be more robust.
If docker outputs whatever message instead of the variables we are going to return this wrong info

pkg/exec/helpers.go
```diff
@@ -62,8 +62,8 @@ func RunErrorForError(err error) *RunError {
 // stderr + stdout, it scans these for lines and returns a slice of output lines
 func CombinedOutputLines(cmd Cmd) (lines []string, err error) {
 	var buff bytes.Buffer
-	cmd.SetStdout(&buff)
-	cmd.SetStderr(&buff)
+	cmd = cmd.SetStdout(&buff)
```
aojea: just curious, we need the assignment to make the code testable with the new `fakeCmder`?

tao12345666333:
> is better to document that you MUST use environment variables if you want to use a proxy instead of open this new can of worms ;) ?

yes, I think so. providing available solutions (use environment variables) is simpler and more effective than covering all possibilities

pkg/cluster/internal/providers/provider/common/proxy.go
BenTheElder: - we control the forms
- error output goes to stderr, not stdout

pkg/exec/helpers.go
BenTheElder: At one point we did. This is more correct regardless.

aojea:
ahhg is parallel, the issue is only happening on serial :man_facepalming:

BenTheElder:
see the issue, this needs a different solution.

PR #1177, opened 2019-12-18
document extra port mapping

BenTheElder:

PR #1178, opened 2019-12-18
reorganize config docs, document networking options

BenTheElder:
adds to #1177 

documents cluster wide networking options


aojea:
/lgtm
nice

PR #1182, opened 2019-12-18
Add Message When No Clusters Exist

danielhelfand:
This pull request adds a simple message of `No clusters available` from `kind get clusters` if no clusters have been created. I think this is more helpful to an end user than returning no message, but I would be open to changing the message if what I have is not clear.

### DIFF
--- a/pkg/cmd/kind/get/clusters/clusters.go
+++ b/pkg/cmd/kind/get/clusters/clusters.go
@@ -50,8 +50,12 @@ func runE(logger log.Logger, streams cmd.IOStreams) error {
 	if err != nil {
 		return err
 	}
-	for _, cluster := range clusters {
-		fmt.Fprintln(streams.Out, cluster)
+	if len(clusters) > 0 {
+		for _, cluster := range clusters {
+			fmt.Fprintln(streams.Out, cluster)
+		}
+	} else {
+		logger.Warn("No clusters available")
 	}
 	return nil
 }


pkg/cmd/kind/get/clusters/clusters.go
```diff
@@ -50,8 +50,12 @@ func runE(logger log.Logger, streams cmd.IOStreams) error {
...
 		return err
 	}
-	for _, cluster := range clusters {
-		fmt.Fprintln(streams.Out, cluster)
+	if len(clusters) > 0 {
+		for _, cluster := range clusters {
+			fmt.Fprintln(streams.Out, cluster)
+		}
+	} else {
+		fmt.Fprintln(streams.Out, "No clusters available")
```
BenTheElder: This should not go to stdout. It's a user facing message, not program output.
We could use Logger.Warn maybe.

BenTheElder:
Ack on the concept 👍

pkg/cmd/kind/get/clusters/clusters.go
danielhelfand: Sure, I just changed to `logger.Warn` in latest push.

BenTheElder:
thanks, I have some bikeshed thoughts for the exact message, but those need not block the PR, this is a net improvement :-)

danielhelfand:
@BenTheElder Sounds good! Thanks for the review.

PR #1183, opened 2019-12-18
disable apparmor confinement for kind nodes

BenTheElder:
cc @tamalsaha


aojea:
/lgtm
hehe, we keep it doing more secure 🙃

PR #1187, opened 2019-12-19
prevent pulling node image that locally exists when using only a SHA

dustinspecker:
Before, running `kind create cluster --image
kindest/node@sha256:70ce6ce09bee5c34ab14aec2b84d6edb260473a60638b1b095470a3a0f95ebec`
would always attempt to pull `kindest/node` which would fail. This is
happening because of the conversion of the image to a friendly image
name is stripping off the SHA portion. If a tag isn't provided then this
causes kind to erroneously attempt to re-pull the image even if the
image with the SHA exists locally.

Now, always explicitly use the original image and not the friendly
image name. This enables the `docker inspect` command to work properly.

### DIFF
--- a/pkg/cluster/internal/providers/docker/images.go
+++ b/pkg/cluster/internal/providers/docker/images.go
@@ -35,10 +35,11 @@ func ensureNodeImages(logger log.Logger, status *cli.Status, cfg *config.Cluster
 	// pull each required image
 	for _, image := range common.RequiredNodeImages(cfg).List() {
 		// prints user friendly message
+		friendlyImageName := image
 		if strings.Contains(image, "@sha256:") {
-			image = strings.Split(image, "@sha256:")[0]
+			friendlyImageName = strings.Split(image, "@sha256:")[0]
 		}
-		status.Start(fmt.Sprintf("Ensuring node image (%s) 🖼", image))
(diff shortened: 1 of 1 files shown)


pkg/cluster/internal/providers/docker/images.go
```diff
@@ -35,21 +35,22 @@ func ensureNodeImages(logger log.Logger, status *cli.Status, cfg *config.Cluster
 	// pull each required image
 	for _, image := range common.RequiredNodeImages(cfg).List() {
 		// prints user friendly message
+		friendlyImageName := image
```
dustinspecker: I opted to keep the friendlyImageName around, but another solution to this is to remove the original logic that removed the sha256 portion. Does mean all the messages/logs would display the full SHA. What are folks' thoughts? I'm happy to adjust this in any way.

BenTheElder:
The friendly image name is only meant for display (the line with the emoji)

dustinspecker:
Ah, okay. Thank you. I'll remove passing it around tonight.

BenTheElder:
this bug is actually really affecting me trying to work from a poor network right now ironically :(, I'm not sure when we introduced it but I'm going to patch up your PR and get it in

BenTheElder:
thanks a ton for fixing this, this was a dumb bug :-)

PR #1193, opened 2019-12-20
Dependency cleanup

BenTheElder:
I upgraded all the upstream binary dependencies and took a sledgehammer to unnecessary content in the images.

Base image is something like 100MB smaller now by removing unneeded content.
Everything is much more up to date.

Cleaned up scripts for managing base image in preparation for standardizing non-node-image build and removing the relatively useless `kind build base-image`.

This should help push back against the eternal binary bloat problem.

will update with a node image bump as soon as that's pushed.


BenTheElder:
This is a breaking change for anyone depending on arbitrary content being in the node image, but we do NOT support this.

It should not be a breaking change versus older kind binaries.

BenTheElder:
:shipit: 
doesn't shrink the node image as much as you'd want but, it helps.

we still need to lighten up kubernetes core.

Issue #1154, opened 2019-12-10
Add new k8s IPv6 configuration improvements to KIND


aojea:
**What would you like to be added**:

Currently we have to overcome kubernetes IPv6 configuraiton limitations in kind, i.e, hardcoding node ip 

- [x] https://github.com/kubernetes/kubernetes/pull/85850

- [x] https://github.com/kubernetes/kubernetes/pull/84727

**Why is this needed**:

This will simplify the Kind glue and test better upstream IPv6 support on k8s

We can get rid of the NodeAddress dependency
https://github.com/kubernetes-sigs/kind/blob/4c7854c1347860ef04a7a8caef625ac4491afa4b/pkg/cluster/internal/kubeadm/config.go#L43

https://github.com/kubernetes-sigs/kind/blob/4c7854c1347860ef04a7a8caef625ac4491afa4b/pkg/cluster/internal/kubeadm/config.go#L408

https://github.com/kubernetes-sigs/kind/blob/4c7854c1347860ef04a7a8caef625ac4491afa4b/pkg/cluster/internal/kubeadm/config.go#L401

BenTheElder:
that'll be nice :-)

BenTheElder:
we can still add this, we just need to guard on the k8s version, right?

aojea:
> we can still add this, we just need to guard on the k8s version, right?

yeah, see https://github.com/kubernetes-sigs/kind/pull/1195 for reference

aojea:
:/

aojea:
/close

Fixed by #1154

PR #1195, opened 2019-12-20
autodetect advertises address with IPv6

aojea:
apiserver and kubelet advertise ipv6 endpoints when binding to ipv6 since 1.17

This removes the need to hardcode the kubelet and kube-apiserver IPs when using IPv6

Fixes : #1154 

~Unblocks https://github.com/kubernetes/kubernetes/pull/86480~


liggitt:
/hold

it's unclear https://github.com/kubernetes/kubernetes/pull/86480 is a compatible change... we should probably wait for a decision there before merging this

aojea:
> /hold
> 
> it's unclear [kubernetes/kubernetes#86480](https://github.com/kubernetes/kubernetes/pull/86480) is a compatible change... we should probably wait for a decision there before merging this

yeah, thanks for the clarification, I've also added the WIP to this PR, 
I'm finding IP families inconsistencies here too :/

aojea:
2/2 failure of the same 2 tests, will investigate later

aojea:
It's failing because the kube-scheduler and the kube-controller-manager are being restarted constantly because they have the wrong http probes configured https://github.com/kubernetes/kubernetes/pull/86493

aojea:
🤔  

https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_kind/1195/pull-kind-conformance-parallel-ipv6/1213414923934109697/artifacts/logs/kind-control-plane/containers/kube-apiserver-kind-control-plane_kube-system_kube-apiserver-b5b1764ec4d32744472affd4653f94243e70947b6b5b888eb9a3b837e9035147.log

> 2020-01-04T11:13:47.951628264Z stderr F I0104 11:13:47.951353       1 server.go:596] external host was not specified, using 172.17.0.4

> 2020-01-04T11:13:48.948109526Z stderr F I0104 11:13:48.947888       1 rest.go:113] the default service ipfamily for this cluster is: IPv6

aojea:
/hold cancel
all the changes needed were merged and we can benefit from the apiserver advertise address autodetection.
just only remains the kubelet to autodetect the node-ip family and we can get rid of hardcoding the node-address
https://github.com/kubernetes/kubernetes/pull/85850
/assign @BenTheElder

BenTheElder:
Is this going to break with older versions of kubernetes then...?

aojea:
> Is this going to break with older versions of kubernetes then...?

it will break only with `kubeadm.k8s.io/v1beta2` , the commit didn't modify previous kubeadm templates so you can still deploy IPv6 clusters with `kubeadm.k8s.io/v1beta1

Also, I'm not sure about this, but it may be possible to patch the preconfigured templates with a config file, no?

aojea:
/hold
let's wait for 
https://github.com/kubernetes/kubernetes/pull/85850

BenTheElder:
That's not good enough, we shouldn't break any released k8s.

This needs to be version gated.

aojea:
this is no autodetecting the node-ip using `::` as parameter of kubelet because the node name doesn't resolve the IPv6 address, using the following code:

```
package main

import (
        "fmt"
        "net"
        "os"
        "runtime"
)

func main() {
        fmt.Println("GOOS:", runtime.GOOS)
        // Resolver
        addrs, _ := net.LookupIP(os.Args[1])
        fmt.Println("net.LookupIP addrs:", addrs)

}
```

we can confirm it

```
 /lookup kind-control-plane
GOOS: linux
net.LookupIP addrs: [172.17.0.2]
```

aojea:
it was fixed in docker recently by this https://github.com/docker/libnetwork/pull/2445/commits/e3bacd69cc4a768f23015135ba85129634b5fe85

aojea:
/hold
it seems we should wait for a new release of docker with the patch mentioned before included
If I understood docker releases correctly, it should be v19.03.6

BenTheElder:
I don't think we can really rely on docker that new for our users?

aojea:
you are absolutely right ... I was so focused in this PR and chasing this odd behavior that forget these things 😅 
/close

BenTheElder:
fixed in https://github.com/kubernetes-sigs/kind/pull/1521

Issue #1213, opened 2019-12-30
Better UX for using local registry


h7kanna:
**What would you like to be added**:

Add a special node to the Kind cluster with 'registry' role. This node can run docker registry without the user having to run any extra scripts to configure Kind cluster

**Why is this needed**:

This can be an alternative to 'kind load docker-image' and naturally integrates with 'docker push' CI workflows.

Also, the registry storage can be mounted from the host and kind load can be avoided for faster setup.

BenTheElder:
We're still designing this, there are some tradeoffs that depend on how the networking changes to fix https://github.com/kubernetes-sigs/kind/issues/148

The obvious "registry" role has some downsides, including that you'll need to push to each cluster.

diafour:
I choose not to run a local registry in-cluster. It is not great for testing against multiple versions and/or clusters.
So I run local registry as a standalone docker container with mounted volume in a home directory. Kind cluster is started with a small script, that does two things after the cluster is ready: it adds `localhost:5000` as a trusted registry to /etc/containerd/config.toml and creates a NodePort Service and a DaemonSet to proxying from that NodePort to docker host IP.

That is an adopted solution back from `kubeadm-dind-cluster` days: https://github.com/kubernetes-retired/kubeadm-dind-cluster/issues/56#issuecomment-387463386

BenTheElder:
thanks for the input!

> ... it adds localhost:5000 as a trusted registry to /etc/containerd/config.toml ...

you can do this at boot time now, xref: https://kind.sigs.k8s.io/docs/user/local-registry/

> ... and creates a NodePort Service and a DaemonSet to proxying from that NodePort to docker host IP. ...

That's a neat approach, thanks :-)

diafour:
Yeah, that `config.toml` override was for 0.5 release. Thanks for the link!

diafour:
Hooray! I've just realized that `endpoint` should not be the same as `mirror` name and simplified my setup: no proxying with DaemonSet required, just `containerdConfigPatches`:
https://gist.github.com/diafour/13cef191b7cf39543393d310dd6353a0

The workflow is like this:
```

docker build -t "localhost:5000/name:tag" .
docker push localhost:5000/name:tag

kind: Pod
...
    image: localhost:5000/name:tag

cluster-15.sh create
kubectl replace ... 
```

There are `create` and `delete` commands to address #148 ;)

@BenTheElder Thanks again!

nicks:
[...]

PR #1214, opened 2019-12-30
WIP: kind-registry node

h7kanna:
This is a demo implementation for #1213


BenTheElder:
Thanks for the PR but I'd like to:
- finish reworking the networking in kind which will affect this
- discuss an agreed detailed design with the community before adding this feature

Will keep this demo in mind :-)

Issue #1215, opened 2019-12-30
Update kind-with-registry.sh to work out of the box.


finn-ball:
https://kind.sigs.k8s.io/docs/user/local-registry/

When following these instructions and create the shell script but get:

```
Error: No such object: kind-registry
```

**What would you like to be documented**:

I'd like the documentation to be updated to reflect how to create a local registry that can be used with Kind.

**Why is this needed**:

For easier local development.

BenTheElder:
fixed by https://github.com/kubernetes-sigs/kind/pull/1220

marcindulak:
FYI: there still seem to be an issue when pulling images from the insecure local registry configured by  `kind-with-registry.sh` https://github.com/containerd/cri/issues/1367

marcindulak:
> FYI: there still seem to be an issue when pulling images from the insecure local registry configured by `kind-with-registry.sh` [containerd/cri#1367](https://github.com/containerd/cri/issues/1367)

It looks like already from the containerd version up in https://hub.docker.com/layers/kindest/node/v1.16.2/images/sha256-2c68d327c2833fa9c0f81b5fd36499cf969646cd50affecd21b4725d37931e21 the problem of pulling from insecure registry has been fixed.

```sh
docker exec -it kind-control-plane bash -c "containerd -v"
containerd github.com/containerd/containerd v1.3.0-27-g54658b88 54658b8831a7382bc79bf0b11ca591951c56aa6b
```

BenTheElder:
sorry about that, yeah we should be on 1.3.2+ now for future images

PR #1220, opened 2020-01-02
Update kind-with-registry.sh to work out of the box

marcindulak:
This PR fixes two problems which prevent " kind-with-registry.sh to work out of the box" https://github.com/kubernetes-sigs/kind/issues/1215

1. the script exited due to `set -o errexit` when no `kind-registry` docker container was present

 ```docker inspect -f '{{.State.Running}}' kind-registry
Error: No such object: kind-registry
```

2. the script used `kind` as the default cluster name, which collided with the default. The registry cluster is now started with the name of `"${KIND_CLUSTER_NAME}"`

### DIFF
--- a/site/static/examples/kind-with-registry.sh
+++ b/site/static/examples/kind-with-registry.sh
@@ -7,7 +7,7 @@ KIND_CLUSTER_NAME="${KIND_CLUSTER_NAME:-kind}"
 # create registry container unless it already exists
 reg_name='kind-registry'
 reg_port='5000'
-running="$(docker inspect -f '{{.State.Running}}' "${reg_name}")"
+running="$(docker inspect -f '{{.State.Running}}' "${reg_name}" 2>/dev/null || true)"
 if [ "${running}" != 'true' ]; then
   docker run \
     -d --restart=always -p "${reg_port}:5000" --name "${reg_name}" \
@@ -15,7 +15,7 @@ if [ "${running}" != 'true' ]; then
 fi
 
 # create a cluster with the local registry enabled in containerd
-cat <<EOF | kind create cluster --config=-
+cat <<EOF | kind create cluster --name "${KIND_CLUSTER_NAME}" --config=-
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 containerdConfigPatches: 

### CODE REVIEW

File: site/static/examples/kind-with-registry.sh [hunk 1]
```diff
@@ -7,15 +7,15 @@ KIND_CLUSTER_NAME="${KIND_CLUSTER_NAME:-kind}"
...
+running="$(docker inspect -f '{{.State.Running}}' "${reg_name}" 2>/dev/null || :)"
 if [ "${running}" != 'true' ]; then
   docker run \
     -d --restart=always -p "${reg_port}:5000" --name "${reg_name}" \
     registry:2
 fi
 
 # create a cluster with the local registry enabled in containerd
-cat <<EOF | kind create cluster --config=-
+cat <<EOF | kind create cluster --name "${reg_name}" --config=-
```
BenTheElder: NACK on this change, `${reg_name}` is the name of the registry container, NOT the cluster.

you can avoid collosion by:
- setting `KIND_CLUSTER_NAME`
- modifying the script

site/static/examples/kind-with-registry.sh
```diff
@@ -29,4 +29,4 @@ ip_fmt='{{.NetworkSettings.IPAddress}}'
 cmd="echo $(docker inspect -f "${ip_fmt}" "${reg_name}") registry >> /etc/hosts"
 for node in $(kind get nodes --name "${KIND_CLUSTER_NAME}"); do
   docker exec "${node}" sh -c "${cmd}"
-done
\ No newline at end of file
+done
```
BenTheElder: this adds unnecessary empty space when rendered in the docs and isn't necessary to parse
for now I'd prefer to not add this

site/static/examples/kind-with-registry.sh
```diff
@@ -7,15 +7,15 @@ KIND_CLUSTER_NAME="${KIND_CLUSTER_NAME:-kind}"
 # create registry container unless it already exists
 reg_name='kind-registry'
 reg_port='5000'
-running="$(docker inspect -f '{{.State.Running}}' "${reg_name}")"
+running="$(docker inspect -f '{{.State.Running}}' "${reg_name}" 2>/dev/null || :)"
```
BenTheElder: :+1: thanks
`true` might be a little better known than `:`

site/static/examples/kind-with-registry.sh
marcindulak: I see now, my understanding of `KIND_CLUSTER_NAME` was incorrect. I though it was a secondary cluster created just for the purpose of demonstrating the registry.

site/static/examples/kind-with-registry.sh
marcindulak: eol should be removed now

site/static/examples/kind-with-registry.sh
marcindulak: : changed to true

BenTheElder:
/lgtm
/approve
/ok-to-test

thank you very much!

PR #1222, opened 2020-01-03
Capitalize Command Descriptions

danielhelfand:
Currently, commands/subcommands are not consistent in terms of whether or not the beginning of the description is capitalized. This pull request changes all lower case description beginnings to capitalized. 

This inconsistency also exists for flags, and I would be happy to follow up with something to also correct those flags. Could do that in this pr too if that would be easier.


aojea:
unrelated failure
thanks LGTM
/assign @BenTheElder

BenTheElder:
/lgtm
/approve
this is consistent with kubectl

PR #1226, opened 2020-01-04
refactor the internal logger and action out to a lib

djgilcrease:
I needed something almost identical to the cluster create actions you use in kind so I created lib for them @ https://gitlab.com/digitalxero/simple-actions based 99% on the actions and logger/status system here in kind.

This is just a PR to switch the internal actions to use that lib as well as add a dry run option to kind

If you do not want to switch feel free to just close this PR.


BenTheElder:
Appreciate the PR but I'd rather not take on dependencies where we can avoid it, kind's dependency set is minimal, things we need for compatibility.

I also expect to eliminate the action abstraction for this particular use case (I can see it being useful elsewhere...) and the logger library will likely evolve more before kind hits 1.0.

Thanks but I think we'll go a different direction :sweat_smile:

PR #1236, opened 2020-01-08
Cleanup Meaningless Variable Declaration

vanou:
This commit cleans up meaningless variable declaration in which underscore (blank identifier) is used as
variable identifier.
And this declaration makes no effect in source code.

So this declaration can be safely removed.

### DIFF
--- a/pkg/internal/cli/logger.go
+++ b/pkg/internal/cli/logger.go
@@ -40,8 +40,6 @@ type Logger struct {
 	isSmartWriter bool
 }
 
-var _ log.Logger = &Logger{}
-
 // NewLogger returns a new Logger with the given verbosity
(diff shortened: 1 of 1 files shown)


BenTheElder:
That's not true. This is a compile time type assertion.

pkg/internal/cli/logger.go
```diff
@@ -40,8 +40,6 @@ type Logger struct {
 	isSmartWriter bool
 }
 
-var _ log.Logger = &Logger{}
-
```
BenTheElder: This ensures that Logger implements log.Logger.
This is a common idiom in go.

vanou:
Thanks @BenTheElder . I understand the intention of this.

BenTheElder:
Thanks for the PR anyhow 🙃

PR #1239, opened 2020-01-09
Bump cni version with portmap fix

aojea:
It includes the fix to the portmap plugin that was showing up in the k8s CI


BenTheElder:
I was just looking for this last night how convenient! :D

PR #1247, opened 2020-01-12
site: Add Contour Ingress Controller as example in /docs/user/ingress

stevesloka:
This adds to the Ingress examples by showing how to deploy with Contour (projectcontour.io). 

Signed-off-by: Steve Sloka <steve@stevesloka.com>

### DIFF
--- a/site/content/docs/user/ingress.md
+++ b/site/content/docs/user/ingress.md
@@ -23,13 +23,17 @@ in the kubeadm `InitConfiguration`, to be used
 by the ingress controller `nodeSelector`.
 
 
-The following ingress controllers are known to work:
+1. [Create a cluster](#create-cluster)
+2. Deploy an Ingress controller, the following ingress controllers are known to work:
+    - [Contour](#contour)
+    - [Ingress NGINX](#ingress-nginx)
 
- - [Ingress NGINX](#ingress-nginx)
-
-### Ingress NGINX
+### Create Cluster
 
 Create a kind cluster with `extraPortMappings` and `node-labels`.
+
+- **extraPortMappings** allow the local host to make requests to the Ingress controller over ports 80/443
+- **node-labels** only allow the ingress controller to run on a specific node(s) matching the label selector
 
 {{< codeFromInline lang="bash" >}}
 cat <<EOF | kind create cluster --config=-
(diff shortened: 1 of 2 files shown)


amwat:
https://deploy-preview-1247--k8s-kind.netlify.com/docs/user/ingress/#contour
/retest

site/content/docs/user/ingress.md
```diff
@@ -25,8 +25,40 @@ by the ingress controller `nodeSelector`.
...
 The following ingress controllers are known to work:
 
+ - [Contour](#contour)
  - [Ingress NGINX](#ingress-nginx)
 
+### Contour
+
+Create a kind cluster with `extraPortMappings` which will expose ports 80 & 443 from your cluster to your local host.
+
+{{< codeFromInline lang="bash" >}}
```
amwat: This part is similar to the ingress nginx portmappings? Can we merge them
You'll also want to use node labels for making it work in multi node clusters.

stevesloka: The configs are similar but Contour doesn't require those extra mappings which is why I split the configs apart. That said, having them for Contour won't break it, just not needed to get it running with Kind. 

How would this work with multi-node clusters in kind? You can't have additional ports map to 80/443?

amwat: What I meant is, in the multi-node case, those node-labels and the deployments patches are required to ensure that the controller gets scheduled correctly to the node which has the ports exposed.

For example I just tried running contour ingress against a cluster with some additional nodes and it doesn't work properly.

```
kind: Cluster
nodes:
- role: control-plane
  extraPortMappings:
  - containerPort: 80
    hostPort: 80
    protocol: TCP
  - containerPort: 443
    hostPort: 443
    protocol: TCP
- role: control-plane
- role: worker
- role: worker
EOF
```

stevesloka: Ahhh ok, I'm following now. =) Let me fix this up.

stevesloka:
@amwat I did a little bit of re-org from your review comments. Please take a look. =)

site/content/docs/user/ingress.md
```diff
@@ -55,6 +59,35 @@ nodes:
...
+```
+
+Apply it by running:
+
+{{< codeFromInline lang="bash" >}}
+kubectl patch daemonsets -n projectcontour envoy -p '{{< minify file="static/examples/ingress/contour/patch.json" >}}' 
+{{< /codeFromInline >}}
+
+Now the Contour is all setup to be used. 
+Refer the [Using Ingress](#using-ingress) for a basic example usage.
```
BenTheElder: S/Refer the/Refer to/ ?

amwat:
/lgtm
Thanks!

BenTheElder:
/approve
@amwat mind fixing the minor typo when you file to fix the patching on older k8s as discussed in slack?

BenTheElder:
Thanks @stevesloka !

PR #1258, opened 2020-01-15
Drop apiversion from kubeadmconfigpatch.

amwat:
Let kubeadm figure out the apiversion itself so we have backwards compatibility.

/cc @BenTheElder

### DIFF
--- a/site/content/docs/user/ingress.md
+++ b/site/content/docs/user/ingress.md
@@ -43,7 +43,6 @@ nodes:
 - role: control-plane
   kubeadmConfigPatches:
   - |
-    apiVersion: kubeadm.k8s.io/v1beta2
     kind: InitConfiguration
     nodeRegistration:
       kubeletExtraArgs:
@@ -82,7 +81,7 @@ kubectl patch daemonsets -n projectcontour envoy -p '{{< minify file="static/exa
 {{< /codeFromInline >}}
 
 Now the Contour is all setup to be used. 
-Refer the [Using Ingress](#using-ingress) for a basic example usage.
+Refer to [Using Ingress](#using-ingress) for a basic example usage.
 
 Additional information about Contour can be found at: [projectcontour.io](https://projectcontour.io)
 


BenTheElder:
/lgtm
/approve
thanks!

BenTheElder:
> Let kubeadm figure out the apiversion itself so we have backwards compatibility.

aside: what actually happens is _kind_ selects the API version for kubeadm regardless based on the kubernetes version, and then the patch either matches or doesn't

since this field is stable between versions so far, we don't need the version (ignoring alpha API versions for older k8s which is a different problem but those are going to unsupported anyhow...)

tao12345666333:
```
Jan 15 23:18:08.370: INFO: >>> kubeConfig: /root/.kube/kind-test-config
Jan 15 23:18:08.373: INFO: Waiting up to 30m0s for all (but 0) nodes to be schedulable
Jan 15 23:18:08.391: INFO: Condition Ready of node kind-worker is false, but Node is tainted by NodeController with [{node.kubernetes.io/not-ready  NoSchedule <nil>} {node.kubernetes.io/not-ready  NoExecute 2020-01-15 23:18:02 +0000 UTC}]. Failure
Jan 15 23:18:08.391: INFO: Condition Ready of node kind-worker2 is false instead of true. Reason: KubeletNotReady, message: [runtime network not ready: NetworkReady=false reason:NetworkPluginNotReady message:Network plugin returns error: cni plugin not initialized, Failed to initialize CSINodeInfo: error updating CSINode annotation: timed out waiting for the condition; caused by: nodes "kind-worker2" not found]
Jan 15 23:18:08.391: INFO: Unschedulable nodes:
Jan 15 23:18:08.391: INFO: -> kind-worker Ready=false Network=false Taints=[{node.kubernetes.io/not-ready  NoSchedule <nil>} {node.kubernetes.io/not-ready  NoExecute 2020-01-15 23:18:02 +0000 UTC}] NonblockingTaints:node-role.kubernetes.io/master
Jan 15 23:18:08.391: INFO: -> kind-worker2 Ready=false Network=false Taints=[{node.kubernetes.io/not-ready  NoSchedule <nil>}] NonblockingTaints:node-role.kubernetes.io/master
```

/retest

BenTheElder:
that's not the actual error, that's just CSI being noisy during bringup

BenTheElder:
Prow just never reported ...?
/shrug

PR #1259, opened 2020-01-15
Expose logger to external writer

Anuj2512:
Expose the kind logger to external writer so that other modules can use the logging functionality

### DIFF
--- a/pkg/cmd/logger.go
+++ b/pkg/cmd/logger.go
@@ -36,6 +36,15 @@ func NewLogger() log.Logger {
 	return cli.NewLogger(writer, 0)
 }
 
+// NewLoggerWithWriter returns the standard logger used by the kind CLI
+// This logger writes to given external writter
+func NewLoggerWithWriter(writer io.Writer) log.Logger {
+	if env.IsSmartTerminal(writer) {
+		writer = cli.NewSpinner(writer)
+	}
+	return cli.NewLogger(writer, 0)
+}
+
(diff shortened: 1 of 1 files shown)


pkg/cmd/logger.go
```diff
@@ -36,6 +36,15 @@ func NewLogger() log.Logger {
 	return cli.NewLogger(writer, 0)
 }
 
+// NewLoggerWithWriter returns the standard logger used by the kind CLI
+// This logger writes to given external writter
+func NewLoggerWithWriter(writer io.Writer) log.Logger {
+	if env.IsSmartTerminal(writer) {
+		writer = cli.NewSpinner(writer)
+	}
+	return cli.NewLogger(writer, 0)
```
BenTheElder: what writer do you intend to pass to this?

you can already implement the logger interface, this particular implemenation makes assumptions about writing to a console :thinking:

Anuj2512: This is something for automation and for passing the logs to UI during the automation.

In my case the writer is custom go writer which is listening on a channel to read the content and pass the content to the UI through web-sockets.

Let me know if you need more details about the use-case. Writer is smart enough to handle the logs written to console.

Anuj2512: I got it. Thanks for the review. I use the logger interface to handle my scenario as you mentioned.

Thanks for the time.

BenTheElder: i think the logger interface is indeed the way to go here 😅 
thanks

Issue #1235, opened 2020-01-08
Feature gates does not apply


tamalsaha:
**What happened**:
I am trying to create a Kind 1.11.10 cluster with rancher/local-path-provisioner. Because, we want to test our operator for Openshift 3.11 which is K8s 1.11 api. From Rancher's project, we see that we need to enable `DynamicProvisioningScheduling` feature flag.

https://github.com/rancher/local-path-provisioner/blob/v0.0.1/README.md#prepare-kubernetes-cluster

**What you expected to happen**:
I checked the logs kind. I don't see the features gates being applied. Also, checked the flags for the kube-apiserver process which is also missing the flags.

**How to reproduce it (as minimally and precisely as possible)**:

You can see my kind config here: https://github.com/tamalsaha/k8s-111-local-path-demo/blob/master/hack/kubernetes/kind.yaml

```
$ git clone https://github.com/tamalsaha/k8s-111-local-path-demo.git
$ cd k8s-111-local-path-demo

$ kind create cluster --config hack/kubernetes/kind.yaml --image kindest/node:v1.11.10 -v 10
$ kubectl apply -f https://github.com/rancher/local-path-provisioner/raw/v0.0.11/deploy/local-path-storage.yaml
$ kubectl apply -f hack/kubernetes/pod.yaml 
```

If the pod stays is pending state, PVC provisioner is not working.

**Anything else we need to know?**:

**Environment:**

- kind version: (use `kind version`): v0.6.1
- Kubernetes version: (use `kubectl version`): 1.11.10
- Docker version: (use `docker info`):
- OS (e.g. from `/etc/os-release`): Ubuntu 18.04

tamalsaha:
Logs from kind create command:

```
$ kind create cluster --config hack/kubernetes/kind.yaml --image kindest/node:v1.11.10 -v 10

Creating cluster "kind" ...
DEBUG: docker/images.go:58] Image: kindest/node:v1.11.10 present locally
 ✓ Ensuring node image (kindest/node:v1.11.10) 🖼
 ✓ Preparing nodes 📦 
DEBUG: config/config.go:90] Using kubeadm config:
api:
  advertiseAddress: 172.17.0.4
  bindPort: 6443
apiServerCertSANs:
- localhost
- 127.0.0.1
apiServerExtraVolumes:
- hostPath: /etc/nsswitch.conf
  mountPath: /etc/nsswitch.conf
  name: nsswitch
  pathType: FileOrCreate
[...]

PR #1261, opened 2020-01-16
Use rancher/local-path-provisioner:v0.0.12

tamalsaha:
This adds local volume support for k8s v1.11.x .
xref: https://github.com/kubernetes-sigs/kind/issues/1235#issuecomment-572023673

Signed-off-by: Tamal Saha <tamal@appscode.com>

### DIFF
--- a/pkg/build/node/storage.go
+++ b/pkg/build/node/storage.go
@@ -25,7 +25,7 @@ NOTE: we have customized it in the following ways:
 - install as the default storage class
 */
 
-var defaultStorageImages = []string{"rancher/local-path-provisioner:v0.0.11", "k8s.gcr.io/debian-base:v2.0.0"}
+var defaultStorageImages = []string{"rancher/local-path-provisioner:v0.0.12", "k8s.gcr.io/debian-base:v2.0.0"}
 
 const defaultStorageManifest = `
 # kind customized https://github.com/rancher/local-path-provisioner manifest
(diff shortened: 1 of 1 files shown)


BenTheElder:
/ok-to-test
exciting!

tamalsaha:
I can send a separate pr to document the patch needed for local PV working in 1.11 . https://github.com/tamalsaha/k8s-111-local-path-demo/blob/master/hack/kubernetes/kind.yaml#L20-L40

Where should it go?

BenTheElder:
it might be reasonable to just flip those feature gates in the config kind generates for v1.11 by default, given that we no longer test Kubernetes 1.11 anyhow this will only affect users trying to test kubernetes apps for which this is probably desirable to have on by default?

we also need to change one or two other bits for v1.11 to pick this up IIRC

tamalsaha:
>  this is probably desirable to have on by default?

That will be really great for users.

BenTheElder:
in the messy build code .. this branch should be dropped: https://github.com/kubernetes-sigs/kind/blob/2e853a85cce9986f5d7fd1eab9202ba96e6bb93f/pkg/build/node/node.go#L433

and then we can put those config options directly in:
https://github.com/kubernetes-sigs/kind/blob/2e853a85cce9986f5d7fd1eab9202ba96e6bb93f/pkg/cluster/internal/kubeadm/config.go#L83

per:
https://github.com/kubernetes-sigs/kind/blob/2e853a85cce9986f5d7fd1eab9202ba96e6bb93f/pkg/cluster/internal/kubeadm/config.go#L450-L451

BenTheElder:
will file a follow-up
/lgtm
/approve
Thanks!!

PR #1262, opened 2020-01-16
enable dynamic PVs in Kubernetes 1.11

BenTheElder:
follow up to https://github.com/kubernetes-sigs/kind/pull/1261

### DIFF
--- a/pkg/cluster/internal/kubeadm/config.go
+++ b/pkg/cluster/internal/kubeadm/config.go
@@ -134,6 +134,20 @@ nodeRegistration:
     node-ip: "{{ .NodeAddress }}"
 networking:
   podSubnet: "{{ .PodSubnet }}"
+# we need this to allow dynamic PVs to work properly
+apiServerExtraArgs:
+  "feature-gates": "DynamicProvisioningScheduling=true"
+controllerManagerExtraArgs:
+  "feature-gates": "DynamicProvisioningScheduling=true"
+schedulerExtraArgs:
+  "feature-gates": "DynamicProvisioningScheduling=true"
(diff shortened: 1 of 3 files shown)


BenTheElder:
thanks @tamalsaha 
/cc @aojea @amwat

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -134,6 +134,20 @@ nodeRegistration:
     node-ip: "{{ .NodeAddress }}"
 networking:
   podSubnet: "{{ .PodSubnet }}"
+# we need this to allow dynamic PVs to work properly
+apiServerExtraArgs:
+  "feature-gates": "DynamicProvisioningScheduling=true"
```
aojea: the docs says it was superseded by `VolumeScheduling`
https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

BenTheElder: this config is for kubernetes v1.11 :upside_down_face:

BenTheElder: it was deprecated in 1.12

aojea: ahh, but only after 1.12 

> DynamicProvisioningScheduling: Extend the default scheduler to be aware of volume topology and handle PV provisioning. This feature is superseded by the VolumeScheduling feature completely in v1.12.

tamalsaha:
Thanks a lot @BenTheElder ! I know you just cut a release for kind. If it is not too much trouble, can we get a release with these changes?

PR #1255, opened 2020-01-14
[WIP][DO NOT MERGE] Test dual stack PR

aojea:
just the Dual Stack PR with an additional commit that forces the IP family to dual stack so we can test it in the CI

PR #1263, opened 2020-01-17
add featuregates to config

BenTheElder:
This allows doing:
```yaml
kind: Cluster
featureGates:
  AllBeta: true
```

instead of writing many kubeadm config patches.
featuregates are frequently toggled for testing in the broader kubernetes CI

/cc @amwat @liggitt

### DIFF
--- a/pkg/apis/config/v1alpha4/types.go
+++ b/pkg/apis/config/v1alpha4/types.go
@@ -31,6 +31,12 @@ type Cluster struct {
 	// Networking contains cluster wide network settings
 	Networking Networking `yaml:"networking,omitempty"`
 
+	// FeatureGates contains a map of Kubernetes feature gates to whether they
+	// are enabled. The feature gates specified here are passed to all Kubernetes components as flags or in config.
+	//
+	// https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/
+	FeatureGates map[string]bool `yaml:"featureGates,omitempty"`
+
 	// KubeadmConfigPatches are applied to the generated kubeadm config as
 	// merge patches. The `kind` field must match the target object, and
 	// if `apiVersion` is specified it will only be applied to matching objects.
--- a/pkg/cluster/internal/create/actions/config/config.go
+++ b/pkg/cluster/internal/create/actions/config/config.go
@@ -77,6 +77,7 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 		ServiceSubnet:        ctx.Config.Networking.ServiceSubnet,
 		ControlPlane:         true,
 		IPv6:                 ctx.Config.Networking.IPFamily == "ipv6",
+		FeatureGates:         ctx.Config.FeatureGates,
 	}
 
 	kubeadmConfigPlusPatches := func(node nodes.Node, data kubeadm.ConfigData) func() error {
@@ -176,6 +177,23 @@ func getKubeadmConfig(cfg *config.Cluster, data kubeadm.ConfigData, node nodes.N
 	}
 	data.KubernetesVersion = kubeVersion
 
+	// TODO: gross hack!
+	// identify node in config by matching name (since these are named in order)
+	// we should really just streamline the bootstrap code and maintain
+	// this mapping ... something for the next major refactor
+	var configNode *config.Node
+	namer := common.MakeNodeNamer("")
+	for i := range cfg.Nodes {
+		n := &cfg.Nodes[i]
+		nodeSuffix := namer(string(n.Role))
+		if strings.HasSuffix(node.String(), nodeSuffix) {
+			configNode = n
+		}
+	}
+	if configNode == nil {
+		return "", errors.Errorf("failed to match node %q to config", node.String())
+	}
+
 	// get the node ip address
 	nodeAddress, nodeAddressIPv6, err := node.IP()
 	if err != nil {
@@ -201,17 +219,11 @@ func getKubeadmConfig(cfg *config.Cluster, data kubeadm.ConfigData, node nodes.N
 		return "", err
 	}
 
-	// since we only need the last portion of the name,
-	// create namer without a clusterName
-	namer := common.MakeNodeNamer("")
-	for _, inode := range cfg.Nodes {
-		nodeSuffix := namer(string(inode.Role))
-		// if needed, apply current node's patches
-		if strings.HasSuffix(node.String(), nodeSuffix) && (len(inode.KubeadmConfigPatches) > 0 || len(inode.KubeadmConfigPatchesJSON6902) > 0) {
-			patchedConfig, err = patch.KubeYAML(patchedConfig, inode.KubeadmConfigPatches, inode.KubeadmConfigPatchesJSON6902)
-			if err != nil {
-				return "", err
-			}
+	// if needed, apply current node's patches
+	if len(configNode.KubeadmConfigPatches) > 0 || len(configNode.KubeadmConfigPatchesJSON6902) > 0 {
+		patchedConfig, err = patch.KubeYAML(patchedConfig, configNode.KubeadmConfigPatches, configNode.KubeadmConfigPatchesJSON6902)
+		if err != nil {
+			return "", err
 		}
 	}
 
--- a/pkg/cluster/internal/kubeadm/config.go
+++ b/pkg/cluster/internal/kubeadm/config.go
@@ -18,6 +18,8 @@ package kubeadm
 
 import (
 	"bytes"
+	"fmt"
+	"sort"
 	"strings"
 	"text/template"
 
@@ -48,7 +50,8 @@ type ConfigData struct {
 	// The subnet used for services
 	ServiceSubnet string
 	// IPv4 values take precedence over IPv6 by default, if true set IPv6 default values
-	IPv6 bool
+	IPv6         bool
+	FeatureGates map[string]bool
 	// DerivedConfigData is populated by Derive()
 	// These auto-generated fields are available to Config templates,
 	// but not meant to be set by hand
@@ -60,6 +63,10 @@ type ConfigData struct {
 type DerivedConfigData struct {
 	// DockerStableTag is automatically derived from KubernetesVersion
 	DockerStableTag string
+	// SortedFeatureGateKeys allows us to iterate FeatureGates deterministically
+	SortedFeatureGateKeys []string
+	// FeatureGatesString is of the form `Foo=true,Baz=false`
+	FeatureGatesString string
 }
 
 // Derive automatically derives DockerStableTag if not specified
@@ -67,6 +74,22 @@ func (c *ConfigData) Derive() {
 	if c.DockerStableTag == "" {
 		c.DockerStableTag = strings.Replace(c.KubernetesVersion, "+", "_", -1)
 	}
+
+	// get sorted list of FeatureGate keys
+	featureGateKeys := make([]string, 0, len(c.FeatureGates))
+	for k := range c.FeatureGates {
+		featureGateKeys = append(featureGateKeys, k)
+	}
+	sort.Strings(featureGateKeys)
+	c.SortedFeatureGateKeys = featureGateKeys
+
+	// create a sorted key=value,... string of FeatureGates
+	var featureGates []string
+	for _, k := range featureGateKeys {
+		v := c.FeatureGates[k]
+		featureGates = append(featureGates, fmt.Sprintf("%s=%t", k, v))
+	}
+	c.FeatureGatesString = strings.Join(featureGates, ",")
 }
 
 // See docs for these APIs at:
@@ -125,8 +148,6 @@ kubeletConfiguration:
       nodefs.available: "0%"
       nodefs.inodesFree: "0%"
       imagefs.available: "0%"
-controllerManagerExtraArgs:
-  enable-hostpath-provisioner: "true"
 nodeRegistration:
   criSocket: "/run/containerd/containerd.sock"
   kubeletExtraArgs:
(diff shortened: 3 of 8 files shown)


pkg/cluster/internal/create/actions/config/config.go
```diff
@@ -176,6 +177,23 @@ func getKubeadmConfig(cfg *config.Cluster, data kubeadm.ConfigData, node nodes.N
 	}
 	data.KubernetesVersion = kubeVersion
 
+	// TODO: gross hack!
```
BenTheElder: this is basically the same gross hack, and now somewhat orthogonal to this PR, but it is slightly more robust now ...

pkg/apis/config/v1alpha4/types.go
```diff
@@ -31,6 +31,12 @@ type Cluster struct {
 	// Networking contains cluster wide network settings
 	Networking Networking `yaml:"networking,omitempty"`
 
+	// FeatureGates contains a map of Kubernetes feature gates to wether they
```
liggitt:
```suggestion
	// FeatureGates contains a map of Kubernetes feature gates to whether they
```

pkg/apis/config/v1alpha4/types.go
```diff
@@ -31,6 +31,12 @@ type Cluster struct {
...
+	// are enabled
```
liggitt:
```suggestion
	// are enabled. The feature gates specified here are passed to all Kubernetes components as flags or in config.
```

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -125,29 +135,29 @@ kubeletConfiguration:
...
 controllerManagerExtraArgs:
-  "feature-gates": "DynamicProvisioningScheduling=true"
+  "feature-gates": "{{ .FeatureGatesString }}"
+  enable-hostpath-provisioner: "true"
 schedulerExtraArgs:
-  "feature-gates": "DynamicProvisioningScheduling=true"
+  "feature-gates": "{{ .FeatureGatesString }}"
 nodeRegistration:
   kubeletExtraArgs:
-    "feature-gates": "DynamicProvisioningScheduling=true"
```
BenTheElder: this needs to get defaulted
/hold

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -60,13 +62,21 @@ type ConfigData struct {
...
 }
 
 // Derive automatically derives DockerStableTag if not specified
 func (c *ConfigData) Derive() {
 	if c.DockerStableTag == "" {
 		c.DockerStableTag = strings.Replace(c.KubernetesVersion, "+", "_", -1)
 	}
+
+	var featureGates []string
+	for k, v := range c.FeatureGates {
```
liggitt: sort key names to be deterministic?

pkg/internal/apis/config/types.go
```diff
@@ -29,6 +29,12 @@ type Cluster struct {
 	// Networking contains cluster wide network settings
 	Networking Networking
 
+	// FeatureGates contains a map of Kubernetes feature gates to wether they
+	// are enabled
+	//
+	// https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/
+	FeatureGates map[string]bool `yaml:"featureGates,omitempty"`
```
liggitt: drop the yaml tag?

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -445,12 +481,19 @@ evictionHard:
   nodefs.available: "0%"
   nodefs.inodesFree: "0%"
   imagefs.available: "0%"
+{{if .FeatureGates}}featureGates:
+{{ range $key, $value := .FeatureGates }}
```
liggitt: not sure if this needs to be deterministically ordered, but if it's a map it won't be

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -251,12 +265,19 @@ evictionHard:
...
```
liggitt: non-deterministic

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -125,29 +135,29 @@ kubeletConfiguration:
...
 nodeRegistration:
   kubeletExtraArgs:
-    "feature-gates": "DynamicProvisioningScheduling=true"
-kubeProxy:
+    "feature-gates": "{{ .FeatureGatesString }}"
+{{if .FeatureGates}}kubeProxy:
   config:
     featureGates:
-      DynamicProvisioningScheduling: true
+{{ range $key, $value := .FeatureGates }}
```
liggitt: non-deterministic

pkg/apis/config/v1alpha4/types.go
```diff
@@ -31,6 +31,12 @@ type Cluster struct {
...
+	//
+	// https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/
+	FeatureGates map[string]bool `yaml:"featureGates,omitempty"`
```
neolit123: there is a common separation between k8s feature gates and kubeadm feature gates.
usually the kubeadm feature gates can hold features only relevant to kubeadm.
also they are not the preferred mechanic for feature graduation on our side.

would it make sense to also support kubeadm feature gates within the same iteration?

```
kubernetesFeatureGates
kubeadmFeatureGates
```

BenTheElder: right now I'm focused on kubernetes component feature gates (motivation is testing with AllBeta=false AllAlpha=false ...), currently it looks like the only kubeadm feature gate documented is related to ipv6 dualstack, which kind will probably choose to toggle based on networking.ipFamily ?

neolit123: yes, kubeadm does not have any interesting feature gates to kind currently.
the alternative for potential future support is:
```
featureGates
kubeadmFeatureGates
```

BenTheElder: this outmodes the kludge at https://kind.sigs.k8s.io/docs/user/quick-start/#enable-feature-gates-in-your-cluster

pkg/cluster/internal/kubeadm/config.go
BenTheElder: whee, ACK

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -60,13 +63,36 @@ type ConfigData struct {
...
 }
 
 // Derive automatically derives DockerStableTag if not specified
 func (c *ConfigData) Derive() {
 	if c.DockerStableTag == "" {
 		c.DockerStableTag = strings.Replace(c.KubernetesVersion, "+", "_", -1)
 	}
+
+	// get sorted list of FeatureGate keys
+	featureGateKeys := make([]string, len(c.FeatureGates))
```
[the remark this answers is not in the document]
BenTheElder: ACK - done

BenTheElder:
> failed to try resolving symlinks in path "/var/log/pods/test-pods_0c1dcd5e-38cb-11ea-9d38-92dcf75bc5bd_18aa6b32-38cb-11ea-b2a9-42010a800040/test/0.log": lstat /var/log/pods/test-pods_0c1dcd5e-38cb-11ea-9d38-92dcf75bc5bd_18aa6b32-38cb-11ea-b2a9-42010a800040/test/0.log: no such file or directory

Exciting prow flake
/retest
/hold cancel

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -125,29 +148,29 @@ kubeletConfiguration:
       nodefs.available: "0%"
       nodefs.inodesFree: "0%"
       imagefs.available: "0%"
-controllerManagerExtraArgs:
-  enable-hostpath-provisioner: "true"
```
liggitt: related/intentional?

liggitt:
Just the one question about the removal of the enable-hostpath-provisioner arg

/lgtm
/hold 
unhold at will

pkg/cluster/internal/kubeadm/config.go
BenTheElder: It moved down to merge with the feature gates. A previous commit toggling DynamicProvisionScheduling failed to combine these.

liggitt: Ah, missed it. Surprised strict parsing didn't catch the duplicate key

BenTheElder: We have strict parsing on our own config but not on any upstream k8s yet as far as I know.

pkg/apis/config/v1alpha4/types.go
aojea: well , the DualStack job fails because of this
https://github.com/kubernetes-sigs/kind/pull/1255
kubaedm DualStack feature gate pass the feature-gate parameter to all the components, but since this merged this is override because it configures
```
  extraArgs:
    feature-gates: ""
```

based on this I think that I will switch in dual stack to the new config and not patch kubaedm, what do you think?

aojea: this is how it looks like now the dualstack feature gate
https://github.com/kubernetes-sigs/kind/blob/9405b2570d081f92da9534ca644ae03e4dd876e2/hack/ci/e2e-k8s.sh#L86-L97

liggitt: Yeah, we should only output these here if feature gates is non empty.

aojea: that was my first thought, but will it merge both feature-gates in case that they are enabled in both places?
I mean, if I configure 
```
 featureGates: 
   EndpointSlice: true 
 kubeadmConfigPatches: 
 - | 
   kind: ClusterConfiguration 
   metadata: 
     name: config 
   featureGates: 
     IPv6DualStack: true 
 EOF 
```

will I have endpointslices and dual-stack, or the former will override the later as it is doing now?

liggitt: That's a good question. If kubeadm already has a config option to set kubernetes feature gates uniformly, making kind use that option under the covers (instead of building a feature-gates flag override) would probably make sense. I'm not familiar enough with kubeadm to know what that option is or how it handles merges

neolit123: if the user has:
```
kind: ClusterConfiguration
featureGates: 
     IPv6DualStack: true 
apiServer:
    extraArgs:
        "feature-gates": "EndpointSlice=true"
```

the resulting flag for the kube-apiserver will be `--feature-gates="EndpointSlice=true"`

BenTheElder: - kubeadm does not have a uniform feature gate option.
- I'm reworking the config generation to not emit the feature gate args in kubeadm config if the kind option is not set.
- using patches for feature-gates is equally as powerful but more verbose, you cannot stack these because it's a string rather than a map (except in component configs)

liggitt: > I'm reworking the config generation to not emit the feature gate args in kubeadm config if the kind option is not set.

That's sensible.

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -464,6 +520,13 @@ func Config(data ConfigData) (config string, err error) {
 	// assume the latest API version, then fallback if the k8s version is too low
 	templateSource := ConfigTemplateBetaV2
 	if ver.LessThan(version.MustParseSemantic("v1.12.0")) {
+		// in Kubernetes 1.11 we need to enable this feature, unless the user
+		// explicitly disabled it. this is considered part of our "base config"
+		// so it's totally not a layering violation, we just don't want this logic
+		// encoded in the text template ;-)
+		if _, set := data.FeatureGates["DynamicProvisioningScheduling"]; !set {
+			data.FeatureGates["DynamicProvisioningScheduling"] = true
```
liggitt: this means the kind option will always be set on old versions and will stomp manual feature gate patches, right? probably not a big deal for older versions but something to keep in mind for future internal auto feature gate enablement

pkg/apis/config/v1alpha4/types.go
BenTheElder: > that was my first thought, but will it merge both feature-gates in case that they are enabled in both places?
I mean, if I configure 
```
 featureGates: 
   EndpointSlice: true 
 kubeadmConfigPatches: 
 - | 
   kind: ClusterConfiguration 
   metadata: 
     name: config 
   featureGates: 
     IPv6DualStack: true 
 EOF 
```
> will I have endpointslices and dual-stack, or the former will override the later as it is doing now?

So amusingly these are the docs:
> - "featureGates" still exists under ClusterConfiguration, but there are no supported feature gates in 1.13.
> https://godoc.org/k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm/v1beta1

If you read the code though there are at least two feature gates.

If you tell kubeadm to enable `IPv6DualStack` then it will set the Kubernetes components feature gate flags with this **only if the user has not already set the flag** so you need to set the ipv6 feature gate at the kind level if you want both.

E.G. https://github.com/kubernetes/kubernetes/blob/f4b6b751cdf2d46940355359443b263cb883d006/cmd/kubeadm/app/phases/kubelet/flags.go#L124-L128

This is actually easier for kind to handle properly now. What should happen instead is:
```
kind: Cluster
networking:
  ipfamily: ipv4ipv6-dualstack # or whatever we want to call this
```
=> internally default ipv6 featuregate to enabled at the kind config level => passed to all components explicitly.

And users should stop relying on patching the kubeadm configs in N different places to do the equivilant of the new featuregate field.

pkg/cluster/internal/kubeadm/config.go
BenTheElder: I think antonio was using the _kubeadm_ feature gate flags which is an additional related way to enable only at most a few specific features. 

A user supplied patch to component feature gates should still be applied on top of this, stomping this value.
Kubeadm feature gates on the other hand that happen to cause defaulting Kubernetes component feature gates are stomped by user supplied feature gate values (including when kind does this).



We do:
- generate base kubeadm config in kind internally, based on kind's cluster config, including this new feature gates field
- take user kubeadm config patches and apply them on top, trumping any genrated config

In general relying on patching kubeadm config is not the most stable but does provide a way to apply overrides to what kind is doing, which is not the most stable plan long term. 

We should highlight that better.

BenTheElder: though we won't stop the kubeadm feature gates for one particular case if we don't set an explicit emptystring :+)

pkg/apis/config/v1alpha4/types.go
aojea: +100

Issue #1260, opened 2020-01-16
kind delete clusters --all Doesn't Account for Different Cluster kubeconfig Paths


danielhelfand:
**What would you like to be added**:

The current implementation of `kind delete clusters --all` only takes into account [one kubeconfig path](https://github.com/kubernetes-sigs/kind/pull/1204#discussion_r367097369) at a time. It would be nice to take into account if clusters created by kind have different kubeconfig paths. 

**Why is this needed**:

The purpose of `kind delete clusters --all` is to delete all clusters, not just clusters associated with a particular kubeconfig. It's by no means a major issue to only delete all clusters by a particular kubeconfig, and I believe this should satisfy most users of kind. 

However, it would be nice to support both options of deleting all by a kubeconfig and actually deleting all clusters regardless of kubeconfig path.

danielhelfand:
Closing this as of this [comment](https://github.com/kubernetes-sigs/kind/pull/1264#discussion_r368183399). If there is a need for a true delete all approach where kubeconfig does not need to be provided explicitly, then this can be reopened.

webern:
> Closing this as of this [comment](https://github.com/kubernetes-sigs/kind/pull/1264#discussion_r368183399). If there is a need for a true delete all approach where kubeconfig does not need to be provided explicitly, then this can be reopened.

I found my way here because a command that does the following would be convenient:

```
kind get clusters | xargs -t -n1 kind delete cluster --name
```

BenTheElder:
Note that a cluster may actually have multiple kubeconfig on the host (see kind export) or this may be done by other tools + kind get.

The rules for which Kubeconfig file is modified based on the inputs follow the official client library (client-go) that nearly all tools are using (we've actually reimplemented this subset of logic because that is a heavy dependency and we only need to manipulate these files). 

Additionally: we don't record where you chose to export Kubeconfig to (which you can do repeatedly into multiple files for a single cluster) and I don't think this is worth the complexity on our end -- the happy path is using a single file and selecting clusters by context. Power users doing something that depends on multiple files could e.g. place all of them in a common directory and rm the contents when deleting the clusters.

webern:
> The rules for which Kubeconfig file is modified

I wouldn't expect the deletion of a cluster to modify the kubeconfig. I wouldn't notice if it did or didn't, in fact. I might be a little weird in that:
- I never have more than one cluster in a kubeconfig (we hates it precious).
[...]

PR #1264, opened 2020-01-17
Edit kind delete clusters --all Description

danielhelfand:
Minor change to the flag description `--all` for `kind delete clusters`. Just being a bit more explicit about what it does at the moment.

See #1260 for more information.

### DIFF
--- a/pkg/cmd/kind/delete/clusters/deleteclusters.go
+++ b/pkg/cmd/kind/delete/clusters/deleteclusters.go
@@ -48,7 +48,7 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 		},
 	}
 	cmd.Flags().StringVar(&flags.Kubeconfig, "kubeconfig", "", "sets kubeconfig path instead of $KUBECONFIG or $HOME/.kube/config")
-	cmd.Flags().BoolVar(&flags.All, "all", false, "delete all clusters")
+	cmd.Flags().BoolVar(&flags.All, "all", false, "delete all nodes (containers) labeled as kind cluster and kubeconfig entries for the nodes")
 	return cmd
 }
 
@@ -58,7 +58,7 @@ func deleteClusters(logger log.Logger, flags *flagpole, clusters []string) error
 	)
 	var err error
 	if flags.All {
-		//Delete all clusters
+		// Delete all clusters
 		if clusters, err = provider.List(); err != nil {
 			return errors.Wrap(err, "failed listing clusters for delete")
 		}


neolit123:
I haven't check the new flag. But I am going to assume it does not delete
all clusters for a kubeconfig but rather all kind clusters associated with
it.

danielhelfand:
@neolit123 Yes, you are correct. Apologies, I was unfamiliar with the distinction in how clusters are referred to throughout the CLI.

pkg/cmd/kind/delete/clusters/deleteclusters.go
```diff
@@ -48,7 +48,7 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 		},
 	}
 	cmd.Flags().StringVar(&flags.Kubeconfig, "kubeconfig", "", "sets kubeconfig path instead of $KUBECONFIG or $HOME/.kube/config")
-	cmd.Flags().BoolVar(&flags.All, "all", false, "delete all clusters")
+	cmd.Flags().BoolVar(&flags.All, "all", false, "delete all kind clusters associated with a kubeconfig")
```
BenTheElder: I think that's even more confusing.
it will delete all kind clusters, just with the asterisk that if you:
`kind export kubeconfig` to some random paths
OR
`kind create cluster --kubeconfig=/some-random-path`

then `kind delete clusters` will not be touching those files, we don't centrally track this.

whereas in the simpler world of `kind create cluster` pre 0.6 `kind delete cluster` was guaranteed to remove the file.

given the `kind export kubeconfig` case, I think maybe we can just close https://github.com/kubernetes-sigs/kind/issues/1260 as WAI

if you set `--kubeconfig` on this command it will remove the matching clusters in that file.

you can leverage this to do `kind delete clusters --all` repeatedly with different `--kubeconfig` to clean all of them (and only the first time remove containers unless more are created)

danielhelfand: Sure, that's what I originally had so I'll just change it back to remove `kind clusters`. 

I can link to your post comment here and close #1260. If someone has a better approach down the road, at least thoughts around it are documented in the closed issue.

pkg/cmd/kind/delete/clusters/deleteclusters.go
```diff
@@ -48,7 +48,7 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 		},
 	}
 	cmd.Flags().StringVar(&flags.Kubeconfig, "kubeconfig", "", "sets kubeconfig path instead of $KUBECONFIG or $HOME/.kube/config")
-	cmd.Flags().BoolVar(&flags.All, "all", false, "delete all clusters")
+	cmd.Flags().BoolVar(&flags.All, "all", false, "delete all clusters associated with a kubeconfig")
```
BenTheElder: to me this reads like it only deletes clusters that are in a kubeconfig which is incorrect.

this will delete any nodes in the backend labeled as a kind cluster and then any kubeconfig entries that match the containers in the specified config.

danielhelfand: Okay, I have updated the description to be more specific about what is actually taking place. Let me know if it reads a little better. Thanks as always for the help.

pkg/cmd/kind/delete/clusters/deleteclusters.go
```diff
@@ -48,7 +48,7 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 		},
 	}
 	cmd.Flags().StringVar(&flags.Kubeconfig, "kubeconfig", "", "sets kubeconfig path instead of $KUBECONFIG or $HOME/.kube/config")
-	cmd.Flags().BoolVar(&flags.All, "all", false, "delete all clusters")
+	cmd.Flags().BoolVar(&flags.All, "all", false, "delete all nodes (containers) labeled as kind cluster and kubeconfig entries for the nodes")
```
danielhelfand: Not sure how necessary saying `nodes (containers)` is, but thought it helped a little.

BenTheElder: this flag still seems like the wrong place to explain what deleting does.
the command usage seems more appropriate

BenTheElder: (or documentation)

danielhelfand: Fair. I like the idea of usage, but it might be worthwhile to also have in docs as well.

BenTheElder:
I think "delete all clusters" is sufficient. if users want to know more about what deleting a cluster does, we should document that in some place other than the `--all` flag description.

thanks for the PR though!

PR #1268, opened 2020-01-17
Bump golangci-lint timeout.

amwat:
The default 1m timeout is resulting in a lot of flakes.


neolit123:
/lgtm
wondering if the linter supports this flag so that it can exit early out of logic that can block indefinitely.

amwat:
yup https://github.com/golangci/golangci-lint#command-line-options

PR #1269, opened 2020-01-18
kind integrated test workflow MVP

BenTheElder:
implements #986


hack/ci/e2e-k8s.sh
```diff
@@ -22,167 +22,9 @@ set -o errexit -o nounset -o xtrace
...
-  # create the cluster and run tests
-  create_cluster && run_tests
-}
-
-main
+# BUILD_TYPE: bazel or docker
+#
+export KIND_BUILD_TYPE="${BUILD_TYPE:-bazel}"
+export KIND_TEST_FOCUS_OVERRIDE="${FOCUS:-}"
+export KIND_TEST_SKIP_OVERRIDE="${SKIP:-}"
```
BenTheElder: in the future we will stop doing this and just control the lists in code centrally (IE `kind test presubmit` will have a fixed list, with the override being meant for local prototyping hacks, not CI)

hack/ci/e2e-k8s.sh
```diff
@@ -22,167 +22,9 @@ set -o errexit -o nounset -o xtrace
...
+kind test presubmit --ip-family="${IP_FAMILY:-ipv4}"
```
BenTheElder: intend to minimize knobs that need touching heavily to avoid the typical confusion we have with our test rigs, however ip-family is one axis that we definitely test along in kubernetes presubmit and can't really handle well otherwise.

when we migrate this into prow config it will be simpler like:
`kind test presubmit`
`kind test presubmit --ip-family=ipv6`
w/o the env stuff.

aojea: just that you mention this topic, want to share something I realized this week.
With dual-stack is it possible to have 2 kind of clusters:
- dualstack ipv4 (default) - ipv6 (sec)
- dualstack ipv6 (default) - ipv4 (sec)

maybe something we should think for the future, the default is giving by the order of the cluster-cidr configured.

pkg/cmd/kind/test/internal/testcmd/testcmd.go
```diff
@@ -0,0 +1,363 @@
...
+			}
+			return errors.New("aborting due to invalid configuration")
+		}
+		return errors.Wrap(err, "failed to create cluster")
+	}
+
+	// In Kubernetes CI we don't have real IPv6 connectivity so we need to
+	// employ some work arounds when testing IPv6 clusters
+	// We do this all environments for consistency.
+	if config.Networking.IPFamily == v1alpha4.IPv6Family {
```
aojea: `config.Networking.IPFamily == v1alpha4.IPv6Family || config.Networking.IPFamily == v1alpha4.DualStackFamily`

pkg/cmd/kind/test/internal/testcmd/testcmd.go
```diff
@@ -0,0 +1,363 @@
...
+		}
+	}
+
+	// run tests
+	skip := details.Skip
+	if details.Parallel {
+		skip = strings.Join([]string{`\[Serial\]`, skip}, "|")
+	}
+	cmd := exec.Command(
+		"./hack/ginkgo-e2e.sh",
```
aojea: do we need the  `./hack/ginkgo-e2e.sh` bash script?
a quick look to the script seems that we can avoid most of the checks and run directly the `e2e.test`binary,

pkg/cmd/kind/test/internal/testcmd/testcmd.go
```diff
@@ -0,0 +1,363 @@
...
+	if err != nil {
+		return errors.Wrap(err, "failed to get control-plane node for coreDNS IPv6 CI fixes")
+	}
+	if len(nodes) < 1 {
+		return errors.New("failed to locate a kind node for coreDNS IPv6 CI fixes")
+	}
+	node := nodes[0]
+
+	// "fix" coreDNS configmap
+	// TODO: refactor this to not be a shell script.
```
aojea: +1000
This can also be a good /help wanted issue

pkg/cmd/kind/test/test.go
```diff
@@ -0,0 +1,42 @@
...
+	"sigs.k8s.io/kind/pkg/log"
+)
+
+// NewCommand returns a new cobra.Command for test
+func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
+	cmd := &cobra.Command{
+		Args: cobra.NoArgs,
+		// TODO(bentheelder): more detailed usage
+		Use:   "test",
+		Short: "Tests one of [presubmit, conformance]",
```
aojea: how can I run my custom tests? this is specially useful to troubleshoot e2e tests or to create new ones, where you only want to run 1 or a small number of tests.
can we add one subcommand that allows to specify my custom FOCUS and SKIP fields?

pkg/cmd/kind/test/internal/testcmd/testcmd.go
```diff
@@ -0,0 +1,363 @@
...
+	skip := details.Skip
+	if details.Parallel {
+		skip = strings.Join([]string{`\[Serial\]`, skip}, "|")
+	}
+	cmd := exec.Command(
+		"./hack/ginkgo-e2e.sh",
+		// don't use the horrid e2e.test """provider""" concept
+		"--provider=skeleton",
+		// need to inform tests of worker count
+		"--num-nodes=", strconv.Itoa(numNodesForE2E),
```
aojea: the jobs are failing with
> I0118 01:25:13.095] [1] invalid value "" for flag -num-nodes: strconv.ParseInt: parsing "": invalid syntax
I0118 01:25:13.095] [1] Usage of /go/src/k8s.io/kubernetes/bazel-bin/test/e2e/e2e.test:
🤷‍♂

aojea: ok, there is the problem, is adding an space
```
W0118 01:25:14.472] Call stack:
W0118 01:25:14.472]   1: ./hack/ginkgo-e2e.sh:143 main(...)
W0118 01:25:14.472] Exiting with status 1
W0118 01:25:22.159] ERROR: command "./hack/ginkgo-e2e.sh --provider=skeleton --num-nodes= 2 --report-dir /workspace/_artifacts --disable-log-dump=true --ginkgo.focus '' --ginkgo.skip '\[Serial\]|'" failed with error: exit status 1
```

BenTheElder: no, the space is a rendering of passing this to the shell.
the problem is that we're passing it as two arguments, but the first contains `=`

marked the PR WIP, I haven't even run this yet. by inspection it's mostly done but it's still a prototype. I'm going to rework things a bit before un-WIP-ing it.

BenTheElder:
e2e is failing because kubectl isn't being found on the path when building with bazel + running ginkgo-e2e.sh. probably not actually fixed upstream :/
will settle on an option for this tomorrow.

pkg/cmd/kind/test/internal/testcmd/testcmd.go
```diff
@@ -0,0 +1,372 @@
...
+	cleanup := func(calledFromSignalHandler bool) {
+		err2 := provider.CollectLogs(flags.Name, filepath.Join(flags.Artifacts, "logs"))
+		err3 := provider.Delete(flags.Name, flags.kubeconfigPath())
+		// don't touch err if we're being called from another goroutine
+		// on program early exit, or if err is already non-nil
+		if calledFromSignalHandler || err != nil {
+			return
+		} else if err2 != nil {
+			err = err2
+		} else {
```
aojea: this is only `if err3 != nil` or we alway assign err3 to err at this point?

pkg/cmd/kind/test/internal/testcmd/testcmd.go
```diff
@@ -0,0 +1,372 @@
...
+		return errors.Wrap(err, "error creating build context")
+	}
+	if err := ctx.Build(); err != nil {
+		return errors.Wrap(err, "error building node image")
+	}
+
+	// build test binaries
+	switch buildType {
+	case "bazel":
+		if err := exec.InheritOutput(
```
aojea: this writes to the logger, right?

pkg/cmd/kind/test/internal/testcmd/testcmd.go
```diff
@@ -0,0 +1,372 @@
...
+	"path/filepath"
+	"strconv"
+	"strings"
+	"sync"
+	"syscall"
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"sigs.k8s.io/kind/pkg/apis/config/v1alpha4"
```
aojea: shouldn't we use the internal APIs?

pkg/cmd/kind/test/internal/testcmd/testcmd.go
```diff
@@ -0,0 +1,372 @@
...
+
+func boolToYorN(y bool) string {
+	if y {
+		return "y"
+	}
+	return "n"
+}
+
+func enableIPv6Workarounds(provider *cluster.Provider, name string) error {
+	// get a control plane node to use run kubectl from
```
aojea: maybe we should check that ipv6 is enabled? or at least ipv6 forwarding?

pkg/cmd/kind/test/internal/testcmd/testcmd.go
```diff
@@ -0,0 +1,372 @@
...
+	nodes, err := provider.ListInternalNodes(name)
+	if err != nil {
+		return errors.Wrap(err, "failed to get nodes for coreDNS IPv6 CI fixes")
+	}
+	nodes, err = nodeutils.ControlPlaneNodes(nodes)
```
aojea: nit, a different name to the assigned variable to distinguish from the previous one?

hack/ci/e2e-k8s.sh
BenTheElder: that isn't really related to this change though.

pkg/cmd/kind/test/test.go
BenTheElder: - see the e2e.sh for details on overriding these
- it's not intended to run _your_ tests. it's intended to run the standard flows in a way a non-power-user can trivially mimic by invoking the command(s). you as a power user can still write a little script with your flows.

pkg/cmd/kind/test/internal/testcmd/testcmd.go
BenTheElder: maybe? probably not. the replacements need to be equivalent. i'm not sure how many casual contributors know sed / shell and go string manipulation deeply 🤷‍♂

pkg/cmd/kind/test/internal/testcmd/testcmd.go
BenTheElder: that's anti-helpful here. we don't want to retain and accidentally use the unfiltered node list.
we could break this out into a helper method but it's just a few lines so that's not super helpful

pkg/cmd/kind/test/internal/testcmd/testcmd.go
BenTheElder: the caller is responsible for having ipv6 enabled docker. we're not going to muck with the host in `kind`

pkg/cmd/kind/test/internal/testcmd/testcmd.go
BenTheElder: that's orthogonal to this PR. but in any case I'm not convinced that this is useful right now, it will skew versus other test callers and this script also does helpful things like locating the e2e binary in the many possible build paths.

pkg/cmd/kind/test/internal/testcmd/testcmd.go
BenTheElder: we can't realy TODO constants that don't exist yet, dual stack my not be a family, it will probably be many families?
dual stack as an API is a TODO.

also, there is validation long before this code rune that errors if not one of ipv4 / ipv6 specifically to handle the fact that this command is not aware of any other families currently

pkg/cmd/kind/test/internal/testcmd/testcmd.go
BenTheElder: no. this wires it to stdout / stderr.

this respecting streams is a broad TODO, but it's not super important for cmd/ at the moment

pkg/cmd/kind/test/internal/testcmd/testcmd.go
BenTheElder: we don't care at this point, because if err3 is nil at this point so are err2 and err /shrug

pkg/cmd/kind/test/internal/testcmd/testcmd.go
BenTheElder: no. internal apis are for:
- translating between versions
- allowing runtime code implementing the API to handle a single version, possibly with extra (computed?) data that is not part of the public API.

the CLI is a client of the runtime.

pkg/cmd/kind/test/internal/testcmd/testcmd.go
aojea: nah, no problem at all, is just there are at least 1 or 2 places more doing this that way, that's why I added the nit  😄

hack/ci/e2e-k8s.sh
aojea: why I can't resolve the review comments to minimize them?
🤔

BenTheElder: you don't have permissions. :github:

pkg/cmd/kind/test/internal/testcmd/testcmd.go
aojea: 👍

pkg/cmd/kind/test/internal/testcmd/testcmd.go
aojea: 👍

pkg/cmd/kind/test/test.go
aojea: I think I've found my answer, just let me rephrase, how can I specify FOCUS and SKIP with this change?
I have to use the env variables `KIND_TEST_SKIP_OVERRIDE` and `KIND_TEST_FOCUS_OVERRIDE` right?

BenTheElder: those are temporary for transitioning the prowjobs
you can run e2e.test / ginkgo e2e.sh yourself currently.

aojea:
this is an awesome improvement, please disregard my comments, I don't know how to minimize them 🙃

BenTheElder:
/test all

https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/1269/pull-kind-e2e-kubernetes/1220495992827154437

that is some weird test behavior

BenTheElder:
[doing some experiments with e2e before moving forward, commits etc. will be cleaned up before hold / WIP are removed]

aojea:
tests are failing because of 🤔 
`    exec: "kubectl": executable file not found in $PATH`

BenTheElder:
poor connectivity and forgot to warm local deps right now so using CI to
debug

BenTheElder:
trailing newline bug :P

BenTheElder:
Just flakes. Will clean this up soon. The PATH hacks are lame and were not supposed to still be required, will write a more portable approach.

aojea:
one small doc to understand how to use it or mostly for own consumption?

BenTheElder:
Docs come when it's not an MVP :+)

BenTheElder:
However it really is just `kind test presubmit` if you have the deps setup for building k8s. Other stuff is for CI, except the optional flags which have help usage ...

BenTheElder:
/retest
Storage test flake in 1.14

aojea:
Do you think that is possible to implement 2 executions of the `e2e.test`?
One using parallel without the serial tests and other only with the serial tests and aggregating  the results of both runs

BenTheElder:
I don't think we should deviate from the rest of upstream currently.

BenTheElder:
okaaay, let's see about getting this up to date again

PR #1274, opened 2020-01-21
only set featuregates string if non empty

BenTheElder:
see this thread: https://github.com/kubernetes-sigs/kind/pull/1263/files#r368245568

untested

### DIFF
--- a/pkg/cluster/internal/kubeadm/config.go
+++ b/pkg/cluster/internal/kubeadm/config.go
@@ -308,11 +308,13 @@ controlPlaneEndpoint: "{{ .ControlPlaneEndpoint }}"
 # to the cluster after rewriting the kubeconfig to point to localhost
 apiServer:
   certSANs: [localhost, "{{.APIServerAddress}}"]
+{{ if .FeatureGates -}}
   extraArgs:
     "feature-gates": "{{ .FeatureGatesString }}"
+{{- end}}
 controllerManager:
-  extraArgs:
 {{ if .FeatureGates -}}
+  extraArgs:
     "feature-gates": "{{ .FeatureGatesString }}"
 {{- end}}
     enable-hostpath-provisioner: "true"
--- a/pkg/cluster/internal/kubeadm/config.go
+++ b/pkg/cluster/internal/kubeadm/config.go
@@ -155,22 +155,22 @@ nodeRegistration:
     node-ip: "{{ .NodeAddress }}"
 networking:
   podSubnet: "{{ .PodSubnet }}"
+controllerManagerExtraArgs:
+  enable-hostpath-provisioner: "true"
+{{ if .FeatureGates }}  "feature-gates": "{{ .FeatureGatesString }}"
 apiServerExtraArgs:
   "feature-gates": "{{ .FeatureGatesString }}"
-controllerManagerExtraArgs:
-  "feature-gates": "{{ .FeatureGatesString }}"
-  enable-hostpath-provisioner: "true"
 schedulerExtraArgs:
   "feature-gates": "{{ .FeatureGatesString }}"
 nodeRegistration:
   kubeletExtraArgs:
     "feature-gates": "{{ .FeatureGatesString }}"
-{{if .FeatureGates}}kubeProxy:
+kubeProxy:
   config:
     featureGates:
 {{ range $key := .SortedFeatureGateKeys }}
       "{{ $key }}": {{ index $.FeatureGates $key }}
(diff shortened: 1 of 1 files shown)


BenTheElder:
cc @liggitt @aojea

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -415,8 +420,10 @@ controlPlaneEndpoint: "{{ .ControlPlaneEndpoint }}"
 # to the cluster after rewriting the kubeconfig to point to localhost
 apiServer:
   certSANs: [localhost, "{{.APIServerAddress}}"]
+  {{ if .FeatureGates -}}
   extraArgs:
     "feature-gates": "{{ .FeatureGatesString }}"
+  {{- end }}
 controllerManager:
   extraArgs:
     "feature-gates": "{{ .FeatureGatesString }}"
```
liggitt: needs wrapping here?

BenTheElder: thanks, done

aojea:
having second thoughts about this, because it only solves the problem when no feature-gate is configured that may be confusing for the users when they try to use additional feature gates and only some are working.

At the end of the day, we are only trying to solve the dual-stack problem that has 2 different features gates, one in kubeadm that configures the corresponding features gates of the components and does validation on the input parameters.

I think that we should go with Ben suggestion of implementing kubeadm DualStack feature gate in kind (validation and population of components featureGates) and avoid this change.

BenTheElder:
> having second thoughts about this, because it only solves the problem when no feature-gate is configured that may be confusing for the users when they try to use additional feature gates and only some are working.

This is a debateably useful config change that just avoids explicitly setting empty strings all over the place, which is sort of noise. On the other hand it adds noise to the config templates...

> At the end of the day, we are only trying to solve the dual-stack problem that has 2 different features gates, one in kubeadm that configures the corresponding features gates of the components and does validation on the input parameters.

I am not trying to solve that here. The solution to that is to have kind internally handle dualstack properly, kind's networking config should drive feature gates etc, which is totally unrelated to this PR.

> I think that we should go with Ben suggestion of implementing kubeadm DualStack feature gate in kind (validation and population of components featureGates) and avoid this change.

Yeah ... To clarify though: 
- We already enable _kubernetes_ feature gate(s) in kind in kubernetes 1.11
- kubeadm feature gates can influence kubernetes feature gates but we should not rely on that if we want to allow users to enable extra feature gates... Regardless of anything kind did special already, this is expected in kubeadm, if you want feature gate foo AND kubeadm's feature gate bar that enables kubernetes feature gate baz then you must set foo and baz as kubernetes feature gates yourself. it won't mutate an explicit setting of foo to append baz.
- kind user supplied feature gates map[name]=enabled should be applied on top of the kind generate feature gates before passing to kubeadm

aojea:
very convincing 
/hold cancel

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -311,15 +312,19 @@ apiServer:
...
```
aojea: this should be added to the `if`

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -311,15 +312,19 @@ apiServer:
...
 controllerManager:
   extraArgs:
+{{ if .FeatureGates -}}
```
aojea: move up a few lines so it includes the apiserver feature-gates

liggitt:
lgtm, isn't this still needed?

BenTheElder:
reviving...

pkg/cluster/internal/kubeadm/config.go
BenTheElder: sigh, this part was correct, moving it up broke the other extraArgs.

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/1274#discussion_r386720341 ... fixing

BenTheElder:
flake
> 

[sig-node] ConfigMap should be consumable via the environment [NodeConformance] [Conformance] expand_less
--


https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_kind/1274/pull-kind-conformance-parallel-1-15/1234627209726005251/
/retest

BenTheElder:
tests green

BenTheElder:
Flakes
/retest

PR #1276, opened 2020-01-22
Drop quiet from docker commands.

amwat:
We shouldn't be using quiet with --format.
/cc @BenTheElder 

xref: #154


BenTheElder:
/lgtm
/approve
Thanks!

Per discussion, docker allows this even though it doesn't make sense, podman does not.
Fixing this is mostly a no-op, except that it also helps experiments regarding podman compatibility..

/retest

Issue #597, opened 2019-06-07
[docs] Is it possible to explicitly specify k8s version?


artemnikitin:
**What would you like to be documented**:
Is it possible to explicitly specify Kubernetes version to run with `kind create cluster`? At the moment, if create cluster,  then version 1.14.2 will be used. It's unclear if it's possible to change that.

**Why is this needed**:
At the moment it's not clear if it's possible or not. If it's possible, then it's not clear how to do that.

tao12345666333:
You can specify the tag of node image
```
kind create cluster  --image kindest/node:v1.14.1
```

tao12345666333:
or follow this https://kind.sigs.k8s.io/#installation-and-usage

artemnikitin:
Thanks for info! Sorry for missing this

ryanhartje:
How can you configure this via a manifest?

BenTheElder:
the `image` field on each node in `nodes`, currently `--image` maps to over-ridding these.
`nodes` has a default like:
```
nodes:
- role: control-plane
   image: <default for kind version>
```

ryanhartje:
Thank you!!!

PR #1282, opened 2020-01-23
Add setting k8s versions in config to docs

ryanhartje:
updated the docs to add some info about how to set the node images in config files.

I needed to test against a specific version matrix, and I was unable to find how to do this in the docs. While I could do this via the command line, it wasn't sufficient when trying to use Kind for testing automation. 

This should resolve this issue as well:
https://github.com/kubernetes-sigs/kind/issues/597

### DIFF
--- a/site/content/docs/user/configuration.md
+++ b/site/content/docs/user/configuration.md
@@ -161,7 +161,7 @@ nodes:
 - role: worker
 {{< /codeFromInline >}}
 
-You can also set a specific Kubernetes version by setting the `node`'s container image. You can find available image tags on the [releases page](https://github.com/kubernetes-sigs/kind/releases). Please use the `sha256` shasum for your desired kubernetes version, as seen in this example:
+You can also set a specific Kubernetes version by setting the `node`'s container image. You can find available image tags on the [releases page](https://github.com/kubernetes-sigs/kind/releases). Please include the `@sha256:` [image digest](https://docs.docker.com/engine/reference/commandline/pull/#pull-an-image-by-digest-immutable-identifier) from the image in the release notes, as seen in this example:
 
 {{< codeFromInline lang="yaml" >}}
 kind: Cluster
--- a/site/content/docs/user/configuration.md
+++ b/site/content/docs/user/configuration.md
@@ -161,16 +161,16 @@ nodes:
 - role: worker
 {{< /codeFromInline >}}
 
-You can also set a specific Kubernetes version by setting the node image. You can find available versions on [DockerHub](https://hub.docker.com/r/kindest/node/tags):
+You can also set a specific Kubernetes version by setting the `node`'s container image. You can find available image tags on the [releases page](https://github.com/kubernetes-sigs/kind/releases). Please use the `sha256` shasum for your desired kubernetes version, as seen in this example:
+
 {{< codeFromInline lang="yaml" >}}
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
 - role: control-plane
 - role: worker
-  image: kindest/node:v1.16.4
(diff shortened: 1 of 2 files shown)


site/content/docs/user/configuration.md
```diff
@@ -161,6 +161,18 @@ nodes:
 - role: worker
 {{< /codeFromInline >}}
 
+You can also set a specific Kubernetes version by setting the node image. You can find available versions on [DockerHub](https://hub.docker.com/r/kindest/node/tags):
```
BenTheElder: Please don't recommend this. We don't currently guarantee that those tags will be compatible. You should check the release notes for a list of images we shipped with the release OR build a custom image

ryanhartje: Okay, can we instead of not describing how to do this, update the verbiage? It seems disadvantageous to not document this.

BenTheElder: right, we should document setting image, just not doing so by checking docker hub tags, please point to the release notes instead and recommend including the `sha256` bit.

BenTheElder: we should also give an example following best practices I.E., including the full documented image.

ryanhartje: I'm not sure what you had in mind. I'm not super familiar with `kind` so I wouldn't consider myself authoritative on best practices.

aojea:
@ryanhartje please sign the Linux Foundation CLA 😄

ryanhartje:
I signed up several years ago. The bot got confused because I pushed a commit from another computer with a different email configured. :/

aojea:
> I signed up several years ago. The bot got confused because I pushed a commit from another computer with a different email configured. :/

heh
/lgtm

site/content/docs/user/configuration.md
```diff
@@ -161,6 +161,18 @@ nodes:
...
 
+You can also set a specific Kubernetes version by setting the `node`'s container image. You can find available image tags on the [releases page](https://github.com/kubernetes-sigs/kind/releases). Please use the `sha256` shasum for your desired kubernetes version, as seen in this example:
+
+{{< codeFromInline lang="yaml" >}}
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+nodes:
+- role: control-plane
+- role: worker
+  image: kindest/node:v1.16.4@sha256:b91a2c2317a000f3a783489dfb755064177dbc3a0b2f4147d50f04825d016f55
```
BenTheElder: I wonder how hard it would be to put this in a data file as a follow up, so we can keep the example updated with a valid one for the current kind version.

can be a follow up

site/content/docs/user/configuration.md
```diff
@@ -161,6 +161,18 @@ nodes:
 - role: worker
 {{< /codeFromInline >}}
 
+You can also set a specific Kubernetes version by setting the `node`'s container image. You can find available image tags on the [releases page](https://github.com/kubernetes-sigs/kind/releases). Please use the `sha256` shasum for your desired kubernetes version, as seen in this example:
```
BenTheElder: I want to go ahead and get this in, but I think we should use more precise language for the sha256 bit:
```suggestion
You can also set a specific Kubernetes version by setting the `node`'s container image. You can find available image tags on the [releases page](https://github.com/kubernetes-sigs/kind/releases). Please include the `@sha256:` [image digest](https://docs.docker.com/engine/reference/commandline/pull/#pull-an-image-by-digest-immutable-identifier) from the image in the release notes, as seen in this example:
```

BenTheElder:
/lgtm
/approve
/hold
holding for at most a day to give a chance to follow up on comments now, otherwise will handle in a seperate PR or two

BenTheElder:
sorry for the delay on this!

BenTheElder:
/lgtm
/hold cancel
sorry that was a long "day" .. I went oncall for the prow.k8s.io etc. the next day and lost track of some things :(

thanks again for this!

PR #1286, opened 2020-01-25
Use logger instead of printf for waitforready

Anuj2512:
- Updates the wait for ready execution to use logger instead of directly printing output to screen.

### DIFF
--- a/pkg/cluster/internal/create/actions/waitforready/waitforready.go
+++ b/pkg/cluster/internal/create/actions/waitforready/waitforready.go
@@ -69,13 +69,13 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 	isReady := waitForReady(node, startTime.Add(a.waitTime))
 	if !isReady {
 		ctx.Status.End(false)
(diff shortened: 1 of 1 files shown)


pkg/cluster/internal/create/actions/waitforready/waitforready.go
```diff
@@ -69,13 +69,13 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 	isReady := waitForReady(node, startTime.Add(a.waitTime))
 	if !isReady {
 		ctx.Status.End(false)
-		fmt.Println(" • WARNING: Timed out waiting for Ready ⚠️")
+		ctx.Logger.V(0).Info(" • WARNING: Timed out waiting for Ready ⚠️\n")
```
BenTheElder: For consistency with other log statements we should maybe drop the trailing newline, the logger ensures a trailing newline.

Same for below

Anuj2512: addressed.

BenTheElder:
/lgtm
/approve
Thanks 🙃

PR #1287, opened 2020-01-25
add get address command

kandros:
add `kind get address` command to print cluster address of API server

Screenshot 2020-01-25 at 16 19 40


neolit123:
you can already do something similar with the following command, no?

`kind get kubeconfig | grep server ...`

kandros:
> you can already do something similar with the following command, no?
> 
> `kind get kubeconfig | grep server ...`

missed that 😅  great

Issue #1292, opened 2020-01-28
extraPortMappings are on 0.0.0.0 when address is not specified with ipv6 cluster


nbartos:
**What happened**:
Using extraPortMappings in the kind cluster create config file works with ipv4 clusters but not ipv6.  When using ipv6, tcp connections just hang.

**What you expected to happen**:
Port forwards to work the same in ipv4 and ipv6 clusters.

**How to reproduce it (as minimally and precisely as possible)**:
Configure /etc/docker/daemon.json for ipv6:
```
{
  "fixed-cidr-v6": "2001:db8:1::/64",
  "ipv6": true
}
```
Restart docker:
```
systemctl restart docker
```

Use cluster create config file called v6test.conf:
```
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
networking:
  ipFamily: ipv6
nodes:
- role: control-plane
  kubeadmConfigPatches:
  - |
    kind: InitConfiguration
    nodeRegistration:
      kubeletExtraArgs:
        node-labels: "ingress-ready=true"
        authorization-mode: "AlwaysAllow"
  extraPortMappings:
  - containerPort: 32444
    hostPort: 32444
    protocol: TCP
```

Then run:
```
export KUBECONFIG=.kind-kubeconfig2
kind create cluster --name v6test --kubeconfig=.kind-kubeconfig2 --config=v6test.conf
helm install ngress https://kubernetes-charts.storage.googleapis.com/nginx-ingress-1.26.2.tgz --set controller.service.type=NodePort --set controller.service.omitClusterIP=true --set defaultBackend.service.omitClusterIP=true --set controller.service.nodePorts.https=32444 --wait
curl -Lfk 'https://[::1]:32444/'
```

The last curl should hang.  Repeating steps with `ipFamily: ipv4` in the config file, the curl should return immediately with a 404.

**Anything else we need to know?**:

**Environment:**

- kind version: (use `kind version`): 0.7.0
- Kubernetes version: (use `kubectl version`): v1.17.0
- Docker version: (use `docker info`): 19.03.1
- OS (e.g. from `/etc/os-release`): Ubuntu 18.04

aojea:
this is interesting, it's mapping the port to IPv4

```
$ sudo  /usr/sbin/ip6tables-save  | grep 32444
$ sudo  /usr/sbin/iptables-save  | grep 32444
[...]

PR #1295, opened 2020-01-28
explicitly default port mapping respecting IP family

BenTheElder:
/cc @amwat @aojea 
fixes https://github.com/kubernetes-sigs/kind/issues/1292

### DIFF
--- a/pkg/cluster/internal/providers/docker/provision.go
+++ b/pkg/cluster/internal/providers/docker/provision.go
@@ -82,11 +82,11 @@ func planCreation(cluster string, cfg *config.Cluster) (createContainerFuncs []f
 						ContainerPort: common.APIServerInternalPort,
 					},
 				)
-				return createContainer(runArgsForNode(node, name, genericArgs))
+				return createContainer(runArgsForNode(node, cfg.Networking.IPFamily, name, genericArgs))
 			})
 		case config.WorkerRole:
 			createContainerFuncs = append(createContainerFuncs, func() error {
-				return createContainer(runArgsForNode(node, name, genericArgs))
+				return createContainer(runArgsForNode(node, cfg.Networking.IPFamily, name, genericArgs))
 			})
 		default:
 			return nil, errors.Errorf("unknown node role: %q", node.Role)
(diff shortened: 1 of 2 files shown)


pkg/cluster/internal/providers/docker/provision.go
```diff
@@ -272,23 +274,39 @@ func generateMountBindings(mounts ...config.Mount) []string {
...
-		if pm.ListenAddress != "" {
-			hostPortBinding = net.JoinHostPort(pm.ListenAddress, fmt.Sprintf("%d", pm.HostPort))
-		} else {
-			hostPortBinding = fmt.Sprintf("%d", pm.HostPort)
+		// do provider internal defaulting
+		// in a future API revision we will handle this at the API level and remove this
+		if pm.ListenAddress == "" {
+			switch clusterIPFamily {
+			case config.IPv4Family:
+				pm.ListenAddress = "0.0.0.0" // this is the docker default anyhow
```
aojea: @liggitt enlightened me about these unspecified addresses, it turns out the go library behaves the same with "0.0.0.0" and "::"

BenTheElder: dockerd doesn't though, it uses a proxy process for `::` and iptables for `0.0.0.0`

BenTheElder:
fixed typo in string unkown -> unknown

pkg/cluster/internal/providers/docker/provision.go
```diff
@@ -272,23 +274,39 @@ func generateMountBindings(mounts ...config.Mount) []string {
...
+			case config.IPv6Family:
```
aojea: why not checking only for IPv6 now and keep defaulting as before?

aojea:
can't we default the pm.	ListenAddress of portmaps in the API validation stage?

BenTheElder:
Please read the code comments 🙃

In the future we should rethink API defaulting but right now that's left up to the provider.

pkg/cluster/internal/providers/docker/provision.go
BenTheElder: Because the same problem will happen in dual stack ...

aojea:
ahhg, it's too late for me, will lgtm tomorrow morning if nobody does it before 💤

Issue #1290, opened 2020-01-27
Rare "failed to init node with kubeadm" on Kind v0.7.0


howardjohn:
**What happened**: Occasionally find errors like https://prow.istio.io/view/gcs/istio-prow/pr-logs/pull/istio_istio/20244/integ-telemetry-k8s-tests_istio/7089 failing to start up the kind cluster. I have seen similar errors before, but its very rare, anecdotally I see one a week or so across 1000s of tests. (meta comment: is there a good way to grep across all test logs in GCS? I cannot find one. I used to download them locally but ran out of disk space).

**What you expected to happen**: ideally no errors

**How to reproduce it (as minimally and precisely as possible)**: I cannot repro

**Anything else we need to know?**:
I don't really expect too much here. As I see errors in the future I'll add more context to maybe root cause this. Right now I am not too concerned with this as its extremely uncommon, mostly just opening this to track or if it helps find some issue

**Environment:**
- kind version: (use `kind version`): v0.7.0
- Kubernetes version: (use `kubectl version`): v1.17.0
- Docker version: (use `docker info`): 19.03.2
- OS (e.g. from `/etc/os-release`): linux

BenTheElder:
> (meta comment: is there a good way to grep across all test logs in GCS? I cannot find one. I used to download them locally but ran out of disk space).

there isn't AFAIK. for kubernetes someone hacked up a dashboard to allow searching test failures at some point but only for junit results and not for pod-utils jobs IIRC ... :|

BenTheElder:
> https://prow.istio.io/view/gcs/istio-prow/pr-logs/pull/istio_istio/20244/integ-telemetry-k8s-tests_istio/7089

This looks like the API server had trouble starting up in a reasonable time frame, kubeadm gave up after about 4 minutes of waiting on the healthz. We could probably ask kubeadm to allow a longer wait but I'm not sure that's desirable.

The results do have logs from the API server, I don't see anything obvious in them suggesting a clear KIND bug, it kinda looks like things were just running to slowly.

Currently my plan for this is to attempt to tweak our etcd configuration to make it lighter on I/O as that's my main suspect other than just ... Kubernetes. I had a breif chat with our team about possible options for this last week but we're not shipping anything just yet.

It could also be prow giving us issues with noisy shared hosts. I'm not seeing this in our prow running kind at HEAD though, we actually haven't had a cluster bringup flake in a long time (plenty of e2e flakes though...).

[...]

Issue #1288, opened 2020-01-27
Cannot create cluster due to `docker exec cat /kind/version` failing


2opremio:
**What happened**:

Kind failed to create a cluster with 

```
ERROR: failed to create cluster: failed to generate kubeadm config content: failed to get kubernetes version from node: failed to get file: command "docker exec --privileged flux-e2e-2-control-plane cat /kind/version" failed with error: exit status 1
```

Full logs at https://circleci.com/gh/fluxcd/flux/9822

**What you expected to happen**:

I expect the cluster to be created without errors

**How to reproduce it (as minimally and precisely as possible)**:

Unfortunately it's not systematic. It only happens from time to time

It happens when running `make e2e` in https://github.com/fluxcd/flux/

**Anything else we need to know?**:

Although things have improved quite a bit since Kind 0.7.0 we are still getting some cluster creation errors from time to time. More info at https://github.com/fluxcd/flux/issues/2672

This may have to do with the fact that we create a few (4) Kind clusters in parallel. See https://github.com/fluxcd/flux/blob/cb319f2f7878f33369cf54139df26867aa6632ca/test/e2e/run.bash#L55

**Environment:**

- kind version: (use `kind version`): 0.7.0
- Kubernetes version: (use `kubectl version`): 1.14.10
- Docker version: (use `docker info`): 18.09.3
- OS (e.g. from `/etc/os-release`): Ubuntu 16.04.5 LTS

BenTheElder:
Can you add `-v 1` to your creation call? It will dump stack and more error output.

This error can happen if the node container exited for some reason.

How much resources are available where you are creating 4 in parallel?

2opremio:
> Can you add `-v 1` to your creation call? It will dump stack and more error output.

Will do


> How much resources are available where you are creating 4 in parallel?

[We are using](https://github.com/fluxcd/flux/blob/019a9bb7fba3b1a82e5bc414f6112ed458e3c380/.circleci/config.yml#L49-L51) CircleCI's machine executor with a large resource class [which is supposed to have 4 cores and 15GB of ram](https://circleci.com/docs/2.0/executor-types/#using-machine)

BenTheElder:
That should be enough RAM & CPU for four single node clusters, but disk I/O which isn't advertised there might be an issue.

2opremio:
@BenTheElder note that we only get errors from time to time, and only at cluster creation. I would expect the failures to be more systematic.

[...]

PR #1302, opened 2020-01-30
Enable basic podman support.

amwat:
Implemented a bunch of changes in the podman provider to make it work.

```
KIND_EXPERIMENTAL_PROVIDER=podman kind create cluster
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.17.0) 🖼
 ✓ Preparing nodes 📦  
 ✓ Writing configuration 📜 
 ✓ Starting control-plane 🕹️ 
 ✓ Installing CNI 🔌 
 ✓ Installing StorageClass 💾 

Have a question, bug, or feature request? Let us know! https://kind.sigs.k8s.io/#community 🙂 
```


pkg/cluster/internal/providers/podman/provider.go
```diff
@@ -53,6 +56,10 @@ func (p *Provider) Provision(status *cli.Status, cluster string, cfg *config.Clu
 	if err := ensureMinVersion(); err != nil {
 		return err
 	}
+	// warnf if podman is running rootless
+	if os.Geteuid() != 0 {
+		p.logger.Warn("podman might not work properly in rootless mode")
```
BenTheElder: maybe something like this:
```suggestion
		p.logger.Warn("kind may not work properly with rootless podman")
```
or this:
```suggestion
		p.logger.Warn("podman provider may not work properly in rootless mode")
```

_podman_ works in rootless, but podman + kind may not

pkg/cluster/internal/providers/podman/provider.go
```diff
@@ -53,6 +56,10 @@ func (p *Provider) Provision(status *cli.Status, cluster string, cfg *config.Clu
 	if err := ensureMinVersion(); err != nil {
 		return err
 	}
+	// warnf if podman is running rootless
```
BenTheElder:
```suggestion
	// warn if podman is running rootless <explain why perhaps?>
	// <kind doesn't currently work with podman rootless XYZ>
```

pkg/cluster/internal/providers/podman/provider.go
```diff
@@ -17,19 +17,22 @@ limitations under the License.
...
 import (
+	"encoding/json"
 	"fmt"
 	"net"
+	"os"
+	"strconv"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/util/sets"
+	"sigs.k8s.io/kind/pkg/cluster/internal/providers/provider/common"
```
BenTheElder: this should be grouped below with the internal imports.

BenTheElder:
generally LGTM, couple nits 🙃

pkg/cluster/internal/providers/podman/node.go
```diff
@@ -52,7 +52,7 @@ func (n *node) Role() (string, error) {
 func (n *node) IP() (ipv4 string, ipv6 string, err error) {
 	// retrieve the IP address of the node using podman inspect
 	cmd := exec.Command("podman", "inspect",
-		"-f", "{{range .NetworkSettings.Networks}}{{.IPAddress}},{{.GlobalIPv6Address}}{{end}}",
+		"-f", "{{.NetworkSettings.IPAddress}},{{.NetworkSettings.GlobalIPv6Address}}",
```
aojea: so podmad does IPv6?
cool

pkg/cluster/internal/providers/podman/provider.go
```diff
@@ -164,13 +170,27 @@ func (p *Provider) GetAPIServerEndpoint(cluster string) (string, error) {
...
+		HostIP        string `json:"hostIP"`
+	}
+
+	var portMappings []portMapping
+	if err := json.Unmarshal([]byte(lines[0]), &portMappings); err != nil {
+		return "", errors.Errorf("invalid network details: %v", err)
+	}
+	for _, pm := range portMappings {
+		if pm.ContainerPort == common.APIServerInternalPort && pm.Protocol == "tcp" {
+			return net.JoinHostPort(pm.HostIP, strconv.FormatInt(int64(pm.HostPort), 10)), nil
```
aojea: nit  `strconv.Itoa(int(pm.HostPort))`
> Itoa is equivalent to FormatInt(int64(i), 10).

pkg/cluster/internal/providers/podman/node.go
BenTheElder: podman uses CNI so ...
this particular diff is just dealing with slightly different structs vs docker https://www.redhat.com/sysadmin/container-networking-podman

pkg/cluster/internal/providers/podman/provision.go
```diff
@@ -307,6 +303,11 @@ func generatePortMappings(clusterIPFamily config.ClusterIPFamily, portMappings .
 		// generate the actual mapping arg
 		protocol := string(pm.Protocol)
 		hostPortBinding := net.JoinHostPort(pm.ListenAddress, fmt.Sprintf("%d", pm.HostPort))
+		// Podman expects empty string instead of 0 to assign a random port
```
aojea: :man_shrugging:

pkg/cluster/internal/providers/podman/provision.go
```diff
@@ -307,6 +303,11 @@ func generatePortMappings(clusterIPFamily config.ClusterIPFamily, portMappings .
...
+		// https://github.com/containers/libpod/blob/master/pkg/spec/ports.go#L68-L69
+		if strings.HasSuffix(hostPortBinding, ":0") {
```
aojea: why don't check directly `pm.HostPort` before creating the `hostPortBinding` variable?

BenTheElder: That's not any less branchy and this is closer to the docker code /shrug

pkg/cluster/internal/providers/podman/node.go
aojea: CNI folks are our friends ... we'll have better support than libnetwork ... but I don't envision anything extraordinary different than this firewalld, IPv6 proxied portMapping, ... issues

pkg/cluster/internal/providers/podman/provision.go
BenTheElder: Yeah ... 🤷‍♂️
This is why we needed a provider and not just a switch between calling docker and podman. They're not actually 1:1 compatible and podman has a number of bugs we need to avoid (mostly currently by requiring a new enough version ...)

pkg/cluster/internal/providers/podman/provision.go
aojea: fair enough

pkg/cluster/internal/providers/podman/node.go
BenTheElder: It won't have any IPv6 network in the default CNI config though.

aojea:
couldn't test it but overall it LGTM
I know a lot of people that will be happy with this :)
Thanks

aojea:
@BenTheElder this failures is interesting https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_kind/1302/pull-kind-conformance-parallel-1-16/1223359035785351172/

> W0131 21:53:03.952] ERROR: failed to create cluster: failed to generate kubeadm config content: failed to get kubernetes version from node: failed to get file: command "docker exec --privileged kind-worker cat /kind/version" failed with error: exit status 1

one of the nodes, in this case `kind-worker` shows this error in most logfiles

> Error response from daemon: Container 2b64f5a22772902e4a64c2eab46a5e7a0d81765235f7897c6b6465213f59f29d is not running

https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_kind/1302/pull-kind-conformance-parallel-1-16/1223359035785351172/artifacts/logs/kind-worker/

BenTheElder:
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_kind/1302/pull-kind-conformance-parallel-1-16/1223359035785351172/artifacts/logs/kind-worker/serial.log

it looks like it was like, randomly terminated?? the container logs have no errors but didn't get to systemd

xref: https://github.com/kubernetes-sigs/kind/issues/1288 https://github.com/kubernetes-sigs/kind/issues/1290

BenTheElder:
test flake https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_kind/1302/pull-kind-conformance-parallel-1-16/1223371864189964289/
/override pull-kind-conformance-parallel-1-16

pkg/cluster/internal/providers/podman/provider.go
```diff
@@ -17,22 +17,24 @@ limitations under the License.
...
-
 	"sigs.k8s.io/kind/pkg/cluster/internal/providers/provider"
 	"sigs.k8s.io/kind/pkg/cluster/internal/providers/provider/common"
+	"sigs.k8s.io/kind/pkg/cluster/nodes"
 	"sigs.k8s.io/kind/pkg/cluster/nodeutils"
+	"sigs.k8s.io/kind/pkg/errors"
+	"sigs.k8s.io/kind/pkg/exec"
 	"sigs.k8s.io/kind/pkg/internal/apis/config"
 	"sigs.k8s.io/kind/pkg/internal/cli"
+	"sigs.k8s.io/kind/pkg/log"
```
BenTheElder: i'm not going to block on this, but prefer to seperate internal imports from public imports, to give some sense of the split.

- stdlib
- github
- k8s
- public kind
- internal kind

=> arranged by most common / public to most specific to this code

PR #1310, opened 2020-02-04
Move log export logic to providers

amwat:
Enabling more things on podman.
/cc @BenTheElder


BenTheElder:
I think we should restructure these a bit to share code, most of it is not provider specific and we'll want N=3 or more soon enough

pkg/exec/helpers.go
```diff
@@ -132,3 +133,23 @@ func RunWithStdinWriter(cmd Cmd, writerFunc func(io.Writer) error) error {
...
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+	cmd.SetStdout(f)
+	cmd.SetStderr(f)
+	return cmd.Run()
+}
+
+// ToPathFn returns the ToPath function, useful for running concurrently
```
BenTheElder: I don't think this should be in exec. it's just a trivial lambda that can be done inline

pkg/exec/helpers.go
```diff
@@ -132,3 +133,23 @@ func RunWithStdinWriter(cmd Cmd, writerFunc func(io.Writer) error) error {
 		},
 	})
 }
+
+// ToPath is a helper to run a cmd and write the output to path
+func ToPath(cmd Cmd, path string) error {
```
BenTheElder: this probably needs a more descriptive title, `exec.ToPath` isn't a very obvious function out of context.

imo this should be an internal helper of the provider log collection

pkg/cluster/nodes/types.go
```diff
@@ -33,4 +33,6 @@ type Node interface {
 	// Possibly remove this method in favor of obtaining this detail with
 	// exec or from the provider
 	IP() (ipv4 string, ipv6 string, err error)
+	// CollectLogs collects the node logs
+	CollectLogs(dir string) error
```
BenTheElder: s/dir/hostDir

maybe "ExportLogs" ?

pkg/cluster/internal/providers/provider/common/logs.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// ExecToWriter is a helper to run a cmd and write the output to writer
+func ExecToWriter(cmd exec.Cmd, w io.Writer) error {
+	cmd.SetStdout(w)
+	cmd.SetStderr(w)
+	return cmd.Run()
+}
+
+// FileOnHost is a helper to create a file at path
```
BenTheElder: even if the directory doesn't exist, in which case it will be created with ModePerm

pkg/cluster/internal/providers/provider/common/logs.go
```diff
@@ -0,0 +1,60 @@
...
+			"containerd.log",
+		),
+	})
+}
+
+// ExecToWriter is a helper to run a cmd and write the output to writer
+func ExecToWriter(cmd exec.Cmd, w io.Writer) error {
+	cmd.SetStdout(w)
+	cmd.SetStderr(w)
+	return cmd.Run()
```
BenTheElder: this can be done inline in other code like `cmd.SetStdout(w).SetStderr(w).Run()`, a one liner like this isn't usually an exported method /shrug

BenTheElder: also technically per the cmd interface this version should be:
```
...
cmd = cmd.SetStdout(w)
cmd = cmd.SetStderr(w)
...
```

pkg/cluster/internal/providers/podman/provider.go
```diff
@@ -200,3 +202,53 @@ func (p *Provider) node(name string) nodes.Node {
...
+	var errs []error
+	for _, n := range nodes {
+		node := n // https://golang.org/doc/faq#closures_and_goroutines
+		name := node.String()
+		path := filepath.Join(dir, name)
+		if err := internallogs.DumpDir(p.logger, node, "/var/log", path); err != nil {
+			errs = append(errs, err)
+		}
+
+		fns = append(fns, []func() error{
```
BenTheElder: you can just `append(fns, func(){ ... }, func(){ ... })` instead of creating a slice and then spreading it?

pkg/cluster/internal/providers/podman/node.go
```diff
@@ -149,3 +150,7 @@ func (c *nodeCmd) SetStderr(w io.Writer) exec.Cmd {
 	c.stderr = w
 	return c
 }
+
+func (n *node) SerialLogs(writer io.Writer) error {
+	return common.ExecToWriter(exec.Command("podman", "logs", n.name), writer)
```
BenTheElder:
```suggestion
	return exec.Command("podman", "logs", n.name).SetStdout(w).SetStderr(w).Run()
```

(also `writer` can just be `w`, there's no extra info in the variable name since it's just the type, e.g. https://golang.org/pkg/io/#WriteString)

amwat: done

pkg/cluster/internal/providers/podman/provider.go
amwat: :man_facepalming:

pkg/cluster/internal/providers/provider/common/logs.go
amwat: yeah I had that previously, guess not really required as a helper.

pkg/cluster/internal/providers/provider/common/logs.go
amwat: done

BenTheElder:
I think we will want to DRY out further in the future but the public APIs look 👍👍👍 and we should get podman fixed!

/lgtm
/approve
Sorry for the delay!

PR #1314, opened 2020-02-06
Parse: apiversion and kind inverted in error output

maelvls:
For example, with:

```sh
# Note: this kind and apiversion do not exist (they existed in early 2019)
% kind create cluster --name b --config - <<EOF 
kind: Node
role: control-plane
replicas: 3
EOF

ERROR: failed to create cluster: unknown kind kind.sigs.k8s.io/v1alpha3 for apiVersion: Node
```

(that's because I was trying to use the old API from early 2019; I realized later that I had to use `kind.x-k8s.io/v1alpha4` and `kind: Cluster` instead 🙂)

Kind version: 8643089fc

I also want to say thank you again for this absolutely fantastic project! 😁


aojea:
/lgtm
/ok-to-test
thanks :upside_down_face:

BenTheElder:
/lgtm
/approve
thank you!

PR #1316, opened 2020-02-06
WIP: Add extra nics to nodes

qinqon:
Signed-off-by: Quique Llorente <ellorent@redhat.com>

### DIFF
--- a/pkg/apis/config/v1alpha3/types.go
+++ b/pkg/apis/config/v1alpha3/types.go
@@ -85,6 +85,9 @@ type Node struct {
 	// ExtraPortMappings describes additional port mappings for the node container
 	// binded to a host Port
 	ExtraPortMappings []PortMapping `yaml:"extraPortMappings,omitempty"`
+
+	// ExtraNetworks describes additional networks for the node container
+	ExtraNetworks []Network `yaml:"extraNetworks,omitempty"`
 }
 
 // NodeRole defines possible role for nodes in a Kubernetes cluster managed by `kind`
@@ -196,6 +199,13 @@ type PortMapping struct {
 	ListenAddress string `yaml:"listenAddress,omitempty"`
 	// Protocol (TCP/UDP)
 	Protocol PortMappingProtocol `yaml:"protocol,omitempty"`
+}
+
+// Network specifies a network to create and attach a container.
+// In yaml this looks like:
+//  name: secondary-nic1
+type Network struct {
+	Name string `yaml:"name,omitempty"`
 }
 
 // MountPropagation represents an "enum" for mount propagation options,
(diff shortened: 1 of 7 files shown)


BenTheElder:
This won't work yet. Please see past discussion of user defined docker networks, they break DNS.

Also, is there a reason if it did work that you couldn't add extra networks after the cluster was created using your own script? KIND supports multiple backends now (podman) and we'd like to keep things focused and generic...

qinqon:
Hi @BenTheElder, thanks for the quick feedback!

We can create the docker networks after kind is up, but was curious about how difficult would be to implent it at kind, hence this PR. It's possible to fix the DNS issue? Or it would increase complexity too much?

Btw code is super easy to follow congrats on that.

BenTheElder:
heh, hoping to get the code a bit more easy to follow imho 🙃 

The DNS issue is being explored, I want to enable using a docker network for #148, but it's been a bit fiddly to lock down something sane for that 😅 Finally finding some more time to sit and hack on this today hopefully.

once that issue is fixed I think this will just work, but i'm not sure what sort of network config / topology we want to do by default (one per cluster? one common one for kind? both should probably have an option for the user to override? do we still want to try to support the default `bridge` somehow...?), it depends on what hacks we settle on for #148 🔥😅

tao12345666333:
past discussion ref: https://github.com/kubernetes-sigs/kind/pull/484

Recently I was also planning to reimplement the restart support.

Adding a custom network is needed for this. Using the bridge network is the easiest way.

The DNS problems I encountered before, I think I can have a more direct way to solve it via `--add-host`.

BenTheElder:
Add host doesn't solve reboot, or the coreDNS loop ...
There are some other possible approaches though.

qinqon:
@BenTheElder after reading a little the issues, issue is with ips changing after restart so DNS are pointing to wrong address right ? but this PR is just about having extra nics to test knmstate [1] it configure networking at host, want to see if we can use kind to test that but for that we need secondary nics.

[1] https://github.com/nmstate/kubernetes-nmstate

aojea:
I'm afraid that DNS will not be the only issue with multiple NIC for the nodes ... there is a lot of logic in KIND/Kindnet/Kubeadm/Kubernetes that works "by default" assuming a simple network setup like the "autodiscovery" of the addresses for binding the servers, the reported node IP address by kubelet, kindnet... 

I suggest to not overcomplicate the KIND network unless we have a great benefit and keep using alternatives like `koko` after the cluster is up as I suggested here
https://github.com/kubernetes-sigs/kind/pull/939#issuecomment-540186911

qinqon:
Ack, closing this PR. Thanks.

PR #1319, opened 2020-02-07
stop bazel after building

BenTheElder:
stopgap measure to hopefully help reduce pressure prow.k8s.io

cc @dims @liggitt


hack/ci/e2e-k8s.sh
```diff
@@ -49,6 +49,9 @@ build_with_bazel() {
   # make sure we have e2e requirements
   bazel build //cmd/kubectl //test/e2e:e2e.test //vendor/github.com/onsi/ginkgo/ginkgo
 
+  # free up memory by terminating bazel
+  bazel shutdown
```
dims:
```
bazel shutdown || true
pkill ^bazel || true
```

BenTheElder: done

PR #1323, opened 2020-02-10
Improve load docker-image logging

### LINKED ISSUE #1322: Improve logging when using kind load docker-image
Here's the current output when loading a docker image:

```
Image: "my-image:latest" with ID "sha256:cc850a8c6a875325fcb385496e7f3eeed15ea8514babdce95f09c03379a07904" not present on node "kind-control-plane"
```

This message looks like an error message. It gets me every time, I'm scanning through the logs of a failed build, I see it and I think "oh it's failed to find the docker image". And then I remember no, that's what kind outputs when everything's ok. People who are less experienced with kind on my team have often assumed that that is the problem, and wasted time going down rabbit holes trying to debug a problem that doesn't exist.

A small change in wording of the message would make a huge difference to UX here, eg:

```
Image: "my-image:latest" with ID "sha256:cc850a8c6a875325fcb385496e7f3eeed15ea8514babdce95f09c03379a07904" not yet present on node "kind-control-plane", loading...
```

Note the addition of the words `yet` and `loading...`.

jroper:
Fixes #1322

Changed the message logged when the image needs to be loaded so that it doesn't read like an error message and won't be confused as such.

### DIFF
--- a/pkg/cmd/kind/load/docker-image/docker-image.go
+++ b/pkg/cmd/kind/load/docker-image/docker-image.go
@@ -120,7 +120,7 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
 		id, err := nodeutils.ImageID(node, imageName)
 		if err != nil || id != imageID {
 			selectedNodes = append(selectedNodes, node)
-			logger.V(0).Infof("Image: %q with ID %q not present on node %q", imageName, imageID, node.String())
+			logger.V(0).Infof("Image: %q with ID %q not yet present on node %q, loading...", imageName, imageID, node.String())
 		}
 	}
 


pkg/cmd/kind/load/docker-image/docker-image.go
```diff
@@ -120,7 +120,7 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
 		id, err := nodeutils.ImageID(node, imageName)
 		if err != nil || id != imageID {
 			selectedNodes = append(selectedNodes, node)
-			logger.V(0).Infof("Image: %q with ID %q not present on node %q", imageName, imageID, node.String())
+			logger.V(0).Infof("Image: %q with ID %q not yet present on node %q, loading...", imageName, imageID, node.String())
```
BenTheElder: `loading ...` implies we're doing this right now which is somewhat misleading, we won't actually load it.

This should probably be V(1) / debug anyhow, that got lost in translation at some point refactoring the logging https://github.com/kubernetes-sigs/kind/blame/5df15c23d3a8520193655460df310540c2f1a7d3/cmd/kind/load/docker-image/docker-image.go#L122

BenTheElder:
will refine further later

BenTheElder:
thanks for the PR :-)

PR #1337, opened 2020-02-17
Doc: KIND complex network scenarios

aojea:
There were several PR and demand to implement this in KIND.
However, I think that KIND can serve better as a building block for complex scenarios that can be easily scripted, avoiding adding complexity to the project.


site/content/docs/user/networking-scenarios.md
```diff
@@ -0,0 +1,127 @@
...
+
+<img src="/docs/user/images/kind-kubernetes-network-kindnet.png"/>
+
+[The Kubernetes network model](https://kubernetes.io/docs/concepts/cluster-administration/networking/#the-kubernetes-network-model) implies end to end connectivity without NAT between Pods.
+
+By default, KIND uses its own CNI plugin, **Kindnet**, that install the corresponding routes and iptables rules on the cluster nodes.
+
+## Multiple clusters
+
+As we explained before, all KIND clusters are sahring the same docker network, that means that all the cluster nodes have direct connectivity.
```
neiljerram: typo "sahring"

site/content/docs/user/networking-scenarios.md
```diff
@@ -0,0 +1,127 @@
...
+
+If we want to spawn multiple cluster and provide Pod to Pod connectivity between different clusters, first we have to configure the cluster networking parameters to avoid address overlapping.
```
neiljerram: "multiple clusters"

site/content/docs/user/networking-scenarios.md
```diff
@@ -0,0 +1,127 @@
...
+    inet 127.0.0.1/8 scope host lo
+       valid_lft forever preferred_lft forever
+    inet6 ::1/128 scope host 
+       valid_lft forever preferred_lft forever
+10: eth0@if11: <BROADCAST,MULTICAST,UP,LOWER_UP,M-DOWN> mtu 1500 qdisc noqueue state UP 
+    link/ether 02:42:ac:11:00:03 brd ff:ff:ff:ff:ff:ff
+    inet 172.17.0.3/16 brd 172.17.255.255 scope global eth0
+{{< /codeFromInline >}}
+
+That means that Pods will be able to reach other dockers containers that does not belong to any KIND cluster, however, the docker container will not be able to answer to the Pod IP address until we intall the correspoding routes.
```
tao12345666333: > reach other dockers containers

maybe should change to:

reach other docker containers

aojea: I had the same doubt, but with the amount of "containers", pods, ... that we have in these virtualized environments I think that maybe is good be explicit about this

neiljerram:
@aojea Would this be a good time to talk more about your comment at https://github.com/kubernetes-sigs/kind/pull/939#issuecomment-540186911 ?  I understand that you and @BenTheElder had concerns about my proposal at the time, but I am not sure you are right that my objective can be easily achieved in existing other ways.

aojea:
> @aojea Would this be a good time to talk more about your comment at [#939 (comment)](https://github.com/kubernetes-sigs/kind/pull/939#issuecomment-540186911) ? I understand that you and @BenTheElder had concerns about my proposal at the time, but I am not sure you are right that my objective can be easily achieved in existing other ways.

/hold

@neiljerram my understanding is that you want to automate in KIND:

* attaching new networks to the KIND nodes
* adding custom static routes to the KIND nodes
* adding loopback interfaces with custom IPs to the KIND nodes

if that's correct I can document how to do it, it seems easy to automate with a script or in the same way that kubeadm friends are doing with kinder https://github.com/kubernetes/kubeadm/tree/master/kinder#usage

IMHO that seems a very intrusive and specific change to target multihomed nodes environments, that are not very common on cloud environments ... bear in mind that main goal of KIND is testing Kubernetes.
Personally, what I'm more afraid of is about having more dependencies on `libnetwork`, is really opinionated for things like IPv6 or DNS behavior, I don know what those `docker network connect` commands will change ... 
However, @BenTheElder may have another opinion ...

neiljerram:
Thanks @aojea for your interest in this.

At the time of that PR, I was modelling a multi-homed infrastructure, with two independent planes of connectivity between any two nodes.  Obviously the idea is that one of the connectivity planes fails in some way, we still have connectivity between all the nodes over the other plane.

My reason for thinking that this needs integration in KIND is as follows.

- For a setup like this to be resilient, it is important that connections to or from a node do not use an interface-specific address as their source or destination IP.  Because then the connection cannot continue if the connectivity fails adjacent to that interface-specific address.  So instead we want to set up a "loopback address" on each node and arrange for all outgoing connections to use that address.

- That includes the connections that are established for the functioning of the Kubernetes control place itself.  For example each node's kubelet connecting to the API server should have src IP = loopback address of kubelet node.

- Therefore, I think, the provisioning of the loopback address, and the routing to loopback addresses on other nodes, must be done as part of the KIND cluster setup.

WDYT?  Am I still missing other possible approaches here?

aojea:
/hold cancel
> Thanks @aojea for your interest in this.
> 
> At the time of that PR, I was modelling a multi-homed infrastructure, with two independent planes of connectivity between any two nodes. Obviously the idea is that if one of the connectivity planes fails in some way, we still have connectivity between all the nodes over the other plane.
> 
> My reason for thinking that this needs integration in KIND is as follows.
> 
> * For a setup like this to be resilient, it is important that connections to or from a node do not use an interface-specific address as their source or destination IP.  Because then the connection cannot continue if the connectivity fails adjacent to that interface-specific address.  So instead we want to set up a "loopback address" on each node and arrange for all outgoing connections to use that address.
> * That includes the connections that are established for the functioning of the Kubernetes control plane itself.  For example each node's kubelet connecting to the API server should have src IP = loopback address of kubelet node.
> * Therefore, I think, the provisioning of the loopback address, and the routing to loopback addresses on other nodes, must be done as part of the KIND cluster setup.
> 
> WDYT? Am I still missing other possible approaches here?

yeah, I totally understand your point from the Network engineering perspective, but that setup needs a routing protocol to work and do the failover, I know that Calico and kube-router gives that possibility allowing you to peer with the leaf switches, but as I've said before this is a very specific scenario for bare metal environments, where you don't have an IaaS handling the infrastructure.

For "cloud-native" environments, the IaaS + cloud-controller-manager and Kubernetes + controller loops handle the "resilience" of the environment, i.e. the VMs only need one interface because the network is "virtual" and the IaaS handles it, for the Kubernetes workloads the controller loops handle the pods and services, restarts containers that fail, replaces containers, kills containers that don’t respond to your user-defined health check, and doesn’t advertise them to clients until they are ready to serve.

Specifically to KIND, the network is a Linux bridge, everything is SW and in the same host, if one interface or bridge fails we'll have a bigger problem :smile:

neiljerram:
@aojea Many thanks.  So if I understand correctly, I think your position can be summarised as:

- You agree that KIND-level support would be needed, for someone to correctly model that topology using KIND.

- But you don't want to add the complexity for that to KIND, because you don't think it's an important enough use case for the Kubernetes community.

Is that right?

aojea:
> * You agree that KIND-level support would be needed, for someone to correctly model that topology using KIND.

my point is that I don't see the need to implement it in KIND because you can do it just after the cluster creation, this is an example with bash, using python, go , ... you can easily build much more complex topologies and parametrize it:

```
LOOPBACK_PREFIX="1.1.1."
MY_BRIDGE="my_net2"
MY_ROUTE=10.0.0.0/24
MY_GW=172.16.17.1
# Create 2nd network
docker network create ${MY_BRIDGE}
# Create kubernetes cluster
kind create cluster
# Configure nodes to use the second network
for n in $(kind get nodes); do
  # Connect the node to the second network
  docker network connect ${MY_BRIDGE} ${n}
  # Configure a loopback address
  docker exec ${n} ip addr add ${LOOPBACK_PREFIX}${i}/32 dev lo
  # Add static routes
   docker exec ${n} ip route add ${MY_ROUTE} via {$MY_GW}
done
```

> * But you don't want to add the complexity for that to KIND, because you don't think it's an important enough use case for the Kubernetes community.

Is not just that, KIND is gating kubernetes and is used as CI in a big amount of the Kubernetes ecosystem projects, I'm afraid that the risk of introducing this change could affect the stability of the project, hence all these CIs . You can't imagine the amount of hours that @BenTheElder mainly, @amwat,  I and others have spent debugging flakiness and optimizing KIND

neiljerram:
@aojea 
> my point is that I don't see the need to implement it in KIND because you can do it just after the cluster creation, 

I described in my previous comment why this is not good enough: I need the Kubernetes control plane connections to be using loopback addresses, and IIRC those are set up *during* cluster creation.  Do you think I've got something wrong there?

aojea:
> @aojea
> 
> > my point is that I don't see the need to implement it in KIND because you can do it just after the cluster creation,
> 
> I described in my previous comment why this is not good enough: I need the Kubernetes control plane connections to be using loopback addresses, and IIRC those are set up _during_ cluster creation. Do you think I've got something wrong there?

ok, now I got it, sorry for the confusion but I wasn't understanding your point ... 

It can be done after the cluster setup, is a bit tricky though.

When creating the cluster add the loopback IP address you are going to use for the control-plane to the certificate SAN  (the apiserver binds to "all-interfaces" by default)
```
kind: Cluster
# add the loopback to apiServer cert SANS
kubeadmConfigPatchesJSON6902:
- group: kubeadm.k8s.io
  kind: ClusterConfiguration
  patch: |
    - op: add
      path: /apiServer/certSANs/-
      value: my-loopback
```

After the cluster has been created, modify the kube-apiserver `--advertise-address` flag in `/etc/kubernetes/manifests/kube-apiserver.yaml`
(is a static pod manifest, once you write the file it restarts the pod with the new config)

```
apiVersion: v1
kind: Pod
metadata:
  creationTimestamp: null
  labels:
    component: kube-apiserver
    tier: control-plane
  name: kube-apiserver
  namespace: kube-system
spec:
  containers:
  - command:
    - kube-apiserver
    - --advertise-address=172.17.0.4
```

and then change in all the kubelet the `node-ip` flag 
```
root@kind-worker:/# more /var/lib/kubelet/kubeadm-flags.env 
KUBELET_KUBEADM_ARGS="--container-runtime=remote --container-runtime-endpoint=/run/containerd/containerd.sock --fail-swap-on=false --node-ip=172.17.0.4"
 ```

and restart them `systemctl restart kubelet` to use the new config

**Comment by aojea** [comment]
/assign @BenTheElder 

please review
https://deploy-preview-1337--k8s-kind.netlify.com/docs/user/networking-scenarios/

File: site/content/docs/user/networking-scenarios.md [hunk 4]
```diff
@@ -0,0 +1,228 @@
...
+title: "Using KIND to emulate complex network scenarios"
+menu:
+  main:
+    parent: "user"
+    identifier: "networking-scenarios"
+    weight: 3
+---
+# Using KIND to emulate complex network scenarios [Linux Only]
+
+KIND runs Kubernetes cluster in Docker, and leverages Docker networking for all the network features: portmapping, IPv6, containers connectivity, ...
```
neolit123: `...connectivity, etc.`

neolit123: portmapping -> port mapping

site/content/docs/user/networking-scenarios.md
```diff
@@ -0,0 +1,228 @@
...
+
+{{< codeFromInline lang="bash" >}}
+$ ip addr show docker0
+4: docker0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default 
+    link/ether 02:42:83:eb:5e:67 brd ff:ff:ff:ff:ff:ff
+    inet 172.17.0.1/16 brd 172.17.255.255 scope global docker0
+       valid_lft forever preferred_lft forever
+{{< /codeFromInline >}}
+
+Docker also creates iptables NAT rules on the docker host that masquerade the traffic from the containers connected to docker0 bridge to connect to the outside world.
```
neolit123: the docker host -> the Docker host

site/content/docs/user/networking-scenarios.md
```diff
@@ -0,0 +1,228 @@
...
+* NodePort
+* LoadBalancer
+* Headless
+* ExternalName
+
+In Linux hosts, you can access directly the Cluster IP address of the services just adding one route to the configured **serviceSubnet** parameters via any of the nodes that belong to the cluster, so there is no need to use NodePort or LoadBalancer services.
+
+## Multiple clusters
+
+As we explained before, all KIND clusters are sharing the same docker network, that means that all the cluster nodes have direct connectivity.
```
neolit123: docker network -> Docker network

site/content/docs/user/networking-scenarios.md
```diff
@@ -0,0 +1,228 @@
...
+ip route add 10.110.2.0/24 via 172.17.0.2
+
+$kubectl --context kind-clusterB get nodes -o=jsonpath='{range .items[*]}{"ip route add "}{.spec.podCIDR}{" via "}{.status.addresses[?(@.type=="InternalIP")].address}{"\n"}{end}'
+ip route add 10.120.0.0/24 via 172.17.0.7
+ip route add 10.120.1.0/24 via 172.17.0.6
+ip route add 10.120.2.0/24 via 172.17.0.5
+
+{{< /codeFromInline >}}
+
+Then we just need to install the routes obtained from cluterA in each node of clusterB and viceversa:
```
neolit123: viceversa -> vice versa

site/content/docs/user/networking-scenarios.md
```diff
@@ -0,0 +1,228 @@
...
+    inet 172.17.0.3/16 brd 172.17.255.255 scope global eth0
+{{< /codeFromInline >}}
+
+That means that Pods will be able to reach other Docker containers that does not belong to any KIND cluster, however, the Docker container will not be able to answer to the Pod IP address until we install the correspoding routes.
+
+We can solve it installing routes in the new containers to the Pod Subnets in each Node.
+
+### Example: Multiple network interfaces and Multi-Home Nodes
+
+There can be scenarios that requite multiple interfaces in the KIND nodes to test multi-homing, VLANS, CNI plugins, ...
```
neolit123: `...CNI plugins, etc.`

site/content/docs/user/networking-scenarios.md
```diff
@@ -0,0 +1,228 @@
...
+    inet 127.0.0.1/8 scope host lo
+       valid_lft forever preferred_lft forever
+    inet6 ::1/128 scope host 
+       valid_lft forever preferred_lft forever
+10: eth0@if11: <BROADCAST,MULTICAST,UP,LOWER_UP,M-DOWN> mtu 1500 qdisc noqueue state UP 
+    link/ether 02:42:ac:11:00:03 brd ff:ff:ff:ff:ff:ff
+    inet 172.17.0.3/16 brd 172.17.255.255 scope global eth0
+{{< /codeFromInline >}}
+
+That means that Pods will be able to reach other Docker containers that does not belong to any KIND cluster, however, the Docker container will not be able to answer to the Pod IP address until we install the correspoding routes.
```
neolit123: correspoding -> corresponding

site/content/docs/user/networking-scenarios.md
```diff
@@ -0,0 +1,228 @@
...
+  name: kube-apiserver
+  namespace: kube-system
+spec:
+  containers:
+  - command:
+    - kube-apiserver
+    - --advertise-address=172.17.0.4
+```
+
+and then change in all the nodes the kubelet `node-ip` flag:
```
neolit123: and then change the `node-ip` flag for the kubelets on all the nodes:

site/content/docs/user/networking-scenarios.md
```diff
@@ -0,0 +1,228 @@
...
+
+```
+root@kind-worker:/# more /var/lib/kubelet/kubeadm-flags.env 
+KUBELET_KUBEADM_ARGS="--container-runtime=remote --container-runtime-endpoint=/run/containerd/containerd.sock --fail-swap-on=false --node-ip=172.17.0.4"
+ ```
+
+and restart them `systemctl restart kubelet` to use the new config
```
neolit123: Finally restart the kubelets to use the new configuration with `systemctl restart kubelet`.

neolit123: important to note here is that calling kubeadm init / join again on the node will override `/var/lib/kubelet/kubeadm-flags.env`. alternative is to use `/etc/default/kubelet`
https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/kubelet-integration/#the-kubelet-drop-in-file-for-systemd

aojea: let me add it as a note, due to the ephemeral nature of the nodes I don't expect people to issue those commands ... but :man_shrugging:

site/content/docs/user/networking-scenarios.md
```diff
@@ -217,12 +217,13 @@ spec:
...
 
 ```
 root@kind-worker:/# more /var/lib/kubelet/kubeadm-flags.env 
 KUBELET_KUBEADM_ARGS="--container-runtime=remote --container-runtime-endpoint=/run/containerd/containerd.sock --fail-swap-on=false --node-ip=172.17.0.4"
  ```
 
-and restart them `systemctl restart kubelet` to use the new config
+Finally restart the kubelets to use the new configuration with `systemctl restart kubelet`.
 
+It's important to note that calling `kubeadm init / join` again on the node will override `/var/lib/kubelet/kubeadm-flags.env`. An [alternative is to use /etc/default/kubelet](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/kubelet-integration/#the-kubelet-drop-in-file-for-systemd).S
```
neolit123: trailing `S` after the `.`

site/content/docs/user/networking-scenarios.md
```diff
@@ -0,0 +1,229 @@
...
+    inet 127.0.0.1/8 scope host lo
+       valid_lft forever preferred_lft forever
+    inet6 ::1/128 scope host 
+       valid_lft forever preferred_lft forever
+10: eth0@if11: <BROADCAST,MULTICAST,UP,LOWER_UP,M-DOWN> mtu 1500 qdisc noqueue state UP 
+    link/ether 02:42:ac:11:00:03 brd ff:ff:ff:ff:ff:ff
+    inet 172.17.0.3/16 brd 172.17.255.255 scope global eth0
+{{< /codeFromInline >}}
+
+That means that Pods will be able to reach other Docker containers that does not belong to any KIND cluster, however, the Docker container will not be able to answer to the Pod IP address until we install the corresponding routes.
```
kihahu: Since you are referring to multiple containers, use `do` instead of `does`

BenTheElder:
sorry for the immense delay. I'd been hoping to get #148 done faster. I'd still like to hold off detailing networking internals until after I'm doing taking a swing at changing them :D

aojea:
/hold
this need to be updated to match current status, we have custom bridges now and cluster restart :smile:

site/content/docs/user/networking-scenarios.md
```diff
@@ -0,0 +1,229 @@
...
+We can obtain the routes using kubectl:
+
+{{< codeFromInline lang="bash" >}}
+$ kubectl --context kind-clusterA get nodes -o=jsonpath='{range .items[*]}{"ip route add "}{.spec.podCIDR}{" via "}{.status.addresses[?(@.type=="InternalIP")].address}{"\n"}{end}'
+ip route add 10.110.0.0/24 via 172.17.0.4
+ip route add 10.110.1.0/24 via 172.17.0.3
+ip route add 10.110.2.0/24 via 172.17.0.2
+
+$kubectl --context kind-clusterB get nodes -o=jsonpath='{range .items[*]}{"ip route add "}{.spec.podCIDR}{" via "}{.status.addresses[?(@.type=="InternalIP")].address}{"\n"}{end}'
+ip route add 10.120.0.0/24 via 172.17.0.7
```
masaeedu, line 115: Is this supposed to be 220? Also why are there three results here when each cluster has two nodes?

aojea: heh, good catch on both things, is 220 and the config should have 3 nodes

PR #1341, opened 2020-02-20
More podman fixes + provider auto detection

amwat:
Fixes node image names to be podman compatible.
Add basic provider auto detection.

/cc @BenTheElder

### DIFF
--- a/pkg/cluster/internal/providers/docker/images.go
+++ b/pkg/cluster/internal/providers/docker/images.go
@@ -36,10 +36,7 @@ func ensureNodeImages(logger log.Logger, status *cli.Status, cfg *config.Cluster
 	// pull each required image
 	for _, image := range common.RequiredNodeImages(cfg).List() {
 		// prints user friendly message
-		friendlyImageName := image
-		if strings.Contains(image, "@sha256:") {
-			friendlyImageName = strings.Split(image, "@sha256:")[0]
-		}
+		friendlyImageName, image := sanitizeImage(image)
 		status.Start(fmt.Sprintf("Ensuring node image (%s) 🖼", friendlyImageName))
 		if _, err := pullIfNotPresent(logger, image, 4); err != nil {
 			status.End(false)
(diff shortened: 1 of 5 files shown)


pkg/cluster/internal/providers/docker/images.go
```diff
@@ -36,10 +36,7 @@ func ensureNodeImages(logger log.Logger, status *cli.Status, cfg *config.Cluster
 	// pull each required image
 	for _, image := range common.RequiredNodeImages(cfg).List() {
 		// prints user friendly message
-		friendlyImageName := image
-		if strings.Contains(image, "@sha256:") {
-			friendlyImageName = strings.Split(image, "@sha256:")[0]
-		}
+		friendlyImageName, image := sanitizeImage(image)
```
aojea: why do we return 2 values, one overwritting `image`?

pkg/cluster/provider.go
```diff
@@ -58,13 +59,22 @@ func NewProvider(options ...ProviderOption) *Provider {
...
-		// TODO: consider auto-detection once more than 1 provider is stable
-		providerEnv := os.Getenv("KIND_EXPERIMENTAL_PROVIDER")
-		if providerEnv == "podman" {
+
+	supportedProviders := map[string]func(log.Logger) internalprovider.Provider{
+		"podman": podman.NewProvider,
+		"docker": docker.NewProvider,
+	}
+
+	// check if provider was overridden
```
aojea: shouldn't this only be checked if `p.provider == nil`?

amwat: yes, but currently at this point `p.provider` is always `nil`.

that being said I will closely be sending a follow up PR that will expose the provider as a ProviderOption, and probably also move the `KIND_EXPERIMENTAL_PROVIDER` up to the cli.

pkg/cluster/internal/providers/docker/images.go
amwat: https://github.com/kubernetes-sigs/kind/blob/460c00b8aa035c49e790541f923d117b7026e0ac/pkg/cluster/internal/providers/podman/images.go#L84-L85

For docker it's not that useful, but for podman all 3 images are different. Added it to both to have the same flow.
We don't _need_ to overwrite `image`  here, but /shrug reused the same variable instead of defining a new one.

pkg/cluster/internal/providers/podman/provider.go
```diff
@@ -42,6 +42,7 @@ import (
 
 // NewProvider returns a new provider based on executing `podman ...`
 func NewProvider(logger log.Logger) provider.Provider {
+	logger.Warn("enabling experimental podman provider")
```
BenTheElder: in the future it might be nice to point to how to contact us along with this message

pkg/cluster/provider.go
BenTheElder: I agree that we should check for nil but also think that follow up should happen

PR #1348, opened 2020-02-21
drain reader when untarring

BenTheElder:
/👏👏 @amwat

### DIFF
--- a/pkg/cluster/internal/logs/logs.go
+++ b/pkg/cluster/internal/logs/logs.go
@@ -20,6 +20,7 @@ import (
 	"archive/tar"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"path"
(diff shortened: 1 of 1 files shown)


pkg/cluster/internal/logs/logs.go
```diff
@@ -63,7 +64,10 @@ func untar(logger log.Logger, r io.Reader, dir string) (err error) {
 
 		switch {
 		case err == io.EOF:
-			return nil
+			// drain the reader, which may have trailing null bytes
+			// we don't want to leave the writer hanging
+			_, err := io.Copy(ioutil.Discard, r)
```
BenTheElder: (tar writes in fixed size blocks, there is padding)

aojea: excellent

PR #1351, opened 2020-02-21
fix abspath resolution on windows

### LINKED ISSUE #1324: Abspath is incorrectly mangling hostPath on Windows
**What happened**:
I was unable to start kind on Windows with `hostPath` set to `/c/temp`

I expected to start kind on Windows with `hostPath` set to `/c/temp` 

Start kind with the following config

*cfg.yml*

```
kind: Cluster
nodes:
- role: control-plane
  extraMounts:
    - containerPath: /opt/local-path-provisioner
      hostPath: /c/temp
      readOnly: false
```

```bash
kind create cluster --config cfg.yml
```

I applied the following workaround

```diff
--- a/pkg/cluster/internal/providers/docker/provision.go	2020-02-10 22:28:08.601828100 +0200
+++ b/pkg/cluster/internal/providers/docker/provision.go	2020-02-10 22:29:04.711607700 +0200
@@ -19,7 +19,6 @@
 import (
 	"fmt"
 	"net"
-	"path/filepath"
 	"strings"
 
 	"sigs.k8s.io/kind/pkg/cluster/constants"
@@ -61,16 +60,6 @@
 		node := node.DeepCopy()              // copy so we can modify
 		name := nodeNamer(string(node.Role)) // name the node
 
-		// fixup relative paths, docker can only handle absolute paths
-		for i := range node.ExtraMounts {
-			hostPath := node.ExtraMounts[i].HostPath
-			absHostPath, err := filepath.Abs(hostPath)
-			if err != nil {
-				return nil, errors.Wrapf(err, "unable to resolve absolute path for hostPath: %q", hostPath)
-			}
-			node.ExtraMounts[i].HostPath = absHostPath
-		}
-
 		// plan actual creation based on role
 		switch node.Role {
 		case config.ControlPlaneRole:
```

**Environment:**

- kind version: (use `kind version`): v0.7.0

  ```
  Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.0", GitCommit:"70132b0f130acc0bed193d9ba59dd186f0e634cf", GitTreeState:"clean", BuildDate:"2019-12-07T21:20:10Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"windows/amd64"}
  Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.0", GitCommit:"70132b0f130acc0bed193d9ba59dd186f0e634cf", GitTreeState:"clean", BuildDate:"2020-01-14T00:09:19Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"linux/amd64"}
  ```

  ```
  Client:
  Debug Mode: false

  Server:
  Containers: 2
    Running: 2
    Paused: 0
    Stopped: 0
  Images: 1
  Server Version: 19.03.5
  Storage Driver: overlay2
    Backing Filesystem: extfs
    Supports d_type: true
    Native Overlay Diff: true
  Logging Driver: json-file
  Cgroup Driver: cgroupfs
  Plugins:
    Volume: local
  Swarm: inactive
  Runtimes: runc
  Default Runtime: runc
  Init Binary: docker-init
  containerd version: b34a5c8af56e510852c35414db4c1f4fa6172339
  runc version: 3e425f80a8c931f88e6d94a8c831b9d5aa481657
  init version: fec3683
  Security Options:
    seccomp
    Profile: default
  Kernel Version: 4.14.154-boot2docker
  Operating System: Boot2Docker 19.03.5 (TCL 10.1)
  OSType: linux
  Architecture: x86_64
  CPUs: 1
  Total Memory: 989.5MiB
  Name: kind
  ID: A4W7:F5TO:SQMC:2XKF:MING:LCHG:ME5N:CURZ:QTXN:5PXZ:F7YW:H3ZS
  Docker Root Dir: /mnt/sda1/var/lib/docker
  Debug Mode: false
  Labels:
    provider=virtualbox
  Experimental: false
  Insecure Registries:
    127.0.0.0/8
  Live Restore Enabled: false
  ```

  ```
  Microsoft Windows [Version 10.0.18363.628]
  ```

BenTheElder:
fixes #1324

### DIFF
--- a/pkg/cluster/internal/providers/docker/provision.go
+++ b/pkg/cluster/internal/providers/docker/provision.go
@@ -25,6 +25,7 @@ import (
 	"sigs.k8s.io/kind/pkg/cluster/constants"
 	"sigs.k8s.io/kind/pkg/errors"
 	"sigs.k8s.io/kind/pkg/exec"
+	"sigs.k8s.io/kind/pkg/fs"
 
 	"sigs.k8s.io/kind/pkg/cluster/internal/loadbalancer"
 	"sigs.k8s.io/kind/pkg/cluster/internal/providers/provider/common"
@@ -64,11 +65,13 @@ func planCreation(cluster string, cfg *config.Cluster) (createContainerFuncs []f
 		// fixup relative paths, docker can only handle absolute paths
 		for i := range node.ExtraMounts {
 			hostPath := node.ExtraMounts[i].HostPath
-			absHostPath, err := filepath.Abs(hostPath)
-			if err != nil {
-				return nil, errors.Wrapf(err, "unable to resolve absolute path for hostPath: %q", hostPath)
+			if !fs.IsAbs(hostPath) {
+				absHostPath, err := filepath.Abs(hostPath)
+				if err != nil {
+					return nil, errors.Wrapf(err, "unable to resolve absolute path for hostPath: %q", hostPath)
+				}
+				node.ExtraMounts[i].HostPath = absHostPath
 			}
-			node.ExtraMounts[i].HostPath = absHostPath
 		}
 
 		// plan actual creation based on role
--- a/pkg/fs/fs.go
+++ b/pkg/fs/fs.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"path"
 	"path/filepath"
 	"runtime"
 	"strings"
@@ -41,6 +42,13 @@ func TempDir(dir, prefix string) (name string, err error) {
 		name = filepath.Join("/private", name)
 	}
 	return name, nil
+}
+
+// IsAbs is like filepath.IsAbs but also considering posix absolute paths
+// to be absolute even if filepath.IsAbs would not
+// This fixes the case of Posix paths on Windows
+func IsAbs(hostPath string) bool {
+	return path.IsAbs(hostPath) || filepath.IsAbs(hostPath)
 }
 
 // Copy recursively directories, symlinks, files copies from src to dst


pkg/cluster/internal/providers/docker/provision.go
```diff
@@ -64,11 +65,13 @@ func planCreation(cluster string, cfg *config.Cluster) (createContainerFuncs []f
...
 			hostPath := node.ExtraMounts[i].HostPath
-			absHostPath, err := filepath.Abs(hostPath)
-			if err != nil {
-				return nil, errors.Wrapf(err, "unable to resolve absolute path for hostPath: %q", hostPath)
+			if !fs.IsAbs(hostPath) {
+				absHostPath, err := filepath.Abs(hostPath)
+				if err != nil {
+					return nil, errors.Wrapf(err, "unable to resolve absolute path for hostPath: %q", hostPath)
+				}
+				node.ExtraMounts[i].HostPath = absHostPath
```
amwat: should be outside the if?

BenTheElder: pseudo-code of the current logic:
```
if not absolute(path):
  make_absolute(path)
```
so no?

aojea: why not detecting the the OS?
`if runtime.GOOS == "windows" `

amwat: oh right, nvm

BenTheElder: the behavior is correct on linux with the current very simple implementation

remote hosts make this tricky and will need a more elaborate logic, which I'm mulling, but the priority is:
- preserve the existing behavior generally
- fix the case of posix absolute path on windows

this is the minimum implementation of that and pretty simple / cheap in the common case, on linux it just checks the first character at most twice (in the case that it is relative)

BenTheElder: remote hosts (even locally remote, think WSL2) make this a bit more complex, I think we can solve it but I wanted to go ahead and see if we could sort out the original issue cheaply and get that verified

PR #1357, opened 2020-02-24
Use text/template instead of html/template

darkowlzz:
Replaces html/template package usage with text/template since we are
parsing yaml manifests, not html.


BenTheElder:
Thanks.
FYI that this is an internal unsupported implementation detail if you are planning to leverage it :+)

Issue #1347, opened 2020-02-21
Getting Kind works on ARM64


zhlhahaha:
We are trying to make kind works on ARM64 platform.
I summarize some task may need to do. Looking forward your suggestion
Tasks:
* [ ] 1 Ensure all codes can be built successfully on Arm64.
* [ ] 2 Enable kind cross build base images and node images for ARM platform on x86.
* [ ] 3 Get kind CI works on ARM64 platform
* [ ] 4 e2e-kind passed on Arm

zhlhahaha:
@BenTheElder

BenTheElder:
There are existing issues for this that should be looked at.

1) kind can be build for arm64, we publish binaries already, just not the node / base image, for which we only pre-publish a small set of images at the moment.

2) This is not as simple as it sounds. The node image build is not a normal image build, and I'm not sure this is actually something we're interested in at the moment.

3) Our CI is prow, see github.com/kubernetes/test-infra. It is not natively hosted on arm and I do not wish to own any additional infra myself at the moment.

4) kind e2e has already been known to work on arm. we helped figure this out before but nobody maintained it... https://testgrid.k8s.io/conformance-all#kind,%20v1.14%20(dev,%20ARM64)

BenTheElder:
please see also https://github.com/kubernetes-sigs/kind/issues/166 https://github.com/kubernetes-sigs/kind/issues/188 ...

BenTheElder:
kind is great in kubernetes CI because we can run it natively in the CI. Given that we need external infra to run ARM CI anyhow, I'm not really sure kind is an improvement over just `kubeadm init` on whatever external ARM box we have anyhow...?

even for development clusters, see https://github.com/kubernetes/minikube/issues/955

BenTheElder:
I do want KIND to work on ARM long term, but from what I've gathered the goal here is testing kubernetes on ARM upstream for "official support" ...? In which case I'm not positive this is the most efficient work stream.

zhlhahaha:
> please see also #166 #188 ...

The issue is more complex than I thought, but we can make it step by step. 
Thanks for your suggestion and informative comments.
I will read though your comments and related link first.

lubinszARM:
From my perspective, I think the enablement of Prow on Arm64 is more helpful to our goal here is  testing k8s on ARM for "official support".
We have tested the local Prow on x86. And I think the enablement of Prow on Arm64 is our focus for the our next job.
[...]

Issue #188, opened 2018-12-19
ARM64 CI


BenTheElder:
per discussion in #kind slack, we should setup some CI with openlab to get `kind` on arm64 xref #166 

@dims was able to get arm64 working, but we'll need some set this up to keep it working once that goes in, as the maintainers do not have access to suitable arm machines to test on otherwise.

/assign
/kind feature
/priority important-longterm

dixudx:
@lubinsz You might help on this?

BenTheElder:
note that we will need to fix #166 first, however that is very doable. Dims previously made a quick patch that worked, but we haven't PRed anything yet.

lubinsz:
@BenTheElder @dixudx 
Let me check it.

dims:
@lubinsz see my previous patch in https://github.com/kubernetes-sigs/kind/issues/166#issuecomment-448766016

hh:
https://github.com/WorksOnArm/cluster/issues/154 gives us access to Packet hardware.
How would we like it configured?

- Running k8s?
This would allow us to specify this cluster to run kind jobs and fit into sig-testings usual approach to testing.

- running docker only
We would need to ssh in, run kind + test, + cleanup.

/cc @dvn-ii

BenTheElder:
I think running docker / SSH only is the most well understood path currently, we can treat these similar to a node or cadvisor e2e job, and put credentials in Prow to access them.

Long term it might be interesting to be able to run prowjobs on these machines directly, but that will require more work to maintain the cluster and it will take figuring out how we want to handle distributing other credentials.

hh:
When trying to kind build, we note that docker-ce and friends are not available directly from the same repos:

```
E: Version '18.06.*' for 'docker-ce' was not found
The command '/bin/sh -c curl -fsSL "https://download.docker.com/linux/$(. /etc/os-release; echo "$ID")/gpg" | apt-key add -     && apt-key fingerprint 0EBFCD88     && ARCH="${ARCH}" add-apt-repository         "deb [arch=${ARCH}] https://download.docker.com/linux/$(. /etc/os-release; echo "$ID") $(lsb_release -cs) stable"     && clean-install "docker-ce=${DOCKER_VERSION}"' returned a non-zero code: 100
ERRO[22:13:41] Docker build Failed! exit status 100         
Error: build failed: exit status 100
```

Probably need some debugging:

```
root@kind:~# kind create cluster
Creating cluster "kind" ...
[...]

PR #1366, opened 2020-02-27
Base image updates

BenTheElder:
- newest containerd 1.3 which has some fixes
- initial base image cross build
ref: #188 #166 #1347 ...

this is _much_ slower and problematic to locally test (the manifest list cannot be exported to local, so you need to do a quick build and test that) but, it is feasible. starting with just amd64 and arm64.


pkg/build/node/node.go
```diff
@@ -42,7 +42,7 @@ import (
 const DefaultImage = "kindest/node:latest"
 
 // DefaultBaseImage is the default base image used
-const DefaultBaseImage = "kindest/base:v20200221-f3727da0@sha256:a9da2114e7ea293d9ac544a0c5b58d63e5ee44b4e76891bf35671fb1f5ffc611"
+const DefaultBaseImage = "kindest/base:v20200226-6d5ec59c@sha256:7fbe0890dab2dfb170fef3bf8ca8dda9bae40c66e6124a59ce2fd2debd18e818"
```
BenTheElder: you cannot browse the digest on docker hub, but the tag is browsable, so we include both.
this is the digest of a manifest list for linux/amd64 + linux/arm64

images/base/Makefile
```diff
@@ -0,0 +1,27 @@
+TAG?=$(shell echo "$$(date +v%Y%m%d)-$$(git describe --always --dirty)")
+IMAGE?=kindest/base:$(TAG)
+
+# build with buildx
+PLATFORMS?=linux/amd64,linux/arm64
```
BenTheElder: we'll probably want to expand this, however I'm trying to wrap this up and make sure the base image works at all, my main priority was the containerd update, this change is more of a diversion due to #1347 #166

aojea:
Cool

BenTheElder:
FYI @neolit123

neolit123:
ack, thanks for the heads up.

BenTheElder:
shouldn't affect anything I think, BUT it is a manifest list image instead now, and it does mean some movement towards arm being a thing.

I'm guessing kinder users are all amd64 but I wouldn't be too surprised if someone was trying to use another architecture :sweat_smile:

BenTheElder:
(and the image build is changing, we really need to remove `kind build base-image` which is outdated and have people leverage the makefile / not need to build it themselves since we'll have pre-cross-built-it for them)

neolit123:
we have quite a few amd64 jobs for kinder right now and the amd64 assumptions are hardcoded.

i'm hoping that one day we will be able to run at least one arm64 job too.

since we don't have a way to provision arm64 with prow today (AFAIK), i was wondering if its possible to try it in github actions, by [requesting a non-amd64 VM in the workflow job](https://github.com/uraimo/run-on-arch-action#optional-parameters) and running kind/kinder there with k8s non-amd64 artifacts.

but this has been in the backlog, due to other priorities.

BenTheElder:
Ack, thanks for the details! I'll follow up on this in the future.

BenTheElder:
Looks like that action is just a qemu wrapper, the performance will probably be too bad.

Travis has ACTUAL arm IIRC but getting it in k8s may be hard.

neolit123:
> Looks like that action is just a qemu wrapper, the performance will probably be too bad.

i didn't know it is using emulation. this might be quite slow, indeed.

BenTheElder:
locally using qemu to build the image is pretty slow, and I'm betting the actions workers are much smaller machines.

PR #1367, opened 2020-02-27
Include `golang` in the base image

chendave:
This is important for a project written by `golang`

see the discussion here: https://github.com/kubernetes/kubernetes/pull/88399#discussion_r383107339

This has blocked the gating for the PR.

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -66,7 +66,7 @@ RUN echo "Ensuring scripts are executable ..." \
     && DEBIAN_FRONTEND=noninteractive clean-install \
       systemd \
       conntrack iptables iproute2 ethtool socat util-linux mount ebtables udev kmod \
-      libseccomp2 \
+      libseccomp2 golang \
       bash ca-certificates curl rsync \
     && find /lib/systemd/system/sysinit.target.wants/ -name "systemd-tmpfiles-setup.service" -delete \
     && rm -f /lib/systemd/system/multi-user.target.wants/* \


BenTheElder:
i'm sorry but we are NOT going to do this.
the base image is used to bundle up dependencies of running kubernetes.
keeping it small is very important.

golang is not a runtime requirement for kubernetes.

BenTheElder:
this discussion? https://github.com/kubernetes/kubernetes/pull/88399#discussion_r383107339

this is the wrong image. the image there is in the test-infra repo, not KIND.

chendave:
okay, so is there any other way to include the "golang" for the gating?

chendave:
see the error here: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/88399/pull-kubernetes-e2e-kind/1232859645786198016#1:build-log.txt%3A881

patch test-infra has been done here and merged.
https://github.com/kubernetes/test-infra/pull/16447

BenTheElder:
Your test-infra changes are not immediately deployed.
I will look at the error when I get back to a computer.

BenTheElder:
If you click the "yaml" on that page you can see that a different image is used.

We should also not use the Debian go package, it is not the right version for Kubernetes

chendave:
Thank you!

test-infra is effective I think since it has been more than 24 hours.
CI failed with pull-kubernetes-e2e-kind and pull-kubernetes-e2e-kind-ipv6 actaully.

see the yaml, the image is gcr.io/k8s-testimages/krte:v20200212-1f7b8ac-master, do you know where is the source for it?

BenTheElder:
> test-infra is effective I think since it has been more than 24 hours.

Not necessarily. There must be a human approved bump, and in this case we rarely use that image directly, it must be bumped in the `FROM` of another image and then that image bumped.

but we're not using that image here anyhow

> see the yaml, the image is gcr.io/k8s-testimages/krte:v20200212-1f7b8ac-master, do you know where is the source for it?

in test infra, image/krte

however I'm currently NACK on the k/k change, will discuss there

PR #1368, opened 2020-02-27
Adds CommandContext to the Cmder interface

### LINKED ISSUE #1364: Add a CommandContext to the exec package
The standard library exposes the `exec.CommandContext` function that takes a context. It would be great if kind's exec library also exposed the context version of the exec.Command function.

This could be an additional function added to `LocalCmder` and either a new interface or a new function added to the `Cmder` interface.

I would be happy to open a PR for this if y'all are interested in adding this to your exec package!

**What should be cleaned up or changed**:
This should be added because there is no convenient way to pass a context into the command library. Sometimes commands can hang indefinitely and it would be really nice to add a timeout in some cases.

**Why is this needed**:
Being able to cancel long running functions in a reasonable amount of time is nice.

chuckha:
Signed-off-by: Chuck Ha <chuckh@vmware.com>

Fixes #1364

/assign @BenTheElder

### DIFF
--- a/pkg/build/node/internal/container/docker/exec.go
+++ b/pkg/build/node/internal/container/docker/exec.go
@@ -17,6 +17,7 @@ limitations under the License.
 package docker
 
 import (
+	"context"
 	"io"
 
 	"sigs.k8s.io/kind/pkg/exec"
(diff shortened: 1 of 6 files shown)


pkg/build/node/internal/container/docker/exec.go
```diff
@@ -83,6 +94,9 @@ func (c *containerCmd) Run() error {
 		c.args...,
 	)
 	cmd := exec.Command("docker", args...)
+	if c.ctx != nil {
```
BenTheElder: Probably should be an if else?

chuckha: sure, no problems!

chuckha: how does this look?

BenTheElder: great, thanks!

BenTheElder:
/lgtm
/approve
thanks!
this will be super useful

BenTheElder:
prow
/retest

PR #1372, opened 2020-02-28
Fix make lint

amwat:
lint inside go_container.

/cc @BenTheElder

### DIFF
--- a/hack/verify/lint.sh
+++ b/hack/verify/lint.sh
@@ -21,10 +21,8 @@ REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)"
 cd "${REPO_ROOT}"
 
 # build golangci-lint
-SOURCE_DIR="${REPO_ROOT}/hack/tools" hack/go_container.sh \
-  go build -o /out/golangci-lint github.com/golangci/golangci-lint/cmd/golangci-lint
 SOURCE_DIR="${REPO_ROOT}/hack/tools" GOOS="linux" hack/go_container.sh \
-  go build -o /out/golangci-lint-linux github.com/golangci/golangci-lint/cmd/golangci-lint
+  go build -o /out/golangci-lint github.com/golangci/golangci-lint/cmd/golangci-lint
 
 # run golangci-lint
 LINTS=(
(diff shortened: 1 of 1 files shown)


amwat:
also IIUC plumbing GOOS through actually breaks on mac, since our container is running inside a linux vm.
but I might be missing something @BenTheElder ?

hack/go_container.sh
```diff
@@ -38,22 +38,7 @@ DOCKER="${DOCKER:-"$(which docker || which podman || echo "docker")"}"
 # ========================== END SCRIPT SETTINGS ===============================
 
 # autodetects host GOOS and GOARCH and exports them if not set
-detect_and_set_goos_goarch() {
```
BenTheElder: No. This breaks make install

hack/verify/lint.sh
```diff
@@ -36,7 +34,8 @@ LINTS=(
 )
 ENABLE=$(sed 's/ /,/g' <<< "${LINTS[@]}")
 # first for the repo in general
-GO111MODULE=on bin/golangci-lint --disable-all --enable="${ENABLE}" run ./pkg/... ./cmd/... .
+SOURCE_DIR="${REPO_ROOT}" hack/go_container.sh \
+  /out/golangci-lint --disable-all --enable="${ENABLE}" run ./pkg/... ./cmd/... .
 # ... and then for kindnetd, which is only on linux
 SOURCE_DIR="${REPO_ROOT}/images/kindnetd" GOOS="linux" "${REPO_ROOT}/hack/go_container.sh" \
```
BenTheElder: We set GOOS here

hack/verify/lint.sh
```diff
@@ -23,8 +23,6 @@ cd "${REPO_ROOT}"
 # build golangci-lint
 SOURCE_DIR="${REPO_ROOT}/hack/tools" hack/go_container.sh \
   go build -o /out/golangci-lint github.com/golangci/golangci-lint/cmd/golangci-lint
-SOURCE_DIR="${REPO_ROOT}/hack/tools" GOOS="linux" hack/go_container.sh \
```
BenTheElder: And here

BenTheElder:
It could make sense to autodetect GOOS only during make build, but the current change leaves that broken.

PR #1376, opened 2020-02-29
remove kind build base-image

BenTheElder:
the base image should be built with the makefile from a checkout, this subcommand is lingering tech debt and out of sync.


tao12345666333:
/lgtm

Some documents need to be updated afterwards.

```
(MoeLove) ➜  kind git:(master) ✗ rg 'kind build base-image'
images/base/README.md
6:The image can be built with `kind build base-image`.

site/content/docs/design/node-image.md
19:and they can be built with `kind build base-image` and `kind build node-image` respectively.

site/content/docs/user/quick-start.md
239:kind build base-image
252:kind build base-image --image base:v0.1.0

```

BenTheElder:
thanks!

PR #1379, opened 2020-03-03
Config fixup

BenTheElder:
fix after https://github.com/kubernetes-sigs/kind/pull/1274 🤦‍♂ `-}}` was eating load-bearing whitespace

cc @liggitt

### DIFF
--- a/pkg/cluster/internal/kubeadm/config.go
+++ b/pkg/cluster/internal/kubeadm/config.go
@@ -308,15 +308,15 @@ controlPlaneEndpoint: "{{ .ControlPlaneEndpoint }}"
 # to the cluster after rewriting the kubeconfig to point to localhost
 apiServer:
   certSANs: [localhost, "{{.APIServerAddress}}"]
-{{ if .FeatureGates -}}
+{{ if .FeatureGates }}
   extraArgs:
     "feature-gates": "{{ .FeatureGatesString }}"
-{{- end}}
+{{ end}}
 controllerManager:
-{{ if .FeatureGates -}}
+{{ if .FeatureGates }}
   extraArgs:
     "feature-gates": "{{ .FeatureGatesString }}"
-{{- end}}
+{{ end}}
     enable-hostpath-provisioner: "true"
     # configure ipv6 default addresses for IPv6 clusters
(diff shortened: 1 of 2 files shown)


pkg/cluster/internal/kubeadm/config.go
```diff
@@ -359,7 +359,7 @@ apiVersion: kubeadm.k8s.io/v1beta1
 kind: JoinConfiguration
 metadata:
   name: config
-{{ if .ControlPlane -}}
```
BenTheElder: stray tab

pkg/cluster/internal/patch/matchinfo.go
```diff
@@ -33,7 +33,7 @@ type matchInfo struct {
 func parseYAMLMatchInfo(raw string) (matchInfo, error) {
 	m := matchInfo{}
 	if err := yaml.Unmarshal([]byte(raw), &m); err != nil {
-		return matchInfo{}, errors.Wrap(err, "failed to parse type meta")
+		return matchInfo{}, errors.Wrapf(err, "failed to parse type meta for %q", raw)
```
BenTheElder: not overly settled on `%q` /shrug, we have a number of other places that need more helpful failure output, I have an enhanced error design stashed for a quiet day :+)

BenTheElder:
prow flake
/retest

BenTheElder:
/retest
...

filbranden:
My rule of thumb for template directives is to always use `{{- ... }}`.

The closing `}}` will *never* eat into indentation of the next line.

On the other hand, the `{{-` will eat into trailing whitespace of the previous line. (There should be none and even if there was some, if it needs to be preserved you have a larger problem.)

Having one side with the `-` means you'll eat one of the newlines (in this case, the one preceding it), so you don't get a spare newline for template directives.

Furthermore, you're now free to indent your template directives, since the `{{-` will eat that indentation too.

Win win!

BenTheElder:
thanks @filbranden that makes sense to me, at this point I was pretty annoyed with go template and just tore out all the `-` but it sounds good to keep it on the closing side 🙃 

also, hi!

PR #1380, opened 2020-03-03
add ci go caching script

BenTheElder:
we can use this in CI like:
```yaml
env:
- name: DOWNLOAD_CACHE
   value: "true"
command:
- hack/ci/cache-wrapper.sh
- make
- build
```
It will preserve the exit code from calling $@, before calling $@ it will download a tarball and populate the go cache.
In postsubmit we have it populate the cache.

### DIFF
--- a/hack/ci/cache-wrapper.sh
+++ b/hack/ci/cache-wrapper.sh
@@ -1,0 +1,57 @@
+#!/bin/bash
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


hack/ci/cache-wrapper.sh
```diff
@@ -0,0 +1,57 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# USAGE: cache-wrapper.sh hack/go_container.sh some-go-command
+# $@ will be executed with this script wrapping cache upload / download
+
+set -o errexit -o nounset -o pipefail
+
+# options for where the cache is stored
+BUCKET="${BUCKET:-bentheelder-kind-ci-builds}"
```
BenTheElder: I'm going to ask for a bucket from wg-k8s and move everything over, but this will do until that lands

PR #1391, opened 2020-03-10
Add docs note on how to list loaded images

JamesLaverack:
This can aid debugging, and help to understand the state of the cluster.

This delightful one-liner courtesy of @jlaswell

### DIFF
--- a/site/content/docs/user/quick-start.md
+++ b/site/content/docs/user/quick-start.md
@@ -151,6 +151,13 @@ kind load docker-image my-custom-image:unique-tag
 kubectl apply -f my-manifest-using-my-image:unique-tag
 ```
 
+**Note**: You can get a list of images present on a cluster node by
+using `docker exec`:
+```
+docker exec -it my-node-name crictl images
+```
+Where `my-node-name` is the name of the Docker container.
+
 **Note**: The Kubernetes default pull policy is `IfNotPresent` unless
 the image tag is `:latest` in which case the default policy is `Always`.
 `IfNotPresent` causes the Kubelet to skip pulling an image if it already exists.


site/content/docs/user/quick-start.md
```diff
@@ -151,6 +151,13 @@ kind load docker-image my-custom-image:unique-tag
 kubectl apply -f my-manifest-using-my-image:unique-tag
 ```
 
+**Note**: You can get a list of images present on a cluster node by
+using `docker exec`:
+```
+docker exec -it my-node-name crictl images
```
munnerz: This exposes some of the implementation details of kind which _could_ change in future. The only way to really avoid that though would be to add a CLI command to kind itself for this, which may or may not be desirable.

Overall I think this is meaningful/helpful to include, but maybe needs a warning on it or needs to be made less prominent?

/cc @BenTheElder 
/ok-to-test

JamesLaverack: I also opened #1393 to propose a change to the kind CLI to achieve the same thing.

BenTheElder: I'm in favor of doing this, I think the future is CRI, and node names are meant to be reliable for this.
if for some reason we support dockershim in the future AND it's baked into kubelet / not CRI compatible still we can just update the note here to show both ways or something

site/content/docs/user/quick-start.md
```diff
@@ -151,6 +151,13 @@ kind load docker-image my-custom-image:unique-tag
...
+```
+Where `my-node-name` is the name of the Docker container.
```
BenTheElder: we should mention `kind get nodes` or `kubectl get nodes` and specifically that we're talking about a node container

BenTheElder: can be done in a follow up.

BenTheElder:
thanks!

BenTheElder:
ignoring irrelevant CI, and triggering our new build postsubmit anyhow...

Issue #1291, opened 2020-01-27
Adding API Options?


robotica72:
Ive been playing around with trying to get kind to deploy using an OIDC provider.  Overall I dont see why it wouldnt work, but I cant get the install to add the options to the API server.  Ive tried using the kubeadm format in my create cluster file, similar to the featuregates but the install fails with:

ERROR: failed to create cluster: failed to init node with kubeadm: command "docker exec --privileged kind-control-plane kubeadm init --ignore-preflight-errors=all --config=/kind/kubeadm.conf --skip-token-print --v=6" failed with error: exit status 1

Happy to add what I tried in the config, but I thought I would ask to make sure it was possible at all before digging into it more.

BenTheElder:
You can set options on the api server.

can you share:
- the full CLI you used
- the kind version

when debugging, increasing verbosity of `kind` helps (`-v 1`)

robotica72:
Hi Ben - That was a quick reply!   Thanks for the information.

I just managed to get it working - when I had an issue I backed off the extra options and I forgot the add the oidc-client-id option, so with both required options it does start the API server with the two API options.

Im using multiple pods, so my install uses a config yaml, so this works:

nodes:
- role: control-plane
  kubeadmConfigPatches:
  - |
    kind: ClusterConfiguration
    apiServer:
      extraArgs:
        oidc-issuer-url: "https://xxxx.xxx.xxx"
        oidc-client-id: "kubernetes"
- role: worker

Thanks again for the quick reply - I really appreciate that.

BenTheElder:
awesome!

yeah, this sort of issue can be frustrating to debug, what you can do is use `kind export logs` while the node is still running (`--retain` will avoid kind deleting it on failure) or yourself look for the pod log in the node. the API server logs would surface the flag issue.

robotica72:
Excellent - Thanks for that note, The API logs would have helped, without a doubt :)

PR #1404, opened 2020-03-13
Add OIDC configuration to docs

int128:
This will add the OpenID Connect configuration to the document. It helps testing Kubernetes OIDC authentication. All example files are based on [the acceptance test of kubelogin](https://github.com/int128/kubelogin/tree/master/acceptance_test).

### DIFF
--- a/site/content/docs/user/configuration.md
+++ b/site/content/docs/user/configuration.md
@@ -177,6 +177,66 @@ nodes:
   image: kindest/node:v1.16.4@sha256:b91a2c2317a000f3a783489dfb755064177dbc3a0b2f4147d50f04825d016f55
 {{< /codeFromInline >}}
 
+### Authentication
+
+#### OpenID Connect
+
+You can set up OpenID Connect (OIDC) authentication by the following steps:
+
+1. Deploy an OIDC provider outside of the cluster.
+1. Create a cluster with the extra arguments for OIDC.
+1. Set up kubectl, e.g. [kubelogin](https://github.com/int128/kubelogin).
+
+You need to consider the following constraints:
+
+- The issuer URL must be HTTPS. A TLS certificate is required.
+- You cannot deploy an OIDC provider in the cluster,
+  because the API server cannot access any pod or service.
+- The API server and kubectl access an OIDC provider via the same URL.
+
+Here is an example of cluster config.
+
+{{< codeFromFile file="static/examples/config-with-oidc.yaml" lang="yaml" >}}
+
+You can use [Dex](https://github.com/dexidp/dex) as an OIDC provider.
+Here is a minimum config of Dex.
+
+{{< codeFromFile file="static/examples/oidc/dex.yaml" lang="yaml" >}}
+
+You can set up the authentication using Dex on Docker as follows:
+
+```sh
+# run a container of Dex
+docker run -d --name dex-server -p 10443:10443 -v $PWD:/cfg quay.io/dexidp/dex:v2.21.0 serve /cfg/dex.yaml
+
+# create a cluster
+kind create cluster --config config-with-oidc.yaml
+
+# set up the hosts so that the API server can access Dex
+docker inspect -f '{{.NetworkSettings.IPAddress}}' dex-server | sed -e 's,$, dex-server,' | \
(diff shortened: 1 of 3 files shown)


aojea:
This seems a very specific configuration and we discussed previously that growing the main page with configurations option will make it more difficult to read.
I defer to @BenTheElder the last decision but I prefer to add this in its own section, same we did with local-registry, ingress, ...

site/content/docs/user/configuration.md
```diff
@@ -177,6 +177,66 @@ nodes:
...
+# set up the kubeconfig (using kubelogin)
+kubectl config set-credentials oidc --exec-api-version=client.authentication.k8s.io/v1beta1 \
+  --exec-command=kubectl \
+  --exec-arg=oidc-login \
+  --exec-arg=get-token \
+  --exec-arg=--oidc-issuer-url=https://dex-server:10443/dex \
+  --exec-arg=--oidc-client-id=YOUR_CLIENT_ID \
+  --exec-arg=--oidc-client-secret=YOUR_CLIENT_SECRET \
+  --exec-arg=--oidc-extra-scope=email \
+  --exec-arg=--certificate-authority=$PWD/dex-ca.crt
```
neolit123: what step would write the `dex-ca.crt`?

neolit123:
curious to find more about the use case here. @int128 can you explain how this can be useful to kind users?

i can see the creation of kind clusters with OIDC to be able to dry-run the creation of real cluster with OIDC. at least one user wanted this [here](https://github.com/kubernetes-sigs/kind/issues/1291), but not clear why.

site/content/docs/user/configuration.md
```diff
@@ -177,6 +177,66 @@ nodes:
...
 
+### Authentication
+
+#### OpenID Connect
+
+You can set up OpenID Connect (OIDC) authentication by the following steps:
+
+1. Deploy an OIDC provider outside of the cluster.
+1. Create a cluster with the extra arguments for OIDC.
+1. Set up kubectl, e.g. [kubelogin](https://github.com/int128/kubelogin).
```
neolit123: this seems like a recommendation to use your own plugin, which is fine as long as we can justify the demand for this documentation.

int128:
This doc may help tool developers to test Kubernetes with real OIDC providers. Actually I test my plugin with various providers (e.g. testing the group based access control). I was struggled to set up Kind with OIDC because there was no doc about OIDC.

site/content/docs/user/configuration.md
int128: It's a long script. I will add it as a separated file in `examples` folder.

site/content/docs/user/configuration.md
int128: It is not essential here and can be removed.

site/content/docs/user/configuration.md
```diff
@@ -177,6 +177,66 @@ nodes:
   image: kindest/node:v1.16.4@sha256:b91a2c2317a000f3a783489dfb755064177dbc3a0b2f4147d50f04825d016f55
 {{< /codeFromInline >}}
 
+### Authentication
```
BenTheElder: this page is meant to document configuration fields in the config file.

int128:
Thank you for your comments.
I think the OIDC section is not needed in Kind docs at this time and close this.

int128:
FYI, I wrote this topic in https://github.com/int128/kind-oidc

spkane:
So, as a kind user, I did find the example in here useful once I tracked it down. I suspected that something like this was possible, but was having trouble figuring out exactly how to do it. The example IS very specific, and could likely be simplified somewhat, I my case I only need a kind config that looked something like this:

```
kind: Cluster
kubeadmConfigPatches:
  - |
    kind: ClusterConfiguration
    metadata:
      name: config
    apiServer:
      extraArgs:
        oidc-issuer-url: 'https://example.okta.com'
        oidc-client-id: '8idycy89sdnfydsy89yj'
        oidc-username-claim: 'email'
        oidc-username-prefix: 'okta:'
        oidc-groups-claim: 'groups'
        oidc-groups-prefix: 'okta:'
```

PR #1405, opened 2020-03-13
[WIP][TEST]: use etcd in memory

aojea:
kind mounts /tmp in memory

```
		// runtime temporary storage
```
use that to provide the etcd storage and avoid writing to disk

### DIFF
--- a/hack/ci/e2e-k8s.sh
+++ b/hack/ci/e2e-k8s.sh
@@ -131,6 +131,9 @@ kubeadmConfigPatches:
   kind: ClusterConfiguration
   metadata:
     name: config
+  etcd:
+    local:
+      dataDir: "/tmp/lib/etcd"
   apiServer:
     extraArgs:
       "runtime-config": "${runtime_config}"


aojea:
ping @BenTheElder

aojea:
1.17 19m13s
1.16 17m51s
1.15 17m10s
ipv6 15m11s
/retest all

aojea:
1.17 18m41s
1.16 30m8s
1.15 15m45s
ipv6 15m31s
/test all

aojea:
1.17 16m6s
1.16 16m7s
1.15 16m40s
ipv6 19m34s

BenTheElder:
> https://github.com/kubernetes-sigs/kind/pull/1405#issuecomment-598728915

FYI I am already github "watching" the repo and go through issues / PRs multiple times a day normally when I have the time :-)

BenTheElder:
going to need those numbers to be compared to without etcd.

I also think we should get some locally, CI is on very noisy shared hosts.

I continue to be concerned about recommending this to users OOTB, memory based storage will be wiped on reboot, this is in conflict with fixing kind to otherwise behave correctly there.

aojea:
/close

I don't think this is a priority right now, we'll retake later

PR #1409, opened 2020-03-16
BUG: fix finding the provider when links are present

slimm609:
when running an ingress on KIND, like the one documented here: https://banzaicloud.com/blog/kind-ingress/

when you run
```
docker run -d --name banzai-kind-proxy-${port} \
      --publish 127.0.0.1:${port}:${port} \
      --link banzai-control-plane:target \
      alpine/socat -dd \
      tcp-listen:${port},fork,reuseaddr tcp-connect:target:${node_port}
```
it adds a link to the control-plane.     

Once this control-plane link exists, you can no longer get information from kind for nodes and can not load images. 
```
kind load docker-image ubuntu --name kind
ERROR: failed to get role for node: command "docker inspect --format '{{ index .Config.Labels "io.x-k8s.kind.role"}}' ingress-proxy-443/target,ingress-proxy-80/target,kind-control-plane" failed with error: exit status 1
```

This is caused by the --no-trunc option which in the docker ps command because it includes linked containers.  Removing the no-trunc option resolves the issue.    

I have tested several different setups to confirm resolution and couldn't find why the --no-trunc was added initially.

### DIFF
--- a/pkg/cluster/internal/providers/docker/provider.go
+++ b/pkg/cluster/internal/providers/docker/provider.go
@@ -77,7 +77,7 @@ func (p *Provider) Provision(status *cli.Status, cluster string, cfg *config.Clu
 func (p *Provider) ListClusters() ([]string, error) {
 	cmd := exec.Command("docker",
 		"ps",
-		"-a",         // show stopped nodes
+		"-a", // show stopped nodes
 		// filter for nodes with the cluster label
 		"--filter", "label="+clusterLabelKey,
 		// format to include the cluster name
@@ -94,7 +94,7 @@ func (p *Provider) ListClusters() ([]string, error) {
 func (p *Provider) ListNodes(cluster string) ([]nodes.Node, error) {
 	cmd := exec.Command("docker",
 		"ps",
-		"-a",         // show stopped nodes
+		"-a", // show stopped nodes
 		// filter for nodes with the cluster label
 		"--filter", fmt.Sprintf("label=%s=%s", clusterLabelKey, cluster),
 		// format to include the cluster name
--- a/pkg/cluster/internal/providers/docker/provider.go
+++ b/pkg/cluster/internal/providers/docker/provider.go
@@ -78,7 +78,6 @@ func (p *Provider) ListClusters() ([]string, error) {
 	cmd := exec.Command("docker",
 		"ps",
 		"-a",         // show stopped nodes
-		"--no-trunc", // don't truncate
 		// filter for nodes with the cluster label
 		"--filter", "label="+clusterLabelKey,
 		// format to include the cluster name
@@ -96,7 +95,6 @@ func (p *Provider) ListNodes(cluster string) ([]nodes.Node, error) {
 	cmd := exec.Command("docker",
 		"ps",
 		"-a",         // show stopped nodes
-		"--no-trunc", // don't truncate
 		// filter for nodes with the cluster label
 		"--filter", fmt.Sprintf("label=%s=%s", clusterLabelKey, cluster),
 		// format to include the cluster name
--- a/pkg/cluster/internal/providers/podman/provider.go
+++ b/pkg/cluster/internal/providers/podman/provider.go
@@ -90,7 +90,7 @@ func (p *Provider) Provision(status *cli.Status, cluster string, cfg *config.Clu
 func (p *Provider) ListClusters() ([]string, error) {
 	cmd := exec.Command("podman",
 		"ps",
-		"-a",         // show stopped nodes
+		"-a", // show stopped nodes
 		// filter for nodes with the cluster label
 		"--filter", "label="+clusterLabelKey,
 		// format to include the cluster name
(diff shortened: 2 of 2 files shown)


neolit123:
we had a discussion at some point that `--link` and linked containers is not really recommended.
next to removing linked containers, are there any side-effects of removing `--no-trunc`?

slimm609:
> we had a discussion at some point that `--link` and linked containers is not really recommended (cannot find the ticket).
> next to removing linked containers, are there any side-effects of removing `--no-trunc`?

not that I could find or that I was able to see during testing.   I also looked at the commit history for that section to code to try to find if the `--no-trunc` was added in response to a issue and i couldn't find one.  (hoping that it would lead me to a reason for it being there)

skymeyer:
@slimm609 @neolit123 I looked into a bit closer, and I suggest to use `{{.ID}}` instead of `{{.Names}}` and leave the `--no-trunc` in place.

As we can reference containers by `id` or `name` this should have no further unexpected fallout.

See https://github.com/docker/cli/blob/master/cli/command/formatter/container.go#L115-L138

pkg/cluster/internal/providers/docker/provider.go
```diff
@@ -100,7 +100,7 @@ func (p *Provider) ListNodes(cluster string) ([]nodes.Node, error) {
 		// filter for nodes with the cluster label
 		"--filter", fmt.Sprintf("label=%s=%s", clusterLabelKey, cluster),
 		// format to include the cluster name
-		"--format", `{{.Names}}`,
+		"--format", `{{.ID}}`,
```
BenTheElder: other things use the cluster names, this is a breaking API change. please look at the node API

BenTheElder:
link is deprecated upstream in docker but we'd be willing to accept a reasonable fix, using ID is not acceptable here at the moment, things key of the node names currently.

I've commented on the bonzai cloud page, we do not recommend using those instructions, instead please see https://kind.sigs.k8s.io/docs/user/ingress/ for a different approach.

BenTheElder:
I think we don't need the `--no-trunc` here per the original discussion, when I get a chance later I'll try to confirm, I'm assuming that container names won't actually be truncated but I'd like evidence to verify 😅

pkg/cluster/internal/providers/docker/provider.go
skymeyer: That comment doesn't seem right to me

BenTheElder: see below where we use the output to create a `Node` with a string value, this value is used to implement the `Name()` method which other things key off of.

skymeyer:
> ... I'm assuming that container names won't actually be truncated

Probably some corner cases.

https://github.com/docker/cli/blob/master/vendor/github.com/docker/docker/pkg/stringid/stringid.go#L25-L37

> using ID is not acceptable here at the moment, things key of the node names currently.

If this is merely for naming convention/derivation which is used to create the nodes, this should be more clear in the API. Regardless, any operations on each container should rely on labels instead which seems to be the case, not how those containers are named.

This all seems the case using `io.x-k8s.kind.cluster` and `io.x-k8s.kind.role` labels which are used for filtering. The container name should be irrelevant to the `provider` abstract.

Not sure this should be addressed on this issue, but that'd create some better structure how `kind` handles things on the `provider` level.

BenTheElder:
> If this is merely for naming convention/derivation which is used to create the nodes, this should be more clear in the API.

This is in the node interface. https://github.com/kubernetes-sigs/kind/blob/2e0a0385bb5a0f11bd7d2c5dba8b749e2a5018a6/pkg/cluster/nodes/types.go#L30

> Regardless, any operations on each container should rely on labels instead which seems to be the case, not how those containers are named.

They don't all do that, and there are reasons they can't currently.

---

Again docker is very clear that they do not want users using the "legacy container link" feature feature upstream anyhow with bold red text:
Screen Shot 2020-03-16 at 5 53 21 PM

https://docs.docker.com/network/links/

BenTheElder:
btw: I'm about to land using user defined networks in kind today or tomorrow ideally, I finished making it work (harder than it sounds...) over the weekend but I'm currently cleaning it up and breaking it into manageable PRs.

skymeyer:
Sure, no problem. I get the `link` deprecation which is hanging there for a while now, but it's only indirectly related though. Not sure what the effect is of dropping `--no-trunc` either ...

We'll take another ingress approach and see checkout your user def netw goodies once available. 

Thx @BenTheElder - stay healthy ;-)

BenTheElder:
you too!

broadly docker `--format` is not well documented but I bet we can inspect the code if nothing else 😅

BenTheElder:
AFAICT we can just drop this flag, opened https://github.com/kubernetes-sigs/kind/pull/1420 since it's trivial but if this PR gets updated before then I'll of course close that one since IIRC the change there is what this PR originally did..

BenTheElder:
holding https://github.com/kubernetes-sigs/kind/pull/1420#issuecomment-600453822 a bit so this one can be updated instead, thanks for finding this @slimm609 👍

slimm609:
Sorry about that.    Updated to reflect the original changes.

aojea:
/lgtm
/ok-to-test
/assign @BenTheElder 
nice catch, thanks

BenTheElder:
/lgtm
/approve
thank you!

PR #1414, opened 2020-03-17
embedded dns compatibility

BenTheElder:
docker doesn't really document the full details but I've done a *ton* of looking into this.

with this change the kind image can handle being on a non default docker network.

this is once again broken out and cleaned up from my prototype for https://github.com/kubernetes-sigs/kind/issues/148

### DIFF
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -154,6 +154,33 @@ select_iptables() {
   update-alternatives --set ip6tables "/usr/sbin/ip6tables-${mode}" > /dev/null
 }
 
+enable_network_magic(){
+  # well-known docker embedded DNS is at 127.0.0.11:53
+  local docker_embedded_dns_ip='127.0.0.11'
+
+  # first we need to detect an IP to use for reaching the docker host
+  local docker_host_ip
+  docker_host_ip="$( (getent ahostsv4 'host.docker.internal' | head -n1 | cut -d' ' -f1) || true)"
+  if [[ -z "${docker_host_ip}" ]]; then
+    docker_host_ip=$(ip -4 route show default | cut -d' ' -f3)
+  fi
+
+  # patch docker's iptables rules to switch out the DNS IP
+  iptables-save \
+    | sed \
+      `# switch docker DNS DNAT rules to our chosen IP` \
+      -e "s/-d ${docker_embedded_dns_ip}/-d ${docker_host_ip}/g" \
+      `# we need to also apply these rules to non-local traffic (from pods)` \
+      -e 's/-A OUTPUT \(.*\) -j DOCKER_OUTPUT/\0\n-A PREROUTING \1 -j DOCKER_OUTPUT/' \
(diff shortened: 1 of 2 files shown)


BenTheElder:
/retest
test flake

images/base/files/usr/local/bin/entrypoint
```diff
@@ -154,6 +154,33 @@ select_iptables() {
...
+      -e "s/-d ${docker_embedded_dns_ip}/-d ${docker_host_ip}/g" \
+      `# we need to also apply these rules to non-local traffic (from pods)` \
+      -e 's/-A OUTPUT \(.*\) -j DOCKER_OUTPUT/\0\n-A PREROUTING \1 -j DOCKER_OUTPUT/' \
+      `# switch docker DNS SNAT rules rules to our chosen IP` \
+      -e "s/--to-source :53/--to-source ${docker_host_ip}:53/g"\
+    | iptables-restore
+
+  # now we can ensure that DNS is configured to use our IP
+  cp /etc/resolv.conf /etc/resolv.conf.original
+  sed -e "s/${docker_embedded_dns_ip}/${docker_host_ip}/g" /etc/resolv.conf.original >/etc/resolv.conf
```
amwat: sed in place instead? `-i`

images/base/files/usr/local/bin/entrypoint
```diff
@@ -154,6 +154,33 @@ select_iptables() {
   update-alternatives --set ip6tables "/usr/sbin/ip6tables-${mode}" > /dev/null
 }
 
+enable_network_magic(){
```
amwat: maybe a more informative name :P (and move magic to a comment)
`replace_docker_dns` ?

BenTheElder:
It's magic ✨

BenTheElder:
Also it does not replace it 🙃

amwat:
seems legit
/lgtm

images/base/files/usr/local/bin/entrypoint
aojea: This implies that ${docker_host_ip} has a DNS resolver working, otherwise it is not going to be able to resolve DNS.
By default, docker copies the resolv.conf from the host

images/base/files/usr/local/bin/entrypoint
```diff
@@ -154,6 +154,33 @@ select_iptables() {
...
+  # well-known docker embedded DNS is at 127.0.0.11:53
+  local docker_embedded_dns_ip='127.0.0.11'
+
+  # first we need to detect an IP to use for reaching the docker host
+  local docker_host_ip
+  docker_host_ip="$( (getent ahostsv4 'host.docker.internal' | head -n1 | cut -d' ' -f1) || true)"
```
aojea: this is not resolving to me in linux

images/base/files/usr/local/bin/entrypoint
BenTheElder: No it does not. Because it is a replace :-)

BenTheElder: You can't.

PR #1418, opened 2020-03-18
Adding UT for port validate

longkb:
Signed-off-by: Kim Bao Long <longkb@vn.fujitsu.com>

### DIFF
--- a/pkg/internal/apis/config/validate_test.go
+++ b/pkg/internal/apis/config/validate_test.go
@@ -243,3 +243,49 @@ func TestNodeValidate(t *testing.T) {
 		})
 	}
 }
+
+func TestPortValidate(t *testing.T) {
+	cases := []struct {
+		TestName    string
+		Port        int32
+		ExpectError string
+	}{
+		{
+			TestName:    "-1 port",
+			Port:        -1,
+			ExpectError: "",
+		},
+		{
+			TestName:    "valid port",
+			Port:        10,
+			ExpectError: "",
+		},
+		{
+			TestName:    "negative port",
+			Port:        -2,
+			ExpectError: "invalid port number: -2",
+		},
+		{
+			TestName:    "extra port",
+			Port:        65536,
+			ExpectError: "invalid port number: 65536",
+		},
+	}
+
+	for _, tc := range cases {
+		tc := tc //capture loop variable
+		t.Run(tc.TestName, func(t *testing.T) {
+			t.Parallel()
+			err := validatePort(tc.Port)
+			// the error can be:
+			// - nil, in which case we should expect no errors or fail
+			if err == nil && len(tc.ExpectError) > 0 {
+				t.Errorf("Test failed, unexpected error: %s", tc.ExpectError)
+			}
+
+			if err != nil && err.Error() != tc.ExpectError {
+				t.Errorf("Test failed, error: %s expected error: %s", err, tc.ExpectError)
+			}
+		})
+	}
+}


pkg/internal/apis/config/validate_test.go
```diff
@@ -243,3 +244,48 @@ func TestNodeValidate(t *testing.T) {
...
+
+	for _, tc := range cases {
+		tc := tc //capture loop variable
+		t.Run(tc.TestName, func(t *testing.T) {
+			err := validatePort(tc.Port)
+			// the error can be:
+			// - nil, in which case we should expect no errors or fail
+			if err == nil && tc.ExpectErrors != nil {
+				t.Error("received no errors but expected errors for case")
+			}else if !reflect.DeepEqual(err, tc.ExpectErrors) {
```
BenTheElder: nit: should be `gofmt`ed (which would be `} else if ...`)

BenTheElder: `hack/update/gofmt.sh` should fix this

longkb: Done. Thanks :)

aojea:
the test is panicking, @łongkb I think that we can simplify the test with this approach and avoid using reflect:

```
func TestPortValidate(t *testing.T) {
	cases := []struct {
		TestName    string
		Port        int32
		ExpectError string
	}{
		{
			TestName:    "-1 port",
			Port:        -1,
			ExpectError: "",
		},
		{
			TestName:    "valid port",
			Port:        10,
			ExpectError: "",
		},
		{
			TestName:    "negative port",
			Port:        -2,
			ExpectError: "invalid port number: -2",
		},
		{
			TestName:    "extra port",
			Port:        65536,
			ExpectError: "invalid port number: 65536",
		},
	}

	for _, tc := range cases {
		tc := tc //capture loop variable
		t.Run(tc.TestName, func(t *testing.T) {
			err := validatePort(tc.Port)
			// the error can be:
			// - nil, in which case we should expect no errors or fail
			if err == nil && len(tc.ExpectError) > 0 {
				t.Errorf("Test failed, unexpected error: %v %s", err, tc.ExpectError)
			}

			if err != nil && err.Error() != tc.ExpectError {
				t.Errorf("Test failed, error: %s expected error: %s", err, tc.ExpectError)
			}
		})
	}
}
```

longkb:
> 
> 
> the test is panicking, @longkb I think that we can simplify the test with this approach and avoid using reflect:
> 
> ```go
> func TestPortValidate(t *testing.T) {
> 	cases := []struct {
> 		TestName    string
> 		Port        int32
> 		ExpectError string
> 	}{
> 		{
> 			TestName:    "-1 port",
> 			Port:        -1,
> 			ExpectError: "",
> 		},
> 		{
> 			TestName:    "valid port",
> 			Port:        10,
> 			ExpectError: "",
> 		},
> 		{
> 			TestName:    "negative port",
> 			Port:        -2,
> 			ExpectError: "invalid port number: -2",
> 		},
> 		{
> 			TestName:    "extra port",
> 			Port:        65536,
> 			ExpectError: "invalid port number: 65536",
> 		},
> 	}
> 
> 	for _, tc := range cases {
> 		tc := tc //capture loop variable
> 		t.Run(tc.TestName, func(t *testing.T) {
> 			err := validatePort(tc.Port)
> 			// the error can be:
> 			// - nil, in which case we should expect no errors or fail
> 			if err == nil && len(tc.ExpectError) > 0 {
> 				t.Errorf("Test failed, unexpected error: %s", tc.ExpectError)
> 			}
> 
> 			if err != nil && err.Error() != tc.ExpectError {
> 				t.Errorf("Test failed, error: %s expected error: %s", err, tc.ExpectError)
> 			}
> 		})
> 	}
> }
> ```

Thanks @aojea . It passed the UT with your great patch :)

aojea:
/retest
/lgtm
/assign @BenTheElder 
Thanks

pkg/internal/apis/config/validate_test.go
```diff
@@ -243,3 +243,49 @@ func TestNodeValidate(t *testing.T) {
...
+		},
+	}
+
+	for _, tc := range cases {
+		tc := tc //capture loop variable
+		t.Run(tc.TestName, func(t *testing.T) {
+			t.Parallel()
+			err := validatePort(tc.Port)
+			// the error can be:
+			// - nil, in which case we should expect no errors or fail
```
BenTheElder: nit: this comment seems to cover the cases on the surface, but doesn't actually list them ...?

BenTheElder:
/hold
holding for a moment to see if we can handle the one comment nit :-)

PR #1420, opened 2020-03-18
drop --no-trunc

BenTheElder:
fixes https://github.com/kubernetes-sigs/kind/pull/1409

I did a bunch of testing, I don't think we ever needed this ... the longest names that you can even make a working kind container with (before setting hostname fails and breaks starting the container in docker...) are not truncated ...

we can just drop `--no-trunc`.

### DIFF
--- a/pkg/cluster/internal/providers/docker/provider.go
+++ b/pkg/cluster/internal/providers/docker/provider.go
@@ -77,8 +77,7 @@ func (p *Provider) Provision(status *cli.Status, cluster string, cfg *config.Clu
 func (p *Provider) ListClusters() ([]string, error) {
 	cmd := exec.Command("docker",
 		"ps",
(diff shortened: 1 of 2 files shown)


BenTheElder:
/hold
I want to give https://github.com/kubernetes-sigs/kind/pull/1409 a chance to revert back to the original state (like this), kudos to the author for finding this.

BenTheElder:
Lacking a response, I'm going to actually just going to modify this commit to include the original author as evidenced by this force push data: https://github.com/kubernetes-sigs/kind/compare/b5a7353d7987dc80b8228048daafc45bd4541a49..35b0870315681259d9cb47513c4c031766fc6efc from https://github.com/kubernetes-sigs/kind/pull/1409

BenTheElder:
we did get the original PR updated.

Issue #1326, opened 2020-02-10
CNI not working properly in ipv6 mode


amwat:
**What happened**:
ref: https://github.com/kubernetes-sigs/kind/issues/1292#issue-555947266

**What you expected to happen**:
Ingress should work properly in ipv6.
**How to reproduce it (as minimally and precisely as possible)**:
Following the ingress-nginx guide on https://kind.sigs.k8s.io/docs/user/ingress/
with the addition of enabling ipv6
```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
networking:
  ipFamily: ipv6
nodes:
- role: control-plane
  kubeadmConfigPatches:
  - |
    kind: InitConfiguration
    nodeRegistration:
      kubeletExtraArgs:
        node-labels: "ingress-ready=true"
        authorization-mode: "AlwaysAllow"
  extraPortMappings:
  - containerPort: 80
    hostPort: 80
    protocol: TCP
  - containerPort: 443
    hostPort: 443
    protocol: TCP
```
**Anything else we need to know?**:
Note that the docker port forwarding seems to work correctly. 
example: just simply running nginx directly on the kind node is accessible from the host.
**Environment:**

- kind version: (use `kind version`): kind v0.8.0-alpha+7b4c4b691a7855 go1.13.7 linux/amd64
Kubernetes version: (use kubectl version): v1.17.0
Docker version: (use docker info): Docker version 19.03.5
OS (e.g. from /etc/os-release): Debian GNU/Linux

BenTheElder:
FYI @aledbf -- pretty sure this is a KIND bug :+)

aledbf:
I tested the configuration and cannot reach ingress-nginx using `::1` but it works if I used the IPV6 address assigned to the container.

aojea:
it seems a bug in the docker proxy, that tries to use IPv4 on the translation when using `::1`as destination.

```
curl -v -6 -I ::1 localhost/foo
* Closing connection -1
curl: (3) URL using bad/illegal format or missing URL
*   Trying ::1:80...
* TCP_NODELAY set
* Connected to localhost (::1) port 80 (#0)
> HEAD /foo HTTP/1.1
> Host: localhost
> User-Agent: curl/7.68.0
> Accept: */*
> 
* Recv failure: Connection reset by peer
* Closing connection 0
curl: (56) Recv failure: Connection reset by peer
```

However, using the container IPv6 address it works as @aledbf says:

```
[...]

PR #1421, opened 2020-03-18
IPv6 docker nat Known issue

aojea:
Docker doesn't implement IPv6 NAT for port forwarding and is not likely to do it.

Fixes: #1326


site/content/docs/user/known-issues.md
```diff
@@ -347,6 +348,15 @@ related to applications you are trying to run in KIND.
 
 See Previous Discussion: [kind#1179]
 
+## IPv6 Port Forwarding
+
+Docker consider that all the IPv6 addresses should be reachable, hence doesn't implement
```
BenTheElder: S/consider/assumes

PR #1430, opened 2020-03-21
Fix hack/update/deps.sh for kindnetd code move

gurjeet:


aojea:
/ok-to-test
/lgtm
/assign @BenTheElder 
Thanks

PR #1431, opened 2020-03-21
Updated contributing doc to correct script names

gurjeet:
Mention the correct verify/* and update/* script names in the
contributing docs to prevent confusion for the new developers.

In passing, surrounded some command names with back-ticks.


aojea:
I think that some of these docs are outdated, i.e., the generated bindata is no longer needed IIRC, but the PR is legit, the scripts moved to folders a while ago 😄 
/lgtm
/assign @BenTheElder

BenTheElder:
/approve
yeah this doc should be rewritten from scratch, it's on my backlog.
it's well intention-ed but overly specific and likely to be eternally stale.

PR #1438, opened 2020-03-24
add unit test for pkg/fs/fs.go

Hellcatlk:
Signed-off-by: Zouyu <zouy.fnst@cn.fujitsu.com>

### DIFF
--- a/pkg/fs/fs_test.go
+++ b/pkg/fs/fs_test.go
@@ -1,0 +1,241 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package fs
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+)
+
+func TestTempDir(t *testing.T) {
+	tempDir := os.TempDir()
+	tests := []struct {
+		name          string
+		dir           string
+		prefix        string
+		expectedError bool
+	}{
+		{
+			name:          "all is right",
+			dir:           tempDir,
+			prefix:        "tmp",
+			expectedError: false,
+		},
+		{
+			name:          "dir is empty",
+			dir:           "",
+			prefix:        "tmp",
+			expectedError: false,
+		},
+		{
+			name:          "dir doesn't exist",
+			dir:           "/abc",
+			prefix:        "tmp",
+			expectedError: true,
+		},
+		{
+			name:          "prefix is empty",
+			dir:           tempDir,
+			prefix:        "",
+			expectedError: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			name, err := TempDir(tt.dir, tt.prefix)
+			os.Remove(name)
+			if (nil != err) != tt.expectedError {
+				t.Errorf("expectedError: %v, got: %v", tt.expectedError, err)
+			}
+		})
+	}
+}
+
+func TestIsAbs(t *testing.T) {
+	tests := []struct {
+		name     string
+		hostPath string
+		expected bool
+	}{
+		{
+			name:     "is absolute path",
+			hostPath: "/home/test",
+			expected: true,
+		},
+		{
+			name:     "is relative path",
+			hostPath: "./test",
+			expected: false,
+		},
+		{
+			name:     "is a home path",
+			hostPath: "~/test",
+			expected: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ret := IsAbs(tt.hostPath)
+			if ret != tt.expected {
+				t.Errorf("expected: %v, got: %v", tt.expected, ret)
+			}
+		})
+	}
+}
+
+func TestCopy(t *testing.T) {
+	testDir, err := ioutil.TempDir(os.TempDir(), "")
+	os.Mkdir(testDir+"/td1", 0777)
+	if nil != err {
+		return
+	}
+	defer os.Remove(testDir)
+
+	tests := []struct {
+		name          string
+		src           string
+		dst           string
+		expectedError bool
+	}{
+		{
+			name:          "src is empty",
+			src:           "",
+			dst:           testDir + "/test.txt",
+			expectedError: true,
+		},
+		{
+			name:          "src doesn't exist",
+			src:           testDir + "/abc",
+			dst:           testDir + "/test.txt",
+			expectedError: true,
+		},
+		// src is dir
+		{
+			name:          "all is right",
+			src:           testDir + "/td1",
+			dst:           testDir + "/test",
+			expectedError: false,
+		},
+		{
+			name:          "dst is empty",
+			src:           testDir + "/td1",
+			dst:           "",
+			expectedError: true,
+		},
+		// src is file
+		{
+			name:          "all is right",
+			src:           "fs_test.go",
+			dst:           testDir + "/test.txt",
+			expectedError: false,
+		},
+		{
+			name:          "dst is dir",
+			src:           "fs_test.go",
+			dst:           testDir + "/td1",
+			expectedError: true,
+		},
+		{
+			name:          "dst is empty",
+			src:           "fs_test.go",
+			dst:           "",
+			expectedError: true,
+		},
+		{
+			name:          "dst is exist",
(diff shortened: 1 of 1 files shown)


pkg/fs/fs_test.go
```diff
@@ -0,0 +1,229 @@
...
+	"os"
+	"testing"
+)
+
+func TestTempDir(t *testing.T) {
+	tests := []struct {
+		name   string
+		dir    string
+		prefix string
+		expect error
```
neolit123: in some cases it is better to have a `expectedError bool` instead of matching the error.

pkg/fs/fs_test.go
```diff
@@ -0,0 +1,229 @@
...
+			expect: nil,
+		},
+		{
+			name:   "dir is nil",
+			dir:    "",
+			prefix: "tmp",
+			expect: nil,
+		},
+		{
+			name:   "dir isn't exist",
```
neolit123: isn't -> doesn't

pkg/fs/fs_test.go
```diff
@@ -0,0 +1,229 @@
...
+		expect error
+	}{
+		{
+			name:   "all is right",
+			dir:    ".",
+			prefix: "tmp",
+			expect: nil,
+		},
+		{
+			name:   "dir is nil",
```
neolit123: `dir` is not really `nil` it is an empty string.

pkg/fs/fs_test.go
```diff
@@ -0,0 +1,229 @@
...
+			expect: nil,
+		},
+		{
+			name:   "dir isn't exist",
+			dir:    "/abc",
+			prefix: "tmp",
+			expect: errors.New("isn't nil"),
+		},
+		{
+			name:   "prefix is nil",
```
neolit123: `prefix` is not really `nil` it is an empty string.

pkg/fs/fs_test.go
```diff
@@ -0,0 +1,229 @@
...
+			dir:    ".",
+			prefix: "",
+			expect: nil,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			name, err := TempDir(tt.dir, tt.prefix)
+			os.Remove(name)
+			if (nil == err) != (nil == tt.expect) {
```
neolit123:
```
if (err != nil) != tt.expectError {
...
```

pkg/fs/fs_test.go
```diff
@@ -0,0 +1,229 @@
...
+		hostPath string
+		expect   bool
+	}{
+		{
+			name:     "is absolute path",
+			hostPath: "/home/test",
+			expect:   true,
+		},
+		{
+			name:     "isn't absolute path",
```
neolit123: `is relative path`

pkg/fs/fs_test.go
```diff
@@ -0,0 +1,229 @@
...
+			hostPath: "./test",
+			expect:   false,
+		},
+		{
+			name:     "isn't absolute path(~)",
```
neolit123: `is a home path`

pkg/fs/fs_test.go
```diff
@@ -0,0 +1,229 @@
...
+			}
+		})
+	}
+}
+
+func TestIsAbs(t *testing.T) {
+	tests := []struct {
+		name     string
+		hostPath string
+		expect   bool
```
neolit123: expected

pkg/fs/fs_test.go
```diff
@@ -0,0 +1,229 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			ret := IsAbs(tt.hostPath)
+			if ret != tt.expect {
+				t.Errorf("expect: %v, got: %v", tt.expect, ret)
+			}
+		})
+	}
+}
+
+func TestCopy(t *testing.T) {
```
neolit123: see the comments for TestTempDir above.

pkg/fs/fs_test.go
```diff
@@ -0,0 +1,229 @@
...
+	defer os.Remove("./test.txt")
+	defer os.RemoveAll("./test")
```
neolit123: tests should not use relative paths, but instead write to https://golang.org/pkg/io/ioutil/#TempDir

pkg/fs/fs_test.go
```diff
@@ -0,0 +1,229 @@
...
+			err := Copy(tt.src, tt.dst)
+			if (nil == err) != (nil == tt.expect) {
+				t.Errorf("expect: %v, got: %v", tt.expect, err)
+			}
+		})
+	}
+}
+
+func TestCopyFile(t *testing.T) {
+	defer os.Remove("./test.txt")
```
neolit123: [1]

pkg/fs/fs_test.go
```diff
@@ -0,0 +1,229 @@
...
+	tests := []struct {
+		name   string
+		src    string
+		dst    string
+		expect error
```
neolit123: similar comments as TestCopy and TestTempDir

pkg/fs/fs_test.go
```diff
@@ -0,0 +1,229 @@
...
+			prefix: "",
+			expect: nil,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			name, err := TempDir(tt.dir, tt.prefix)
+			os.Remove(name)
+			if (nil == err) != (nil == tt.expect) {
+				t.Errorf("expect: %v, got: %v", tt.expect, err)
```
neolit123: `expected:...`

pkg/fs/fs_test.go
```diff
@@ -0,0 +1,229 @@
...
+			name:     "isn't absolute path(~)",
+			hostPath: "~/test",
+			expect:   false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ret := IsAbs(tt.hostPath)
+			if ret != tt.expect {
+				t.Errorf("expect: %v, got: %v", tt.expect, ret)
```
neolit123: `expected:...`

pkg/fs/fs_test.go
```diff
@@ -0,0 +1,229 @@
...
+			src:    "./fs.go",
+			dst:    "./test.txt",
+			expect: nil,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := Copy(tt.src, tt.dst)
+			if (nil == err) != (nil == tt.expect) {
+				t.Errorf("expect: %v, got: %v", tt.expect, err)
```
neolit123: `expected:...`

pkg/fs/fs_test.go
```diff
@@ -0,0 +1,229 @@
...
+			src:    "./fs.go",
+			dst:    "./test.txt",
+			expect: nil,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := CopyFile(tt.src, tt.dst)
+			if (nil == err) != (nil == tt.expect) {
+				t.Errorf("expect: %v, got: %v", tt.expect, err)
```
neolit123: `expected:...`

pkg/fs/fs_test.go
```diff
@@ -0,0 +1,229 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
tao12345666333: Can be updated to 2020 by the way :smile:

pkg/fs/fs_test.go
```diff
@@ -0,0 +1,229 @@
...
+		},
+		{
+			name:   "src is dir",
+			src:    ".",
+			dst:    "./test.txt",
+			expect: errors.New("isn't nil"),
+		},
+		{
+			name:   "src is nil",
+			src:    "",
```
tao12345666333: same as neolit123's comments above.

pkg/fs/fs_test.go
```diff
@@ -0,0 +1,241 @@
...
+
+package fs
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+)
+
+func TestTempDir(t *testing.T) {
```
BenTheElder: this doesn't seem very useful, this unit test is just testing `os.TempDir` ...

if we want to unit test this, we need to retool so we can actually test the behavior we're alterting

BenTheElder: we should not unit test the go standard library, they have their own unit tests

pkg/fs/fs_test.go
```diff
@@ -0,0 +1,241 @@
...
+			name, err := TempDir(tt.dir, tt.prefix)
+			os.Remove(name)
+			if (nil != err) != tt.expectedError {
+				t.Errorf("expectedError: %v, got: %v", tt.expectedError, err)
+			}
+		})
+	}
+}
+
+func TestIsAbs(t *testing.T) {
```
BenTheElder: again we only seem to be testing the standard library behavior instead of our special behavior

BenTheElder:
I do think this package should have tests, but only if they're actually testing the altered behavior we have. As far as I can tell this PR currently just tests the standard library behavior which is redundant to go itself.

PR #1446, opened 2020-03-27
return error when attemp to delete a non-existent cluster

zhijianli88:
the behaviors become:
$ ./bin/kind delete clusters aaaa bb cc
failed to delete cluster "aaaa": No such cluster
failed to delete cluster "bb": No such cluster
failed to delete cluster "cc": No such cluster
Deleted clusters: []
$ ./bin/kind delete cluster --name ccc
Deleting cluster "ccc" ...
ERROR: failed to delete cluster "ccc": No such cluster

Signed-off-by: Li Zhijian <lizhijian@cn.fujitsu.com>

### DIFF
--- a/pkg/cluster/internal/delete/delete.go
+++ b/pkg/cluster/internal/delete/delete.go
@@ -39,6 +39,9 @@ func Cluster(logger log.Logger, c *context.Context, explicitKubeconfigPath strin
 		logger.Errorf("failed to update kubeconfig: %v", kerr)
 	}
 
+	if len(n) == 0 {
+		return errors.New("No such cluster")
+	}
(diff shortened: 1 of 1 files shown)


aojea:
I vaguely remember a discussion about this, and @BenTheElder  commenting that this was the expected behaviour ... but I can be wrong 😅

BenTheElder:
This is working as intended.
Delete also cleans up kubeconfig entries and is idempotent.
This is not an error.

Thanks for the PR though!

zhijianli88:
> This is working as intended.
> Delete also cleans up kubeconfig entries and is idempotent.
> This is not an error.
> 
> Thanks for the PR though!

thanks for the explanation. :)

PR #1454, opened 2020-03-31
[Doc] update ingress annotation to proper nginx prefix

rajalokan:
- Update Ingress annotation to proper nginx prefix one

### DIFF
--- a/site/static/examples/ingress/usage.yaml
+++ b/site/static/examples/ingress/usage.yaml
@@ -51,7 +51,7 @@ kind: Ingress
 metadata:
   name: example-ingress
   annotations:
-    ingress.kubernetes.io/rewrite-target: /
+    nginx.ingress.kubernetes.io/rewrite-target: /
 spec:
   rules:
   - http:


site/static/examples/ingress/usage.yaml
```diff
@@ -51,7 +51,7 @@ kind: Ingress
 metadata:
   name: example-ingress
   annotations:
-    ingress.kubernetes.io/rewrite-target: /
+    nginx.ingress.kubernetes.io/rewrite-target: /$2
```
aledbf: You can't use $2 if you don't have groups defined in the path

rajalokan: So I should rather keep it `/`?

site/static/examples/ingress/usage.yaml
```diff
@@ -51,7 +51,7 @@ kind: Ingress
 metadata:
   name: example-ingress
   annotations:
-    ingress.kubernetes.io/rewrite-target: /
+    nginx.ingress.kubernetes.io/rewrite-target: /
```
BenTheElder, line 54: What versions is this compatible with?

site/static/examples/ingress/usage.yaml
```diff
@@ -64,4 +64,4 @@ spec:
         backend:
           serviceName: bar-service
           servicePort: 5678
----
\ No newline at end of file
+---
```
BenTheElder, line 67: This is intentional IIRC, because of now the snippets are currently rendered

site/static/examples/ingress/usage.yaml
amwat: we should also make sure we don't break contour.
although looking at https://github.com/projectcontour/contour/issues/64 i don't think they ever used that annotation.

basic experimentation with removing that annotation altogether also worked for me both on contour and nginx.
we should probably just drop it? to have a minimal example as much as possible.

rajalokan: I've tried this with kubernetes `1.17`. If you are looking for ingress controller version, it is `0.30.0`. Any specific version you want me to try this out ? 

I see the issue here. Let me try out contour as well. Probably we need to break this ingress spec into two parts. Nginx specific and Contour specific. 
```
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: example-ingress
  annotations:
    ingress.kubernetes.io/rewrite-target: /
spec:
  rules:
  - http:
      paths:
      - path: /foo
        backend:
          serviceName: foo-service
          servicePort: 5678
      - path: /bar
        backend:
          serviceName: bar-service
          servicePort: 5678
```

This documentation might seem trivial and easy to be resolved for someone who knows this. But I couldn't get pass it and had to ask in slack. So I feel this should be clear enough for anyone trying this out?

rajalokan: I'm sorry to bug you both for such a simple issue. But I can't get ingress working with removing this annotation. I'm getting 404 ( for both contour and nginx-ingress).  Below is how I'm creating my cluster 
```
cat <<EOF | kind create cluster --config -
kind: Cluster
nodes:
- role: control-plane
  kubeadmConfigPatches:
  - |
    kind: InitConfiguration
    nodeRegistration:
      kubeletExtraArgs:
        authorization-mode: "AlwaysAllow"
  extraPortMappings:
  - containerPort: 80
    hostPort: 80
    protocol: TCP
  - containerPort: 443
    hostPort: 443
    protocol: TCP
EOF
```

BenTheElder: the answer seems to be 0.22 ?
https://kubernetes.github.io/ingress-nginx/examples/rewrite/#rewrite-target

rajalokan: Yes. For ingress after 0.22 we must add ingress annotation `nginx.ingress.kubernetes.io/rewrite-target`. Hence following current documentation fails. 

Reason for this is our documentation mentions `ingress.kubernetes.io/rewrite-target` (notice missing **nginx**). Fixing this at common ingress manifest can be an issue with `Contour` example. I tried my hand with contour but couldn't get it working using our doc.

BenTheElder:
fixed by https://github.com/kubernetes-sigs/kind/pull/1499

BenTheElder:
thanks for the PR! we didn't actually need the rewrite target so it's been removed 👍

PR #1455, opened 2020-03-31
[Doc] Added TOC for config page

rajalokan:
- Adding table of contents for configuration page
- Let me know if leve 2 bullets need to be added to toc?

### DIFF
--- a/site/content/docs/user/configuration.md
+++ b/site/content/docs/user/configuration.md
@@ -15,7 +15,9 @@ This guide covers how to configure KIND cluster creation.
 
 We know this is currently a bit lacking right now and will expand it over time - PRs welcome!
 
-<!--TODO: TOC?-->
+1. [Getting Started](#getting-started)
+2. [Cluster-Wide Options](#clusterwide-options)
+3. [Per-Node Options](#pernode-options)
 
 ## Getting Started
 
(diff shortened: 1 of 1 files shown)


site/content/docs/user/configuration.md
```diff
@@ -15,7 +15,9 @@ This guide covers how to configure KIND cluster creation.
 
 We know this is currently a bit lacking right now and will expand it over time - PRs welcome!
 
-<!--TODO: TOC?-->
+1. [Getting Started](#getting-started)
+2. [Cluster-Wide Options](#clusterwide-options)
```
BenTheElder, line 19: I think we should prefer a TOC that won't get stale.

rajalokan: Got it. Quick look gives me 
- We can introduce `toc` from hugo side using something like [this](https://gohugo.io/content-management/toc/). But I tried that and it breaks our UI. We would need to bring in CSS changes to accommodate this `toc` ( probably something collapsible either in sidebar or at top of the page). 
- Tried looking at something from markdown side. But that would require adding new plugin. 

Both of this is beyond my knowledge level ( don't even want to try my hand in CSS side of things. :smile: ). Let me know if anything simple can be done.

BenTheElder: I haven't done this with hugo yet either

BenTheElder:
thanks for the PR but i'm going to hold off for a more future proof approach for now, we already have staleness issues with this sort of content 😅

PR #1456, opened 2020-03-31
Add kubeadmConfigPatches and metrics-server references to docs

rikatz:
This PR adds the following documentations:

* kubeadmConfigPatches (that can be used to put kube-proxy in IPVS mode) 
* deploying metrics-server in KinD as being a known issue

### DIFF
--- a/site/content/docs/user/known-issues.md
+++ b/site/content/docs/user/known-issues.md
@@ -32,6 +32,7 @@ It may additionally be helpful to:
 * [Chrome OS](#chrome-os)
 * [AppArmor](#apparmor)
 * [IPv6 port forwarding](#ipv6-port-forwarding)
+* [Deploying Metrics Server inside KinD](#metrics-server-inside-kind)
 
 ## Kubectl Version Skew
 
@@ -382,4 +383,40 @@ See Previous Discussion: [kind#1326]
 [version skew]: https://kubernetes.io/docs/setup/release/version-skew-policy/#supported-version-skew
 [Quick Start]: /docs/user/quick-start
 [AppArmor]: https://en.wikipedia.org/wiki/AppArmor
+
+## Metrics Server inside KinD
+
+Metrics Server needs to reach the metrics of each kubelet node before starting. Supposing the hostname of the container running kind is ``kind-control-plane`` and its IP inside docker is ``172.17.0.2`` you can edit the metrics-server deployment to force its /etc/hosts to point to the right kubelet IPs, as the following:
+
+{{< codeFromInline lang="yaml" >}}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: metrics-server
+  namespace: kube-system
+  labels:
+    k8s-app: metrics-server
+spec:
+ template:
+    metadata:
+      name: metrics-server
+      labels:
+        k8s-app: metrics-server
+    spec:
+      hostAliases:
+      - ip: "172.17.0.2"
+        hostnames:
+        - "kind-control-plane"
+      containers:
+      - args:
+        - --kubelet-insecure-tls
+        [...]
+
+{{< /codeFromInline >}}
+
+To check which IP your kind-control-plane is using you can verify with ``docker inspect kind-control-plane |grep IPAddress``, remembering to change the name of the container to whatever name your control-plane is using.
+
+{{< securitygoose >}}**NOTE**: --kubelet-insecure-tls is for testing only and should never be used in production{{</ securitygoose >}}
+
+
 
--- a/site/content/docs/user/configuration.md
+++ b/site/content/docs/user/configuration.md
@@ -177,6 +177,24 @@ nodes:
   image: kindest/node:v1.16.4@sha256:b91a2c2317a000f3a783489dfb755064177dbc3a0b2f4147d50f04825d016f55
 {{< /codeFromInline >}}
 
+### Patching Kubeadm configuration
+KinD uses kubeadm configuration structures to bootstrap the cluster. You can use the ``kubeadmConfigPatches`` snippet to patch any [configuration used by kubeadm](https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-config/) inside KinD.
+
+As an example, you can put kube-proxy in IPVS mode using the following example:
+
+{{< codeFromInline lang="yaml" >}}
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+kubeadmConfigPatches:
+- |
+  apiVersion: kubeproxy.config.k8s.io/v1alpha1
+  kind: KubeProxyConfiguration
+  metadata:
+    name: config
+  mode: ipvs
+{{< /codeFromInline >}}
+
+
 
 ## Per-Node Options
 


site/content/docs/user/known-issues.md
```diff
@@ -383,3 +384,39 @@ See Previous Discussion: [kind#1326]
...
+      labels:
+        k8s-app: metrics-server
+    spec:
+      hostAliases:
+      - ip: "172.17.0.2"
+        hostnames:
+        - "kind-control-plane"
+      containers:
+      - args:
+        - --kubelet-insecure-tls
```
BenTheElder, line 412: this seems substantially less than ideal, why do we need to do this?

I haven't looked at metrics server recently.

rikatz: Hi @BenTheElder,

It seems the CA that signs kubelet is different from the CA used by the APIServer. 

Metrics-server is able to connect without any problem to apiserver and fetch all the nodes  (``E0407 12:22:00.252207       1 reststorage.go:135] unable to fetch node metrics for node "kind118-control-plane": no metrics known for node``) but when it tries to connect to the kubelet it receiver the certificate signed by unknown authority error. It seems that the certificates from the endpoints are signed by different CA.

As an example, the following is the CA of the nodes from my cluster called kind118: 
``issuer=CN = kind118-control-plane-ca@1586194038``

And from APIServer:
``issuer=CN = kubernetes``

I don't know really if this is the case with KinD. Fetching the certs with openssl I really can see they're different. 

To be honest I couldn't find into KinD kubelet config where it's using a specific CA, so I'm guessing it auto generates a certificate for its secure port (10250) and that's why the metrics-server Pod doesn't know which CA is that and I'm skipping the tls validation.

aojea: certificates are handled by kubeadm https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/

rikatz: But also the certificates from the secure metrics endpoints (10250)? This is related to controller signing the certificates of each kubelet to communicate to api-server right?

aojea: no idea :sweat_smile: , @neolit123 may help us

rikatz: I've tested here and it generates an automatic certificate for the secure port:

```
cartman #  nohup ./kubelet & 
cartman # openssl s_client -connect 127.0.0.1:10250 |grep issuer\=
issuer=CN = cartman-ca@1586267640
```

BenTheElder: https://github.com/kubernetes-sigs/metrics-server/issues/146

this seems to be a relatively unresolved kubeadm / metrics-server issue. we should not be flipping off kubelet auth though.

rikatz: Indeed but in this case I'm skipping the validation of the CA from kubelet when metric-server connects into it (skip the unknown authority error) just to gather informations from kubelet.

I'm not changing or suggesting in anyway to change any kubelet parameter :)

Anyway I might take a better look into how use the same API Server CA into the kubelet HTTP/metrics certificates

rikatz: Also I've found the part which says this is an auto generated certificate:

```
    --tls-cert-file string                                                                                      File containing x509 Certificate used for serving HTTPS (with intermediate certs, if any, concatenated after server cert). If --tls-cert-file and --tls-private-key-file are not provided, a self-signed certificate and key are generated for the public address and saved to the directory passed to --cert-dir. (DEPRECATED: This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.)
```
So what if we change the configMap from kubelet to point to the same location that it uses to authenticate against api-server?

BenTheElder: I did a little more looking and haven't spotted anyone using metrics server w/ kubeadm without disabling auth which seems a bit alarming, we should probably try to contribute upstream here, I'm sort of over capacity but I'd appreciate us at least trying to see if the kubeadm folks have a solution here.

BenTheElder: we can probably reach out in the kubeadm slack instead of always pinging neolit123 as well :-)

aojea: 🤔 _I thought that neolit was ubiquitous_ 🙃 
 yeah, if this grows better using another channel

neolit123: well, the bots requested my review on this PR already.

currently, kubeadm needs rearchitecture to workaround an issue in the kubelet (and the trust of singing serving certificates in general).

[this](https://github.com/kubernetes-sigs/metrics-server/issues/146#issuecomment-460093946) is doable today in kubeadm, but i'm encouraged by some folks that understand the problem better than me to not take this particular action for now.

in general, higher level tools can also do [this](https://github.com/kubernetes-sigs/metrics-server/issues/146#issuecomment-472655656), but it has to be a running process (e.g. privileged controller or some Pod / CronJob). also the proposal there, as is, is going to need some tweaks once the k8s certificates API goes v1.

the following works too:
https://github.com/brancz/kube-rbac-proxy
[but it's pending documentation upstream](https://github.com/kubernetes/kubeadm/issues/1602) and is a bit entangled, but it allows you to not touch kubeadm or the kubelet and workaround this on the side of SARs.

with respect to kind, given the target cluster is not supposed to be in production, enabling `--kubelet-insecure-tls` seems tolerable, as long as there is the disclaimer. the disclaimer can also link to https://github.com/kubernetes-sigs/metrics-server/issues/146 saying "here you can find alternatives to this".

rikatz: @BenTheElder ok, so last comment and to be practical:

* I can split this PR into two, one dealing only with kubeadm config patching docs, and other dealing with metrics-server

* About @neolit123 suggestions, I think at least to move forward we can put into the ``security-goose`` alert the referenced issue, OR we can wait until a better solution is achieved for the integration between metrics-server and kubelet, and disregard this by now.

WDYT?

rikatz:
@BenTheElder do you want me to split this PR, so we got the kubeadmPatch part and can discuss a little bit more about metrics-server and kubelet certs? :)

BenTheElder:
yes please, let's do that!

site/content/docs/user/known-issues.md
BenTheElder: the one thing is I'd very much like to actually enable this core API in kind and I'm not inclined to do that by making it even more insecure.

whatever we document will get used widely, I feel like if metrics are important (... seems so) we can actually solve the problem

I also don't actually want people relying on kubeadm patches too heavily, they're available for power users that know what they're doing but they're a bit brittle across versions and I think most users have no idea what kubeadm config looks like etc.

BenTheElder: every time I find a heavily patched target I'm trying to introduce a way for kind to actually be aware that you e.g. want to enable a feature gate, and handle it intelligently (like toggling it throughout kubeadm config for all the components, being able to default certain gates in kind so our system components work, etc...)

ereslibre:
A comment from my side, I was able to deploy the `metrics-server` without any specific configuration, just some added parameters to it:

```
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    k8s-app: metrics-server
  name: metrics-server
  namespace: kube-system
spec:
  progressDeadlineSeconds: 600
  replicas: 1
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      k8s-app: metrics-server
  strategy:
    rollingUpdate:
      maxSurge: 25%
      maxUnavailable: 25%
    type: RollingUpdate
  template:
    metadata:
      creationTimestamp: null
      labels:
        k8s-app: metrics-server
      name: metrics-server
    spec:
      containers:
      - args:
        - --cert-dir=/tmp
        - --secure-port=4443
        - --kubelet-insecure-tls
        - --kubelet-preferred-address-types=InternalIP
        image: k8s.gcr.io/metrics-server-amd64:v0.3.6
        imagePullPolicy: IfNotPresent
        name: metrics-server
        ports:
        - containerPort: 4443
          name: main-port
          protocol: TCP
        resources: {}
        securityContext:
          readOnlyRootFilesystem: true
          runAsNonRoot: true
          runAsUser: 1000
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
        volumeMounts:
        - mountPath: /tmp
          name: tmp-dir
      dnsPolicy: ClusterFirst
      nodeSelector:
        kubernetes.io/arch: amd64
        kubernetes.io/os: linux
      restartPolicy: Always
      schedulerName: default-scheduler
      securityContext: {}
      serviceAccount: metrics-server
      serviceAccountName: metrics-server
      terminationGracePeriodSeconds: 30
      volumes:
      - emptyDir: {}
        name: tmp-dir
```

In my case, `--kubelet-insecure-tls` and `--kubelet-preferred-address-types=InternalIP` made the trick for me. The first will just ignore the presented certificate by the kubelets (not containing the IP address on the SAN), and the second will use the internal IP directly instead of trying to resolve the hostname with CoreDNS (that of course cannot resolve).

site/content/docs/user/known-issues.md
rikatz: So maybe disregarding the metrics server stuff, is it a better idea to implement a way that kind understands that kube-proxy should run with IPVS instead of iptables? I can try making this happen :)

BenTheElder:
The hostnames should resolve now, I think (in v0.8.0).

site/content/docs/user/known-issues.md
rikatz: About the heavy users, anyway wouldn't be good to have the patching method documented? Or is this some feature that KinD is targeting to remove later?

rikatz:
I've opened PR #1595 so instead of patching the kubeadm config KinD will support this as an internal configuration :D

If this is fine, I can close this PR and maybe open an issue so we can discuss about the metrics-server stuff before putting this into code or docs

rikatz:
Closing, as KinD now supports ipvs mode without the need of patching.

Issue #1416, opened 2020-03-18
kind doesn't work on btrfs


sudoforge:
**What happened**:

After installing `kind`, I ran a basic `create cluster`, which failed to start the control plane:

```zsh
➜ kind create cluster
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.17.0) 🖼
 ✓ Preparing nodes 📦
 ✓ Writing configuration 📜
 ✗ Starting control-plane 🕹️
ERROR: failed to create cluster: failed to init node with kubeadm: command "docker exec --privileged kind-control-plane kubeadm init --ignore-preflight-errors=all --config=/kind/kubeadm.conf --skip-token-print --v=6" failed with error: exit status 1
```

**What you expected to happen**:

I expected a cluster to be created.

**How to reproduce it (as minimally and precisely as possible)**:

Run `kind create cluster`.

**Anything else we need to know?**:

Using `-v 1`, I'm able to see that the error is related to a failed health check:

```zsh
➜ kind -v 1 create cluster
Creating cluster "kind" ...
DEBUG: docker/images.go:61] Image: kindest/node:v1.17.0@sha256:9512edae126da271b66b990b6fff768fbb7cd786c7d39e86bdf55906352fdf62 present locally
 ✓ Ensuring node image (kindest/node:v1.17.0) 🖼
 ✓ Preparing nodes 📦
 ✓ Writing configuration 📜
⠈⠑ Starting control-plane 🕹️
...
I0318 02:34:54.181668     119 waitcontrolplane.go:80] [wait-control-plane] Waiting for the API server to be healthy
I0318 02:34:54.183466     119 loader.go:375] Config loaded from file:  /etc/kubernetes/admin.conf
[wait-control-plane] Waiting for the kubelet to boot up the control plane as static Pods from directory "/etc/kubernetes/manifests". This can take up to 4m0s
I0318 02:34:54.187200     119 round_trippers.go:443] GET https://172.17.0.3:6443/healthz?timeout=10s  in 0 milliseconds
...
[kubelet-check] It seems like the kubelet isn't running or healthy.
[kubelet-check] The HTTP call equal to 'curl -sSL http://localhost:10248/healthz' failed with error: Get http://localhost:10248/healthz: dial tcp 127.0.0.1:10248: connect: connection refused.
```

_Output is truncated for brevity_

`docker logs kind-control-plane` doesn't seem to provide any useful information:

```zsh
[...]

PR #1464, opened 2020-04-06
Support BTRFS and ZFS docker storage

aojea:
It seems that mounting /dev/mapper from the host to the kind
nodes remove the limitation on Docker of using BTRFS or ZFS

https://kind.sigs.k8s.io/docs/user/known-issues/#docker-on-btrfs-or-zfs

Fixes: https://github.com/kubernetes-sigs/kind/issues/1416

### DIFF
--- a/pkg/cluster/internal/providers/docker/provision.go
+++ b/pkg/cluster/internal/providers/docker/provision.go
@@ -164,6 +164,13 @@ func commonArgs(cluster string, cfg *config.Cluster) ([]string, error) {
 	if usernsRemap() {
 		args = append(args, "--userns=host")
 	}
+
+	// handle Docker on Btrfs or ZFS
+	// https://github.com/kubernetes-sigs/kind/issues/1416#issuecomment-606514724
+	if mountDevMapper() {
+		args = append(args, "--volume", "/dev/mapper", "/dev/mapper")
+	}
+
 	return args, nil
 }
 
--- a/pkg/cluster/internal/providers/docker/util.go
+++ b/pkg/cluster/internal/providers/docker/util.go
@@ -36,3 +36,17 @@ func usernsRemap() bool {
 	}
 	return false
 }
+
+// mountDevMapper checks if the Docker storage driver is Btrfs or ZFS
+func mountDevMapper() bool {
+	storage := ""
+	cmd := exec.Command("docker", "info", "-f", "{{.Driver}}")
+	lines, err := exec.CombinedOutputLines(cmd)
+	if err != nil {
+		return false
+	}
+	if len(lines) > 0 {
+		storage = strings.ToLower(strings.TrimSpace(lines[0]))
+	}
+	return storage == "btrfs" || storage == "zfs"
+}


aojea:
/hold
I need to test it
/assign @BenTheElder

pkg/cluster/internal/providers/docker/provision.go
```diff
@@ -164,9 +164,32 @@ func commonArgs(cluster string, cfg *config.Cluster) ([]string, error) {
...
 		args = append(args, "--userns=host")
 	}
+
+	// handle Docker on Btrfs or ZFS
+	dockerStorage, err := getDockerStorage()
+	if err != nil {
+		return nil, errors.Wrap(err, "can't get Docker Storage")
+	}
+	if dockerStorage == "btrfs" || dockerStorage == "zfs" {
+		args = append(args, "--volume", "/dev/mapper", "/dev/mapper")
```
neolit123, line 174: implementation wise - LGTM.
i'd add a note here linking to external documentation why `/dev/mapper` solves the issue (if it indeed does).

pkg/cluster/internal/providers/docker/provision.go
```diff
@@ -164,9 +164,32 @@ func commonArgs(cluster string, cfg *config.Cluster) ([]string, error) {
 	if usernsRemap() {
 		args = append(args, "--userns=host")
 	}
+
+	// handle Docker on Btrfs or ZFS
+	dockerStorage, err := getDockerStorage()
```
BenTheElder, line 169: in the future this should be refactored to get all the docker info we want, we should avoid unnecessarily making extra info calls.

pkg/cluster/internal/providers/docker/provision.go
```diff
@@ -164,9 +164,32 @@ func commonArgs(cluster string, cfg *config.Cluster) ([]string, error) {
...
+	if err != nil {
+		return nil, errors.Wrap(err, "can't get Docker Storage")
```
BenTheElder, line 171: this is inconsistent with usernsRemap

BenTheElder:
/approve
nit: re consistency with usernsremap

pkg/cluster/internal/providers/docker/provision.go
aojea: fixed

pkg/cluster/internal/providers/docker/util.go
```diff
@@ -36,3 +36,17 @@ func usernsRemap() bool {
 	}
 	return false
 }
+
+// mountDevMapper checks if the Docker storage driver is Btrfs or ZFS
+func mountDevMapper() bool {
+	storage := ""
+	cmd := exec.Command("docker", "info", "-f", "{{.Driver}}")
+	lines, err := exec.CombinedOutputLines(cmd)
+	if err != nil {
```
BenTheElder, line 45: i'm not sure if this is the right approach here, but at least it's consistent.

this needs to get refactored anyhow, kind bootstrap has a lot of unnecessary complications I want to clobber after #148

BenTheElder:
/lgtm
/approve
thanks!

pkg/cluster/internal/providers/docker/util.go
aojea: is consistent and is less disruptive, in case of failure it works as usual 
🤷‍♂

PR #1488, opened 2020-04-17
drop extraneous dot from path

BenTheElder:
/shrug
we're going to get this right someday


BenTheElder:
and now it's building :tada:

PR #1492, opened 2020-04-17
create redirect for release downloads

BenTheElder:
After this you can use `https://kind.sigs.k8s.io/dl/v0.7.0/kind-linux-amd64` (and generically any of our release binaries are served by this redirect)

Updated our docs to use this as it's shorter and it gives us more control over hosting (e.g. we have a wg-k8s prod GCS bucket now, and GCS seems to much faster than github artifacts).

### DIFF
--- a/site/layouts/index.redirects
+++ b/site/layouts/index.redirects
@@ -2,4 +2,5 @@
 {{- range $apiVersions }}
 /{{ . }}     https://godoc.org/sigs.k8s.io/kind/pkg/apis/config/{{ . }}
 {{- end }}
+/dl/v* https://github.com/kubernetes-sigs/kind/releases/download/v:splat
 /dl/* https://storage.googleapis.com/k8s-staging-kind/:splat
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ into your `$PATH`:
 On Mac & Linux:
 
 ```console
-curl -Lo ./kind "https://github.com/kubernetes-sigs/kind/releases/download/v0.7.0/kind-$(uname)-amd64"
+curl -Lo ./kind "https://kind.sigs.k8s.io/dl/v0.7.0/kind-$(uname)-amd64"
 chmod +x ./kind
 mv ./kind /some-dir-in-your-PATH/kind
 ```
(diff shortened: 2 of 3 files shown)


BenTheElder:
https://docs.netlify.com/routing/redirects/#rule-processing-order

> The redirects engine will process the first matching rule it finds, reading from top to bottom. Rules in the _redirects file are always processed first, followed by rules in the Netlify configuration file.

tao12345666333:
I have a worry, there are currently many users of KIND from China, but GCS bucket cannot be downloaded in China.

BenTheElder:
This currently downloads from GitHub.

BenTheElder:
Though, fwiw, Kubernetes's official releases are in GCS as I remember, and
there are many install methods.

This is just *one* download url. We can always host mirrors etc.

site/layouts/index.redirects
```diff
@@ -2,4 +2,5 @@
 {{- range $apiVersions }}
 /{{ . }}     https://godoc.org/sigs.k8s.io/kind/pkg/apis/config/{{ . }}
 {{- end }}
-/dl/* https://storage.googleapis.com/k8s-staging-kind/:splat
+/dl/v* https://github.com/kubernetes-sigs/kind/releases/download/v:splat
+/dl/* https://storage.googleapis.com/k8s-staging-kind/:splat
```
aojea, line 6: do we need an end of line here?

BenTheElder:
You can see that the redirects already work on the preview site.

It's a toml file, the newlines don't matter

site/layouts/index.redirects
BenTheElder: no

BenTheElder:
the nightly builds are in GCS (and always have been) though.

is netlify available in china? I see some indication that it also may not be.

preview URLs (will download binaries)
https://deploy-preview-1492--k8s-kind.netlify.app/dl/latest/kind-linux-amd64
https://deploy-preview-1492--k8s-kind.netlify.app/dl/v0.7.0/kind-linux-amd64

aojea:
@tao12345666333 can you confirm everything is ok and working well from China?

tao12345666333:
@aojea only **/dl/v** works well.

BenTheElder:
FYI @tao12345666333 we can use:
https://docs.netlify.com/routing/redirects/redirect-options/#redirect-by-country-or-language
to enable GCS to host releases w/ better speed etc., but on `Country=cn` use github releases.

I'll probably do this at some point. we may also be able to leverage https://github.com/kubernetes/release/tree/master/cmd/gh2gcs

BenTheElder:
for nightlies though, github isn't terribly well suited to pushing CI builds from prow, I'm not sure of a good way to host those

tao12345666333:
> FYI @tao12345666333 we can use:
> https://docs.netlify.com/routing/redirects/redirect-options/#redirect-by-country-or-language
> to enable GCS to host releases w/ better speed etc., but on `Country=cn` use github releases.

Wow, SGTM!





> for nightlies though, github isn't terribly well suited to pushing CI builds from prow, I'm not sure of a good way to host those

I think as long as there is a officially released version on GitHub as a backup, it will be fine.

PR #1493, opened 2020-04-17
Support KIND_CLUSTER_NAME.

amwat:
fixes https://github.com/kubernetes-sigs/kind/issues/1094

/cc @BenTheElder

### DIFF
--- a/pkg/cmd/kind/create/cluster/createcluster.go
+++ b/pkg/cmd/kind/create/cluster/createcluster.go
@@ -30,6 +30,7 @@ import (
 	"sigs.k8s.io/kind/pkg/errors"
 	"sigs.k8s.io/kind/pkg/log"
 
+	"sigs.k8s.io/kind/pkg/internal/cli"
 	"sigs.k8s.io/kind/pkg/internal/runtime"
 )
 
@@ -51,6 +52,7 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 		Short: "Creates a local Kubernetes cluster",
 		Long:  "Creates a local Kubernetes cluster using Docker container 'nodes'",
 		RunE: func(cmd *cobra.Command, args []string) error {
+			cli.OverrideDefaultName(cmd.Flags())
 			return runE(logger, streams, flags)
 		},
 	}
--- a/pkg/internal/cli/override.go
+++ b/pkg/internal/cli/override.go
@@ -1,0 +1,18 @@
+package cli
+
+import (
+	"os"
+
+	"github.com/spf13/pflag"
+)
+
+// OverrideDefaultName conditionally allows overriding the default cluster name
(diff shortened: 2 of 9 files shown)


pkg/cmd/kind/create/cluster/createcluster.go
```diff
@@ -51,6 +52,9 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 		Short: "Creates a local Kubernetes cluster",
 		Long:  "Creates a local Kubernetes cluster using Docker container 'nodes'",
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if !cmd.Flags().Changed("name") {
```
BenTheElder, line 55: we could dry this out by just passing the `cmd.Flags()` (flagset pointer) to `OverrideDefaultName` and then internally check `flags.Changed("name")` then call `flags.Set("name", value)`

pkg/internal/cli/override.go
```diff
@@ -0,0 +1,14 @@
...
+import (
+	"os"
+
+	"sigs.k8s.io/kind/pkg/log"
+)
+
+func OverrideDefaultName(logger log.Logger, defaultName *string) {
+	if name := os.Getenv("KIND_CLUSTER_NAME"); name != "" {
+		*defaultName = name
+		logger.V(1).Infof("using %q due to KIND_CLUSTER_NAME", name)
```
BenTheElder, line 12: using what? this message seems a little obtuse.

pkg/cmd/kind/create/cluster/createcluster.go
amwat: Done.
I initially had that, but figured that checking the specific `name` flag logic should probably live near the command than in the method that just checks the env.
/shrug

pkg/internal/cli/override.go
amwat: Clarified.

pkg/internal/cli/override.go
```diff
@@ -0,0 +1,17 @@
...
+
+	"github.com/spf13/pflag"
+	"sigs.k8s.io/kind/pkg/log"
+)
+
+func OverrideDefaultName(logger log.Logger, fs *pflag.FlagSet) {
+	if !fs.Changed("name") {
+		if name := os.Getenv("KIND_CLUSTER_NAME"); name != "" {
+			fs.Set("name", name)
+			logger.V(1).Infof("using default cluster name as %q due to KIND_CLUSTER_NAME", name)
```
BenTheElder, line 14: Last nit: it's not the default right?
using cluster name %q from KIND_CLUSTER_NAME

Or just drop this all together. I don't think most apps log this sort of thing.

amwat: done

BenTheElder:
/lgtm
/approve
Maybe write elaborate unit tests in a follow up ;-)
Thanks!

PR #1494, opened 2020-04-17
fix config featuregate handling

BenTheElder:
fixes a panic on some kubernetes versions
we need to not be potentially assigning to a nil map lower down.

/priority critical-urgent
/cc @amwat @aojea


aojea:
⚡

PR #1495, opened 2020-04-17
support NO_COLOR

BenTheElder:
thanks @liggitt 
https://github.com/kubernetes/kubernetes/pull/73032/files#r410491521

### DIFF
--- a/pkg/internal/env/term_test.go
+++ b/pkg/internal/env/term_test.go
@@ -99,6 +99,15 @@ func TestIsSmartTerminal(t *testing.T) {
 			Writer:  &testFakeTTY{},
 		},
 		{
+			Name: "tty, NO_COLOR=",
+			FakeEnv: map[string]string{
+				"NO_COLOR": "",
+			},
+			GOOS:    "linux",
+			IsSmart: false,
+			Writer:  &testFakeTTY{},
+		},
+		{
 			Name: "tty, Travis CI",
 			FakeEnv: map[string]string{
 				"TRAVIS":                      "true",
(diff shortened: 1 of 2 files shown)


pkg/internal/env/term_test.go
```diff
@@ -112,8 +121,9 @@ func TestIsSmartTerminal(t *testing.T) {
 	for _, tc := range cases {
 		tc := tc // capture tc
 		t.Run(tc.Name, func(t *testing.T) {
-			res := isSmartTerminal(tc.Writer, tc.GOOS, func(s string) string {
-				return tc.FakeEnv[s]
+			res := isSmartTerminal(tc.Writer, tc.GOOS, func(s string) (string, bool) {
+				k, set := tc.FakeEnv[s]
```
BenTheElder, line 125: irritatingly it seems if you want the 2 return value map mode you have to use assignment, can't just return

aojea: this is what I found:

> Go won't let you use the return value from a function called with a *go* statement, because there's no guarantee the return value will be ready before we attempt to use it

BenTheElder: this is not a function call or a go statement.
this is about maps having single or multiple assignment

aojea: Are you sure? I see a lot of stuff here :) https://github.com/golang/go/blob/master/src/runtime/map.go

BenTheElder: yes? semantically indexing a map is not a function call. this is https://golang.org/ref/spec#Index_expressions

obviously it's internally implemented as one.

BenTheElder: the behavior change on using single or multiple assignment is widely regarded as one of the most surprising aspects of the language. it's just biting us more than usual here.

PR #1503, opened 2020-04-22
Set listenAddress in example to 127.0.0.1

zegl:
Docker for mac fails with the following if --publish is set to 127.0.0.0:

> bind: can't assign requested address.

Updating the example to be working out of the box.


BenTheElder:
/ok-to-test
/lgtm
/approve
thanks! I don't think this was intended

Issue #1461, opened 2020-04-02
Cluster does not initialize properly, kindnet times out and crashes.


salasrod:
**What happened**:

CoreDNS fails to start, and Kindnet has timeouts in their logs:

```
I0402 17:39:58.069948       1 main.go:64] hostIP = 172.17.0.2
podIP = 172.17.0.2
I0402 17:40:33.166911       1 main.go:104] Failed to get nodes, retrying after error: Get https://10.96.0.1:443/api/v1/nodes: dial tcp 10.96.0.1:443: i/o timeout
I0402 17:41:03.167486       1 main.go:104] Failed to get nodes, retrying after error: Get https://10.96.0.1:443/api/v1/nodes: dial tcp 10.96.0.1:443: i/o timeout
I0402 17:41:34.167858       1 main.go:104] Failed to get nodes, retrying after error: Get https://10.96.0.1:443/api/v1/nodes: dial tcp 10.96.0.1:443: i/o timeout
I0402 17:42:06.168271       1 main.go:104] Failed to get nodes, retrying after error: Get https://10.96.0.1:443/api/v1/nodes: dial tcp 10.96.0.1:443: i/o timeout
I0402 17:42:39.168601       1 main.go:104] Failed to get nodes, retrying after error: Get https://10.96.0.1:443/api/v1/nodes: dial tcp 10.96.0.1:443: i/o timeout
panic: Reached maximum retries obtaining node list: Get https://10.96.0.1:443/api/v1/nodes: dial tcp 10.96.0.1:443: i/o timeout
```

**What you expected to happen**:

CoreDNS containers to start, and kindnet not to crash

**How to reproduce it (as minimally and precisely as possible)**:

`kind create cluster`
`kubectl -n kube-system logs kindnet-xxxxx`

```kubectl -n kube-system get pods -o wide
NAME                                         READY   STATUS             RESTARTS   AGE   IP           NODE                 NOMINATED NODE   READINESS GATES
coredns-6955765f44-tvtk5                     0/1     Pending            0          20m   <none>       <none>               <none>           <none>
coredns-6955765f44-vbpxc                     0/1     Pending            0          20m   <none>       <none>               <none>           <none>
etcd-kind-control-plane                      1/1     Running            0          20m   172.17.0.2   kind-control-plane   <none>           <none>
kindnet-wgsbc                                0/1     CrashLoopBackOff   5          20m   172.17.0.2   kind-control-plane   <none>           <none>
[...]

Issue #284, opened 2019-02-08
node level DNS may need override


BenTheElder:
cc @floreks 

TLDR floreks had kind in dind on kubeadm, in which case the kind "node" DNS was not configured in a working way. We may need to allow DNS to be overridden at the node level to fix EG mage pulling

BenTheElder:
cc @alejandrox1 I think we should probably add something about DNS mismatch to the known-issues.
In short:
- DNS from the host environment will propagate to the kind "nodes"
- If the host DNS is not appropriate for the nodes (EG DNS in an outer Kubernetes cluster dind pod) then this will cause a problem

Work around for the pod case is to configure the outer dind pod DNS to use upstream resolvers instead of the in-cluster resolvers.

More general solution will probably require that kind support configuring node-level DNS options.

mikkeloscar:
I just hit this issue when trying to run https://github.com/kubernetes-incubator/metrics-server in a `kind` cluster.

The metrics server tries to go to the kubelets via the node hostname and fails:

```
$ kubectl --namespace kube-system logs -f metrics-server-cd4946b-fxgzf 
I0321 13:44:56.710455       1 serving.go:273] Generated self-signed cert (apiserver.local.config/certificates/apiserver.crt, apiserver.local.config/certificates/apiserver.key)
[restful] 2019/03/21 13:44:57 log.go:33: [restful/swagger] listing is available at https://:443/swaggerapi
[restful] 2019/03/21 13:44:57 log.go:33: [restful/swagger] https://:443/swaggerui/ is mapped to folder /swagger-ui/
I0321 13:44:57.461103       1 serve.go:96] Serving securely on [::]:443
E0321 13:45:07.834706       1 reststorage.go:144] unable to fetch pod metrics for pod e2e/es-operator-74b79684df-kkfjl: no metrics known for pod
E0321 13:45:07.834733       1 reststorage.go:144] unable to fetch pod metrics for pod e2e/es-master-c895c6f54-nzzfw: no metrics known for pod
E0321 13:45:08.733047       1 reststorage.go:144] unable to fetch pod metrics for pod e2e/es-operator-74b79684df-kkfjl: no metrics known for pod
E0321 13:45:08.733074       1 reststorage.go:144] unable to fetch pod metrics for pod e2e/es-master-c895c6f54-nzzfw: no metrics known for pod
[...]

PR #1508, opened 2020-04-24
Use the network luke

BenTheElder:
implement #148 

NOTE: breaking change for docker backend, requires new node images, switches nodes from the default bridge to a `kind` network. Also causes clusters to start back up on reboot...

### DIFF
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -181,7 +181,7 @@ enable_network_magic(){
   sed -e "s/${docker_embedded_dns_ip}/${docker_host_ip}/g" /etc/resolv.conf.original >/etc/resolv.conf
 
   # fixup IPs in manifests ...
-  # TODO: dual-stack LOL :this-is-fine:
+  # TODO: ensure this supports rebooting the host w/ IPv6 clusters
   curr_ip=$(hostname --ip-address)
   if [ -f /kind/old-ip ]; then
     old_ip=$(cat /kind/old-ip)
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -179,6 +179,15 @@ enable_network_magic(){
   # now we can ensure that DNS is configured to use our IP
   cp /etc/resolv.conf /etc/resolv.conf.original
   sed -e "s/${docker_embedded_dns_ip}/${docker_host_ip}/g" /etc/resolv.conf.original >/etc/resolv.conf
+
+  # fixup IPs in manifests ...
+  # TODO: dual-stack LOL :this-is-fine:
+  curr_ip=$(hostname --ip-address)
+  if [ -f /kind/old-ip ]; then
+    old_ip=$(cat /kind/old-ip)
+    sed -i "s#${old_ip}#${curr_ip}#" /etc/kubernetes/manifests/*.yaml /var/lib/kubelet/kubeadm-flags.env
+  fi
+  echo -n "${curr_ip}" >/kind/old-ip
 }
 
 # run pre-init fixups
--- a/pkg/build/nodeimage/const_storage.go
+++ b/pkg/build/nodeimage/const_storage.go
@@ -21,7 +21,7 @@ The default PV driver manifest and images are provisionally rancher.io/local-pat
 NOTE: we have customized it in the following ways:
 - storage is under /var instead of /opt
 - debian-base is used as the helper image (k8s already ships this upstream as the base for many images) instead of busybox
-- schedule to "master" kubeadm nodes (control-plane host)
+- schedule to linux nodes
 - install as the default storage class
 */
 
@@ -87,7 +87,7 @@ spec:
         app: local-path-provisioner
     spec:
       nodeSelector:
-        node-role.kubernetes.io/master: ''
+        kubernetes.io/os: linux
       tolerations:
       - key: node-role.kubernetes.io/master
         operator: Equal
--- a/pkg/cluster/internal/create/actions/config/config.go
+++ b/pkg/cluster/internal/create/actions/config/config.go
@@ -19,8 +19,6 @@ package config
 
 import (
 	"bytes"
-	"fmt"
-	"net"
 	"strings"
 
 	"sigs.k8s.io/kind/pkg/cluster/constants"
@@ -53,26 +51,10 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 		return err
 	}
 
-	// get the control plane endpoint, in case the cluster has an external load balancer in
-	// front of the control-plane nodes
-	controlPlaneEndpoint, controlPlaneEndpointIPv6, err := nodeutils.GetControlPlaneEndpoint(allNodes)
-	if err != nil {
-		// TODO(bentheelder): logging here
-		return err
-	}
-
-	// configure the right protocol addresses
-	if ctx.Config.Networking.IPFamily == "ipv6" {
-		controlPlaneEndpoint = controlPlaneEndpointIPv6
-	}
-
-	// TODO: clean this up and make it optional
-
-	endpointNode, err := nodeutils.APIServerEndpointNode(allNodes)
+	controlPlaneEndpoint, err := ctx.ClusterContext.GetAPIServerInternalEndpoint()
 	if err != nil {
 		return err
 	}
-	controlPlaneEndpoint = net.JoinHostPort(endpointNode.String(), fmt.Sprintf("%d", common.APIServerInternalPort))
 
 	// create kubeadm init config
 	fns := []func() error{}
--- a/pkg/cluster/internal/create/actions/config/config.go
+++ b/pkg/cluster/internal/create/actions/config/config.go
@@ -19,6 +19,8 @@ package config
 
 import (
 	"bytes"
+	"fmt"
+	"net"
 	"strings"
 
 	"sigs.k8s.io/kind/pkg/cluster/constants"
@@ -63,6 +65,14 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 	if ctx.Config.Networking.IPFamily == "ipv6" {
 		controlPlaneEndpoint = controlPlaneEndpointIPv6
 	}
+
+	// TODO: clean this up and make it optional
+
+	endpointNode, err := nodeutils.APIServerEndpointNode(allNodes)
+	if err != nil {
+		return err
+	}
+	controlPlaneEndpoint = net.JoinHostPort(endpointNode.String(), fmt.Sprintf("%d", common.APIServerInternalPort))
 
 	// create kubeadm init config
 	fns := []func() error{}
--- a/pkg/cluster/internal/providers/docker/network.go
+++ b/pkg/cluster/internal/providers/docker/network.go
@@ -20,6 +20,8 @@ import (
 	"regexp"
 
 	"sigs.k8s.io/kind/pkg/exec"
+
+	"sigs.k8s.io/kind/pkg/internal/apis/config"
 )
 
 // TODO: we'll probably allow configuring this
@@ -33,7 +35,9 @@ import (
 const fixedNetworkName = "kind"
 
 // ensureNetwork checks if docker network by name exists, if not it creates it
-func ensureNetwork(name string) error {
+func ensureNetwork(name string, ipFamily config.ClusterIPFamily) error {
+	// TODO: the network might already exist and not have ipv6 ... :|
+	// discussion: https://github.com/kubernetes-sigs/kind/pull/1508#discussion_r414594198
 	out, err := exec.Output(exec.Command(
 		"docker", "network", "ls",
 		"--filter=name=^"+regexp.QuoteMeta(name)+"$",
@@ -45,6 +49,9 @@ func ensureNetwork(name string) error {
 	// network already exists
 	if string(out) == name+"\n" {
 		return nil
+	}
+	if ipFamily == config.IPv6Family {
+		return exec.Command("docker", "network", "create", "-d=bridge", "--ipv6", name).Run()
 	}
 	return exec.Command("docker", "network", "create", "-d=bridge", name).Run()
 }
--- a/pkg/cluster/internal/providers/docker/network.go
+++ b/pkg/cluster/internal/providers/docker/network.go
@@ -17,6 +17,8 @@ limitations under the License.
 package docker
 
 import (
+	"regexp"
+
 	"sigs.k8s.io/kind/pkg/exec"
 )
 
@@ -34,7 +36,7 @@ const fixedNetworkName = "kind"
 func ensureNetwork(name string) error {
 	out, err := exec.Output(exec.Command(
 		"docker", "network", "ls",
-		"--filter=name=^"+name,
+		"--filter=name=^"+regexp.QuoteMeta(name)+"$",
 		"--format={{.Name}}",
 	))
 	if err != nil {
--- a/pkg/cluster/internal/providers/docker/network.go
+++ b/pkg/cluster/internal/providers/docker/network.go
@@ -22,7 +22,7 @@ import (
 
 // TODO: we'll probably allow configuring this
 //
-// however currently picking a single network is equivilant to the previous
+// however currently picking a single network is equivalent to the previous
 // behavior *except* that we moved from the default bridge to a user defined
 // network because the default bridge is actually special versus any other
 // docker network and lacks the emebdded DNS
--- a/pkg/cluster/internal/providers/docker/network.go
+++ b/pkg/cluster/internal/providers/docker/network.go
@@ -50,8 +50,11 @@ func ensureNetwork(name string, ipFamily config.ClusterIPFamily) error {
 	if string(out) == name+"\n" {
 		return nil
 	}
+	// TODO: ipv6 subnet should probably not be fixed
+	// Though maybe just require the user to handle this by creating the network
+	// as they desire before running kind ...
 	if ipFamily == config.IPv6Family {
-		return exec.Command("docker", "network", "create", "-d=bridge", "--ipv6", name).Run()
+		return exec.Command("docker", "network", "create", "-d=bridge", "--ipv6", "--subnet=fc00:db8:2::/64", name).Run()
 	}
 	return exec.Command("docker", "network", "create", "-d=bridge", name).Run()
 }
--- a/pkg/cluster/internal/providers/docker/network.go
+++ b/pkg/cluster/internal/providers/docker/network.go
@@ -1,0 +1,48 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package docker
+
+import (
+	"sigs.k8s.io/kind/pkg/exec"
+)
+
+// TODO: we'll probably allow configuring this
+//
+// however currently picking a single network is equivilant to the previous
+// behavior *except* that we moved from the default bridge to a user defined
+// network because the default bridge is actually special versus any other
+// docker network and lacks the emebdded DNS
+//
+// for now this also makes it easier for apps to join the same network
+const fixedNetworkName = "kind"
+
+// ensureNetwork checks if docker network by name exists, if not it creates it
+func ensureNetwork(name string) error {
+	out, err := exec.Output(exec.Command(
+		"docker", "network", "ls",
+		"--filter=name=^"+name,
+		"--format={{.Name}}",
+	))
+	if err != nil {
+		return err
+	}
+	// network already exists
+	if string(out) == name+"\n" {
+		return nil
+	}
+	return exec.Command("docker", "network", "create", "-d=bridge", name).Run()
+}
--- a/pkg/cluster/internal/providers/docker/provider.go
+++ b/pkg/cluster/internal/providers/docker/provider.go
@@ -58,6 +58,10 @@ func (p *Provider) Provision(status *cli.Status, cluster string, cfg *config.Clu
 		return err
 	}
 
+	if err := ensureNetwork(fixedNetworkName); err != nil {
+		return errors.Wrap(err, "failed to ensure docker network")
+	}
+
 	// actually provision the cluster
 	icons := strings.Repeat("📦 ", len(cfg.Nodes))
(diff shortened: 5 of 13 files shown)


BenTheElder:
this almost definitely:
- doesn't work with ipv6 single stack (should still  work, host reboot / dockerd reboot won't work though, which is the status quo)
- may entirely break podman ipv6 single stack (I don't think we'd tested this yet anyhow cc @amwat)

on docker node provider and ipv4 nodes though this should work great
it also resolves using the host DNS properly #284 and probably fixes #1461 (need to confirm on a setup with the issue)

both should be resolvable in a future iteration, but

pkg/cluster/internal/create/actions/config/config.go
```diff
@@ -19,6 +19,8 @@ package config
 
 import (
 	"bytes"
+	"fmt"
+	"net"
```
BenTheElder: this whole commit can be dropped but I'm not fighting with git at 2am and it's not going to show up in the git history anyhow.

images/base/files/usr/local/bin/entrypoint
```diff
@@ -179,6 +179,15 @@ enable_network_magic(){
   # now we can ensure that DNS is configured to use our IP
   cp /etc/resolv.conf /etc/resolv.conf.original
   sed -e "s/${docker_embedded_dns_ip}/${docker_host_ip}/g" /etc/resolv.conf.original >/etc/resolv.conf
+
+  # fixup IPs in manifests ...
+  # TODO: dual-stack LOL :this-is-fine:
+  curr_ip=$(hostname --ip-address)
```
BenTheElder, line 185: we'll want to swap this out with something else. this was sufficient for bootstrapping it but we'll need to actually change this to look at /kind/old-ip and get an IP of the same family.

not going to be too hard, but wasn't relevant to the bigger problems. bash wrangling for ipv6 singlestack can be a follow up, but should happen prior to releasing any official kind updates.

pkg/cluster/internal/providers/docker/network.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+// TODO: we'll probably allow configuring this
+//
+// however currently picking a single network is equivilant to the previous
+// behavior *except* that we moved from the default bridge to a user defined
+// network because the default bridge is actually special versus any other
+// docker network and lacks the emebdded DNS
+//
+// for now this also makes it easier for apps to join the same network
```
BenTheElder, line 30: I.E. you can do `docker run --network=kind --name=metallb-test-cluster-sidecar ....` when doing fun test stuff.
Having a predictable default network makes that saner.

We could do one per cluster but not only is that more complex, it's going to cause IP allocation issues more readily when you have lots of small clusters (which is much moreso the usage pattern than one cluster with a large number of nodes).

In the future this can be configurable and this can just be the default, but it's roughly equivilant to the current state and keeps the PR focused on the necessary changes.

pkg/cluster/internal/providers/docker/provision.go
```diff
@@ -144,7 +144,33 @@ func commonArgs(cluster string, cfg *config.Cluster) ([]string, error) {
...
+		// - unless-stopped will also retry failures indefinitely, similar to always
+		// except that it won't restart when the container is `docker stop`ed
+		// - on-failure is not great, we're only interested in restarting on
+		// reboots, not failures. *however* we can limit the number of retries
+		// *and* it forgets all state on dockerd restart and retries anyhow.
+		// - on-failure:0 is what we want .. restart on failures, except max
+		// retries is 0, so only restart on reboots.
+		// however this _actually_ means the same thing as always
+		// so the closest thing is on-failure:1, which will retry *once*
+		"--restart=on-failure:1",
```
BenTheElder, line 173: this means if the node fails the entrypoint it will actually quickly start again, which is not quite ideal, but doesn't seem too problematic.

I did quite a bit of testing around this behavior before settling on this as "ok".

BenTheElder:
hehe surprise surprise we did actually break ipv6
/hold

BenTheElder:
will fix ipv6 in the morning :sleeping:

images/base/files/usr/local/bin/entrypoint
```diff
@@ -179,6 +179,15 @@ enable_network_magic(){
   # now we can ensure that DNS is configured to use our IP
   cp /etc/resolv.conf /etc/resolv.conf.original
   sed -e "s/${docker_embedded_dns_ip}/${docker_host_ip}/g" /etc/resolv.conf.original >/etc/resolv.conf
+
+  # fixup IPs in manifests ...
+  # TODO: dual-stack LOL :this-is-fine:
```
aojea, line 184: dual stack control plane will use only one ip family, the default one, no problem here from the top of my head

pkg/cluster/internal/providers/docker/network.go
```diff
@@ -0,0 +1,48 @@
...
+
+package docker
+
+import (
+	"sigs.k8s.io/kind/pkg/exec"
+)
+
+// TODO: we'll probably allow configuring this
+//
+// however currently picking a single network is equivilant to the previous
```
aojea, line 25: typo for the typohunters :)

pkg/cluster/internal/providers/docker/network.go
aojea: also it will modify current behaviour where multiple cluster are in the same network.
IIRC istio was using it

pkg/cluster/internal/providers/docker/network.go
```diff
@@ -0,0 +1,48 @@
...
+// docker network and lacks the emebdded DNS
+//
+// for now this also makes it easier for apps to join the same network
+const fixedNetworkName = "kind"
+
+// ensureNetwork checks if docker network by name exists, if not it creates it
+func ensureNetwork(name string) error {
+	out, err := exec.Output(exec.Command(
+		"docker", "network", "ls",
+		"--filter=name=^"+name,
```
aojea, line 37: we should use a stricter regex or another way to detect the network

```
docker network ls --filter=name=^k
NETWORK ID          NAME                DRIVER              SCOPE
2054b2e0e513        kind                bridge              local
efc7ccd36802        kindy               bridge              local
```

```
docker network ls --filter=name=^kind$
NETWORK ID          NAME                DRIVER              SCOPE
2054b2e0e513        kind                bridge              local
 aojeagarcia  ~  Downloads  docker network ls --filter=name=^kindy$
NETWORK ID          NAME                DRIVER              SCOPE
efc7ccd36802        kindy               bridge              local
```
or with docker inspect
```
docker inspect kin
[]
Error: No such object: kin
 aojeagarcia  ~  Downloads  1  echo $? 
1
```
```
 docker inspect kind
[
    {
        "Name": "kind",
        "Id": "2054b2e0e513111b323738371afaa6ac80ae22ad4d58e78107f9d4ca6beff7b0",
        "Created": "2020-04-24T15:40:05.557245181+02:00",
...
echo $? 
0
```

pkg/cluster/internal/providers/docker/network.go
```diff
@@ -0,0 +1,48 @@
...
+		"--format={{.Name}}",
+	))
+	if err != nil {
+		return err
+	}
+	// network already exists
+	if string(out) == name+"\n" {
```
aojea, line 44: using `docker inspect` we can avoid this

pkg/cluster/internal/providers/docker/network.go
```diff
@@ -0,0 +1,48 @@
...
+		return nil
+	}
+	return exec.Command("docker", "network", "create", "-d=bridge", name).Run()
```
aojea, line 47: TODO

```go
if ipv6 {
exec.Command("docker", "network", "create", "--ipv6","-d=bridge", name).Run()
```

pkg/cluster/internal/providers/docker/provider.go
```diff
@@ -58,6 +58,10 @@ func (p *Provider) Provision(status *cli.Status, cluster string, cfg *config.Clu
 		return err
 	}
 
+	if err := ensureNetwork(fixedNetworkName); err != nil {
+		return errors.Wrap(err, "failed to ensure docker network")
```
aojea, line 62: should we fall back to the default docker network?

pkg/exec/helpers.go
```diff
@@ -86,6 +86,14 @@ func OutputLines(cmd Cmd) (lines []string, err error) {
 	return lines, err
 }
 
+// Output is like os/exec's cmd.Output, but over our Cmd interface
+func Output(cmd Cmd) ([]byte, error) {
```
aojea, line 90: we don't need this if we use `docker inspect` to detect the network

BenTheElder: There's no reason not to have this, and docker inspect involves moving more data that we don't use.

pkg/cluster/internal/providers/docker/provider.go
BenTheElder: No?
The default network has no embedded DNS and docker had had user mode networks for a very long time.

pkg/cluster/internal/providers/docker/network.go
BenTheElder: I don't think so, but this isn't really a problem anyhow. I'm aware of inspect and chose this method.

This method is more precise, it only errors if docker cannot be called, and the value it returns is either "" or "name\n"

pkg/cluster/internal/providers/docker/network.go
BenTheElder: We should fix the regex.

pkg/cluster/internal/providers/docker/provider.go
```diff
@@ -173,6 +173,20 @@ func (p *Provider) GetAPIServerEndpoint(cluster string) (string, error) {
...
+	// locate the node that hosts this
+	allNodes, err := p.ListNodes(cluster)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to list nodes")
+	}
+	n, err := nodeutils.APIServerEndpointNode(allNodes)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to get api server endpoint")
+	}
+	return net.JoinHostPort(n.String(), fmt.Sprintf("%d", common.APIServerInternalPort)), nil
```
aojea, line 187: it took me a while to get this, maybe a comment explaining that we use `hostnames` instead of IPs?

pkg/cluster/internal/providers/docker/provision.go
aojea: I think we have to assume several restarts for the worker nodes until the control-plane node is up, or will they keep trying?
We can't guarantee the order in which the containers are created, right?

pkg/cluster/internal/providers/docker/network.go
aojea: that's a good point, we can differentiate between command errors and a not found network

pkg/cluster/internal/providers/docker/network.go
aojea: can we add a comment?

pkg/cluster/internal/providers/docker/provider.go
BenTheElder: fair 🙃

pkg/cluster/internal/providers/docker/provider.go
aojea: I'm curious about the DNS behavior with IPv6, seems we are going to have some fun :partying_face:

pkg/exec/helpers.go
aojea: yeah, answered above, the current method is better

aojea:
this is great, only some nits to add comments explaining some changes and make the regex to obtain  the network name more restrictive

pkg/cluster/internal/providers/docker/provision.go
BenTheElder: The entrypoint isn't aware of other nodes and ultimately becomes systemd. kubelet restarts over and over w/o exiting the container.

The containers only fail when killed or when the entry point fails prior to systemd (systemd will only exit for a signal).

So this means when a kind node crashes in the entrypoint we try to run it again once ~immediately... And it's just going to fail again.

This makes it slightly racy / slightly longer to when kind realizes the container is dead and fails / rms it unless --retain.

It will only retry *once* unless dockerd is restarted though.

images/base/files/usr/local/bin/entrypoint
BenTheElder: I realized this later but forgot to remove this comment 😅
Thanks

pkg/cluster/internal/providers/docker/network.go
BenTheElder: I'll probably actually fix it :sweat_smile:

BenTheElder:
fixed the typo and the regex (and for completeness escaping since we don't need it)

@neolit123 lol, my sleepy self let github auto fill the PR title from the branch :+)

pkg/cluster/internal/providers/docker/network.go
BenTheElder: fixed

pkg/cluster/internal/providers/docker/network.go
BenTheElder: done, thanks

pkg/cluster/internal/providers/docker/provider.go
BenTheElder: that's true, if embedded DNS misbehaves on IPv6 we may have to use the bridge for IPv6 singlestack :grimacing:

BenTheElder: I have some thoughts about how we'd make that work, but I hope we don't need to and we can deal with podman + ipv6 singlestack later.

pkg/cluster/internal/providers/docker/network.go
BenTheElder: ohhhh ugh, if we want to support simultaneous ipv4 and ipv6 we're going to have to either:
- always enable ipv6 on the kind bridge
- require that users ensure the bridge exists w/ ipv6 before kind
- require that users create an ipv6 cluster first before any ipv4
- use a special network for ipv6 clusters

requiring that the ipv6 one is created first seems sanest?
that or having a `kind-ipv6` network ...

always enabling ipv6 won't work because docker OOTB will cause:
> Error response from daemon: could not find an available, non-overlapping IPv6 address pool among the defaults to assign to the network

due to ipv6 not being configured on the daemon.

BenTheElder: i implemented:
> require that users create an ipv6 cluster first before any ipv4

just to see how far CI gets.

images/base/files/usr/local/bin/entrypoint
```diff
@@ -179,6 +179,15 @@ enable_network_magic(){
...
+  curr_ip=$(hostname --ip-address)
+  if [ -f /kind/old-ip ]; then
+    old_ip=$(cat /kind/old-ip)
+    sed -i "s#${old_ip}#${curr_ip}#" /etc/kubernetes/manifests/*.yaml /var/lib/kubelet/kubeadm-flags.env
+  fi
+  echo -n "${curr_ip}" >/kind/old-ip
```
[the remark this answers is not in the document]
BenTheElder: it contains the old-ip at reboot time, which is the only time we use it, the curr IP should always be discovered from the live interface

pkg/cluster/internal/providers/docker/network.go
aojea: can we just retry and fallback to an ipv4 only bridge?

BenTheElder: not sure what you mean.
ipv6 has to be enabled at network creation time.

the old default one has it IF the daemon has ipv6 configured. but no embedded DNS.

if the user creates a kind cluster with ipv4 then creates ipv6 we'll have created a v4 only bridge.

we could try to check if the daemon has v6 configured when creating v4 clusters and opportunistically enable v6 maybe.

let's see how CI does first.

BenTheElder:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_kind/1508/pull-kind-conformance-parallel-ipv6/1253784670273802243/

> Error response from daemon: could not find an available, non-overlapping IPv6 address pool among the defaults to assign to the network

that's a problem.
we can configure this in the job runner I guess, but that will take a bit to roll out.
I think we can put a quick hack in here first to test ...

pkg/cluster/internal/providers/docker/network.go
aojea:
```go

if err:=  exec.Command("docker", "network", "create", "--ipv6","-d=bridge", name).Run(): err != nil {
 return exec.Command("docker", "network", "create", "-d=bridge", name).Run()
}
return nil
```

BenTheElder:
https://github.com/moby/moby/issues/29443

sigh.

using an IPv6 network *always* requires specifying the subnet for each network, no IPAM.

the default bridge network has a special config for this that you must set. https://github.com/moby/moby/issues/29443#issuecomment-495808871

I'm going to test how far we get if we hack the CI scripts to do:
> require that users ensure the bridge exists w/ ipv6 before kind

aojea:
> [moby/moby#29443](https://github.com/moby/moby/issues/29443)
> 
> sigh.
> 
> using an IPv6 network _always_ requires specifying the subnet for each network, no IPAM.
> 
> the default bridge network has a special config for this that you must set. [moby/moby#29443 (comment)](https://github.com/moby/moby/issues/29443#issuecomment-495808871)
> 
> I'm going to test how far we get if we hack the CI scripts to do:
> 
> > require that users ensure the bridge exists w/ ipv6 before kind

no problem , it's doable and we have lots of room to avoid collisions, 

we can take a random number if necessary

`ocker network create --driver bridge --ipv6 --subnet 2a02:6b8:b010:9020:1::/104 kind`

BenTheElder:
I tossed in the subnet from wrapper.sh for now, technically it's customizable if you just create the network before using kind ;-)

pkg/cluster/internal/providers/docker/network.go
```diff
@@ -50,8 +50,11 @@ func ensureNetwork(name string, ipFamily config.ClusterIPFamily) error {
 	if string(out) == name+"\n" {
 		return nil
 	}
+	// TODO: ipv6 subnet should probably not be fixed
+	// Though maybe just require the user to handle this by creating the network
+	// as they desire before running kind ...
 	if ipFamily == config.IPv6Family {
-		return exec.Command("docker", "network", "create", "-d=bridge", "--ipv6", name).Run()
+		return exec.Command("docker", "network", "create", "-d=bridge", "--ipv6", "--subnet=fc00:db8:1::/64", name).Run()
```
aojea, line 57: we can use a /104 , it means 255 addresses and we have a low risk of collisions based on this https://github.com/kubernetes/kubernetes/pull/90115#issuecomment-616670315

BenTheElder: i'm using a /104 just above this now. i'm not sure this is the value we want to settle on.

BenTheElder: test are running though https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/1508/pull-kind-conformance-parallel-1-16/1253795238921637894

BenTheElder:
Subpath flake
/retest

BenTheElder:
IPv6 green.
Not sure if reboot will work yet though.. haven't tested. I'd bet not quite.

BenTheElder:
that flake is similar to the ones we were worried about in the pause pr ...
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/1508/pull-kind-e2e-kubernetes/1253795238921637893
/retest

BenTheElder:
/hold cancel
IPv6 is not perfect but this is fairly reasonable now.

amwat:
FYI i tested podman ipv6 and it wasn't working before this PR as well, so no regression there.

BenTheElder:
this does appear to fix https://github.com/kubernetes-sigs/kind/pull/1508, i tested with a host using iptables_nft

one outstanding issue I found that @amwat confirmed is this:
```console
$ kubectl get po -A
NAMESPACE            NAME                                         READY   STATUS         RESTARTS   AGE
kube-system          coredns-66bff467f8-fxzqz                     1/1     Running        1          14m
kube-system          coredns-66bff467f8-kg9sf                     1/1     Running        1          14m
kube-system          etcd-kind-control-plane                      1/1     Running        1          14m
kube-system          kindnet-mr62p                                1/1     Running        1          14m
kube-system          kube-apiserver-kind-control-plane            1/1     Running        1          14m
kube-system          kube-controller-manager-kind-control-plane   1/1     Running        1          14m
kube-system          kube-proxy-gpp6v                             1/1     Running        1          14m
kube-system          kube-scheduler-kind-control-plane            1/1     Running        1          14m
local-path-storage   local-path-provisioner-774f7f8fdb-j8mmd      0/1     NodeAffinity   0          14m
local-path-storage   local-path-provisioner-774f7f8fdb-wxp7w      1/1     Running        0          9s
```

... a stale `local-path-storage` pod will be left around with broken NodeAffinity a new one seems to come up healthy though.

haven't debugged it further yet. this is after one physical host reboot.

pkg/cluster/internal/providers/podman/provider.go
```diff
@@ -195,6 +195,26 @@ func (p *Provider) GetAPIServerEndpoint(cluster string) (string, error) {
...
+	allNodes, err := p.ListNodes(cluster)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to list nodes")
+	}
+	n, err := nodeutils.APIServerEndpointNode(allNodes)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to get apiserver endpoint")
+	}
+	// TODO: check cluster IP family and return the correct IP
+	// This means IPv6 singlestack is broken on podman
```
BenTheElder, line 210: noting that per @amwat this never worked, leaving the TODO for the future though.

aojea:
/test all
before final lgtm

images/base/files/usr/local/bin/entrypoint
BenTheElder: updated

pkg/build/nodeimage/const_storage.go
```diff
@@ -87,7 +87,7 @@ spec:
         app: local-path-provisioner
     spec:
       nodeSelector:
-        node-role.kubernetes.io/master: ''
```
BenTheElder, line 90: this label is deprecated, however we do want to only run on linux nodes for our friends doing windows join hijinks.

pkg/cluster/internal/providers/docker/network.go
BenTheElder: this will report odd errors when the user was actually trying to do ipv6.
I will think on this some more and follow up

BenTheElder:
I think the local-path-storage issue should be [resoolllved](https://www.youtube.com/watch?v=T6b83-eLEWc) now.

BenTheElder:
:shipit: 
will follow up regarding IPv6 improvements

aojea:
/lgtm
fantastic

Issue #1410, opened 2020-03-16
Create cluster fails without apparent information outside debug mode


khos2ow:
**What happened**:

I was trying to spin up a kind cluster with `kind create cluster` which kept failing over and over with the following error:

```
$ kind create cluster
Creating cluster "kind" ...                        
 ✓ Ensuring node image (kindest/node:v1.17.0) 🖼
 ✗ Preparing nodes 📦
docker run error: command "docker run --hostname kind-control-plane --name kind-control-plane --label io.x-k8s.kind.role=control-plane --privileged --security-opt seccomp=unconfined --security-opt apparmor=unconfined --tmpfs /tmp --tmpfs /run --volume /var --volume /lib/modules:/lib/modules:ro --detach --tty --label io.x-k8s.kind.cluster=kind -e '<redacted>' --publish=127.0.0.1:0:6443/TCP kindest/node:v1.17.0@sha256:9512edae126da271b66b990b6fff768fbb7cd786c7d39e86bdf55906352fdf62" failed with error: exit status 125  
```

It was only when I ran with `-v 1` that I noticed I had forgotten to remove the previous cluster which I had spun up.

```
Output:
docker: Error response from daemon: Conflict. The container name "/kind-control-plane" is already in use by container "9df1a7a82171e3d329be23b8f63a1bd24f075e54f7a7d1591f94f4610293df3b". You have to remove (or re
name) that container to be able to reuse that name.
See 'docker run --help'.
```

**What you expected to happen**:

~~I'm not exactly sure if this has been discussed before or not, but I feel like this specific error (or some other similar in nature) can be caught and bubbled up in normal execution rather than only exposing them in debug mode.~~

**How to reproduce it (as minimally and precisely as possible)**:

- `kind-0.6.1 create cluster`
- let it spin up completely
- leave it alone running
- `kind-0.7.0 create cluster`
- the error above which only visible with `-v 1`

**Anything else we need to know?**:

Note that after a little bit of investigation to write up this issue I've figured out that this **only** happens if the existing container is of a different version than the cluster you're trying to spin up now. For example spin up with `v0.6.1` and the try again later with `v0.7.0`. Now I'm not sure if it made sense for this to be fixed within the code rather than a specific documentation for it (if there isn't any already)

**Environment:**

- kind version: (use `kind version`): `kind v0.7.0 go1.13.6 linux/amd64`
- Kubernetes version: (use `kubectl version`): `v1.16.4`
[...]

PR #1510, opened 2020-04-25
Better default info

BenTheElder:
fixes: https://github.com/kubernetes-sigs/kind/issues/1410

we still guard the stacktrace behind at least -v 1. possibly higher in the future...?

failed command output is *critical* to understanding what went wrong in many cases, we shouldn't hide this by default.

### DIFF
--- a/pkg/build/nodeimage/internal/container/docker/pull.go
+++ b/pkg/build/nodeimage/internal/container/docker/pull.go
@@ -55,8 +55,5 @@ func Pull(logger log.Logger, image string, retries int) error {
 			}
 		}
 	}
-	if err != nil {
-		logger.V(1).Infof("Failed to pull image: %q %v", image, err)
-	}
 	return err
 }
--- a/pkg/cluster/internal/create/create.go
+++ b/pkg/cluster/internal/create/create.go
@@ -100,7 +100,6 @@ func Cluster(logger log.Logger, ctx *context.Context, opts *ClusterOptions) erro
 	// Create node containers implementing defined config Nodes
 	if err := ctx.Provider().Provision(status, ctx.Name(), opts.Config); err != nil {
 		// In case of errors nodes are deleted (except if retain is explicitly set)
-		logger.Errorf("%v", err)
 		if !opts.Retain {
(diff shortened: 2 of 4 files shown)


pkg/cmd/kind/create/cluster/createcluster.go
```diff
@@ -98,12 +98,6 @@ func runE(logger log.Logger, streams cmd.IOStreams, flags *flagpole) error {
 		cluster.CreateWithDisplayUsage(true),
 		cluster.CreateWithDisplaySalutation(true),
 	); err != nil {
-		if errs := errors.Errors(err); errs != nil {
```
BenTheElder, line 101: we have more suitable logging for this in main:
image

pkg/cluster/internal/create/create.go
```diff
@@ -100,7 +100,6 @@ func Cluster(logger log.Logger, ctx *context.Context, opts *ClusterOptions) erro
 	// Create node containers implementing defined config Nodes
 	if err := ctx.Provider().Provision(status, ctx.Name(), opts.Config); err != nil {
 		// In case of errors nodes are deleted (except if retain is explicitly set)
-		logger.Errorf("%v", err)
```
BenTheElder, line 103: we return this error, so we shouldn't log it. it will get logged up the stack.

pkg/build/nodeimage/internal/container/docker/pull.go
```diff
@@ -55,8 +55,5 @@ func Pull(logger log.Logger, image string, retries int) error {
 			}
 		}
 	}
-	if err != nil {
-		logger.V(1).Infof("Failed to pull image: %q %v", image, err)
```
BenTheElder, line 59: we return this error and log it elsewhere up the stack.

BenTheElder:
example of why we need this:
image

BenTheElder:
[k8s / bazel build flake](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_kind/1510/pull-kind-conformance-parallel-1-13/1253965358142853125/)
/retest

PR #1512, opened 2020-04-25
Use autodiscovery addresses

aojea:
Fixes https://github.com/kubernetes-sigs/kind/issues/1154

PR #1513, opened 2020-04-25
Use node ip autodiscovery

### LINKED ISSUE #1154: Add new k8s IPv6 configuration improvements to KIND

aojea:
Alternative to #1512

 Fixes #1154

### DIFF
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -180,6 +180,10 @@ enable_network_magic(){
   cp /etc/resolv.conf /etc/resolv.conf.original
   sed -e "s/${docker_embedded_dns_ip}/${docker_host_ip}/g" /etc/resolv.conf.original >/etc/resolv.conf
 
+  # use dns to resolve hostnames with more priority than files
+  cp /etc/nsswitch.conf /etc/nsswitch.conf.original
+  sed -e "s/^hosts:.*/hosts:          dns files/" /etc/nsswitch.conf.original >/etc/nsswitch.conf
+
   # fixup IPs in manifests ...
   # TODO: ensure this supports rebooting the host w/ IPv6 clusters
   curr_ip=$(hostname --ip-address)
--- a/pkg/build/nodeimage/defaults.go
+++ b/pkg/build/nodeimage/defaults.go
@@ -20,7 +20,7 @@ package nodeimage
 const DefaultImage = "kindest/node:latest"
 
 // DefaultBaseImage is the default base image used
-const DefaultBaseImage = "kindest/base:v20200423-30be2258"
+const DefaultBaseImage = "aojea/kindbase:auto"
 
 // DefaultMode is the default kubernetes build mode for the built image
 // see pkg/build/kube.Bits
--- a/pkg/cluster/internal/kubeadm/config.go
+++ b/pkg/cluster/internal/kubeadm/config.go
@@ -535,6 +535,15 @@ func Config(data ConfigData) (config string, err error) {
 		data.FeatureGates = make(map[string]bool)
 	}
 
+	if ver.AtLeast(version.MustParseSemantic("v1.18.0")) {
+		// Use IP autodiscovery
+		if data.IPv6 {
+			data.NodeAddress = "::"
+		} else {
+			data.NodeAddress = ""
+		}
+	}
+
 	// assume the latest API version, then fallback if the k8s version is too low
 	templateSource := ConfigTemplateBetaV2
 	if ver.LessThan(version.MustParseSemantic("v1.12.0")) {


aojea:
/hold
still exploring different options, there are things doesn't work yet

aojea:
the apiserver autodetection is working

```
I0425 21:38:14.703573     152 interface.go:262] Found valid IPv6 address fc00:db8:2::4 for interface "eth0".
I0425 21:38:14.703577     152 interface.go:411] Found active IP fc00:db8:2::4
```

kubelet is failing because of https://kubernetes.slack.com/archives/DGW6DCL23/p1587850042058100

aojea:
only resolve IPv6 addresses non-local nodes 🤔 

```
root@kind-worker:/# ./lookup kind-worker
GOOS: linux
net.LookupIP addrs: [172.19.0.3]
root@kind-worker:/# ./lookup kind-control-plane
GOOS: linux
net.LookupIP addrs: [172.19.0.4 fc00:db8:2::4]
root@kind-worker:/# ./lookup kind-worker2
GOOS: linux
net.LookupIP addrs: [172.19.0.2 fc00:db8:2::2]
```

images/base/files/usr/local/bin/entrypoint
```diff
@@ -180,6 +180,10 @@ enable_network_magic(){
   cp /etc/resolv.conf /etc/resolv.conf.original
   sed -e "s/${docker_embedded_dns_ip}/${docker_host_ip}/g" /etc/resolv.conf.original >/etc/resolv.conf
 
+  # use dns to resolve hostnames with more priority than files
+  cp /etc/nsswitch.conf /etc/nsswitch.conf.original
+  sed -e "s/^hosts:.*/hosts:          dns files/" /etc/nsswitch.conf.original >/etc/nsswitch.conf
```
aojea, line 185: @BenTheElder how can we test this in the CI?

aojea: ok, I think I know

aojea:
/assign @BenTheElder 
please take a look, autodiscovery works with this PR

pkg/build/nodeimage/defaults.go
```diff
@@ -20,7 +20,7 @@ package nodeimage
 const DefaultImage = "kindest/node:latest"
 
 // DefaultBaseImage is the default base image used
-const DefaultBaseImage = "kindest/base:v20200423-30be2258"
+const DefaultBaseImage = "aojea/kindbase:auto"
```
aojea, line 23: this is built with current PR, it's being added on a temporal commit just to test it works on CI

BenTheElder: I'll rebase and push an image after reviewing the other changes 👍

images/base/files/usr/local/bin/entrypoint
```diff
@@ -180,6 +180,10 @@ enable_network_magic(){
   cp /etc/resolv.conf /etc/resolv.conf.original
   sed -e "s/${docker_embedded_dns_ip}/${docker_host_ip}/g" /etc/resolv.conf.original >/etc/resolv.conf
 
+  # use dns to resolve hostnames with more priority than files
+  cp /etc/nsswitch.conf /etc/nsswitch.conf.original
```
BenTheElder, line 184: what was the original value?
why aren't we statically rewriting this in the baseimage build?

BenTheElder: debian *should* be `hosts: dns [!UNAVAIL=return] files` I think, which is actually a little surprising, I thought most distros were shipping ~`files dns`

https://github.com/golang/go/issues/35305
https://manpages.debian.org/stretch/manpages/nsswitch.conf.5.en.html

BenTheElder:
rewrote the second commit and pushed a new PR https://github.com/kubernetes-sigs/kind/pull/1518

images/base/files/usr/local/bin/entrypoint
aojea: You are correct, the original was **files dns**

aojea: I didn't think about the original bare honestly, added to the entry point because is where we are adding these hacks

BenTheElder:
superseded https://github.com/kubernetes-sigs/kind/pull/1521

PR #1518, opened 2020-04-27
Use node ip autodiscovery [forked]

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/1513 but with the base image cleaned up.

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -108,7 +108,9 @@ RUN echo "Ensuring scripts are executable ..." \
  && echo "Ensuring /etc/kubernetes/manifests" \
     && mkdir -p /etc/kubernetes/manifests \
  && echo "Adjusting systemd-tmpfiles timer" \
-    && sed -i /usr/lib/systemd/system/systemd-tmpfiles-clean.timer -e 's#OnBootSec=.*#OnBootSec=1min#'
+    && sed -i /usr/lib/systemd/system/systemd-tmpfiles-clean.timer -e 's#OnBootSec=.*#OnBootSec=1min#' \
+ && echo "Modifying /etc/nsswitch.conf to prefer hosts" \
+    && sed -i /etc/nsswitch.conf -re 's#^(hosts:\s*).*#\1dns files#'
 
 # tell systemd that it is in docker (it will check for the container env)
 # https://www.freedesktop.org/wiki/Software/systemd/ContainerInterface/
--- a/pkg/build/nodeimage/defaults.go
+++ b/pkg/build/nodeimage/defaults.go
@@ -20,7 +20,7 @@ package nodeimage
 const DefaultImage = "kindest/node:latest"
 
 // DefaultBaseImage is the default base image used
-const DefaultBaseImage = "kindest/base:v20200423-30be2258"
+const DefaultBaseImage = "kindest/base:v20200426-fa727959"
 
 // DefaultMode is the default kubernetes build mode for the built image
 // see pkg/build/kube.Bits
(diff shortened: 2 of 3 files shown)


images/base/Dockerfile
```diff
@@ -108,7 +108,9 @@ RUN echo "Ensuring scripts are executable ..." \
  && echo "Ensuring /etc/kubernetes/manifests" \
     && mkdir -p /etc/kubernetes/manifests \
  && echo "Adjusting systemd-tmpfiles timer" \
-    && sed -i /usr/lib/systemd/system/systemd-tmpfiles-clean.timer -e 's#OnBootSec=.*#OnBootSec=1min#'
+    && sed -i /usr/lib/systemd/system/systemd-tmpfiles-clean.timer -e 's#OnBootSec=.*#OnBootSec=1min#' \
+ && echo "Modifying /etc/nsswitch.conf to prefer hosts" \
+    && sed -i /etc/nsswitch.conf -re 's#^(hosts:\s*).*#\1dns files#'
```
BenTheElder, line 113: preserve the indent and do this at build time instead of runtime

aojea:
Could you check if reboot works?

images/base/Dockerfile
aojea: This seems better indeed

BenTheElder:
I didn't yet, and I'm currently booted to windows on my desktop 😱 /s (no IPv6 clusters on mac...)

aojea:
no problem, I'm going to do it now

aojea:
Before restart
```
ubectl get nodes -o wide
NAME                 STATUS   ROLES    AGE     VERSION                                    INTERNAL-IP   EXTERNAL-IP   OS-IMAG
E       KERNEL-VERSION                CONTAINER-RUNTIME
kind-control-plane   Ready    master   3m54s   v1.19.0-alpha.2.172+ce65f0e7ddeb60-dirty   172.19.0.2    <none>        Ubuntu 
19.10   4.12.14-lp151.28.13-default   containerd://1.3.3-14-g449e9269
kind-worker          Ready    <none>   3m19s   v1.19.0-alpha.2.172+ce65f0e7ddeb60-dirty   172.19.0.3    <none>        Ubuntu 
19.10   4.12.14-lp151.28.13-default   containerd://1.3.3-14-g449e9269
kind-worker2         Ready    <none>   3m19s   v1.19.0-alpha.2.172+ce65f0e7ddeb60-dirty   172.19.0.4    <none>        Ubuntu 
19.10   4.12.14-lp151.28.13-default   containerd://1.3.3-14-g449e9269

```

After only control-plane goes up, the workers fail with the following error

```
sed: can't read /etc/kubernetes/manifests/*.yaml: No such file or directory
```

BenTheElder:
I can fix that in the morning I think.

aojea:
> I can fix that in the morning I think.

don't worry, I'll send a patch

aojea:
/hold

This has a side effect, now we have two ip addresses

```
root@kind-control-plane:/# hostname --ip-address
fc00:db8:2::3 172.19.0.3
```

aojea:
and it always return ipv6 first, I just checked in an IPv4 node
```
docker exec -it tewst-control-plane hostname --ip-address
fc00:db8:2::5 172.19.0.5
```

aojea:
it does work with ipv6 :partying_face:

BenTheElder:
Can we pass kubelet 0.0.0.0 for ipv4 discovery?

aojea:
> Can we pass kubelet 0.0.0.0 for ipv4 discovery?

the option was implemented only for `::`

I think is better to pass the `empty` value, I think that the option has `omitempty`

BenTheElder:
superseded https://github.com/kubernetes-sigs/kind/pull/1521

PR #1519, opened 2020-04-27
add a flag for the path in export logs

BenTheElder:
`kind export logs` is currently _the only_ CLI command accepting an argument, and it only accepts this one argument. All other commands require zero arguments.

This PR creates a migration path off of that argument by adding a flag to control it instead. 
We can explore supporting arguments for `--name` in v0.9.0, which is more intuitive / ecosystem compatible in retrospect. `minikube`'s `--profile` was probably not desirable to mimic.

i.e. we can enable `kind get cluster kind2` <-> `kubectl get po baz` in a PR after v0.8.0 is out (so v0.8.0 only warns about the argument but still supports it and provides time to migrate).

### DIFF
--- a/pkg/cmd/kind/export/logs/logs.go
+++ b/pkg/cmd/kind/export/logs/logs.go
@@ -18,7 +18,9 @@ limitations under the License.
 package logs
 
 import (
+	"errors"
 	"fmt"
+	"os"
 
 	"github.com/spf13/cobra"
 
@@ -32,7 +34,8 @@ import (
 )
 
 type flagpole struct {
-	Name string
+	Name      string
+	Directory string
 }
 
 // NewCommand returns a new cobra.Command for getting the cluster logs
@@ -46,10 +49,15 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 		Long:  "Exports logs to a tempdir or [output-dir] if specified",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cli.OverrideDefaultName(cmd.Flags())
+			// TODO: remove this once we no longer support the arg
+			if !cmd.Flags().Lookup("directory").Value.Set() {
+				flags.Dir = ""
+			}
 			return runE(logger, streams, flags, args)
 		},
 	}
 	cmd.Flags().StringVar(&flags.Name, "name", cluster.DefaultName, "the cluster context name")
+	cmd.Flags().StringVarP(&flags.Direcory, "directory", "C", "", "change to directory dir before exporting")
 	return cmd
 }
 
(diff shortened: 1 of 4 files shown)


pkg/cmd/kind/root.go
```diff
@@ -111,11 +111,7 @@ func runE(logger log.Logger, flags *flagpole, command *cobra.Command) error {
 	maybeSetVerbosity(logger, log.Level(flags.Verbosity))
 	// warn about deprecated flag if used
 	if setLogLevel {
-		if cmd.ColorEnabled(logger) {
-			logger.Warn("\x1b[93mWARNING\x1b[0m: --loglevel is deprecated, please switch to -v and -q!")
-		} else {
-			logger.Warn("WARNING: --loglevel is deprecated, please switch to -v and -q!")
-		}
+		cmd.FancyWarn(logger, "--loglevel is deprecated, please switch to -v and -q!")
```
BenTheElder, line 114: not related, just DRYing this out.

pkg/cmd/kind/export/logs/logs.go
```diff
@@ -70,14 +73,26 @@ func runE(logger log.Logger, streams cmd.IOStreams, flags *flagpole, args []stri
 
 	// get the optional directory argument, or create a tempdir
 	var dir string
-	if len(args) == 0 {
+	if flags.Out != "" {
+		// prefer the flag
+		dir = flags.Out
+		if len(args) != 0 {
+			cmd.FancyWarn(logger, "Ignoring argument since preferred --out is set. Please stup using an arugment for the export path!")
```
BenTheElder, line 80: probably error instead?

neolit123: typos in `stup` and `arugment`

i guess, if it's deprecated it should continue to work until the deprecation period passed.

BenTheElder: This would be an error if you set *both*

neolit123: ok, makes sense.

BenTheElder:
there is also `kind load ...` which take one argument (the image name / path).
/hold

I'm going to tweak this slightly, but also taking a stronger look at what the ecosystem is doing these days.

BenTheElder:
I do think we should rethink this tool a bit regardless though, the tempdir thing is a bit strange in retrospect.

pkg/cmd/kind/export/logs/logs.go
BenTheElder: instead it's all warnings with a preference chain. I updated it slightly.

BenTheElder:
Updated:
- `-C` / `--directory` flag inspired by GNU utils since kubernetes tools don't seem to have this concept.
  - https://www.gnu.org/software/make/manual/html_node/Options-Summary.html
  - http://man7.org/linux/man-pages/man1/tar.1.html

Behavior:
- if the flag is set, do a chdir and then export to the current dir, `.` will be the default directory once the migration period is over
- if the argument is set but not the flag, write to the argument and warn about migrating to the flag
- if the flag and argument are not set, continue to use a tempdir but warn about the impending change to `.` as the default and the flag

Should be non-breaking for now (though warnings) with an easy path to switch your arg usage to a flag.

This helps free up options in the future for arguments and brings behavior more in line with other archiving / building tools.

Users of the tempdir mode are probably quite rare, we could add a special alternative flag for them if we find someone unable of using something like `kind export logs -C $(mktemp -d)`.

It is not optimal to dump to a tempdir generally, you will wind up having to copy all of it to somewhere useful.

Not 100% convinced of this change though. Leaving the hold for now for others to inspect. Need to think more about how we handle other commands and try to find more usage in the wild.

BenTheElder:
I never commented here, but `kind load ...` is still an issue.
considering options.

BenTheElder:
I think we want to add a new `kind load image` anyhow that intelligently processes the image arguments looking to see if they should come from docker, podman, or appear to be a tar file path. cc @aojea 

the issue is, do we take this opportunity to move to a more normal kubernetes UX with `$tool $verb $proper-noun ...` and make it `kind load image [cluster] [images...]` in this new command?

we could retrofit `proper-noun` into most of the other commands easily as they don't use arguments.
that leaves this command I think, which this PR already creates a solution for.

what about `KIND_CLUSTER_NAME`?

xref: #1766

aojea:
> We can explore supporting arguments for --name in v0.9.0

do you think `--name` is better as an argument? I see it more like the `namespace` flag in kubectl

`kubectl -n myname get pods`

BenTheElder:
i'm not sure honestly, but I think I'd still prefer to move this (directory) to a flag.

Issue #1475, opened 2020-04-14
Release v0.8.0


BenTheElder:
We've been slow getting this one out, but some gnarly changes are landing.

It's past time to prioritize these and get a release out.

- An earnest attempt at finishing up the PRs and good-enough #148 needs to happen.
- #1471 must be fixed.

Integrated test workflow can be punted.

/assign
I've mostly finished catching up on some administrative stuff, I'm going to start temporarily de-prioritizing things like PR review to actually ship these.

BenTheElder:
needed to fix up https://github.com/kubernetes-sigs/kind/issues/1392, this got a bit more pressing. nearly done with that.

BenTheElder:
eep https://github.com/kubernetes-sigs/kind/issues/1498

BenTheElder:
double eep, newer k8s is broken on iptables-nft which is becoming more common. going to see about fixing.

https://github.com/kubernetes-sigs/kind/issues/1461

BenTheElder:
https://github.com/kubernetes-sigs/kind/issues/1471 is sorted. though we need to follow up on why the containerd upgrade (abandoned) seemingly caused more CI timeouts. holding off on that for now because the iptables-nft issue is alarming and the fix I hope to apply also solves host reboot and most DNS issues.

BenTheElder:
Pr open for https://github.com/kubernetes-sigs/kind/issues/148

BenTheElder:
getting closer on #148, PR more or less does what we want but has some rough spots particularly around ipv6. did some more testing.

BenTheElder:
#148 is nominally resolved, but we need a bit of follow-up with network management and IPv6 clusters in particular.

BenTheElder:
https://github.com/kubernetes-sigs/kind/issues/1520 / ipv6 being worked on

BenTheElder:
https://github.com/kubernetes/kubernetes/issues/90568 today's main distraction. back at it. had a good improptu soundboard / brainstorming session w/ @amwat PRs on the way..

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/1526 solves some of the ipv6 woes

BenTheElder:
#1526 is in, fixing our CI, then #273 + https://github.com/kubernetes-sigs/kind/pull/1527/ / handling ipv6 subnet collisions to avoid failing to create clusters when the user happens to be using our preferred subnet

[...]

PR #1521, opened 2020-04-27
Fixreboot

aojea:


images/base/files/usr/local/bin/entrypoint
```diff
@@ -181,13 +181,28 @@ enable_network_magic(){
   sed -e "s/${docker_embedded_dns_ip}/${docker_host_ip}/g" /etc/resolv.conf.original >/etc/resolv.conf
 
   # fixup IPs in manifests ...
-  # TODO: ensure this supports rebooting the host w/ IPv6 clusters
-  curr_ip=$(hostname --ip-address)
-  if [ -f /kind/old-ip ]; then
-    old_ip=$(cat /kind/old-ip)
-    sed -i "s#${old_ip}#${curr_ip}#" /etc/kubernetes/manifests/*.yaml /var/lib/kubelet/kubeadm-flags.env
+  curr_ipv4="$( (getent ahostsv4 $(hostname) | head -n1 | cut -d' ' -f1) || true)"
```
aojea, line 184: I've tried to put it in an array but is too brittle, this option has more boilerplate code bit is more resilient ... the order of the returned address may change depending on the node configuration, or someone can enable ipv6 addresses later on, who knows :shrug:

aojea:
@BenTheElder please take the last commit , it fixes the reboot on ipv6 and in worker nodes

aojea:
I've tested and it worked to me on Linux, IPv6 and multinode, can't test in mac or windows

images/base/files/usr/local/bin/entrypoint
BenTheElder: if we depended on this for something other than host reboot I'd be pretty unhappy with how much `|| true` is going on here, but given the context this seems reasonable..

BenTheElder:
xref: #148 #1475

BenTheElder:
rebased, droped the old base bump, added a new base bump

BenTheElder:
/lgtm cancel
/approve cancel
/hold
this needs to be reworked, this is not safe when there is not an IPv6 for one.

images/base/files/usr/local/bin/entrypoint
```diff
@@ -181,13 +181,28 @@ enable_network_magic(){
...
+  echo "Current ip $curr_ipv4"
+  if [ -f /kind/old-ipv4 ]; then
+      old_ipv4=$(cat /kind/old-ipv4)
+      # kubernetes manifests are only present on control-plane nodes
+      sed -i "s#${old_ipv4}#${curr_ipv4}#" /etc/kubernetes/manifests/*.yaml || true
+      # this is no longer required with autodiscovery
+      sed -i "s#${old_ipv4}#${curr_ipv4}#" /var/lib/kubelet/kubeadm-flags.env || true
   fi
-  echo -n "${curr_ip}" >/kind/old-ip
+  [[ ! -z "${curr_ipv4}" ]] && echo -n "${curr_ipv4}" >/kind/old-ipv4
```
BenTheElder, line 193: this form is cute but exits. an if statement does not set the exit code.

BenTheElder:
fixing

images/base/files/usr/local/bin/entrypoint
BenTheElder: guarding against this more in another commit

BenTheElder:
pushing a new node image because this change will otherwise break the default :-)

aojea:
Is still on hold , any particular reason?

Issue #1412, opened 2020-03-17
Provide cluster name as part of the Configuration and not as flag


jorgemoralespou:
**What would you like to be added**:
I would like to have the cluster name being specified in the Config object that is provided to Kind rather than specified as a flag.

You would be able to create a cluster this way:
```
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
metadata:
  name: my-cluster
```

Then you can create programatically multiple clusters with any available mechanism, but only changing the definition file, which is stored in VCS:
```
kind create cluster --config=config.yaml
```

**Why is this needed**:
By default, you can only specify the name of the cluster via the `--name` flag. If you want to programatically create multiple clusters for testing purposes, you need to change the invocation arguments, instead of the configuration file. For automation, we want everything stored in VCS and so the name of the cluster should be part of the configuration.
This would also be consistent with how Kubernetes resources work, and since Kind is a tool for creating a k8s cluster, consistency should be desired for the usage.

There's some other places that consistency should be desirable, but those will go to other issues:
- labels (cluster and node labels)
- CLI arguments alignment to kubectl (so that it can be eventually used as a plugin)
- `-o output` for certain commands (e.g. kind export kubeconfig)

BenTheElder:
the downside to this is that you're going to need to specify `--name` for any other command, how will you delete these clusters?

BenTheElder:
> There's some other places that consistency should be desirable, but those will go to other issues:
> labels (cluster and node labels)

hmm? these are internal details and have changed before. please do NOT use these!

> CLI arguments alignment to kubectl (so that it can be eventually used as a plugin)

I don't think "be a kubectl plugin" is really a goal for us right now and i'm not really sure how helpful that is, but I admit i'm curious about what this means.

> -o output for certain commands (e.g. kind export kubeconfig)

?? we have `--kubeconfig` and `KUBECONFIG` to control where the output goes, this is 1:1 with kubectl ...

jorgemoralespou:
> the downside to this is that you're going to need to specify `--name` for any other command, how will you delete these clusters?

One of the things that I'm finding quite interesting lately in the k8s ecosystem is the lack of consistency when developing a solution. This comment doesn't try to be targeted at kind, but it can be applied to many projects, it just happens to be posted here.

[...]

PR #1525, opened 2020-04-29
name in config

BenTheElder:
This was supposed to be a small PR to enable #1412.

On the upside it's not much new code, and we delete a bunch of junk.

What actually happened in order to make this sane:
- drop v1alpha3 config
- eliminate the internal cluster / provider "context" thing (this had it's own name defaulting...)
- move a few things around
  - ListInternalNodes to nodeutils
  - Some of create cluster error handling on nodes already existing / logging => inside create.Cluster impl
- refactor to just pass a config object in more places instead of (name, config)

TODO:
- [ ] support config in `kind delete cluster`
- [ ] document
- [x] add to config testdata

/hold


pkg/cluster/provider.go
```diff
@@ -36,6 +36,14 @@ import (
 // DefaultName is the default cluster name
 const DefaultName = constants.DefaultClusterName
 
+// defaultName is a helper that given a neame defaults it if unset
```
aojea, line 39: typo s/neame/name/

BenTheElder: fixed

BenTheElder:
I think we can actually punt documenting for later when this is more settled, but this refactor cleans / unlocks the ability to clean up a LOT so i'm going to rebase it and go ahead

BenTheElder:
It's that kubectl test flake.
/retest

PR #1526, opened 2020-04-29
always create network with ipv6

BenTheElder:
/shrug
turns out "enabling ipv6" is only for the default bridge ...?
so we can just always create the kind network with ipv6 enabled.

### DIFF
--- a/pkg/cluster/internal/providers/docker/network.go
+++ b/pkg/cluster/internal/providers/docker/network.go
@@ -20,8 +20,6 @@ import (
 	"regexp"
 
 	"sigs.k8s.io/kind/pkg/exec"
-
-	"sigs.k8s.io/kind/pkg/internal/apis/config"
 )
 
 // TODO: we'll probably allow configuring this
@@ -35,7 +33,7 @@ import (
 const fixedNetworkName = "kind"
 
 // ensureNetwork checks if docker network by name exists, if not it creates it
-func ensureNetwork(name string, ipFamily config.ClusterIPFamily) error {
+func ensureNetwork(name string) error {
 	// TODO: the network might already exist and not have ipv6 ... :|
 	// discussion: https://github.com/kubernetes-sigs/kind/pull/1508#discussion_r414594198
 	out, err := exec.Output(exec.Command(
@@ -53,8 +51,5 @@ func ensureNetwork(name string, ipFamily config.ClusterIPFamily) error {
 	// TODO: ipv6 subnet should probably not be fixed
 	// Though maybe just require the user to handle this by creating the network
 	// as they desire before running kind ...
-	if ipFamily == config.IPv6Family {
-		return exec.Command("docker", "network", "create", "-d=bridge", "--ipv6", "--subnet=fc00:db8:2::/64", name).Run()
-	}
-	return exec.Command("docker", "network", "create", "-d=bridge", name).Run()
+	return exec.Command("docker", "network", "create", "-d=bridge", "--ipv6", "--subnet=fc00:db8:2::/64", name).Run()
 }
--- a/pkg/cluster/internal/providers/docker/provider.go
+++ b/pkg/cluster/internal/providers/docker/provider.go
@@ -58,7 +58,7 @@ func (p *Provider) Provision(status *cli.Status, cluster string, cfg *config.Clu
 		return err
 	}
 
-	if err := ensureNetwork(fixedNetworkName, cfg.Networking.IPFamily); err != nil {
+	if err := ensureNetwork(fixedNetworkName); err != nil {
 		return errors.Wrap(err, "failed to ensure docker network")
 	}
 


BenTheElder:
FYI @amwat lol :joy:

BenTheElder:
/hold
checking on some other systems

pkg/cluster/internal/providers/docker/network.go
```diff
@@ -53,8 +51,5 @@ func ensureNetwork(name string, ipFamily config.ClusterIPFamily) error {
 	// TODO: ipv6 subnet should probably not be fixed
 	// Though maybe just require the user to handle this by creating the network
 	// as they desire before running kind ...
-	if ipFamily == config.IPv6Family {
-		return exec.Command("docker", "network", "create", "-d=bridge", "--ipv6", "--subnet=fc00:db8:2::/64", name).Run()
-	}
-	return exec.Command("docker", "network", "create", "-d=bridge", name).Run()
+	return exec.Command("docker", "network", "create", "-d=bridge", "--ipv6", "--subnet=fc00:db8:2::/64", name).Run()
```
aojea, line 54: we can't create the same range in all networks, we should add some bytes hashing the name or something

BenTheElder:
sanity checked that this PR works fine on docker desktop (macOS), ubuntu 20.04 / docker-ce, neither with ipv6 """enabled""".

pkg/cluster/internal/providers/docker/network.go
BenTheElder: that is true, however this is not new to this PR, and currently there is a single caller with a fixed value. there's an existing TODO here, I'm limiting this PR to *just* resolving the aspect of "ipv4 clusters before ipv6 clusters => bad state"

BenTheElder:
we'll want to update the docs after the release, you don't need to muck with daemon config to do ipv6 anymore :tada: 

we can also consider removing that hack from KRTE

pkg/cluster/internal/providers/docker/network.go
aojea: ups, I forgot that was hardcoded now XD

aojea: what happens if the host does not have ipv6?
the command will fail?

BenTheElder: https://github.com/kubernetes-sigs/kind/pull/1526#issuecomment-621084660

BenTheElder:
/override pull-kind-conformance-parallel-1-12
/override pull-kind-conformance-parallel-1-13
/override pull-kind-conformance-parallel-1-14
will fix CI in a bit.

pkg/cluster/internal/providers/docker/network.go
aojea: bad wording, not that IPv6, the host IPv6
`net.ipv6.conf.all.disable_ipv6 = 1`

BenTheElder: it still works but i'm not sure that setting is sufficient to test e.g. lack of modules

BenTheElder:
```console
$ sudo sysctl net.ipv6.conf.all.disable_ipv6
net.ipv6.conf.all.disable_ipv6 = 1

$ docker exec kind-control-plane ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: veth94b73da2@if3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default 
    link/ether 82:b9:38:07:5d:5b brd ff:ff:ff:ff:ff:ff link-netns cni-7e2721d4-185b-b646-b43f-41f797ddfa65
    inet 10.244.0.1/32 brd 10.244.0.1 scope global veth94b73da2
       valid_lft forever preferred_lft forever
3: vetheb4ad147@if3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default 
    link/ether 4e:c3:91:c2:4a:c0 brd ff:ff:ff:ff:ff:ff link-netns cni-ad9efa8f-e278-bed2-a72d-f634c116ede3
    inet 10.244.0.1/32 brd 10.244.0.1 scope global vetheb4ad147
       valid_lft forever preferred_lft forever
4: veth39517578@if3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default 
    link/ether 02:c6:7a:67:8c:b0 brd ff:ff:ff:ff:ff:ff link-netns cni-fbf3ca43-0688-1ff0-09ba-c90621b6407e
    inet 10.244.0.1/32 brd 10.244.0.1 scope global veth39517578
       valid_lft forever preferred_lft forever
17: eth0@if18: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default 
    link/ether 02:42:ac:12:00:02 brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet 172.18.0.2/16 brd 172.18.255.255 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fc00:db8:2::2/64 scope global nodad 
       valid_lft forever preferred_lft forever
    inet6 fe80::42:acff:fe12:2/64 scope link 
       valid_lft forever preferred_lft forever

```

aojea: that's fair enough, I think that we can move forward with this :)

BenTheElder: i'm definitely nervous about that, maybe I can fire up a COS GKE node and see if it still works

amwat:
hmm are we sure?
atleast their docs say you still need to enable it on the daemon before creating any networks with ipv6
https://docs.docker.com/network/bridge/#use-ipv6

BenTheElder:
```
$ ps aux | grep docker
root       12194  0.0  0.2 1824648 96092 ?       Ssl  11:32   0:07 /usr/bin/dockerd -H fd:// --containerd=/run/containerd/containerd.sock
root       13538  0.0  0.0 548940  3572 ?        Sl   11:33   0:00 /usr/bin/docker-proxy -proto tcp -host-ip 127.0.0.1 -host-port 35601 -container-ip 172.18.0.2 -container-port 6443
root       13545  0.0  0.0 109104  6564 ?        Sl   11:33   0:00 containerd-shim -namespace moby -workdir /var/lib/containerd/io.containerd.runtime.v1.linux/moby/0705010047ec8100ec5ac99f6a5d0a24e389234cd4c5af275961986aaca92bf5 -address /run/containerd/containerd.sock -containerd-binary /usr/bin/containerd -runtime-root /var/run/docker/runtime-runc
benthee+   51134  0.0  0.0   9032  2604 pts/1    S+   14:48   0:00 grep --color=auto docker


$ sudo cat /etc/docker/daemon.json
[sudo] password for bentheelder: 
cat: /etc/docker/daemon.json: No such file or directory

```

BenTheElder:
i'm having difficulty proving it, but AFAICT docker ipv6 daemon options are just for the default bridge. (`fixed-cidr` definitely is)

aojea:
> 'm having difficulty proving it, but AFAICT docker ipv6 daemon options are just for the default bridge. (`fixed-cidr` definitely is)
yeah, I think that they've moved to libnetwork the new networking and those are legacy options, but is just my impression

BenTheElder:
I brought up a multi-node ipv6 cluster on a machine with default docker (no config file).

BenTheElder:
it worked on a COS node (1.14.10-gke.27), though it appears the ip6tables modules are loaded on my clean cluster :shrug:

marcosnils:
Hey there, 

This just broke my kind deployment since I have ipv6 disable from `ipv6.disable=1` in grub config and now I can't create clusters anymore :(

BenTheElder:
I tested this with ipv6 disable=1 (see discussion above), can you share more details about your environment?

BenTheElder:
As an immediate workaround you can:
- `docker network rm kind && docker network create kind` (skip the other flags)

BenTheElder:
kind will only create the network if it does not exist.

aojea:
> I tested this with ipv6 disable=1 (see discussion above), can you share more details about your environment?

Is because it has disable it in grub  .... I'm at mobile now but I think we should to probe ipv6 or try to create the network in ipv6 and fall back to ipv4

BenTheElder:
i'll send a patch and we can cut an 0.8.1 ...

marcosnils:
> Is because it has disable it in grub .... I'm at mobile now but I think we should to probe ipv6 or try to create the network in ipv6 and fall back to ipv4

^ this. ipv6 is disabled at kernel level and not sysctl. I liked the previous way where you could configure your ipFamily to `ipv4` in config.yaml and kind wouldn't try to create an ipv6 network. Can we bring that back?

marcosnils:
Created https://github.com/kubernetes-sigs/kind/issues/1544 to keep track.

BenTheElder:
> ^ this. ipv6 is disabled at kernel level and not sysctl. I liked the previous way where you could configure your ipFamily to ipv4 in config.yaml and kind wouldn't try to create an ipv6 network. Can we bring that back?.

:facepalm: 

thanks, I think we can come up with something equivilant. the tricky part is that the dockerd may not be running on the same machine as `kind`...

BenTheElder:
> Btw, thanks for the almost immediate reply and for the amazing project :)

:upside_down_face: thanks

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/1545

PR #1527, opened 2020-04-29
generate ipv6 subnet based on the network name

aojea:
Following up on IPv6 changes

### DIFF
--- a/pkg/cluster/internal/providers/docker/network.go
+++ b/pkg/cluster/internal/providers/docker/network.go
@@ -17,6 +17,8 @@ limitations under the License.
 package docker
 
 import (
+	"crypto/sha1"
+	"net"
 	"regexp"
 
 	"sigs.k8s.io/kind/pkg/exec"
@@ -48,8 +50,28 @@ func ensureNetwork(name string) error {
 	if string(out) == name+"\n" {
 		return nil
 	}
-	// TODO: ipv6 subnet should probably not be fixed
-	// Though maybe just require the user to handle this by creating the network
-	// as they desire before running kind ...
-	return exec.Command("docker", "network", "create", "-d=bridge", "--ipv6", "--subnet=fc00:db8:2::/64", name).Run()
+
+	// generate unique subnet per network based on the name
+	// obtained from the ULA fc00::/8 range
+	subnet := generateULASubnetFromName(name)
+
+	return exec.Command("docker", "network", "create", "-d=bridge", "--ipv6", "--subnet", subnet, name).Run()
+}
+
+// generateULASubnetFromName generate an IPv6 subnet based on the name passed as parameter
+func generateULASubnetFromName(name string) string {
+	ip := make([]byte, 16)
+	ip[0] = 0xfc
+	ip[1] = 0x00
+	h := sha1.New()
+	h.Write([]byte(name))
+	bs := h.Sum(nil)
+	for i := 2; i < 8; i++ {
+		ip[i] = bs[i]
+	}
+	subnet := &net.IPNet{
+		IP:   net.IP(ip),
+		Mask: net.CIDRMask(64, 128),
+	}
+	return subnet.String()
 }
--- a/pkg/cluster/internal/providers/docker/network.go
+++ b/pkg/cluster/internal/providers/docker/network.go
@@ -20,8 +20,6 @@ import (
 	"regexp"
 
 	"sigs.k8s.io/kind/pkg/exec"
-
-	"sigs.k8s.io/kind/pkg/internal/apis/config"
 )
 
 // TODO: we'll probably allow configuring this
@@ -35,7 +33,7 @@ import (
 const fixedNetworkName = "kind"
 
 // ensureNetwork checks if docker network by name exists, if not it creates it
-func ensureNetwork(name string, ipFamily config.ClusterIPFamily) error {
+func ensureNetwork(name string) error {
 	// TODO: the network might already exist and not have ipv6 ... :|
 	// discussion: https://github.com/kubernetes-sigs/kind/pull/1508#discussion_r414594198
 	out, err := exec.Output(exec.Command(
@@ -53,8 +51,5 @@ func ensureNetwork(name string, ipFamily config.ClusterIPFamily) error {
 	// TODO: ipv6 subnet should probably not be fixed
 	// Though maybe just require the user to handle this by creating the network
 	// as they desire before running kind ...
-	if ipFamily == config.IPv6Family {
-		return exec.Command("docker", "network", "create", "-d=bridge", "--ipv6", "--subnet=fc00:db8:2::/64", name).Run()
-	}
-	return exec.Command("docker", "network", "create", "-d=bridge", name).Run()
+	return exec.Command("docker", "network", "create", "-d=bridge", "--ipv6", "--subnet=fc00:db8:2::/64", name).Run()
 }
--- a/pkg/cluster/internal/providers/docker/network_test.go
+++ b/pkg/cluster/internal/providers/docker/network_test.go
@@ -1,0 +1,57 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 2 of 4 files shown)


pkg/cluster/internal/providers/docker/network_test.go
```diff
@@ -0,0 +1,57 @@
...
+			name:   "kin",
+			subnet: "fc00:f804:3187:7254::/64",
+		},
+		{
+			name:   "mysupernetwork",
+			subnet: "fc00:1dd9:697c:314e::/64",
+		},
+	}
+	for _, tc := range cases {
+		tc := tc
```
neolit123, line 47: is the local variable needed here?

pkg/cluster/internal/providers/docker/network.go
```diff
@@ -50,11 +50,28 @@ func ensureNetwork(name string, ipFamily config.ClusterIPFamily) error {
...
-		return exec.Command("docker", "network", "create", "-d=bridge", "--ipv6", "--subnet=fc00:db8:2::/64", name).Run()
+
+	// generate unique subnet per network based on the name
+	// obtained from the ULA fc00::/8 range
+	subnet := generateULASubnetFromName(name)
+
+	return exec.Command("docker", "network", "create", "-d=bridge", "--ipv6", "--subnet", subnet, name).Run()
+}
+
+// generateULASubnetFromName generate an IPvs subnet based on the name passed as parameter
```
neolit123, line 61: typo `IPv6`
i'd move the comments above the call to here and clarify a little more why this is needed.

pkg/cluster/internal/providers/docker/network_test.go
```diff
@@ -0,0 +1,57 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
neolit123, line 2: 2020

site/content/docs/user/quick-start.md
```diff
@@ -366,8 +366,7 @@ kubeadmConfigPatches:
 {{< /codeFromInline >}}
 
 #### IPv6 clusters
-You can run ipv6 only clusters using `kind`, but first you need to enable ipv6 in your docker daemon by editing `/etc/docker/daemon.json` [as described here][docker enable ipv6].
-Ensure you `systemctl restart docker` to pick up the changes.
+You can run ipv6 only clusters using `kind`, but first you need to check that ou have enable ipv6 in your host.
```
neolit123, line 369: `you have enabled`

pkg/cluster/internal/providers/docker/network_test.go
aojea: https://github.com/golang/go/wiki/CommonMistakes#using-reference-to-loop-iterator-variable

neolit123: i don't think it's needed because the iteration happens over a slice of values.

BenTheElder: see the second example under the link above, there's a single loop variable so you need to capture it. normally I leave a comment something like:
```go
tc := tc  // capture variable
```

site/content/docs/user/quick-start.md
```diff
@@ -366,8 +366,7 @@ kubeadmConfigPatches:
 {{< /codeFromInline >}}
 
 #### IPv6 clusters
-You can run ipv6 only clusters using `kind`, but first you need to enable ipv6 in your docker daemon by editing `/etc/docker/daemon.json` [as described here][docker enable ipv6].
```
BenTheElder, line 369: don't edit docs based on an unreleased PR. file a second PR and we'll merge when the release is done.

site/content/docs/user/quick-start.md
```diff
@@ -366,8 +366,7 @@ kubeadmConfigPatches:
...
-Ensure you `systemctl restart docker` to pick up the changes.
+You can run ipv6 only clusters using `kind`, but first you need to check that ou have enabled ipv6 in your host.
```
BenTheElder, line 369: I don't think this comment is complete, for one thing it's the *docker* host (people do use remote docker) and second there's docker desktop.

site/content/docs/user/quick-start.md
aojea: yeah, this was waiting for https://github.com/kubernetes-sigs/kind/pull/1526

aojea:
/hold 
waiting for https://github.com/kubernetes-sigs/kind/pull/1526

site/content/docs/user/quick-start.md
aojea: right

pkg/cluster/internal/providers/docker/network_test.go
neolit123: ok, right this is the go routine in a loop gotcha.
it's required when using t.Parallel() in a sub-test.

BenTheElder: it's also more generally true when capturing either when passing a reference explicitly or when doing a closure.

you can explicitly capture instead for the closure case but this assignment / scoping hack is convenient for all cases :P

site/content/docs/user/quick-start.md
```diff
@@ -366,8 +366,14 @@ kubeadmConfigPatches:
...
-You can run ipv6 only clusters using `kind`, but first you need to enable ipv6 in your docker daemon by editing `/etc/docker/daemon.json` [as described here][docker enable ipv6].
-Ensure you `systemctl restart docker` to pick up the changes.
+You can run IPv6 only clusters using `kind`, if the host that runs the docker containers support IPv6. Nowadays, all Operating Systems use to have IPv6 enabled by deafult, but you can check it in linux with the following command:
+
+```sh
+sudo sysctl net.ipv6.conf.all.disable_ipv6
+net.ipv6.conf.all.disable_ipv6 = 0
+```
+
+If you are using docker in Windows or Mac, you don't have to do anything, because the containers run inside a VM that already has IPv6 enabled.
```
BenTheElder, line 376: IPv6 clusters don't work on mac. You can't port forward.

BenTheElder: Confusingly IPv6 Networks do seem fine? but the port forward will fail.

BenTheElder: AFAIK only linux can run IPv6 clusters even after these changes.

pkg/cluster/internal/providers/docker/network.go
```diff
@@ -50,11 +50,28 @@ func ensureNetwork(name string, ipFamily config.ClusterIPFamily) error {
...
-	// Though maybe just require the user to handle this by creating the network
-	// as they desire before running kind ...
-	if ipFamily == config.IPv6Family {
-		return exec.Command("docker", "network", "create", "-d=bridge", "--ipv6", "--subnet=fc00:db8:2::/64", name).Run()
+
+	// generate unique subnet per network based on the name
+	// obtained from the ULA fc00::/8 range
+	subnet := generateULASubnetFromName(name)
+
+	return exec.Command("docker", "network", "create", "-d=bridge", "--ipv6", "--subnet", subnet, name).Run()
```
BenTheElder, line 58: // TODO: check for failure due to overlapping subnet and "probe" another hash N times

BenTheElder:
1-13 status seems stuck even though it should be gone after https://github.com/kubernetes/test-infra/pull/17443
/override pull-kind-conformance-parallel-1-13

pkg/cluster/internal/providers/docker/network.go
aojea: That's a good exercise, What are the chances to have a collision with different names in a 48 bits hash?

site/content/docs/user/quick-start.md
aojea: I'm confused, the cluster works but with limitations because of the host to container port forwarding?

pkg/cluster/internal/providers/docker/network.go
BenTheElder: the problem is that it's not just us creating networks. I have this about implemented, PR in a sec

aojea:
/close 
superseded by #1531

PR #1534, opened 2020-04-30
Fix the problem of container restart failure

panzhc:
1. When restart docker, the kind container restart failed. Docker logs:
```
INFO: Detected IPv4 address: 172.18.0.2
INFO: Detected old IPv4 address: 172.18.0.2
ERROR: Have an old IPv4 address but no current IPv4 address (!)
```
2. After modifying the file and replacing it, the container starts normally:
```
docker cp entrypoint cluster-control-plane:/usr/local/bin/entrypoint
docker start cluster-control-plane
```

### DIFF
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -187,7 +187,7 @@ enable_network_magic(){
       old_ipv4=$(cat /kind/old-ipv4)
       echo "INFO: Detected old IPv4 address: ${old_ipv4}" >&2
       # sanity check that we have a current address
-      if [[ -n $curr_ipv4 ]]; then
+      if [[ -z $curr_ipv4 ]]; then
         echo "ERROR: Have an old IPv4 address but no current IPv4 address (!)" >&2
         exit 1
       fi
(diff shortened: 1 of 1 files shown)


images/base/files/usr/local/bin/entrypoint
```diff
@@ -187,7 +187,7 @@ enable_network_magic(){
       old_ipv4=$(cat /kind/old-ipv4)
       echo "INFO: Detected old IPv4 address: ${old_ipv4}" >&2
       # sanity check that we have a current address
-      if [[ -n $curr_ipv4 ]]; then
+      if [[ -z $curr_ipv4 ]]; then
```
aojea, line 190: yeah, good catch , we have to test that it does not exist

aojea:
/ok-to-test
/assing @BenTheElder 

Ben , I didn't know that the `-n` test requires  the string to be quoted 
https://tldp.org/LDP/abs/html/comparison-ops.html#STRTEST

images/base/files/usr/local/bin/entrypoint
BenTheElder: thanks!

BenTheElder:
bypassing CI, will follow up with image updates

BenTheElder:
thank you!

PR #1537, opened 2020-04-30
drop stale note

BenTheElder:


BenTheElder:
this was a bad experimental feature, we haven't supported this in many versions.

PR #1538, opened 2020-04-30
experimentally allow custom networks

### LINKED ISSUE #273: Allow configurable docker network for kind cluster nodes

BenTheElder:
fixes #273 
- we allow overriding the network
- it's not officially supported
- currently we do no validation
- it must be done via an experimental env, which comes with a warning

### DIFF
--- a/pkg/cluster/internal/providers/docker/provider.go
+++ b/pkg/cluster/internal/providers/docker/provider.go
@@ -19,6 +19,7 @@ package docker
 import (
 	"fmt"
 	"net"
+	"os"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


BenTheElder:
bazel build flake
/retest

amwat:
missed changing
https://github.com/kubernetes-sigs/kind/blob/1f9ac837b8db20081642f81fcf9d796da1dfa574/pkg/cluster/internal/providers/docker/provision.go#L147-L148

BenTheElder:
done

pkg/cluster/internal/providers/docker/provider.go
```diff
@@ -58,7 +59,14 @@ func (p *Provider) Provision(status *cli.Status, cluster string, cfg *config.Clu
 		return err
 	}
 
-	if err := ensureNetwork(fixedNetworkName); err != nil {
+	// ensure the pre-requesite network exists
+	networkName := fixedNetworkName
+	if n, set := os.LookupEnv("KIND_EXPERIMENTAL_DOCKER_NETWORK"); set {
```
amwat, line 64: optional nit: maybe just `os.getenv()!=""` to also handle deliberate empty string?

BenTheElder: They should not set the env? 🤔

BenTheElder: done

BenTheElder:
re-applying, made trivial change per comments. this is the last PR before the v0.8.0 tag

jieyu:
can we make this a hidden flag, instead of an environment variable? It's not very easy to use environment variable when using the library to create the kind cluster.

BenTheElder:
> can we make this an experimental option in the API, instead of an environment variable? It's not very easy to use environment variable when using the library to create the kind cluster.

experimental options are only exposed via environment variable and are *highly* discouraged.

making it easy would defeat the purpose.

what's your use-case in not using the standard network?

we'd actually prefer to use the default "bridge" network but we can't due to the lack of embedded DNS.

PR #1542, opened 2020-05-01
fix kind-with-registry for v0.8

### LINKED ISSUE #1541: local registry doesn't work anymore with kind v0.8.0
I'm having trouble getting Kind with a local registry to work on Kind v0.8.0. I'm using the instructions here:
https://kind.sigs.k8s.io/docs/user/local-registry/

I've tried running the registry with `--net=kind` (per the v0.8.0 release notes) and that didn't appear to help. 

Are these instructions expected to still work?

BenTheElder:
fixes #1541

### DIFF
--- a/site/static/examples/kind-with-registry.sh
+++ b/site/static/examples/kind-with-registry.sh
@@ -1,9 +1,6 @@
 #!/bin/sh
 set -o errexit
 
-# desired cluster name; default is "kind"
-KIND_CLUSTER_NAME="${KIND_CLUSTER_NAME:-kind}"
-
 # create registry container unless it already exists
 reg_name='kind-registry'
 reg_port='5000'
@@ -15,10 +12,10 @@ if [ "${running}" != 'true' ]; then
 fi
 
 # create a cluster with the local registry enabled in containerd
-cat <<EOF | kind create cluster --name "${KIND_CLUSTER_NAME}" --config=-
+cat <<EOF | kind create cluster --config=-
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
-containerdConfigPatches: 
+containerdConfigPatches:
 - |-
   [plugins."io.containerd.grpc.v1.cri".registry.mirrors."localhost:${reg_port}"]
     endpoint = ["http://${reg_name}:${reg_port}"]
@@ -28,6 +25,6 @@ EOF
 docker network connect "kind" "${reg_name}"
 
 # tell https://tilt.dev to use the registry
-for node in $(kind get nodes --name "${KIND_CLUSTER_NAME}"); do
+for node in $(kind get nodes); do
   kubectl annotate node "${node}" "tilt.dev/registry=localhost:${reg_port}";
 done
--- a/site/static/examples/kind-with-registry.sh
+++ b/site/static/examples/kind-with-registry.sh
@@ -26,3 +26,8 @@ EOF
 
 # connect the registry to the cluster network
 docker network connect "kind" "${reg_name}"
+
+# tell https://tilt.dev to use the registry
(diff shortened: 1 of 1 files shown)


site/static/examples/kind-with-registry.sh
```diff
@@ -13,7 +13,6 @@ if [ "${running}" != 'true' ]; then
     -d --restart=always -p "${reg_port}:5000" --name "${reg_name}" \
     registry:2
 fi
-reg_ip="$(docker inspect -f '{{.NetworkSettings.IPAddress}}' "${reg_name}")"
 
 # create a cluster with the local registry enabled in containerd
 cat <<EOF | kind create cluster --name "${KIND_CLUSTER_NAME}" --config=-
```
BenTheElder, line 18: we could actually stop using the `--name` flag

site/static/examples/kind-with-registry.sh
```diff
@@ -1,9 +1,6 @@
 #!/bin/sh
 set -o errexit
 
-# desired cluster name; default is "kind"
-KIND_CLUSTER_NAME="${KIND_CLUSTER_NAME:-kind}"
```
BenTheElder, line 5: this is handled internally to kind now.

site/static/examples/kind-with-registry.sh
```diff
@@ -13,14 +10,21 @@ if [ "${running}" != 'true' ]; then
...
-    endpoint = ["http://${reg_ip}:${reg_port}"]
+    endpoint = ["http://${reg_name}:${reg_port}"]
 EOF
+
+# connect the registry to the cluster network
+docker network connect "kind" "${reg_name}"
+
+# tell https://tilt.dev to use the registry
+for node in $(kind get nodes); do
+  kubectl annotate node "${node}" "tilt.dev/registry=localhost:${reg_port}";
```
BenTheElder, line 29: @nicks is this documented anywhere?

nicks: nope, I'm hoping to promote a shared standard that is not tilt-specific, this is a hack to unblock people

BenTheElder: Awesome! 👍

nicks: would be equally happy with an annotation like `kind.sigs.k8s.io/registry=localhost:${reg_port}` if you're uncomfortable with the tilt.dev annotation

BenTheElder: that sounds good to me :+1: 
I'm fine with tilt.dev but it probably looks better to use a kind annotation

nicks: added some documentation
https://docs.tilt.dev/choosing_clusters.html#discovering-the-registry
that you can point people to if anyone asks

BenTheElder: i'm going to leave this for now since it already works and doesn't seem all that "offensive" in a demo script until I can think of how we want to be doing this in the future when we actually integrate something properly which is probably soonish anyhow.

BenTheElder: thanks for documenting this!

BenTheElder: is there somewhere we can follow the standardization effort? I think this is a really great idea

nicks: nope, it's been mostly off-the-cuff conversations I've been having with people. It falls at the intersection of a few concerns (DevEx teams and cluster providers and image-build tools).

But this conversation is the kick in the pants I probably needed to start some more formal conversations. I'll kick that off by filing some issues in this repo + the k3d repo + the microk8s repo and see where that goes.

BenTheElder:
we're not changing kind sources here

Issue #1544, opened 2020-05-01
Can't create ipv4 clusters if ipv6 is disabled at kernel level


marcosnils:
**What happened**:
Due to https://github.com/kubernetes-sigs/kind/pull/1526, `docker network create` will constantly fail if `ipv6.disable=1` is set in boot time at kernel level.

**What you expected to happen**:

Kind cluster to be created

**How to reproduce it (as minimally and precisely as possible)**:

Set `ipv6.disable=1` in grub, and try to create a cluster

**Anything else we need to know?**:

Nope, this issue has been identified already

**Environment:**
All

BenTheElder:
/assign
/priority critical-urgent

workaround for now is:
```
docker network rm kind
docker network create -d=bridge kind
```

we'll fix this and cut 0.8.1

BenTheElder:
OK, set to repro and check exactly what behavior we can expect I think:
```console
$ cat /proc/cmdline
BOOT_IMAGE=/boot/vmlinuz-5.4.0-28-generic root=UUID=<snip> ro quiet splash vt.handoff=7 ipv6.disable=1
```

BenTheElder:
```console
$ kind create cluster
Creating cluster "kind" ...
ERROR: failed to create cluster: failed to ensure docker network: command "docker network create -d=bridge --ipv6 --subnet fc00:f853:ccd:e793::/64 kind" failed with error: exit status 1
Command Output: Error response from daemon: Cannot read IPv6 setup for bridge br-bc6b11ad1b57: open /proc/sys/net/ipv6/conf/br-bc6b11ad1b57/disable_ipv6: no such file or directory
```
I think we can detect this error and abort out of trying to do an IPv6 enabled network based on that.

@marcosnils is this what you see (presumably with a different bridge name)

marcosnils:
> @marcosnils is this what you see (presumably with a different bridge name)

Exactly!

BenTheElder:
in theory if you install (`go install .` or `make build && cp bin/kind /somewhere/kind`) this is fixed in HEAD (to confirm you'll want to delete the clusters and network first and see what happens creating a cluster)

BenTheElder:
https://github.com/kubernetes-sigs/kind/releases/tag/v0.8.1 contains the fix

PR #1545, opened 2020-05-01
handle hosts that cannot create ipv6 networks by falling back to ipv4…

BenTheElder:
… only

fixes https://github.com/kubernetes-sigs/kind/issues/1544

### DIFF
--- a/pkg/cluster/internal/providers/docker/network.go
+++ b/pkg/cluster/internal/providers/docker/network.go
@@ -48,27 +48,54 @@ func ensureNetwork(name string) error {
 	if err != nil {
 		return err
 	}
+	// network already exists, we're good
 	if exists {
 		return nil
 	}
 
-	// generate unique subnet per network based on the name
+	// Generate unique subnet per network based on the name
 	// obtained from the ULA fc00::/8 range
-	// make N attempts with "probing" in case we happen to collide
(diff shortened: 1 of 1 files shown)


BenTheElder:
I tested this on an ubuntu box with temporary grub parameter to disable ipv6 at boot

pkg/cluster/internal/providers/docker/network.go
```diff
@@ -48,27 +48,54 @@ func ensureNetwork(name string) error {
...
+	}
+
+	// On the first try check if ipv6 fails entirely on this machine
+	// https://github.com/kubernetes-sigs/kind/issues/1544
+	// Otherwise if it's not a pool overlap error, fail
+	// If it is, make more attempts below
+	if isIPv6UnavailableError(err) {
+		// only one attempt, IPAM is automatic in ipv4 only
+		return createNetwork(name, "")
+	} else if isPoolOverlapError(err) {
```
aojea, line 73: If ! We return, is missing the ! , No?

BenTheElder: Yeah, this was missed in review.
We should fix this, but it just means the probing for alternative space doesn't work which while unfortunate just means users should precreate the network with a different subnet if the one we pick is taken by them for something else.

PR #1573, opened 2020-05-08
fix NO_PROXY

BenTheElder:
- inspect the correct docker network
- include node names in no_proxy

### DIFF
--- a/pkg/cluster/internal/providers/docker/provision.go
+++ b/pkg/cluster/internal/providers/docker/provision.go
@@ -179,7 +179,7 @@ func commonArgs(cluster string, cfg *config.Cluster, networkName string) ([]stri
 	}
 
 	// pass proxy environment variables
-	proxyEnv, err := getProxyEnv(cfg)
+	proxyEnv, err := getProxyEnv(cfg, networkName)
 	if err != nil {
 		return nil, errors.Wrap(err, "proxy setup error")
 	}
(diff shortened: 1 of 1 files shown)


pkg/cluster/internal/providers/docker/provision.go
```diff
@@ -271,18 +283,20 @@ func runArgsForLoadBalancer(cfg *config.Cluster, name string, args []string) ([]
 	return append(args, loadbalancer.Image), nil
 }
 
-func getProxyEnv(cfg *config.Cluster) (map[string]string, error) {
+func getProxyEnv(cfg *config.Cluster, networkName string, nodeNames []string) (map[string]string, error) {
```
aojea, line 286: should `[]nodeNames` be in `common` and not provider specific?

BenTheElder: nah, there's no DNS in the others at the moment. we can DRY out later if that becomes the case

BenTheElder:
hrm NFS flakes ... https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/1573/pull-kind-e2e-kubernetes/1258906611737432064

going to need to devote some time to test deflake very soon.

mkimuram:
@BenTheElder 

Is my understanding right that in kind v0.8.x until this commit is included, or in kind v0.8.0 and v0.8.1, users need to specify `NO_PROXY` in a different way from what is described in [the current document](https://kind.sigs.k8s.io/docs/user/quick-start/#configure-kind-to-use-a-proxy), when deploying kind behind a proxy?

The document asks to add "172.17.0.0/16" to `NO_PROXY`, however these versions seem to require to add docker network's subnet and all nodes' name, like "172.18.0.0/16,kind-control-plane". (And after this commit is included, users won't need to specify anything special to `NO_PROXY` for changing kind's behavior.)

I've tested with v0.8.1, v0.8.0, v0.7.0, and master branch and reached to above conclusion, which took me a bit of time. If above is correct, I think that it is worth documented in somewhere. (If next version is planed to release soon, there won't be many people confused by this, though.)

BTW, this PR is very nice work, for it removed much complexity for users in deploying kind behind a  proxy. Thank you very much for your work!

BenTheElder:
> The document asks to add "172.17.0.0/16" to NO_PROXY, however these versions seem to require to add docker network's subnet and all nodes' name, like "172.18.0.0/16,kind-control-plane". (And after this commit is included, users won't need to specify anything special to NO_PROXY for changing kind's behavior.)



That's correct, sorry about that.

The next release is tentatively due by june 1st, though we got a bit less done than I hoped so far .. https://github.com/kubernetes-sigs/kind/milestones

> BTW, this PR is very nice work, for it removed much complexity for users in deploying kind behind a proxy. Thank you very much for your work!

thanks :-)

PR #1576, opened 2020-05-10
Add support for dockerless builds in kind

dims:
Support added in https://github.com/kubernetes/kubernetes/pull/87746


BenTheElder:
Do you know what the size reduction is? 👍

Issue #1572, opened 2020-05-08
Update documentation with new feature gate configuration


aojea:
**What should be cleaned up or changed**:

Documentation explains how to enable feature gates in KIND patching the kubeadm configuration

**Why is this needed**:

Since 0.8.0 feature gates can be enabled from the KIND configuration

https://github.com/kubernetes-sigs/kind/pull/1263

BenTheElder:
/help
I'll try to get to this soon but I have a few ongoing PRs. if anyone wants to beat me to it, basically just make sure the example configs are all using the v1alpha4 api version and then use the `featureGates` field instead of config patches, the featureGates field is a map of string feature gate name to bool enabled

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/1577

PR #1577, opened 2020-05-10
fix featureGate docs to use the new top level field in Cluster config

yashbhutwala:


tao12345666333:
fixes #1572

site/content/docs/user/quick-start.md
```diff
@@ -330,36 +330,8 @@ Feature gates are a set of key=value pairs that describe alpha or experimental f
...
-- |
-  kind: KubeletConfiguration
-  featureGates:
-    FeatureGateName: true
-- |
-  kind: KubeProxyConfiguration
-  featureGates:
-    FeatureGateName: true
+featureGates:
+ FeatureGateName: true
```
aojea, line 334: indent needs another space, right?

yashbhutwala: yes sir, you are correct, sorry about that 😛

site/content/docs/user/quick-start.md
```diff
@@ -330,36 +330,8 @@ Feature gates are a set of key=value pairs that describe alpha or experimental f
...
-      "feature-gates": "FeatureGateName=true"
-- |
-  kind: KubeletConfiguration
-  featureGates:
-    FeatureGateName: true
-- |
-  kind: KubeProxyConfiguration
-  featureGates:
-    FeatureGateName: true
+featureGates:
```
aojea, line 333: this is the config used in the GA jobs
```yaml
# config for 1 control plane node and 2 workers (necessary for conformance)
kind: Cluster
networking:
  ipFamily: ipv4
nodes:
- role: control-plane
- role: worker
- role: worker
featureGates: {"AllAlpha":false,"AllBeta":false,"RotateKubeletClientCertificate":true}
```

I have doubts now, better ask @BenTheElder directly :sweat_smile:

yashbhutwala: @aojea as far as I see, that's the same data structure expressed in two different ways. (just key: value)

aojea: that's my understanding, but I prefer to defer when I'm not sure 😉

BenTheElder: this is just yaml differences.

BenTheElder:
/lgtm
/approve
thanks!

Issue #1558, opened 2020-05-04
Need a way to get kubeconfig with node IP


BenTheElder:
**What would you like to be added**:
Some way to get the kubeconfig with the server field using the node container IP rather than the port forward on the hosts loopback or the internal DNS.

**Why is this needed**:

You could do this in 0.7 and on Linux it works from both host and inside the node / pods

howardjohn:
`kind get kubeconfig --name "${CLUSTER_NAME}" --internal | sed "s/${CLUSTER_NAME}-control-plane/$(docker inspect "${CLUSTER_NAME}-control-plane" --format "{{ .NetworkSettings.Networks.kind.IPAddress }}")/g"`

Is a workaround

aojea:
/close
I can't see now the utility of having the ip, there are multiple ways to obtain it, 
- kubectl get nodes
- nslookup
-

howardjohn:
The utility is that you can use the kubeconfig from within docker. Common in multicluster. Getting the IP is not so hard, but having to `sed` patch it isn't ideal

aojea:
```sh
kubectl get nodes -o=jsonpath='{range .items[*]}{.metadata.name}{":"}{.status.addresses[?(@.type=="InternalIP")].address}{"\n"}{end}'
testapi-control-plane:172.19.0.6
testapi-control-plane2:172.19.0.5
testapi-control-plane3:172.19.0.4
```

aojea:
is not valid using kubectl directly?

aojea:
I think this was already fixed

PR #1583, opened 2020-05-13
return IP address using kubeconfig --internal

aojea:
Before 0.8.0 kind get kubeconfig --internal returned the container
IP address of the node with the API server.

Since we added docker custom bridges, the behavior changed
returning the hostname.

We recover the previous behavior returning an IP address, however,
it will return always an IPv4 address.

Fixes: #1558

### DIFF
--- a/pkg/cluster/internal/kubeconfig/kubeconfig.go
+++ b/pkg/cluster/internal/kubeconfig/kubeconfig.go
@@ -20,8 +20,11 @@ package kubeconfig
 
 import (
 	"bytes"
+	"fmt"
+	"net"
 
 	"sigs.k8s.io/kind/pkg/cluster/internal/context"
+	"sigs.k8s.io/kind/pkg/cluster/internal/providers/provider/common"
 	"sigs.k8s.io/kind/pkg/cluster/nodeutils"
 	"sigs.k8s.io/kind/pkg/errors"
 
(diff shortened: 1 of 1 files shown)


aojea:
/assign @BenTheElder 

I'm not super happy but I couldn't find a "clean" way of getting the cluster IP family without adding a new method to the provider or a new variable indicating it.

Totally open to suggestions

BenTheElder:
I don't think most internal clients should use IP still.

aojea:
> I don't think most internal clients should use IP still.
> […](#)

IIRC that the original purpose of the flag was for obtaining the docker IP, it may be used for other containers that do not belong to the cluster.
Internal clients on the nodes can access directly the `/etc/kubernetes/kubeadm.config`

aojea:
adding a new parameter is too gross? 
`--internal (ipv4|ipv6|dns) (default ipv4)`

BenTheElder:
> IIRC that the original purpose of the flag was for obtaining the docker IP, it may be used for other containers that do not belong to the cluster.

no, the purpose of the flag is to obtain a kubeconfig for use inside the cluster network as opposed to the one we export for the host.

wanting it to be an IP instead of whatever the internal default management is is ... different.
it should probably be it's own flag, technically it could apply to the host but probably have a different default (host default to IP to avoid issues where we don't control /etc/hosts, internal default to dns...)

BenTheElder:
btw when I tag an issue `kind/design` that's because we need to design something like this new flag, I wouldn't have tagged a trivial fix this way :upside_down_face: 
I'm not sure what the clearest flags would be.

aojea:
heh my fault, honestly, didn't see the tags and make my mind around it thinking that they only wanted the IPs, let's close and discuss then
/close

PR #1585, opened 2020-05-13
Docs for using Ambassador as Ingress Controller

inercia:
This PR documents how to install [Ambassador](https://getambassador.io/) as an Ingress Controller.

Ambassador is installed via the Ambassador Operator which is the recommended way of installing Ambassador (it makes sure users always have the latest version of Ambassador installed and takes care of updates as well).

### DIFF
--- a/site/content/docs/user/ingress.md
+++ b/site/content/docs/user/ingress.md
@@ -25,6 +25,7 @@ by the ingress controller `nodeSelector`.
 
 1. [Create a cluster](#create-cluster)
 2. Deploy an Ingress controller, the following ingress controllers are known to work:
+    - [Ambassador](#ambassador)
     - [Contour](#contour)
     - [Ingress NGINX](#ingress-nginx)
 
@@ -57,6 +58,37 @@ nodes:
 EOF
 {{< /codeFromInline >}}
 
+### Ambassador
+
+[Ambassador](https://www.getambassador.io/) will be installed with the help of
+the [Ambassador operator](https://www.getambassador.io/docs/latest/topics/install/aes-operator/).
+
+First install the CRDs with
+
+{{< codeFromInline lang="bash" >}}
+kubectl apply -f https://github.com/datawire/ambassador-operator/releases/latest/download/ambassador-operator-crds.yaml
+{{< /codeFromInline >}}
+
+Now install the kind-specific manifest for installing Ambassador with the operator
+in the `ambassador` namespace:
+
+{{< codeFromInline lang="bash" >}}
+kubectl apply -n ambassador -f https://github.com/datawire/ambassador-operator/releases/latest/download/ambassador-operator-kind.yaml
+kubectl wait --timeout=180s -n ambassador --for=condition=deployed ambassadorinstallations/ambassador
+{{< /codeFromInline >}}
+
+Ambassador is now ready for use. You can try the example in [Using Ingress](#using-ingress) at this moment,
+but Ambassador will not automatically load the `Ingress` defined there. `Ingress` resources must include
+the annotation `kubernetes.io/ingress.class: ambassador` for being recognized by Ambassador (otherwise they are just ignored).
+So once the example has been loaded you can add this annotation with:
+
+{{< codeFromInline lang="bash" >}}
+kubectl annotate ingress example-ingress kubernetes.io/ingress.class=ambassador
+{{< /codeFromInline >}}
+
+Ambassador should be exposing your Ingress now. Please find additional documentation on
+Ambassador [here](https://www.getambassador.io/docs/latest/).
+
 ### Contour
 
 Deploy [Contour components](https://projectcontour.io/quickstart/contour.yaml).
@@ -83,6 +115,7 @@ Now the Contour is all setup to be used.
 Refer to [Using Ingress](#using-ingress) for a basic example usage.
 
 Additional information about Contour can be found at: [projectcontour.io](https://projectcontour.io)
+
 
 ### Ingress NGINX
 


aojea:
/ok-to-test
/lgtm
Thanks @inercia

aojea:
/hold
@inercia the current ingresses are adding some specific configuration to pin the ingress to the control plane node, so the port forwarding defined in the kind-config works.
Is this docs addressing it?

site/content/docs/user/ingress.md
```diff
@@ -26,6 +26,7 @@ by the ingress controller `nodeSelector`.
 1. [Create a cluster](#create-cluster)
 2. Deploy an Ingress controller, the following ingress controllers are known to work:
     - [Contour](#contour)
+    - [Ambassador](#ambassador)
```
amwat, line 29: nit: maybe move it above contour so it's in alphabetical order

inercia: Sure. I'll fix that.

inercia:
Hi @aojea! Thanks for spending some time reviewing this. I didn't realize about the `ingress-ready: "true"` label. I have modified the code for adding a `nodeSelector` with that label, so it should happen automatically.

@amwat You are right, it was not working because I forgot to note that `Ingress` resources must include an annotation `kubernetes.io/ingress.class: "ambassador"` for being recognized by Ambassador. Otherwise, they will be ignored. I have updated the docs accordingly.

aojea:
It's not working for me :(

Also, I think that for this section we should give more detailed steps, most users need more guidance on how to annotate the ingress resource

>  I forgot to note that Ingress resources must include an annotation kubernetes.io/ingress.class: "ambassador"

site/content/docs/user/ingress.md
```diff
@@ -84,6 +85,31 @@ Refer to [Using Ingress](#using-ingress) for a basic example usage.
 
 Additional information about Contour can be found at: [projectcontour.io](https://projectcontour.io)
 
+### Ambassador
```
amwat, line 88: nit: I also meant move this entire section above contour too (alphabetical) 😬

inercia:
> It's not working for me :(

What version of Kubernetes are you running? I have discovered it wasn't working with 1.18 (not completely sure the reason), but it should be working fine now. I have also added a `kubectl wait` in the instruction for being sure Ambassador is installed before going further. I would appreciate it if you could try the new code.

> Also, I think that for this section we should give more detailed steps, most users need more guidance on how to annotate the ingress resource

Yes, in fact, Ingress Controllers should use the `ingress.class` annotation for differentiating who serves an `Ingress`, but most controllers assume that, if the annotation is not there, they should take over that `Ingress` (see [Nginx](https://kubernetes.github.io/ingress-nginx/user-guide/multiple-ingress/) and [contour](https://projectcontour.io/docs/master/annotations/#ingress-class)). Ambassador does the opposite (this behavior has not been standardized). I have added a note on how to add the annotation.

site/content/docs/user/ingress.md
inercia: Sure, I can do that.

aojea:
> What version of Kubernetes are you running? 

it is installing 1.18 by default now,

inercia:
> You will also need to add tolerations for node-role.kubernetes.io/master similar to that shown in the contour section.

@amwat I have added the toleration for running in masters. The new Ambassador deployment will have:

```yaml
       tolerations:
       - effect: NoSchedule
         key: node-role.kubernetes.io/master
         operator: Equal
```

@aojea I have added a `kubectl patch` command for adding the `annotation` in the Ingress example. I have also added a `kubectl wait` for waiting for Ambassador to be ready, so it should be more reliable now...

aojea:
> KIND is installing 1.18 by default now,

no luck yet running the example with 1.18.2
pods are running, ingress patched but the curl to localhost/foo hangs :(

concaf:
@aojea I just tried deploying Ambassador with Kind following the instructions and it seems to work for me.
I've tested this locally on Fedora 30, Kube 1.18, Docker 19.

I tried it again on a fresh VM on GCE (Ubuntu 16.04) with nothing installed except docker, kind and kubectl, and it works there as well. I have created a short screencast of the same for reference - https://asciinema.org/a/5ju0SBnNiP9s6NAhB6a0Sv4TT - PTAL.

aojea:
> @aojea I just tried deploying Ambassador with Kind following the instructions and it seems to work for me.
> I've tested this locally on Fedora 30, Kube 1.18, Docker 19.
> 
> I tried it again on a fresh VM on GCE (Ubuntu 16.04) with nothing installed except docker, kind and kubectl, and it works there as well. I have created a short screencast of the same for reference - https://asciinema.org/a/5ju0SBnNiP9s6NAhB6a0Sv4TT - PTAL.

awesome, I think that my problem yesterday was that quay was down :thinking: .

```
quay.io/datawire/ambassador-operator:v1.2.3
```

Anyway, thanks for the detailed walkthrough, and sorry for being a bit picky, but it is in the best for everybody that things work out of the box, or that will cause a bad experience to users.

/lgtm

inercia:
Apologies for the `push -f`: just removed some spurious blanks lines...

aojea:
/lgtm
/hold cancel
:+1:

site/content/docs/user/ingress.md
```diff
@@ -57,6 +58,37 @@ nodes:
...
+kubectl wait --timeout=180s -n ambassador --for=condition=deployed ambassadorinstallations/ambassador
+{{< /codeFromInline >}}
+
+Ambassador is now ready for use. You can try the example in [Using Ingress](#using-ingress) at this moment,
+but Ambassador will not automatically load the `Ingress` defined there. `Ingress` resources must include
+the annotation `kubernetes.io/ingress.class: ambassador` for being recognized by Ambassador (otherwise they are just ignored).
+So once the example has been loaded you can add this annotation with a `patch`:
+
+{{< codeFromInline lang="bash" >}}
+kubectl patch ingress example-ingress -p '{"metadata":{"annotations":{"kubernetes.io/ingress.class": "ambassador"}}}'
```
amwat, line 86: nit: `kubectl annotate ingress example-ingress kubernetes.io/ingress.class=ambassador` is a bit simpler?

inercia: Damn :facepalm: Should I change it?

BenTheElder: I think we should

inercia: Fixed.

inercia:
Thanks @aojea, @BenTheElder, @amwat, and @concaf for your patience and your time. :-)

PR #1587, opened 2020-05-13
add kindnetd readme with history

BenTheElder:
/cc @aojea @amwat 
ref: https://twitter.com/BenTheElder/status/1260697719506472961

I also added a readme for the haproxy image while I was at it


aojea:
/lgtm
this is really nice, thanks

Issue #1581, opened 2020-05-12
docker vs podman emptyDir mount options (exec vs noexec)


joelddiaz:
**What happened**:
Running a pod with an emptyDir defined results in 'noexec' when run under Podman, and the 'noexec' is missing when run under Docker.

**What you expected to happen**:
Same behavior.

**How to reproduce it (as minimally and precisely as possible)**:
Start up kind with both docker and podman, create the namespace 'testexec' and run this pod on both docker and podman environments with `kubectl create -f ./<file with pod contents>`:
```
apiVersion: v1
kind: Pod
metadata:
  name: testpod
  namespace: testexec
spec:
  containers:
  - args:
    - --output
    - SOURCE,TARGET,FSTYPE,OPTIONS
    command:
    - /usr/bin/findmnt
    image: fedora:latest
    imagePullPolicy: Always
    name: test
    volumeMounts:
    - mountPath: /output
      name: output
  restartPolicy: Never
  volumes:
  - emptyDir: {}
    name: output
```

When the Pod has finished running compare the log output showing the `/output` mountpoint with `kubectl logs testpod`.

Under Podman:
```
/dev/mapper/ssd-root[/var/lib/containers/storage/volumes/9fbd845d68a713e0cf8d22d236a22fc3a8062b21dfe471680992148f62beac2b/_data/lib/kubelet/pods/cc549f1d-935b-4f7c-8525-0d022f43302c/volumes/kubernetes.io~empty-dir/output]                                    |-/output                                   ext4    rw,nosuid,nodev,noexec,relatime,seclabel
```

Under Docker:
```
/dev/mapper/ssd-root[/var/lib/docker/volumes/8ca5f64dd3e81bfe12742a866a6fa9fc8355fc945a235d28b2af018be89c8d31/_data/lib/kubelet/pods/ea1d98af-68c6-4c60-9677-52e332a331ca/volumes/kubernetes.io~empty-dir/output]                                    |-/output                                   ext4    rw,relatime,seclabel
```

**Anything else we need to know?**:

**Environment:**

- kind version: (use `kind version`): kind v0.8.1 go1.14.2 linux/amd64
- Kubernetes version: (use `kubectl version`):
```
[...]

PR #1589, opened 2020-05-14
[Podman] Fix volume mount options

amwat:
fixes: https://github.com/kubernetes-sigs/kind/issues/1581

Unfortunately podman doesn't support adding volume mount options to anonymous volumes during container run time.
Precreated anonymous volumes, labeled them with appropriate node name as key (for later retrieval and deletion)
Also added other default options that docker volume mounts have.

### DIFF
--- a/pkg/cluster/internal/providers/podman/util.go
+++ b/pkg/cluster/internal/providers/podman/util.go
@@ -17,6 +17,7 @@ limitations under the License.
 package podman
 
 import (
+	"fmt"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/util/version"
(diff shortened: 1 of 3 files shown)


neolit123:
logic looks fine, but i hardly know anything about podman. LGTM

pkg/cluster/internal/providers/podman/util.go
```diff
@@ -58,3 +60,47 @@ func ensureMinVersion() error {
...
+func createAnonymousVolume(label string) (string, error) {
+	cmd := exec.Command("podman",
+		"volume",
+		"create",
+		// podman only support filter on key during list
+		// so we use the unique id as key
+		"--label", fmt.Sprintf("%s=true", label))
+	var name bytes.Buffer
+	cmd.SetStdout(&name)
+	if err := cmd.Run(); err != nil {
```
BenTheElder, line 76: We have exec.Output now

BenTheElder:
seems fine. only touches podman code.
thanks.

dgoodwin:
Major thanks for this!

PR #1592, opened 2020-05-14
masquerade docker network traffic

aojea:
It turns out that docker custom networks does not enable masquerade by default.
This is a change in the behaviour that we have with the docker default network and we should keep masquerading the traffic, so the nodes and pods can have external access.


aojea:
/priority important-soon

@mauilion is very right, if the current network was not masquerading most users will be complaining.

This way we enforce it, that is not bad if we deal with some distros/users that does not create masquerade rules by default, however, I don't have clear if this may happen

aojea:
/remove priority/critical-urgent
/remove priority/important-soon

it seems it's not needed, removing priorities and letting others comment

BenTheElder:
Ugh.
Thanks

PR #1595, opened 2020-05-14
Add support to configure kube-proxy mode

rikatz:
Signed-off-by: Ricardo Pchevuzinske Katz <ricardo.katz@gmail.com>

This PR adds support to configure the kube-proxy mode in KinD.

Instead of patching the kubeadm configuration, the user now just need to use the following in KinD config:

```
kind: Cluster
networking:
  kubeProxyMode: ipvs
```

And hopefully I won't break anything :D

### DIFF
--- a/pkg/apis/config/v1alpha4/types.go
+++ b/pkg/apis/config/v1alpha4/types.go
@@ -174,6 +174,9 @@ type Networking struct {
 	// If DisableDefaultCNI is true, kind will not install the default CNI setup.
 	// Instead the user should install their own CNI after creating the cluster.
 	DisableDefaultCNI bool `yaml:"disableDefaultCNI,omitempty"`
+	// KubeProxyMode defines if kube-proxy should operate in iptables or ipvs mode
+	// Defaults to 'iptables' mode
+	KubeProxyMode ProxyMode `yaml:"kubeProxyMode,omitempty"`
 }
 
 // ClusterIPFamily defines cluster network IP family
@@ -184,6 +187,16 @@ const (
 	IPv4Family ClusterIPFamily = "ipv4"
 	// IPv6Family sets ClusterIPFamily to ipv6
 	IPv6Family ClusterIPFamily = "ipv6"
+)
+
+// ProxyMode defines a proxy mode for kube-proxy
+type ProxyMode string
+
+const (
+	// IPTablesMode sets ProxyMode to iptables
+	IPTablesMode ProxyMode = "iptables"
+	// IPVSMode sets ProxyMode to iptables
+	IPVSMode ProxyMode = "ipvs"
 )
 
 // PatchJSON6902 represents an inline kustomize json 6902 patch
--- a/pkg/internal/apis/config/default.go
+++ b/pkg/internal/apis/config/default.go
@@ -67,6 +67,10 @@ func SetDefaultsCluster(obj *Cluster) {
 			obj.Networking.ServiceSubnet = "fd00:10:96::/112"
 		}
 	}
+	// default the KubeProxyMode using iptables as it's already the default
+	if obj.Networking.KubeProxyMode == "" {
+		obj.Networking.KubeProxyMode = IPTablesMode
+	}
 }
 
 // SetDefaultsNode sets uninitialized fields to their default value.
--- a/pkg/internal/apis/config/validate.go
+++ b/pkg/internal/apis/config/validate.go
@@ -45,6 +45,11 @@ func (c *Cluster) Validate() error {
 		errs = append(errs, errors.Wrapf(err, "invalid serviceSubnet"))
 	}
 
+	// KubeProxyMode should be iptables or ipvs
+	if c.Networking.KubeProxyMode != IPTablesMode && c.Networking.KubeProxyMode != IPVSMode {
+		errs = append(errs, errors.Errorf("invalid kubeProxyMode: %s", c.Networking.KubeProxyMode))
+	}
+
 	// validate nodes
 	numByRole := make(map[NodeRole]int32)
 	// All nodes in the config should be valid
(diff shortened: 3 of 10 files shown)


pkg/internal/apis/config/validate.go
```diff
@@ -45,6 +45,11 @@ func (c *Cluster) Validate() error {
 		errs = append(errs, errors.Wrapf(err, "invalid serviceSubnet"))
 	}
 
+	// KubeProxyMode should be iptables or ipvs
+	if c.Networking.KubeProxyMode != "ipvs" && c.Networking.KubeProxyMode != "iptables" {
+		errs = append(errs, errors.Errorf("invalid kubeProxyMode"))
```
neolit123, line 50: `errors.Errorf("invalid kubeProxyMode: %s, c.Networking.KubeProxyMode)`

pkg/apis/config/v1alpha3/types.go
```diff
@@ -126,6 +126,8 @@ type Networking struct {
 	// If DisableDefaultCNI is true, kind will not install the default CNI setup.
 	// Instead the user should install their own CNI after creating the cluster.
 	DisableDefaultCNI bool `yaml:"disableDefaultCNI,omitempty"`
+	// KubeProxyMode defines if kube-proxy should operate in iptables or ipvs mode
+	KubeProxyMode string `yaml:"kubeProxyMode,omitempty"`
```
neolit123, line 130: shouldn't API changes be added only to the latest version - v1alpha4?

BenTheElder: yes, we've held onto v1alpha3 for a while, it's been dropped now.

pkg/internal/apis/config/default.go
```diff
@@ -67,6 +67,10 @@ func SetDefaultsCluster(obj *Cluster) {
 			obj.Networking.ServiceSubnet = "fd00:10:96::/112"
 		}
 	}
+	// default the KubeProxyMode using iptables as it's already the default
+	if obj.Networking.KubeProxyMode == "" {
+		obj.Networking.KubeProxyMode = "iptables"
```
aojea, line 72: we should use an enum, see the IPFamily for reference https://github.com/kubernetes-sigs/kind/pull/1595/files#diff-4686ef90e1f4b2babdb7c7bf7997b62dL136-L143

aojea:
this will make happy some people ;)
/cc @andrewsykim

pkg/apis/config/v1alpha3/types.go
rikatz: Yes, thank you. Corrected here and removed from v1alpha3.

Also, in https://github.com/kubernetes-sigs/kind/pull/1595/files#diff-50194a7023ae768e6fec0e8c8b122fbbR293 I'm putting the config in ``ConfigTemplateAlphaV3`` and also the Beta ones. Should I remove from here and keep only for the Beta ones, assuming those are the Go templates that will be used, or it depends on which version of k8s I'm creating with KinD?

In this same subject, I didn't changed the ``ConfigTemplateAlphaV2`` so just need a direction here :)

pkg/internal/apis/config/default.go
rikatz: I've made some adjustments here, let me know if something else is needed :)

rikatz:
> added a couple of comments.
> could you clarify on the benefits of exposing this for all kind users?

@neolit123 IPVS being used as an alternative mode to iptables could be used by users to test behaviors, also could be used in some kubernetes test jobs that deal with IPVS.

I've got to test this at least two times in some internal stuff, and needed to use kubeadmPatch (as per #1456 and opened the PR as a suggestion from @BenTheElder [here](https://github.com/kubernetes-sigs/kind/pull/1456#discussion_r419025152) I've opened the PR :)

Let me know if you need any other information, and thanks!

rikatz:
/retest

Just in case this is not related with kube-proxy but some flake

pkg/internal/apis/config/validate.go
rikatz: Corrected

BenTheElder:
> When the proxy mode is ipvs, can we have ipvsadm pre-installed on the worker nodes?

we avoid installing things that are not dependencies of running kubernetes because it's already a huge struggle to keep the node size down. is this a requirement? can you elaborate?

we also do not currently do runtime install, kind works fully offline.

BenTheElder:
should add: it's also fully valid to take the node or base image and do `FROM <pre-build kind image>; RUN clean-install ipvsadm` and then use that.

pkg/apis/config/v1alpha4/types.go
```diff
@@ -174,6 +174,8 @@ type Networking struct {
 	// If DisableDefaultCNI is true, kind will not install the default CNI setup.
 	// Instead the user should install their own CNI after creating the cluster.
 	DisableDefaultCNI bool `yaml:"disableDefaultCNI,omitempty"`
+	// KubeProxyMode defines if kube-proxy should operate in iptables or ipvs mode
+	KubeProxyMode ProxyMode `yaml:"kubeProxyMode,omitempty"`
```
BenTheElder, line 178: doc the default?

andrewsykim:
> we avoid installing things that are not dependencies of running kubernetes because it's already a huge struggle to keep the node size down. is this a requirement? can you elaborate?

This is nice to have but not required. It doesn't sound like it's worth the trade-off based on what you said. `ipvsadm` is a CLI to read / update / delete virtual servers and their backends. It's mainly just useful for debugging purposes, kube-proxy talks to IPVS via netlink and never calls out to ipvsadm so it's not required for a working cluster.

pkg/apis/config/v1alpha4/types.go
rikatz: Added right now in my last commit

aojea:
/lgtm
it also works in the CI per https://github.com/kubernetes-sigs/kind/pull/1607

BenTheElder:
thanks for this 😄 
/approve
/lgtm

BenTheElder:
CI is still broken by the quay thing so.

PR #1600, opened 2020-05-15
implement dns caching

aojea:
since we implemented custom docker bridges, we use hostnames
instead of ip addresses on all the components.
This allows us to not care about the nodes IP addresses, so there
is no problem if the cluster is rebooted and other IPs are assigned
to the node.
However, this is not free, and it comes at the cost that every
hostname resolution is a DNS request, and by default uses UDP.
On noise environments this can cause packet drops, the most common
solution to avoid this is implement caching.
Linux has implemented caching in the glibc library using the nscd
daemon, that seems a good fit for us because does not require to
change the /etc/resolv.conf, instead other alternatives like dnsmasq.

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -70,6 +70,7 @@ RUN echo "Ensuring scripts are executable ..." \
       libseccomp2 \
       bash ca-certificates curl rsync \
       nfs-common \
+      nscd \
     && find /lib/systemd/system/sysinit.target.wants/ -name "systemd-tmpfiles-setup.service" -delete \
     && rm -f /lib/systemd/system/multi-user.target.wants/* \
     && rm -f /etc/systemd/system/*.wants/* \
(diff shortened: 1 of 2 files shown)


aojea:
/assign @BenTheElder 
please do your magic so we can run it in the CI, this should alleviate a lot the DNS load.
I tested locally and seems to work fine
/priority important-soon

images/base/Dockerfile
```diff
@@ -112,7 +113,9 @@ RUN echo "Ensuring scripts are executable ..." \
  && echo "Adjusting systemd-tmpfiles timer" \
     && sed -i /usr/lib/systemd/system/systemd-tmpfiles-clean.timer -e 's#OnBootSec=.*#OnBootSec=1min#' \
  && echo "Modifying /etc/nsswitch.conf to prefer hosts" \
-    && sed -i /etc/nsswitch.conf -re 's#^(hosts:\s*).*#\1dns files#'
+    && sed -i /etc/nsswitch.conf -re 's#^(hosts:\s*).*#\1dns files#' \
+ && echo "Enabling DNS caching" \
+    && systemctl enable nscd
```
BenTheElder, line 118: i'm not so sure this is the way to go about this
https://unix.stackexchange.com/a/6240

aojea: basically we only want to cache the calls to the node names `kind-control-plane`, those are hostnames are not going to change and we don't care about its TTL (if they change is because the container was restarted and the cache is empty).

nscd intercepts the libc calls (gethostby*), in this case if we can avoid things like kubelet asking for the node ip periodically (every 60 sec???) to get the node IP.

let me double check how nscd interact with docker and pods and kubernetes 🤔 , it's important that does not intercepts the pods querys to coredns and coredns to external

BenTheElder: is the cache actually empty? /var is persistent.

aojea: don't know the internals, but I think that /var/run/nscd is only needed to create the socket

aojea:
will come back later to this if the DNS resolutions show up as a problem,

PR #1605, opened 2020-05-15
stop using IP auto detection

BenTheElder:
https://github.com/kubernetes/kubernetes/pull/91125#discussion_r426051191


BenTheElder:
CI is going to flake, let's get this one in as part of preventing flakes. this change is safe ...

Issue #1543, opened 2020-05-01
Standard way for Kind to communicate about a local registry


nicks:
Many Kubernetes clusters support a local registry, to make pushing images directly to the cluster fast. But every cluster documents it with manual shell scripts:

- https://kind.sigs.k8s.io/docs/user/local-registry/
- https://microk8s.io/docs/registry-built-in
- https://github.com/rancher/k3d/blob/master/docs/registries.md#using-a-local-registry
- https://docs.openshift.com/container-platform/4.3/registry/accessing-the-registry.html#registry-accessing-directly_accessing-the-registry

Because every cluster is slightly different, this makes it hard for tools to automate this work for users.

## The big idea

Clusters should use annotations on the `kube-system` namespace to document how to configure a registry

## Detailed design

I propose 3 new "standard" annotations:

`x-k8s.io/registry`: the host of the registry (hostname and port), as seen from outside of the cluster
`x-k8s.io/registry-from-cluster`: the host of the registry (hostname and port), as seen from inside the cluster
`x-k8s.io/registry-help`: The URL of documentation where users can read more about how to set a registry up. 

Tools that would benefit from local registries should read this annotation and display it to the user

## What this means for Kind

All Kind clusters would have a `registry-help` annotation pointing users to https://kind.sigs.k8s.io/docs/user/local-registry/

We'd also update the example script at https://kind.sigs.k8s.io/docs/user/local-registry/ to add the `x-k8s.io/registry` annotation during setup.

## Prior Art

This proposal is based on what we've been recommending to people in Tilt, with a lot of success: https://docs.tilt.dev/choosing_clusters.html#custom-clusters

*update 5/4* - changed the proposal to use the `kube-system` namespace instead of node annotations

nicks:
I also filed this feature request here https://github.com/rancher/k3d/issues/234 and here https://github.com/ubuntu/microk8s/issues/1173

BenTheElder:
FWIW x-k8s.io currently is typically used by community projects under something like `<project>.x-k8s.io`, I think there's guidance about this somewhere but I haven't needed to look in a while.

i'm not sure if there's something this would neatly fall under.

I like the general idea though, if not a standard annotation maybe something like checking for `.*/registry-help` etc?

BenTheElder:
sending a poke to cluster-lifecycle, which seems like they might be the appropriate group for this, given that it's sort of a cluster-addon detail.

neolit123:
[...]

PR #1606, opened 2020-05-16
Fixes kind-with-registry.sh

h0tbird:
Makes it more idempotent when connecting the network.
Switches the node annotation from `kind.x-k8s.io/registry` to `tilt.dev/registry`.

### DIFF
--- a/site/static/examples/kind-with-registry.sh
+++ b/site/static/examples/kind-with-registry.sh
@@ -22,10 +22,10 @@ containerdConfigPatches:
 EOF
 
 # connect the registry to the cluster network
-docker network connect "kind" "${reg_name}"
+docker network connect "kind" "${reg_name}" 2>/dev/null || true
 
 # tell https://tilt.dev to use the registry
 # https://docs.tilt.dev/choosing_clusters.html#discovering-the-registry
 for node in $(kind get nodes); do
-  kubectl annotate node "${node}" "kind.x-k8s.io/registry=localhost:${reg_port}";
+  kubectl annotate node "${node}" "tilt.dev/registry=localhost:${reg_port}"
 done


site/static/examples/kind-with-registry.sh
```diff
@@ -22,10 +22,10 @@ containerdConfigPatches:
 EOF
 
 # connect the registry to the cluster network
-docker network connect "kind" "${reg_name}"
+docker network connect "kind" "${reg_name}" 2>/dev/null || true
```
BenTheElder, line 25: we should instead skip this step if the registry was already running. we can remember this from earlier in the script.
doing it this way means we ignore all errors.

site/static/examples/kind-with-registry.sh
```diff
@@ -22,10 +22,10 @@ containerdConfigPatches:
...
 
 # tell https://tilt.dev to use the registry
 # https://docs.tilt.dev/choosing_clusters.html#discovering-the-registry
 for node in $(kind get nodes); do
-  kubectl annotate node "${node}" "kind.x-k8s.io/registry=localhost:${reg_port}";
+  kubectl annotate node "${node}" "tilt.dev/registry=localhost:${reg_port}"
```
BenTheElder, line 30: tilt had previously agreed to use this annotation, but /shrug, there's a KEP to do this in a better way soon anyhow. #1543

h0tbird:
Thank you @BenTheElder.
Closing this one, I am watching the KEP task instead.

site/static/examples/kind-with-registry.sh
nicks: haha, sorry, we didn't get a release out fast enough with this support :facepalm:

PR #1607, opened 2020-05-17
test IPVS mode

aojea:
based on https://github.com/kubernetes-sigs/kind/pull/1595

KIND can improve testing for kube-proxy using IPVS , only recently it started to run a periodic job and in presubmits.
In order to do that we have to pass an environment variable to choose the mode.

### DIFF
--- a/hack/ci/e2e-k8s.sh
+++ b/hack/ci/e2e-k8s.sh
@@ -114,6 +114,8 @@ create_cluster() {
     ;;
   esac
 
+KUBE_PROXY_MODE="ipvs"
+
   # create the config file
   cat <<EOF > "${ARTIFACTS}/kind-config.yaml"
 # config for 1 control plane node and 2 workers (necessary for conformance)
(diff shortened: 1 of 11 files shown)


aojea:
/hold

@BenTheElder if you agree I'll ask Ricardo to cherry-pick 7b8ae51b736bad6514c5d031c660c1c982c2aaf9 into his PR

BenTheElder:
please don't add anything to this script. opening a PR to temporarily vet it in CI is fine, but I would prefer that we:
- stick to defaults where possible
- not complicate this script, I'm going to replace it with an actual program before it grows out of control #1269

aojea:
> please don't add anything to this script. opening a PR to temporarily vet it in CI is fine, but I would prefer that we:
> 
> * stick to defaults where possible
> * not complicate this script, I'm going to replace it with an actual program before it grows out of control #1269

sounds like a good plan :)
/close

Issue #1601, opened 2020-05-15
Uppercased cluster names not properly working in 0.8.X version


angelbarrera92:
**What happened**:
Seems like any cluster created with kind 0.8.X having a capital letter in the cluster-name breaks the cluster creation process.
It works in 0.7.0

**What you expected to happen**:
I expect to get a validation error message if the capital letter is not allowed or a created cluster with the proposed name

**How to reproduce it (as minimally and precisely as possible)**:
Just run `kind create cluster  --name Bug` with both versions:  0.7.0 and 0.8.1.

**Anything else we need to know?**:
```
$ kind create cluster  --name Bug
Creating cluster "Bug" ...
 ✓ Ensuring node image (kindest/node:v1.18.2) 🖼
 ✓ Preparing nodes 📦
 ✓ Writing configuration 📜
 ✗ Starting control-plane 🕹️
ERROR: failed to create cluster: failed to init node with kubeadm: command "docker exec --privileged Bug-control-plane kubeadm init --ignore-preflight-errors=all --config=/kind/kubeadm.conf --skip-token-print --v=6" failed with error: exit status 1
Command Output: I0515 14:29:03.437459     145 initconfiguration.go:200] loading configuration from "/kind/kubeadm.conf"
[config] WARNING: Ignored YAML document with GroupVersionKind kubeadm.k8s.io/v1beta2, Kind=JoinConfiguration
I0515 14:29:03.499819     145 interface.go:400] Looking for default routes with IPv4 addresses
I0515 14:29:03.500189     145 interface.go:405] Default route transits interface "eth0"
I0515 14:29:03.500872     145 interface.go:208] Interface eth0 is up
I0515 14:29:03.500977     145 interface.go:256] Interface "eth0" has 3 addresses :[172.20.0.2/16 fc00:f853:ccd:e793::2/64 fe80::42:acff:fe14:2/64].
I0515 14:29:03.501209     145 interface.go:223] Checking addr  172.20.0.2/16.
I0515 14:29:03.501232     145 interface.go:230] IP found 172.20.0.2
I0515 14:29:03.501278     145 interface.go:262] Found valid IPv4 address 172.20.0.2 for interface "eth0".
I0515 14:29:03.501301     145 interface.go:411] Found active IP 172.20.0.2
hostport Bug-control-plane:6443: host 'Bug-control-plane' must be a valid IP address or a valid RFC-1123 DNS subdomain
[...]

PR #1610, opened 2020-05-18
disallow capitals in cluster names

BenTheElder:
"fixes" https://github.com/kubernetes-sigs/kind/issues/1601
I'll follow up with the other commit making this regex more accurate when I'm more confident in it, filing this to ensure the trivial fix doesn't slip.

### DIFF
--- a/pkg/cluster/internal/create/create.go
+++ b/pkg/cluster/internal/create/create.go
@@ -53,7 +53,7 @@ const (
 // and suffix this name, we can relax it a little
 // see NewContext() for usage
 // https://godoc.org/github.com/docker/docker/daemon/names#pkg-constants
-var validNameRE = regexp.MustCompile(`^[a-zA-Z0-9_.-]+$`)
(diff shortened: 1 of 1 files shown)


pkg/cluster/internal/create/create.go
```diff
@@ -53,7 +53,7 @@ const (
 // and suffix this name, we can relax it a little
 // see NewContext() for usage
 // https://godoc.org/github.com/docker/docker/daemon/names#pkg-constants
-var validNameRE = regexp.MustCompile(`^[a-zA-Z0-9_.-]+$`)
+var validNameRE = regexp.MustCompile(`^[a-z0-9_.-]+$`)
```
BenTheElder, line 56: this regex is going to change further, but we want to get that particular bug corrected already ...

aojea: there is also a TODO to move this check to the config validation, is that the plan?

BenTheElder: yeah, it'll work here for now but

Issue #1611, opened 2020-05-18
podman detection commonly foiled by fake "docker"


BenTheElder:
**What happened**:
There's a package `podman-docker` that symlinks `docker` to podman, defeating the auto detection of podman vs docker being installed and breaking the node provider because it thinks it's using docker when it's not.

**What you expected to happen**: We should auto detect this correctly. We can probably just check if docker is a symlink and resolve it.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment:**

- kind version: (use `kind version`):
- Kubernetes version: (use `kubectl version`):
- Docker version: (use `docker info`):
- OS (e.g. from `/etc/os-release`):

BenTheElder:
correction: it's a script with roughly `exec /usr/bin/podman "$@"`

BenTheElder:
though I think some users do symlink as well.
we might have to do something more slow like invoke the version subcommand and check the output.

BenTheElder:
cc @amwat @aojea

aojea:
if @amwat does not beat me ;) I can send something end of the week

BenTheElder:
https://github.com/containers/libpod/blob/master/docker

PR #1612, opened 2020-05-19
improved provider detection

aojea:
Two changes in this PR:

* Fail if podman is running rootless, currently it throws a warning and later it fails or exits with noop operation ... it's confusing/annoying :upside_down_face: , I think is better failing fast and inform about the problem

* Detect the provider by checking the `command -v` output, we were looking for the path of the binary, but since we are executing the commands directly this should be safe. Also, we avoid the tricks done to replace one or the other command

Fixes: https://github.com/kubernetes-sigs/kind/issues/1611

### DIFF
--- a/pkg/cluster/internal/providers/docker/util.go
+++ b/pkg/cluster/internal/providers/docker/util.go
@@ -22,6 +22,16 @@ import (
 	"sigs.k8s.io/kind/pkg/exec"
 )
 
+// IsAvailable checks if docker is available in the system
+func IsAvailable() bool {
+	cmd := exec.Command("docker", "-v")
+	lines, err := exec.OutputLines(cmd)
+	if err != nil || len(lines) != 1 {
+		return false
+	}
+	return strings.HasPrefix(lines[0], "Docker version")
+}
+
 // usernsRemap checks if userns-remap is enabled in dockerd
 func usernsRemap() bool {
 	cmd := exec.Command("docker", "info", "--format", "'{{json .SecurityOptions}}'")
--- a/pkg/cluster/internal/providers/podman/util.go
+++ b/pkg/cluster/internal/providers/podman/util.go
@@ -26,6 +26,16 @@ import (
 	"sigs.k8s.io/kind/pkg/exec"
 )
 
+// IsAvailable checks if podman is available in the system
+func IsAvailable() bool {
+	cmd := exec.Command("podman", "-v")
+	lines, err := exec.OutputLines(cmd)
+	if err != nil || len(lines) != 1 {
+		return false
+	}
+	return strings.HasPrefix(lines[0], "podman version")
+}
+
 func getPodmanVersion() (*version.Version, error) {
 	cmd := exec.Command("podman", "--version")
 	lines, err := exec.CombinedOutputLines(cmd)
--- a/pkg/cluster/provider.go
+++ b/pkg/cluster/provider.go
@@ -17,7 +17,6 @@ limitations under the License.
 package cluster
 
 import (
-	"os/exec"
 	"sort"
 
 	"sigs.k8s.io/kind/pkg/cluster/constants"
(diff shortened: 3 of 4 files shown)


pkg/cluster/provider.go
```diff
@@ -69,12 +70,20 @@ func NewProvider(options ...ProviderOption) *Provider {
 		}
 	}
 
+	isLegitProvider := func(prog string) bool {
+		out, err := exec.Command(prog, "-v").Output()
+		if err != nil {
+			return false
+		}
+		return strings.Contains(string(out), prog)
```
BenTheElder, line 78: does this actually work? `podman --version` spits out $0

BenTheElder: per @amwat, so if a user did the symlink trick to avoid installing a package, I'm not sure this works.

BenTheElder:
general +1, one question
also ignore CI, it's the quay.io image thing

pkg/cluster/provider.go
aojea:
```sh
more docker
#!/bin/sh

exec /usr/bin/podman "$@"

 aojea  /  usr  bin  sudo ./docker -v
podman version 1.9.1
```

BenTheElder: i mean if you put a symlink in path like `/usr/bin/docker -> /usr/bin/podman` ?

BenTheElder:
```
root@instance-4:/home/amwat/amwat/kind# ls -al /usr/bin/docker
lrwxrwxrwx 1 root root 15 May 18 23:02 /usr/bin/docker -> /usr/bin/podman
root@instance-4:/home/amwat/amwat/kind# which docker
/usr/bin/docker
root@instance-4:/home/amwat/amwat/kind# docker --version
docker version 1.8.0
root@instance-4:/home/amwat/amwat/kind# podman --version
podman version 1.8.0
```

BenTheElder: also on my machine (macOS):
```
$ docker -v
Docker version 19.03.8, build afacb8b
```
note the capital D

aojea: you are dight, for reference
```
 aojea  /  usr  bin  ./podman -v
podman version 1.9.1
 aojea  /  usr  bin  ./docker -v
docker version 1.9.1
 aojea  /  usr  bin  ls -al docker
lrwxrwxrwx. 1 root root 6 may 19 18:49 docker -> podman
```

pkg/cluster/internal/providers/podman/util.go
```diff
@@ -26,6 +26,16 @@ import (
 	"sigs.k8s.io/kind/pkg/exec"
 )
 
+// IsAvailable checks if podman is available in the system
+func IsAvailable() bool {
```
aojea, line 30: amwat has almost the same code below :1st_place_medal:

pkg/cluster/internal/providers/podman/util.go
```diff
@@ -26,6 +26,16 @@ import (
...
+	cmd := exec.Command("podman", "-v")
```
BenTheElder, line 31: maybe make these consistent in a follow up

pkg/cluster/internal/providers/docker/util.go
```diff
@@ -22,6 +22,16 @@ import (
...
 )
 
+// IsAvailable checks if docker is available in the system
+func IsAvailable() bool {
+	cmd := exec.Command("docker", "-v")
+	lines, err := exec.OutputLines(cmd)
+	if err != nil || len(lines) != 1 {
+		return false
+	}
+	return strings.HasPrefix(lines[0], "Docker version")
```
BenTheElder, line 32: in the future I think this could probably be relaxed to more like:
- get output (not lines)
- check if output has substring

BenTheElder:
CI is pending https://github.com/kubernetes/kubernetes/pull/91257

PR #1627, opened 2020-05-27
trap INT and TERM in CI

BenTheElder:
EXIT is not handled consistently across shells

this will be obviated by #986 but the need to debug some timeouts is a bit more pressing and this change is tiny ...

### DIFF
--- a/hack/ci/e2e.sh
+++ b/hack/ci/e2e.sh
@@ -37,7 +37,7 @@ install_kind() {
 main() {
   # create temp dir and setup cleanup
   TMP_DIR=$(mktemp -d)
-  trap cleanup EXIT
+  trap cleanup INT TERM EXIT
 
   # install kind
   install_kind
--- a/hack/ci/e2e-k8s.sh
+++ b/hack/ci/e2e-k8s.sh
@@ -27,8 +27,12 @@ set -o errexit -o nounset -o xtrace
 #          false - (default) APIs and features left at defaults
 # 
 
-# our exit handler (trap)
+# cleanup logic for cleanup on exit
+CLEANED_UP=false
 cleanup() {
+  if [ "$CLEANED_UP" = "true" ]; then
+    return
+  fi
   # KIND_CREATE_ATTEMPTED is true once we: kind create
   if [ "${KIND_CREATE_ATTEMPTED:-}" = true ]; then
     kind "export" logs "${ARTIFACTS}/logs" || true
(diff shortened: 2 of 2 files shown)


hack/ci/e2e.sh
```diff
@@ -37,7 +37,7 @@ install_kind() {
 main() {
   # create temp dir and setup cleanup
   TMP_DIR=$(mktemp -d)
-  trap cleanup EXIT
+  trap cleanup INT TERM EXIT
```
BenTheElder, line 40: this one should already be doing this because it's bash, but the other script may be ash/dash (it's using sh)

aojea: this comment is confusing me regarding trapping EXIT
https://unix.stackexchange.com/a/149093

aojea:
/lgtm
it seems worth to try

hack/ci/e2e.sh
BenTheElder: > There's also what @Shawn said: Ash and Dash don't trap signals with EXIT.

that's the only relevant part really.

POSIX has requirements about what it looks like when EXIT is called, but when it is called is under specified. we explicitly want these

aojea:
/hold

we were luck to hit a timeout https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/1627/pull-kind-e2e-kubernetes/1265745915679870979

and is not working :/

aojea:
2 jobs timed out without logs :/
https://github.com/kubernetes-sigs/kind/pull/1627#issuecomment-634951877

BenTheElder:
/hold cancel
there's other reasons we may not work during timeouts, this change is still correct

aojea:
/retest
then 😄

BenTheElder:
notably I'm concerned about how podutils / our config is handling these, I think we need a longer grace period.


>May 27 21:06:34.545: INFO: Running AfterSuite actions on all nodes
>
>{"component":"entrypoint","file":"prow/entrypoint/run.go:164","func":"k8s.io/test-infra/prow/entrypoint.Options.ExecuteProcess","level":"error","msg":"Process did not finish before 40m0s timeout","time":"2020-05-27T21:29:25Z"}
> {"component":"entrypoint","file":"prow/entrypoint/run.go:245","func":"k8s.io/test-infra/prow/entrypoint.gracefullyTerminate","level":"error","msg":"Process did not exit before 15s grace period","time":"2020-05-27T21:29:40Z"}

note those times, 15s is not even enough time for the ginkgo after-suite to finish.
the GCE jobs are on bootstrap.py

BenTheElder:
/test all
seems to be working as intended so far, though we haven't seen the timeout behavior

aojea:
> /test all
> seems to be working as intended so far, though we haven't seen the timeout behavior

Jordan merged a PR related to the test with timeouts, I don't think it will show up again, if I read it correctly the test should timeout now, not the job

BenTheElder:
timeouts are happening on other branches too.
/test all

BenTheElder:
At the very least this doesn't misbehave so far, merging so we can maybe start collecting logs elsewhere

PR #1628, opened 2020-05-28
Add host path mount support for k8s e2e test script

saschagrunert:
If the environment variable `$EXTRA_MOUNTS` is set, then we now add
`extraMounts` to the kind configuration in the format
`HOST_PATH:CONTAINER_PATH[,…]`.

This allows us to use custom localhost seccomp profiles for extended e2e
testing in Kubernetes.

For example, if `EXTRA_MOUNTS="host1:ctr1,host2:ctr2"`, then the
resulting cluster config would look like this:

```yaml
kind: Cluster
networking:
  ipFamily: ipv4
nodes:
- role: control-plane
  extraMounts:
  - hostPath: host1
    containerPath: /ctr1
  - hostPath: host2
    containerPath: /ctr2
- role: worker
  extraMounts:
  - hostPath: host1
    containerPath: /ctr1
  - hostPath: host2
    containerPath: /ctr2
- role: worker
  extraMounts:
  - hostPath: host1
    containerPath: /ctr1
  - hostPath: host2
    containerPath: /ctr2
```

As a follow-up, I would like:

1. Add static seccomp test profiles to k/k
1. Modify the `pull-kubernetes-e2e-kind` job to set the `EXTRA_MOUNTS` to point to `<kubelet-root-dir>/seccomp`
1. Add additional e2e seccomp GA tests which now use those localhost profiles

Refers to https://github.com/kubernetes/kubernetes/pull/91442
/cc @pjbgf @hasheddan @evrardjp

### DIFF
--- a/hack/ci/e2e-k8s.sh
+++ b/hack/ci/e2e-k8s.sh
@@ -25,7 +25,9 @@ set -o errexit -o nounset -o xtrace
 # BUILD_TYPE: bazel or make
 # GA_ONLY: true  - limit to GA APIs/features as much as possible
 #          false - (default) APIs and features left at defaults
-# 
+# EXTRA_MOUNTS: optional host path mounts to be attached to each node in the
+#               format: "HOST_PATH:CONTAINER_PATH,..."
+#
 
 # our exit handler (trap)
 cleanup() {
@@ -72,6 +74,15 @@ build() {
   make all WHAT='cmd/kubectl test/e2e/e2e.test vendor/github.com/onsi/ginkgo/ginkgo'
 }
 
+# echo extraMounts for the provided parameter
+add_extra_mounts() {
+    if [ -n "$1" ]; then
+        printf "\n  extraMounts:\n"
+        echo "$1" |
+            sed -r "s;(\w*):(\w*)?,?;  - hostPath: \1\n    containerPath: /\2\n;g"
+    fi
+}
+
 # up a cluster with kind
 create_cluster() {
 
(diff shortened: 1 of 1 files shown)


saschagrunert:
Hm, I don't think that the latest failures are related.
/retest

hack/ci/e2e-k8s.sh
```diff
@@ -122,9 +135,9 @@ apiVersion: kind.x-k8s.io/v1alpha4
 networking:
   ipFamily: ${IP_FAMILY:-ipv4}
 nodes:
-- role: control-plane
-- role: worker
-- role: worker
+- role: control-plane $(add_extra_mounts "$EXTRA_MOUNTS")
```
pjbgf, line 138: More of a question: would it be worth it to have different mounts for control-plane and worker notes?

saschagrunert: Hm, maybe for even more enhanced testing. But this would increase the complexity. I just thought that we statically mount only one seccomp test-profile dir for now.

BenTheElder:
/hold

I'm very supportive of expanding the testing we can do :upside_down_face: 

However, I want to stop expanding the environment variable "API surface" especially with #1269 pending

Does this work as expected? What happens if the host doesn't have selinux enabled (or does?)

saschagrunert:
> /hold
> 
> I'm very supportive of expanding the testing we can do 
> 
> However, I want to stop expanding the environment variable "API surface" especially with #1269 pending
> 
> Does this work as expected? What happens if the host doesn't have selinux enabled (or does?)

Thank you for the hint, I think we could integrate the feature on top #1269 of there:
https://github.com/kubernetes-sigs/kind/pull/1269/files#diff-1d7c2061d0c44e9273294f69d17cd09dR277

Do you have any time-schedule when it will be supported? If not, then I think we do not have to rush and I can see if I can integrate my test-feature into the GCE tests (GCI image).

saschagrunert:
Closing in favor of an alternative approach tested in: https://github.com/kubernetes/kubernetes/pull/91571#issuecomment-636099782

If we still want the host paths added I will follow-up on top of #1269

Issue #1555, opened 2020-05-04
HA cluster with kind on Windows errors on copying certificate.


nielsvbrecht:
**What happened**:
When i try to create a cluster with three control-plane nodes I get an error on copying the ca certificate to the 2nd control plane node.

ERROR: failed to create cluster: failed to copy admin kubeconfig: failed to write "/etc/kubernetes/pki/ca.crt" to node: command "docker exec --privileged -i cluster1-control-plane2 cp /dev/stdin /etc/kubernetes/pki/ca.crt" failed with error: exit status 1

Command Output: cp: cannot create regular file '/etc/kubernetes/pki/ca.crt': No such file or directory

Stack Trace:
sigs.k8s.io/kind/pkg/errors.WithStack
        /src/pkg/errors/errors.go:51
sigs.k8s.io/kind/pkg/exec.(*LocalCmd).Run
        /src/pkg/exec/local.go:124
sigs.k8s.io/kind/pkg/cluster/internal/providers/docker.(*nodeCmd).Run
        /src/pkg/cluster/internal/providers/docker/node.go:146
sigs.k8s.io/kind/pkg/cluster/nodeutils.CopyNodeToNode
        /src/pkg/cluster/nodeutils/util.go:70
sigs.k8s.io/kind/pkg/cluster/internal/create/actions/kubeadminit.(*action).Execute
        /src/pkg/cluster/internal/create/actions/kubeadminit/init.go:95
sigs.k8s.io/kind/pkg/cluster/internal/create.Cluster
        /src/pkg/cluster/internal/create/create.go:135
sigs.k8s.io/kind/pkg/cluster.(*Provider).Create
        /src/pkg/cluster/provider.go:138
sigs.k8s.io/kind/pkg/cmd/kind/create/cluster.runE
        /src/pkg/cmd/kind/create/cluster/createcluster.go:91
sigs.k8s.io/kind/pkg/cmd/kind/create/cluster.NewCommand.func1
        /src/pkg/cmd/kind/create/cluster/createcluster.go:56
github.com/spf13/cobra.(*Command).execute
        /go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:842
github.com/spf13/cobra.(*Command).ExecuteC
        /go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:950
github.com/spf13/cobra.(*Command).Execute
        /go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:887
[...]

PR #1630, opened 2020-05-28
use path package for container paths

BenTheElder:
cc @aojea @amwat 

Note that `filepath` should only be used for local paths on the local host (e.g. when exporting logs to the current machine), `path` should be used for container paths.

Fixes https://github.com/kubernetes-sigs/kind/issues/1555


BenTheElder:
CI is having issues again, I'm looking into it but have not found the root cause yet.

PR #1639, opened 2020-05-31
Do the same kind of cgroup fix also for podman

afbjorklund:
The mounts that were fixed for docker, also needs fixing
for podman (which uses the name "libpod_parent" for them)

This is needed for minikube, that supports podman as well
Basically it is the same as docker, with 6 characters changed

See https://github.com/kubernetes/minikube/issues/8033

### DIFF
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -70,6 +70,18 @@ fix_cgroup() {
       mount --bind "${subsystem}" "${subsystem}${docker_cgroup}"
     done
   fi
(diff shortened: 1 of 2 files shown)


BenTheElder:
thanks for the PR!
/ok-to-test

BenTheElder:
FYI @amwat @aojea

BenTheElder:
failure is a bug in the CI which has a patch set to roll out monday + some timeouts due to resource contention issues in CI. ignoring.

BenTheElder:
bypassing CI issues.

PR #1640, opened 2020-05-31
Look for update-alternatives before using it

afbjorklund:
It is only used with dpkg, not with other systems such as
buildroot (that already has iptables legacy as the default)

I experimented with doing an image based on "buildroot",
which uses busybox rather than ubuntu for the packages.

See https://github.com/kubernetes/minikube/issues/6942

Also had issues with "hostname --ip-address", but it seems gone

### DIFF
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -150,8 +150,10 @@ select_iptables() {
   fi
 
   echo "INFO: setting iptables to detected mode: ${mode}" >&2
+  if [ -x /usr/bin/update-alternatives ]; then
   update-alternatives --set iptables "/usr/sbin/iptables-${mode}" > /dev/null
   update-alternatives --set ip6tables "/usr/sbin/ip6tables-${mode}" > /dev/null
+  fi
 }
 
 enable_network_magic(){


images/base/files/usr/local/bin/entrypoint
```diff
@@ -150,8 +150,10 @@ select_iptables() {
   fi
 
   echo "INFO: setting iptables to detected mode: ${mode}" >&2
+  if [ -x /usr/bin/update-alternatives ]; then
```
BenTheElder, line 153: not setting the correct mode is an issue though, in BOTH directions

BenTheElder:
busybox is broken with kubernetes last I checked, e.g. `mount` works improperly ...

It seems like it would also be possible to place a fake / custom `update-alternatives` in this modified image.

I'd prefer not to silently not update the iptables to match, failing to do so is a bug and will cause problems without surfacing them as obviously.

afbjorklund:
The image was based on the new minikube.iso, if you want to check it out...

Basically we are doing the equivalent of update-alternatives at build time.

afbjorklund:
> busybox is broken with kubernetes last I checked, e.g. `mount` works improperly ...

This was only for hosting the image, kubernetes itself runs inside the kindbase/kicbase

> It seems like it would also be possible to place a fake / custom `update-alternatives` in this modified image.

That would work too

afbjorklund:
> > busybox is broken with kubernetes last I checked, e.g. `mount` works improperly ...
> 
> This was only for hosting the image, kubernetes itself runs inside the kindbase/kicbase

Hmm, wrote too fast...

I think we replaced _some_ of the busybox with other variants, like util-linux and friends.

This was indeed intended for the image itself, too many wrappers to keep them apart :-)

afbjorklund:
Note that we are still running the docker containers with ubuntu image, just like kind.

This was an experiment, to see if we could have the _same_ OS for both VM and KIC

aojea:
this was already considered in the original iptables detection scripts
https://github.com/kubernetes-sigs/iptables-wrappers/blob/5be13bb4f92d23b3bc963857b163105f00740a6e/iptables-wrapper-installer.sh#L43-L53

afbjorklund:
I suppose a fedora-based image is also a possibility, in the sake of diversity (like podman)

BenTheElder:
> I suppose a fedora-based image is also a possibility, in the sake of diversity (like podman)

That seems like an enormous waste of resources. The OS of the image is an implementation detail, we're definitely not going to maintain against multiple OS-es for the image contents.

As noted in https://github.com/kubernetes-sigs/kind/pull/1640#issuecomment-636512052 this behavior was considered and rejected in the upstream iptables-wrapper, and as mentioned previously failing to set this is a major bug.
Failing to implement this is going to make the image unusable on many hosts soon. Plenty of users are sticking to legacy mode for compatibility, but most large distros are shipping nft mode by default in their stable releases.

afbjorklund:
> > I suppose a fedora-based image is also a possibility, in the sake of diversity (like podman)
> 
> That seems like an enormous waste of resources. The OS of the image is an implementation detail, we're definitely not going to maintain against multiple OS-es for the image contents.

In one aspect you are right, both podman and cri-o _are_ an enormous waste of resources...
It would be much easier to just ignore the CRI and run everything on docker and containerd.

And so far we haven't added any infrastructure for actually testing on CentOS (or Fedora)
So it just breaks as a part of the so called "none" driver, and we let `kubeadm` worry about it.

There are multiple OS for minikube being used at the moment, but we might change that soon.

But for **this** issue, I'm just going to add a fake `update-alternatives` script for Debian compat.

images/base/files/usr/local/bin/entrypoint
afbjorklund: Turns out that this functionality is handled by a simple symlink

afbjorklund:
> As noted in [#1640 (comment)](https://github.com/kubernetes-sigs/kind/pull/1640#issuecomment-636512052) this behavior was considered and rejected in the upstream iptables-wrapper, and as mentioned previously failing to set this is a major bug.
>
> Failing to implement this is going to make the image unusable on many hosts soon. Plenty of users are sticking to legacy mode for compatibility, but most large distros are shipping nft mode by default in their stable releases.

Right now it _only_ has "legacy", but it turns out that it is simple to implement this "update-alternatives" once both binaries are present. It is just a question of where the `/usr/sbin/iptables` symlink is pointing to, whether it is xtables-legacy-multi or xtables-nft-multi

Just have to set `BR2_PACKAGE_IPTABLES_NFTABLES=y`, in order to actually build it without `--disable-nftables`. But then it should be trivial to support a Debian compability script, if you don't want to make it distribution-agnostic (like that code snippet)

https://github.com/kubernetes-sigs/iptables-wrappers/blob/5be13bb4f92d23b3bc963857b163105f00740a6e/iptables-wrapper-installer.sh#L144_L170

```sh
# Update links to point to the selected binaries
for cmd in iptables iptables-save iptables-restore ip6tables ip6tables-save ip6tables-restore; do
    rm -f "${sbin}/\${cmd}"
    ln -s "${sbin}/xtables-\${mode}-multi" "${sbin}/\${cmd}"
done

```

PR #1642, opened 2020-06-01
ExportKubeConfig: provide option to not update kubecontext

thavlik:
Tests are updated and passing. There wasn't any discussion about this, so perhaps it's not the way you want it. If you'd prefer a separate function, e.g. ExportKubeConfigNoChangeContext, I'd be happy to submit the PR.

Thanks

### DIFF
--- a/pkg/cluster/internal/create/create.go
+++ b/pkg/cluster/internal/create/create.go
@@ -155,7 +155,7 @@ func Cluster(logger log.Logger, p provider.Provider, opts *ClusterOptions) error
 		return nil
 	}
 
-	if err := kubeconfig.Export(p, opts.Config.Name, opts.KubeconfigPath); err != nil {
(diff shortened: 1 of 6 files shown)


thavlik:
I signed the CLA. To clarify, `go test` runs, but the Makefile tests do not respect $GOPATH.

thavlik:
```
bentheelder:palm_tree:  36 minutes ago
For generic KUBECONFIG management I'd point to client-go, currently the to API surface is only minimally targeted for resuse, mostly by some kubeadm development. We stopped advertising the API in the docs long ago.




bentheelder:palm_tree:  35 minutes ago
It's brittle enough reimplementing KUBECONFIG management as is (there are open issues) but KIND has a  chicken and egg problem with using core k8s code

Tom Havlik  < 1 minute ago
Gotcha
```

Issue #1650, opened 2020-06-04
[proxy] Ingress nginx example doesn't work - failed calling webhook (Forbidden) while creating Ingress


ddziuda:
**What happened**:
Error from server (InternalError): error when creating "https://kind.sigs.k8s.io/examples/ingress/usage.yaml": Internal error occurred: failed calling webhook "validate.nginx.ingress.kubernetes.io": Post https://ingress-nginx-controller-admission.ingress-nginx.svc:443/extensions/v1beta1/ingresses?timeout=30s: Forbidden

**What you expected to happen**:
Working ingress nginx example.

**How to reproduce it (as minimally and precisely as possible)**:
Just to follow [Ingress nginx example](https://kind.sigs.k8s.io/docs/user/ingress/).

**Anything else we need to know?**:
I'm behind a proxy. I raised #1579 previously and I'm using kind from HEAD - kind v0.9.0-alpha go1.14.3 linux/amd64.

**Environment:**

- kind version: (use `kind version`): kind v0.9.0-alpha go1.14.3 linux/amd64
- Kubernetes version: (use `kubectl version`): 1.18
- Docker version: (use `docker info`): 19.03.2
- OS (e.g. from `/etc/os-release`): CentOS Linux 7 (Core)

Logs look ok for me:
```
[iid@iid helm]$ kubectl wait --namespace ingress-nginx   --for=condition=ready pod   --selector=app.kubernetes.io/component=controller   --timeout=90s
pod/ingress-nginx-controller-cc8dd9868-sgr4x condition met
[iid@iid helm]$ kubectl get pods -n ingress-nginx
NAME                                       READY   STATUS      RESTARTS   AGE
ingress-nginx-admission-create-rktpn       0/1     Completed   0          37s
ingress-nginx-admission-patch-vp5db        0/1     Completed   1          37s
ingress-nginx-controller-cc8dd9868-sgr4x   1/1     Running     0          47s
[iid@iid helm]$ kubectl logs ingress-nginx-admission-create-rktpn -n ingress-nginx
{"err":"secrets \"ingress-nginx-admission\" not found","level":"info","msg":"no secret found","source":"k8s/k8s.go:101","time":"2020-06-04T12:21:17Z"}
{"level":"info","msg":"creating new secret","source":"cmd/create.go:23","time":"2020-06-04T12:21:17Z"}
[iid@iid helm]$ kubectl logs ingress-nginx-admission-patch-vp5db -n ingress-nginx
[...]

PR #1661, opened 2020-06-10
Add in-cluster pod and services dns names to no_proxy

amwat:
ref: https://github.com/kubernetes-sigs/kind/issues/1650#issuecomment-639793233
xref: https://github.com/kubernetes/kubernetes/pull/91651#issuecomment-641319426

### DIFF
--- a/pkg/cluster/internal/providers/docker/provision.go
+++ b/pkg/cluster/internal/providers/docker/provision.go
@@ -294,6 +294,9 @@ func getProxyEnv(cfg *config.Cluster, networkName string, nodeNames []string) (m
 
 		noProxyList := append(subnets, envs[common.NOProxy])
 		noProxyList = append(noProxyList, nodeNames...)
+		// Add .svc explicitly to no_proxy to allow in cluster
+		// pod and service dns resolution
+		noProxyList = append(noProxyList, ".svc")
 		noProxyJoined := strings.Join(noProxyList, ",")
 		envs[common.NOProxy] = noProxyJoined
 		envs[strings.ToLower(common.NOProxy)] = noProxyJoined
--- a/pkg/cluster/internal/providers/docker/provision.go
+++ b/pkg/cluster/internal/providers/docker/provision.go
@@ -294,9 +294,12 @@ func getProxyEnv(cfg *config.Cluster, networkName string, nodeNames []string) (m
 
 		noProxyList := append(subnets, envs[common.NOProxy])
 		noProxyList = append(noProxyList, nodeNames...)
-		// Add .svc explicitly to no_proxy to allow in cluster
-		// pod and service dns resolution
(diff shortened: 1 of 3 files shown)


pkg/cluster/internal/providers/docker/provision.go
```diff
@@ -294,6 +294,9 @@ func getProxyEnv(cfg *config.Cluster, networkName string, nodeNames []string) (m
 
 		noProxyList := append(subnets, envs[common.NOProxy])
 		noProxyList = append(noProxyList, nodeNames...)
+		// Add .svc explicitly to no_proxy to allow in cluster
+		// pod and service dns resolution
+		noProxyList = append(noProxyList, ".svc")
```
aojea, line 299: I think that we have to no_proxy the cluster domain , if we look at the standard coredns config 
```yaml
apiVersion: v1
data:
  Corefile: |
    .:53 {
        errors
        health {
           lameduck 5s
        }
        ready
        kubernetes cluster.local in-addr.arpa ip6.arpa {
           pods insecure
           fallthrough in-addr.arpa ip6.arpa
           ttl 30
        }
        prometheus :9153
        forward . /etc/resolv.conf
        cache 30
        loop
        reload
        loadbalance
    }
```
we should add `.cluster.local`

amwat: yup, thanks for linking to the spec 👍 
I also found a couple of issues related to this: https://github.com/kubernetes/kubeadm/issues/2114

I think we'll need to add all of those since k8s also resolves partially qualified names
`.svc,.svc.cluster,.svc.cluster.local` especially for things like admission webhook which only use `.svc` by default https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#service-reference, and this still won't solve the problem where services should be resolvable through `http://<service>` and `http://<service>.<ns>`

Also, I wasn't sure if `cluster.local` is static or something that we should be picking up from the coredns config, but looks like it is the default.

aojea: > I think we'll need to add all of those since k8s also resolves partially qualified names
> `.svc,.svc.cluster,.svc.cluster.local` especially for things like admission webhook which only use `.svc` by default https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#service-reference, and this still won't solve the problem where services should be resolvable through `http://<service>` and `http://<service>.<ns>`

TIL, thanks for the info

> Also, I wasn't sure if cluster.local is static or something that we should be picking up from the coredns config, but looks like it is the default.

it is not, one example is the kubernetes CI that changes it, but I think that for kind environments we should be safe, maybe just add a comment mentioning this, don't know , an environment with proxies and changing the default kind domain seems a very specific case

BenTheElder:
sorry for the wait
/lgtm
/approve
thanks for this!

PR #1669, opened 2020-06-16
SCTP

BenTheElder:
xref: https://github.com/kubernetes/kubernetes/issues/92041
picks up containerd with the CRI patch to enable SCTP


BenTheElder:
looks like the test only runs on kubenet though ... asking about that https://github.com/kubernetes-sigs/kind/pull/1669

PR #1675, opened 2020-06-18
[WIP] disable build with bazel in kubernetes CI

BenTheElder:
we're seeing lots of issues with bazel caching.
testing just removing it
cc @dims @aojea 

if this works out, we should consider it across the other e2e jobs.

### DIFF
--- a/hack/ci/e2e-k8s.sh
+++ b/hack/ci/e2e-k8s.sh
@@ -244,11 +244,13 @@ main() {
   BUILD_TYPE="${BUILD_TYPE:-bazel}"
 
   # build kubernetes
-  if [ "${BUILD_TYPE:-}" = "bazel" ]; then
-    build_with_bazel
-  else
-    build
-  fi
+  #if [ "${BUILD_TYPE:-}" = "bazel" ]; then
+  #  build_with_bazel
+  #else
+  #  build
+  #fi
+  echo "WARNING: building without bazel due to CI breakage / ignoring BUILD_TYPE"
+  build
 
   # in CI attempt to release some memory after building
   if [ -n "${KUBETEST_IN_DOCKER:-}" ]; then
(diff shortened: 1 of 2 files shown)


BenTheElder:
we're still running the tests without the new skips here.
most of these jobs also did not get the expanded CPU request.

BenTheElder:
that's good to know, the default GOFLAGS is unsafe on k8s 1.15 build, we need to gate https://github.com/kubernetes-sigs/kind/pull/1576 on version.

https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/1675/pull-kind-conformance-parallel-1-15/1273658488270098434

BenTheElder:
https://github.com/kubernetes/test-infra/pull/17992

sigh

dims:
LGTM

BenTheElder:
if this works out instead of this blunt change maybe we should roll it out job by job I think, (in pairs, matching kind jobs and k/k jobs each time).

still using this PR to collect data at least

BenTheElder:
/retest
pod pending timeouts (prow)

BenTheElder:
most of these went pretty quick.
the job on master went slow, but we don't have great insight into the timings.
added some `time` calls

BenTheElder:
ipv6 on master only took 25m, so I'm thinking bazel is really not helping here.

BenTheElder:
> /home/prow/go/src/sigs.k8s.io/kind/hack/ci/e2e-k8s.sh: 253: /home/prow/go/src/sigs.k8s.io/kind/hack/ci/e2e-k8s.sh: time: not found

... why did I implement this in POSIX sh.
#1269

BenTheElder:
bazel build times serious leveled off in the other jobs (not instrumented in kind yet because bash...)

https://testgrid.k8s.io/presubmits-kubernetes-nonblocking#pull-kubernetes-e2e-gce-device-plugin-gpu&graph-metrics=test-duration-minutes&include-filter-by-regex=Build&width=2
speed ~3m again ...

PR #1676, opened 2020-06-18
use iptables-nft to avoid xtables.lock

aojea:
kube-proxy images uses the same version that is using in the host where it is running, so it should pick nft if we default the nodes to nft.

iptables-legacy use an external lock 
```
openat(AT_FDCWD, "/run/xtables.lock", O_RDONLY|O_CREAT, 0600) = 4
flock(4, LOCK_EX|LOCK_NB)  
```

iptables-nft does not, and is namespaced, that should avoid interference when running cluster in different pods 

```
iptables vs. iptables-nft vs. nft
In comparison to legacy iptables, iptables-nft has a few clear benefits: With back end transactions being atomic, there is no need for the global xtables lock which has proven problematic in environments with large and/or rapidly changing rulesets. 
```
xref https://www.redhat.com/en/blog/using-iptables-nft-hybrid-linux-firewall

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -79,6 +79,9 @@ RUN echo "Ensuring scripts are executable ..." \
     && rm -f /lib/systemd/system/basic.target.wants/* \
     && echo "ReadKMsg=no" >> /etc/systemd/journald.conf \
     && ln -s "$(which systemd)" /sbin/init \
+ && echo "Using iptables-nft ..." \
+    && update-alternatives --set iptables /usr/sbin/iptables-nft \
+    && update-alternatives --set ip6tables /usr/sbin/ip6tables-nft \
  && echo "Installing containerd ..." \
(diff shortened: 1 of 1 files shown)


aojea:
/hold
this needs testing, I'm just guessing here

images/base/Dockerfile
```diff
@@ -79,6 +79,9 @@ RUN echo "Ensuring scripts are executable ..." \
     && rm -f /lib/systemd/system/basic.target.wants/* \
     && echo "ReadKMsg=no" >> /etc/systemd/journald.conf \
     && ln -s "$(which systemd)" /sbin/init \
+ && echo "Using iptables-nft ..." \
```
BenTheElder, line 82: should we not do this in the job level?
the docker network stuff injects rules to the kind "node"

BenTheElder:
discussed offline, has to be done elsewhere.

PR #1677, opened 2020-06-18
Revert "Add support for dockerless builds in kind"

BenTheElder:
temporarily, until we sort out version gating this, I need to unblock #1319, this PR broke building old releases under default `kind build node-image` without being noticed, somehow. will investigate later.


BenTheElder:
FYI @dims 
you can still override this by setting `GOFLAGS` env when calling it, IIRC. host env is passed after the env we set and last wins.

PR #1678, opened 2020-06-19
use iptables-nft to avoid xtables.lock

aojea:
iptables-nft and iptables-legacy are compatible, however, kubernetes
needs that all the components: kubelet and kube-proxy agree on the
userspace tools.

In addition, Docker insert rules in the container to do some "magic"
with the DNS that we modify in order to survive reboots.

We use nft always for kubernetes and pick the same version used by the
host to configure the magic DNS rules.


aojea:
@BenTheElder it works for me locally, 
can you give it a spin in the CI?

aojea:
/hold

I have to address some comments from @danwinship about nft dependencies on user space and kernel

aojea:
@BenTheElder is possible the CI nodes does not have the nft modules?

```
lsmod | grep nft
nft_chain_nat          16384  24
nft_counter            16384  220
nft_compat             20480  395
nf_tables             135168  583 nft_compat,nft_counter,nft_chain_nat
nfnetlink              16384  8 nft_compat,nfnetlink_cttimeout,nf_conntrack_netlink,nf_tables
nf_nat                 40960  6 ip6table_nat,xt_nat,openvswitch,nft_chain_nat,iptable_nat,xt_MASQUERADE
x_tables               40960  18 ip6table_filter,xt_conntrack,xt_statistic,iptable_filter,nft_compat,xt_tcpudp,xt_addrtype,xt_recent,xt_nat,xt_comment,ip6_tables,ipt_REJECT,ip_tab
les,ip6table_mangle,xt_MASQUERADE,ip6t_REJECT,iptable_mangle,xt_mark
```

danwinship:
> iptables-nft and iptables-legacy are compatible, however, kubernetes
> needs that all the components: kubelet and kube-proxy agree on the
> userspace tools.

It's not just that; _everything_ on the "host" needs to agree. You need to choose the legacy/nft default before anything runs in the container that uses iptables. (ie, you need to make sure docker is using the same mode as kubernetes is)

> is possible the CI nodes does not have the nft modules?

`nf_tables` is the important one; most of the others are shared between legacy and nft iptables.

aojea:
> > iptables-nft and iptables-legacy are compatible, however, kubernetes
> > needs that all the components: kubelet and kube-proxy agree on the
> > userspace tools.
> 
> It's not just that; _everything_ on the "host" needs to agree. You need to choose the legacy/nft default before anything runs in the container that uses iptables. (ie, you need to make sure docker is using the same mode as kubernetes is)

heh, it's magic, I was thinking the same, but it turns out it does not require everything using the same mode.
I have this patch working locally and passing e2e tests
```
$ docker exec -it kind-control-plane iptables-nft-save | grep KUBE | wc
     75     728    6699
$ docker exec -it kind-control-plane iptables-nft-save | grep DOCKER | wc
      0       0       0
$ docker exec -it kind-control-plane iptables-legacy-save | grep DOCKER | wc
      9      80     610
$ docker exec -it kind-control-plane iptables-legacy-save | grep KUBE | wc
      0       0       0
``` 
and the e2e network tests are passing (Those 6 failures are unrelated)

```
Ran 44 of 5070 Specs in 8834.317 seconds
FAIL! -- 38 Passed | 6 Failed | 0 Pending | 5026 Skipped
```

BenTheElder:
You say you have it locally, but what does your host / docker use?

I don't think this change is safe. We should reconfigure it at the level above (in the prowjob) to nft as I previously discussed.

That should be a simple, safe change that we only toggle for jobs running k8s 1.17+
Remember that we have CI that is not prepared for iptables_nft at all yet (1.16).

aojea:
this is too hacky, funny thing it worked but as Dan suggested this is not expected to be used this way

danwinship:
> You say you have it locally, but what does your host / docker use?

OK, this is confusing because of the whole multiple-levels-of-virtualization thing. When I said 'everything on the "host" needs to agree', I didn't mean that what's used inside kind has to match what's used outside kind; I just meant that within the kind environment you have to use only a single mode.

Maybe I should have said "within any given kernel network namespace, everything needs to agree (but different namespaces don't need to agree with each other)".

>  Remember that we have CI that is not prepared for iptables_nft at all yet (1.16).

well, the later version of Antonio's patch would have still used iptables-legacy on systems where iptables-nft wasn't available or didn't work

BenTheElder:
> OK, this is confusing because of the whole multiple-levels-of-virtualization thing. When I said 'everything on the "host" needs to agree', I didn't mean that what's used inside kind has to match what's used outside kind; I just meant that within the kind environment you have to use only a single mode.

right, the problem is that during container creation (talking about the kind nodes) `docker` use the "host" iptables (the prowjob here, or antonio's local machine / OS locally), to inject some rules.

so at least some of the rules on the "host" (kind "node") are going to depend on the iptables from wherever the docker used to create the nodes is running.

that is being set in https://github.com/kubernetes/test-infra/pull/18019

BenTheElder:
er and since these rules are related to our "upstream DNS resolver" / for resolving node name to IP we probably don't want those to not agree with e.g. kubelet rules.

BenTheElder:
we perform iptables / nft auto selection in the container entrypoint prior to starting systemd, to match the injected rules from docker (and actually modify them slightly..)

aojea:
for reference
my host uses iptables-legacy -> 
docker creates kind containers and insert the dns iptables rules using iptables-legacy ->
kind nodes use iptables-nft by default -> 
k8s inside kind use iptables nft only

my observation is that the cluster worked, I ran several e2e test and it was working, I restarted the cluster and it was working
it may have problems, I don't know 😄 , but fact is that the cluster was functional despite the weird iptables setup 🙃

PR #1680, opened 2020-06-20
add circle macOS CI

BenTheElder:
/hold


BenTheElder:
circle's OSS plan does not have macos

PR #1683, opened 2020-06-22
Improve help message

zhijianli88:


pkg/cmd/kind/load/image-archive/image-archive.go
```diff
@@ -49,7 +49,7 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 			}
 			return nil
 		},
-		Use:   "image-archive",
+		Use:   "image-archive /path/to/tar-archive-file",
 		Short: "Loads docker image from archive into nodes",
```
BenTheElder, line 53: i think it would be better to put `IMAGE-PATH` or something more consistent with the other command but /shrug

BenTheElder:
/hold cancel
/retest
Can tweak in follow up

pkg/cmd/kind/load/image-archive/image-archive.go
zhijianli88: it sounds good, i have changed it to <IMAGE.tar>[1]
[1]: <> is a [Unix-style](https://en.wikipedia.org/wiki/Usage_messag)

PR #1684, opened 2020-06-22
Remove duplicated Loadimage

zhijianli88:


BenTheElder:
Thanks but this wasn't exported in it's current state Intentionally and was duplicated intentionally (it's a very small bit of code and only twice).

This isn't ready to be a public API in it's current form.

This project follows WET -- Write Everything Thrice before DRYing things out.  

It does not satisfy the TODO, which is about having it with the other cluster level methods / part of the cluster's API, which needs more thought (e.g. how are nodes selected?).

Issue #1468, opened 2020-04-09
Allow users to configure MTU on kindnet


faiq:
**What would you like to be added**:

There should be an easy way to configure MTU values on the bridge that kind net creates.

**Why is this needed**

We're using KIND heavily for our CI testing purposes. Our CI workflows run on Kubernetes and in order to get kind to run properly we run a pod with a dind container that starts kind and our tests run against this cluster. We were running into many issues running our tests that make network request from this docker in docker in docker environment. After hours of debugging we realized that the MTU was configured too high and once we ran our inner most container with the following args 

 `args: ["-c", "ip link set mtu 1400 dev eth0 && /manager --enable-leader-election"]`

the issue was mitigated.

If we were able to pass in some configuration via our kind-config.yaml it would be greatly appreciated! 

Please let me know if you'd like me to add in more information.

Thank you for you time and consideration.

BenTheElder:
cc @aojea 
/priority backlog

aojea:
>  Our CI workflows run on Kubernetes

what CNI uses that Kubernetes? is it an overlay?

Because you should not have MTU issues if there is no overlay involved, because kindnet does not encapsulate anything, neither docker does. 

Anyway, kindnet uses the ptp CNI plugin and it does have an option to configure the MTU :smile: 

https://github.com/containernetworking/plugins/tree/master/plugins/main/ptp

aojea:
Ok, I've tested and the MTU option works, we just need to add `"mtu": 1400, ` to the CNI config manifest

https://github.com/kubernetes-sigs/kind/blob/54471c23b742b75c8c9baaa6979bd63576c618f3/images/kindnetd/cmd/kindnetd/cni.go#L60-L62

@BenTheElder what will be the easier way to make it configurable? 

passing it as an environment variable to kindnet and templating it, like we are doing with the POD_SUBNET?
https://github.com/kubernetes-sigs/kind/blob/54471c23b742b75c8c9baaa6979bd63576c618f3/pkg/build/node/cni.go#L157-L158

faiq:
@aojea I think passing it in via an environment variable would be completely fine :) i didn't realize that it was using the ptp module (or that it even existed) thanks so much for looking into this!

aojea:
> @aojea I think passing it in via an environment variable would be completely fine :) i didn't realize that it was using the ptp module (or that it even existed) thanks so much for looking into this!

[...]

PR #1686, opened 2020-06-23
WIP configure mtu 

faiq:
This is my attempt at solving the problem I was outlining in https://github.com/kubernetes-sigs/kind/issues/1468. I imagine that that the MTU should be configured via the networking object 

```yaml
kind: Cluster
networking:
  mtu: 1480
```

I tried following the example of POD_SUBNET outlined in this [comment](https://github.com/kubernetes-sigs/kind/issues/1468#issuecomment-614625988) Where it is passed in as an envvar and templated from there. 

Please let me know how all of this looks and give suggestions

### DIFF
--- a/images/kindnetd/cmd/kindnetd/cni.go
+++ b/images/kindnetd/cmd/kindnetd/cni.go
@@ -34,18 +34,22 @@ import (
 type CNIConfigInputs struct {
 	PodCIDR      string
 	DefaultRoute string
+	MTU          string
 }
 
 // ComputeCNIConfigInputs computes the template inputs for CNIConfigWriter
 func ComputeCNIConfigInputs(node corev1.Node) CNIConfigInputs {
 	podCIDR := node.Spec.PodCIDR
 	defaultRoute := "0.0.0.0/0"
+	defaultMtu := "1480"
+
 	if net.IsIPv6CIDRString(podCIDR) {
 		defaultRoute = "::/0"
 	}
 	return CNIConfigInputs{
 		PodCIDR:      podCIDR,
 		DefaultRoute: defaultRoute,
+		MTU:          defaultMtu,
 	}
 }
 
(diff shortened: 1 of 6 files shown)


BenTheElder:
I think possibly we should actually just be setting MTU to something like this by default.
Are you setting this to multiple different values?

`networking.mtu` seems like a misleading field name, as it won't apply to the node networking and it will only apply to pod networking when using the default CNI.

faiq:
> I think possibly we should actually just be setting MTU to something like this by default.

From my experience 1480 has been a good default value, but as it's your project you can decide


> Are you setting this to multiple different values?

I don't think I am! What I think I am doing is taking the information from `kind-config.yaml`, templating the relevant fields to `/kind/manifests/default-cni.yaml`, which puts them as environment variables. Finally,  the container that does the reconciling of the CNI, picks up the environment variable (if set) and templates it into the `"/etc/cni/net.d/10-kindnet.conflist"` 

However, there are some points where I think I might have gotten tripped up. 

ie: im not sure if this block is doing the templating of the yaml file as expected 

https://github.com/kubernetes-sigs/kind/pull/1686/files#diff-b62036194925aa18fc3a7b63f83eb890R70-R81


Furthermore, I'm unclear if the value [here](https://github.com/kubernetes-sigs/kind/pull/1686/files#diff-a13aa51a1611d98cfe638c15f201177cR185) in `v1alpha4/types.go` is picked up in [here](https://github.com/kubernetes-sigs/kind/pull/1686/files#diff-4686ef90e1f4b2babdb7c7bf7997b62dR140) `config/types.go` which is the start of the execution for the templating of the mtu value.


Regarding `networking.mtu` being misleading, I completely see that. What about `networking.kindNetMTU` which suggests to me that this is the value for the default CNI's MTU

Thanks,

Faiq

faiq:
I am closing this to open a pull request that configures the CNI's MTU from the node's interface directly as mentioned in this comment 
https://github.com/kubernetes-sigs/kind/issues/1468#issuecomment-649051368

PR #1688, opened 2020-06-25
containerd updates

BenTheElder:
- correct containerd sock
- upgrade to latest
- explicitly tolerate missing hugepages https://github.com/containerd/cri/pull/1501 https://github.com/containerd/cri/issues/1500

### DIFF
--- a/images/base/files/etc/crictl.yaml
+++ b/images/base/files/etc/crictl.yaml
@@ -1,1 +1,1 @@
-runtime-endpoint: unix:///var/run/containerd/containerd.sock
+runtime-endpoint: unix:///run/containerd/containerd.sock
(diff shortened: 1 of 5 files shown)


images/base/files/etc/crictl.yaml
```diff
@@ -1 +1 @@
-runtime-endpoint: unix:///var/run/containerd/containerd.sock
\ No newline at end of file
+runtime-endpoint: unix:///run/containerd/containerd.sock
```
aojea, line 1: heh, didn't know that
```
root@kind-control-plane:/# ls -al /var/run
lrwxrwxrwx 1 root root 4 Apr 10 20:34 /var/run -> /run
```

BenTheElder: Yeah, in our case and on most current distros I would expect them to be the same BUT I checked and containerd appears to nominally use /run canonically.

Lack of unix:// has left warnings in the kubelet logs.

PR #1690, opened 2020-06-25
Have CNI assume mtu from eth0

faiq:
As per my conversation with @aojea in the following [issue](https://github.com/kubernetes-sigs/kind/issues/1468#issuecomment-649051368) We figured the best path going forward would be to auto configure the MTU of kindnet from the host's device 

This pull request replaces the one here 
https://github.com/kubernetes-sigs/kind/pull/1686

### DIFF
--- a/images/kindnetd/cmd/kindnetd/cni.go
+++ b/images/kindnetd/cmd/kindnetd/cni.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"io"
+	stdnet "net"
 	"os"
 	"reflect"
 	"text/template"
@@ -34,12 +35,14 @@ import (
 type CNIConfigInputs struct {
 	PodCIDR      string
 	DefaultRoute string
+	Mtu          int
 }
 
 // ComputeCNIConfigInputs computes the template inputs for CNIConfigWriter
 func ComputeCNIConfigInputs(node corev1.Node) CNIConfigInputs {
 	podCIDR := node.Spec.PodCIDR
 	defaultRoute := "0.0.0.0/0"
+
 	if net.IsIPv6CIDRString(podCIDR) {
 		defaultRoute = "::/0"
 	}
@@ -47,6 +50,21 @@ func ComputeCNIConfigInputs(node corev1.Node) CNIConfigInputs {
 		PodCIDR:      podCIDR,
 		DefaultRoute: defaultRoute,
 	}
+}
+
+//computeBridgeMTU finds the mtu for the eth0 interface
+//otherwise it defults to ptp default behavior of being set by kernel
+func computeBridgeMTU() (int, error) {
+	interfaces, err := stdnet.Interfaces()
+	if err != nil {
+		return 0, err
+	}
+	for _, inter := range interfaces {
+		if inter.Name == "eth0" {
+			return inter.MTU, nil
+		}
+	}
+	return 0, errors.New("Found no eth0 device")
 }
 
 // cniConfigPath is where kindnetd will write the computed CNI config
@@ -76,6 +94,9 @@ const cniConfigTemplate = `
 			]
 		]
 		}
+		{{if .Mtu}},
+		"mtu": {{ .Mtu }}
+		{{end}}
 	},
 	{
 		"type": "portmap",
@@ -92,10 +113,12 @@ const cniConfigTemplate = `
 type CNIConfigWriter struct {
 	path       string
 	lastInputs CNIConfigInputs
+	mtu        int
 }
 
 // Write will write the config based on
 func (c *CNIConfigWriter) Write(inputs CNIConfigInputs) error {
+	inputs.Mtu = c.mtu
 	if reflect.DeepEqual(inputs, c.lastInputs) {
 		return nil
 	}
--- a/images/kindnetd/cmd/kindnetd/main.go
+++ b/images/kindnetd/cmd/kindnetd/main.go
@@ -70,9 +70,15 @@ func main() {
 		))
 	}
 
+	mtu, err := computeBridgeMTU()
+	klog.Infof("setting mtu %d for CNI \n", mtu)
+	if err != nil {
+		klog.Infof("Failed to get MTU size from interface eth0, using kernel default MTU size error:%v", err)
+	}
 	// used to track if the cni config inputs changed and write the config
 	cniConfigWriter := &CNIConfigWriter{
 		path: cniConfigPath,
+		mtu:  mtu,
 	}
 
 	// enforce ip masquerade rules
--- a/pkg/build/nodeimage/const_cni.go
+++ b/pkg/build/nodeimage/const_cni.go
@@ -20,7 +20,7 @@ package nodeimage
 The default CNI manifest and images are our own tiny kindnet
 */
 
-var defaultCNIImages = []string{"kindest/kindnetd:v20200619-15f5b3ab"}
+var defaultCNIImages = []string{"kindest/kindnetd:v20200707-3aec2c7e"}
 
 const defaultCNIManifest = `
 # kindnetd networking manifest
@@ -93,7 +93,7 @@ spec:
       serviceAccountName: kindnet
       containers:
       - name: kindnet-cni
-        image: kindest/kindnetd:v20200619-15f5b3ab
+        image: kindest/kindnetd:v20200707-3aec2c7e
         env:
         - name: HOST_IP
           valueFrom:


images/kindnetd/cmd/kindnetd/cni.go
```diff
@@ -75,7 +98,8 @@ const cniConfigTemplate = `
 				}
 			]
 		]
-		}
+		},
+		"mtu": {{ .Mtu }}
```
aojea, line 102: can you please check if this renders correctly? 
I never remember if it needs a trailing comma

BenTheElder: JSON should not have a trailing comma.

images/kindnetd/cmd/kindnetd/cni.go
```diff
@@ -28,27 +29,49 @@ import (
...
 }
 
+//computeBridgeMTU finds the mtu for the eth0 interface
+//otherwise it defults to a sensible value - 1480
+func computeBridgeMTU() int {
+	interfaces, err := stdnet.Interfaces()
+	if err != nil {
+		return defaultMtu
+	}
+	for _, inter := range interfaces {
```
aojea, line 67: let's make it smarter, 
we can default to eth0 but if there are any other interface with an ip we can use it
```go
		for _, i := range ifaces {
			addrs, err := i.Addrs()
			if err != nil {
				continue
			}

			for _, addr := range addrs {
				var ip net.IP
				switch v := addr.(type) {
				case *net.IPNet:
					ip = v.IP
				case *net.IPAddr:
					ip = v.IP
				}

				if ip.IsGlobalUnicast() {
					return i.MTU
				}
```

images/kindnetd/cmd/kindnetd/cni.go
```diff
@@ -28,27 +29,49 @@ import (
 	"k8s.io/utils/net"
 )
 
+//sensibleMTU value to fall back on - if we're not able to configure from eth0
+const defaultMtu = 1480
```
aojea, line 33: why do you think we should change the default?

images/kindnetd/cmd/kindnetd/cni.go
faiq: From my understanding, this `,` should be here. After looking at the example here https://github.com/containernetworking/plugins/tree/master/plugins/main/ptp#example-network-configuration

it looks like `mtu` is a field on the as the `ipam` dictionary.

images/kindnetd/cmd/kindnetd/cni.go
faiq: We could allow the kernel to set the default, as documented here: https://github.com/containernetworking/plugins/tree/master/plugins/main/ptp#example-network-configuration 


I've typically found that mtu is configured to 1500, which has been problematic for many people.

Maybe it should be 1400 instead.

aojea: I prefer to keep current behaviour, and allow the kernel to set it

images/kindnetd/cmd/kindnetd/cni.go
aojea: Dismiss this comment, let's go with current approach, this can be done Ina follow up, no need to complicate it more unnecessarily.
Sorry

images/kindnetd/cmd/kindnetd/cni.go
aojea: Yeah, it's ok, I was referring after MTU

BenTheElder:
explicit +1 on the approach, btw, since I neglected to put that in github anywhere.
will take a closer look soon.

aojea:
ping @faiq , it's only missing not defaulting the mtu to 1480, maybe just not inserting that field in the json if we can't get the MTU value from eth0 is enough.

please disregard my other comment https://github.com/kubernetes-sigs/kind/pull/1690#discussion_r445873639

faiq:
Thanks for the ping @aojea. I'll be completing this one today. I believe we just need to template it if we get an error from getting the MTU from `eth0`. We might need to log this error as well for anyone looking to debug why their MTU isn't being set. Thoughts?

aojea:
> Thanks for the ping @aojea. I'll be completing this one today. I believe we just need to template it if we get an error from getting the MTU from `eth0`. We might need to log this error as well for anyone looking to debug why their MTU isn't being set. Thoughts?

yeah, that sounds nice, what about something like
```
klog.Infof("Failed to get MTU size from interface eth0, using kernel default MTU size")
```

images/kindnetd/cmd/kindnetd/cni.go
```diff
@@ -34,19 +36,41 @@ import (
...
 	PodCIDR      string
 	DefaultRoute string
+	Mtu          int
 }
 
 // ComputeCNIConfigInputs computes the template inputs for CNIConfigWriter
 func ComputeCNIConfigInputs(node corev1.Node) CNIConfigInputs {
 	podCIDR := node.Spec.PodCIDR
 	defaultRoute := "0.0.0.0/0"
+	mtu, err := computeBridgeMTU()
```
aojea, line 46: @BenTheElder if I read it correctly this is going to be executed periodically, can you confirm?

I don't like the idea of iterating over the interfaces periodically, it can cause issues, I had issue with the net library before.
Should be better to try to get the mtu only once? in the main loop? and pass it as a parameter of ComputeCNIConfigInputs?

faiq: Actually, yes it will be computed multiple times and I think it's a more sound approach to compute this once.

images/kindnetd/cmd/kindnetd/cni.go
```diff
@@ -34,19 +36,47 @@ import (
 type CNIConfigInputs struct {
 	PodCIDR      string
 	DefaultRoute string
+	Mtu          int
+}
+
+var mtu int
+
+func init() {
```
BenTheElder, line 44: init() and global data is a big anti-pattern.
mtu should be plumbed down from main.

faiq: I didn't realize this was an anti-pattern, sorry. If you have a quick article on best practices, I'd love to read it! If not, no sweat. 


I can pull it down from main, but I think it would need to go in `makeNodesReconciler` as an extra parameter which then propagates to the `ComputeCNIConfigInputs` function. 

Alternatively we can put `mtu` as a field in `CNIConfigWriter` and in this method `func (c *CNIConfigWriter) Write(inputs CNIConfigInputs)` 

we can have `inputs.Mtu = c.mtu`

Let me know what you would prefer.

aojea: I lean to the `CNIConfigWriter` since mtu only matter to that type.
what do you think Ben?

BenTheElder: no sweat :-)
I would say "avoid globals" is a good practice for all languages, I don't have a reference handy just this moment though, sorry.

`init` is best avoided in code because it's both global behavior and unpredictable import side-effects (importing the package causes this code to be run during the binary startup, in some random order versus other packages)

CNIConfigWriter makes sense to me 👍

BenTheElder: https://stackoverflow.com/a/56039373 is a pretty decent short reference on this point specifically 😅

images/kindnetd/cmd/kindnetd/cni.go
```diff
@@ -76,6 +94,9 @@ const cniConfigTemplate = `
 			]
 		]
 		}
+		{{if .Mtu}},
```
aojea, line 97: if the eth0 interface is not found mtu will be 0, 
and {{if .Mtu}}, will skip this based on the go template docs
```
{{if pipeline}} T1 {{end}}
	If the value of the pipeline is empty, no output is generated;
	otherwise, T1 is executed. The empty values are false, 0, any
	nil pointer or interface value, and any array, slice, map, or
	string of length zero.
	Dot is unaffected.
```

faiq: Yes, and then it will default to the ptp behavior of being set by the kernel. I thought that's what we wanted https://github.com/kubernetes-sigs/kind/pull/1690#discussion_r445998059 

In the docs it says it is an optional field and will be set by Kernel if its not there: https://github.com/kubernetes-sigs/kind/pull/1690#discussion_r445998059

```
Network configuration reference
name (string, required): the name of the network
type (string, required): "ptp"
ipMasq (boolean, optional): set up IP Masquerade on the host for traffic originating from ip of this network and destined outside of this network. Defaults to false.
mtu (integer, optional): explicitly set MTU to the specified value. Defaults to value chosen by the kernel.
ipam (dictionary, required): IPAM configuration to be used for this network.
dns (dictionary, optional): DNS information to return as described in the Result.
```

aojea: right, perfect

aojea:
this looks good now, thanks for your patience.
Do you mind squashing all commits?

faiq:
I can squash all the commits, sure.

aojea:
/retest
/lgtm 
thanks

aojea:
tested locally and works
```
3: eth0@if5: <BROADCAST,MULTICAST,UP,LOWER_UP,M-DOWN> mtu 1400 qdisc noqueue 
```

```
root@kind-control-plane:/# crictl logs f438b5bc17747
I0702 08:52:26.361506       1 main.go:65] hostIP = 172.18.0.2
podIP = 172.18.0.2
I0702 08:52:26.361748       1 main.go:74] setting mtu 1400 for CNI 
```

BenTheElder:
Sorry I was out the end of last week (holidays at employer / USA). Looking at this again now.

BenTheElder:
rebased to HEAD and pushed a kindnetd image
/lgtm
/approve

BenTheElder:
/retest
Test pod didn't start. CI issue.

BenTheElder:
thanks for this!

PR #1694, opened 2020-06-28
Add unit test for func BoolEqual

Hellcatlk:
Signed-off-by: zouyu <zouy.fnst@cn.fujitsu.com>

### DIFF
--- a/pkg/internal/assert/assert_test.go
+++ b/pkg/internal/assert/assert_test.go
@@ -64,6 +64,26 @@ func TestExpectError(t *testing.T) {
 	})
 }
 
+func TestBoolEqual(t *testing.T) {
+	t.Parallel()
+	t.Run("not equal", func(t *testing.T) {
+		t.Parallel()
+		var f fakeT
+		BoolEqual(&f, true, false)
+		if int(f) == 0 {
+			t.Fatalf("Expected t.Errorf to be called but it was not")
+		}
+	})
+	t.Run("equal", func(t *testing.T) {
+		t.Parallel()
+		var f fakeT
+		BoolEqual(&f, true, true)
+		if int(f) != 0 {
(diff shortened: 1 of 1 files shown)


pkg/internal/assert/assert_test.go
```diff
@@ -64,6 +64,34 @@ func TestExpectError(t *testing.T) {
...
+func TestBoolEqual(t *testing.T) {
+	t.Parallel()
+	t.Run("not equal", func(t *testing.T) {
+		t.Parallel()
+		var f fakeT
+		BoolEqual(&f, true, false)
+		if int(f) == 0 {
+			t.Fatalf("Expected t.Errorf to be called but it was not")
+		}
+		BoolEqual(&f, false, true)
```
neolit123, line 76: second test case with the swapped arguments can be omitted?

pkg/internal/assert/assert_test.go
```diff
@@ -64,6 +64,34 @@ func TestExpectError(t *testing.T) {
...
+		}
+	})
+	t.Run("equal", func(t *testing.T) {
+		t.Parallel()
+		var f fakeT
+		BoolEqual(&f, true, true)
+		if int(f) != 0 {
+			t.Fatalf("Expected t.Errorf not to be called but it was")
+		}
+		BoolEqual(&f, false, false)
```
neolit123, line 88: same question here.

pkg/internal/assert/assert_test.go
Hellcatlk: Thanks, done.

pkg/internal/assert/assert_test.go
Hellcatlk: Thanks, done.

BenTheElder:
e2e tests aren't necessary here. saving CI resources.

PR #1696, opened 2020-06-29
Add unit test for pkg/build/nodeimage/internal/kube/bits.go

Hellcatlk:
Signed-off-by: zouyu <zouy.fnst@cn.fujitsu.com>


BenTheElder:
this isn't really expected to be stable API, adding these tests mostly seems to be a shortcut to inflate code coverage without testing much functionality.

I appreciate the PR, but I think I'm going to pass on this one.
this interface is going to be ripped out for #381

PR #1697, opened 2020-06-29
upgrade to go 1.15beta1

BenTheElder:
go 1.15 has some mitigations for various EINTR
this is only used for the kind binary, but we're getting ahead of k8s
k/k will possibly try to do something similar for 1.19 (adopt 1.15rc1, ship with 1.15)

kindnetd is still built with go1.13, using the multi-stage image build there, and indeed each image has it's own.


BenTheElder:
TBR
minor dep change.

PR #1703, opened 2020-07-04
Site: add scroll on sidebar

ericsyh:
### Why for change
On the current [kind docs](https://kind.sigs.k8s.io/)  cannot see the full nav list of sidebar as below : 

image

So i wanna add a scroll to the sidebar to fix this issue.

### What did i do
I added overflow style on `sidebar.html` .

### Test
Test on my local: 
image

### DIFF
--- a/site/assets/css/inline.css
+++ b/site/assets/css/inline.css
@@ -138,6 +138,8 @@ body {
   border-right: 1px solid rgba(0, 0, 0, 0.1);
   background-color: #fafafa;
   z-index: 99999;
+  overflow-y: scroll;
+  overflow-x: hidden;
 }
 .sidebar-collapsed #sidebar {
   display: none;
--- a/site/layouts/partials/sidebar.html
+++ b/site/layouts/partials/sidebar.html
@@ -1,4 +1,4 @@
-<div id="sidebar" style="overflow-y: scroll; overflow-x: hidden">
+<div id="sidebar" >
   <a href="{{.Site.BaseURL}}" id="sidebar-logo"><img src="{{ "logo/logo.png" | relURL }}" width="100%" /></a>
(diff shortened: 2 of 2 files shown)


site/layouts/partials/sidebar.html
```diff
@@ -1,4 +1,4 @@
-<div id="sidebar">
+<div id="sidebar" style="overflow: scroll">
```
BenTheElder, line 1: Let's only do this on y axis

BenTheElder: Also we should do this in the stylesheet

ericsyh: @BenTheElder 
Hi Benjamin, i updated in https://github.com/kubernetes-sigs/kind/pull/1703/commits/dfb524f2f3f6344679e27c1d701c8e717e3ada2a, PTAL again.

site/assets/css/inline.css
```diff
@@ -138,6 +138,8 @@ body {
   border-right: 1px solid rgba(0, 0, 0, 0.1);
   background-color: #fafafa;
   z-index: 99999;
+  overflow-y: scroll;
```
BenTheElder, line 141: this should be:
```suggestion
  overflow-y: auto;
```
we don't want to show a scrollbar when it's not needed

site/assets/css/inline.css
```diff
@@ -138,6 +138,8 @@ body {
...
+  overflow-x: hidden;
```
BenTheElder, line 142: I don't think we need to set this at all?

BenTheElder:
in the future squash your commits to only contain meaningful git history, it makes it easier to dig through later.
in this case the PR should be a single commit as it's one small change.

thanks for the PR!

BenTheElder:
merged in #1717 with some minor cleanup + an additional fix

thanks for the PR!

PR #1706, opened 2020-07-06
increase iptables sync period

aojea:
xref https://kubernetes.slack.com/archives/CEKK1KTN2/p1594028231477100

fixes: #1707

### DIFF
--- a/pkg/cluster/internal/kubeadm/config.go
+++ b/pkg/cluster/internal/kubeadm/config.go
@@ -295,6 +295,8 @@ mode: "{{ .KubeProxyMode }}"
 {{ range $key := .SortedFeatureGateKeys }}
   "{{ $key }}": {{$.FeatureGates $key }}
 {{end}}{{end}}
+iptables:
+  minSyncPeriod: 1s
 `
 
 // ConfigTemplateBetaV1 is the kubadm config template for API version v1beta1
@@ -410,6 +412,8 @@ mode: "{{ .KubeProxyMode }}"
 {{ range $key := .SortedFeatureGateKeys }}
   "{{ $key }}": {{ index $.FeatureGates $key }}
 {{end}}{{end}}
+iptables:
+  minSyncPeriod: 1s
 `
 
 // ConfigTemplateBetaV2 is the kubadm config template for API version v1beta2
@@ -525,6 +529,8 @@ mode: "{{ .KubeProxyMode }}"
 {{ range $key := .SortedFeatureGateKeys }}
   "{{ $key }}": {{ index $.FeatureGates $key }}
 {{end}}{{end}}
+iptables:
+  minSyncPeriod: 1s
 `
 
 // Config returns a kubeadm config generated from config data, in particular


pkg/cluster/internal/kubeadm/config.go
```diff
@@ -525,6 +525,9 @@ mode: "{{ .KubeProxyMode }}"
 {{ range $key := .SortedFeatureGateKeys }}
   "{{ $key }}": {{ index $.FeatureGates $key }}
 {{end}}{{end}}
+iptables:
+  minSyncPeriod: 10s
+  syncPeriod: 60s
```
BenTheElder, line 530: do these match GCE?
should we do it for the other versions?

danwinship: `syncPeriod` is mostly irrelevant these days and should just be left as whatever the default is

`minSyncPeriod: 10s` seems way too high. There is a huge difference between `minSyncPeriod: 0s` and `minSyncPeriod: 1s`, but anything beyond that is mostly just causing iptables updates to lag behind.

BenTheElder:
Hmm
/retest

aojea:
/retest

[sig-storage] PersistentVolumes-local [Volume type: dir-link-bindmounted] Two pods mounting a local volume at the same time should be able to write from pod1 and read from pod2 expand_less
--

aojea:
it failed with
```
wrapper.sh] [EARLY EXIT] Interrupted, entering handler ...
wrapper.sh] [CLEANUP] Cleaning up after docker in docker ...
Stopping Docker: dockerProgram process in pidfile '/var/run/docker-ssd.pid', 1 process(es), refused to die.
wrapper.sh] [CLEANUP] Done cleaning up after docker in docker.
wrapper.sh] [EARLY EXIT] Completed handler ...
```

there are still messages failing to acquire the lock in the jobs, will try with 10 sec to see the diff

aojea:
this is definitively worse, because the are still lock issues, but this ways there is no time to recover

aojea:
failed because of the lock issue
pull-kind-conformance-parallel-ipv6 — Job failed.

let's try 5 secs now, is the time used by the iptables --wait flag

aojea:
/test all
it went good this time

aojea:
/hold cancel

this will not solve the lock issues but per Dan's comment it will solve/alleviate them

I think that this will fix the lock issues, https://github.com/kubernetes/kubernetes/pull/92860
waiting for Dan to confirm

aojea:

don-t remember to see this before

> ✗ Preparing nodes 📦 📦 📦 
ERROR: failed to create cluster: docker run error: command "docker run --hostname kind-worker --name kind-worker --label io.x-k8s.kind.role=worker --privileged --security-opt seccomp=unconfined --security-opt apparmor=unconfined --tmpfs /tmp --tmpfs /run --volume /var --volume /lib/modules:/lib/modules:ro --detach --tty --label io.x-k8s.kind.cluster=kind --net kind --restart=on-failure:1 kindest/node:latest" failed with error: exit status 125
Command Output: 2fa3e8539065d01a7c1ca649b1a70e36ea4f074f9be71e400f9e50dd34c81ed1
docker: Error response from daemon: failed to retreive labels: unexpected EOF: unknown.

this failure is weird too, it ends correctly but the error says it times out

```
Ran 278 of 5208 Specs in 992.382 seconds
SUCCESS! -- 278 Passed | 0 Failed | 0 Pending | 4930 Skipped
Ginkgo ran 1 suite in 16m44.424951814s
Test Suite Passed
+ cleanup
+ [ false = true ]
+ [ true = true ]
+ kind export logs /logs/artifacts/logs
Exported logs for cluster "kind" to:
/logs/artifacts/logs
+ kind delete cluster
Deleting cluster "kind" ...
{"component":"entrypoint","file":"prow/entrypoint/run.go:164","func":"k8s.io/test-infra/prow/entrypoint.Options.ExecuteProcess","level":"error","msg":"Process did not finish before 40m0s timeout","severity":"error","time":"2020-07-07T12:32:19Z"}
{"component":"entrypoint","file":"prow/entrypoint/run.go:245","func":"k8s.io/test-infra/prow/entrypoint.gracefullyTerminate","level":"error","msg":"Process did not exit before 15s grace period","severity":"error","time":"2020-07-07T12:32:34Z"}
```

pkg/cluster/internal/kubeadm/config.go
BenTheElder: @aojea did we ever adjust the default upstream?

aojea: we did https://github.com/kubernetes-sigs/kind/pull/1706

BenTheElder: That's a link back to this very PR?
I see that https://github.com/kubernetes/kubernetes/pull/92860 landed with a cross reference here and I remember that, but that's not this.

aojea: Lol I pasted the wrong link 
https://github.com/kubernetes/kubernetes/pull/92836

BenTheElder: thanks! I thought maybe we had but couldn't find it. glad to see that we did. :-)

PR #1713, opened 2020-07-08
Bump haproxy version

tao12345666333:
.. wait for upstream docker image published.

### DIFF
--- a/images/haproxy/Dockerfile
+++ b/images/haproxy/Dockerfile
@@ -13,6 +13,6 @@
 # limitations under the License.
 
 # standard haproxy image + minimal config so the container will not exit
-ARG BASE="haproxy:2.1.7-alpine"
+ARG BASE="haproxy:2.2.0-alpine"
 FROM ${BASE}
 COPY haproxy.cfg /usr/local/etc/haproxy/haproxy.cfg
--- a/site/content/docs/user/working-offline.md
+++ b/site/content/docs/user/working-offline.md
@@ -158,8 +158,7 @@ EOF
 Note that in an offline environment, in addition to preparing the node image,
 you also need to prepare HAProxy image in advance.
 
-Currently KIND uses `kindest/haproxy:2.1.1-alpine`,
-you can find the specific tag in the [loadbalancer source code][loadbalancer source code].
+You can find the specific tag currently in use at [loadbalancer source code][loadbalancer source code].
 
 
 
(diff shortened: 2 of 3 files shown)


images/haproxy/Dockerfile
```diff
@@ -13,6 +13,6 @@
 # limitations under the License.
 
 # standard haproxy image + minimal config so the container will not exit
-ARG BASE="haproxy:2.1.7-alpine"
+ARG BASE="haproxy:2.2.0-alpine"
```
BenTheElder, line 16: any breaking changes we should know about?

site/content/docs/user/working-offline.md
```diff
@@ -158,7 +158,7 @@ EOF
 Note that in an offline environment, in addition to preparing the node image,
 you also need to prepare HAProxy image in advance.
 
-Currently KIND uses `kindest/haproxy:2.1.1-alpine`,
+Currently KIND uses `kindest/haproxy:2.2.0-alpine`,
```
BenTheElder, line 161: this is out of date and we probably need a way to fix that. we use date-sha tags for all images/

images/haproxy/Dockerfile
tao12345666333: In our case (KIND project) I did not find any break change.

The new version includes some performance optimization, dynamic error handling, enhanced observability, etc.  In some cases, we can use it for debugging HA traffic.

site/content/docs/user/working-offline.md
tao12345666333: Or can I directly state that it can be found here? https://github.com/kubernetes-sigs/kind/blob/master/pkg/cluster/internal/loadbalancer/const.go#L20

neolit123: if most use cases don't mind the underlying haproxy version, there is also the option to not mention the exact version in the docs.

tao12345666333: great! let me change it.

tao12345666333: updated, PTAL. 
thanks!

images/haproxy/Dockerfile
BenTheElder: SGTM :-)

BenTheElder:
hmm cross building the image is failing, looking into it

tao12345666333:
oh. upstream image has not been published. wait https://github.com/docker-library/official-images/pull/8318

BenTheElder:
subscribed

PR #1720, opened 2020-07-09
use new portmap plugin

aojea:
the portmap plugin is buggy on deletion, it just blindly tries to
delete everything once it gets a DEL, no matter the IP family.

Added to the problem caused because it is being executed with the
-w option without any timeout (see go-iptables), it may hold the
lock or cause contention, making all services in the cluster to
fail.

Use the new portmap version
https://github.com/containernetworking/plugins/pull/509

### DIFF
--- a/pkg/build/nodeimage/defaults.go
+++ b/pkg/build/nodeimage/defaults.go
@@ -20,7 +20,7 @@ package nodeimage
 const DefaultImage = "kindest/node:latest"
 
 // DefaultBaseImage is the default base image used
-const DefaultBaseImage = "kindest/base:v20200709-8b878ace"
+const DefaultBaseImage = "aojea/kindbase:portmap"
 
 // DefaultMode is the default kubernetes build mode for the built image
 // see pkg/build/kube.Bits


aojea:
/priority critical-urgent
/kind bug
/assign @BenTheElder 
I'll try to fix the portmap plugin but is going to take me some time, 
in the meantime we can run without it, my grep FU does not show any e2e exercising hosport

aojea:
/hold
I'll work in the portmap plugin to solve the problem, just use this in case is urgent

aojea:
just validating hypothesis

pkg/build/nodeimage/const_cni.go
```diff
@@ -20,7 +20,7 @@ package nodeimage
 The default CNI manifest and images are our own tiny kindnet
 */
 
-var defaultCNIImages = []string{"kindest/kindnetd:v20200707-3aec2c7e"}
+var defaultCNIImages = []string{"aojea/kindnetd:noportmap"}
```
BenTheElder, line 23: Image doesn't match below FYI

aojea: :)

aojea: .... waking up ...

aojea:
this didn't run is and is the more prone to show the errors

nothing in the finished jobs so far :crossed_fingers:

aojea:
promising, no trace (I'm almost crying :smile: one year chasing this and we are so close)
/test all

aojea:
ok, we got it, no deadlock errors again
/test all

aojea:
ok, after 3 rounds and 6 different jobs not locking problems, I think that this time we can confirm the portmap plugin was the one causing the iptables locking problem
/cc @danwinship 
Will work on fixing the portmap plugin

danwinship:
This will break the kube e2e tests that use kind won't it?

aojea:
> This will break the kube e2e tests that use kind won't it?

there are no hostPort e2e tests, at least I didn't find them :shrug:

aojea:
here we go, with the fix here https://github.com/containernetworking/plugins/pull/509

aojea:
pull-kind-e2e-kubernetes and pull-kind-e2e-kubernetes-1-18 clean

aojea:
@BenTheElder this is ready to go, what will be the best way to include the plugin with the fix meanwhile is not merged in CNI?

BenTheElder:
maybe we can build the CNI plugins in a multi-step base-image dockerfile from your fork? With a TODO to switch to upstream once merged, and to switch to just consuming the release binaries once released.

aojea:
/close
by https://github.com/kubernetes-sigs/kind/pull/1729

PR #1725, opened 2020-07-12
Fix Mac OS X url

jayunit100:
the `Darwin` uname value on OS X breaks kind install from the URL, so I separated it out from the linux one.  I thought about doing something fancy to sed D/d  lower case but... seems like its better to just keep it simple

### DIFF
--- a/site/content/docs/user/quick-start.md
+++ b/site/content/docs/user/quick-start.md
@@ -37,7 +37,7 @@ into your `$PATH`.
 On Linux:
 
 {{< codeFromInline lang="bash" >}}
-curl -Lo ./kind https://kind.sigs.k8s.io/dl/v0.8.1/kind-$(uname)-amd64
+curl -Lo ./kind https://kind.sigs.k8s.io/dl/v0.8.1/kind-linux-amd64
 chmod +x ./kind
 mv ./kind /some-dir-in-your-PATH/kind
 {{< /codeFromInline >}}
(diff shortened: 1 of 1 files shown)


site/content/docs/user/quick-start.md
```diff
@@ -34,18 +34,21 @@ Stable releases are generally recommended for CI usage in particular.
 To install, download the binary for your platform from "Assets" and place this
 into your `$PATH`.
 
-On macOS / Linux:
+On Linux:
 
 {{< codeFromInline lang="bash" >}}
 curl -Lo ./kind https://kind.sigs.k8s.io/dl/v0.8.1/kind-$(uname)-amd64
```
BenTheElder, line 40: might as well drop this uname then?

I think this is an unintended side-effect of the download alias, github was serving these case-insensitive.

jayunit100: Your call: IMO at some point this will work again, if we rename or alias the download aliases , so I think its good to keep the generic functionality.

BenTheElder: I don't think there's a good way to rename this, all of the naming is generic, and I think splitting it up is probably cleaner anyhow, so we can list all the options for a platform.

IIRC on some systems uname returns `Linux` as well ...?

aojea: I think that removing `uname` is more consistent with the rest of the doc now

aojea:
uff it fails here too, I was worried I broke something
@BenTheElder is there a possible regression in 1.16?

site/content/docs/user/quick-start.md
jayunit100: done

aojea:
/lgtm
Thanks Jay

BenTheElder:
Thanks so much!

PR #1728, opened 2020-07-13
providers, podman: support podman 2.0

giuseppe:
fix some issues that prevented using podman 2.0

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>

### DIFF
--- a/pkg/cluster/internal/providers/podman/provider.go
+++ b/pkg/cluster/internal/providers/podman/provider.go
@@ -182,20 +182,45 @@ func (p *Provider) GetAPIServerEndpoint(cluster string) (string, error) {
 		return "", errors.Errorf("network details should only be one line, got %d lines", len(lines))
 	}
 
-	// portMapping maps to the standard CNI portmapping capability
+	// portMapping19 maps to the standard CNI portmapping capability used in podman 1.9
 	// see: https://github.com/containernetworking/cni/blob/spec-v0.4.0/CONVENTIONS.md
-	type portMapping struct {
+	type portMapping19 struct {
 		HostPort      int32  `json:"hostPort"`
 		ContainerPort int32  `json:"containerPort"`
 		Protocol      string `json:"protocol"`
 		HostIP        string `json:"hostIP"`
 	}
+	// portMapping20 maps to the podman 2.0 portmap type
+	// see: https://github.com/containers/podman/blob/05988fc74fc25f2ad2256d6e011dfb7ad0b9a4eb/libpod/define/container_inspect.go#L134-L143
+	type portMapping20 struct {
+		HostPort string `json:"HostPort"`
+		HostIP   string `json:"HostIp"`
+	}
 
-	var portMappings []portMapping
-	if err := json.Unmarshal([]byte(lines[0]), &portMappings); err != nil {
+	portMappings20 := make(map[string][]portMapping20)
+	if err := json.Unmarshal([]byte(lines[0]), &portMappings20); err == nil {
+		for k, v := range portMappings20 {
+			protocol := "tcp"
+			parts := strings.Split(k, "/")
+			if len(parts) == 2 {
+				protocol = strings.ToLower(parts[1])
+			}
+			containerPort, err := strconv.Atoi(parts[0])
+			if err != nil {
+				return "", err
+			}
+			for _, pm := range v {
+				if containerPort == common.APIServerInternalPort && protocol == "tcp" {
+					return net.JoinHostPort(pm.HostIP, pm.HostPort), nil
+				}
+			}
+		}
+	}
+	var portMappings19 []portMapping19
+	if err := json.Unmarshal([]byte(lines[0]), &portMappings19); err != nil {
 		return "", errors.Errorf("invalid network details: %v", err)
 	}
-	for _, pm := range portMappings {
+	for _, pm := range portMappings19 {
 		if pm.ContainerPort == common.APIServerInternalPort && pm.Protocol == "tcp" {
 			return net.JoinHostPort(pm.HostIP, strconv.Itoa(int(pm.HostPort))), nil
 		}
--- a/pkg/cluster/internal/providers/podman/provision.go
+++ b/pkg/cluster/internal/providers/podman/provision.go
@@ -353,7 +353,7 @@ func generatePortMappings(clusterIPFamily config.ClusterIPFamily, portMappings .
 		if strings.HasSuffix(hostPortBinding, ":0") {
 			hostPortBinding = strings.TrimSuffix(hostPortBinding, "0")
 		}
-		args = append(args, fmt.Sprintf("--publish=%s:%d/%s", hostPortBinding, pm.ContainerPort, protocol))
+		args = append(args, fmt.Sprintf("--publish=%s:%d/%s", hostPortBinding, pm.ContainerPort, strings.ToLower(protocol)))
 	}
 	return args, nil
 }
--- a/pkg/cluster/internal/providers/podman/provision.go
+++ b/pkg/cluster/internal/providers/podman/provision.go
@@ -179,8 +179,6 @@ func runArgsForNode(node *config.Node, clusterIPFamily config.ClusterIPFamily, n
 		// including some ones podman would otherwise do by default.
 		// for now this is what we want. in the future we may revisit this.
 		"--privileged",
-		"--security-opt", "seccomp=unconfined", // also ignore seccomp
-		"--security-opt", "apparmor=unconfined", // also ignore apparmor
 		// runtime temporary storage
 		"--tmpfs", "/tmp", // various things depend on working /tmp
 		"--tmpfs", "/run", // systemd wants a writable /run


pkg/cluster/internal/providers/podman/provider.go
```diff
@@ -183,19 +183,41 @@ func (p *Provider) GetAPIServerEndpoint(cluster string) (string, error) {
...
 	// portMapping maps to the standard CNI portmapping capability
-	// see: https://github.com/containernetworking/cni/blob/spec-v0.4.0/CONVENTIONS.md
-	type portMapping struct {
+	type portMapping19 struct {
 		HostPort      int32  `json:"hostPort"`
 		ContainerPort int32  `json:"containerPort"`
 		Protocol      string `json:"protocol"`
 		HostIP        string `json:"hostIP"`
 	}
+	type portMapping20 struct {
```
aojea, line 192: can you link to some place explaining this new struct?
is podman 2.0 not using CNI?
portMapping20 seems a subset of portMapping19, is not possible to use the same struct?

pkg/cluster/internal/providers/podman/provision.go
```diff
@@ -179,8 +179,8 @@ func runArgsForNode(node *config.Node, clusterIPFamily config.ClusterIPFamily, n
 		// including some ones podman would otherwise do by default.
 		// for now this is what we want. in the future we may revisit this.
 		"--privileged",
+		"--cgroupns=private",
```
aojea, line 182: is this a breaking change?
https://github.com/containers/podman/issues/4363

aojea:
sorry, but we don't have CI for podman, is this going to work for both 1.9 and 2.0?

pkg/cluster/internal/providers/podman/provision.go
giuseppe: thanks for pointing it out. I think it is better to drop it.  On cgroup v2 it is the default configuration anyway.

pkg/cluster/internal/providers/podman/provider.go
giuseppe: it is coming from: https://github.com/containers/podman/blob/master/libpod/define/container_inspect.go#L130-L139

@mheon should we consider it a breaking change?

They also have different types (one is int32 and the other is string)

giuseppe:
> sorry, but we don't have CI for podman, is this going to work for both 1.9 and 2.0?

yes, I've tested it both with 1.9 and 2.0

pkg/cluster/internal/providers/podman/provider.go
mheon: 2.0 changed the output of inspect port mapping to match Docker, as opposed to previous versions which directly dumped the JSON form of CNI port mappings. Generally speaking, in the past, Podman has considered these bugfixes (we declare that our CLI is Docker compatible, so unintentional breakages like this are bugs).

pkg/cluster/internal/providers/podman/provider.go
```diff
@@ -183,19 +183,41 @@ func (p *Provider) GetAPIServerEndpoint(cluster string) (string, error) {
 	}
 
 	// portMapping maps to the standard CNI portmapping capability
-	// see: https://github.com/containernetworking/cni/blob/spec-v0.4.0/CONVENTIONS.md
-	type portMapping struct {
```
aojea, line 187: please add a comment to the new type
```go
 	// portMapping19 maps to the standard CNI portmapping capability used in podman 1.9
	// see: https://github.com/containernetworking/cni/blob/spec-v0.4.0/CONVENTIONS.md
```

pkg/cluster/internal/providers/podman/provider.go
aojea: please add comments to the portMapping structs explaining its sources, something along these lines
```go
 	// portMapping20 maps to the podman 2.0 portmap type
	// see:  https://github.com/containers/podman/blob/master/libpod/define/container_inspect.go#L130-L139
```

giuseppe: added

pkg/cluster/internal/providers/podman/provider.go
giuseppe: added

pkg/cluster/internal/providers/podman/provider.go
aojea: > Generally speaking, in the past, Podman has considered these bugfixes (we declare that our CLI is Docker compatible, so unintentional breakages like this are bugs).

@mheon podman will have the same policy for 2.0 onwards?

pkg/cluster/internal/providers/podman/provision.go
```diff
@@ -180,7 +180,6 @@ func runArgsForNode(node *config.Node, clusterIPFamily config.ClusterIPFamily, n
 		// for now this is what we want. in the future we may revisit this.
 		"--privileged",
 		"--security-opt", "seccomp=unconfined", // also ignore seccomp
-		"--security-opt", "apparmor=unconfined", // also ignore apparmor
```
aojea, line 183: is there any side effect removing this?

aojea: ic 
```
Command Output: Error: invalid config provided: AppArmorProfile and privileged are mutually exclusive options
```
I'm not very much into apparmor, so my question is about the user experience, does this means the users need to disable apparmor or use specific profiles, or just will work?

aojea:
@mheon talking about compatibility , if I use the same command we are using in the docker provider, podman 2.0 fails :/

https://github.com/kubernetes-sigs/kind/blob/c9469c692b05baee85d27984f23670a1c9b68ff6/pkg/cluster/internal/providers/docker/provider.go#L160-L166

pkg/cluster/internal/providers/podman/provision.go
AkihiroSuda: Let's just specify `--privileged` and drop all `--security-opt *`.

giuseppe: > Let's just specify `--privileged` and drop all `--security-opt *`.

that is better, thanks for the suggestion.  I've updated the PR

aojea:
/retest
/lgtm
hope we don't have more differences between 1.9 and 2.x
Thanks
/assign @BenTheElder

pkg/cluster/internal/providers/podman/provider.go
mheon: Yes, but I don't think we're expecting any further changes here. Inspect now matches Docker to a high degree, so we've attained our goal of compatibility.

pkg/cluster/internal/providers/podman/provision.go
BenTheElder: how does this apply to older versions of podman?

BenTheElder: were these flags just ignored? docker does not throw this error.

BenTheElder:
1.16 has a regression somewhere in the stack, I'll bypass that job when we're ready to merge.

pkg/cluster/internal/providers/podman/provider.go
```diff
@@ -182,20 +182,45 @@ func (p *Provider) GetAPIServerEndpoint(cluster string) (string, error) {
 		return "", errors.Errorf("network details should only be one line, got %d lines", len(lines))
 	}
 
-	// portMapping maps to the standard CNI portmapping capability
+	// portMapping19 maps to the standard CNI portmapping capability used in podman 1.9
```
BenTheElder, line 185: this also applies to other < 2.0, right?

pkg/cluster/internal/providers/podman/provision.go
```diff
@@ -355,7 +353,7 @@ func generatePortMappings(clusterIPFamily config.ClusterIPFamily, portMappings .
 		if strings.HasSuffix(hostPortBinding, ":0") {
 			hostPortBinding = strings.TrimSuffix(hostPortBinding, "0")
 		}
-		args = append(args, fmt.Sprintf("--publish=%s:%d/%s", hostPortBinding, pm.ContainerPort, protocol))
+		args = append(args, fmt.Sprintf("--publish=%s:%d/%s", hostPortBinding, pm.ContainerPort, strings.ToLower(protocol)))
```
BenTheElder, line 356: aside: this really seems like a bug in podman.

mheon: Already fixed on master, but it slipped into a few published releases

BenTheElder: docker accepts uppercase AND this is the canonical value in CRI, which is why we're using uppercase https://github.com/kubernetes/cri-api/blob/34366a3c19379c566a82895ce1bcb75d9a502fac/pkg/apis/runtime/v1alpha2/api.pb.go#L58

pkg/cluster/internal/providers/podman/provider.go
giuseppe: yes, this is the struct used by podman <= 1.9

pkg/cluster/internal/providers/podman/provision.go
giuseppe: I think it is just a regression in Podman that raises an error when both `--privileged` and `--security-opt apparmor=*` are used.  There is already an issue to track the problem in podman 2.0 so that `--security-opt apparmor=` overrides `--privileged`.  For our use case though it doesn't matter since `--privileged` implies no apparmor and selinux

BenTheElder: SGTM, I just want to make sure we're not not-reporting these :sweat_smile:

pkg/cluster/internal/providers/podman/provision.go
BenTheElder: ack, thanks

pkg/cluster/internal/providers/podman/provider.go
BenTheElder: maybe we should note that /shrug

BenTheElder:
We need to file a bug about 1.16
/override pull-kind-conformance-parallel-1-16

Issue #1723, opened 2020-07-11
port-forwarding broken in kindest/node:v1.18.2 but works on kindest/node:v1.18.0


nicks:
**What happened**:

port-forwarding to a Kind cluster doesn't appear to work after 1.18.2 (works OK on 1.18.0)

**How to reproduce it (as minimally and precisely as possible)**:

1. Run:

```
kind create cluster --image=kindest/node:v1.18.2
```

2. Apply this YAML

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: example-html
  labels:
    app: example-html
spec:
  selector:
    matchLabels:
      app: example-html
  template:
    metadata:
      labels:
        app: example-html
    spec:
      containers:
      - name: example-html
        image: busybox
        command: ["busybox", "httpd", "-f", "-p", "8000"]
        ports:
        - containerPort: 8000
```

3. Wait until the pod is ready

4. Run

```
kubectl port-forward deployment/example-html 8000:8000
```

5. Visit http://localhost:8000/

Expected result:
a 404 response

Actual result:
Connection hangs indefinitely

In the kubectl port-forward terminal, I see:
> E0710 19:56:07.104125 2250228 portforward.go:400] an error occurred forwarding 8000 -> 8000: error forwarding port 8000 to pod 7e6be7fb9a3b95c5c80f206f931b339d955f4fd9423a393f7ee072c477d9f370, uid : failed to execute portforward in network namespace "/var/run/netns/cni-1b6beba4-efb4-40fc-741e-fcce3506a3d3": socat command returns error: exit status 1, stderr: "2020/07/10 23:56:07 socat[1814] E connect(5, AF=2 92.242.140.21:8000, 16): Connection timed out\n"

**Anything else we need to know?**:

If I start Kind with `kind create cluster --image=kindest/node:v1.18.0`, port-forwarding works fine

I am so curious what the problem is here, and would love tips on how to debug this further. I've been banging my head against the wall for awhile

**Environment:**

- kind version: (use `kind version`): kind v0.8.1 go1.14.3 linux/amd64
- Kubernetes version: (use `kubectl version`):

```
[...]

PR #1733, opened 2020-07-14
use files first to resolve dns on node images

aojea:
change back nsswitch.conf to use files first to resolve dns

Fixes: https://github.com/kubernetes-sigs/kind/issues/1723


aojea:
/hold

let's see first how is working the new image

BenTheElder:
do you remember why we actually did this before? seems like we must have had a reason 😱

BenTheElder:
I'm usually running around doing this and not the opposite like before

BenTheElder:
Intend to double check thoroughly and ship later this weekend.

BenTheElder:
/retest
sheesh pod pending timeouts w/ prow ...

BenTheElder:
flake
/retest

PR #1737, opened 2020-07-19
minor fixup

BenTheElder:
we don't actually need all of /var to be a mount, just some things we know we write to a lot / are doing nested containers


BenTheElder:
surprise surprise pod-pending timeout

BenTheElder:
CI is in bad shape https://github.com/kubernetes/kubernetes/issues/92937, there's no issues related to this in the runs that actually start though.

BenTheElder:
flakes

PR #1739, opened 2020-07-20
Fix doc for Fedora 32 workaround

rcarrillocruz:


site/content/docs/user/known-issues.md
```diff
@@ -293,7 +293,7 @@ You can work around this by changing the `FirewallBackend` in the `/etc/firewall
 
 ```console
 sed -i /etc/firewalld/firewalld.conf 's/FirewallBackend=.*/FirewallBackend=iptables/'
-systemctl restart docker
+systemctl restart firewalld
```
aojea, line 296: Line above says `restarting docker` , seems the user submitted it found that restarting docker solved the problem.
Is ok restarting only firewalld?

aojea: please change line 292 too :)

BenTheElder:
Thanks for the PR!

Unfortunately Kubernetes projects do require the CNCF CLA, we'll need that before we can merge this, in addition to review comments. https://github.com/kubernetes-sigs/kind/pull/1739#issuecomment-660892590

site/content/docs/user/known-issues.md
rcarrillocruz: > Line above says `restarting docker` , seems the user submitted it found that restarting docker solved the problem.
> Is ok restarting only firewalld?

yep, I just needed to restart firewalld

BenTheElder:
/approve
that 1.16 thing still needs a bug, sigh.

BenTheElder:
filed https://github.com/kubernetes/kubernetes/issues/93301

PR #1747, opened 2020-07-23
Support configuring a cgroup root for kubelet

### LINKED ISSUE #1614: kubelet commits suicide when running KIND in a DinD pod on a k8s cluster
**What happened**:

Kubelet gets confused about some unexpected cgroups and SIGKILLs everything including itself.

KIND cluster creation to work... :unicorn: and :carousel_horse: 

https://github.com/porridge/kind-dind-kind-repro contains the minimum instructions I could come up with.

I [reported this on slack](https://kubernetes.slack.com/archives/CEKK1KTN2/p1587988851297500?thread_ts=1575376979.206200&cid=CEKK1KTN2) earlier but only managed to come up with a fairly simple repro case now.

This was also described in [a blog post by Jie](https://d2iq.com/blog/running-kind-inside-a-kubernetes-cluster-for-continuous-integration) - look for

> However, when we try to run this in CI (in the production Kubernetes cluster), things start to fail.

That post describes a [workaround](https://github.com/jieyu/docker-images/tree/master/kind-cluster) that injects [some additional code](https://github.com/jieyu/docker-images/blob/master/kind-cluster/node/entrypoint-wrapper.sh) *around* the `kind` entrypoint to further massage the cgroup filesystem and requires [patching the kubelet config](https://github.com/jieyu/docker-images/blob/master/kind-cluster/kind-config.yaml).

Because of the maintenance burden of these rather invasive changes I'm looking for a better alternative.

**Environment:**

- kind version: (use `kind version`): kind v0.8.1 go1.14.2 linux/amd64
```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-16T11:56:40Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.8", GitCommit:"ec6eb119b81be488b030e849b9e64fda4caaf33c", GitTreeState:"clean", BuildDate:"2020-03-12T20:52:22Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
```
```
Client:
 Debug Mode: false

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 0
 Server Version: 19.03.9
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Plugins:
  Volume: local
 Swarm: inactive
 Runtimes: runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 7ad184331fa3e55e52b890ea95e65ba581ae3429
 runc version: dc9208a3303feef5b3839f4323d9beb36df0a9dd
 init version: fec3683
 Security Options:
  apparmor
  seccomp
   Profile: default
 Kernel Version: 5.5.0-0.bpo.2-amd64
 Operating System: Debian GNU/Linux 10 (buster) (containerized)
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 31.15GiB
 Name: 5bad58a9e2fb
 ID: HUJB:QY7M:643T:7OXM:L3I2:VU5C:5DHW:EEO7:53Y4:TBA4:DIXQ:A2RE
 Debug Mode: false
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
```

```
PRETTY_NAME="Debian GNU/Linux 10 (buster)"
NAME="Debian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
ID=debian
```

jbarrick-mesosphere:
Adds a cgroupRoot option to the kind configuration file that supports
specifying a cgroup root for kubelet to use.

Part of #1614

Also see https://d2iq.com/blog/running-kind-inside-a-kubernetes-cluster-for-continuous-integration

### DIFF
--- a/images/base/files/etc/systemd/system/kubelet.service.d/10-kubeadm.conf
+++ b/images/base/files/etc/systemd/system/kubelet.service.d/10-kubeadm.conf
@@ -9,4 +9,4 @@ EnvironmentFile=-/var/lib/kubelet/kubeadm-flags.env
 # the .NodeRegistration.KubeletExtraArgs object in the configuration files instead. KUBELET_EXTRA_ARGS should be sourced from this file.
 EnvironmentFile=-/etc/default/kubelet
 ExecStart=
-ExecStart=/usr/bin/kubelet $KUBELET_KUBECONFIG_ARGS $KUBELET_CONFIG_ARGS $KUBELET_KUBEADM_ARGS $KUBELET_EXTRA_ARGS
+ExecStart=/usr/bin/kubelet $KUBELET_KUBECONFIG_ARGS $KUBELET_CONFIG_ARGS $KUBELET_KUBEADM_ARGS $KUBELET_EXTRA_ARGS --cgroup-root=/kubelet
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -66,7 +66,7 @@ mount_kubelet_cgroup_root() {
   cgroup_root=$1
   subsystem=$2
 
-  if [ ! -z ${cgroup_root} ]; then
+  if [ -n "${cgroup_root}" ]; then
     # This is because we set Kubelet's cgroup-root to `/kubelet` by
     # default. We have to do that because otherwise, it'll collide
     # with the cgroups used by the Kubelet running on the host if we
@@ -109,6 +109,8 @@ fix_cgroup() {
     while IFS= read -r subsystem; do
       mkdir -p "${subsystem}${docker_cgroup}"
       mount --bind "${subsystem}" "${subsystem}${docker_cgroup}"
+      # shellcheck disable=SC2153
+      # CGROUP_ROOT is set in environment
       mount_kubelet_cgroup_root "${CGROUP_ROOT}" "${subsystem}"
     done
   fi
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -109,9 +109,7 @@ fix_cgroup() {
     while IFS= read -r subsystem; do
       mkdir -p "${subsystem}${docker_cgroup}"
       mount --bind "${subsystem}" "${subsystem}${docker_cgroup}"
-      # shellcheck disable=SC2153
-      # CGROUP_ROOT is set in environment
-      mount_kubelet_cgroup_root "${CGROUP_ROOT}" "${subsystem}"
+      mount_kubelet_cgroup_root "/kubelet" "${subsystem}"
     done
   fi
   local podman_cgroup_mounts
@@ -124,7 +122,7 @@ fix_cgroup() {
     while IFS= read -r subsystem; do
       mkdir -p "${subsystem}${podman_cgroup}"
       mount --bind "${subsystem}" "${subsystem}${podman_cgroup}"
-      mount_kubelet_cgroup_root "${CGROUP_ROOT}" "${subsystem}"
+      mount_kubelet_cgroup_root "/kubelet" "${subsystem}"
     done
   fi
 }
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -62,6 +62,30 @@ fix_mount() {
   mount --make-rshared /
 }
 
+mount_kubelet_cgroup_root() {
+  cgroup_root=$1
+  subsystem=$2
+
+  if [ ! -z ${cgroup_root} ]; then
+    # This is because we set Kubelet's cgroup-root to `/kubelet` by
+    # default. We have to do that because otherwise, it'll collide
+    # with the cgroups used by the Kubelet running on the host if we
+    # run kind cluster within a Kubernetes pod, resulting in random
+    # processes to be killed.
+    mkdir -p "${subsystem}/${cgroup_root}"
+    if [ "${subsystem}" == "/sys/fs/cgroup/cpuset" ]; then
+      # This is needed. Otherwise, assigning process to the cgroup
+      # (or any nested cgroup) would result in ENOSPC.
+      cat "${subsystem}/cpuset.cpus" > "${subsystem}/${cgroup_root}/cpuset.cpus"
+      cat "${subsystem}/cpuset.mems" > "${subsystem}/${cgroup_root}/cpuset.mems"
+    fi
+    # We need to perform a self bind mount here because otherwise,
+    # systemd might delete the cgroup unintentionally before the
+    # kubelet starts.
+    mount --bind "${subsystem}/${cgroup_root}" "${subsystem}/${cgroup_root}"
+  fi
+}
+
 fix_cgroup() {
   echo 'INFO: fix cgroup mounts for all subsystems'
   # For each cgroup subsystem, Docker does a bind mount from the current
@@ -85,10 +109,11 @@ fix_cgroup() {
     while IFS= read -r subsystem; do
       mkdir -p "${subsystem}${docker_cgroup}"
       mount --bind "${subsystem}" "${subsystem}${docker_cgroup}"
+      mount_kubelet_cgroup_root "${CGROUP_ROOT}" "${subsystem}"
     done
   fi
   local podman_cgroup_mounts
-  podman_cgroup_mounts=$(grep /sys/fs/cgroup /proc/self/mountinfo | grep libpod_parent || true)
+  podman_cgroup_mounts=$(grep /sys/fs/cgroup /proc/self/mountinfo | grep libpod || true)
   if [[ -n "${podman_cgroup_mounts}" ]]; then
     local podman_cgroup cgroup_subsystems subsystem
     podman_cgroup=$(echo "${podman_cgroup_mounts}" | head -n 1 | cut -d' ' -f 4)
@@ -97,6 +122,7 @@ fix_cgroup() {
     while IFS= read -r subsystem; do
       mkdir -p "${subsystem}${podman_cgroup}"
       mount --bind "${subsystem}" "${subsystem}${podman_cgroup}"
+      mount_kubelet_cgroup_root "${CGROUP_ROOT}" "${subsystem}"
     done
   fi
 }
--- a/pkg/cluster/internal/kubeadm/config.go
+++ b/pkg/cluster/internal/kubeadm/config.go
@@ -70,6 +70,9 @@ type ConfigData struct {
 	// IPv4 values take precedence over IPv6 by default, if true set IPv6 default values
 	IPv6 bool
 
+	// CgroupRoot specifies the cgroup root to configure the kubelet to use
+	CgroupRoot   string
+
 	// DerivedConfigData is populated by Derive()
 	// These auto-generated fields are available to Config templates,
 	// but not meant to be set by hand
@@ -195,6 +198,7 @@ nodeRegistration:
     fail-swap-on: "false"
     node-ip: "{{ .NodeAddress }}"
     provider-id: "kind://{{.NodeProvider}}/{{.ClusterName}}/{{.NodeName}}"
+    cgroup-root: "{{ .CgroupRoot }}"
 ---
 # no-op entry that exists solely so it can be patched
 apiVersion: kubeadm.k8s.io/v1beta1
@@ -213,6 +217,7 @@ nodeRegistration:
     fail-swap-on: "false"
     node-ip: "{{ .NodeAddress }}"
     provider-id: "kind://{{.NodeProvider}}/{{.ClusterName}}/{{.NodeName}}"
+    cgroup-root: "{{ .CgroupRoot }}"
 discovery:
   bootstrapToken:
     apiServerEndpoint: "{{ .ControlPlaneEndpoint }}"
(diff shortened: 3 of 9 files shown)


BenTheElder:
this is not a complete fix for https://github.com/kubernetes-sigs/kind/issues/1614

BenTheElder:
thanks though! updated the PR to avoid closing that issue.

pkg/apis/config/v1alpha4/types.go
```diff
@@ -76,6 +76,9 @@ type Cluster struct {
 	// in the order listed.
 	// These should be YAML or JSON formatting RFC 6902 JSON patches
 	ContainerdConfigPatchesJSON6902 []string `yaml:"containerdConfigPatchesJSON6902,omitempty"`
+
+	// CgroupRoot specifies the cgroup root to configure the kubelet to use
+	CgroupRoot string `yaml:"cgroupRoot,omitempty"`
```
BenTheElder, line 81: Is there a reason for this to be configurable, instead of us just picking one?

aojea:
naive question, is there any intersection with cgroupsv2 or rootless that can impact them?

pkg/apis/config/v1alpha4/types.go
jbarrick-mesosphere: Yeah, I guess not, we can just pick one. I'll update it!

images/base/files/usr/local/bin/entrypoint
```diff
@@ -232,6 +232,39 @@ enable_network_magic(){
...
+
+  echo "INFO: configuring cgroup root ${CGROUP_ROOT}"
+
+  CURRENT_CGROUP=$(grep systemd /proc/self/cgroup | cut -d: -f3)
+  CGROUP_SUBSYSTEMS=$(findmnt -lun -o source,target -t cgroup | grep "${CURRENT_CGROUP}" | awk '{print $2}')
+
+  mount --make-rprivate /sys/fs/cgroup
+
+  echo "${CGROUP_SUBSYSTEMS}" |
+  while IFS= read -r SUBSYSTEM; do
```
BenTheElder, line 248: this should get collapsed with the other cgroup stuff eventually, we already do a nearly identical loop.

BenTheElder: somewhere I have a branch with this mostly written, but I understand if this functionality is pressing.

images/base/files/usr/local/bin/entrypoint
```diff
@@ -232,6 +232,39 @@ enable_network_magic(){
...
 }
 
+configure_cgroup_root(){
+  if [ "${CGROUP_ROOT}" == "" ]; then
+    return
+  fi
+
+  echo "INFO: configuring cgroup root ${CGROUP_ROOT}"
+
+  CURRENT_CGROUP=$(grep systemd /proc/self/cgroup | cut -d: -f3)
```
BenTheElder, line 242: IIRC this mechanism was not very robust. I'm a bit concerned about this.

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -158,6 +158,7 @@ nodeRegistration:
   kubeletExtraArgs:
     fail-swap-on: "false"
     node-ip: "{{ .NodeAddress }}"
+    cgroup-root: "/kubelet"
```
BenTheElder, line 161: setting this hardcoded breaks using older images right?

BenTheElder: that is something we can do, but avoiding it is preferable.
I think maybe information flows the wrong direction here.

images/base/files/usr/local/bin/entrypoint
jbarrick-mesosphere: What did you do instead before?

pkg/cluster/internal/kubeadm/config.go
jbarrick-mesosphere: Sure, what did you have in mind? Do we want to try to determine the current cgroup parent from the docker daemon's configuration?

images/base/files/usr/local/bin/entrypoint
BenTheElder: I think this was the bit I was trying to sort out last and didn't get it solved yet, but doing some experiments suggested it might be problematic, I'm having difficulty recalling exactly.

pkg/cluster/internal/kubeadm/config.go
BenTheElder: I'm thinking maybe for changes like this that we can gracefully upgrade behavior on we could autodetect it.
Let's not block this PR on this though, I think that's a more general problem worth solving more generally.

jbarrick-mesosphere: Sounds good, I'll do some manual testing and make sure this works (though perhaps your test suite will already catch if this breaks older images) and, if not, figure out some way to toggle it.

images/base/files/usr/local/bin/entrypoint
jbarrick-mesosphere: Alright, I'll see what I can figure out!

jbarrick-mesosphere: Actually looks like we only do this for detecting the CGROUP_SUBSYSTEMS.. so this gets solved by rolling this stuff into the loop in the other method.

pkg/cluster/internal/kubeadm/config.go
jbarrick-mesosphere: Yeah, looks like it breaks older images. Do you have an idea how you want this feature gated already?

pkg/cluster/internal/providers/docker/provision.go
```diff
@@ -199,6 +199,8 @@ func commonArgs(cluster string, cfg *config.Cluster, networkName string, nodeNam
 		args = append(args, "-e", fmt.Sprintf("%s=%s", key, val))
 	}
 
+	args = append(args, "-e", "CGROUP_ROOT=/kubelet")
```
aojea, line 202: podman does not need it?

jbarrick-mesosphere: True, I probably do need to ensure podman support is still working.

jbarrick-mesosphere: I'm trying to get this working with podman, but I can't get the podman provider working at all on my machine, unfortunately: https://github.com/kubernetes-sigs/kind/issues/1757 any ideas how I can get past that error?

aojea: Then just add a comment in the podman file with a TODO explaining what is needed, don't worry about it. 
Thanks for the effort

jbarrick-mesosphere: Okay cool, I have made it internally configurable so that a cgroup root is not passed through to podman and added a TODO comment.

pkg/cluster/internal/kubeadm/config.go
jbarrick-mesosphere: Okay, it's not internally configurable (not exposed to the user). It gets toggled on here: https://github.com/kubernetes-sigs/kind/pull/1747/files#diff-f9926f17f93e2c4abbb007effdbe351eR56

Any ideas how we can toggle this off for legacy images?

jbarrick-mesosphere:
@BenTheElder is there anything I can do to help move this along?

jbarrick-mesosphere:
@BenTheElder @aojea anything we can do to move this along?

pkg/cluster/internal/providers/docker/provider.go
```diff
@@ -53,6 +53,8 @@ type Provider struct {
 
 // Provision is part of the providers.Provider interface
 func (p *Provider) Provision(status *cli.Status, cfg *config.Cluster) (err error) {
+	cfg.CgroupRoot = "/kubelet"
```
aojea, line 56: if is the default, I think it should go here `pkg/internal/apis/config/default.go`

jbarrick-mesosphere: It should only be set for the Docker provider, so I don't think I can set it in there unless there's a good way to check if the Docker provider is in-use.

jbarrick-mesosphere: @aojea any thoughts? I'd like to keep this moving.

aojea: I'm in deep waters here with the cgroups stuff :smile: ,
 I was reviewing to try to keep it as compatible as possible between the different provider, for the final review you need @BenTheElder

BenTheElder:
Sorry I don't have anything more useful but I'm still keeping an eye on this PR, I'm just not able to spend a lot of time on it right now (kubecon, performance review season, kubernetes CI work, intern final evaluations etc...), I expect to have my attention significantly less divided again in a 1-2 weeks 🤞

pkg/cluster/internal/providers/docker/provider.go
BenTheElder: > It should only be set for the Docker provider, so I don't think I can set it in there unless there's a good way to check if the Docker provider is in-use.

Why wouldn't we set this unconditionally? 
We want things to be very reproducible between CI and local, behavior branching is kept to a minimum.

jbarrick-mesosphere: We do unconditionally set it for Docker, but I previously wasn't able to get it working with podman so we disabled it there, per https://github.com/kubernetes-sigs/kind/pull/1747#discussion_r462606918. But, it's been a few weeks, let me try a rebase and have another go at podman.

jbarrick-mesosphere: Alright, rebased and got podman working with the cgroup root. So we set "/kubelet" in `default.go` so that it can be propagated to the kubeadm config and the container that gets created. This allows us to avoid hardcoding "/kubelet" in more than one place (the kubeadm configs and the entrypoint script) as well as disable the cgroup root in the kubeadm configuration for legacy images if we have a way to detect them (https://github.com/kubernetes-sigs/kind/pull/1747#discussion_r463273752).

Can you give this another review?

jbarrick-mesosphere: @BenTheElder is it possible to get some feedback here?

jbarrick-mesosphere: @BenTheElder  what is the status of this PR?

pkg/cluster/internal/kubeadm/config.go
jbarrick-mesosphere: I moved this configuration entirely into the image by adding `--cgroup-root=/kubelet` to the kubelet flags in `images/base/files/etc/systemd/system/kubelet.service.d/10-kubeadm.conf`, so this is backwards compatible since it we're not setting the `cgroup-root` via the generated Kubeadm YAMLs and it is entirely contained in the image.

Any thoughts?

BenTheElder:
Sorry -- this is going into the next release in some form (v0.10.0), if need be I will pick up any further changes.
I apologize for the terrible experience here, this keeps falling off my radar intermittently 😞

images/base/files/etc/systemd/system/kubelet.service.d/10-kubeadm.conf
```diff
@@ -9,4 +9,4 @@ EnvironmentFile=-/var/lib/kubelet/kubeadm-flags.env
 # the .NodeRegistration.KubeletExtraArgs object in the configuration files instead. KUBELET_EXTRA_ARGS should be sourced from this file.
 EnvironmentFile=-/etc/default/kubelet
 ExecStart=
-ExecStart=/usr/bin/kubelet $KUBELET_KUBECONFIG_ARGS $KUBELET_CONFIG_ARGS $KUBELET_KUBEADM_ARGS $KUBELET_EXTRA_ARGS
\ No newline at end of file
+ExecStart=/usr/bin/kubelet $KUBELET_KUBECONFIG_ARGS $KUBELET_CONFIG_ARGS $KUBELET_KUBEADM_ARGS $KUBELET_EXTRA_ARGS --cgroup-root=/kubelet
```
BenTheElder, line 12: for a follow-up: we should inject this with kubelet config, flags are basically deprecated.

BenTheElder: there are tricker per-node kubelet flags injected right now (providerID) that will require a larger overhaul (kubeadm doesn't support per-node kubelet config, only flags...), so we can punt this until then.

and when we do that, the flag here should override the config, which would make sense given that the image uses a fixed location for now.

so this should work quite well.

BenTheElder:
this looks good. thank you. working on getting this running in CI.

we still need to improve prow.k8s.io's hacky docker in docker... it lacks some of the fixes in https://d2iq.com/blog/running-kind-inside-a-kubernetes-cluster-for-continuous-integration

kind does have the MTU match at the node / kindnetd level now at least.

BenTheElder:
(pushed rebase to latest + base image bump, this will run in CI now)

BenTheElder:
not sure what happened there, CI seems to be having issues today, I haven't quite kept up with that with the dockershim fuss 🙃

BenTheElder:
(we got podman fixed up, rebased again so we can verify this on podman)

PR #1748, opened 2020-07-24
switch to go 1.15rc1

BenTheElder:


BenTheElder:
really tempted to switch us to something like https://github.com/syndbg/goenv

@justaugustus did you wind up looking into that for k/k?

BenTheElder:
our friend` [sig-cli] Kubectl client Simple pod should support inline execution and attach `
/retest

BenTheElder:
TBR
going to start soaking this alongside k/k

aojea:
wow, is not risky using rcs?

BenTheElder:
no, rc1 is used by all of google internally for example.
kubernetes is working on making the same switch

dims:
See notes [here](https://github.com/golang/go/wiki/Go-Release-Cycle) @aojea : 

```
One of the criteria for issuing a release candidate is that Google be using that version of the code for new production builds by default: if we at Google are not willing to run it for production use, we shouldn't be asking others to. We may issue the release candidate a few days in advance of Google changing over, depending on how the calendar falls. For example, the change inside Google makes more sense to do on Mondays, so we may issue the release candidate the Wednesday before or the Wednesday after Google converts to the new release by default.
```

aojea:
There is not better test that running things in production :smile
Great

Issue #1625, opened 2020-05-27
increase kubernetes component verbosity


BenTheElder:
**What would you like to be added**:

during e2e testing we should increase the component log verbosity to match the GCE testing

**Why is this needed**:

so components can be more easily debugged

it might? not make sense to do this in clusters used for other purposes (e.g. istio development) but it definitely makes sense here.

requested by @wojtek-t
https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/config-test.sh#L204

/assign

PR #1750, opened 2020-07-25
allow CI to set the Cluster loglevel

aojea:
GCE jobs have log level 4 
https://github.com/kubernetes/kubernetes/blob/17a2e5e990d88cc3f5700d3bfab302b5a2523aba/cluster/gce/config-test.sh#L211-L218

this has performance implications, let's see how much affect

Fixes: https://github.com/kubernetes-sigs/kind/issues/1625

### DIFF
--- a/hack/ci/e2e-k8s.sh
+++ b/hack/ci/e2e-k8s.sh
@@ -129,6 +129,9 @@ create_cluster() {
     ;;
   esac
 
+# Default Log level for all components in test clusters
+KIND_CLUSTER_LOG_LEVEL=${KIND_CLUSTER_LOG_LEVEL:-4}
+
   # create the config file
   cat <<EOF > "${ARTIFACTS}/kind-config.yaml"
 # config for 1 control plane node and 2 workers (necessary for conformance)
@@ -149,6 +152,23 @@ kubeadmConfigPatches:
   apiServer:
     extraArgs:
       "runtime-config": "${runtime_config}"
+      "v": "${KIND_CLUSTER_LOG_LEVEL}"
+  controllerManager:
+    extraArgs:
+      "v": "${KIND_CLUSTER_LOG_LEVEL}"
+  scheduler:
+    extraArgs:
+      "v": "${KIND_CLUSTER_LOG_LEVEL}"
+  ---
+  kind: InitConfiguration
+  nodeRegistration:
+    kubeletExtraArgs:
+      "v": "${KIND_CLUSTER_LOG_LEVEL}"
+  ---
+  kind: JoinConfiguration
+  nodeRegistration:
+    kubeletExtraArgs:
+      "v": "${KIND_CLUSTER_LOG_LEVEL}"
 EOF
   # NOTE: must match the number of workers above
   NUM_NODES=2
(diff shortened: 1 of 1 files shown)


aojea:
/retest
sig-cli failures

hack/ci/e2e-k8s.sh
```diff
@@ -161,6 +181,10 @@ EOF
     --wait=1m \
     -v=3 \
     "--config=${ARTIFACTS}/kind-config.yaml"
+
+  # Patch kube-proxy to set the verbosity level
```
BenTheElder, line 185: eww, we can't do this in kubeadm?
kube-proxy component config?

aojea: I didn't see how, the `command` field is hardcoded
https://github.com/kubernetes/kubernetes/blob/9a20f30745477423e8776bb0d05316c00648f49b/cmd/kubeadm/app/phases/addons/proxy/manifests.go#L78-L81

maybe @neolit123 knows a better way

hack/ci/e2e-k8s.sh
```diff
@@ -161,6 +181,10 @@ EOF
...
+  kubectl patch -n kube-system daemonset/kube-proxy \
+    --type='json' -p='[{"op": "add", "path": "/spec/template/spec/containers/0/command/3", "value": "--v='${KIND_CLUSTER_LOG_LEVEL}'" }]'
```
neolit123, line 187: `/spec/template/spec/containers/0/command/3`
->
`/spec/template/spec/containers/0/command/-`
to add at the end

hack/ci/e2e-k8s.sh
neolit123: the kube-proxy config does not expose klog verbosity. updating klog verbosity *after* a config file is parsed, means one has to do, something like: `flag.Set("v", someValue)`

i don't think any component (that uses componentConfig) supports that yet.

BenTheElder: And kubeadm provides no way to customize kube-proxy args?

neolit123: no, there are no plans to support that.
kubeadm might be deploying a kube-proxy operator in the future:
https://github.com/kubernetes-sigs/cluster-addons/tree/master/kubeproxy

BenTheElder: To clarify: I don't object to the mechanism itself, except that this should really be an initial value, and not involve a restart of all of kube-proxy. _That_ is non-ideal.

I don't think an operator will help this, and if anything will increase the core overhead, I'd love to know more about what motivates this operator.

neolit123:
It should be set initially. Using the kubeadm component config is a no-go,
so the alternative is user patches. This is good idea and we can do it
today, but long term we dont want kubeadm managing kube-proxy component
config upgrades and also such extra container flags overrides could be done
in a custom resource.

The operator is one way of delegating that addon management responsibility
away, but it comes with its own complexities and the configuration flow is
not exactly clear yet.

BenTheElder:
let's try it for now, but I think longterm this will not be the right approach (bash or otherwise).
unblocking, but thinking about alternatives
/lgtm
/approve

hack/ci/e2e-k8s.sh
```diff
@@ -129,6 +129,9 @@ create_cluster() {
     ;;
   esac
 
+# Default Log level for all components in test clusters
+KIND_CLUSTER_LOG_LEVEL=${KIND_CLUSTER_LOG_LEVEL:-4}
```
BenTheElder, line 133: TODO: consider switching this to something that matches the rest of CI, if we have it. (kube-up?)

aojea:
```
hack/local-up-cluster.sh:LOG_LEVEL=${LOG_LEVEL:-3}
```
 or
```
cluster/gce/config-test.sh:TEST_CLUSTER_LOG_LEVEL=${TEST_CLUSTER_LOG_LEVEL:---v=4}
```

BenTheElder:
(we have a proposal out about CI issues, not going to block debugging things on waiting for that to be fixed)

PR #1752, opened 2020-07-27
move kubelet systemd to base image

BenTheElder:
another small piece broken out of node image build rework


munnerz:
Trying to pick up a few more reviews again.. good to see more of the build move to Dockerfiles rather than Go 😄

/lgtm

PR #1753, opened 2020-07-27
Bring back shellcheck

BenTheElder:
... I think I unintentionally removed this before.

Also fix failures.

### DIFF
--- a/hack/verify/shellcheck.sh
+++ b/hack/verify/shellcheck.sh
@@ -28,12 +28,13 @@ SHELLCHECK_IMAGE='koalaman/shellcheck:v0.7.1'
 
 # Find all shell scripts excluding:
 # - Anything git-ignored - No need to lint untracked files.
-# - ./_* - No need to lint output directories.
 # - ./.git/* - Ignore anything in the git object store.
+# - ./vendor/* - Ignore vendored contents.
+# - ./bin/* - No need to lint output directories.
 all_shell_scripts=()
 while IFS=$'\n' read -r script;
   do git check-ignore -q "$script" || all_shell_scripts+=("$script");
-done < <(grep -irl '#!.*sh' . --exclude-dir={'./bin/*','./.git/*'})
+done < <(grep -irl '#!.*sh' . | grep -Ev '^(\./\.git/)|(\./vendor/)|(\./bin/)')
 
 # common arguments we'll pass to shellcheck
 SHELLCHECK_OPTIONS=(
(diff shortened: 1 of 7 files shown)


hack/verify/shellcheck.sh
```diff
@@ -0,0 +1,58 @@
...
+  # 2126 - disabled because grep -c exits error when there are zero matches,
+  # unlike grep | wc -l
+  '--exclude=2126'
+  # set colorized output
+  '--color=auto'
+)
+
+# actually shellcheck
+docker run \
+  --rm -it -v "${REPO_ROOT}:${REPO_ROOT}" -w "${REPO_ROOT}" \
```
BenTheElder, line 56: Probably best to just drop the -it for now, but later on if we can do TTY detection here we can get nicer results locally ... This fails in CI currently.

images/base/files/usr/local/bin/entrypoint
```diff
@@ -193,7 +193,7 @@ enable_network_magic(){
   sed -e "s/${docker_embedded_dns_ip}/${docker_host_ip}/g" /etc/resolv.conf.original >/etc/resolv.conf
 
   # fixup IPs in manifests ...
-  curr_ipv4="$( (getent ahostsv4 $(hostname) | head -n1 | cut -d' ' -f1) || true)"
+  curr_ipv4="$( (getent ahostsv4 "$(hostname)" | head -n1 | cut -d' ' -f1) || true)"
```
aojea, line 196: is not a problem the nested double quotes?

aojea: it works on my pc, it should be fine then :)

hack/verify/shellcheck.sh
aojea: :roll_eyes: 

> verifying shellcheck ...
> the input device is not a TTY
>

images/base/files/usr/local/bin/entrypoint
BenTheElder: in shell double quotes nest the same as parens: balanced quotes close. this is because double quotes are a context in which things expand.

single quotes do not.

BenTheElder: not having these double quotes before means we word-split on hostname's output.
now hostnames should not have spaces, but it's still functionally incorrect bash to not have the quotes here.

BenTheElder:
Huh, verify is not ignoring ./.git in CI despite the grep. hmm

BenTheElder:
OK verify works in CI again. Some difference in mac / linux glob matching, moved away from using globs to exclude. Verified against different machines.

BenTheElder:
skipping unrelated CI, which will free up starting the retest for other PRs.

PR #1756, opened 2020-07-28
Fix lint to run recursively

amwat:
Turns out lint wasn't running recursively.
Mostly gosec issues were uncovered.

/cc @BenTheElder

### DIFF
--- a/hack/ci/e2e-k8s.sh
+++ b/hack/ci/e2e-k8s.sh
@@ -184,7 +184,7 @@ EOF
 
   # Patch kube-proxy to set the verbosity level
   kubectl patch -n kube-system daemonset/kube-proxy \
-    --type='json' -p='[{"op": "add", "path": "/spec/template/spec/containers/0/command/-", "value": "--v='${KIND_CLUSTER_LOG_LEVEL}'" }]'
+    --type='json' -p='[{"op": "add", "path": "/spec/template/spec/containers/0/command/-", "value": "--v='"${KIND_CLUSTER_LOG_LEVEL}"'" }]'
 }
 
 # run e2es with ginkgo-e2e.sh
--- a/hack/verify/lint.sh
+++ b/hack/verify/lint.sh
@@ -30,10 +30,10 @@ LINTS=(
   deadcode errcheck gosimple govet ineffassign staticcheck \
   structcheck typecheck unused varcheck \
   # additional lints
-  golint gofmt misspell gochecknoinits unparam scopelint gosec
+  golint gofmt misspell gochecknoinits unparam scopelint
 )
 LINTS_JOINED="$(IFS=','; echo "${LINTS[*]}")"
 
 # first for the repo in general
 SOURCE_DIR="${REPO_ROOT}" hack/go_container.sh \
-  /out/golangci-lint --disable-all --enable="${LINTS_JOINED}" --timeout=2m run .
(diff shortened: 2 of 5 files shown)


pkg/cluster/internal/providers/docker/network.go
```diff
@@ -17,7 +17,7 @@ limitations under the License.
 package docker
 
 import (
-	"crypto/sha1"
```
BenTheElder, line 20: This is not actually a cyptographic hash and sha256 is overkill.

pkg/internal/cli/override.go
```diff
@@ -12,7 +12,7 @@ import (
 func OverrideDefaultName(fs *pflag.FlagSet) {
 	if !fs.Changed("name") {
 		if name := os.Getenv("KIND_CLUSTER_NAME"); name != "" {
-			fs.Set("name", name)
+			_ = fs.Set("name", name)
```
BenTheElder, line 15: This can error?? What's the case?

pkg/cluster/internal/providers/docker/network.go
BenTheElder: I don't want to change the default CIDR to satisfy some overzealous linter.

amwat: let's just drop gosec instead

pkg/internal/cli/override.go
amwat: if the flag doesn't exist or something

BenTheElder: seems like we should like, panic then.

pkg/cluster/internal/providers/docker/network.go
```diff
@@ -128,8 +128,8 @@ func generateULASubnetFromName(name string, attempt int32) string {
 	ip[0] = 0xfc
 	ip[1] = 0x00
 	h := sha1.New()
-	h.Write([]byte(name))
-	binary.Write(h, binary.LittleEndian, attempt)
+	_, _ = h.Write([]byte(name))
+	_ = binary.Write(h, binary.LittleEndian, attempt)
```
BenTheElder, line 132: I think this can only happen on failure to allocate, I guess this is OK.

hack/verify/lint.sh
```diff
@@ -30,10 +30,10 @@ LINTS=(
   deadcode errcheck gosimple govet ineffassign staticcheck \
   structcheck typecheck unused varcheck \
   # additional lints
-  golint gofmt misspell gochecknoinits unparam scopelint gosec
+  golint gofmt misspell gochecknoinits unparam scopelint
```
BenTheElder, line 33: should probably check if there are more new ones we should enable in a follow up

hack/ci/e2e-k8s.sh
```diff
@@ -184,7 +184,7 @@ EOF
 
   # Patch kube-proxy to set the verbosity level
   kubectl patch -n kube-system daemonset/kube-proxy \
-    --type='json' -p='[{"op": "add", "path": "/spec/template/spec/containers/0/command/-", "value": "--v='${KIND_CLUSTER_LOG_LEVEL}'" }]'
+    --type='json' -p='[{"op": "add", "path": "/spec/template/spec/containers/0/command/-", "value": "--v='"${KIND_CLUSTER_LOG_LEVEL}"'" }]'
```
BenTheElder, line 187: ahhh bash

BenTheElder:
bypassing e2e flakes

PR #1762, opened 2020-08-05
Fix issue with Podman

izzyleung:
Unfortunately, change introduced in https://github.com/kubernetes-sigs/kind/commit/f5accf2b8e3429842796b7d9865e21d85060cc8a#diff-0c2fe782e83627c7cff34af1e84124d0 rendered Podman functionality broken.

Reason being: 
```sh
$ podman --version
podman version 2.0.4

$ podman help run
...
  -v, --volume stringArray                       Bind mount a volume into the container
      --volumes-from strings                     Mount volumes from the specified container(s)
```

Podman is expecting a comma separated list of volumes to bind mount, however, the mount options for `/var/lib/containerd`, `/var/lib/kubelet`, and `/var/log` will confuse the command line arg parser, making it trying to decypher what it is not supposed to do and not being able to create the container (it just sits at "Preparing nodes 📦" in an endless loop).

Tested with:
```
$ uname -a
Linux Debian 5.7.0-2-amd64 #1 SMP Debian 5.7.10-1 (2020-07-26) x86_64 GNU/Linux

$ cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux bullseye/sid"
NAME="Debian GNU/Linux"
ID=debian

$ ./kind --version
kind version 0.9.0-alpha+add83858a0adde

$ podman --version
podman version 2.0.4
```

### DIFF
--- a/images/base/files/etc/default/kubelet
+++ b/images/base/files/etc/default/kubelet
@@ -1,0 +1,1 @@
+KUBELET_EXTRA_ARGS=--fail-swap-on=false
--- a/pkg/build/nodeimage/build_impl.go
+++ b/pkg/build/nodeimage/build_impl.go
@@ -148,14 +148,6 @@ func (c *buildContext) buildImage(dir string) error {
 		return err
 	}
 
-	// ensure we don't fail if swap is enabled on the host
-	if err = execInBuild("/bin/sh", "-c",
-		`echo "KUBELET_EXTRA_ARGS=--fail-swap-on=false" >> /etc/default/kubelet`,
-	); err != nil {
-		c.logger.Errorf("Image build Failed! Failed to add kubelet extra args: %v", err)
-		return err
-	}
-
 	// pre-pull images that were not part of the build
 	images, err := c.prePullImages(dir, containerID)
 	if err != nil {
--- a/pkg/build/nodeimage/defaults.go
+++ b/pkg/build/nodeimage/defaults.go
@@ -20,7 +20,7 @@ package nodeimage
 const DefaultImage = "kindest/node:latest"
 
 // DefaultBaseImage is the default base image used
-const DefaultBaseImage = "kindest/base:v20200726-5f02d4ed"
+const DefaultBaseImage = "kindest/base:v20200729-c7fbe8df"
 
 // DefaultMode is the default kubernetes build mode for the built image
 // see pkg/build/kube.Bits


izzyleung:
I signed it

izzyleung:
An alternative way would be to just do things like what we do with the `docker` provider and yolo the mounts with

```go
func runArgsForNode(node *config.Node, clusterIPFamily config.ClusterIPFamily, name string, args []string) ([]string, error) {
	args = append([]string{
		"run",
		"--hostname", name, // make hostname match container name
		"--name", name, // ... and set the container name
		"--label", fmt.Sprintf("%s=%s", nodeRoleLabelKey, node.Role),
		// running containers in a container requires privileged
		// NOTE: we could try to replicate this with --cap-add, and use less
		// privileges, but this flag also changes some mounts that are necessary
		"--privileged",
		// runtime temporary storage
		// runtime persistent storage
		// this ensures that E.G. pods, logs etc. are not on the container
		// filesystem, which is not only better for performance, but allows
		// running kind in kind for "party tricks"
		// (please don't depend on doing this though!)
		"--volume", "/var/lib/containerd",
		"--volume", "/var/lib/kubelet",
		"--volume", "/var/log",
		// some k8s things want to read /lib/modules
	},
		args...,
	)

	// convert mounts and port mappings to container run args
	args = append(args, generateMountBindings(node.ExtraMounts...)...)
	mappingArgs, err := generatePortMappings(clusterIPFamily, node.ExtraPortMappings...)
	if err != nil {
		return nil, err
	}
	args = append(args, mappingArgs...)

	// finally, specify the image to run
	_, image := sanitizeImage(node.Image)
	return append(args, image), nil
}
```

Although I don't know how to test out the party trick... 🤔

pkg/cluster/internal/providers/podman/provision.go
```diff
@@ -191,9 +191,7 @@ func runArgsForNode(node *config.Node, clusterIPFamily config.ClusterIPFamily, n
 		// suid: SUID applications on the volume will be able to change their privilege
 		// exec: executables on the volume will be able to executed within the container
 		// dev: devices on the volume will be able to be used by processes within the container
-		"--volume", fmt.Sprintf("%s:/var/lib/containerd:suid,exec,dev", volume),
-		"--volume", fmt.Sprintf("%s:/var/lib/kubelet:suid,exec,dev", volume),
-		"--volume", fmt.Sprintf("%s:/var/log:suid,exec,dev", volume),
+		"--volume", fmt.Sprintf("%s:/var:suid,exec,dev", volume),
```
neolit123, line 194: i'm slightly confused about this change.
instead of mounting separate directories under /var we now mount /var as a whole, which seem fine, but i don't understand what actual problem with podman we are fixing.

if podman is not accepting multiple --volume arguments properly this seems like a podman bug?

- for the infinite loop your are seeing, is there an indication of what is the actual problem?
- is this reproducible with different podman versions?

BenTheElder:
What do you mean "just yolo the mounts"?
Do you mean not supply options to the mounts? Because we only need to supply those in podman because of podman breaking with docker on them.

If podman doesn't support repeated volume flags, that seems like a bug in podman. They should fix their parser 🤨

BenTheElder:
Ah, my brain is awake now 🙃

Instead the fix is we need to generate N named volumes per node instead of one and delete them all.

We can't use Anonymous Volumes (I now assume yolo refers to this) in podman because previous podman versions had extremely bad behavior for this, instead of an anonymous volume the path is bind mounted to the same path on the host.

izzyleung:
Closing temporarily as I have another fix for Podman incoming... Sorry about the trouble. 😢

BenTheElder:
thanks for working on it!

PR #1763, opened 2020-08-05
Fix issue with Podman deleting volumes

izzyleung:
Fix the issue with Podman failing to delete cluster as it failed to correctly delete all anonymous volumes.

# Old behavior
``` $ sudo ./kind delete cluster -v 1000000
enabling experimental podman provider
Deleting cluster "kind" ...
ERROR: failed to delete cluster "kind": command "podman volume rm --force '000fcd56f075c129d8c0fbfb6190923cafd20a65ed1bdbb9f27777e99bc25129
022fe51cd05427cd741a0355d0cbfdee230f8a314fa2fbcccc0c57b2e66cb7c4
10597316f16a8fb4227da9d6ccbd986074937e0ec3be2f088cc53419c1070d73
1933e143a1ca617b4002a3d55647a6eb3a11d7eb725887ee62ad363b0fc61be8
20e2090dc98ba9f278049d58b985d63bb7ffad7a5bf1b16b7cf14d5434055e25
2fa3d617bac948e7c3703e600a87a1e10b6543ca7530d4b1bd2434dc47dda2f9
2fbee920787e978c391756150a5b0b5725b5d4739c2bf717326203ddac545170
3b757825f62a14aae723a61d94d5d9ed3fb24d61a0418935efa7bc3b69810c81
56c369d51bc61bac4aecfa9023f39dc03882152c5b4f62f9f6bd67d1a489139f
643b0a80296d1e8d4d03db48961d135eca0ea67f358f5ae3579c9a3fd2252c72
6eb42729f37ae9ace9bc63af94be298dbcf562ba9d637f19e096e675c5e16a75
7d0826c99eb0fa7eff86bf60a2d674b11d985c95e9a5889423356712c264f569
82e607be9e77f7c843d27d6dc45e77bd3ae430de0395e9212d1cead335aca231
86ae7ed35e2f1fb168e9e44f97f9f1abfcc15875ee07beb3d845f3fcb65702ef
935f5df8fa29b9090cb4c29461e125ccbbe7303767598277e9bf9dc9f194a426
99b8b2e78642aa44403efd2d86c44d4eb12d1c9996584af01d01b4220ef72c4b
aa6f601dd8410240dc0a777617511c499f3477069bb89ae3a72635f6490cfa8b
ae13c601621fe905dabe75b460857f4e503adf420744b083401cd995fae78fe7
ae30d747734b266bec0302539b332232bd8d33a415d12dfbe713747ed20162c4
[...]

PR #1764, opened 2020-08-06
Separate volumes for the different mounts.

amwat:
xref: https://github.com/kubernetes-sigs/kind/pull/1762#issue-463255891

/cc @BenTheElder

### DIFF
--- a/pkg/cluster/internal/providers/podman/provision.go
+++ b/pkg/cluster/internal/providers/podman/provision.go
@@ -162,7 +162,19 @@ func commonArgs(cfg *config.Cluster) ([]string, error) {
 }
 
 func runArgsForNode(node *config.Node, clusterIPFamily config.ClusterIPFamily, name string, args []string) ([]string, error) {
-	volume, err := createAnonymousVolume(name)
+	// Pre-create anonymous volumes to enable specifying mount options
+	// during container run time
+	containerdVolume, err := createAnonymousVolume(name)
+	if err != nil {
+		return nil, err
+	}
+
+	kubeletVolume, err := createAnonymousVolume(name)
+	if err != nil {
+		return nil, err
+	}
+
+	logVolume, err := createAnonymousVolume(name)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


izzyleung:
Haha, you beat me to it, was about to push my change 😅 

But I belive this PR depends on the change in #1763

aojea:
> Haha, you beat me to it, was about to push my change 
> 
> But I believe this PR depends on the change in #1763

without this you are not able to create a cluster :), 
without #1763 you are not able to delete it
we should have both in

izzyleung:
> > Haha, you beat me to it, was about to push my change
> > But I believe this PR depends on the change in #1763
> 
> without this you are not able to create a cluster :),
> without #1763 you are not able to delete it
> we should have both in

https://giphy.com/gifs/leonardo-dicaprio-ellen-page-inception-bUkXxGkGYb5bq

Seriously, was testing out various way of creating volumes with Podman and planned to either push to my PR or create a new one. 

But I think Amit was able to see through my intention in #1763 and snapped it before me. 😅

aojea:
there are still more thing to do in podman, don't worry about that ;)

aojea:
@izzyleung i.e. https://github.com/kubernetes-sigs/kind/pull/1747#discussion_r463273049

amwat:
@izzyleung Apologies, I didn't realize that you were working on it simultaneously 😅 
You are right we need both. I added your commit (preserving author) to this PR to make it self-contained, if that's okay.

izzyleung:
> @izzyleung Apologies, I didn't realize that you were working on it simultaneously 😅
> You are right we need both. I added your commit (preserving author) to this PR to make it self-contained, if that's okay.

@amwat Please go ahead, I have closed my PR in favor of yours (I don't have the permission to run re-tests as my PR is showing failing some tests, and since this PR is passing all the tests, it has better chance of getting merged soon).

pkg/cluster/internal/providers/podman/provision.go
```diff
@@ -162,7 +162,19 @@ func commonArgs(cfg *config.Cluster) ([]string, error) {
 }
 
 func runArgsForNode(node *config.Node, clusterIPFamily config.ClusterIPFamily, name string, args []string) ([]string, error) {
-	volume, err := createAnonymousVolume(name)
+	// Pre-create anonymous volumes to enable specifying mount options
+	// during container run time
+	containerdVolume, err := createAnonymousVolume(name)
```
BenTheElder, line 167: for future cleanup: instead of N calls to this I think we should have a wrapper generate N volumes to distribute amongst mounts / basically abstract """"anonymous volumes""" to more closely map to what we're doing in docker.

I think it wouldn't be hard to make a little API around this that does the allocation in one call (and one error check) to make it less painful to add more of these.

BenTheElder:
/lgtm
/approve
thank you all

pkg/cluster/internal/providers/podman/provision.go
izzyleung: After this PR gets merged, I can help to create another one that refactor this code. 

My idea is to create an array/map in `constants.go` that keeps track of all the volumes with their mount point, then destroy them when destroying the cluster by looking up the very field.

Let me know if this sounds like a good idea.

BenTheElder:
i think it should be fully dynamic and deletion shouldn't depend on it

>

izzyleung:
> i think it should be fully dynamic and deletion shouldn't depend on it
> […](#)

That is fine, they are `AnonymousVolume`s anyway.

aojea:
/retest
duh, 1 hour crazy trying to understand why podman was not working until I realized this was not merged
:upside_down_face:

PR #1766, opened 2020-08-06
Podman load docker images

aojea:
podman: save and load images
    
    the kind load docker-image was not working in podman because it
    was using docker commands directly.
    
    This moves the helper functions used by the kind command to methods
    in the Provider interface, so it works both for podman and docker.

### DIFF
--- a/pkg/cluster/internal/providers/provider/provider.go
+++ b/pkg/cluster/internal/providers/provider/provider.go
@@ -45,4 +45,8 @@ type Provider interface {
 	GetAPIServerInternalEndpoint(cluster string) (string, error)
 	// CollectLogs will populate dir with cluster logs and other debug files
 	CollectLogs(dir string, nodes []nodes.Node) error
+	// SaveImage saves image to dest, as in `docker save`
+	SaveImage(image, dest string) error
+	// ImageID return the Id of the container image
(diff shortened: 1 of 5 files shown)


BenTheElder:
/hold
I think this breaks the meaning of the command, I have a better idea, one sec.

pkg/cluster/internal/providers/provider/provider.go
```diff
@@ -45,4 +45,8 @@ type Provider interface {
 	GetAPIServerInternalEndpoint(cluster string) (string, error)
 	// CollectLogs will populate dir with cluster logs and other debug files
 	CollectLogs(dir string, nodes []nodes.Node) error
+	// SaveImage saves image to dest, as in `docker save`
+	SaveImage(image, dest string) error
```
BenTheElder, line 49: Node providers cannot all provide this anyhow, if I have a `kubectl` + pods based provider, then I want the images to come from the local machine.

Instead we need to cover all the local machine when obtaining images. We can provide a unified UX for that.

aojea:
test
/area provider/podman

aojea:
@BenTheElder how do you suggest to move on with this?
load images is important for CIs

PR #1767, opened 2020-08-06
improve node images documentation: building, customizing, ...

aojea:
kind allows to build your own images, but if you only wants to add
additional software to your images you just can create new images
on top of the current ones.

It also moves the node images building docs to its own section to stop
growing the quickstart.

### DIFF
--- a/site/content/docs/user/node-images.md
+++ b/site/content/docs/user/node-images.md
@@ -1,0 +1,143 @@
+---
+title: "Node Images"
+menu:
+  main:
+    parent: "user"
+    identifier: "node-images"
+    weight: 3
+---
+# Node Images
+
+kind runs a local Kubernetes cluster by using Docker containers as "nodes".
+kind uses the [`node-image`][node image] to run Kubernetes artifacts, such
+as `kubeadm` or `kubelet`.
+The `node-image` in turn is built off the [`base-image`][base image], which
+installs all the dependencies needed for Docker and Kubernetes to run in a
+container.
+
+See [building the base image](#building-the-base-image) for more advanced information.
+
+## Building Images
+
+> Note: If you're using Docker Desktop, be sure to read [Settings for Docker Desktop](#settings-for-docker-desktop) first.
+
+
+Currently, kind supports two different ways to build a `node-image`
+if you have the [Kubernetes][kubernetes] source in your host machine
+(`$GOPATH/src/k8s.io/kubernetes`), by using `docker` or `bazel`.
+To specify the build type use the flag `--type`.
+Note however that using `--type=bazel` on Windows or MacOS will not work
+currently due to Kubelet using [CGO] which requires GCC/glibc for linux.
+A workaround may be enabled in the future.
+
+kind will default to using the build type `docker` if none is specified.
+
+```
+kind build node-image --type bazel
+```
+
+Similarly as for the base-image command, you can specify the name and tag of
+the resulting node image using the flag `--image`.
+
+If you previously changed the name and tag of the base image, you can use here
+the flag `--base-image` to specify the name and tag you used.
+
+### Settings for Docker Desktop
+
+If you are building Kubernetes (for example - `kind build node-image`) on MacOS or Windows then you need a minimum of 6GB of RAM
+dedicated to the virtual machine (VM) running the Docker engine. 8GB is recommended.
+
+To change the resource limits for the Docker on Mac, you'll need to open the
+**Preferences** menu.
+<img src="/docs/user/images/docker-pref-1.png"/>
+
+Now, go to the **Advanced** settings page, and change the
+settings there, see [changing Docker's resource limits][Docker resource lims].
+<img src="/docs/user/images/docker-pref-2.png" alt="Setting 8Gb of memory in Docker for Mac" />
+
+
+To change the resource limits for the Docker on Windows, you'll need to right-click the Moby
+icon on the taskbar, and choose "Settings". If you see "Switch to Linux Containers", then you'll need
+to do that first before opening "Settings"
+
+<img src="/docs/user/images/docker-pref-1-win.png"/>
+
+Now, go to the **Advanced** settings page, and change the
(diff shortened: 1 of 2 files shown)


site/content/docs/user/quick-start.md
```diff
@@ -232,14 +231,64 @@ to do that first before opening "Settings"
...
 
 <img src="/docs/user/images/docker-pref-build-win.png" alt="Setting 8Gb of memory in Docker for Windows" />
 
 
 You may also try removing any unused data left by the Docker engine - e.g.,
 `docker system prune`.
 
+## Customizing node images
+
+kind images are heavily optimized to be as lighter as possible, this has the downside that some common programs
```
neolit123, line 244: > light as possible. This has ...

site/content/docs/user/quick-start.md
```diff
@@ -232,14 +231,64 @@ to do that first before opening "Settings"
...
+are missing, i.e. an editor, ping, ... There is also users that need more specific tools for their clusters or
```
neolit123, line 245: > , such as a editor, `ping`, etc.

site/content/docs/user/quick-start.md
```diff
@@ -232,14 +231,64 @@ to do that first before opening "Settings"
...
+want to embed additional container images to avoid loading once the cluster is up.
+
+It is easy to create a new node image based on a existing one with the new software, just leveraging the
```
neolit123, line 248: "new" -> "additional" ?

site/content/docs/user/quick-start.md
```diff
@@ -232,14 +231,64 @@ to do that first before opening "Settings"
...
+containers capabilities, per example, creating a new Dockerfile that uses the official image as base and adds
```
neolit123, line 249: > . For example,

site/content/docs/user/quick-start.md
```diff
@@ -232,14 +231,64 @@ to do that first before opening "Settings"
...
+an step to install the additional software we need:
```
neolit123, line 250: > a step to install the additional software that is needed:

(removed the personalization to be consistent with the above context)

site/content/docs/user/quick-start.md
```diff
@@ -232,14 +231,64 @@ to do that first before opening "Settings"
...
+FROM kindest/node:v1.18.2
+RUN apt-get update && apt-get install -y \
+    tcpdump \
+    vim \
+ && rm -rf /var/lib/apt/lists/*
+
+COPY my-docker-image.tar /kind
+```
+
+We just need to build the new image with the tag that we want to use
```
neolit123, line 262: it would be best to either personalize the context above or:

> Then to build the new node image:

site/content/docs/user/quick-start.md
```diff
@@ -232,14 +231,64 @@ to do that first before opening "Settings"
...
+
+```sh
+docker build -t mykindnode:image .
```
neolit123, line 265: should this be something like `mykindnodeimage:sometag`?

site/content/docs/user/quick-start.md
```diff
@@ -232,14 +231,64 @@ to do that first before opening "Settings"
...
+```
+
+And use it to create our cluster:
```
neolit123, line 268: "our" vs "the"

site/content/docs/user/quick-start.md
```diff
@@ -232,14 +231,64 @@ to do that first before opening "Settings"
...
+ ✓ Preparing nodes 📦
+ ✓ Writing configuration 📜
+ ✓ Starting control-plane 🕹️
+ ✓ Installing CNI 🔌
+ ✓ Installing StorageClass 💾
+Set kubectl context to "kind-kind"
+You can now use your cluster with:
+kubectl cluster-info --context kind-kind
+Have a nice day! 👋
+$ docker exec -it kind-control-plane tcpdump -i any -c 1
```
neolit123, line 283: should we include a less technical example, such as `ping`?

BenTheElder:
I'm really not sure we should do this.

If users are familiar with docker they'll do this anyhow, but I don't see what we gain by advertising this

Billy99:
I spent a couple of hours trying to add tcpdump to kind/images/base by editing the Dockerfile and still had issues. Then @aojea showed me this easier route. Seems obvious after the fact, but I think it is a good hint to add. My 2 cents.

site/content/docs/user/quick-start.md
```diff
@@ -232,14 +231,64 @@ to do that first before opening "Settings"
...
-settings there, see [changing Docker's resource limits][Docker resource lims].  
+settings there, see [changing Docker's resource limits][Docker resource lims].
 
 <img src="/docs/user/images/docker-pref-build-win.png" alt="Setting 8Gb of memory in Docker for Windows" />
 
 
 You may also try removing any unused data left by the Docker engine - e.g.,
 `docker system prune`.
 
+## Customizing node images
```
BenTheElder, line 242: I would like to stop making this document longer. this is not quick-start material, most users should not ever do this.

site/content/docs/user/node-images.md
```diff
@@ -0,0 +1,135 @@
...
+Build the new image and tag it:
+
+```sh
+docker build -t mykindnode:mytag .
+```
+
+And use it to create the cluster:
+
+```sh
+kind create cluster --image mykindnode:mytag
```
izzyleung, line 100: Nit, I think this line is supposed to be `$ kind create cluster --image mykindnode:mytag`.

Notice the missing `$` at front?

site/content/docs/user/quick-start.md
aojea: I've moved it to its own section

site/content/docs/user/node-images.md
aojea: a quick look at the docs seems to not use it, so I split it in 2 code sections instead of adding the dollar, seems less confusing.
Good catch

PR #1768, opened 2020-08-06
Make rules

BenTheElder:
we need this for a follow-up, where we stop running these shell scripts directly and improve how we build with go.
they're also handy to have.

I moved the shell scripts to make it clearer that they're meant to be run under make now.

/cc @amwat @aojea


BenTheElder:
we need https://github.com/kubernetes/test-infra/pull/18704 to update CI

aojea:
/lgtm
/hold
so you can unhold when CI get the changes

BenTheElder:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/1768/pull-kind-verify/1291428015640678401 there we go

PR #1769, opened 2020-08-06
replace go_container.sh with gimme

BenTheElder:
speeds up builds. streamlines a lot of the tools (they can just invoke go without containerization hoops).
still should be reproducible with go1.13+

also the .go-version file should be goenv compatible


Makefile
```diff
@@ -13,36 +13,48 @@
...
+# record the source commit in the binary, overridable
+COMMIT?=$(shell git rev-parse HEAD 2>/dev/null)
+################################################################################
+# ========================= Setup Go With Gimme ================================
+# go version to use for build etc.
+# setup correct go version with gimme
+PATH:=$(shell . hack/build/setup-go.sh && echo "$${PATH}")
+export PATH
+# work around broken PATH export
+SHELL:=env PATH=$(PATH) $(SHELL)
```
BenTheElder, line 37: this is the trickiest bit, at least on macOS default `make`, without this you fail to actually set path in trivial commands.

https://stackoverflow.com/questions/8941110/how-i-could-add-dir-to-path-in-makefile put me on the right track

Makefile
```diff
@@ -13,36 +13,48 @@
...
 # the output binary name, overridden when cross compiling
 KIND_BINARY_NAME?=kind
-# the container cli to use e.g. docker,podman
-DOCKER?=$(shell which docker || which podman || echo "docker")
-export DOCKER
-# standard "make" target -> builds
-all: build
+# build flags for the kind binary
+# - reproducible builds: -trimpath and -ldlflags=-buildid=
+# - smaller binaries: -w (trim debugger data, but not panics)
```
BenTheElder, line 48: forgot to mention in the PR body, 9.1M => 7M on macOS

Makefile
```diff
@@ -13,36 +13,48 @@
...
+# get the repo root and output path
 REPO_ROOT:=${CURDIR}
 OUT_DIR=$(REPO_ROOT)/bin
+# record the source commit in the binary, overridable
+COMMIT?=$(shell git rev-parse HEAD 2>/dev/null)
+################################################################################
+# ========================= Setup Go With Gimme ================================
+# go version to use for build etc.
+# setup correct go version with gimme
+PATH:=$(shell . hack/build/setup-go.sh && echo "$${PATH}")
```
BenTheElder, line 34: - GOROOT is not required since go.19, go will look next to the go binary on it's own
- I don't think GOOS / GOARCH are relevant here
- PATH is how we inject the go version we want

This part is also not _striclty_ necessary after a later commit, where scripts try to ensure this themselves

aledbf:
@BenTheElder why are you adding the gimme script?

BenTheElder:
> @BenTheElder why are you adding the gimme script?

to version it and avoid having to download it from the internet each time

Makefile
BenTheElder: @amwat pointed out that gvm sets this (and probably other tools, including gimme when used natively), and when set to a different value in the host environment that causes us to potentially use the gimme go binary against some other goroot. we mitigate that below now by clearing GOROOT so it will be auto-detected instead.

BenTheElder:
not really related to this PR I think, but amwat discovered some issues with go1.15 and trying to use vendor if vendor is present.

aojea:
I prefer this to the docker build honestly, defer to @amwat the final lgtm
LGTM

BenTheElder:
/retest
pod pending timeout (prow)

BenTheElder:
verify now has an inexplicable go.sum issue after https://github.com/kubernetes-sigs/kind/commit/da5b1900e82bff7279998f6c3c9e605ab0f02235
https://prow.k8s.io/pr-history/?org=kubernetes-sigs&repo=kind&pr=1769

Makefile
```diff
@@ -53,17 +69,14 @@ install: build
...
-clean-cache:
-	$(DOCKER) volume rm -f kind-build-cache >/dev/null
-# cleans the output directory
-clean-output:
-	rm -rf $(OUT_DIR)/
 # standard cleanup target
-clean: clean-output clean-cache
+clean:
+	rm -rf $(OUT_DIR)/
+	# TODO: remove this in the future. We no longer populat vendor
```
amwat, line 78: nit: dedent
this currently prints out as a shell command

BenTheElder:
fixed it https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/1769/pull-kind-verify/1291498273856557056

hack/build/setup-go.sh
```diff
@@ -0,0 +1,32 @@
...
+
+# script to setup go version with gimme as needed
+# MUST BE RUN FROM THE REPO ROOT DIRECTORY
+
+# read go-version file unless GO_VERSION is set
+GO_VERSION="${GO_VERSION:-"$(cat .go-version)"}"
+
+# only setup go if we haven't set FORCE_HOST_GO, or `go version` doesn't match
+# go version output looks like:
+# go version go1.14.5 darwin/amd64
```
amwat, line 24: something somewhere is outputting the go version, can't seem to pinpoint where :/

BenTheElder: it's related to gimme, I've failed to silence it properly
gimme has an env to disable this, but it only works if used during the first time setting up that version, they don't regenerate the .evn file. not sure what the best fix is yet.

BenTheElder: it is mostly cosmetic though

BenTheElder: fixed in https://github.com/kubernetes-sigs/kind/pull/1769/commits/83d19a41281d0eeea0871c94039a0a3fc7c73bd3

Makefile
BenTheElder: fixed

BenTheElder:
what's one more flake
/retest

justaugustus:
(I speedread this, but I just wanted to drop by and say I'm excited to see it!)

BenTheElder:
this is allowing us to upgrade go versions immediately, instead of waiting a fairly long time for the `golang` docker images to be live at a given version.

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/1771 is within 30m of the release announcement (just when I got around to it) instead of checking back repeatedly for the image to be live

Issue #1778, opened 2020-08-12
Podman roadmap to stable


aojea:
## Podman roadmap to stable

Currently, Podman provider is set as Experimental, mainly because of subtle differences with Docker that breaks some of the features.

To fix this, KIND implements an interface, so providers can diverge in their implementation maintaining its compatibility at the user level.

In order to move to stable the provider it has to have exactly the same capabilities that the Docker provider.

These are the work items I have identified:

- [x] Implement CI. This runs every night from the KIND master branch testing IPv4, IPv6 single and multinode clusters with HA.
https://github.com/aojea/kind/actions?query=workflow%3APodman
- [ ] Podman networking, this is a topic where podman deviates completely from docker, it's a long history, libnetwork vs CNI. Since 0.8.0 KIND moved to their own bridge, this is missing in podman yet, assign @aojea 
- [ ] Load docker images, this needs further discussion https://github.com/kubernetes-sigs/kind/pull/1766
- ...

aojea:
/area provider/podman
/cc @amwat 

For those interested please report the missing features that you find

amwat:
This looks great!
Also we should definitely try to see if we can get those github actions running upstream. 
thoughts @BenTheElder ?

BenTheElder:
I'm in favor of this.

afbjorklund:
>  Windows, Mac support (I do not know if this is even possible

It is _possible_, but you would have to manage the VM yourself since podman only supports a "remote" and not a integrated solution

There are also lots of question marks when it comes to port mapping and such, so thinking of dropping the feature from minikube:

https://github.com/kubernetes/minikube/issues/8003

Right now we are pinning 1.9.3, due to the 2.0.x bugs

BenTheElder:
I don't think we should worry about windows / Mac for now.

We're supporting 2.X in kind 0.8

afbjorklund:
Our issues with podman 2.x was mostly about `podman-remote`

The podman driver (for minikube) is more or less back now,
but we will probably support both of them for a while (2020)

We are hoping that most of the worst regressions are fixed in 2.1

aojea:
here we go https://github.com/containers/podman/pull/8143

afbjorklund:
I have added a similar story for minikube, we are hoping for podman v3....

https://github.com/kubernetes/minikube/issues/10237

BenTheElder:
[...]

PR #1781, opened 2020-08-13
add podman ci

aojea:
add basic ci to test that kind can create clusters with podman as provider.

It can be improved to:
- run e2e tests
- ...
xref: #1778

### DIFF
--- a/.github/workflows/podman.yml
+++ b/.github/workflows/podman.yml
@@ -1,0 +1,100 @@
+name: Podman
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+
+  podman:
+    name: Podman
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        ipFamily: [ipv4, ipv6]
+        deployment: [singleNode, multiNode]
+        exclude:
+          # exclude IPv6
+          - ipFamily: ipv6
+    env:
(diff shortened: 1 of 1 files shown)


neolit123:
IIRC, one downside of actions is that there is no way to retest them by regular contributors without merge access in a repository.

aojea:
> why not run the multi node variants only?

multinode supersedes single node but are harder to debug, also this multinodes exercise HA that adds another layer of complexity with the load balancer

> IIRC, one downside of actions is that there is no way to retest them by regular contributors without merge access in a repository.

I had to go deep into actions lately, and the good thing is that they run on forks, so contributors can test their PR in their forks, this line allows to run the tests manually choosing a branch
https://github.com/kubernetes-sigs/kind/pull/1781/files#diff-4efd11c22b7577e551fe0058f6f7ed81R4

neolit123:
> I had to go deep into actions lately, and the good thing is that they run on forks, so contributors can test their PR in their forks, this line allows to run the tests manually choosing a branch

that is a win, for sure.

aojea:
/retest
pvc test

BenTheElder:
Do we need to do anything on the repo to enable this?

aojea:
/hold
Will enable only the working jobs

I don't thinking we need anything, I already see an actions tab in the repo, but I don't know if there are additional steps, it always worked for me

aojea:
/hold cancel
it seems that is enabled by default but it may be disabled per org and repo
https://docs.github.com/en/actions/getting-started-with-github-actions/about-github-actions#disabling-or-limiting-github-actions-for-your-repository-or-organization

BenTheElder:
Odd...
/retest

PR #1782, opened 2020-08-14
rework build interfaces

BenTheElder:
the build code is still kludgey, but this refines the interfaces around the kubernetes build / artifacts

it does simplify the build a tiny bit as well (no more install context)

xref: #396 #166


pkg/build/nodeimage/build_impl.go
```diff
@@ -133,23 +87,34 @@ func (c *buildContext) buildImage(dir string) error {
 	}
 
 	// copy artifacts in
-	if err = execInBuild("rsync", "-r", "/build/bits/", "/kind/"); err != nil {
```
aojea, line 136: is this a win? moving from `rsync` to `docker cp`?

pkg/build/nodeimage/internal/kube/builder_bazel.go
```diff
@@ -0,0 +1,135 @@
...
+		return nil, err
+	}
+	// make sure we cd back when done
+	defer func() {
+		// TODO(bentheelder): set return error?
+		_ = os.Chdir(cwd)
+	}()
+	if err := os.Chdir(b.kubeRoot); err != nil {
+		return nil, err
+	}
```
aojea, line 64: this is basically
```
pushd /my/folder/ 
....
popd
```

TIL windows have them https://en.wikipedia.org/wiki/Pushd_and_popd

aojea:
I've got this output running with the docker builder
```
+++ [0814 16:34:13] Running build command...
stat /go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/bazel-bin: no such file or directory                    
+++ [0814 16:34:49] Building go targets for linux/amd64:
    cmd/kube-apiserver
```
I suspect it is coming from the k/k scripts, right?

BenTheElder:
yes

pkg/build/nodeimage/internal/kube/builder_bazel.go
BenTheElder: A) this code isn't new.
B) we should eventually stop CD-ing the whole binary
C) those are builtins in a shell, we need to change it in the current process

pkg/build/nodeimage/build_impl.go
BenTheElder: yes and no. rsync on a volume mount might? be faster. but what's done here is less generic and simpler if you consider beyond this line.

but now we are not setting up this complex temp directory thing, it only remains for exporting tarballs from pulled images.

instead of syncing in a directory tree and setting up symlinks into path, we just move executables into path and make them executable, and images loaded directly. specific speciality files written specially (like version).

we can explore doing the equivilant of this build without docker commit later now.

Issue #1732, opened 2020-07-14
[docker for mac]: custom apiServerAddress yields kubeconfig pointing to 127.0.0.1 due to docker regression


vdice:
Greetings!  I've been utilizing the ability to specify custom `apiServerAddress` and `apiServerPort` values as listed in https://kind.sigs.k8s.io/docs/user/configuration/#api-server

While the API Server's listening address indeed appears updated (API server communication is successful after manual updates mentioned below), neither variant of the resulting kubeconfig has a `server` value that uses the custom settings.

With kind version `kind v0.9.0-alpha+2b0b8c07769f88 go1.15beta1 darwin/amd64`, say I set the aforementioned values to:

```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
networking:
  apiServerAddress: 10.0.1.4
  apiServerPort: 6443
```

and create the cluster as usual (`kind create cluster --config kind-config.yaml`),

`kind get kubeconfig` has `server: https://127.0.0.1:32777` and
`kind get kubeconfig --internal` has `server: https://kind-control-plane:6443`

The first is automatically merged/placed into the `$KUBECONFIG` location, but neither work to communicate with the cluster.

Only after manually updating the kubeconfig will communications work:

```console
 $ kubectl config view -o json | jq -r '.clusters[0].cluster.server'
https://127.0.0.1:32778

 $ kubectl cluster-info

To further debug and diagnose cluster problems, use 'kubectl cluster-info dump'.
The connection to the server 127.0.0.1:32778 was refused - did you specify the right host or port?

 $ kubectl config set clusters.kind-kind.server https://10.0.1.4:6443
Property "clusters.kind-kind.server" set.

 $ kubectl cluster-info
Kubernetes master is running at https://10.0.1.4:6443
KubeDNS is running at https://10.0.1.4:6443/api/v1/namespaces/kube-system/services/kube-dns:dns/proxy

To further debug and diagnose cluster problems, use 'kubectl cluster-info dump'.
```

My question is thus: Should not the stock kubeconfig returned via `kind get kubeconfig` use a `server` address built from the provided `apiServerAddress` and `apiServerPort` values?

BenTheElder:
Hi, can you share some more details about your environment please?
It should be using those values in the default.

vdice:
Definitely.  Let me know if there is any other info that would help.

[...]

PR #1785, opened 2020-08-17
feat(docker/provider): defer to desktop.docker.io label for api endpoint, if present

vdice:
* For the Docker provider, adds logic to defer to the `docker.desktop.io/ports/6443/tcp` label value for determining the API server endpoint

Proposed fix for https://github.com/kubernetes-sigs/kind/issues/1732 (also reference this issue for background context)

### DIFF
--- a/pkg/cluster/internal/providers/docker/provider.go
+++ b/pkg/cluster/internal/providers/docker/provider.go
@@ -157,15 +157,37 @@ func (p *Provider) GetAPIServerEndpoint(cluster string) (string, error) {
 		return "", errors.Wrap(err, "failed to get api server endpoint")
 	}
 
-	// retrieve the specific port mapping using docker inspect
+	// if the 'desktop.docker.io/ports/<PORT>/tcp' label is present,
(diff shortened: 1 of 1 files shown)


BenTheElder:
I think we could do this with one call but it might be a little gross and can be sorted out in a follow-up.

I plan to rewrite the bootstrapping logic to retain more information and abstract things out less internally anyhow, so it's  perhaps not worth it.

BenTheElder:
thank you!

vdice:
> I think we could do this with one call

Ah yes, good idea... all via an extended go template?  Will defer to you, but sounds like perhaps a follow-up and/or rework is currently the favored option.

vdice:
I'm having trouble parsing the test failure(s), but able/ready to address if pointers can be provided.  Thanks!

BenTheElder:
CI / Kubernetes test issues, they're not related to your PR, don't worry about it, sorry about that! :sweat_smile: 

let's get this in as-is

PR #1786, opened 2020-08-18
increase etcd backend quota to match GCE / kube-up

BenTheElder:
xref: https://github.com/kubernetes/kubernetes/issues/94029

TODO:
- should this be bumped in kubeadm upstream in future versions?
- should we change the healthcheck behavior?
- can we reduce the storage usage, should we adjust the compaction behavior?

### DIFF
--- a/pkg/cluster/internal/kubeadm/config.go
+++ b/pkg/cluster/internal/kubeadm/config.go
@@ -140,6 +140,14 @@ scheduler:
 networking:
   podSubnet: "{{ .PodSubnet }}"
   serviceSubnet: "{{ .ServiceSubnet }}"
+etcd:
+  # TODO: external etcd / HA ..?
+  local:
+    extraArgs:
+      # this matches GCE
+      # https://github.com/kubernetes/kubernetes/commit/ca99cbca02542ae63e72f5be84d23078e7422f85
+      # https://github.com/kubernetes/kubernetes/issues/94029
+      "quota-backend-bytes": "4294967296"
 ---
 apiVersion: kubeadm.k8s.io/v1beta1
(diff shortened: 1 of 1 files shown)


BenTheElder:
/retest
bazel build errors downloadding images

aojea:
I think is better to do this in the e2e script if we want to bump it and not make it default in kind.

As an example I have one project running kind with etcd in memory, and most CI and users don't have lot of resources and disk.

The problem only happens in a test that is hard anybody using kind may hit it, and I don't the that the solution to the issue is just bumping the database size, since we are storing 1g of useless information.

Please hold this, I have an idea to fix the original issue

aojea:
/hold
I suggest to hold until we have more reports about this

BenTheElder:
> As an example I have one project running kind with etcd in memory, and most CI and users don't have lot of resources and disk.

We will still be compacting AIUI, this just avoids etcd refusing to write further when we hit this.
Since you're enabling in-memory, you can also patch this flag to have a different limit.

That said, the subnet change is better, and hopefully we don't _need_ this now.
I may re-open it in the future. I don't want arbitrary barriers to kind succeeding vs GCE, and this limit is exceedingly arbitrary.

BenTheElder:
See discussion on: https://github.com/kubernetes/test-infra/pull/18956 and https://kubernetes.slack.com/archives/C09QZTRH7/p1598291758010000?thread_ts=1597992971.004900&cid=C09QZTRH7

BenTheElder:
willing to close this for now, but i'm still suspicious of the defaults here. for testing we should consider more regular apiserver compaction as well.

Issue #1567, opened 2020-05-06
[podman] External Load Balancer for HA cluster


aojea:
**What should be cleaned up or changed**:

with https://github.com/kubernetes-sigs/kind/pull/1564 the external load balancer is tied to the docker provisioner

**Why is this needed**:

allow loadbalancer to restart gracefully with podman too

PR #1790, opened 2020-08-18
Podman network

aojea:
Since 0.8.0, KIND uses custom networks with docker to leverage the
embedded DNS server and other features.

This PR provides podman with the same functionality  impemented in docker
Fixes: #1567

### DIFF
--- a/pkg/cluster/internal/create/actions/config/config.go
+++ b/pkg/cluster/internal/create/actions/config/config.go
@@ -201,7 +201,7 @@ func getKubeadmConfig(cfg *config.Cluster, data kubeadm.ConfigData, node nodes.N
 	// configure the right protocol addresses
 	if cfg.Networking.IPFamily == "ipv6" {
 		if nodeAddressIPv6 == "" {
-			return "", errors.Errorf("failed to get IPV6 address; is the docker daemon configured to use IPV6 correctly?")
+			return "", errors.Errorf("failed to get IPV6 address; is the container provider (docker,podman) configured to use IPV6 correctly?")
 		}
 		data.NodeAddress = nodeAddressIPv6
 	}
--- a/pkg/cluster/internal/providers/podman/node.go
+++ b/pkg/cluster/internal/providers/podman/node.go
@@ -53,7 +53,7 @@ func (n *node) Role() (string, error) {
 func (n *node) IP() (ipv4 string, ipv6 string, err error) {
 	// retrieve the IP address of the node using podman inspect
 	cmd := exec.Command("podman", "inspect",
-		"-f", "{{.NetworkSettings.IPAddress}},{{.NetworkSettings.GlobalIPv6Address}}",
+		"-f", "{{range .NetworkSettings.Networks}}{{.IPAddress}},{{.GlobalIPv6Address}}{{end}}",
 		n.name, // ... against the "node" container
 	)
 	lines, err := exec.OutputLines(cmd)
--- a/.github/workflows/podman.yml
+++ b/.github/workflows/podman.yml
@@ -7,7 +7,6 @@ on:
       - master
 
 jobs:
-
   podman:
     name: Podman
     runs-on: ubuntu-latest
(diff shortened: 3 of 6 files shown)


pkg/cluster/internal/providers/docker/provider.go
```diff
@@ -253,3 +253,8 @@ func (p *Provider) CollectLogs(dir string, nodes []nodes.Node) error {
 	errs = append(errs, errors.AggregateConcurrent(fns))
 	return errors.NewAggregate(errs)
 }
+
+// DeleteNetwork does nothing for docker
```
BenTheElder, line 257: We should not do this.
DeleteNodes in podman should clean up the network if no nodes are left.

BenTheElder: Or the API at provider level should be DeleteCluster (resources only not kubeconfig) and we figure out the per node stuff later.

aojea: ack

aojea:
in functionality terms, the only thing missing is the HA deployment, haproxy is hardcoded to use the docker resolver 127.0.0.1:53 , in podman it has the ip of the gw :shrug:

aojea:
missing the port forwarding on ipv6 only :)

aojea:
/assign @amwat @BenTheElder 
PTAL

aojea:
I've decided to stop wrestling with IPv6, I can take a look later, 
but this would give podman "feature parity" with docker and IPv4

pkg/cluster/internal/providers/podman/util.go
```diff
@@ -102,7 +103,12 @@ func getVolumes(label string) ([]string, error) {
 	// Trim away the last `\n`.
 	trimmedOutput := strings.TrimSuffix(string(output), "\n")
 	// Get names of all volumes by splitting via `\n`.
-	return strings.Split(string(trimmedOutput), "\n"), nil
+	for _, v := range strings.Split(string(trimmedOutput), "\n") {
+		if v != "" {
```
aojea, line 107: it parses the load balancer container and adds an empty volume to the list, we can remove all the empty volumes to prevent the same issue in the future

pkg/cluster/internal/providers/podman/node.go
```diff
@@ -63,7 +63,10 @@ func (n *node) IP() (ipv4 string, ipv6 string, err error) {
 	if len(lines) != 1 {
 		return "", "", errors.Errorf("file should only be one line, got %d lines", len(lines))
 	}
-	ips := strings.Split(lines[0], ",")
+	// TODO: investigate where the double quotes are added
+	// it does not seem to happen running from the CLI
```
BenTheElder, line 67: a call specifying format has different output with TTY or not!?

pkg/cluster/internal/loadbalancer/config.go
```diff
@@ -37,8 +38,12 @@ global
   log /dev/log local1 notice
   daemon
 
-resolvers docker
+resolvers kinddns
+  {{ if $.Docker -}}
```
BenTheElder, line 42: : /

pkg/cluster/internal/create/actions/loadbalancer/loadbalancer.go
```diff
@@ -74,11 +75,19 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
...
+	dockerProvider := true
+	if _, ok := ctx.Provider.(*podman.Provider); ok {
+		dockerProvider = false
+	}
 	// create loadbalancer config data
 	loadbalancerConfig, err := loadbalancer.Config(&loadbalancer.ConfigData{
 		ControlPlanePort: common.APIServerInternalPort,
 		BackendServers:   backendServers,
 		IPv6:             ctx.Config.Networking.IPFamily == config.IPv6Family,
+		Docker:           dockerProvider,
```
BenTheElder, line 90: this code should not be provider aware, this sort of thing should be encapsulated in the node provider.

pkg/cluster/internal/providers/podman/util.go
BenTheElder: there's definitely a more efficient way to do this, multiple actually.
https://github.com/golang/go/wiki/SliceTricks#filter-in-place

BenTheElder:
> It creates one network per cluster. The network has the same lifecycle than the nodes

are we unable to use one network?

using one network means less CIDRs need to be allocated by the user and I would expect it to generally be easier to work with.

aojea:
> > It creates one network per cluster. The network has the same lifecycle than the nodes
> 
> are we unable to use one network?
> 
> using one network means less CIDRs need to be allocated by the user and I would expect it to generally be easier to work with.

podman only support single stack networks, ipv4 or ipv6, that means that for the one bridge model we have to create one per each family
 podman does not isolate container networks, like docker, and I found this model more powerful for networking setups and I like it more.
I think that each provider is different with different user base, and I'm start to think that podman will be more attractive in the networking area, since it has less constraints and it's Linux only,

pkg/cluster/internal/providers/podman/node.go
aojea: solved removing the quotes and using a rune

pkg/cluster/internal/providers/podman/util.go
aojea: done

pkg/cluster/internal/loadbalancer/config.go
aojea: it works for both providers adding both resolvers :shrug:

BenTheElder:
/hold
we chatted a bit about the approach last night, need to summarize here but TLDR need to rethink per-cluster network.
let's see how #1831 lands in 0.9 and visit this in 0.10 along with a broader aim to stabilize podman

aojea:
changed Podman to use same approach than docker
blocked by
https://github.com/containers/podman/issues/8444

aojea:
/hold cancel
Forget to remove it, this is ok for now, I will fix ipv6 later
Ping @BenTheElder

pkg/cluster/internal/create/actions/config/config.go
```diff
@@ -201,7 +201,7 @@ func getKubeadmConfig(cfg *config.Cluster, data kubeadm.ConfigData, node nodes.N
 	// configure the right protocol addresses
 	if cfg.Networking.IPFamily == "ipv6" {
 		if nodeAddressIPv6 == "" {
-			return "", errors.Errorf("failed to get IPV6 address; is the docker daemon configured to use IPV6 correctly?")
+			return "", errors.Errorf("failed to get IPV6 address; is the container provider (docker,podman) configured to use IPV6 correctly?")
```
BenTheElder, line 204: TODO: we can insert the name of the one we're using here, there's a .String() on provider

BenTheElder: (we want to avoid useless detail / sending people looking at the wrong thing)

BenTheElder:
one nit for a follow-up
/lgtm
/approve

pkg/cluster/internal/create/actions/config/config.go
aojea: https://github.com/kubernetes-sigs/kind/pull/1960

Issue #1792, opened 2020-08-19
go get sigs.k8s.io/kind@v0.8.1: can't request explicit version of path in main module


shay-berman:
```
$ go version
go version go1.13.8 linux/amd64
$ GO111MODULE="on" go get sigs.k8s.io/kind@v0.8.1
go get sigs.k8s.io/kind@v0.8.1: can't request explicit version of path in main module

```

**What happened**:
failed to run `go get sigs.k8s.io/kind@v0.8.1` as mentioned above.

**What you expected to happen**:
to be able to get kind.

**How to reproduce it (as minimally and precisely as possible)**:
as mentioned above.

**Anything else we need to know?**:
its really the first few lines in the Kind README -> https://github.com/kubernetes-sigs/kind/blob/master/README.md

**Environment:**

- kind version: (use `kind version`):
- Kubernetes version: (use `kubectl version`):
- Docker version: (use `docker info`):
docker info
Client:
 Debug Mode: false

Server:
 Containers: 25
  Running: 0
  Paused: 0
  Stopped: 25
 Images: 698
 Server Version: 19.03.6
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 7ad184331fa3e55e52b890ea95e65ba581ae3429
 runc version: dc9208a3303feef5b3839f4323d9beb36df0a9dd
 init version: fec3683
 Security Options:
  apparmor
  seccomp
   Profile: default
 Kernel Version: 5.4.0-40-generic
 Operating System: Ubuntu 20.04 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 46.73GiB
 Name: shayb-thinkpad-t490
 ID: VU4S:2J5A:TLEO:ZKO4:A646:CINX:WHL3:VJHM:E7UZ:RB4A:OAKU:JYZM
 Docker Root Dir: /var/lib/docker
 Debug Mode: false
 Username: shayberman
 Registry: https://index.docker.io/v1/
 Labels:
 Experimental: false
 Insecure Registries:
[...]

PR #1793, opened 2020-08-19
add a note about go get inside go module projects

BenTheElder:
fixes https://github.com/kubernetes-sigs/kind/issues/1792


BenTheElder:
podman multinode fails on a documentation PR @aojea
logs are not too helpful.

BenTheElder:
at least it doesn't block.
what is the ACL for re-running these?

PR #1798, opened 2020-08-20
Improve Dockerfile for base

izzyleung:
# Changes:
- Unify binary/utility installation logic
  - Unify `curl` syntax/arguments.
  - Unify `tar` syntax/arguments
- Remove a duplicate export statement for `ARCH`

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -95,14 +95,16 @@ RUN echo "Ensuring scripts are executable ..." \
     && runc --version \
     && systemctl enable containerd \
  && echo "Installing crictl ..." \
-    && curl -fSL "https://github.com/kubernetes-sigs/cri-tools/releases/download/${CRICTL_VERSION}/crictl-${CRICTL_VERSION}-linux-${ARCH}.tar.gz" | tar xzC /usr/local/bin \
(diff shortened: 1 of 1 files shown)


images/base/Dockerfile
```diff
@@ -95,14 +95,16 @@ RUN echo "Ensuring scripts are executable ..." \
     && runc --version \
     && systemctl enable containerd \
  && echo "Installing crictl ..." \
-    && curl -fSL "https://github.com/kubernetes-sigs/cri-tools/releases/download/${CRICTL_VERSION}/crictl-${CRICTL_VERSION}-linux-${ARCH}.tar.gz" | tar xzC /usr/local/bin \
+    && export CRICTL_URL="https://github.com/kubernetes-sigs/cri-tools/releases/download/${CRICTL_VERSION}/crictl-${CRICTL_VERSION}-linux-${ARCH}.tar.gz" \
+    && curl -sSL --retry 5 --output /tmp/crictl.tgz "${CRICTL_URL}" \
+    && tar -C /usr/local/bin -xzvf /tmp/crictl.tgz \
+    && rm -rf /tmp/crictl.tgz \
  && echo "Installing CNI binaries ..." \
-    && export ARCH=$(dpkg --print-architecture | sed 's/ppc64el/ppc64le/' | sed 's/armhf/arm/') \
```
BenTheElder, line 100: we shouldn't have a "step" depend on the contents of another step.
if we're going to dedupe this, we should move it much earlier.

izzyleung: Sounds good, I can move this statement up.

PR #1805, opened 2020-08-24
set providerID

### LINKED ISSUE #1711: set providerID
set kubelet's `--provider-id` (or the component config equivalent)

**Why is this needed**:
"Real" clusters typically set this, and tools may read it for identification.

NOTE: this may not be so trivial, given kubeadm not having per-node kubelet config ... xref: https://github.com/kubernetes-sigs/kind/issues/1424 / https://github.com/kubernetes-sigs/kind/issues/1270 => https://github.com/kubernetes/kubeadm/issues/1682

BenTheElder:
closes #1711 

```
$ kubectl describe no | grep -i provider
ProviderID:                  kind://docker/kind/kind-control-plane
```

### DIFF
--- a/pkg/cluster/internal/create/actions/config/config.go
+++ b/pkg/cluster/internal/create/actions/config/config.go
@@ -28,6 +28,7 @@ import (
 	"sigs.k8s.io/kind/pkg/cluster/internal/create/actions"
 	"sigs.k8s.io/kind/pkg/cluster/internal/kubeadm"
 	"sigs.k8s.io/kind/pkg/cluster/internal/patch"
+	"sigs.k8s.io/kind/pkg/cluster/internal/providers/podman"
 	"sigs.k8s.io/kind/pkg/cluster/internal/providers/provider/common"
 	"sigs.k8s.io/kind/pkg/cluster/nodeutils"
 	"sigs.k8s.io/kind/pkg/internal/apis/config"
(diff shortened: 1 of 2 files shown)


pkg/cluster/internal/create/actions/config/config.go
```diff
@@ -59,7 +60,16 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 	// create kubeadm init config
 	fns := []func() error{}
 
+	// TODO(bentheelder): settle on a better abstraction for this
+	// While we don't want much code aware of the provider name,
+	// we do want to surface it in the node providerID field
+	providerName := "docker"
```
BenTheElder, line 66: really hesitant to encourage further reliance on this, but I think it makes sense to surface this data in the providerID field and I want to ship that.
We can replace this with something else later (maybe a public method that isn't officially part of the interface or something, need to think on that)

BenTheElder: Perhaps `String()`

amwat: Yeah, Stringer probably seems good enough.

BenTheElder: https://github.com/kubernetes-sigs/kind/pull/1809 done

PR #1810, opened 2020-08-26
Bump images

BenTheElder:
catches up on minor base image changes, so we're shipping the latest for v0.9.0


BenTheElder:
CI flake (pod pending timeout)

we should consider guaranteed QoS for our presubmits, matching k/k.

Issue #1425, opened 2020-03-19
Retry kubeconfig modification if locking fails


Arnavion:
**What would you like to be added**:

Currently, if `kind` is unable to lock the kubeconfig during `create cluster` or `delete cluster`, it just gives up on the modification and fails. It would be nice if it retried until it succeeded. It's fine if it's opt-in via a new option to the two commands instead of the default.

**Why is this needed**:

My workflow involves deploying many clusters to the same machine, and it would be nice if I was able to parallelize it by trivially running `kind` in parallel.

---

Workaround:

- For `create cluster`, pass it `--kubeconfig` with a throwaway tempfile so that it doesn't modify the real `~/.kube/config`. Then, after all the clusters have been created, run `kind export kubeconfig` serialized to update the real `~/.kube/config`.

- For `delete cluster`, pass it `--kubeconfig` with a throwaway tempfile so that it doesn't modify the real `~/.kube/config`. Then, after all the clusters have been deleted, run `kubectl config delete-cluster`, `delete-context` and `unset users.<>` serialized to update the real `~/.kube/config`.

BenTheElder:
I think a single retry if the failure is specifically due to locking would be a reasonable approach, we don't want to hang forever on this but it does stink to fail setting up kubeconfig or cluster creation due to this race.

The current code is from a very close following of `client-go`, the official implementation of this for general purpose usage (which we don't pull in to avoid chicken-and-egg issues with testing kubernetes and to keep our tiny binary size down), this was definitely an oversight.

Arnavion:
A single retry wouldn't be enough. I agree that you don't want to hang forever, which is why I suggested adding an option for it so that it's opt-in.

BenTheElder:
I don't think users should be specifying yet another flag for this. Though we should follow whatever the best practices from the existing tooling. A lot of tools write to this file; I'll look again into how other tools in the project handle this (e.g. `kops`).

We need to lock it for a pretty short period of time, and only once per invocation.

Arnavion:
>We need to lock it for a pretty short period of time, and only once per invocation.

Yes, and nevertheless I see the lock errors very reliably every time I deploy, so just retrying once will probably also only let one instance of `kind` make progress and the rest will fail again.

BenTheElder:
only if there's insufficient jitter to the retry. s/once/some bounded number of times/

PR #1811, opened 2020-08-26
retry kubeconfig export in kind create cluster

BenTheElder:
fixes https://github.com/kubernetes-sigs/kind/issues/1425
/milestone v0.9.0

### DIFF
--- a/pkg/cluster/internal/create/create.go
+++ b/pkg/cluster/internal/create/create.go
@@ -155,7 +155,17 @@ func Cluster(logger log.Logger, p providers.Provider, opts *ClusterOptions) erro
 		return nil
 	}
 
-	if err := kubeconfig.Export(p, opts.Config.Name, opts.KubeconfigPath); err != nil {
+	// try exporting kubeconfig with backoff for locking failures
+	// TODO: factor out into a public errors API w/ backoff handling?
+	// for now this is easier than coming up with a good API
+	var err error
+	for _, b := range []time.Duration{time.Millisecond, time.Millisecond * 50, time.Millisecond * 100} {
+		if err = kubeconfig.Export(p, opts.Config.Name, opts.KubeconfigPath); err == nil {
+			break
+		}
+		time.Sleep(b)
+	}
+	if err != nil {
 		return err
 	}
 


pkg/cluster/internal/create/create.go
```diff
@@ -155,7 +155,17 @@ func Cluster(logger log.Logger, p providers.Provider, opts *ClusterOptions) erro
 		return nil
 	}
 
-	if err := kubeconfig.Export(p, opts.Config.Name, opts.KubeconfigPath); err != nil {
+	// try exporting kubeconfig with backoff for locking failures
+	// TODO: factor out into a public errors API w/ backoff handling?
+	// for now this is easier than coming up with a good API
+	var err error
+	for _, b := range []time.Duration{time.Millisecond, time.Millisecond * 50, time.Millisecond * 100} {
```
BenTheElder, line 162: these times are fairly arbitrary, but a strict improvement over the current behavior.

we don't need to do any significant computation during this time, just lock the files, read them in, write out the patched file with the kubeconfig merged in.

BenTheElder: it would probably be best to not retry for certain classes of failure, but this is much simpler to get implemented, and the harm is minimal, it will take slightly longer to fail if it's e.g. a permissions issue.

BenTheElder:
I didn't touch `kind export kubeconfig`, users/callers can retry that themselves for now since that's the only thing being done.

something to consider for the future, but much less critical than this, which will fail cluster creation and doesn't otherwise have a user option to granularly ignore / retry

BenTheElder:
this is not going to affect the e2e tests.

pkg/cluster/internal/create/create.go
```diff
@@ -155,7 +155,17 @@ func Cluster(logger log.Logger, p providers.Provider, opts *ClusterOptions) erro
...
+		if err = kubeconfig.Export(p, opts.Config.Name, opts.KubeconfigPath); err == nil {
+			break
+		}
+		time.Sleep(b)
```
amwat, line 166: nit: a sleep after the last failed attempt is unnecessary maybe move it first?

BenTheElder: done

BenTheElder: oops it already merged, filing a follow-up

BenTheElder: https://github.com/kubernetes-sigs/kind/pull/1814

PR #1813, opened 2020-08-26
revert back to persisting all of /var

BenTheElder:
we forgot some things, after digging some more I think we just want to stick to this for now.


aojea:
less is more :+1: 
/lgtm

Issue #1271, opened 2020-01-20
Document howto enable deprecated k8s APIs


nbartos:
For kubeadm I can do something like this to enable the deprecated APIs in k8s 1.16:

```
apiVersion: kubeadm.k8s.io/v1beta1
kind: ClusterConfiguration
apiServer:
  extraArgs:
    runtime-config: "apps/v1beta1=true,apps/v1beta2=true,extensions/v1beta1/daemonsets=true,extensions/v1beta1/deployments=true,extensions/v1beta1/replicasets=true,extensions/v1beta1/networkpolicies=true,extensions/v1beta1/podsecuritypolicies=true"
```

How would I go about creating a kind 1.16 cluster with these APIs enabled?

neolit123:
kind supports patches:
https://kind.sigs.k8s.io/docs/user/quick-start/#enable-feature-gates-in-your-cluster

/triage support
/kind documentation

nbartos:
Thanks, adding this to the config seems to work:
```
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
  kubeadmConfigPatches:
  - |
    apiVersion: kubeadm.k8s.io/v1beta2
    kind: ClusterConfiguration
    apiServer:
      extraArgs:
        runtime-config: "apps/v1beta1=true,apps/v1beta2=true,extensions/v1beta1/daemonsets=true,extensions/v1beta1/deployments=true,extensions/v1beta1/replicasets=true,extensions/v1beta1/networkpolicies=true,extensions/v1beta1/podsecuritypolicies=true"
```

I think this is a fairly common use case, having it in the docs would be nice.

neolit123:
> Thanks, adding this to the config seems to work

excellent.

> I think this is a fairly common use case, having it in the docs would be nice.

docs PRs are always welcome.

BenTheElder:
with kind v0.7.0 you can do the following kind config:
```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
kubeadmConfigPatches:
- |
  kind: ClusterConfiguration
  apiServer:
    extraArgs:
      runtime-config: "apps/v1beta1=true,apps/v1beta2=true,extensions/v1beta1/daemonsets=true,extensions/v1beta1/deployments=true,extensions/v1beta1/replicasets=true,extensions/v1beta1/networkpolicies=true,extensions/v1beta1/podsecuritypolicies=true"
```
[...]

PR #1816, opened 2020-08-26
Runtime config

BenTheElder:
ref: #1271


BenTheElder:
FYI @liggitt

BenTheElder:
/hold
make generated is broken. fixing

BenTheElder:
some weird output path issue with deepcopy-gen, I think it's trying to use the whole module path relative to the repo root.

pushed a work-around, also fixed a spelling error.

Issue #1719, opened 2020-07-09
Create cluster fails - kind-control-plane does not work on zfs


johnlane:
**What happened**: cluster create failed
```
failed to create cluster: failed to init node with kubeadm: command "docker exec --privileged kind-control-plane kubeadm init --ignore-preflight-errors=all --config=/kind/kubeadm.conf --skip-token-print --v=6" failed with error: exit status 1
```

**What you expected to happen**: successful cluster creation

**How to reproduce it (as minimally and precisely as possible)**:
```
$ kind create cluster 
```

**Anything else we need to know?**:

**I'm using ZFS**. I have read the other (now closed as resolved) issue about using zfs and I can see that `/dev/mapper` is bind-mounted into the `kind-control-plane` container. I'm running 0.8.1 so I believe that I should have the version that should work with ZFS.

**Environment:**

- kind version: (use `kind version`): 
`kind v0.8.1 go1.14.2 linux/amd64`
- Kubernetes version: (use `kubectl version`):
`Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"archive", BuildDate:"2020-04-23T22:11:11Z", GoVersion:"go1.14.2", Compiler:"gc", Platform:"linux/amd64"}`

- Docker version: (use `docker info`):
` Server Version: 19.03.8-ce
 containerd version: d76c121f76a5fc8a462dc64594aea72fe18e1178.m`

- OS (e.g. from `/etc/os-release`): Arch Linux
`5.6.11-arch1-1 #1 SMP PREEMPT Wed, 06 May 2020 17:32:37 +0000 x86_64 GNU/Linux`

To capture more information, I ran with `kind create cluster --loglevel=debug --retain`. Here is some further log data...

In `kind-control-plane/containerd.log`
```
[...]

PR #1818, opened 2020-08-26
rework containerd config

BenTheElder:
now that we skip preflight we can simply use a fixed pause image version controlled by containerd.

fixes: #1719 as well


pkg/build/nodeimage/build_impl.go
```diff
@@ -120,22 +120,6 @@ func (c *buildContext) buildImage(bits kube.Bits) error {
 		return err
 	}
 
-	// find the pause image and inject containerd config
```
aojea, line 123: seems you don't need know to prePullImags in Line 117

BenTheElder: yeah, will fix after the conformance ga job is fixed

BenTheElder:
> [sig-cli] Kubectl client Simple pod should support port-forward 

everything checks out now except one thing: the kubelet flag regarding the pause image may not match.
this flag does two things:
- configures dockershim to use it to implement pod sandboxes (not relevant to CRI, it's up to the implementation)
- configures the imageGC to exclude this image (and only this image... it's a special codepath)

the latter we might care about if we we enable imageGC someday, but we also have that issue with other images

images/base/files/usr/local/bin/entrypoint
```diff
@@ -18,6 +18,23 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+configure_containerd() {
+  # we need to switch to the 'native' snapshotter on zfs
+  if [[ "$(stat -f -c %T /kind)" == 'zfs' ]]; then
```
aojea, line 23: do we still need this https://github.com/kubernetes-sigs/kind/blob/master/pkg/cluster/internal/providers/docker/provision.go#L208 ?

BenTheElder: yes

aojea: do we need to handle btrfs too?

BenTheElder: we have prior art for zfs's best option being falling back to the native driver because the zfs mode is sketchy at least in the kind context and overlay driver doesn't work.
we don't for btrfs.

if someone reports this with evidence, we can add it. I _think_ btrfs is already fixed but can't confirm.

BenTheElder: see: https://github.com/kubernetes-sigs/kind/issues/1719

aojea:
/lgtm
/hold
unhold when necessary :)

images/base/files/usr/local/bin/entrypoint
```diff
@@ -18,6 +18,23 @@ set -o errexit
...
+    sed -i 's/snapshotter = "overlayfs"/snapshotter = "native"/' /etc/containerd/config.toml
```
BenTheElder, line 24: this line itself is fine except we forgot to put the explicit snapshotter in the config. https://github.com/kubernetes-sigs/kind/issues/1891

PR #1819, opened 2020-08-26
fix passing runtime_config

BenTheElder:
https://github.com/kubernetes/kubernetes/issues/94263


hack/ci/e2e-k8s.sh
```diff
@@ -144,7 +144,7 @@ nodes:
 - role: worker
 - role: worker
 featureGates: ${feature_gates}
-runtimeConfig: ${runtime_config}
```
BenTheElder, line 147: https://github.com/kubernetes-sigs/kind/pull/1820

PR #1823, opened 2020-08-27
Fix problem with spaces in $PATH

steigr:


Makefile
```diff
@@ -38,7 +38,7 @@ GOROOT:=
 GO111MODULE=on
 export PATH GOROOT GO111MODULE
 # work around broken PATH export
-SHELL:=env PATH=$(PATH) $(SHELL)
+SHELL:=env PATH="$(PATH)" $(SHELL)
```
BenTheElder, line 41: I'm pretty sure this adds literal quotes, this is a makefile, not a shell, so it's setting a literal single argument of `PATH="`+ contents of $PATH + `"` to `env`

BenTheElder:
can you elaborate on the problem you experienced?

steigr:
Ah. Sorry for missing that.

If there any spaces in paths inside your $PATH environment variable, make fails to build

```
> echo $PATH | cut -f10- -d: | cut -f-7 -d:
/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Applications/VMware Fusion.app/Contents/Public:/Library/TeX/texbin
                                                                ^^^ this breaks
>  make
env: Fusion.app/Contents/Public:/Library/TeX/texbin: No such file or directory
go build -v -o /Users/steigr/code/kind/bin/kind -trimpath -ldflags="-buildid= -w -X=sigs.k8s.io/kind/pkg/cmd/kind/version.GitCommit="
```

Makefile
steigr: Yes. Make is not as cumbersome as shell regarding quotes. But this will end up in `sh`, thus should suffice shell's syntax.

BenTheElder: the value `$(PATH)` winds up in sh, but if you add the quotes then the first and last path elements are actually just incorrect.

you can see how this PR breaks the path in CI for example.

BenTheElder:
that's no good, can you share your `make` version / details and anything else about your host?

Makefile
BenTheElder: ah just saw your edit, let's try the space subsitution trick in CI, I'll take a look locally as well.

BenTheElder:
the space trick works in my local environment, including if I add a path with spaces to PATH.

TIL how tricky it is to get a literal space in make...

steigr:
thanks for the example.

```
docker run --rm -ti -v$PWD:/Users/steigr/code/kind -w/Users/steigr/code/kind golang:1.15
root@cab16af00a8c:/Users/steigr/code/kind# export PATH=/usr/local/go/bin:/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/this\ breaks/make:/bin
#                                                               ^ 'go' lives here                                                                              ^  basic tools
root@cab16af00a8c:/Users/steigr/code/kind# head -n 45 Makefile | tail -n 5
SPACE:= $(empty) $(empty)
SHELL:=env PATH=$(subst $(SPACE),\$(SPACE),$(PATH)) $(SHELL)
	################################################################################
# ============================== OPTIONS =======================================
# install tool
root@cab16af00a8c:/Users/steigr/code/kind# make
go build -v -o /Users/steigr/code/kind/bin/kind -trimpath -ldflags="-buildid= -w -X=sigs.k8s.io/kind/pkg/cmd/kind/version.GitCommit=b80a914d0e0721a98d66967ce3e9e2181daedca4"
...
```

BenTheElder:
as an extra sanity check for your env: `go version bin/kind` (using any recent go version on the host) will tell us if the build picked up the correct go toolchain from the makefile.

it works correctly for me with the space trick:
```console
$ go version bin/kind
bin/kind: go1.15
```
```patch
diff --git a/Makefile b/Makefile
index ef02af51..dff405cb 100644
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,8 @@ GOROOT:=
 GO111MODULE=on
 export PATH GOROOT GO111MODULE
 # work around broken PATH export
-SHELL:=env PATH=$(PATH) $(SHELL)
+SPACE:=$(subst ,, )
+SHELL:=env PATH=$(subst $(SPACE),\$(SPACE),$(PATH)) $(SHELL)
 ################################################################################
 # ============================== OPTIONS =======================================
 # install tool
@@ -58,6 +59,7 @@ KIND_BUILD_FLAGS?=-trimpath -ldflags="-buildid= -w -X=sigs.k8s.io/kind/pkg/cmd/k
 all: build
 # builds kind in a container, outputs to $(OUT_DIR)
 kind:
+       go version
        go build -v -o $(OUT_DIR)/$(KIND_BINARY_NAME) $(KIND_BUILD_FLAGS)
 # alias for building kind
 build: kind
```

Makefile
```diff
@@ -38,7 +38,8 @@ GOROOT:=
 GO111MODULE=on
 export PATH GOROOT GO111MODULE
 # work around broken PATH export
-SHELL:=env PATH=$(PATH) $(SHELL)
+SPACE:= $(empty) $(empty)
```
BenTheElder, line 41: What do you think about `$(subst ,, )`?
It feels a little gross relying on `empty` being unset. https://stackoverflow.com/a/10571900

BenTheElder:
thanks for this, I didn't realize the makefile hack was introducing this bug, all of the scripts should be correctly handling this sort of thing otherwise 🙃 

/approve

Makefile
steigr: As I hope not too much people will look inside Makefile: I'm fine with subst, but $(empty) is more obvious (despite the variable should make things clear).

BenTheElder:
/lgtm
thanks!

BenTheElder:
FYI @amwat for kubernetes-sigs/kubetest2

Issue #1824, opened 2020-08-28
Unable to pull kindest/node with userns remapping enabled


vixus0:
**What happened**:
Attempted to start `kind` using a newer version of the `kindest/node` image. Pulling the image causes the following error:
```
failed to register layer: Error processing tar file(exit status 1): Container ID 484827 cannot be mapped to a host ID
```

**What you expected to happen**:
The image is pulled successfully and the cluster is created.

**How to reproduce it (as minimally and precisely as possible)**:
Setup user namespace remapping as described here: https://docs.docker.com/engine/security/userns-remap/
Then run:
```
docker pull kindest/node:v1.19.0@sha256:6a6e4d588db3c2873652f382465eeadc2644562a64659a1da4db73d3beaa8848
```

**Anything else we need to know?**:
I can successfully run kind with the default image version `kindest/node:v1.18.2`.

**Environment:**

- kind version: (use `kind version`): kind v0.8.1 go1.14.2 linux/amd64
- Kubernetes version: (use `kubectl version`): v1.18.8
- Docker version: (use `docker info`): 19.03.12
- OS (e.g. from `/etc/os-release`): Ubuntu 20.04.1

BenTheElder:
Neat. Looks like that means we have some high UID file in the export. Maybe the updated base image.

BenTheElder:
It appears to be due to the new build mechanism
```console
$ docker run --rm -it --entrypoint bash kindest/node:v1.19.0@sha256:6a6e4d588db3c2873652f382465eeadc2644562a64659a1da4db73d3beaa8848
root@1e216bec9c15:/# find / \( -uid 484827 \)  -ls 2>/dev/null
 26087029  81952 -rwxr-xr-x   1 484827   89939    83918152 Aug 26 18:54 /usr/bin/kubelet
 26087028  58964 -rwxr-xr-x   1 484827   89939    60373229 Aug 26 18:54 /usr/bin/kubectl
 26087027  54196 -rwxr-xr-x   1 484827   89939    55490758 Aug 26 18:54 /usr/bin/kubeadm
```

BenTheElder:
`kindest/node:v1.19.0@sha256:3b0289b2d1bab2cb9108645a006939d2f447a10ad2bb21919c332d06b548bbc6` should not have this issue.

BenTheElder:
https://circleci.com/docs/2.0/high-uid-error/ was very helpful for this, for future reference.

PR #1827, opened 2020-08-28
ensure root owns binaries in node image

BenTheElder:
fixes: https://github.com/kubernetes-sigs/kind/issues/1824

root was the owner in previous images. some build changes caused these to have a different UID.
see https://github.com/kubernetes-sigs/kind/issues/1824#issuecomment-683165201

### DIFF
--- a/pkg/build/nodeimage/build_impl.go
+++ b/pkg/build/nodeimage/build_impl.go
@@ -97,6 +97,9 @@ func (c *buildContext) buildImage(bits kube.Bits) error {
 		if err := execInBuild("chmod", "+x", nodePath); err != nil {
 			return err
 		}
+		if err := execInBuild("chown", "root:root", nodePath); err != nil {
+			return err
+		}
 	}
 
 	// write version


BenTheElder:
confirmed that built image no longer has the high UID kubelet/kubeadm/kubectl

pkg/build/nodeimage/build_impl.go
```diff
@@ -97,6 +97,9 @@ func (c *buildContext) buildImage(bits kube.Bits) error {
 		if err := execInBuild("chmod", "+x", nodePath); err != nil {
 			return err
 		}
+		if err := execInBuild("chown", "root:root", nodePath); err != nil {
```
aojea, line 100: do we need to do this recursive? `chown -R root:root`

BenTheElder: no, this is a full filepath.

aojea: ahh, nodePath points to binary, scratch thath

PR #1828, opened 2020-08-31
Add LICENSE to site examples

babilen5:
Adding an explicit license to the example directory clarifies the
licensing of code snippets that have been included in the
documentation.

The license of these examples was previously up to interpretation as
the presence of a LICENSE file in `site/` (cf. [0]) implied CC-BY-SA
4.0 even though licensing code under that license is recommended
against even by creative commons themselves (cf. [1]).

Including this LICENSE file makes it clear that examples are licensed
under ASLv2 which makes it easier to adapt and integrate their code
into other projects.

[0]: https://github.com/kubernetes-sigs/kind/blob/master/site/LICENSE
[1]: https://creativecommons.org/faq/#can-i-apply-a-creative-commons-license-to-software

### DIFF
--- a/site/static/examples/LICENSE
+++ b/site/static/examples/LICENSE
@@ -1,0 +1,395 @@
+Attribution 4.0 International
+
+=======================================================================
+
+Creative Commons Corporation ("Creative Commons") is not a law firm and
+does not provide legal services or legal advice. Distribution of
+Creative Commons public licenses does not create a lawyer-client or
+other relationship. Creative Commons makes its licenses and related
+information available on an "as-is" basis. Creative Commons gives no
+warranties regarding its licenses, any material licensed under their
+terms and conditions, or any related information. Creative Commons
+disclaims all liability for damages resulting from their use to the
+fullest extent possible.
+
+Using Creative Commons Public Licenses
+
+Creative Commons public licenses provide a standard set of terms and
+conditions that creators and other rights holders may use to share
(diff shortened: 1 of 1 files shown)


nikhita:
/hold

> Including this LICENSE file makes it clear that examples are licensed
under ASLv2 which makes it easier to adapt and integrate their code
into other projects.

ASLv2 is not an allowed third party license under the CNCF - https://github.com/cncf/foundation/blob/master/allowed-third-party-license-policy.md#approved-licenses-for-allowlist

babilen5:
Thank you, @nikhita I will clarify this with upstream.

neolit123:
Hi, so this PR adds a CC-BY-4.O license in the examples package, which overrides the implied parent ASLv2 of the project, the same way the site package is doing.

I wonder how the CNCF is allowing k/website to use that, but I am suspecting an exeption.
Should kind not be using cc-by-4 for its website?

> ASLv2 is not an allowed third party license under the CNCF -

Did you mean CC-BY-4.O?

> Thank you, @nikhita I will clarify this with upstream.

Did you mean downstream here? Upstream to kind are projects that kind uses like k/kubernetes.

nikhita:
@neolit123 there is a follow up PR here #1829 :)

babilen5:
Thanks for the clarification, the intention is indeed to include the Apache License, Version 2.0 in the examples directory

PR #1831, opened 2020-09-01
fix ambiguous networks

### LINKED ISSUE #1596: network kind is ambiguous
**What happened**:
Kicked off multiple builds in our CI environment, some tests use KIND to spin up clusters. Saw s bunch of failures:
```
  ✗ Preparing nodes 📦 
 ERROR: failed to create cluster: docker run error: command "docker run --hostname ci-a510791-control-plane --name ci-a510791-control-plane --label io.x-k8s.kind.role=control-plane --privileged --security-opt seccomp=unconfined --security-opt apparmor=unconfined --tmpfs /tmp --tmpfs /run --volume /var --volume /lib/modules:/lib/modules:ro --detach --tty --label io.x-k8s.kind.cluster=ci-a510791 --net kind --restart=on-failure:1 --volume=/workspace/pr-113/e2e/etc/rootca1.crt:/usr/local/share/ca-certificates/rootca1.crt:ro --volume=/workspace/pr-113/e2e/etc/rootca2.crt:/usr/local/share/ca-certificates/rootca2.crt:ro --publish=127.0.0.1:40759:6443/TCP kindest/node:v1.17.5" failed with error: exit status 125
 Command Output: 4614c0b36ac6a3e641b0a300d07b6b0bc7317132fab3d494d21a3e4777aa5d5a
 docker: Error response from daemon: network kind is ambiguous (2 matches found on name).
```

No interference between tests, as we experienced w/ KIND 0.7.x

* create many KIND clusters at the same time, starting from a clean box (w/o a kind network already created)?

```
# docker network ls
NETWORK ID          NAME                DRIVER              SCOPE
98fdc5e5af26        bridge              bridge              local
6f8de42fadad        host                host                local
de87cb7dd35e        kind                bridge              local
158648a47e91        kind                bridge              local
75afe274f8f8        none                null                local
```

**Environment:**

- kind version: (use `kind version`): `kind v0.8.1 go1.13.9 linux/amd64`
```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.5", GitCommit:"e0fccafd69541e3750d460ba0f9743b90336f24f", GitTreeState:"clean", BuildDate:"2020-04-16T11:44:03Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```
```
$ docker info
Client:
 Debug Mode: false

Server:
 Containers: 1
  Running: 1
  Paused: 0
  Stopped: 0
 Images: 122
 Server Version: 19.03.6
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Plugins:
  Volume: local
 Swarm: inactive
 Runtimes: runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version:
 runc version:
 init version:
 Security Options:
  apparmor
  seccomp
   Profile: default
 Kernel Version: 5.3.0-46-generic
 Operating System: Ubuntu 19.10
 OSType: linux
 Architecture: x86_64
 CPUs: 104
 Total Memory: 754.6GiB
 Name: 7959d5c46f-m9c7p
 ID: UMPE:ZM2Z:POMD:VQDB:7JAM:7OV5:LNJE:XP5W:EX4Z:CA5N:GO35:IBFY
 Debug Mode: false
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
```
```
NAME="Ubuntu"
VERSION="19.10 (Eoan Ermine)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 19.10"
VERSION_ID="19.10"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=eoan
UBUNTU_CODENAME=eoan
```

BenTheElder:
fixes #1596 

Rough algorithm:
- check if network exists if it does, end here
- create network
- check if there are now > 1 networks
- sort networks based on:
  - \# of attached containers (available when inspecting the network)
  - CreatedAt (creation timestamp, standard docker object metadata)
  - Id comparison (dockerd assigns UIDs as the primary identifier centrally)
- delete N-1 networks, keeping the one that sorts first

This ensures that we will remove the ambiguous extra networks when we have two `kind create cluster` "win the race" of `docker network create` and simultaneously create identically named networks within the timeframe that docker doesn't catch that the name already exists during the internal non-atomic check for an existing network by the name.

This should be safe because we deterministically sort based on "is this network being used somehow" then "when was this network created" and finally if those are identical on the UID docker assigns them.

TODO: 
- [x] I need to rework the network list / sort logic to list then inspect and sort by \# of attached containers, currently it only sorts by ID and CreatedAt
- ~[ ] There's still perhaps a race when doing the cleanup, where process A and B win the race, A wins it faster and somehow continues on to creating containers attached to this while B is going to cleanup. I'm not sure if this can actually happen and it's difficult to reproduce, but if it did we might get an error during container creation, in which case we should retry because the cleanup is coming. I think this isn't feasible as in all my experiments when I pull out the network logic and "win the race" by running it simultaneously in two goroutines I see that the following list check observes both networks and handles it, but this is probably still worth adding as a mitigation~

EDIT: The second case should not happen now. If somehow I'm wrong it will be a simple mitigation in the future though and this already provides a significant improvement (complete?) in avoiding #1596.


pkg/cluster/internal/providers/docker/network.go
```diff
@@ -102,6 +130,55 @@ func createNetwork(name, ipv6Subnet string) error {
...
+			break
+		} else if err != nil {
+			return nil, errors.Wrap(err, "failed to decode networks list")
+		}
+		networks = append(networks, network)
+	}
+
+	// deterministically sort networks
+	// NOTE: THIS PART IS IMPORTANT!
+	// TODO(fixme): we should be sorting on active usage first!
```
BenTheElder, line 165: this TODO is the one thing keeping this WIP.
I will come back and rework this later. the overall code will be more or less the same otherwise.

aojea:
let's verify it

```diff
diff --git a/pkg/cluster/internal/providers/docker/network_test.go b/pkg/cluster/internal/providers/docker/network_test.go
index 789a0d93..a5fb1d2b 100644
--- a/pkg/cluster/internal/providers/docker/network_test.go
+++ b/pkg/cluster/internal/providers/docker/network_test.go
@@ -19,8 +19,46 @@ package docker
 import (
        "fmt"
        "testing"
+
+       "sigs.k8s.io/kind/pkg/exec"
 )
 
+func TestEnsureNetworkConcurrent(t *testing.T) {
+       defer func() {
+               cmd := exec.Command(
+                       "docker", "network", "rm", "test-kind",
+               )
+               cmd.Run()
+       }()
+
+       // Create multiple networks concurrenctly
+       errCh := make(chan error, 3)
+       for i := 0; i < 3; i++ {
+               go func() {
+                       errCh <- ensureNetwork("test-kind")
+               }()
+       }
+       for i := 0; i < 3; i++ {
+               if err := <-errCh; err != nil {
+                       t.Errorf("error creating network: %v", err)
+               }
+       }
+
+       cmd := exec.Command(
+               "docker", "network", "ls",
+               "--filter=name=^test-kind$",
+               "--format={{.Name}}",
+       )
+
+       lines, err := exec.OutputLines(cmd)
+       if err != nil {
+               t.Errorf("obtaining the docker networks")
+       }
+       if len(lines) != 1 {
+               t.Errorf("wrong number of networks created: %d", len(lines))
+       }
+}
+
 func Test_generateULASubnetFromName(t *testing.T) {
        t.Parallel()
        cases := []struct {
```

aojea:
maybe the name has not to be unique but the subnet should be

```
=== RUN   TestEnsureNetworkConcurrent
    TestEnsureNetworkConcurrent: /home/aojea/go/src/sigs.k8s.io/kind/pkg/cluster/internal/providers/docker/network_test.go:43: error creating network: command "docker network create -d=bridge -o com.docker.network.bridge.enable_ip_masquerade=true --ipv6 --subnet fc00:3587:1532:ea5a::/64 test-kind" failed with error: exit status 1
    TestEnsureNetworkConcurrent: /home/aojea/go/src/sigs.k8s.io/kind/pkg/cluster/internal/providers/docker/network_test.go:43: error creating network: command "docker network create -d=bridge -o com.docker.network.bridge.enable_ip_masquerade=true --ipv6 --subnet fc00:3587:1532:ea5a::/64 test-kind" failed with error: exit status 1
--- FAIL: TestEnsureNetworkConcurrent (0.36s)
FAIL
coverage: 15.7% of statements
exit status 1
```

pkg/cluster/internal/providers/docker/network.go
```diff
@@ -42,13 +47,15 @@ const fixedNetworkName = "kind"
 
 // ensureNetwork checks if docker network by name exists, if not it creates it
 func ensureNetwork(name string) error {
-	// TODO: the network might already exist and not have ipv6 ... :|
-	// discussion: https://github.com/kubernetes-sigs/kind/pull/1508#discussion_r414594198
-	exists, err := checkIfNetworkExists(name)
+	// check if network exists already and remove any duplicate networks
+	exists, err := removeDuplicateNetworks(name)
```
aojea, line 51: at this point is hard to guess what is the good one, if there are multiple networks, there should have been created previously.
I think that this case is complex to handle and maybe we should just check if a network exists and pick one,

pkg/cluster/internal/providers/docker/network.go
```diff
@@ -91,6 +98,27 @@ func ensureNetwork(name string) error {
...
+func createNetworkNoDuplicates(name, ipv6Subnet string) error {
+	if err := createNetwork(name, ipv6Subnet); err != nil {
+		return err
+	}
+	_, err := removeDuplicateNetworks(name)
+	return err
+}
+
+func removeDuplicateNetworks(name string) (bool, error) {
+	networks, err := sortedNetworksWithName(name)
```
aojea, line 110: networksIDs?

pkg/cluster/internal/providers/docker/network.go
BenTheElder: The TODO is not new, only moved up.

This TODO is actually about the case that we previously created a network and it may have suboptimal config (like missing the ipmasquerade flag we added).

In this case we will NOT run the risk of duplicate networks (this can only happen if they're created concurrently) but we currently will continue to use the old suboptimal config.

What would possibly be ideal is something like: when the network is ensured we diff the settings vs desired and then automatically recreate it if it doesn't match and nothing is currently on the network. It needs some thought and still isn't handled, so the TODO was preserved when hoisting the existence check into this network list call here.

pkg/cluster/internal/providers/docker/network.go
BenTheElder: sure

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/1831#issuecomment-684977764

yeah, so AFAICT this only happens when we use the default IPAM (AKA the no-IPv6 case).

I'm not overly inclined to add unit tests that depend on docker / aren't self-contained just yet, we probably want to consider having an integration test pattern similar to kubernetes vs our unit and e2e testing.

aojea:
> I'm not overly inclined to add unit tests that depend on docker / aren't self-contained just yet, we probably want to consider having an integration test pattern similar to kubernetes vs our unit and e2e testing.

not asking to include it , just to run it before merge to be sure, it is not working for me ... I couldn't spend much time though

pkg/cluster/internal/providers/docker/network.go
aojea: if anybody is running with suboptimal config and didn't notice is that is optimal for him :smile: 

I will keep going with the TODO and iterate with more feedback, unless you had feedback about this and/or saw this behavior causing issues

BenTheElder:
picking this back up now.

BenTheElder:
testing with `go test -count=40 -run ^TestIntegrationEnsureNetworkConcurrent -v ./pkg/cluster/internal/providers/docker` is cropping up the odd occasion that no network exists when all the ensure calls are done. I haven't figured out that race yet.

BenTheElder:
I think it may have been in the timestamp parse/comparison. Since dropping that I've not had any failures in 160 iterations.

pkg/cluster/internal/providers/docker/network.go
BenTheElder: > if anybody is running with suboptimal config and didn't notice is that is optimal for him 😄

I mean some users are not aware of kind managing the network. In the future it might be good to at least hint that they should have it recreated.

pkg/cluster/internal/providers/docker/network_test.go
```diff
@@ -18,9 +18,65 @@ package docker
...
+	"strings"
 	"testing"
+
+	"sigs.k8s.io/kind/pkg/errors"
+	"sigs.k8s.io/kind/pkg/exec"
+
+	"sigs.k8s.io/kind/pkg/internal/integration"
 )
 
+func TestIntegrationEnsureNetworkConcurrent(t *testing.T) {
```
BenTheElder, line 31: We can consider reworking the make targets based on this pattern.

to run only "unit": `go test -short ...`
to run only "integration": `go test -run ^TestIntegration ...`

pkg/cluster/internal/providers/docker/network.go
```diff
@@ -69,8 +75,19 @@ func ensureNetwork(name string) error {
 	// If it is, make more attempts below
 	if isIPv6UnavailableError(err) {
 		// only one attempt, IPAM is automatic in ipv4 only
-		return createNetwork(name, "")
-	} else if !isPoolOverlapError(err) {
+		return createNetworkNoDuplicates(name, "")
```
aojea, line 78: why don't we check ` if isPoolOverlapError(err) ` like in the other places?

aojea: ahh, just read IPAM is automatic, hope ipam handles overlapping too

BenTheElder: When we pass "" for the IPv6 subnet to this method we don't enable IPv6, in IPv4 we let docker's internal allocator handle it. This is from before the PR.

BenTheElder:
/retest
CI flakes from being on the legacy build cluster. We should talk to CI signal about moving KIND testing over with the other jobs setting resource limits.

amwat (commented):
Great! This should mitigate the main ambiguous docker network issue :+1:
The other remaining possibilities of race conditions seem like weird corner cases which we'll most likely not run into in practice.

pkg/cluster/internal/providers/docker/network.go
```diff
@@ -154,51 +154,57 @@ func createNetwork(name, ipv6Subnet string) error {
...
-		"--format={{json .}}",
+		"--format={{.ID}}", // output as unambiguous IDs
 	))
 	if err != nil {
 		return nil, err
 	}
 
+	// now inspect each
+	ids := strings.Split(strings.TrimSuffix(string(lsOut), "\n"), "\n")
+	inspectOut, err := exec.Output(exec.Command("docker", append([]string{"network", "inspect"}, ids...)...))
```
aojea, line 167: _it is fine_  :smile:

aojea:
SGTM
just suggesting to squash some of the first commits so we can have a safe revert if necessary, not criticizing the PR, just the opposite, I think that this is a big improvement that tries to solve a well known docker broken behavior

BenTheElder:
currently refactoring to make https://github.com/kubernetes-sigs/kind/pull/1831#discussion_r487281833 a bit cleaner to address.

initially I didn't want to expose the inspect details but I think it's worth it to break these up and make them more testable, upon reflection.

pkg/cluster/internal/providers/docker/network.go
```diff
@@ -102,6 +151,55 @@ func createNetwork(name, ipv6Subnet string) error {
...
+
+	networks := []networkInspectEntry{}
+	if err := json.Unmarshal(inspectOut, &networks); err != nil {
+		return nil, errors.Wrap(err, "failed to decode networks list")
+	}
+
+	// deterministically sort networks
+	// NOTE: THIS PART IS IMPORTANT!
+	sort.Slice(networks, func(i, j int) bool {
+		if len(networks[i].Containers) < len(networks[j].Containers) {
```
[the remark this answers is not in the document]
BenTheElder: done.

amwat (commented):
/lgtm

BenTheElder:
I profiled this before the last commit and improved the performance slightly, the happy path is just as cheap as before now

We're going to need to do something about those two CI jobs, but they're unrelated to this PR.

PR #1836, opened 2020-09-08
Update to use stable shellcheck image

MaXinjian:
If we always use stable shellcheck image, there is no need to change shellcheck image version any more

Signed-off-by: Ma Xinjian <maxj.fnst@cn.fujitsu.com>


BenTheElder:
Thanks but I'd rather not.

manually updating the version prevents surprises where PRs start failing because shellcheck changed.

we pin all of our dependencies.

PR #1841, opened 2020-09-09
bazel: derive exec path of go binaries from action graph

mikedanese:
Output paths are not a stable API and all our conditional path
construction here is fragile. This replaces the old path construction
with queries to the action graph.

Part of unblocking: kubernetes/kubernetes#94449

cc @justaugustus @BenTheElder

### DIFF
--- a/pkg/build/nodeimage/internal/kube/builder_bazel.go
+++ b/pkg/build/nodeimage/internal/kube/builder_bazel.go
@@ -17,10 +17,12 @@ limitations under the License.
 package kube
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
 
+	"sigs.k8s.io/kind/pkg/errors"
 	"sigs.k8s.io/kind/pkg/exec"
 	"sigs.k8s.io/kind/pkg/log"
 )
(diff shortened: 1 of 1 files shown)


pkg/build/nodeimage/internal/kube/builder_bazel.go
```diff
@@ -81,42 +82,24 @@ func (b *bazelBuilder) Build() (Bits, error) {
...
-	// for building from release binaries?
-	// https://github.com/kubernetes/kubernetes/pull/73930
-	strippedCommandPath := func(command string) string {
-		return filepath.Join(
-			binDir, "cmd", command,
-			fmt.Sprintf("%s_stripped", bazelGoosGoarch), command,
-		)
+	kubeletPath, err := findGoBinary("//cmd/kubelet:kubelet")
+	if err != nil {
+		return nil, fmt.Errorf("could not find kubelet: %w", err)
```
neolit123, line 87: kind is using a custom package for errors:
https://github.com/kubernetes-sigs/kind/tree/master/pkg/errors/errors.go

e.g. the above should be `errors.Wrap(err, "could not find kubelet")`

mikedanese: Fixed.

pkg/build/nodeimage/internal/kube/builder_bazel.go
```diff
@@ -81,42 +83,27 @@ func (b *bazelBuilder) Build() (Bits, error) {
...
-	// https://github.com/kubernetes/kubernetes/pull/73930
-	strippedCommandPath := func(command string) string {
-		return filepath.Join(
-			binDir, "cmd", command,
-			fmt.Sprintf("%s_stripped", bazelGoosGoarch), command,
-		)
+	// This is a manual dereference of the alias created here:
+	//
+	// https://github.com/kubernetes/kubernetes/blob/b56d0acaf5bead31bd17d3b88d4a167fcbac7866/build/go.bzl#L45
+	kubeletPath, err := findGoBinary("//cmd/kubelet:_kubelet-cgo")
```
mikedanese, line 89: **Question for reviewers:** How is this binary used? Is it ok to always pick the linux build target or should do a bit of extra work to deference the //cmd/kubelet alias?

BenTheElder:
>Output paths are not a stable API 

That's a bit ridiculous. The bazel output directories are documented. In any sane build system I would expect to be able to obtain built binaries without having to construct some query and parse the output :/

This is another one for https://github.com/kubernetes/kubernetes/issues/88553 @dims @thockin @justaugustus 

> and all our conditional path construction here is fragile. 

I also take issue with this. We're only conditionally checking against one breakage ever when upstream introduced limited cross compilation in 1.14+, which changed the binary names. There's nothing fragile about the existing logic, bazel's output is fragile.

BenTheElder:
/lgtm
/approve

I appreciate filing a fix for https://github.com/kubernetes/kubernetes/pull/94449, but I think this highlights why we have near zero adoption of this build outside of CI. 

A consumer should be able to predict something like `bazel-bin/cmd/kubelet` without needing to construct a command and parse json etc.

mikedanese:
> That's a bit ridiculous.

Agree, this one had me scratching my head...

justaugustus:
I agree that bazel makes me feel ridiculous on a regular basis.
Opened https://github.com/kubernetes/test-infra/pull/19166 to add 1.19 kind tests and drop the tests for 1.16 and 1.15 branches (which are now EOL).

justaugustus:
@BenTheElder -- you may want/need to override the `pull-kind-conformance-parallel-1-15` job if it continues to fail.

mikedanese:
1.15 is still on bazel 0.23 it looks like and aquery didn't support json output. I modified this to use proto instead of json since it seemed the easiest but if you don't like this, we could all:

1. munge the info out of textproto
2. less ideally, stop supporting 1.15... what's the support policy for kind?

justaugustus:
> I agree that bazel makes me feel ridiculous on a regular basis.
> Opened [kubernetes/test-infra#19166](https://github.com/kubernetes/test-infra/pull/19166) to add 1.19 kind tests and drop the tests for 1.16 and 1.15 branches (which are now EOL).

@mikedanese -- Already opened a PR to drop those tests.

justaugustus:
> what's the support policy for kind?

Probably(/hopefully) the same as Kubernetes, so dev (1.20), 1.19, 1.18, and 1.17.

mikedanese:
Ok, then I'll drop the last commit and we can skip 1.15.

BenTheElder:
KIND currently supports 1.13+ (kubeadm beta versions) but we only need bazel in Kubernetes CI.

I'm fine with dropping bazel-only support from anything Kubernetes isn't running CI for.

BenTheElder:
https://github.com/kubernetes/test-infra/pull/19168 disables bazel in that job.
I'm OK bypassing broken CI to merge this though.

BenTheElder:
we're also hitting pod pending timeouts on prow w/ running the jobs :/
/test pull-kind-conformance-parallel-1-16
(prow pending timeout)

mikedanese:
For posterity since GH UI lost it, the dropped commit that made this work for 1.15 is here: https://github.com/kubernetes-sigs/kind/commit/29491d74d125af1175d2d8c0386e4d4c14a47d48

BenTheElder:
bypassing e2e flakes and ensuring this one goes in now.

Issue #1834, opened 2020-09-05
Customise Docker volume


mausch:
It would be great if users could customise the docker volume created by kind on cluster creation.

Currently the volume is anonymous so it's hard to identify it in `docker volume ls`.
Also because it grows big with all the images and stuff I'd like to put it on a different host partition instead of the default /var/lib/docker/volumes .

I guess the implementation for this would go in https://github.com/kubernetes-sigs/kind/blob/4d5b73e9158bb9e09711fd4d6065c38fd46709fd/pkg/cluster/internal/providers/podman/util.go#L76 ?
(or just write a different function)

BenTheElder:
All docker volumes can grow arbitrarily large, is there a reason to not move all of them?

What particular volumes we create is an implementation detail subject to change at any time.
Making them configurable would prevent us from changing that.

mausch:
Hi @BenTheElder , thanks for the quick reply!
Like many people, my desktop machines have several disks/partitions with different properties (speed, capacity, filesystem, etc)
I know exactly what containers I'm running and which volumes will _actually_ grow large and what they contain so I will schedule them accordingly on my disks.
For example, I have a machine with a primary 256GB SSD and a secondary 2TB HDD. At work I have a Postgres database that won't fit on the SSD so I put that volume on the HDD. I also run Elasticsearch but I know that the data I index there will fit on the SSD so I leave it there.
As for kind, its volume on the machine I'm testing it is 6GB just from infrastructure images, I haven't even started deploying my own applications (some of which are pretty large unfortunately, one in particular is 5GB!).

BenTheElder:
I don't think docker provides a mechanism to place an anonymous volume on a different disk.

If there is a mechanism that I'm not aware of to place _anonymous_ volumes on another disk, we could support an enviornment variable to point kind to a different anonymous volume location.

I don't think we should switch to named volumes for these as that guarantees users will expect to depend on them, and this is a highly internal detail that can and has changed in order to keep kind functioning well.

mausch:
> I don't think docker provides a mechanism to place an anonymous volume on a different disk.

This works for me:

`docker volume create --opt type=none --opt device=/path/to/some/other/disk --opt o=bind`

> If there is a mechanism that I'm not aware of to place anonymous volumes on another disk, we could support an enviornment variable to point kind to a different anonymous volume location.

That sounds great! :+1: 

[...]

PR #1843, opened 2020-09-10
Custom volume options

mausch:
#1834

### DIFF
--- a/pkg/cluster/internal/providers/docker/provision.go
+++ b/pkg/cluster/internal/providers/docker/provision.go
@@ -236,7 +236,7 @@ func runArgsForNode(node *config.Node, clusterIPFamily config.ClusterIPFamily, n
 		// filesystem, which is not only better for performance, but allows
 		// running kind in kind for "party tricks"
 		// (please don't depend on doing this though!)
-		"--volume", "/var",
+		"--mount", "type=volume,target=/var",
 		// some k8s things want to read /lib/modules
 		"--volume", "/lib/modules:/lib/modules:ro",
 	},
--- a/pkg/cluster/internal/providers/docker/provision.go
+++ b/pkg/cluster/internal/providers/docker/provision.go
@@ -19,6 +19,7 @@ package docker
 import (
 	"fmt"
 	"net"
+	"os"
 	"path/filepath"
 	"strings"
 
@@ -26,6 +27,7 @@ import (
 	"sigs.k8s.io/kind/pkg/errors"
 	"sigs.k8s.io/kind/pkg/exec"
 	"sigs.k8s.io/kind/pkg/fs"
+	"sigs.k8s.io/kind/pkg/log"
 
 	"sigs.k8s.io/kind/pkg/cluster/internal/loadbalancer"
 	"sigs.k8s.io/kind/pkg/cluster/internal/providers/common"
@@ -33,7 +35,7 @@ import (
 )
 
 // planCreation creates a slice of funcs that will create the containers
-func planCreation(cfg *config.Cluster, networkName string) (createContainerFuncs []func() error, err error) {
+func planCreation(logger log.Logger, cfg *config.Cluster, networkName string) (createContainerFuncs []func() error, err error) {
 	// we need to know all the names for NO_PROXY
 	// compute the names first before any actual node details
 	nodeNamer := common.MakeNodeNamer(cfg.Name)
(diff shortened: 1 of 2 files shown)


BenTheElder:
/hold
We're definitely not binding ourselves to the docker CLI, in the future we may use the client library.

Also as I said these volumes are an internal detail subject to change.

mausch:
>  We're definitely not binding ourselves to the docker CLI, in the future we may use the client library.

This would be a problem if the client library didn't support these options, right? But that doesn't seem to be the case I don't think? [ContainerCreate](https://godoc.org/github.com/docker/docker/client#Client.ContainerCreate) can be passed a [HostConfig](https://godoc.org/github.com/docker/docker/api/types/container#HostConfig) which has a [Mounts](https://godoc.org/github.com/docker/docker/api/types/mount#Mount) property which has the necessary [VolumeOptions](https://godoc.org/github.com/docker/docker/api/types/mount#VolumeOptions) 

> Also as I said these volumes are an internal detail subject to change.

Well, this is explicitly an unsupported feature with a big warning... meaning you could just remove this feature altogether if there is some actual reason to do so :slightly_smiling_face:

pkg/cluster/internal/providers/docker/provision.go
```diff
@@ -213,7 +215,13 @@ func commonArgs(cluster string, cfg *config.Cluster, networkName string, nodeNam
 	return args, nil
 }
 
-func runArgsForNode(node *config.Node, clusterIPFamily config.ClusterIPFamily, name string, args []string) ([]string, error) {
+func runArgsForNode(logger log.Logger, node *config.Node, clusterIPFamily config.ClusterIPFamily, name string, args []string) ([]string, error) {
+	varVolume := "type=volume,target=/var"
+	if n := os.Getenv("KIND_EXPERIMENTAL_DOCKER_VOLUME_OPTIONS"); n != "" {
```
BenTheElder, line 220: if you look in the recent history this was actually N volumes covering specific subsets of /var, we've temporarily reverted to one volume.
how is this handled if we switch back to N volumes ...?

mausch: Good question! I see that happened in https://github.com/kubernetes-sigs/kind/commit/8201cd9e69c2a4909afb5d2fa1c10e737acf7363 . 
If you had to switch back to N volumes you could just remove this environment variable altogether. I think it being explicitly unsupported gives you the right to do that at any moment and nobody has the right to complain about it ;-)
Tbh personally I'm mostly interested in /var/lib/containerd specifically, so if it were ever removed I'd maybe reimplement it (using a different env var name to avoid confusion!) and send another PR.

mausch:
3+ months later and the project detail that caused the main criticism to this PR ("volumes are an internal detail subject to change") has not been changed.
Anyway personally speaking I no longer have a need for this change. I'm lucky that I could afford some 1TB drives so disk space is no longer an issue for me. However this isn't the case for most people in the world, and I really think we should try to make every effort to accommodate for people with less resources.
Anyway it's pretty clear that you're against a change like this and like I said I no longer need it so feel free to close it.

aojea:
> Anyway personally speaking I no longer have a need for this change. I'm lucky that I could afford some 1TB drives so disk space is no longer an issue for me. However this isn't the case for most people in the world, and I really think we should try to make every effort to accommodate for people with less resources.
> Anyway it's pretty clear that you're against a change like this and like I said I no longer need it so feel free to close it.

I'm sad that you have that impression, is not common to need such amount of disk space, actually you were the first requesting it, lots of CIs are running thousands of KIND clusters daily in less that 14 Gb without any issues
https://docs.github.com/es/free-pro-team@latest/actions/reference/specifications-for-github-hosted-runners

/close

Ben explained the technical problems to maintain a solutions like yours in the long term, anyway, we hope to see you again
Thanks

mausch:
@aojea Hi Antonio, I think there's been a misunderstanding here, just wanted to clarify things. 
Firstly saying that I bought some 1TB drives doesn't mean that I actually need 1TB for KIND clusters specifically. But KIND is a tool for local development, and people use their computers for many different things not just KIND :slightly_smiling_face: 
So KIND competes for disk space with everything else on the machine e.g. databases, code repositories, build artifacts, and also non-development stuff.
In my case for example, a KIND volume having all the images I need is around 100GB. I don't think this is a particularly big cluster really. And for, say, a 250GB drive this is a nontrivial amount of disk space. Therefore the need to be flexible about where KIND puts its volume.

Anyway have a great 2021 :tada:

aojea:
lol, thanks for the clarification, I was totally getting it wrong :D
I have a similar issue, I have 2 partitions: one for the system and other for the `home` folder.
My solution is to move the `docker` data directory to the new partition using the `-g` option:
`ExecStart =/usr/bin/dockerd -g /new/docker/root/dir -H fd://`
Moving only `kind` data doesn't solve my problem

PR #1844, opened 2020-09-11
use kube::util::find-binary to find kubectl

mikedanese:
To centralize binary selection in k/k. Also drop the e2e.test copying
which is no longer needed according to the comment.


hack/ci/e2e-k8s.sh
```diff
@@ -71,12 +71,12 @@ build_with_bazel() {
...
-  # https://github.com/kubernetes/kubernetes/issues/68306
-  # TODO: remove this, it was fixed in 1.13+
-  mkdir -p '_output/bin/'
-  cp 'bazel-bin/test/e2e/e2e.test' '_output/bin/'
-  PATH="$(dirname "$(find "${PWD}/bazel-bin/" -name kubectl -type f)"):${PATH}"
+  local kubectl_path
+  # This is run from k/k root, so shellcheck shouldn't expect to be able to
+  # follow hack/lib/util.sh.
+  # shellcheck disable=SC1091
+  kubectl_path="$(source hack/lib/util.sh; kube::util::find-binary '.test')"
```
BenTheElder, line 78: we're going to be up a tree when we port this out of bash :|
https://github.com/kubernetes-sigs/kind/issues/986 https://github.com/kubernetes-sigs/kind/pull/1269

hack/ci/e2e-k8s.sh
```diff
@@ -71,12 +71,12 @@ build_with_bazel() {
   bazel shutdown || true
   pkill ^bazel || true
 
-  # ensure the e2e script will find our binaries ...
-  # https://github.com/kubernetes/kubernetes/issues/68306
-  # TODO: remove this, it was fixed in 1.13+
-  mkdir -p '_output/bin/'
-  cp 'bazel-bin/test/e2e/e2e.test' '_output/bin/'
-  PATH="$(dirname "$(find "${PWD}/bazel-bin/" -name kubectl -type f)"):${PATH}"
```
BenTheElder, line 79: why isn't this sufficient? most of the `kube::util::find-binary` logic is for the makefile output.
is there a non-executable file in bazel-bin named `kubectl` now?

hack/ci/e2e-k8s.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
BenTheElder, line 1: this was intentionally written in posix sh since the logic is simple.

mikedanese: Switched in order to source hack/lib/util.sh. Thought process: bash is fairly universal nowadays, used in most places in k/*, and all other shell scripts in this directory use bash. It can still be simple even if it's bash :)

hack/ci/e2e-k8s.sh
mikedanese: I suspect it would work fine to use find right now but I'm not very familiar with how we use the bazel cache and future changes (e.g. cross building in the same CI pool) might subtly cause issues. `kube::util::find-binary` is more complex but it does a reasonable job. I don't feel too strongly either way. Yet another option is to use cluster/kubectl.sh like other CI jobs.

mikedanese:
cc @justaugustus

hack/ci/e2e-k8s.sh
mikedanese: Pushed a fourth option that doesn't depend on kube::util::find-binary

hack/ci/e2e-k8s.sh
```diff
@@ -67,16 +67,16 @@ build_with_bazel() {
   # make sure we have e2e requirements
   bazel build //cmd/kubectl //test/e2e:e2e.test //vendor/github.com/onsi/ginkgo/ginkgo
 
+  kubectl_path="$(bazel aquery 'mnemonic("GoLink", //cmd/kubectl)' \
+      | grep '^  Outputs: \[.*\]$' \
+      | sed 's/^  Outputs: \[\(.*\)\]$/\1/')"
```
BenTheElder, line 72: neat ...

PR #1847, opened 2020-09-14
bump debian iptables base image

aojea:
new iptables 1.8.5 release
https://lwn.net/Articles/822157/

xref: https://github.com/kubernetes/kubernetes/pull/94733

### DIFF
--- a/images/kindnetd/Dockerfile
+++ b/images/kindnetd/Dockerfile
@@ -24,6 +24,6 @@ COPY . .
 RUN CGO_ENABLED=0 go build -o ./kindnetd ./cmd/kindnetd
 
 # build real kindnetd image
-FROM k8s.gcr.io/build-image/debian-iptables:v12.1.0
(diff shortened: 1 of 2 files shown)


aojea:
/hold

we ned https://github.com/kubernetes/kubernetes/pull/94733

BenTheElder:
I've been keeping an eye on this but intentionally not going to ship this in v0.9, no more suprises until that's out the door.
Almost ready to go. The remaining fix PR should be ready and I squared away most of the rest related to the release last night.

justaugustus:
https://github.com/kubernetes/kubernetes/pull/94733 has merged :)

aojea:
what happens with 1.15? it is timing out
/retest
last time we had a similar problem with an old version it turned out something was broken, did it deserve to be investigated or is a known issue?

BenTheElder:
> last time we had a similar problem with an old version it turned out something was broken, did it deserve to be investigated or is a known issue?

it's because we had to stop using bazel.

PR #1854, opened 2020-09-15
export logs kind version

### LINKED ISSUE #1850: export logs: add kind version

`export logs` adds a `kubernetes-version.txt` with the Kubernetes version. It could be nice to add a `kind-version.txt` with the version of kind.

**Why is this needed**:

Can be useful to debug in some scenarios. This is not really a major issue at all though, very low priority.

kensipe:
Resolves #1850 in a way that is consistent with existing code

### DIFF
--- a/pkg/cluster/internal/providers/common/logs.go
+++ b/pkg/cluster/internal/providers/common/logs.go
@@ -5,6 +5,7 @@ import (
 	"path/filepath"
 
 	"sigs.k8s.io/kind/pkg/cluster/nodes"
+	"sigs.k8s.io/kind/pkg/cmd/kind/version"
 	"sigs.k8s.io/kind/pkg/errors"
 	"sigs.k8s.io/kind/pkg/exec"
 )
@@ -22,8 +23,24 @@ func CollectLogs(n nodes.Node, dir string) error {
 			return cmd.SetStdout(f).SetStderr(f).Run()
 		}
 	}
+	writeToPathFn := func(s string, path string) func() error {
+		return func() error {
+			f, err := FileOnHost(path)
+			if err != nil {
+				return err
+			}
+			defer f.Close()
+			_, err = f.WriteString(s)
+			return err
+		}
+	}
+
 	return errors.AggregateConcurrent([]func() error{
 		// record info about the node container
+		writeToPathFn(
+			version.DisplayVersion(),
+			filepath.Join(dir, "kind-version.txt"),
+		),
 		execToPathFn(
 			n.Command("cat", "/kind/version"),
 			"kubernetes-version.txt",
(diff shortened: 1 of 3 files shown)


BenTheElder:
Thanks for the PR!

This should be implemented somewhere else, we want it to happen for all of the node providers (docker,podman,...)

kensipe:
@BenTheElder I took your todo comment as the place to perform the action.   does it make sense to move it to align with the creation of kubernetes-version.txt?     that was my initial intention before seeing the todo comment.

BenTheElder:
That seems like a good place to put it 👍 
I think this TODO has just been carried over since before there the provider abstraction.

BenTheElder:
we should drop the TODO line and see if it was copied into the podman code and drop it there too if so

kensipe:
@BenTheElder moved to common logs which seems better.   Somewhat dissatisfying to have a cmd import in common/logs.go specifically `"sigs.k8s.io/kind/pkg/cmd/kind/version"`.   let me know if this needs refactoring.  I might need some guidance on preferences here if this is desired.

One alternative was to call `kind` from `kind` which I thought was not preferred.   For clarity

```
		writeToPathFn(
			version.DisplayVersion(),
			filepath.Join(dir, "kind-version.txt"),
		),
```
could be replaced with:
```
		execToPathFn(
			n.Command("kind", "version"),
			"kind-version.txt",
		),
```
This has the advantage of removing the `writeToPathFn` and removed the cmd import... it seems to call out for a value which is known within the process.

BenTheElder:
There's no kind binary on the node to call, and even if there was we want to know the kind version on the host doing the log dumping.

This import is fine.
Please squash these commits and remove the identical TODO from the other (podman) location.

kensipe:
makes sense... you got it!

aojea:
it now logs the kind version inside the nodes,
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/1854/pull-kind-e2e-kubernetes-1-18/1306291584945360898/artifacts/logs/kind-worker/kind-version.txt

I would prefer to have it in the root path, same as docker.info, 
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/1854/pull-kind-e2e-kubernetes-1-18/1306291584945360898/artifacts/logs/docker-info.txt

but this seems to reduce the duplicate code, so

/lgtm

aojea:
both jobs timed out? 
/retest

BenTheElder:
@aojea we probably want to switch to smoke-test style jobs for older kubernetes releases in the future, still thinking about how best to do that. I asked @amwat to write a proposal for version support policy first.
/override pull-kind-conformance-parallel-1-15

BenTheElder:
/lgtm
/approve
thank you!

aojea:
> @aojea we probably want to switch to smoke-test style jobs for older kubernetes releases in the future, still thinking about how best to do that. I asked @amwat to write a proposal for version support policy first.
> /override pull-kind-conformance-parallel-1-15

what about the github actions?

kensipe:
thanks team!

PR #1855, opened 2020-09-16
add config command

yylt:
show config about default cluster

Signed-off-by: yylt <yang8518296@163.com>

### DIFF
--- a/pkg/cmd/kind/config/config.go
+++ b/pkg/cmd/kind/config/config.go
@@ -1,0 +1,38 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


aojea:
naive question, is this really useful?

yylt:
When add  more extra arg about kubernetes components, and I find it's too many parameters which have to set.

should print more args on kubernetes components, but not sure this pr needed, so it's just simple now

BenTheElder:
Hmm, I'm not sure if we want to add this, but if we did it would still be something like `kind [verb] [noun]` for consistency.
In the future it's generally preferable to open an issue to discuss feature additions first so we can discuss that sort of thing up front :sweat_smile:

BenTheElder:
Thanks for the PR but in it's current state we're not going to merge this.
We will need to identify at minimum what the command should be first.
If you'd like to pursue this further, please open an issue with a proposal.

PR #1856, opened 2020-09-16
Update ingress example to use networking.k8s.io

arangamani:
extensions api group is deprecated in newer versions of Kubernetes.


amwat:
/lgtm
/ok-to-test

Thanks!

BenTheElder:
/lgtm
/approve
thanks!

PR #1857, opened 2020-09-16
Updating Code based on Additional Linters

kensipe:
Slowing adding a couple of linters listed beflow, there are a number of lint issues to resolve.    Fixed in this PR:
1. some `goconst` around existing consts that were not used
2. `unconvert` for unnecessary conversations

history:   The current project configuration is such that a user with a `.golangci-lint.yml` in their home dir would override existing configurations.   Remaining on this PR are reasonable noncontroversial updates to code based on some additional linters (unknownly at the time) being applied.   There is another PR coming to change to a configuration which will eliminate this environmental issue.

### DIFF
--- a/pkg/apis/config/v1alpha4/default.go
+++ b/pkg/apis/config/v1alpha4/default.go
@@ -43,14 +43,14 @@ func SetDefaultsCluster(obj *Cluster) {
 	// and [::1]:randomPort on ipv6
 	if obj.Networking.APIServerAddress == "" {
 		obj.Networking.APIServerAddress = "127.0.0.1"
-		if obj.Networking.IPFamily == "ipv6" {
+		if obj.Networking.IPFamily == IPv6Family {
 			obj.Networking.APIServerAddress = "::1"
 		}
 	}
 	// default the pod CIDR
 	if obj.Networking.PodSubnet == "" {
 		obj.Networking.PodSubnet = "10.244.0.0/16"
-		if obj.Networking.IPFamily == "ipv6" {
+		if obj.Networking.IPFamily == IPv6Family {
 			obj.Networking.PodSubnet = "fd00:10:244::/64"
 		}
 	}
@@ -60,7 +60,7 @@ func SetDefaultsCluster(obj *Cluster) {
 	// we allocate a /16 subnet that allows 65535 services (current Kubernetes tested limit is O(10k) services)
 	if obj.Networking.ServiceSubnet == "" {
 		obj.Networking.ServiceSubnet = "10.96.0.0/16"
-		if obj.Networking.IPFamily == "ipv6" {
+		if obj.Networking.IPFamily == IPv6Family {
 			obj.Networking.ServiceSubnet = "fd00:10:96::/112"
 		}
 	}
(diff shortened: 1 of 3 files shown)


BenTheElder:
hmm, the linters should be running and passing in CI already?

pkg/build/nodeimage/build_impl.go
```diff
@@ -349,6 +329,32 @@ func (c *buildContext) prePullImages(bits kube.Bits, dir, containerID string) ([
 	return importer.ListImported()
 }
 
+func pullImageFns(requiredImages []string, builtImages sets.String, imagesDir string, pulledImages chan string, logger log.Logger) []func() error {
```
BenTheElder, line 332: I'm -1 on this, this loses a ton of context, this method takes way too many parameters.

this whole file does need refactoring, but I'd rather not just do enough to silence some linter.
we're going to continue restructuring build to solve some open issues as-is.

pkg/cluster/internal/kubeconfig/internal/kubeconfig/paths.go
```diff
@@ -40,7 +40,7 @@ https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands
 - If $KUBECONFIG environment variable is set, then it is used as a list of paths (normal path delimiting rules for your system). These paths are merged. When a value is modified, it is modified in the file that defines the stanza. When a value is created, it is created in the first file that exists. - If no files in the chain exist, then it creates the last file in the list.
 
 - Otherwise, ${HOME}/.kube/config is used and no merging takes place.
-*/
+*/ // nolint:lll
```
BenTheElder, line 43: this linter should probably just be disabled. we're not using 80 character-width hardware terminals in 2020.

pkg/cluster/internal/providers/docker/network.go
```diff
@@ -18,7 +18,7 @@ package docker
 
 import (
 	"bytes"
-	"crypto/sha1"
+	"crypto/sha1" //nolint:gosec // used as a seed to subnet generator
```
BenTheElder, line 21: I think we should probably just disable this linter wholesale, I'm not aware of it catching much useful and it complains about a lot of intentional, well-meant behavior

BenTheElder (commented):
we should verify that the verify job is doing it's job, lint passing is supposed to be enforced in presubmit.

I'm not a fan of linter-comments, but if we're going to leave linter comments we should leave an explanation of why, so in the future it can be removed if appropriate.
In all cases here I think we should disable the linter, they don't seem particularly useful.

pkg/build/nodeimage/build_impl.go
kensipe: yeah... I thought this might be controversial

I'll revert and mark ignore for this method

pkg/cluster/internal/kubeconfig/internal/kubeconfig/paths.go
kensipe: totally align with you there!  on it

pkg/cluster/internal/providers/docker/network.go
kensipe: roger that

kensipe:
@BenTheElder thanks for the guidance and context.  I updated the title and desc of this PR with current details.  Also squashed during rebase and ready.
I will be pulling together a PR specific to linter configuration in a yaml config file NOT in the root of the project.

PR #1859, opened 2020-09-17
Lint Configuration via YAML

kensipe:
Moving golangci-lint configuration to yaml vs sh which provides a number of benefits.
1. It stops dev env with user or global configurations from highjacking the configuration of the linters 
2. it provides a better / easier to maintain configurations with granular controls not available with shell execution

There are NO changes to the configured linters for this PR.
There is an upgrade to golangci-lint to the latest which requires no additional code updates.

### DIFF
--- a/hack/tools/.golangci.yml
+++ b/hack/tools/.golangci.yml
@@ -1,0 +1,27 @@
+run:
+  timeout: 2m
+
+linters:
+  disable-all: true
+  enable:
+    # default golangci-lint lints
+    - deadcode
+    - errcheck
+    - gosimple
+    - govet
+    - ineffassign
+    - staticcheck
+    - structcheck
+    - typecheck
+    - unused
+    - varcheck
+
+    # additional lints
+    - gochecknoinits
+    - gofmt
+    - golint
+    - misspell
+    - scopelint
+    - unparam
+
+linters-settings:
--- a/hack/tools/go.sum
+++ b/hack/tools/go.sum
@@ -86,6 +86,7 @@ github.com/go-toolsmith/strparse v1.0.0 h1:Vcw78DnpCAKlM20kSbAyO4mPfJn/lyYA4BJUD
 github.com/go-toolsmith/strparse v1.0.0/go.mod h1:YI2nUKP9YGZnL/L1/DLFBfixrcjslWct4wyljWhSRy8=
 github.com/go-toolsmith/typep v1.0.0 h1:zKymWyA1TRYvqYrYDrfEMZULyrhcnGY3x7LDKU2XQaA=
 github.com/go-toolsmith/typep v1.0.0/go.mod h1:JSQCQMUPdRlMZFswiq3TGpNp1GMktqkR2Ns5AIQkATU=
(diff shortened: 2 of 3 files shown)


hack/tools/.golangci.yml
```diff
@@ -0,0 +1,24 @@
...
+    - ineffassign
+    - misspell
+    - staticcheck
+    - structcheck
+    - typecheck
+    - unparam
+    - unused
+    - varcheck
+    - golint
+    - scopelint
```
neolit123, line 22: ideally the comments:
```
# default golangci-lint lints

...

# additional lints

...
```

from the shell configuration should be preserved in the yaml.

neolit123 (commented):
the change seems fine to me.
added one request about preserving the comments explaining which linters are default / additional.

hack/tools/go.mod
```diff
@@ -3,7 +3,7 @@ module sigs.k8s.io/kind/hack/tools
 go 1.13
 
 require (
-	github.com/golangci/golangci-lint v1.27.0
+	github.com/golangci/golangci-lint v1.31.0
```
neolit123, line 6: ideally such changes should be separate commits.

BenTheElder:
/approve
agree with @neolit123, going ahead and approving but leaving LGTM for the comments.

kensipe:
@neolit123 @BenTheElder thanks!  comments added back in... also reorganized order.  The original order of "default" linters was alpha (updated this list to that).  In the sh, the "additional" lints were not in any order I could make sense of.. this are now alpha as well.   Removed linter version bump and rebased.

neolit123 (commented):
/lgtm

hack/tools/go.sum
```diff
@@ -1,8 +1,25 @@
 cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
+cloud.google.com/go v0.34.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
aojea, line 2: this file should not be here, right?

hack/tools/go.sum
```diff
@@ -86,6 +86,7 @@ github.com/go-toolsmith/strparse v1.0.0 h1:Vcw78DnpCAKlM20kSbAyO4mPfJn/lyYA4BJUD
 github.com/go-toolsmith/strparse v1.0.0/go.mod h1:YI2nUKP9YGZnL/L1/DLFBfixrcjslWct4wyljWhSRy8=
 github.com/go-toolsmith/typep v1.0.0 h1:zKymWyA1TRYvqYrYDrfEMZULyrhcnGY3x7LDKU2XQaA=
 github.com/go-toolsmith/typep v1.0.0/go.mod h1:JSQCQMUPdRlMZFswiq3TGpNp1GMktqkR2Ns5AIQkATU=
+github.com/go-xmlfmt/xmlfmt v0.0.0-20191208150333-d5b6f63a941b h1:khEcpUM4yFcxg4/FHQWkvVRmgijNXRfzkIDHh23ggEo=
```
kensipe, line 89: these 2 go.sum changes are after a `go mod tidy`.   looks legit... but if you want these removed for the previous go.sum file, let me know.

kensipe:
good catch @aojea !   @neolit123 could use another lgtm after rebase... thanks!

PR #1884, opened 2020-10-03
Error if cluster to delete does not exist

whoan:
If we have a typo in the name of the cluster to delete, the current output is (the same as when we actually delete a cluster):

    $ kind delete cluster --name=anything
    Deleting cluster "anything" ...

After this change, we have this:

    $ kind delete cluster --name=anything
    Deleting cluster "anything" ...
    ERROR: failed to delete cluster "anything": Cluster not found

### DIFF
--- a/pkg/cluster/internal/kubeconfig/internal/kubeconfig/remove.go
+++ b/pkg/cluster/internal/kubeconfig/internal/kubeconfig/remove.go
@@ -48,6 +48,8 @@ func RemoveKIND(kindClusterName string, explicitPath string) error {
 				if err := write(existing, configPath); err != nil {
 					return err
 				}
+			} else {
+				return errors.New("Cluster not found")
 			}
 
 			return nil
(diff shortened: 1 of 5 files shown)


BenTheElder:
Thanks for the PR!
Please see past discussion of why this is the intentional behavior.

In short: clusters cannot exist or not exist. Nodes and kubeconfig entries do, and cleanup is idempotent.

whoan:
Got it @BenTheElder . Thanks for the clarification.
I think there are still some changes that you can pick from the commits I made in the branch.

Thanks for this useful tool!

pkg/cluster/internal/kubeconfig/internal/kubeconfig/remove.go
```diff
@@ -48,6 +48,8 @@ func RemoveKIND(kindClusterName string, explicitPath string) error {
 				if err := write(existing, configPath); err != nil {
 					return err
 				}
+			} else {
+				return errors.New("Cluster not found")
```
BenTheElder, line 52: the cluster is not necessarily expected to be found in all config files.

BenTheElder:
thanks :smiley:, one comment

aojea:
what is the status here?
should we close?

whoan:
@aojea I think so. Apparently, current behavior is intentional.

aojea:
then thanks for the PR @whoan 
/close

PR #1886, opened 2020-10-06
add gpus support in docker provider

cgeers:
Exposes a new external api node parameter, `gpus` that, when present, is passed through as the `--gpus` flag of the associated node.  Only the docker provider is supported currently, though podman support should be doable as well.

### DIFF
--- a/pkg/apis/config/v1alpha4/types.go
+++ b/pkg/apis/config/v1alpha4/types.go
@@ -142,6 +142,9 @@ type Node struct {
 	// The node-level patches will be applied after the cluster-level patches
 	// have been applied. (See Cluster.KubeadmConfigPatchesJSON6902)
 	KubeadmConfigPatchesJSON6902 []PatchJSON6902 `yaml:"kubeadmConfigPatchesJSON6902,omitempty"`
+
+	// ExtraArguments to be given to the provider
+	ExtraArguments []string `yaml:"extraArguments,omitempty"`
 }
 
 // NodeRole defines possible role for nodes in a Kubernetes cluster managed by `kind`
--- a/pkg/cluster/internal/providers/docker/provision.go
+++ b/pkg/cluster/internal/providers/docker/provision.go
@@ -251,6 +251,8 @@ func runArgsForNode(node *config.Node, clusterIPFamily config.ClusterIPFamily, n
 	}
 	args = append(args, mappingArgs...)
 
+	args = append(args, node.ExtraArguments...)
+
 	// finally, specify the image to run
 	return append(args, node.Image), nil
 }
--- a/pkg/cluster/internal/providers/podman/provision.go
+++ b/pkg/cluster/internal/providers/podman/provision.go
@@ -209,6 +209,8 @@ func runArgsForNode(node *config.Node, clusterIPFamily config.ClusterIPFamily, n
 	}
 	args = append(args, mappingArgs...)
 
+	args = append(args, node.ExtraArguments...)
+
 	// finally, specify the image to run
 	_, image := sanitizeImage(node.Image)
 	return append(args, image), nil
--- a/pkg/internal/apis/config/convert_v1alpha4.go
+++ b/pkg/internal/apis/config/convert_v1alpha4.go
@@ -55,6 +55,7 @@ func convertv1alpha4Node(in *v1alpha4.Node, out *Node) {
 	out.ExtraMounts = make([]Mount, len(in.ExtraMounts))
 	out.ExtraPortMappings = make([]PortMapping, len(in.ExtraPortMappings))
 	out.KubeadmConfigPatchesJSON6902 = make([]PatchJSON6902, len(in.KubeadmConfigPatchesJSON6902))
+	out.ExtraArguments = in.ExtraArguments
 
 	for i := range in.ExtraMounts {
 		convertv1alpha4Mount(&in.ExtraMounts[i], &out.ExtraMounts[i])
--- a/pkg/internal/apis/config/types.go
+++ b/pkg/internal/apis/config/types.go
@@ -104,6 +104,9 @@ type Node struct {
 	// KubeadmConfigPatchesJSON6902 are applied to the generated kubeadm config
 	// as patchesJson6902 to `kustomize build`
 	KubeadmConfigPatchesJSON6902 []PatchJSON6902
+
+	// ExtraArguments allow for custom options to be passed to the provider
+	ExtraArguments []string
 }
 
 // NodeRole defines possible role for nodes in a Kubernetes cluster managed by `kind`
--- a/site/content/docs/user/configuration.md
+++ b/site/content/docs/user/configuration.md
@@ -294,3 +294,22 @@ nodes:
 
 [YAML]: https://yaml.org/
 [feature gates]: https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/
+
+### Extra Arguments
+
+Extra arguments can optionally be used augment the default behavior of the container runtime that provisions cluster nodes.
+
+NOTE: Extra Arguments are specifc to the underlying container management system. They are *not* necessarily portable between container management systems (IE: docker vs podman), and may not be supported at all in future releases. Use this configuration option with care, and as a last resort only.
+
+For example, the following configuration is valid for docker, and will expose all GPUs devices present to the cluster node, as well as adjust the limit on open file handles.
+{{< codeFromInline lang="yaml" >}}
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+nodes:
+- role: control-plane
+  extraArguments:
+  - --gpus
+  - all
+  - --ulimit
+  - nofile=8192:8192
+{{< /codeFromInline >}}


pkg/cluster/internal/providers/docker/provision.go
```diff
@@ -251,6 +251,10 @@ func runArgsForNode(node *config.Node, clusterIPFamily config.ClusterIPFamily, n
 	}
 	args = append(args, mappingArgs...)
 
+	if node.Gpus != "" {
+		args = append(args, "--gpus", node.Gpus)
```
BenTheElder, line 255: what does this flag do vs passing in the device as an extraMount path, given the node already has full capabilities?

cgeers: This docker flag seems to be fairly poorly documented since its inception many months ago, so believe me I wish I had a better answer for you.

It appears that as of recent docker versions, forcing the use of nvidia's `nvidia-container-runtime` runc wrapper is no longer sufficient in itself to expose identified GPU devices to containers under that runtime. This flag appears to be recently required (as of 19.03.13?) in addition the nv runtime configuration in order to present the identified devices to containers under docker. As best as I can tell, exposing devices via mount alone does not appear to be a supported path to obtain the necessary NVML libs and utils.

cgeers:
I have signed the CLA as an individual.

BenTheElder:
I go around doing reviews regularly but the stack is pretty deep (just happened to be checking my inbox) https://gubernator.k8s.io/pr/BenTheElder has some approximation of the stack.

I will get to this one.

My quick note for now is: we prefer to follow the precedents set by CRI in terms of the API for talking to containers. I'm hesitant to add a config field that is passed directly to `docker`, in the future we may use the API client for example.
https://github.com/kubernetes/cri-api

In the future it would be helpful to file a feature request issue to discuss details before implementing.

BenTheElder:
So I did some digging last night. Couple notes:
- Kubernetes doesn't support --gpus and still uses the alternative oci runtime

- The actual API in kubernetes is largely the device plugin API
  - device managers can tell kubelet that devices exist
  - devices are requested by amount not by specific devices
  - libraries etc. being present isn't handled by any k8s API

The --gpu flag doesn't seem very well documented as you said and there's nothing like this in any of the other container tools (podman, cri-o, containerd, ....)

links:
https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/
https://docs.docker.com/config/containers/resource_constraints/#gpu

---

I think we want support for GPU passthrough, but I don't think there's good precedent for a portable abstraction so I'm not sure what it should look like.

Currently we don't have runtime specific options in config.
At minimum, the podman provider needs to thrown an error if this field is set.

More ideally I think we'd think about how to make this a portable abstraction that _could_ be implemented across providers fairly easily, or a good pattern for very clearly bundling non-portable options.

The former seems preferable but harder, I presume people GPUs with other container tools as well, a cursory glance seems to confirm this.

cgeers:
Thanks for the analysis @BenTheElder. I'm happy to implement any changes you deem appropriate to make this acceptable.

BenTheElder:
@aojea can you comment re: podman?

aojea:
> @aojea can you comment re: podman?

I spoke to one podman hacker @mheon, and seems that podman does not need the flag 

> Adding the GPU as a device should be enough. If you're using an nvidia card you need their OCI hook installed (I believe you also need a label to trigger said hook, but I'm not 100% on that)

aojea:
@cgeers I think that for podman we must pass the device id with the `--device` flag

BenTheElder:
xref: https://github.com/NVIDIA/gpu-operator/issues/7#issuecomment-737306014 we'll also want this to progress.

---

> Adding the GPU as a device should be enough. If you're using an nvidia card you need their OCI hook installed (I believe you also need a label to trigger said hook, but I'm not 100% on that)

What does that entail?

> @cgeers I think that for podman we must pass the device id with the --device flag , so same as in docker but replace args = append(args, "--gpus", node.Gpus) by

Presumably the flag values are not compatible though? So we'd want to consider coming up with some format that can be mapped to both.

aojea:
> Presumably the flag values are not compatible though? So we'd want to consider coming up with some format that can be mapped to both.

I´m guessing here that gpus field pass the device id, so in podman you must pass it as ` --device` , but I can be totally wrong

cgeers:
Thanks for the input @aojea. I will look into podman to confirm the behavior. 

If these providers end up requiring very different arguments and values, the idea the @BenTheElder floated regarding provider specific options sounds like the most forwards compatible path. I'll take a look and report back.

cgeers:
I've updated his PR with a more generic, albeit lower effort approach that can service the GPU exposure use case, as well as other use cases, such as adjusting cluster node ulimits. I've also updated documentation accordingly.

After discussion with team members, because this exposes configuration possibilities beyond those that the CRI targets, it opens up the possibility of users creating clusters configurations that are not portable between container management systems. I've added warnings to this effect in the documentation. 

I understand this is a tricky proposition, and as such it may take some time consider the benefits and drawbacks that this more flexible approach provides.

If this is deemed not acceptable, I'm happy to pursue both the GPU device exposure and ulimit settings as independent config options who's provisioner implementation handles accordingly, even if in doing so, kind must forever play catch up with new, different and missing container management system capabilities.

aojea:
this is similar to the kubeadm `extraArgs` parameter, I think that this offers flexibility to users to try different runtime options, and could be useful for us to get a taste on now features before adding them to the API.
Just one question, should it be a map[string]string instead of []string?

cgeers:
> this is similar to the kubeadm `extraArgs` parameter, I think that this offers flexibility to users to try different runtime options, and could be useful for us to get a taste on now features before adding them to the API.
> Just one question, should it be a map[string]string instead of []string?

Hrm. map[string]string might be problematic for supporting repeating arguments, and arguments without values.

pkg/apis/config/v1alpha4/types.go
```diff
@@ -142,6 +142,9 @@ type Node struct {
 	// The node-level patches will be applied after the cluster-level patches
 	// have been applied. (See Cluster.KubeadmConfigPatchesJSON6902)
 	KubeadmConfigPatchesJSON6902 []PatchJSON6902 `yaml:"kubeadmConfigPatchesJSON6902,omitempty"`
+
+	// ExtraArguments to be given to the provider
+	ExtraArguments []string `yaml:"extraArguments,omitempty"`
```
BenTheElder, line 147: I'm not in favor of supporting this, the fact that we use exec is an implementation detail.

BenTheElder:
We _may_ have to do provider specific options.
We still should not add "pass arbitrary flags through to exec", that's been previously rejected for good reason. Not only is that a support nightmare but it locks us into the implementation detail that we use exec currently.

pkg/apis/config/v1alpha4/types.go
aojea: > I'm not in favor of supporting this, the fact that we use exec is an implementation detail.

yeah, but is the only way, no? 
you can not do arbitrary patching ala kubeadm with the providers

cgeers:
> We _may_ have to do provider specific options.
> We still should not add "pass arbitrary flags through to exec", that's been previously rejected for good reason. Not only is that a support nightmare but it locks us into the implementation detail that we use exec currently.

Understood. I will look into an alternative approach.

pkg/apis/config/v1alpha4/types.go
BenTheElder: you don't need to, we can come up with a good way to enable GPUs.

I think the problem is: kubernetes has a pretty good model where you just request N GPUs and GPUs are uniform within a node. One option would be to only support passing through all GPUs to *one* node and just have an option for this.

BenTheElder:
This particular approach I do not want to merge: https://github.com/kubernetes-sigs/kind/pull/1886/files#r545434435

I will try to find time to follow up on the linked issue soon, it looks like there's at least movement from nvidia re: containerd support.

closing to clear backlog for #2120 as well, but we should definitely revisit this.

romeokienzler:
hey @BenTheElder @cgeers is somebody of you planning to make any progress on this soon?

BenTheElder:
I don't have time for this currently unfortunately. 
Please file a feature issue with a detailed design plan we can discuss if you are interested in taking this up.
https://kind.sigs.k8s.io/docs/contributing/getting-started

pkg/apis/config/v1alpha4/types.go
romeokienzler: dear @BenTheElder - before I start working on a proposal - can you please confirm that I understood your concerns correctly? is your concern that passing the --gpus option is specific to docker and doesn't work on portman or that you can't control the amount of GPUs assigned to the kind node (e.g. other applications or other kind clusters won't get access to them anymore)

klueska:
Are there any plans to eventually allow passing a list of `devices` to inject from the host?

If so, that is a path towards eventual GPU support through CDI: https://github.com/docker/cli/issues/3864

We are in the processing of extending docker with CDI support (as presented in the issue above) right now.

PR #1896, opened 2020-10-16
site: update the registry docs to use the config map

nicks:
I still think it's worth having a native way to do this in Kind (as per https://github.com/kubernetes-sigs/kind/issues/1543 and https://github.com/kubernetes-sigs/kind/issues/1213) but for now I'm just updating the docs to match current best practices


BenTheElder:
sorry for the delay and thank you!

nicks:
Yay thanks @BenTheElder !

Issue #1899, opened 2020-10-19
APIServer Model


stevensu1977:
**What would you like to be added**:
I need create multiple k8s environment ,so I add "server" subcommand to kind , I think may be there have other one have same requirement like me .
if you use "kind server" you can get full REST API server 

POST|DELETE   /api/v1/cluster/{name}
GET /api/v1/clusters

**Why is this needed**:
create multiple k8s environment for some meetup  .

BenTheElder:
You can implement this server by importing the kind packages or wrapping the kind CLI yourself.

A cluster management rest service is pretty majorly out of scope for kind and would have some serious security issues (be careful allowing user specified node images!)

stevensu1977:
Hi @BenTheElder  I got your point ,   I think kind is very useful tools for meetup/education ,  if kind can be run as  daemon, client model both , that's be very cool !  I can use kind access other remote kind server create temp k8s cluster for meetup .
may be my idea is already out of scope for kind,  :<

BenTheElder:
you can support this without it being in kind AND without forking.

you can create a binary called `kind-server` that imports `sigs.k8s.io/kind/pkg/...` the same as the upstream CLI.

I don't think this is something we want to support upstream right now. besides security concerns it's not really a goal for us to overlap further with e.g. CAPI.

for prior art with provider wrappers importing the library packages see: https://github.com/kyma-incubator/terraform-provider-kind

thanks anyhow! in the future we can revisit this.

PR #1900, opened 2020-10-19
#1899 APIServer Model 

stevensu1977:
I need create multiple k8s environment ,so I add "server" subcommand to kind , I think may be there have other one have same requirement like me .
if you use "kind server" you can get full REST API server

POST|DELETE /api/v1/cluster/{name}
GET /api/v1/clusters


neolit123:
@stevensu1977
ideally this should have been first discussed in a kind issue ticket or a Google doc proposal, instead of sending the PR first.
nevertheless an interesting idea.

BenTheElder:
thanks for the PR, https://github.com/kubernetes-sigs/kind/issues/1899#issuecomment-717069865

PR #1903, opened 2020-10-20
default ipv6 subnet to /56

aojea:
we were defaulting IPv6 subnet to a /64, this was working because
kubeadm was calculating and modifying directly the node-cidr-mask,
so the controller manager does not complain.

Kubeadm should only validate that the config is correct and fail
in case of error, this behavior will be changed soon, but once
this happen it will break kind default deployment.

### DIFF
--- a/pkg/apis/config/v1alpha4/default.go
+++ b/pkg/apis/config/v1alpha4/default.go
@@ -51,7 +51,9 @@ func SetDefaultsCluster(obj *Cluster) {
 	if obj.Networking.PodSubnet == "" {
 		obj.Networking.PodSubnet = "10.244.0.0/16"
 		if obj.Networking.IPFamily == "ipv6" {
-			obj.Networking.PodSubnet = "fd00:10:244::/64"
+			// node-mask cidr default is /64 so we need a larger subnet, we use /56 following best practices
+			// xref: https://www.ripe.net/publications/docs/ripe-690#4--size-of-end-user-prefix-assignment---48---56-or-something-else-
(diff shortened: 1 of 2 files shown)


BenTheElder:
/approve
fun fun :+)

aojea:
> 2020/10/20 16:21:34 Image pull was unsuccessful: unable to save remote image k8s.gcr.io/build-image/debian-iptables@sha256:e30919918299988b318f0208e7fd264dee21a6be9d74bbd9f7fc15e78eade9b4: unable to write image layers: unable to write image layer: unable to write the contents of layer 1 to /bazel-scratch/.cache/bazel/_bazel_root/cae228f2a89ef5ee47c2085e441a3561/external/debian-iptables-amd64/image/001.tar.gz: read tcp 10.60.181.111:51148->108.177.112.128:443: read: connection reset by peer

is not the first time I see this

neolit123:
> xref: kubernetes/kubernetes#95723

maybe we should hold until that PR is merged, its a breaking change for ipv6 users as far as i can tell, so i'd appreciate more feedback on the PR.

/hold
?

aojea:
> maybe we should hold until that PR is merged, its a breaking change for ipv6 users as far as i can tell, so i'd appreciate more feedback on the PR.
> 
> /hold

is the other way around, we can not merge the other without this, and this is not breaking change since we only change the default ipv6 subnet, that has not any external dependencies

neolit123:
ok, feel free to remove hold.

aojea:
Thanks Lubomir
/hold cancel

PR #1904, opened 2020-10-20
WIP: Modify hack/ci/e2e-k8s.sh to support audit logging

hh:
This is to get feedback and test a version of e2e-k8s.sh that supports creating ARTIFACTS/audit/audit.log without any filters.

See: https://github.com/kubernetes/test-infra/issues/19613#issuecomment-713091006

### DIFF
--- a/hack/ci/e2e-k8s.sh
+++ b/hack/ci/e2e-k8s.sh
@@ -171,7 +171,7 @@ kubeadmConfigPatches:
     extraArgs:
       "v": "${KIND_CLUSTER_LOG_LEVEL}"
       "audit-policy-file": "/etc/kubernetes/pki/audit/policy.yaml"
-      "audit-log-path": "/etc/kubernetes/pki/audit.log"
+      "audit-log-path": "/etc/kubernetes/pki/audit/audit.log"
   controllerManager:
     extraArgs:
       "v": "${KIND_CLUSTER_LOG_LEVEL}"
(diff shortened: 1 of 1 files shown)


hack/ci/e2e-k8s.sh
```diff
@@ -176,12 +194,15 @@ EOF
...
   # TODO(BenTheElder): settle on verbosity for this script
   KIND_CREATE_ATTEMPTED=true
+  # ./audit/ is be relative to where kind is run
+  pushd ${ARTIFACTS}
   kind create cluster \
     --image=kindest/node:latest \
     --retain \
     --wait=1m \
     -v=3 \
     "--config=${ARTIFACTS}/kind-config.yaml"
```
aojea, line 204: then you should remove ${ARTIFACTS} here

hack/ci/e2e-k8s.sh
```diff
@@ -142,6 +154,10 @@ networking:
   kubeProxyMode: ${KUBE_PROXY_MODE:-iptables}
 nodes:
 - role: control-plane
+  extraMounts:
+  - hostPath: audit/
```
aojea, line 158: I still think that you should keep at is and use the relative path here appending the artifacts folder
```suggestion
  - hostPath: ${ARTIFACTS}/audit/
```

hack/ci/e2e-k8s.sh
aojea: Ups sorry, didn't realize this is the full path

hack/ci/e2e-k8s.sh
aojea: Scratch it, my fault

aojea:
@hh the apiserver fails to start with this error
., You know why it fails to mount it.?

2020-10-21T11:47:25.307808917Z stderr F Error: loading audit policy file: failed to read file path "/var/audit/policy.yaml": open /var/audit/policy.yaml: no such file or directory

hh:
I'm going to close and let this drop. We opted for new jobs that enable full audit logging.

PR #1907, opened 2020-10-22
Upgrade Ingress resource definition from v1beta1 to v1

electrocucaracha:
The Kubernetes [Ingress API went GA in v1.19](https://opensource.googleblog.com/2020/09/kubernetes-ingress-goes-ga.html). Using **v1beta1** version on v1.19+ Kubernetes clusters raises the following message: `Warning: networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress`. This change updates the syntax of that resource definition.

### DIFF
--- a/site/static/examples/ingress/usage.yaml
+++ b/site/static/examples/ingress/usage.yaml
@@ -46,7 +46,7 @@ spec:
   # Default port used by the image
   - port: 5678
 ---
-apiVersion: networking.k8s.io/v1beta1
+apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   name: example-ingress
@@ -55,11 +55,17 @@ spec:
   - http:
       paths:
       - path: /foo
+        pathType: Prefix
         backend:
-          serviceName: foo-service
-          servicePort: 5678
+          service:
+            name: foo-service
+            port:
+              number: 5678
       - path: /bar
+        pathType: Prefix
         backend:
-          serviceName: bar-service
-          servicePort: 5678
+          service:
+            name: bar-service
+            port:
+              number: 5678
 ---


neolit123 (commented):
/lgtm
as per https://kubernetes.io/docs/concepts/services-networking/ingress/#the-ingress-resource

amwat:
PR looks good thanks!
Although, currently we don't have separate versioned docs and we probably want them to be applicable to most of the versions kind supports (which is back till k8s 1.14 currently)
We are currently planning on having a well defined version policy (cc @BenTheElder docs is probably another thing to consider)

For now can you add a separate section for 1.19+ ?

BenTheElder:
I think we should go ahead and switch since the default version is 1.19, but it might be worth following up with an example for the older API and a note about when to use that.

cc @aledbf I'm curious how you think this should be approached

aledbf:
Just in case this change breaks the Ingress example if someone uses k8s < v1.19.

aojea:
I prefer a warning that a broken deployment :)
I will wait until 1.19 is EOL 

```
Warning: networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+
```

BenTheElder:
we're going to be migrating the default branch and additional stale PRs are problematic for that.
we can reopen this later if we think we should do it.

1.19 still won't be EOL for another 4 months at least https://github.com/kubernetes-sigs/kind/pull/1907#issuecomment-716485697

BenTheElder:
this got picked up again in https://github.com/kubernetes-sigs/kind/pull/2306, and is being re-discussed there, FYI.

PR #1920, opened 2020-11-09
issue 1905/support for multiple images to kind load

### LINKED ISSUE #1905: "kind load docker-image" supports loading multiple images at once
`kind load docker-image` supports loading multiple images at once.
Usage: `kind load docker-image [OPTIONS] IMAGE [IMAGE...]`

**Why is this needed**:
When I create the test cluster using `kind`, I have to import a lot of custom images. It will be much easier if batch operations can be supported.

yashvardhan-kukreja:
closes #1905 

**Usage:**

Multiple image names provided (all images present locally):
```
./kind load docker-image nginx,busybox --name master    

```
Multiple image names provided (all images are not present locally):
```
./kind load docker-image nginx,busybox,python --name master
ERROR: image: "python" not present locally
```
One image name provided without commas **(backwards compatibility)**:
```
./kind load docker-image nginx --name master               

```

### DIFF
--- a/pkg/cmd/kind/load/docker-image/docker-image.go
+++ b/pkg/cmd/kind/load/docker-image/docker-image.go
@@ -48,13 +48,13 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 	cmd := &cobra.Command{
 		Args: func(cmd *cobra.Command, args []string) error {
 			if len(args) < 1 {
-				return errors.New("name of image is required")
+				return errors.New("a list of image names is required")
 			}
 			return nil
 		},
-		Use:   "docker-image <IMAGE>",
-		Short: "Loads docker image from host into nodes",
-		Long:  "Loads docker image from host into all or specified nodes by name",
+		Use:   "docker-image <IMAGE> [IMAGE...]",
+		Short: "Loads docker images from host into nodes",
+		Long:  "Loads docker images from host into all or specified nodes by name",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cli.OverrideDefaultName(cmd.Flags())
 			return runE(logger, flags, args)
@@ -82,10 +82,14 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
 	)
 
 	// Check that the image exists locally and gets its ID, if not return error
-	imageName := args[0]
-	imageID, err := imageID(imageName)
-	if err != nil {
-		return fmt.Errorf("image: %q not present locally", imageName)
+	imageNames := args
+	var imageIDs []string
+	for _, imageName := range imageNames {
+		imageID, err := imageID(imageName)
+		if err != nil {
+			return fmt.Errorf("image: %q not present locally", imageName)
+		}
+		imageIDs = append(imageIDs, imageID)
 	}
 
 	// Check if the cluster nodes exist
@@ -121,37 +125,40 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
 
 	// pick only the nodes that don't have the image
 	selectedNodes := []nodes.Node{}
-	for _, node := range candidateNodes {
-		id, err := nodeutils.ImageID(node, imageName)
-		if err != nil || id != imageID {
-			selectedNodes = append(selectedNodes, node)
-			logger.V(0).Infof("Image: %q with ID %q not yet present on node %q, loading...", imageName, imageID, node.String())
+	fns := []func() error{}
+	for i, imageName := range imageNames {
+		imageID := imageIDs[i]
+		for _, node := range candidateNodes {
+			id, err := nodeutils.ImageID(node, imageName)
+			if err != nil || id != imageID {
+				selectedNodes = append(selectedNodes, node)
+				logger.V(0).Infof("Image: %q with ID %q not yet present on node %q, loading...", imageName, imageID, node.String())
+			}
+		}
+		if len(selectedNodes) == 0 {
+			logger.V(0).Infof("Image: %q with ID %q found to be already present on all nodes.", imageName, imageID)
+			continue
 		}
 	}
 
-	if len(selectedNodes) == 0 {
-		return nil
-	}
-
-	// Save the image into a tar
-	dir, err := fs.TempDir("", "image-tar")
+	// Setup the tar path where the images will be saved
+	dir, err := fs.TempDir("", "images-tar")
 	if err != nil {
 		return errors.Wrap(err, "failed to create tempdir")
 	}
 	defer os.RemoveAll(dir)
-	imageTarPath := filepath.Join(dir, "image.tar")
-
-	err = save(imageName, imageTarPath)
+	imagesTarPath := filepath.Join(dir, "images.tar")
+	// Save the images into a tar
+	err = save(imageNames, imagesTarPath)
 	if err != nil {
 		return err
 	}
 
-	// Load the image on the selected nodes
-	fns := []func() error{}
+	// Load the images on the selected nodes
 	for _, selectedNode := range selectedNodes {
 		selectedNode := selectedNode // capture loop variable
 		fns = append(fns, func() error {
-			return loadImage(imageTarPath, selectedNode)
+			return loadImage(imagesTarPath, selectedNode)
 		})
 	}
 	return errors.UntilErrorConcurrent(fns)
@@ -169,9 +176,10 @@ func loadImage(imageTarName string, node nodes.Node) error {
 	return nodeutils.LoadImageArchive(node, f)
 }
 
-// save saves image to dest, as in `docker save`
-func save(image, dest string) error {
-	return exec.Command("docker", "save", "-o", dest, image).Run()
+// save saves images to dest, as in `docker save`
+func save(images []string, dest string) error {
+	commandArgs := append([]string{"save", "-o", dest}, images...)
+	return exec.Command("docker", commandArgs...).Run()
 }
 
 // imageID return the Id of the container image
--- a/site/content/docs/user/quick-start.md
+++ b/site/content/docs/user/quick-start.md
@@ -151,11 +151,12 @@ context name `kind` and delete that cluster.
 ## Loading an Image Into Your Cluster
 
 Docker images can be loaded into your cluster nodes with:
-`kind load docker-image my-custom-image`
 
-> **NOTE**: If using a named cluster you will need to specify the name of the 
-cluster you wish to load the image into:
-`kind load docker-image my-custom-image --name kind-2`
+`kind load docker-image my-custom-image-0 my-custom-image-1`
+
+**Note**: If using a named cluster you will need to specify the name of the 
+cluster you wish to load the images into:
+`kind load docker-image my-custom-image-0 my-custom-image-1 --name kind-2`
 
 Additionally, image archives can be loaded with:
 `kind load image-archive /my-image-archive.tar`


go.sum
```diff
@@ -205,8 +205,6 @@ golang.org/x/sys v0.0.0-20191005200804-aed5e4c7ecf9/go.mod h1:h1NjWce9XRLGQEsW7w
 golang.org/x/sys v0.0.0-20200116001909-b77594299b42/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
 golang.org/x/sys v0.0.0-20200323222414-85ca7c5b95cd/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
 golang.org/x/sys v0.0.0-20200622214017-ed371f2e16b4/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
-golang.org/x/sys v0.0.0-20200923182605-d9f96fdee20d h1:L/IKR6COd7ubZrs2oTnTi73IhgqJ71c9s80WsQnh0Es=
-golang.org/x/sys v0.0.0-20200923182605-d9f96fdee20d/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
```
neolit123, line 209: `go mod tidy` should be part of CI in general, so that similar updates only happen when dependencies change.

pkg/cmd/kind/load/docker-image/docker-image.go
```diff
@@ -48,13 +49,13 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 	cmd := &cobra.Command{
 		Args: func(cmd *cobra.Command, args []string) error {
 			if len(args) < 1 {
-				return errors.New("name of image is required")
+				return errors.New("name of image or comma-separated names of images required")
```
neolit123, line 52:
```suggestion
				return errors.New("a comma-separated list of images required")
```

go.sum
yashvardhan-kukreja: alright. reverting this change.

pkg/cmd/kind/load/docker-image/docker-image.go
```diff
@@ -121,38 +126,40 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
...
-		fns = append(fns, func() error {
-			return loadImage(imageTarPath, selectedNode)
-		})
+		// Load the image on the selected nodes
+		for _, selectedNode := range selectedNodes {
+			selectedNode := selectedNode // capture loop variable
+			fns = append(fns, func() error {
+				return loadImage(imageTarPath, selectedNode)
+			})
+		}
```
neolit123, line 162: potentially, this code can be made to run in parallel via UntilErrorConcurrent:
https://github.com/kubernetes-sigs/kind/blob/a2282e90b85656f44eb4ee82a25dbd78eb7b1a6c/pkg/errors/concurrent.go#L25

site/content/docs/user/quick-start.md
```diff
@@ -140,14 +140,17 @@ kind delete cluster
 If the flag `--name` is not specified, kind will use the default cluster
 context name `kind` and delete that cluster.
 
-## Loading an Image Into Your Cluster
+## Loading Images Into Your Cluster
+
+Docker images can be loaded into your cluster nodes by specifying their names as comma-separated:
```
neolit123, line 145:
```suggestion
Docker images can be loaded into your cluster nodes by specifying their names as a comma-separated list:
```

pkg/cmd/kind/load/docker-image/docker-image.go
yashvardhan-kukreja: alright. So, just to confirm, should I add the snippet of code performing the `// Save the image into a tar`, a part of all the functions.

So, the new code would look like:

```go
		// Load the image on the selected nodes
		for _, selectedNode := range selectedNodes {
			selectedNode := selectedNode // capture loop variable
			fns = append(fns, func() error {
				// Save the image into a tar
				dir, err := fs.TempDir("", "image-tar")
				if err != nil {
					return errors.Wrap(err, "failed to create tempdir")
				}
				defer os.RemoveAll(dir)
				imageTarPath := filepath.Join(dir, "image.tar")

				err = save(imageName, imageTarPath)
				if err != nil {
					return err
				}
				return loadImage(imageTarPath, selectedNode)
			})
		}
```
And then, at the end `return errors.UntilErrorConcurrent(fns)` will run all the `fns` (involving saving and loading) in parallel

Is that what you mean?

go.sum
yashvardhan-kukreja: I also reverted the changes for `hack/tools/go.sum` and `images/kindnetd/go.sum`

pkg/cmd/kind/load/docker-image/docker-image.go
yashvardhan-kukreja: alright. I made the changes and brought the code performing the "saving of image into tar" just along the code where loading of images is happening. (Just above `return loadImage(imageTarPath, selectedNode)`)

pkg/cmd/kind/load/docker-image/docker-image.go
```diff
@@ -121,38 +126,40 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
...
+		for _, selectedNode := range selectedNodes {
+			selectedNode := selectedNode // capture loop variable
+			imageName := imageName       // capture variable (for lint checks)
+			fns = append(fns, func() error {
+				// Save the image into a tar
+				dir, err := fs.TempDir("", "image-tar")
+				if err != nil {
+					return errors.Wrap(err, "failed to create tempdir")
+				}
+				defer os.RemoveAll(dir)
```
neolit123, line 153: i think we can re-use the same temp dir for all images.

pkg/cmd/kind/load/docker-image/docker-image.go
neolit123: saving + loading in the same `fn` "should" work as long as the docker engine is fine with that.

pkg/cmd/kind/load/docker-image/docker-image.go
yashvardhan-kukreja: yes yes.. just noticed that few seconds ago. Was fixing that only.

neolit123 (commented):
/lgtm
/assign @BenTheElder

yashvardhan-kukreja:
> please, keep the commits squashed to one.
> 
> /ok-to-test

squashed... the lgtm label got removed. Don't know if that's ok or not.

neolit123:
> squashed... the lgtm label got removed. Don't know if that's ok or not.

it's fine. pending review from @BenTheElder

pkg/cmd/kind/load/docker-image/docker-image.go
```diff
@@ -121,38 +126,41 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
...
+		imageID := imageIDs[i]
+		for _, node := range candidateNodes {
+			id, err := nodeutils.ImageID(node, imageName)
+			if err != nil || id != imageID {
+				selectedNodes = append(selectedNodes, node)
+				logger.V(0).Infof("Image: %q with ID %q not yet present on node %q, loading...", imageName, imageID, node.String())
+			}
+		}
+		if len(selectedNodes) == 0 {
+			return nil
```
aojea, line 140: I think you should `continue` not return here, one image can be present in all nodes, but the next image maybe not

aojea: i.e kind load `docker-image busybox,alpine,httpd`, if busybox is present in all nodes you return and alpine and httpd are not loaded

yashvardhan-kukreja: oh right, nice catch. That's true. My bad. Missed to test on a multiple node-setup

yashvardhan-kukreja: made the changes.

BenTheElder:
why would it be a single comma delimited argument? I don't think that was the ask in #1905 and we haven't had a chance to discuss yet.

yashvardhan-kukreja:
> why would it be a single comma delimited argument? I don't think that was the ask in #1905 and we haven't had a chance to discuss yet.

>> **What would you like to be added:**
>> kind load docker-image supports loading multiple images at once.
>> Usage: kind load docker-image [OPTIONS] IMAGE [IMAGE...]

@BenTheElder , @SataQiu seems to have asked for a support to load multiple images in one go a `kind load docker-image` command instead of needing to execute the kind load docker-image N number of times for loading N images.

And comma-delimited argument seemed reliable to me in the sense that image names/tags don't allow to include commas so, the user providing the multiple images as a single string of image names delimited by commas seemed simple to use and implement.

yashvardhan-kukreja:
> > why would it be a single comma delimited argument? I don't think that was the ask in #1905 and we haven't had a chance to discuss yet.
> 
> > > **What would you like to be added:**
> > > kind load docker-image supports loading multiple images at once.
> > > Usage: kind load docker-image [OPTIONS] IMAGE [IMAGE...]
> > > **Why is this needed:**
> > > When I create the test cluster using kind, I have to import a lot of custom images. It will be much easier if batch operations can be supported.
> 
> @BenTheElder , @SataQiu seems to have asked for a support to load multiple images in one go of a `kind load docker-image` command instead of needing to execute the `kind load docker-image` command N number of times for loading N images.
> 
> And comma-delimited argument seemed reliable to me in the sense that image names/tags don't allow to include commas so, the user providing the multiple images as a single string of image names delimited by commas seemed simple to use and implement, and reliable coz the commas would never conflict with the image name/tag
> 
> Also, the saving/loading of all the images (when multiple images are provided) will be done parallelly (concurrently) via the `errors.UntilErrorConcurrent(fns)`. So, the batch-operation based execution seems to going to happen efficiently too.

OR @BenTheElder , @aojea , @neolit123 , @SataQiu  do you folks think that I should migrate the command to work for a space-separated list of image name ( `kind load docker-image image-1 image-2 image-3` ) instead of comma-separated string of image names ( `kind load docker-image image-1,image-2,image-3` )

It is an extremely small change, so I implement in a matter of seconds. 
So, whatever you folks suggest 😄

jaredallard:
@yashvardhan-kukreja Couldn't the names be space delimited? I wouldn't expect input to be split by commas unless... spaces somehow were needed? Even then, I'd use quotes if I needed to pass spaces, I don't think we'd want to be doing the system argument processors' job by trying to split on `,` unless needed.

yashvardhan-kukreja:
> @yashvardhan-kukreja Couldn't the names be space delimited? I wouldn't expect input to be split by commas unless... spaces somehow were needed? Even then, I'd use quotes if I needed to pass spaces, I don't think we'd want to be doing the system argument processors' job by trying to split on `,` unless needed.

haha.. just a sec ago, I commented that too. Sure, it's an extremely small change on my side. I can do that too.

yashvardhan-kukreja:
> > @yashvardhan-kukreja Couldn't the names be space delimited? I wouldn't expect input to be split by commas unless... spaces somehow were needed? Even then, I'd use quotes if I needed to pass spaces, I don't think we'd want to be doing the system argument processors' job by trying to split on `,` unless needed.
> 
> haha.. just a sec ago, I commented that too. Sure, it's an extremely small change on my side. I can do that too.

@jaredallard , @BenTheElder , @aojea , @neolit123 , in the above commit, I shifted to using space-separated list of images.
`kind load docker-image image1 image2 image3`

If you still feel that this is not satisfactory, I am up for any other suggestions and git resets :)

neolit123:
I like separating by comma more. This is consistent with k8s core
components e.g. for feature gates. Defering to Ben, thought.

yashvardhan-kukreja:
> I like separating by comma more. This is consistent with k8s core
> components e.g. for feature gates. Defering to Ben, thought.
> 

alright, let's see what @BenTheElder says

pkg/cmd/kind/load/docker-image/docker-image.go
```diff
@@ -121,38 +125,42 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
...
-
-	// Save the image into a tar
-	dir, err := fs.TempDir("", "image-tar")
-	if err != nil {
-		return errors.Wrap(err, "failed to create tempdir")
-	}
-	defer os.RemoveAll(dir)
-	imageTarPath := filepath.Join(dir, "image.tar")
-
-	err = save(imageName, imageTarPath)
```
BenTheElder, line 144: docker save supports multiple images and dedupes layers.
we should be saving once still.

yashvardhan-kukreja: ohh nice. Will do that.

BenTheElder:
> I like separating by comma more. This is consistent with k8s core
components e.g. for feature gates. Defering to Ben, thought.

I like comma over "spaces", but ... I like actually using separate arguments better (as the original feature request usage shows).

I was fishing for arguments as to why we wouldn't do that.

To be extra clear: spaces between strings in a shell typically means they're split into separate shell words I.E. arguments. So if we call it from another program it is like `exec.Command("kind", "load", "docker-image", "image1", "image2")` not `exec.Command("kind", "load", "docker-image", "image1 image2")`.

We should do _that_, not:
```shell
kind load docker-image 'image1 image2'
```

unless someone has a compelling argument.

yashvardhan-kukreja:
> > I like separating by comma more. This is consistent with k8s core
> > components e.g. for feature gates. Defering to Ben, thought.
> 
> I like comma over "spaces", but ... I like actually using separate arguments better (as the original feature request usage shows).
> 
> I was fishing for arguments as to why we wouldn't do that.
> 
> To be extra clear: spaces between strings in a shell typically means they're split into separate shell words I.E. arguments. So if we call it from another program it is like `exec.Command("kind", "load", "docker-image", "image1", "image2")` not `exec.Command("kind", "load", "docker-image", "image1 image2")`.
> 
> We should do _that_ (`kind load docker-image image1 image2`), not:
> 
> ```shell
> kind load docker-image 'image1 image2'
> ```
> 
> unless someone has a compelling argument.
> 
> EDIT: and if someone does have a compelling argument, spaces would probably be confused all the time by users in shells etc. so commas would be better, besides consistency arguments.

I am also in favor of commas. I believe that in terms of user experience (for eg. while entering it as a command) won't be an issue coz it's just commas instead of spaces.

Now, from dev's experience, if commas are being used, then, the dev can anyways code like this `exec.Command("kind", "load", "docker-image", "image1,imag2,image3,image4")`

So, I don't think that would be an issue too.

yashvardhan-kukreja:
So, @BenTheElder , @aojea , @neolit123 , I am in favor of commas over spaces. What do you guys think?

aojea:
> EDIT: and if someone does have a compelling argument, spaces would probably be confused all the time by users in shells etc. so commas would be better, besides consistency arguments.

commas, but is a personal taste, I can not backup it :smile:

yashvardhan-kukreja:
> > EDIT: and if someone does have a compelling argument, spaces would probably be confused all the time by users in shells etc. so commas would be better, besides consistency arguments.
> 
> commas, but is a personal taste, I can not backup it

I guess we can go with commas then, coz all of us have unanimously favored for commas only.

yashvardhan-kukreja:
> > > EDIT: and if someone does have a compelling argument, spaces would probably be confused all the time by users in shells etc. so commas would be better, besides consistency arguments.
> > 
> > 
> > commas, but is a personal taste, I can not backup it
> 
> I guess we can go with commas then, coz all of us have unanimously favored for commas only.

I can do something of an overkill too (might be useful) where the user can either enter as a space-separated way or comma-separated way.

For example, if the command sees multiple arguments, then, space-separated way would be considered but if the comma sees a single argument, then, kind will considered the list of images to be comma-separated.

So, all of the below will work fine then, 
`kind load docker-image image1`
`kind load docker-image image1 image2 image3`
`kind load docker-image image1,image2,image3`

sounds, may be, like an overkill but thought of mentioning this.

yashvardhan-kukreja:
FYI, my latest commit to this PR corresponds to using comma-separated image name.
If that's fine, then, please let me know. I'll squash my commits. @BenTheElder , @neolit123 , @aojea .

jaredallard:
I'm still incredibly confused as to why commas would ever be used when there is no reason to do so, since now you're still... acting like the shell and having to split the argv essentially. I mean, ultimately here it doesn't matter as much (since commas can't be used in image names), but when it comes to repeated arguments to a process... I don't know why we're using commas when spaces can never be used in an image name, this seems like non-standard behaviour (compared to other CLI tools) for not really any reason?

BenTheElder:
To be clear I did not vote for commas over splitting args, as @jaredallard said, and I don't see a compelling argument for forcing all images into a single argument.

BenTheElder:
Requiring spaces within a single argument would be fairly awkward (versus say, commas which is _less_ awkward but only a bit), but I don't see why anyone would do that to begin with. Kubernetes uses commas within _flag_ values that cannot be multi-argument.

I think there's some conflation of "how I type things into the shell" and "what the process receives" going on here... The fact that shells are typically space separated is irrelevant, users know how to use their shell and multiple arguments, I've not heard a single reason this should be done with a single argument.

Perhaps also some confusion around "backwards compatibility"? Supplying a single image is compatible between old and new kind with any of these options, but supplying multiple images is going to require using the newer version of kind _regardless_ ...

aojea:
> > EDIT: and if someone does have a compelling argument, spaces would probably be confused all the time by users in shells etc. so commas would be better, besides consistency arguments.
> 
> commas, but is a personal taste, I can not backup it

discard my previous comment about commas, I commented with a limited understanding of the topic, but last 3 comments explain it perfectly

yashvardhan-kukreja:
got it guys. Thanks. Updating the code accordingly.

yashvardhan-kukreja:
Squashed the commits with the new changes.
Multiple arguments supported instead of single comma-separared arguments.

Tested it the multiple arguments manually as well as with "xargs" (just to play around with). Working perfectly fine as requested.

pkg/cmd/kind/load/docker-image/docker-image.go
```diff
@@ -48,13 +49,13 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
...
 			if len(args) < 1 {
-				return errors.New("name of image is required")
+				return errors.New("a single image name or a space-separated list of image names is required")
 			}
 			return nil
 		},
-		Use:   "docker-image <IMAGE>",
-		Short: "Loads docker image from host into nodes",
-		Long:  "Loads docker image from host into all or specified nodes by name",
+		Use:   "docker-image <IMAGE1 IMAGE2 IMAGE3>",
```
jaredallard, line 56: nit: this could be more standard by doing this instead:


```suggestion
		Use:   "docker-image <IMAGE> [IMAGE...],
```

jaredallard: Not sure of the standards on this project but `<ARG>` generally denotes requires and `[ARG]` being optional, with `...` being multiple. We could probably shorten this to `<IMAGE...>` but I haven't seen that before and am not sure 😆

pkg/cmd/kind/load/docker-image/docker-image.go
```diff
@@ -169,9 +177,10 @@ func loadImage(imageTarName string, node nodes.Node) error {
 	return nodeutils.LoadImageArchive(node, f)
 }
 
-// save saves image to dest, as in `docker save`
-func save(image, dest string) error {
-	return exec.Command("docker", "save", "-o", dest, image).Run()
+// save saves images to dest, as in `docker save`
+func save(images []string, dest string) error {
+	commandArgs := strings.Split(fmt.Sprintf("save -o %s %s", dest, strings.Join(images, " ")), " ")
```
jaredallard, line 182: nit: this seems a little convoluted

```suggestion
	commandArgs := []string{"save", "-o", dest, strings.Join(images, " ")}
```

pkg/cmd/kind/load/docker-image/docker-image.go
yashvardhan-kukreja: aah cool cool. Got  it.

neolit123:
> kind load docker-image image1 image2

oh, if this is about separate arguments then spaces are the way to go.
if these are in a single flag then commas are preferred:

> kind load docker-image --images="image1,image2"

yashvardhan-kukreja:
> > kind load docker-image image1 image2
> 
> oh, if this is about separate arguments then spaces are the way to go.
> if these are in a single flag then commas are preferred:
> 
> > kind load docker-image --images="image1,image2"

yes. And in the above conversation and in the issue's description too, the former approach of separate args was suggested and asked for. :)

yashvardhan-kukreja:
> > > kind load docker-image image1 image2
> > 
> > 
> > oh, if this is about separate arguments then spaces are the way to go.
> > if these are in a single flag then commas are preferred:
> > > kind load docker-image --images="image1,image2"
> 
> yes. And in the above conversation and in the issue's description too, the former approach of separate args was suggested and asked for. :)

And that's the change I made in the latest commit. So, it's working as expected :)

yashvardhan-kukreja:
@neolit123 @aojea , can u please review this?

yashvardhan-kukreja:
@neolit123 @aojea , sorry for pestering but can u please review this? 🙈

neolit123 (commented):
LGTM

yashvardhan-kukreja:
Please have a look at this PR @BenTheElder . It has already received one LGTM status.

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/1926#issuecomment-744751196

jaredallard:
Also ignore this, if we're demoing prow features, this is always a must

/meow

pkg/cmd/kind/load/docker-image/docker-image.go
```diff
@@ -48,13 +48,13 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 	cmd := &cobra.Command{
 		Args: func(cmd *cobra.Command, args []string) error {
 			if len(args) < 1 {
-				return errors.New("name of image is required")
+				return errors.New("a single image name or a space-separated list of image names is required")
```
BenTheElder, line 51: this is still inaccurate, the spaces are a detail of the shell.

site/content/docs/user/quick-start.md
```diff
@@ -148,14 +148,15 @@ kind delete cluster
 If the flag `--name` is not specified, kind will use the default cluster
 context name `kind` and delete that cluster.
 
-## Loading an Image Into Your Cluster
+## Loading Images Into Your Cluster
```
BenTheElder, line 151: This will break inbound heading links. The title change doesn't seem worth it in that light.

BenTheElder (commented):
(See comment on your other PR) And thank you!

two nits we can address in follow-up given this lengthy delay 😞

pkg/cmd/kind/load/docker-image/docker-image.go
yashvardhan-kukreja: Ohh, so I suppose `a list of image names is required` message would be good. `a single image name` part feels redundant in my opinion but I'd love to hear your opinion.

site/content/docs/user/quick-start.md
yashvardhan-kukreja: Ahh, good point. Although I just found that there is only one inbound heading link which I guess we can simply change to point to `Loading images into your cluster` instead.
In my opinion, I guess we can go forwards with changing that inbound link as its a tiny change anyways but let me know what do you think :)

PS: [under this heading only is that one inbound heading link](https://github.com/kubernetes-sigs/kind/blob/master/site/content/docs/user/private-registries.md#pull-to-the-host-and-side-load)

yashvardhan-kukreja:
@BenTheElder  I have the changes ready as per my above comments. If these sound good, let me know, I ll push them under this PR : )
Thanks!

pkg/cmd/kind/load/docker-image/docker-image.go
BenTheElder: that SGTM, I think a list is fine, given that one image is just a list of only one image 👍

site/content/docs/user/quick-start.md
BenTheElder: > Ahh, good point. Although I just found that there is only one inbound heading link which I guess we can simply change to point to Loading images into your cluster instead.

Er my concern is not the links within the site, but all the places we've handed out this link to users.

On most pages I'd agree, but this page has too much content (before we started making individual guides) and has been around a while, I'm certain there are many more links to this floating around out there 😅

yashvardhan-kukreja: Coolz, sounds legit. I ll revert this one.

BenTheElder:
test flakes are known / unrelated, don't worry about that.

site/content/docs/user/quick-start.md
```diff
@@ -150,12 +150,13 @@ context name `kind` and delete that cluster.
 
 ## Loading an Image Into Your Cluster
 
-Docker images can be loaded into your cluster nodes with:
-`kind load docker-image my-custom-image`
+Docker images can be loaded into your cluster nodes by specifying their names as a space-separated list:
```
BenTheElder, line 153: the "space-seperate list" bit should also be dropped here
```suggestion
Docker images can be loaded into your cluster nodes with:
```
the example and `--help` text should be sufficient to define the format (and again, it may not be spaces, it's actually controlled by `$IFS` in most shells, and in a calling program it will be an argument array)

yashvardhan-kukreja: Aah, forgot to update this one :see_no_evil

PR #1923, opened 2020-11-12
bump go to 1.15.5

BenTheElder:


aojea:
/lgtm
what happen? it was close to the previous one

BenTheElder:
security release, but as a general rule kind just keeps up with latest where possible regardless.

aojea:
was lazy to go to the release notes :)

BenTheElder:
/retest
TLDR some nasty security bugs that mostly shouldn't apply to our usage _currently_, but all the same upgrading.

PR #1925, opened 2020-11-13
Fix zfs

BenTheElder:
let's try this one more time
fixes https://github.com/kubernetes-sigs/kind/issues/1891
(well it will with a new node image, which I'll add once CI passes)

### DIFF
--- a/images/base/files/etc/containerd/config.toml
+++ b/images/base/files/etc/containerd/config.toml
@@ -18,3 +18,5 @@ version = 2
   # allow hugepages controller to be missing
   # see https://github.com/containerd/cri/pull/1501
   tolerate_missing_hugepages_controller = true
+  # explicitly use default snapshotter so we can sed it in entrypoint
+  snapshotter = "overlayfs"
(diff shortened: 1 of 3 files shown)


images/base/files/etc/containerd/config.toml
```diff
@@ -17,4 +17,6 @@ version = 2
   sandbox_image = "k8s.gcr.io/pause:3.3"
   # allow hugepages controller to be missing
   # see https://github.com/containerd/cri/pull/1501
-  tolerate_missing_hugepages_controller = true
\ No newline at end of file
+  tolerate_missing_hugepages_controller = true
+  # explicitly use default snapshotter so we can sed it in entrypoint
+  snapshotter = "overlayfs"
```
aojea, line 22: it says that depends on the platform, is this something we worry about?
https://godoc.org/github.com/containerd/containerd#pkg-constants

aojea: nevermind, this will be always linux

PR #1931, opened 2020-11-18
split unit and integration tests in makefile

BenTheElder:
we're going to need to switch up the CI jobs, but this should be a better end state.

cc @aojea

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,7 +70,13 @@ install: build
 # ================================= Testing ====================================
 # unit tests (hermetic)
 unit:
-	hack/make-rules/unit.sh
+	MODE=unit hack/make-rules/test.sh
+# integration tests
+integration:
+	MODE=integration hack/make-rules/test.sh
+# all tests
+test:
+	hack/make-rules/test.sh
 ################################################################################
 # ================================= Cleanup ====================================
 # standard cleanup target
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -70,7 +70,13 @@ install: build
 # ================================= Testing ====================================
 # unit tests (hermetic)
 unit:
```
BenTheElder, line 72: ci runs this target
we want to rename the job and run the test target.

pkg/cluster/internal/providers/docker/network_integration_test.go
```diff
@@ -0,0 +1,81 @@
+// +build !nointegration
```
BenTheElder, line 1: This allows us to run unit tests only without the annoying skip messages. we could drop the skip mechanism entirely, but it's not super onerous for now.

hack/make-rules/test.sh
```diff
@@ -13,30 +13,49 @@
...
 go build -o "${REPO_ROOT}/bin/gotestsum" gotest.tools/gotestsum
 cd "${REPO_ROOT}"
 
+go_test_opts=(
+  "-coverprofile=${REPO_ROOT}/bin/${MODE}.cov"
+  '-covermode' 'count'
+  '-coverpkg' 'sigs.k8s.io/kind/...'
+)
+if [[ "${MODE}" = 'unit' ]]; then
+  go_test_opts+=('-run' '^TestIntegration')
```
aojea, line 38: Are they flipped?

BenTheElder: yes

PR #1727, opened 2020-07-13
Support Rootless Docker

AkihiroSuda:
This PR adds support for running kind with Rootless Docker provider.
Requires cgroup v2 hosts.

### 1. "podman: unlock rootless"
Turn off `"podman provider does not work properly in rootless mode"` error and print a warning message instead.
However, Podman provider still doesn't work . (see the bottom of this PR)

### 2. "base: ignore EACCES from `mount -o remount,ro /sys`"
`mount -o remount,ro /sys` fails with `permission denied` on rootless Docker and on rootless Podman, but the error is negligible.
### 3. "containerd: add /etc/containerd/config-rootless.toml"
`/etc/containerd/config-rootless.toml` is the config for running kind in rootless Docker/Podman.

 * `ociwrapper` script is used to remove `.linux.resources.devices` from `config.json`,  because `.linux.resources.devices` is meaningless on rootless and yet produces errors. Workaround until we get proper fixes in containerd and runc.
 * `restrict_oom_score_adj` is set to true to ignore oom_score_adj errors

The entrypoint overrides `/etc/containerd/config.toml` with `config-rootless.toml` when running in rootless Docker/Podman.
The rootless-ness is detected by comparing `/proc/1/uid_map` with `0 0 4294967295`.

#### Base

```console
$ docker build -t kind-base ./images/base
```

Available on Docker Hub as `	akihirosuda/tmp-kind-base:g554d2e07`.
Built from https://github.com/AkihiroSuda/kind/commits/554d2e076b1ea0fb55fcdff5cf8d972933bb78df .

#### Node

Needs PR https://github.com/kubernetes/kubernetes/pull/93012 and PR https://github.com/kubernetes/kubernetes/pull/92863. 
The `kubelet: new feature gate: Rootless` commit is not necessary for `kind`, because Rootless Docker itself sets up cgroup fs.

```console
$ kind build node-image --base-image kind-base
```

Available on Docker Hub as `akihirosuda/tmp-kind-node:g554d2e07-g3c1dda52`
Built from https://github.com/AkihiroSuda/kind/commits/554d2e076b1ea0fb55fcdff5cf8d972933bb78df  +  https://github.com/AkihiroSuda/kubernetes/commits/3c1dda52bb3a931acb4810e34fbfa1afee949ec5

> **NOTE**
> 
> Rootful Docker is still required for `kind build node-image`.

### Rootless Docker

* Boot Ubuntu 20.04 host with `systemd.unified_cgroup_hierarchy=1`
[...]

PR #1935, opened 2020-11-18
Support Rootless Docker and Rootless Podman, without patching Kubernetes

AkihiroSuda:
This PR adds support for running kind with Rootless Docker provider.
Requires Docker 20.10 / Podman 3.0, with cgroup v2.

Unlike the previous PR (https://github.com/kubernetes-sigs/kind/pull/1727), this version works without patching Kubernetes (1.20.4).

However, this version has dirty hacks such as faking sysctl keys by bind-mounting regular files under `/proc/sys`.
So I still want the Kubernetes PR to be merged: https://github.com/kubernetes/kubernetes/pull/92863

## Restrictions

The restrictions of Rootless Docker apply to kind clusters as well.

e.g.
- OverlayFS cannot be used unless the host is Ubuntu or Debian
- Cannot mount block storages
- Cannot mount NFS

To workaround the OverlayFS issue, we could use fuse-overlayfs on kernel >= 4.18: https://github.com/AkihiroSuda/containerd-fuse-overlayfs
However, to decrease complexity of PR, the support for fuse-overlayfs is not included in this PR, and will be introduced in a separate PR after this PR gets merged.

## How this PR works
When the entrypoint script detect that it is running inside user namespace (i.e. rootless), it modifies `/etc/containerd/config.toml` to:

- Set `restrict_oom_score_adj` to true to adjust oomScoreAdj value
- Change the snapshotter from overlay to native, unless running with Ubuntu/Debian kernel

The entrypoint script also does:
- Bind-mount several fake regular files inside `/proc/sys` to make them writable. Workaround until https://github.com/kubernetes/kubernetes/pull/92863 gets merged.
- Set `kubeProxyConfiguration.conntrack.maxPerCore` to 0, for avoiding an error during setting sysctl value `net.netfilter.nf_conntrack_max`

### Step 1: Prepare host
- Install Ubuntu 20.10 host. 

- Add `GRUB_CMDLINE_LINUX="systemd.unified_cgroup_hierarchy=1"` to `/etc/default/grub`.

- Create `/etc/systemd/system/user@.service.d/delegate.conf` with the following content:
```
[Service]
Delegate=yes
```

- Run `sudo update-grub` and reboot.

- Install Docker 20.10 or Podman 3.0.
```console
$ curl -fsSL https://get.docker.com | sh
$ sudo apt-get install -y docker-ce-rootless-extras
$ dockerd-rootless-setuptool.sh install -f
```

- Run `docker info` with `DOCKER_HOST=unix://$XDG_RUNTIME_DIR/docker.sock`, and make sure it shows "rootless" as a Security Option:
```console
$ export DOCKER_HOST=unix://$XDG_RUNTIME_DIR/docker.sock

$ docker info
...
Cgroup Driver: systemd
 Cgroup Version: 2
...
 Security Options:
  seccomp
   Profile: default
  rootless
  cgroupns
...
```

### Step 2: Prepare the node image

```console
$ (cd $GOPATH/src/k8s.io/kubernetes && git checkout v1.20.4)
```

```console
export DOCKER_HOST=unix://$XDG_RUNTIME_DIR/docker.sock
docker build -t kindest/base:latest ./images/base && kind build node-image --base-image kindest/base:latest --type=bazel
```

The image is registered as `kindest/node:latest` in Rootless Docker's image store.

NOTE: `--type=bazel` is required for running `kind build node-image` with Rootless Docker.

### Step 3: Start kind with Rootless Docker
- Run `kind create cluster --image kindest/node:latest`
- Run `ps auxw` on the hosts, and make sure the kind processes are running as unprivileged users
- Make sure `kubectl get pods -A` shows all pods as `Running`

```console
$ export DOCKER_HOST=unix://$XDG_RUNTIME_DIR/docker.sock
$ kind create cluster --image kindest/node:latest
...
$ kubectl get pods -A
NAMESPACE            NAME                                         READY   STATUS    RESTARTS   AGE
kube-system          coredns-74ff55c5b-cgdfs                      1/1     Running   0          113s
kube-system          coredns-74ff55c5b-hpldh                      1/1     Running   0          113s
kube-system          etcd-kind-control-plane                      1/1     Running   0          114s
kube-system          kindnet-qqm9j                                1/1     Running   0          113s
kube-system          kube-apiserver-kind-control-plane            1/1     Running   0          114s
kube-system          kube-controller-manager-kind-control-plane   1/1     Running   0          114s
kube-system          kube-proxy-qx849                             1/1     Running   0          113s
kube-system          kube-scheduler-kind-control-plane            1/1     Running   0          114s
local-path-storage   local-path-provisioner-78776bfc44-68n5f      1/1     Running   0          113s
```

- - -

Pre-built image is temporarily available on my Docker Hub: https://hub.docker.com/r/akihirosuda/tmp-kind-node/tags

### DIFF
--- a/images/base/files/etc/containerd/config.toml
+++ b/images/base/files/etc/containerd/config.toml
@@ -20,3 +20,5 @@ version = 2
   tolerate_missing_hugepages_controller = true
   # explicitly use default snapshotter so we can sed it in entrypoint
   snapshotter = "overlayfs"
+  # restrict_oom_score_adj needs to be true when running inside UserNS (rootless)
+  restrict_oom_score_adj = false
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -18,10 +18,96 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+# If /proc/self/uid_map 4294967295 mappings, we are in the initial user namespace, i.e. the host.
+# Otherwise we are in a non-initial user namespace.
+# https://github.com/opencontainers/runc/blob/v1.0.0-rc92/libcontainer/system/linux.go#L109-L118
+userns=""
+if grep -Eqv "0[[:space:]]+0[[:space:]]+4294967295" /proc/self/uid_map; then
+  userns="1"
+  echo 'INFO: running in a user namespace (experimental)'
+fi
+
+validate_userns() {
+  if [[ -z "${userns}" ]]; then
+    return
+  fi
+
+  local nofile_hard
+  nofile_hard="$(ulimit -Hn)"
+  local nofile_hard_expected="64000"
+  if [[ "${nofile_hard}" -lt "${nofile_hard_expected}" ]]; then
+    echo "WARN: UserNS: expected RLIMIT_NOFILE to be at least ${nofile_hard_expected}, got ${nofile_hard}" >&2
+  fi
+
+  if [[ ! -f "/sys/fs/cgroup/cgroup.controllers" ]]; then
+    echo "ERROR: UserNS: cgroup v2 needs to be enabled" >&2
+    exit 1
+  fi
+  for f in cpu memory pids; do
+    if ! grep -qw $f /sys/fs/cgroup/cgroup.controllers; then
+      echo "ERROR: UserNS: $f controller needs to be delegated" >&2
+    exit 1
+    fi
+  done
+}
+
+fake_file_with_content(){
+  local path="$1"
+  local content="$2"
+  local base="/run/fake"
+  local fake_path="${base}/${path}"
+  mkdir -p "$(dirname "${fake_path}")"
+  echo "INFO: UserNS: faking ${path} to be \"${content}\" (writable)"
+  echo "${content}" > "${fake_path}"
+  mount --bind "${fake_path}" "${path}"
+}
+
+fake_sysctl() {
+  local key="$1"
+  local key_slash
+  # shellcheck disable=SC2001
+  key_slash="$(echo "${key}" | sed -e s@\\.@/@g)"
+  local path="/proc/sys/${key_slash}"
+  if [[ -f "${path}" ]]; then
+    local content
+    content="$(cat "${path}")"
+    fake_file_with_content "${path}" "${content}"
+  fi
+}
+
 configure_containerd() {
   # we need to switch to the 'native' snapshotter on zfs
   if [[ "$(stat -f -c %T /kind)" == 'zfs' ]]; then
     sed -i 's/snapshotter = "overlayfs"/snapshotter = "native"/' /etc/containerd/config.toml
+  fi
+
+  # userns (rootless) configs
+  if [[ -n "$userns" ]]; then
+    # Adjust oomScoreAdj
+    sed -i 's/restrict_oom_score_adj = false/restrict_oom_score_adj = true/' /etc/containerd/config.toml
+
+    # mounting overlayfs inside userns requires patching kernel.
+    # Ubuntu kernel is patched by default.
+    # Debian kernel is patched by default as well, but Debian needs `sudo modprobe overlay permit_mounts_in_userns=1`.
+    local tmp
+    tmp=$(mktemp -d)
+    mkdir -p "${tmp}"/{l,u,w,m}
+    if mount -t overlay overlay -o "lowerdir=${tmp}/l,upperdir=${tmp}/u,workdir=${tmp}/w" "${tmp}/m"; then
+      umount "${tmp}/m"
+    else
+      echo 'INFO: UserNS: this kernel does not support mounting overlayfs inside userns. Disabling overlayfs'
+      sed -i 's/snapshotter = "overlayfs"/snapshotter = "native"/' /etc/containerd/config.toml
+    fi
+    rm -rf "${tmp}"
+
+    # To run vanilla kubelet inside UserNS, we need to fake several unwritable sysctl to be writable.
+    # Workaround until https://github.com/kubernetes/kubernetes/pull/92863 gets merged in the upstream.
+    fake_sysctl "vm.overcommit_memory"
+    fake_sysctl "vm.panic_on_oom"
+    fake_sysctl "kernel.panic"
+    fake_sysctl "kernel.panic_on_oops"
+    fake_sysctl "kernel.keys.root_maxkeys"
+    fake_sysctl "kernel.keys.root_maxbytes"
   fi
 }
 
@@ -50,12 +136,16 @@ fix_mount() {
     sync
   fi
 
-  echo 'INFO: remounting /sys read-only'
-  # systemd-in-a-container should have read only /sys
-  # https://systemd.io/CONTAINER_INTERFACE/
-  # however, we need other things from `docker run --privileged` ...
-  # and this flag also happens to make /sys rw, amongst other things
-  mount -o remount,ro /sys
+  if [[ -z "${userns}" ]]; then
+    echo 'INFO: remounting /sys read-only'
+    # systemd-in-a-container should have read only /sys
+    # https://systemd.io/CONTAINER_INTERFACE/
+    # however, we need other things from `docker run --privileged` ...
+    # and this flag also happens to make /sys rw, amongst other things
+    #
+    # This step is skipped when running inside UserNS, because it fails with EACCES.
+    mount -o remount,ro /sys
+  fi
 
   echo 'INFO: making mounts shared' >&2
   # for mount propagation
@@ -212,6 +302,13 @@ fix_kmsg() {
     else
       echo 'WARN: /dev/kmsg does not exist, nor does /dev/console!' >&2
     fi
+  elif [[ -n "${userns}" ]]; then
+    if [[ -f "/proc/sys/kernel/dmesg_restrict" ]]; then
+      if [[ "$(cat /proc/sys/kernel/dmesg_restrict)" = "1" ]]; then
+        echo 'WARN: UserNS: /dev/kmsg is not readable, faking with /dev/null (hint: set sysctl value "kernel.dmesg_restrict" to 0)' >&2
+        mount --bind /dev/null /dev/kmsg
+      fi
+    fi
   fi
 }
 
@@ -298,6 +395,9 @@ enable_network_magic(){
     echo -n "${curr_ipv6}" >/kind/old-ipv6
   fi
 }
+
+# validate state
+validate_userns
 
 # run pre-init fixups
 # NOTE: it's important that we do configure* first in this order to avoid races
--- a/pkg/cluster/internal/providers/docker/provider.go
+++ b/pkg/cluster/internal/providers/docker/provider.go
@@ -17,6 +17,8 @@ limitations under the License.
 package docker
 
 import (
+	"encoding/csv"
+	"encoding/json"
 	"fmt"
 	"net"
 	"os"
@@ -280,4 +282,34 @@ func (p *provider) CollectLogs(dir string, nodes []nodes.Node) error {
 	// run and collect up all errors
 	errs = append(errs, errors.AggregateConcurrent(fns))
 	return errors.NewAggregate(errs)
+}
+
+// Info returns the provider info.
+func (p *provider) Info() (*providers.ProviderInfo, error) {
+	cmd := exec.Command("docker", "info", "--format", "{{json .SecurityOptions}}")
+	out, err := exec.Output(cmd)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get docker info")
+	}
+	var securityOptions []string
+	if err := json.Unmarshal(out, &securityOptions); err != nil {
+		return nil, err
+	}
+	var info providers.ProviderInfo
+	for _, o := range securityOptions {
+		// o is like "name=seccomp,profile=default", or "name=rootless",
+		csvReader := csv.NewReader(strings.NewReader(o))
+		sliceSlice, err := csvReader.ReadAll()
+		if err != nil {
+			return nil, err
+		}
+		for _, f := range sliceSlice {
+			for _, ff := range f {
+				if ff == "name=rootless" {
+					info.Rootless = true
+				}
+			}
+		}
+	}
+	return &info, nil
 }
--- a/site/content/docs/user/rootless.md
+++ b/site/content/docs/user/rootless.md
@@ -1,0 +1,59 @@
+---
+title: "Running kind with Rootless Docker"
+menu:
+  main:
+    parent: "user"
+    identifier: "rootless"
+    weight: 3
+---
+Starting with kind 0.11.0, [Rootless Docker](https://docs.docker.com/go/rootless/) and [Rootless Podman](https://github.com/containers/podman/blob/master/docs/tutorials/rootless_tutorial.md) can be used as the node provider of kind.
+
+## Provider requirements
+- Docker: 20.10 or later
+- Podman: 3.0 or later
+
+## Host requirements
+The host needs to be running with cgroup v2.
+
+cgroup v2 is enabled by default on Fedora.
+On other distros, cgroup v2 can be typically enabled by adding `GRUB_CMDLINE_LINUX="systemd.unified_cgroup_hierarchy=1"` to `/etc/default/grub` and
+running `sudo update-grub`.
+
+Also, depending on the host configuration, the following steps might be needed:
+
+- Create `/etc/systemd/system/user@.service.d/delegate.conf` with the following content, and then run `sudo systemctl daemon-reload`:
+```ini
+[Service]
+Delegate=yes
+```
+
+- Create `/etc/modules-load.d/iptables.conf` with the following content:
+```
+iptables_nat
+ip6tables_nat
+```
+
+## Restrictions
+
+The restrictions of Rootless Docker apply to kind clusters as well.
+
+e.g.
+- OverlayFS cannot be used unless the host is using kernel >= 5.11, or Ubuntu/Debian kernel
+- Cannot mount block storages
+- Cannot mount NFS
+
+## Creating a kind cluster with Rootless Docker
+
+To create a kind cluster with Rootless Docker, just run:
+```console
+$ export DOCKER_HOST=unix://${XDG_RUNTIME_DIR}/docker.sock
+$ kind create cluster
+```
+
+To create a kind cluster with Rootless Podman, just run:
+```console
+$ KIND_EXPERIMENTAL_PROVIDER=podman kind create cluster
+```
+
+## Tips
+- To enable OOM watching, allow `dmesg` by running `sysctl -w kernel.dmesg_restrict=0`.
--- a/pkg/cluster/internal/create/actions/config/config.go
+++ b/pkg/cluster/internal/create/actions/config/config.go
@@ -48,6 +48,11 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 	ctx.Status.Start("Writing configuration 📜")
 	defer ctx.Status.End(false)
 
+	providerInfo, err := ctx.Provider.Info()
+	if err != nil {
+		return err
+	}
+
 	allNodes, err := ctx.Nodes()
 	if err != nil {
 		return err
@@ -76,6 +81,7 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 		IPv6:                 ctx.Config.Networking.IPFamily == "ipv6",
 		FeatureGates:         ctx.Config.FeatureGates,
 		RuntimeConfig:        ctx.Config.RuntimeConfig,
+		RootlessProvider:     providerInfo.Rootless,
 	}
 
 	kubeadmConfigPlusPatches := func(node nodes.Node, data kubeadm.ConfigData) func() error {
--- a/pkg/cluster/internal/kubeadm/config.go
+++ b/pkg/cluster/internal/kubeadm/config.go
@@ -74,6 +74,10 @@ type ConfigData struct {
 	// These auto-generated fields are available to Config templates,
 	// but not meant to be set by hand
 	DerivedConfigData
+
+	// Provider is running with rootless mode, so kube-proxy needs to be configured
+	// not to fail on sysctl error.
+	RootlessProvider bool
 }
 
 // DerivedConfigData fields are automatically derived by
@@ -385,7 +389,14 @@ mode: "{{ .KubeProxyMode }}"
 {{end}}{{end}}
 iptables:
   minSyncPeriod: 1s
-{{end}}
+{{if .RootlessProvider}}conntrack:
+# Skip setting sysctl value "net.netfilter.nf_conntrack_max"
+  maxPerCore: 0
+# Skip setting "net.netfilter.nf_conntrack_tcp_timeout_established"
+  tcpEstablishedTimeout: 0s
+# Skip setting "net.netfilter.nf_conntrack_tcp_timeout_close"
+  tcpCloseWaitTimeout: 0s
+{{end}}{{end}}
 `
 
 // Config returns a kubeadm config generated from config data, in particular
@@ -404,6 +415,9 @@ func Config(data ConfigData) (config string, err error) {
 	// assume the latest API version, then fallback if the k8s version is too low
 	templateSource := ConfigTemplateBetaV2
 	if ver.LessThan(version.MustParseSemantic("v1.15.0")) {
+		if data.RootlessProvider {
+			return "", errors.Errorf("version %q is not compatible with rootless provider", ver)
+		}
 		templateSource = ConfigTemplateBetaV1
 	}
 
--- a/pkg/cluster/internal/providers/podman/provider.go
+++ b/pkg/cluster/internal/providers/podman/provider.go
@@ -68,12 +68,6 @@ func (p *provider) Provision(status *cli.Status, cfg *config.Cluster) (err error
 		return err
 	}
 
-	// kind doesn't work with podman rootless, surface an error
-	if os.Geteuid() != 0 {
-		p.logger.Errorf("podman provider does not work properly in rootless mode")
-		os.Exit(1)
-	}
-
 	// TODO: validate cfg
 	// ensure node images are pulled before actually provisioning
 	if err := ensureNodeImages(p.logger, status, cfg); err != nil {
@@ -349,4 +343,12 @@ func (p *provider) CollectLogs(dir string, nodes []nodes.Node) error {
 	// run and collect up all errors
 	errs = append(errs, errors.AggregateConcurrent(fns))
 	return errors.NewAggregate(errs)
+}
+
+// Info returns the provider info.
+func (p *provider) Info() (*providers.ProviderInfo, error) {
+	info := &providers.ProviderInfo{
+		Rootless: os.Geteuid() != 0,
+	}
+	return info, nil
 }
--- a/pkg/cluster/internal/providers/provider.go
+++ b/pkg/cluster/internal/providers/provider.go
@@ -45,4 +45,11 @@ type Provider interface {
 	GetAPIServerInternalEndpoint(cluster string) (string, error)
 	// CollectLogs will populate dir with cluster logs and other debug files
 	CollectLogs(dir string, nodes []nodes.Node) error
+	// Info returns the provider info
+	Info() (*ProviderInfo, error)
+}
+
+// ProviderInfo is the info of the provider
+type ProviderInfo struct {
+	Rootless bool
 }


images/base/Dockerfile
```diff
@@ -12,6 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# runc (v1.0.0-rc93-pre)
+FROM golang:1.15-alpine AS runc
```
BenTheElder, line 16: if we're going to build with cgo shouldn't we build with matching libc?

images/base/Dockerfile
```diff
@@ -12,6 +12,16 @@
...
+RUN apk add --no-cache git build-base libseccomp-dev
+RUN git clone -q https://github.com/opencontainers/runc.git /go/src/github.com/opencontainers/runc
+WORKDIR /go/src/github.com/opencontainers/runc
+# Nov 18, 2020
+ARG RUNC_VERSION="689513cc0946d8e968abbe35544d5a7827d2bcfa"
```
BenTheElder, line 21: can we get this runc into containerd instead? we track the version that matches containerd currently and don't need to build it every time we build the base image.

images/base/files/usr/local/bin/entrypoint
```diff
@@ -50,12 +130,16 @@ fix_mount() {
...
-  # and this flag also happens to make /sys rw, amongst other things
-  mount -o remount,ro /sys
+  if [[ -z "${userns}" ]]; then
+    echo 'INFO: remounting /sys read-only'
+    # systemd-in-a-container should have read only /sys
+    # https://systemd.io/CONTAINER_INTERFACE/
+    # however, we need other things from `docker run --privileged` ...
+    # and this flag also happens to make /sys rw, amongst other things
+    #
+    # This step is skipped when running inside UserNS, because it fails with EACCES.
```
BenTheElder, line 140: we need to disable udev

images/base/files/usr/local/bin/entrypoint
```diff
@@ -18,11 +18,91 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+userns=""
+if egrep -qv "0[[:space:]]+0[[:space:]]+4294967295" /proc/self/uid_map; then
```
BenTheElder, line 22: comment?

images/base/files/etc/containerd/config.toml
```diff
@@ -20,3 +22,5 @@ version = 2
   tolerate_missing_hugepages_controller = true
   # explicitly use default snapshotter so we can sed it in entrypoint
   snapshotter = "overlayfs"
+  # restrict_oom_score_adj needs to be true when running inside UserNS (rootless)
+  restrict_oom_score_adj = false
```
BenTheElder, line 26: is there a reason to not enable it by default?

site/content/docs/user/quick-start.md
```diff
@@ -425,6 +425,50 @@ The structure of the logs will look more or less like this:
 The logs contain information about the Docker host, the containers running 
 kind, the Kubernetes cluster itself, etc.
 
+### Rootless Docker
```
BenTheElder, line 428: I think this should maybe be a separate guide in the docs similar to WSL2, this page is a bit lengthy already.

images/base/Dockerfile
AkihiroSuda: It is built as a static binary

images/base/Dockerfile
AkihiroSuda: Yes, after the release of runc rc93. It will be very soon: https://github.com/opencontainers/runc/issues/2659

images/base/files/usr/local/bin/entrypoint
AkihiroSuda: What happens if not?

BenTheElder: udev is automatically disabled by /sys being `ro`, which is one of the systemd-in-a-container requirements.
when udev is not disabled, sometimes startup is broken by trying to "start" devices which cannot be started.
e.g. from the docker for mac xhyve.

site/content/docs/user/quick-start.md
```diff
@@ -425,6 +425,50 @@ The structure of the logs will look more or less like this:
...
+The host needs to be running with cgroup v2.
+
+cgroup v2 is enabled by default on Fedora.
+On other distros, cgroup v2 can be typically enabled by adding `GRUB_CMDLINE_LINUX="systemd.unified_cgroup_hierarchy=1"` to `/etc/default/grub` and
+running `sudo update-grub`.
+
+Also, depending on the host configuration, the following steps might be needed:
+
+- Create `/etc/systemd/system/user@.service.d/delegate.conf` with the following content, and then run `sudo systemctl daemon-reload`:
+```
```
jglick, line 442:
```suggestion
```ini
```

site/content/docs/user/quick-start.md
```diff
@@ -425,6 +425,50 @@ The structure of the logs will look more or less like this:
...
+[Service]
+Delegate=yes
+EOF
```
jglick, line 445: Typo?

```suggestion
```

site/content/docs/user/quick-start.md
```diff
@@ -425,6 +425,50 @@ The structure of the logs will look more or less like this:
...
+
+#### Creating a kind cluster with Rootless Docker
+
+To create a kind cluster with Rootless Docker, just run `kind create cluster` command with
+`DOCKER_HOST=unix://${XDG_RUNTIME_DIR}/docker.sock`.
+
+```console
+$ export DOCKER_HOST=unix://${XDG_RUNTIME_DIR}/docker.sock
+$ kind create cluster
+```
```
jglick, line 470:
```suggestion
To create a kind cluster with Rootless Docker, just run:

```bash
export DOCKER_HOST=unix://${XDG_RUNTIME_DIR}/docker.sock
kind create cluster
```
```

File: images/base/files/usr/local/bin/entrypoint (hunk 3 above)
Comment (BenTheElder) [reply images/base/files/usr/local/bin/entrypoint:140]: https://github.com/kubernetes-sigs/kind/issues/1474#issuecomment-613278644

File: images/base/files/etc/containerd/config.toml (hunk 5 above)
Comment (AkihiroSuda) [reply images/base/files/etc/containerd/config.toml:26]: If we enable it by default even for rootful mode, users cannot set oomScoreAdj that is stricter than containerd.

File: images/base/files/usr/local/bin/entrypoint (hunk 3 above)
Comment (AkihiroSuda) [reply images/base/files/usr/local/bin/entrypoint:170]: updated

File: images/base/files/usr/local/bin/entrypoint (hunk 4 above)
Comment (AkihiroSuda) [reply images/base/files/usr/local/bin/entrypoint]: updated

File: site/content/docs/user/quick-start.md (hunk 6 above)
Comment (AkihiroSuda) [reply site/content/docs/user/quick-start.md]: updated

File: site/content/docs/user/quick-start.md (hunk 7 above)
Comment (AkihiroSuda) [reply site/content/docs/user/quick-start.md]: updated

File: site/content/docs/user/quick-start.md (hunk 8 above)
Comment (AkihiroSuda) [reply site/content/docs/user/quick-start.md]: updated

File: site/content/docs/user/quick-start.md (hunk 9 above)
Comment (AkihiroSuda) [reply site/content/docs/user/quick-start.md]: updated

File: images/base/files/usr/local/bin/entrypoint [hunk 10]
```diff
@@ -50,12 +158,18 @@ fix_mount() {
...
   fi
 
-  echo 'INFO: remounting /sys read-only'
-  # systemd-in-a-container should have read only /sys
-  # https://systemd.io/CONTAINER_INTERFACE/
-  # however, we need other things from `docker run --privileged` ...
-  # and this flag also happens to make /sys rw, amongst other things
-  mount -o remount,ro /sys
+  if [[ -n "${userns}" ]]; then
+    systemctl disable udev.service
```
BenTheElder, line 162: should we just statically disable this in the base image?

images/base/Dockerfile
BenTheElder: [watching for this / subsrcribed, it doesn't seem _too_ far off and it would be preferable if possible]

images/base/files/usr/local/bin/entrypoint
AkihiroSuda: Do you mean we should disable udev even for rootful mode?

BenTheElder: yeah, we can do it at build-time in the base image right?
currently it's defacto ~disabled by ro /sys anyhow.

AkihiroSuda: opened a separate PR:  https://github.com/kubernetes-sigs/kind/pull/1948

AkihiroSuda:
Now this PR needs to wait for cgroup v2 fix: https://github.com/kubernetes-sigs/kind/issues/2013

images/base/files/usr/local/bin/entrypoint
BenTheElder: this is merged now 👍

AkihiroSuda: Thanks for merging!

AkihiroSuda:
Rebased. Tested with Kubernetes v1.20.2.

BenTheElder:
runc update upstream should be picked up in https://github.com/kubernetes-sigs/kind/pull/2057

images/base/Dockerfile
```diff
@@ -12,6 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# runc (v1.0.0-rc93-pre)
```
BenTheElder, line 15: this change should be unecessary after https://github.com/kubernetes-sigs/kind/pull/2057 (latest containerd build should have rc93)

AkihiroSuda: seems to have caused regression :( https://github.com/kubernetes-sigs/kind/issues/2058

AkihiroSuda: The issue #2058 is now fixed.

AkihiroSuda:
Rebased. Tested with Kubernetes v1.20.4.

images/base/Dockerfile
AkihiroSuda: We have rc93 now

AkihiroSuda:
Podman CI is failing, but unrelated: https://github.com/kubernetes-sigs/kind/issues/2085

images/base/files/usr/local/bin/entrypoint
```diff
@@ -18,11 +18,119 @@ set -o errexit
...
+    echo "ERROR: UserNS: expected RLIMIT_NOFILE to be at least ${nofile_hard_expected}, got ${nofile_hard}" >&2
+    exit 1
+  fi
+
+  local nf_conntrack_max
+  nf_conntrack_max="$(cat /proc/sys/net/netfilter/nf_conntrack_max)"
+  local nf_conntrack_max_expected="131072"
+  if [[ "${nf_conntrack_max}" != "${nf_conntrack_max_expected}" ]]; then
+    # This ERROR can be demoted to WARNING when k/k PR gets merged: https://github.com/kubernetes/kubernetes/pull/92863
+    echo "ERROR: UserNS: expected net.netfilter.nf_conntrack_max to be ${nf_conntrack_max_expected}, got ${nf_conntrack_max}" >&2
```
aojea, line 49: I don't have the historical context, but I don't think this should be errors because users should be able to tweak this parameter, is not that something is going to break if this is not matching this specific value ... it can perform much worse, but that is another conversation

AkihiroSuda: Without this value, kube-proxy error out during attempting to set sysctl. So until 
 https://github.com/kubernetes/kubernetes/pull/92863 gets merged we need this check.

aojea: ic, this is a configuration parameter to the kube-proxy, so this will ONLY work for the default value "131072", I don't think nobody will tweak this parameter and that you can get it beforehand so this seems the only way

aojea: yeah, agree, I was curious only :)

site/content/docs/user/rootless.md
```diff
@@ -0,0 +1,49 @@
+---
+title: "Running kind with Rootless Docker"
+menu:
+  main:
+    parent: "user"
+    identifier: "rootless"
+    weight: 3
+---
+Starting with kind 0.11.0 and Docker 20.10, Rootless Docker can be used as the node provider of kind.
```
aojea, line 9: Should we link to the docker rootless install docs?
https://docs.docker.com/engine/security/rootless/

aojea:
/ok-to-test
I'm trying to launch it by myself, hitting some issues with docker ...

BenTheElder:
@aojea note that testing this doesn't test this 🙃
- there's not base image yet
- we don't have rootless CI yet

(So I left it not ok-to-test to avoid CI noise that isn't telling us much)

aojea:
> @aojea note that testing this doesn't test this 
> 
> * there's not base image yet
> * we don't have rootless CI yet
> 
> (So I left it not ok-to-test to avoid CI noise that isn't telling us much)

yeah, I'm testing it locally, I've removed the ok-to-test just to indicate somehow that I was actively reviewing ... not giving the impression it was stuck

aojea:
> docker build -t kindest/base:latest ./images/base 

@BenTheElder should that works or should it be done with the `make`

> f54b8973bdc7b119eb4e7ac642cf7652d35954b37f97a7c1fa71f3d062badc29  /tmp/containerd.ppc64le.tgz                                                                        sha256sum: /tmp/containerd.sha256: no file was verified                                                                                                              The command '/bin/sh -c echo "Ensuring scripts are executable ..."     && chmod +x /usr/local/bin/clean-install /usr/local/bin/entrypoint  && echo "Installing Packages ..."     && DEBIAN_FRONTEND=noninteractive clean-install       systemd       conntrack iptables iproute2 ethtool socat util-linux mount ebtables udev kmod       libseccomp2 pigz       bash ca-certificates curl rsync       nfs-common       jq     && find /lib/systemd/system/sysinit.target.wants/ -name "systemd-tmpfiles-setup.service" -delete     && rm -f /lib/systemd/system/multi-user.target.wants/*     && rm -f /etc/systemd/system/*.wants/*     && rm -f /lib/systemd/system/local-fs.target.wants/*     && rm -f /lib/systemd/system/sockets.target.wants/*udev*     && rm -f /lib/systemd/system/sockets.target.wants/*initctl*     && rm -f /lib/systemd/system/basic.target.wants/*     && echo "ReadKMsg=no" >> /etc/systemd/journald.conf     && ln -s "$(which systemd)" /sbin/init  && echo "Enabling kubelet ... "     && systemctl enable kubelet.service  && echo "Installing containerd ..."     && curl -sSL --retry 5 --output /tmp/containerd.${TARGETARCH}.tgz "${CONTAINERD_URL}"     && echo "${CONTAINERD_AMD64_SHA256SUM}  /tmp/containerd.amd64.tgz" | tee /tmp/containerd.sha256     && echo "${CONTAINERD_ARM64_SHA256SUM}  /tmp/containerd.arm64.tgz" | tee -a /tmp/containerd.sha256     && echo "${CONTAINERD_PPC64LE_SHA256SUM}  /tmp/containerd.ppc64le.tgz" | tee -a /tmp/containerd.sha256     && sha256sum --ignore-missing -c /tmp/containerd.sha256     && rm -f /tmp/containerd.sha256     && tar -C /usr/local -xzvf /tmp/containerd.${TARGETARCH}.tgz     && rm -rf /tmp/containerd.${TARGETARCH}.tgz     && rm -f /usr/local/bin/containerd-stress /usr/local/bin/containerd-shim-runc-v1     && curl -sSL --retry 5 --output /tmp/runc.${TARGETARCH} "${RUNC_URL}"     && echo "${RUNC_AMD64_SHA256SUM}  /tmp/runc.amd64" | tee /tmp/runc.sha256     && echo "${RUNC_ARM64_SHA256SUM}  /tmp/runc.arm64" | tee -a /tmp/runc.sha256     && echo "${RUNC_PPC64LE_SHA256SUM}  /tmp/runc.ppc64le" | tee -a /tmp/runc.sha256     && sha256sum --ignore-missing -c /tmp/runc.sha256     && mv /tmp/runc.${TARGETARCH} /usr/local/sbin/runc     && chmod 755 /usr/local/sbin/runc     && containerd --version     && runc --version     && systemctl enable containerd  && echo "Installing crictl ..."     && curl -sSL --retry 5 --output /tmp/crictl.${TARGETARCH}.tgz "${CRICTL_URL}"     && echo "${CRICTL_AMD64_SHA256SUM}  /tmp/crictl.amd64.tgz" | tee /tmp/crictl.sha256     && echo "${CRICTL_ARM64_SHA256SUM}  /tmp/crictl.arm64.tgz" | tee -a /tmp/crictl.sha256     && echo "${CRICTL_PPC64LE_SHA256SUM}  /tmp/crictl.ppc64le.tgz" | tee -a /tmp/crictl.sha256     && sha256sum --ignore-missing -c /tmp/crictl.sha256     && rm -f /tmp/crictl.sha256     && tar -C /usr/local/bin -xzvf /tmp/crictl.${TARGETARCH}.tgz     && rm -rf /tmp/crictl.${TARGETARCH}.tgz  && echo "Installing CNI plugin binaries ..."     && curl -sSL --retry 5 --output /tmp/cni.${TARGETARCH}.tgz "${CNI_PLUGINS_URL}"     && echo "${CNI_PLUGINS_AMD64_SHA256SUM}  /tmp/cni.amd64.tgz" | tee /tmp/cni.sha256     && echo "${CNI_PLUGINS_ARM64_SHA256SUM}  /tmp/cni.arm64.tgz" | tee -a /tmp/cni.sha256     && echo "${CNI_PLUGINS_PPC64LE_SHA256SUM}  /tmp/cni.ppc64le.tgz" | tee -a /tmp/cni.sha256     && sha256sum --ignore-missing -c /tmp/cni.sha256     && rm -f /tmp/cni.sha256     && mkdir -p /opt/cni/bin     && tar -C /opt/cni/bin -xzvf /tmp/cni.${TARGETARCH}.tgz     && rm -rf /tmp/cni.${TARGETARCH}.tgz     && find /opt/cni/bin -type f -not \(          -iname host-local          -o -iname ptp          -o -iname portmap          -o -iname loopback       \)       -delete  && echo "Ensuring /etc/kubernetes/manifests"     && mkdir -p /etc/kubernetes/manifests  && echo "Adjusting systemd-tmpfiles 
> timer"     && sed -i /usr/lib/systemd/system/systemd-tmpfiles-clean.timer -e 's#OnBootSec=.*#OnBootSec=1min#'  && echo "Disabling udev"     && systemctl disable udev
> .service' returned a non-zero code: 1

BenTheElder:
> @BenTheElder should that works or should it be done with the make

no, use the make command. `make -C images/base quick` for testing. (I think this image even has docs?)

aojea:
I have to be doing something wrong creating the image, but I'm stuck on this error
```
 ./kind create cluster --image c6d0593a9058                                                                                                   
Creating cluster "kind" ...                                                                                                                                          
 ✓ Ensuring node image (c6d0593a9058) 🖼                                                                                                                              
 ✗ Preparing nodes 📦                                                                                                                                                
ERROR: failed to create cluster: docker run error: command "docker run --hostname kind-control-plane --name kind-control-plane --label io.x-k8s.kind.role=control-pla
ne --privileged --security-opt seccomp=unconfined --security-opt apparmor=unconfined --tmpfs /tmp --tmpfs /run --volume /var --volume /lib/modules:/lib/modules:ro --
detach --tty --label io.x-k8s.kind.cluster=kind --net kind --restart=on-failure:1 --init=false --publish=127.0.0.1:44957:6443/TCP c6d0593a9058" failed with error: ex
it status 125                                                                                                                                                        
Command Output: docker: Error response from daemon: No command specified.
```

AkihiroSuda:
> Error response from daemon: No command specified.

I guess you might have accidentally specified an invalid image that lacks ENTRYPOINT?

aojea:
> I guess you might have accidentally specified an invalid image that lacks ENTRYPOINT?

I'm building the images from this PR, I've tried twice with the same result, let me check again, bear in mind that the base image building process is not the same as the one you have in the description https://github.com/kubernetes-sigs/kind/pull/1935#issuecomment-786786318

AkihiroSuda:
bazel seems to have been removed in https://github.com/kubernetes/kubernetes/pull/99561, so the most robust way of testing this PR would be to build the node image on rootful and then push it to some registry. (Or just `docker save` and `docker load`).

Eventually we will be able to support rootless & bazelless image building in a follow-up PR.

aojea:
I've tried building it locally and pushing the images to dockerhub, one image from master and other from 1.20.4, both failed :(
```
docker images
REPOSITORY       TAG              IMAGE ID       CREATED          SIZE
aojea/kindnode   rootless1.20.4   b4ab21eda6f9   24 minutes ago   1.25GB
aojea/kindnode   rootless         c76acaecb5e3   7 hours ago      1.17GB
```

I'm running docker in an ubuntu vm following your configuration Steps 1:

```
docker info
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Build with BuildKit (Docker Inc., v0.5.1-docker)

Server:
 Containers: 12
  Running: 0
  Paused: 0
  Stopped: 12
 Images: 6
 Server Version: 20.10.4
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 269548fa27e0089a8b8278fc4fc781d7f65a939b
 runc version: ff819c7e9184c13b7c2607fe6c30ae19403a7aff
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
  rootless
  cgroupns
 Kernel Version: 5.4.0-66-generic
 Operating System: Ubuntu 20.04.2 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 2
 Total Memory: 981MiB
 Name: insane-tudela
 ID: ZDF4:IIET:65OQ:QO4Q:K2E6:N7JQ:USUS:WKUJ:GDTH:JCN3:YRJW:26W6
 Docker Root Dir: /home/ubuntu/.local/share/docker
 Debug Mode: false
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false

WARNING: Support for cgroup v2 is experimental
```

I've checked the images have your scripts:
```
docker run -it --privileged --entrypoint bash aojea/kindnode:rootless
root@5201ca84504d:/# grep fake_sysc /usr/local/bin/entrypoint 
fake_sysctl() {
    fake_sysctl "vm.overcommit_memory"
    fake_sysctl "vm.panic_on_oom"
    fake_sysctl "kernel.panic"
    fake_sysctl "kernel.panic_on_oops"
    fake_sysctl "kernel.keys.root_maxkeys"
    fake_sysctl "kernel.keys.root_maxbytes"
```
The weird thing is that it fails with this error now:
```
./kind create cluster --image aojea/kindnode:rootless1.20.4 --retain
Creating cluster "kind" ...
 ✓ Ensuring node image (aojea/kindnode:rootless1.20.4) 🖼
 ✓ Preparing nodes 📦  
 ✗ Writing configuration 📜 
ERROR: failed to create cluster: failed to generate kubeadm config content: failed to get kubernetes version from node: failed to get file: command "docker exec --privileged kind-control-plane cat /kind/version" failed with error: exit status 1
Command Output: Error response from daemon: Container f09b51a9400b5bc77738ac9bff7ff0c2c4f2d8dd7c2e81d1f09203cedba3ade6 is not 
```

It logs:
```
docker logs f09b51a9400b
INFO: running in a user namespace (experimental)
cat: /proc/sys/net/netfilter/nf_conntrack_max: No such file or directory
INFO: running in a user namespace (experimental)
cat: /proc/sys/net/netfilter/nf_conntrack_max: No such file or directory
```

I can see the `/kind/version` file inside the node image:
```
docker run -it --privileged --entrypoint bash aojea/kindnode:rootless 
root@51dcb78c682b:/# cat /kind/version 
v1.21.0-beta.0.217+f79795d71850abroot@51dcb78c682b:/# 
````
```
docker run -it --privileged --entrypoint bash aojea/kindnode:rootless1.24 
root@03feab04371d:/# cat /kind/version 
v1.20.4root@03feab04371d:/#
```

I feel I'm missing something, can you try with the images I've published in dockerhub  @AkihiroSuda or publish one that I can try?

**Comment by AkihiroSuda** [comment]
`/proc/sys/net/netfilter/nf_conntrack_max` seems to need kernel >= 5.7 https://github.com/torvalds/linux/commit/d0febd81ae77a0e13717f1412ff9589e43fc4f8b

Could you try more recent distro? I'm using Ubuntu 20.10 (kernel 5.8).

**Comment by AkihiroSuda** [comment]
Modified the PR to check kernel version on start up. Also mentioned the requirement in `site/content/docs/user/rootless.md`.

**Comment by AkihiroSuda** [comment]
Pushed a new node image to my Docker Hub: akihirosuda/tmp-kind-node:g99077d5d-v1.20.4

**Comment by AkihiroSuda** [comment]
Updated PR to remove prerequisite of "net.netfilter.nf_conntrack_max" , by setting `kubeProxyConfiguration.conntrack.maxPerCore` to 0 . 

Image: `akihirosuda/tmp-kind-node:gf4eef8b2-v1.20.4`

This version may work on Ubuntu 20.04.

**Comment by aojea** [comment]
Great, it works for me now, nice,  I never used rootless before, it has some networking edges, i.e. you can't access containers from the host ...

I this this approach is better because we reduce bash and is easier to maintain in the long term, however, we need @BenTheElder to check the new changes introduced in the provider interface

**Comment by AkihiroSuda** [comment]
Updated again
- Allow `kernel.dmesg_restrict=1`, by bind-mounting `/dev/null` to `/dev/kmsg`
- Unlock Podman

**Comment by AkihiroSuda** [comment]
Rebased and squashed commits.

Image: `akihirosuda/tmp-kind-node:g85d51d89-v1.20.4`

File: site/content/docs/user/rootless.md [hunk 14]
```diff
@@ -0,0 +1,59 @@
+---
+title: "Running kind with Rootless Docker"
+menu:
+  main:
+    parent: "user"
+    identifier: "rootless"
+    weight: 3
+---
+Starting with kind 0.11.0, [Rootless Docker](https://docs.docker.com/go/rootless/) and [Rootless Podman](https://github.com/containers/podman/blob/master/docs/tutorials/rootless_tutorial.md) can be used as the node provider of kind.
```
BenTheElder, line 9: nit: 
```suggestion
Starting with kind v0.11.0, [Rootless Docker](https://docs.docker.com/go/rootless/) and [Rootless Podman](https://github.com/containers/podman/blob/master/docs/tutorials/rootless_tutorial.md) can be used as the node provider of kind.
```
(we should try to be consistent about this)
(non-blocking nit, can easily handle this sort of thing in a follow-up)

pkg/cluster/internal/providers/docker/provider.go
```diff
@@ -281,3 +283,33 @@ func (p *provider) CollectLogs(dir string, nodes []nodes.Node) error {
 	errs = append(errs, errors.AggregateConcurrent(fns))
 	return errors.NewAggregate(errs)
 }
+
+// Info returns the provider info.
+func (p *provider) Info() (*providers.ProviderInfo, error) {
```
BenTheElder, line 288: I'm not sure I love this API, but the point of node providers being internal is precisely so we can iterate on stuff like this in isolated implementation packages (still exporting) without worrying about users depending on it, versus the public cluster provider. (comment directed at @aojea)

images/base/files/usr/local/bin/entrypoint
```diff
@@ -18,11 +18,97 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+# If /proc/self/uid_map 4294967295 mappings, we are in the initial user namespace, i.e. the host.
+# Otherwise we are in a non-initial user namespace.
+# https://github.com/opencontainers/runc/blob/v1.0.0-rc92/libcontainer/system/linux.go#L109-L118
+userns=""
```
BenTheElder, line 24: now that we have to detect rootless provider on the host anyhow, is there a reason to do this inside the container?
do we expect this to vary within rootless?
or should we just start passing a KIND_ROOTLESS_NODE_PROVIDER=true ?

BenTheElder:
thank you for keeping after this. 
we've been having fun with kubernetes code freeze which has kept me a bit low bandwidth here. I'd like to go ahead and merge this this week and iterate from where we're at rather than let this linger longer.

images/base/files/usr/local/bin/entrypoint
AkihiroSuda: Checking userns here is beneficial for potential support of `dockerd --userns-remap` and LXD driver.

pkg/cluster/internal/providers/docker/provider.go
aojea: yeah, but that makes a big assumption about providers compatibility, and we have right now 2 providers that try to be completely compatible, I can't see how this will go with kata or others, ...

BenTheElder: we can pretty trivially rework this method later though, this PR has already seen a lot of back and forth

BenTheElder:
/lgtm
/approve
I think there's room to bikeshed some more for sure, but erring on the side of unblocking for now. Thank you.

PR #1937, opened 2020-11-18
disable CGO by default

BenTheElder:
This also shrinks the binary another bit on linux 7.1M to 7.0M
Fixes a build failure I encountered.

iI someone really wants CGO (no real reason currently) they can build with `make build CGO_ENABLED=1`


krzyzacy (approved):

amwat (commented):
/lgtm
/retest

PR #1940, opened 2020-11-22
CI podman IPv6 multinode

aojea:
Podman 2.2.x has support for IPv6 compatible with docker

xref: https://github.com/containers/podman/issues/7302


aojea:
duplicate of https://github.com/kubernetes-sigs/kind/pull/1940

PR #729, opened 2019-07-23
Give yet another zsh completion help/example

RafalKorepta:
Add one line example for users who install zsh-completion using homebrew.

Issue #719, opened 2019-07-18
zsh completions throws an error


jecnua:
Hi all,

**What happened**:

The kind zsh completions throws and error when you try to use it.

**What you expected to happen**:

I expect the completions to work :)

**How to reproduce it (as minimally and precisely as possible)**:

```
kind completion zsh > /usr/local/share/zsh/site-functions/_kind
autoload -U compinit && compinit
kind<TAB>
```

**Environment:**

- kind version: (use `kind version`): v0.3.0
- Kubernetes version: (use `kubectl version`): v1.15.0
- Docker version: (use `docker info`): 19.03.0-rc2
- OS (e.g. from `/etc/os-release`): Darwin xxx 18.6.0 Darwin Kernel Version 18.6.0

```
zsh --version
zsh 5.3 (x86_64-apple-darwin18.0)
```

```
brew info zsh-completions
zsh-completions: stable 0.30.0, HEAD
```

Let me know if I can help with more info.

aojea:
is this a duplicate of https://github.com/kubernetes-sigs/kind/issues/522?

jecnua:
No. I have read the other ticket before. That is just asking for a way to source the autocomplete in zsh :) This problem I have is following the normal way that is shown in:

`kind completion -h`

BenTheElder:
This is unfortunate. The zsh completions are auto generated by a dependency (https://github.com/spf13/cobra), I think they've been generally problematic upstream :/

BenTheElder:
ref: https://github.com/kubernetes-sigs/kind/pull/729

BenTheElder:
I think https://github.com/kubernetes-sigs/kind/pull/729 documented this (please correct me, I don't use zsh...)

Issue #522, opened 2019-05-13
zsh completions can't be sourced


Mikewl:
**What happened**:
Running `source <(kind completion zsh)` fails with the following error
```
_arguments:comparguments:325: can only be called from completion function
_arguments:comparguments:325: can only be called from completion function
```

**What you expected to happen**:
The completions to import and be available

**How to reproduce it (as minimally and precisely as possible)**:
Run `source <(kind completion zsh)` in zsh

**Anything else we need to know?**:

**Environment:**
- zsh version:  zsh 5.7.1 (x86_64-apple-darwin18.2.0)
- kind version: (use `kind version`): 0.3.0-alpha
- Kubernetes version: (use `kubectl version`): 1.14
- Docker version: (use `docker info`): 18.09.2
- OS (e.g. from `/etc/os-release`): MacOS 10.14

BenTheElder:
hmm, I wonder if this is a bug upstream in cobra, we are generating these with that library
https://github.com/spf13/cobra

nirnanaaa:
Might be related to https://github.com/spf13/cobra/pull/646

rsteube:
Did this work before? As afaik you don't source completions (as in bash) but instead add the folders containing them to fpath: https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions-howto.org#telling-zsh-which-function-to-use-for-completing-a-command

e.g:
`kind completion zsh > /somedir/_kind` (once to generate it)
and in .zshrc
```
fpath=(/somedir $fpath)`
# and something like this (normally done by plugin managers - ideally only once for performance)
autoload -Uz compinit && \
   compinit -C
```

paranoidaditya:
👋 

I added the autocompletion subcommand. 

If you run kind completion

 `
Outputs kind shell completion for the given shell (bash or zsh)
This depends on the bash-completion binary.  Example installation instructions:
# for bash users
	$ kind completion bash > ~/.kind-completion
	$ source ~/.kind-completion
# for zsh users
	% kind completion zsh > /usr/local/share/zsh/site-functions/_kind
	% autoload -U compinit && compinit
Additionally, you may want to output the completion to a file and source in your .bashrc
Note for zsh users: [1] zsh completions are only supported in versions of zsh >= 5.2
`

There's a PR in `spf13/cobra` which lets you source autocompletion in zsh directly like bash but that's hasn't been merged yet

BenTheElder:
[...]

PR #1942, opened 2020-11-25
Add completion for oh-my-zsh

FedericoAntoniazzi:
### Issue
I've installed kind few days ago and never been able to setup completion for zsh using the commands printed by `kind completion zsh` because both paths are incorrect.
The first one tells the user to create the `/usr/local/share/zsh/site-functions/_kind` file but in my case there isn't that path in my system, sudo privileges are required and zsh doesn't read files from there. The second option uses fpath to determine the path in which the file should be located but, in my case it was the `~/.oh-my-zsh/` folder and _kind was completely ignored.

### Proposed solution
My solution is an extended but more precise version of the second method and uses a predefined path (already present in fpath) `~/.oh-my-zsh/completions` but requires the plugin [zsh-completions](https://github.com/zsh-users/zsh-completions) to be installed.

#### Acknowledgments
1. #522
2. #719
3. #729

### DIFF
--- a/pkg/cmd/kind/completion/completion.go
+++ b/pkg/cmd/kind/completion/completion.go
@@ -53,6 +53,9 @@ This depends on the bash-completion binary.  Example installation instructions:
 	% autoload -U compinit && compinit
 # or if zsh-completion is installed via homebrew
     % kind completion zsh > "${fpath[1]}/_kind"
+# or if you use oh-my-zsh (needs zsh-completions plugin)
+	% mkdir ~/.oh-my-zsh/completions/
+	% kind completion zsh > ~/.oh-my-zsh/completions/_kind
 
 # for fish users
 	% kind completion fish > ~/.config/fish/completions/kind.fish
--- a/pkg/cmd/kind/completion/completion.go
+++ b/pkg/cmd/kind/completion/completion.go
@@ -54,8 +54,8 @@ This depends on the bash-completion binary.  Example installation instructions:
 # or if zsh-completion is installed via homebrew
     % kind completion zsh > "${fpath[1]}/_kind"
 # or if you use oh-my-zsh (needs zsh-completions plugin)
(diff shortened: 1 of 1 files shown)


FedericoAntoniazzi:
> It may take a couple minutes for the CLA signature to be fully registered; after that, please reply here with a new comment and we'll verify.  Thanks.

I've signed up to CLA

BenTheElder:
Hi and thank you, I don't think we have any ZFS using maintainers so it's all been user contributed ❤️

Unfortunately the Kubernetes organization requires the CNCF CLA check to pass, more on that here https://github.com/kubernetes-sigs/kind/pull/1942#issuecomment-733357124

FedericoAntoniazzi:
> Hi and thank you, I don't think we have any ~ZFS~ ZSH (brain fart ... I use ZFS but not ZSH 🙃) using maintainers so it's all been user contributed ❤️
> 
> Unfortunately the Kubernetes organization requires the CNCF CLA check to pass, more on that here https://github.com/kubernetes-sigs/kind/pull/1942#issuecomment-733357124

I had problems with the CLA but now it's ok. 😀

I've proposed a solution that worked on my main pc, tested in a VM and a second laptop all using a fresh installation of linux systems (Arch Linux and Ubuntu 20.04) and worked perfectly.
I would like it if some other people could test the commands in order to be sure I haven't missed anything or an edge/particular case 😅

FedericoAntoniazzi:
@BenTheElder I thought the CI was failing because there were new commits on master so I used git rebase but now the error seems to be caused by podman. 
I don't know what to do now, do someone could tell me a hint for a possible solution or something related?

BenTheElder:
sorry, was out for a bit.
podman failing was due to a breakage in podman. kind has since worked around it.

the most recent failure is just a flake in kubernetes testing, it's fine.

BenTheElder:
/lgtm
thank you.

PR #1944, opened 2020-11-26
Prevent remote access to insecure registry

nicholasdille:
**Issue**

The example to use a local insecure registry in a kind cluster also exposes the registry on the local network. See https://github.com/kubernetes-sigs/kind/blob/master/site/static/examples/kind-with-registry.sh#L10

**Solution**

Limit the port publishing for the local insecure registry to localhost. The change was checked by running the following:

```
docker pull nginx
docker tag nginx localhost:5000/nginx
docker push localhost:5000/nginx
kubectl run --image=localhost:5000/nginx nginx
```


nicholasdille:
I have signed the CLA :-)

BenTheElder:
/lgtm
/approve
thank you for this, sorry for the delay, we've had all PRs backlogged with people on vacation and chasing a tricky breakage inbetween.

PR #1960, opened 2020-12-04
clarify error getting ipv6 address on node

aojea:
Improve the validation when getting an IPv6 address from a node, and clarify the error message, specifying the container provider that is failing

### DIFF
--- a/pkg/cluster/internal/create/actions/config/config.go
+++ b/pkg/cluster/internal/create/actions/config/config.go
@@ -20,6 +20,7 @@ package config
 import (
 	"bytes"
 	"fmt"
+	"net"
 	"strings"
 
 	"sigs.k8s.io/kind/pkg/cluster/constants"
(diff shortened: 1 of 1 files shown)


pkg/cluster/internal/create/actions/config/config.go
```diff
@@ -166,7 +168,7 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 
 // getKubeadmConfig generates the kubeadm config contents for the cluster
 // by running data through the template and applying patches as needed.
-func getKubeadmConfig(cfg *config.Cluster, data kubeadm.ConfigData, node nodes.Node) (path string, err error) {
+func getKubeadmConfig(cfg *config.Cluster, data kubeadm.ConfigData, node nodes.Node, provider string) (path string, err error) {
```
aojea, line 171: I doubted between passing the whole context or adding a new argument, since it is only a string and doesn't look like this is going to grow much, I pick the later

BenTheElder: I still need to take a sledgehammer to this code anyhow :-)

this is a reasonable approach for now :+1:

pkg/cluster/internal/create/actions/config/config.go
```diff
@@ -60,8 +61,9 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 	// create kubeadm init config
 	fns := []func() error{}
 
+	provider := fmt.Sprintf("%s", ctx.Provider)
```
BenTheElder, line 64: huh did I do that? This is a pretty roundabout way to call `ctx.Provider.String()` ...

aojea: ` type provider.Provider has not method or type String()` 
how does the stringer works with interfaces by the way? 
is not necessary to export it in the interface?

BenTheElder: ah, no so, the interface doesn't expose it, but `Sprintf` for `"%s"` will check if it can assert `fmt.Stringer` and call String anyhow.

this is why I did this 🙃

PR #1965, opened 2020-12-07
Add CI for docker in Github Actions

aojea:
It also fixes a bug where the images failed to run after they were created locally

When we added `Support configuring a cgroup root for kubelet #1747` , we checked for podman or docker to remount the cgroups. However, docker in Github doesn't show docker in its cgroup path, causing that the script didn't remounted the cgroups subsystems under /kubelet, hence the `kubelet`, that is configured to use `cgroup-root: /kubelet` fails to start because there is nothing in that path.

```
grep /sys/fs/cgroup /proc/self/mountinfo | grep docker
root@kind-control-plane:/usr/local/bin# 
```

Fixes: #1969

### DIFF
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -1,0 +1,86 @@
+name: Compatibility
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  images:
+    name: Node Images
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+
+      - name: Verify
+        run: make verify
+
+      - name: Install kind from current repository
+        run: sudo make install INSTALL_DIR=/usr/local/bin
+
+      - name: Checkout Kubernetes
+        uses: actions/checkout@v2
+        with:
+          path: src/k8s.io/kubernetes
+          repository: kubernetes/kubernetes
+          ref: v1.20.0
+
+      - name: Install kubectl
+        run: |
+          curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl
+          chmod +x ./kubectl
+          sudo mv ./kubectl /usr/local/bin/kubectl
+
+      - name: Build kind node image from current repository
+        run: |
+          KUBE_PATH="${{ github.workspace }}/src/k8s.io/kubernetes"
+          kind build node-image --kube-root ${KUBE_PATH} --image node:test -v7
+
+      - name: Create cluster from current repository
+        run: |
+          kind create cluster --name kind-pr --wait 1m --image node:test -v7 --retain
+
+      - name: Create cluster with last stable version
+        run: |
+          curl -Lo ./kind https://kind.sigs.k8s.io/dl/v0.9.0/kind-linux-amd64
+          chmod +x ./kind
+          cat <<EOF | ./kind create cluster -v7 --wait 1m --image node:test --retain --config=-
+          kind: Cluster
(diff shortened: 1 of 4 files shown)


.github/workflows/docker.yml
```diff
@@ -0,0 +1,95 @@
...
+      fail-fast: false
+      matrix:
+        ipFamily: [ipv4, ipv6]
+        deployment: [singleNode, multiNode]
+    env:
+      JOB_NAME: "docker-${{ matrix.deployment }}-${{ matrix.ipFamily }}"
+      IP_FAMILY: ${{ matrix.ipFamily }}
+    steps:
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
```
aojea, line 24: @BenTheElder FYI https://github.com/actions/checkout/issues/27#issuecomment-535897113

BenTheElder:
interesting that it does not break here 🤔

aojea:
for reference, it fails here for the cgroup you can get the logs from here https://github.com/aojea/ovn-kubernetes/runs/1512853588?check_suite_focus=true with following error

> > Dec 07 19:16:44 ovn-control-plane kubelet[994]: F1207 19:16:44.170173     994 server.go:269] failed to run Kubelet: invalid configuration: cgroup-root ["kubelet"] doesn't exist
Dec 07 19:16:44 ovn-control-plane kubelet[994]: goroutine 1 [running]:

aojea:
we have a reproduced and a job to test that we don´t break last stable version
https://github.com/kubernetes-sigs/kind/pull/1965/checks?check_run_id=1529704560

.github/workflows/docker.yml
```diff
@@ -0,0 +1,95 @@
...
+          sudo kubectl wait --for=condition=ready pods --namespace=kube-system -l k8s-app=kube-dns
+          sudo kubectl get nodes -o wide
+          sudo kubectl get pods -A
+
+      - name: Load docker image
+        run: kind load docker-image busybox:2
+        continue-on-error: true
+
+      - name: Export logs
+        if: always()
```
lotharschulz, line 81:
```suggestion
if: ${{ always() }}
```

lotharschulz: The _double curly braces_ syntax is used in line 56 as well and is in line with the [documentation](https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#example-8). If accepted here the same change should apply to line 88.

aojea: oh, thanks

pkg/build/nodeimage/defaults.go
```diff
@@ -20,7 +20,7 @@ package nodeimage
 const DefaultImage = "kindest/node:latest"
 
 // DefaultBaseImage is the default base image used
-const DefaultBaseImage = "kindest/base:v20201130-23777eca"
+const DefaultBaseImage = "kindest/base:v20201204-49cad832"
```
aojea, line 23: ping @BenTheElder , however, this doesn't solve the problem

BenTheElder: where did you pull that from? note: random images in this repo are not supported.
I dont think we missed any bumps merging.

aojea: pulled from here https://hub.docker.com/r/kindest/base/tags?page=1&ordering=last_updated , it says updated 11 days ago

aojea:
/close
https://github.com/kubernetes-sigs/kind/pull/1979

.github/workflows/compatibility.yml
```diff
@@ -0,0 +1,86 @@
+name: Compatibility
```
BenTheElder, line 1: this we don't guarantee and was a red-herring

.github/workflows/docker.yml
```diff
@@ -0,0 +1,95 @@
+name: Docker
```
BenTheElder, line 1: a variant of this we should probably keep somewhere.

PR #1967, opened 2020-12-08
Move cluster name validation to Validate method

knabben:
Quick chore: moving the name regex parsing to `cluster.Validate()` method

### DIFF
--- a/pkg/cluster/internal/create/create.go
+++ b/pkg/cluster/internal/create/create.go
@@ -19,7 +19,6 @@ package create
 import (
 	"fmt"
 	"math/rand"
-	"regexp"
 	"time"
 
 	"github.com/alessio/shellescape"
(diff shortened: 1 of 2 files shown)


pkg/cluster/internal/create/create.go
```diff
@@ -83,14 +76,6 @@ func Cluster(logger log.Logger, p providers.Provider, opts *ClusterOptions) erro
...
-	// TODO: move to config validation
-	// validate the name
-	if !validNameRE.MatchString(opts.Config.Name) {
-		return errors.Errorf(
-			"'%s' is not a valid cluster name, cluster names must match `%s`",
-			opts.Config.Name, validNameRE.String(),
-		)
-	}
 	// warn if cluster name might typically be too long
 	if len(opts.Config.Name) > clusterNameMax {
```
knabben, line 80: should this warning be moved to the validation as well?

BenTheElder: maybe, I'm not sure if it makes sense for Validate() to warn, and this is not inherent in the API (we don't enforce this restriction ourselves), it's a limitation of the underlying implementation currently

knabben:
@neolit123 can you take a look at it? thanks

BenTheElder:
sorry I didn't get back to this myself
thank you for the PR
/lgtm
/approve

PR #1987, opened 2020-12-21
Fix podman inspect NetworkSettings.Ports

tossmilestone:
/assign @aojea

### DIFF
--- a/pkg/cluster/internal/providers/podman/provider.go
+++ b/pkg/cluster/internal/providers/podman/provider.go
@@ -196,7 +196,7 @@ func (p *provider) GetAPIServerEndpoint(cluster string) (string, error) {
 		cmd := exec.Command(
 			"podman", "inspect",
 			"--format",
-			"{{ json .NetworkSettings.Ports }}",
+			"json {{ .NetworkSettings.Ports }}",
 			n.String(),
 		)
 		lines, err := exec.OutputLines(cmd)


pkg/cluster/internal/providers/podman/provider.go
```diff
@@ -196,7 +196,7 @@ func (p *provider) GetAPIServerEndpoint(cluster string) (string, error) {
 		cmd := exec.Command(
 			"podman", "inspect",
 			"--format",
-			"{{ json .NetworkSettings.Ports }}",
+			"json {{ .NetworkSettings.Ports }}",
```
aojea, line 199: previous format was correct, we want to get the subsection in json format, this is the same syntax that podman uses for testing it 
https://github.com/containers/podman/blob/5c6b5ef34905f40562b518799c35be8d06694e65/test/e2e/inspect_test.go#L440
It´s also how docker documents it https://docs.docker.com/engine/reference/commandline/inspect/#get-a-subsection-in-json-format

tossmilestone: I got error when execute the command:
```
[root@localhost ~]# podman inspect --format="{{ json .NetworkSettings.Ports }}" kind-control-plane
ERRO[0000] Error printing inspect output: template: all inspect:1: function "json" not defined
```
Is it a not supported podman version?

aojea: I've already fixed it in kind
https://github.com/kubernetes-sigs/kind/blob/e36bf401e3d9792956f216855712606b3069d41e/pkg/cluster/internal/providers/podman/provider.go#L254

And podman has fixed it upstream recently.

You have to use kind from master

tossmilestone: Thanks @aojea

Issue #1969, opened 2020-12-08
Kubelet: invalid configuration: cgroup-root ["kubelet"] doesn't exist


aojea:
**What happened**:

Running a node-image built with kind `92f54895`, in github actions, using kind 0.9.0 makes kubelet fail to start with the following error:

>  Dec 07 19:16:44 ovn-control-plane kubelet[994]: F1207 19:16:44.170173     994 server.go:269] failed to run Kubelet: invalid configuration: cgroup-root ["kubelet"] doesn't exist
Dec 07 19:16:44 ovn-control-plane kubelet[994]: goroutine 1 [running]:

**How to reproduce it (as minimally and precisely as possible)**:

I've tried to reproduced it locally without success:
```
./kind.v0.9.0 create cluster --image aojea/kindnode:kind92f54895_k8se1c617a88ec
```

The failing job with the errors and logs is here
https://github.com/aojea/ovn-kubernetes/runs/1512853588?check_suite_focus=true

Bear in mind that github actions have its own docker and we had issues with KIND running there before
```
Client:
 Debug Mode: false
 Plugins:
  buildx: Build with BuildKit (Docker Inc., 0.4.2+azure)

Server:
 Containers: 3
  Running: 3
  Paused: 0
  Stopped: 0
 Images: 22
 Server Version: 19.03.13+azure
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 8fba4e9a7d01810a393d5d25a3621dc101981175
 runc version: ff819c7e9184c13b7c2607fe6c30ae19403a7aff
 init version: 
 Security Options:
  apparmor
  seccomp
   Profile: default
 Kernel Version: 5.4.0-1031-azure
 Operating System: Ubuntu 18.04.5 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 2
 Total Memory: 6.791GiB
 Name: fv-az183-477
[...]

PR #1995, opened 2020-12-29
Fix entrypoint cgroups mounts

aojea:
The cgroups mount points are different depending on the provider or the platform where kind is running.
We need to be able to handle this different mountpoints, or kubelet will fail to start.
Fixes: #1969

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -1,0 +1,94 @@
+name: Docker
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  docker:
+    name: Docker
+    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


aojea:
@BenTheElder it is working now :smile: https://github.com/kubernetes-sigs/kind/pull/1995/checks?check_run_id=1627892273

BenTheElder:
Excellent! I'll take a closer look soon, popping my head up today but I'm back on the 5th.

BenTheElder:
rebased, updated base image bump, etc.
tests look good so far, looks reasonable enough to me for now.
/lgtm
/approve

AkihiroSuda:
This change seems to broke cgroup v2 ( https://github.com/kubernetes-sigs/kind/issues/2013)

PR #1996, opened 2021-01-04
document loadbalancer

micarlise:


aojea:
/ok-to-test
I think that @BenTheElder wanted to document the limitations for the loadbalancer in Windows and Mac too

> On mac/windows we should explain the limitations (namely that you can't currently access it cleanly from the host due to docker containers not being routable, only via port forwards), and point to alternatives (like the ingress guide).

BenTheElder:
preview: https://deploy-preview-1996--k8s-kind.netlify.app/

sorry I've been unavailable-ish, catching up on PRs today.

site/content/docs/user/loadbalancer.md
```diff
@@ -0,0 +1,100 @@
...
+    identifier: "user-loadbalancer"
+    weight: 3
+description: |-
+    This guide covers how to get service of type LoadBalancer working in a kind cluster using [Metallb].
+
+    This guide complements metallb [installation docs], and sets up metallb using layer2 protocol.  For other protocols check metallb [configuration docs].
+
+    With Docker on Linux, you can send traffic directly to the loadbalancer's external IP if the IP space is within the docker IP space.  
+    
+    On macOS and Windows, docker does not expose the docker network to the host.  Because of this limitation, loadbalancers are only accessable via port fowards.  You may want to check out the [Ingress Guide] as a cross-platform workaround.  You can also expose pods and services using extra port mappings as shown in the extra port mappings section of the [Configuration Guide].
```
BenTheElder, line 15: I think we should clarify this slightly, something like :

> Because of this limitation, loadbalancers are only accessable via port fowards.

=>

>  Because of this limitation, containers (including kind nodes) are only reachable from the host via port-forwards, however other containers / pods can reach other things running in docker including loadbalancers.

Technically loadbalancers work on docker desktop, they're just not _easily_ reachable _from the host_, for some approaches to testing this is acceptable.

BenTheElder:
thank you so much for this PR, I have one request for clarification.

/approve
I'm going to leave it approved so anyone can LGTM it either when that clarification is made or so we can go ahead and merge it and clarify as a follow-up if you don't have time to get to it first.

Thank you!

site/content/docs/user/loadbalancer.md
micarlise: Thanks for the feedback.  I updated the wording per your suggestion.

PR #2011, opened 2021-01-20
bump node image

BenTheElder:
following substantial changes to entrypoint, so we can better test this


BenTheElder:
following https://github.com/kubernetes-sigs/kind/pull/1995 most recently

PR #2014, opened 2021-01-20
entrypoint: fix regression for cgroup v2

### LINKED ISSUE #2013: regression: entrypoint is broken on cgroup v2
**What happened**:

`fix_cgroup()` change in https://github.com/kubernetes-sigs/kind/pull/1995/files broke cgroup v2

It should work

Minimal repro:
```bash
$ docker build -t kindest/base:latest ./images/base
$ docker run --rm --privileged kindest/base
stat: cannot read file system information for '/kind': No such file or directory
INFO: ensuring we can execute mount/umount even with userns-remap
INFO: remounting /sys read-only
INFO: making mounts shared
INFO: fix cgroup mounts for all subsystems
$ echo $?
1
```

**Environment:**

- kind version: (use `kind version`): 26637e29987e1759f0bfb6025f335396960c8565 (Jan 19, 2021)
- Ubuntu 20.10, cgroup v2
- moby/moby@0456e058d2, containerd/containerd@092f9e607a0fe430d6daa4dd5d2be9231bf4966e, opencontainers/runc@1a9011689467283662592a54e1e544a310c03967

AkihiroSuda:
v2 support for `fix_cgroup()` was broken: #2013

As CgroupNS is enabled by default on v2 hosts, we do not need to
mess around the cgroup mounts.

However, at least we need to create "/kubelet" cgroup (#1969).

Fix #2013

### DIFF
--- a/images/base/files/etc/systemd/system/kubelet.service.d/10-kubeadm.conf
+++ b/images/base/files/etc/systemd/system/kubelet.service.d/10-kubeadm.conf
@@ -8,5 +8,8 @@ EnvironmentFile=-/var/lib/kubelet/kubeadm-flags.env
 # This is a file that the user can use for overrides of the kubelet args as a last resort. Preferably, the user should use
 # the .NodeRegistration.KubeletExtraArgs object in the configuration files instead. KUBELET_EXTRA_ARGS should be sourced from this file.
 EnvironmentFile=-/etc/default/kubelet
+# On cgroup v1, the /kubelet cgroup is created in the entrypoint script before running systemd.
+# On cgroup v2, the /kubelet cgroup is created here. (See the comments in the entrypoint script for the reason.)
+ExecStartPre=/bin/sh -euc "if [ -f /sys/fs/cgroup/cgroup.controllers ]; then create-kubelet-cgroup-v2; fi"
 ExecStart=
 ExecStart=/usr/bin/kubelet $KUBELET_KUBECONFIG_ARGS $KUBELET_CONFIG_ARGS $KUBELET_KUBEADM_ARGS $KUBELET_EXTRA_ARGS --cgroup-root=/kubelet
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -83,6 +83,26 @@ mount_kubelet_cgroup_root() {
 }
 
 fix_cgroup() {
+  if [[ -f "/sys/fs/cgroup/cgroup.controllers" ]]; then
+    echo 'INFO: detected cgroup v2'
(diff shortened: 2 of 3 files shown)


images/base/files/usr/local/bin/entrypoint
```diff
@@ -83,7 +83,34 @@ mount_kubelet_cgroup_root() {
 }
 
 fix_cgroup() {
-  echo 'INFO: fix cgroup mounts for all subsystems'
+  if [[ -f "/sys/fs/cgroup/cgroup.controllers" ]]; then
+    echo 'INFO: detected cgroup v2'
+    # Both Docker and Podman enable CgroupNS on cgroup v2 hosts by default.
```
BenTheElder, line 88: thank you for leaving the detailed comments!

BenTheElder:
LGTM

AkihiroSuda:
WIP. The current revision has a bug that moves the systemd process tree from `/` group to `/init` group.

AkihiroSuda:
Updated. The current revision works fine with Kubernetes 1.20.2.

AkihiroSuda:
cc @giuseppe @mrunalp FYI

images/base/files/etc/systemd/system/kubelet.service.d/10-kubeadm.conf
```diff
@@ -8,5 +8,8 @@ EnvironmentFile=-/var/lib/kubelet/kubeadm-flags.env
 # This is a file that the user can use for overrides of the kubelet args as a last resort. Preferably, the user should use
 # the .NodeRegistration.KubeletExtraArgs object in the configuration files instead. KUBELET_EXTRA_ARGS should be sourced from this file.
 EnvironmentFile=-/etc/default/kubelet
+# On cgroup v1, the /kubelet cgroup is created in the entrypoint script before running systemd.
+# On cgroup v2, the /kubelet cgroup is created here. (See the comments in the entrypoint script for the reason.)
+ExecStartPre=/bin/sh -euc "if [ -f /sys/fs/cgroup/cgroup.controllers ]; then create-kubelet-cgroup-v2; fi"
```
BenTheElder, line 13: is this script re-entrant? it tentatively looks like it is not.
under kubeadm kubelet will crash and restart many times before the config is written and it finally starts

AkihiroSuda: Only the first execution of `create-kubelet-cgroup-v2` is meaningful. So, I don't see any issue about restarting.

BenTheElder:
/lgtm
/approve
thanks for the prompt diagnosis, fix, and detailed comments! 😄

PR #2020, opened 2021-01-21
Added release builder

micahhausler:
The release builder fetches pre-built Kubernetes artifacts for a node image from a standard upstream release URL.

**Reviewer notes**: 

* I modified the `New{Docker,Bazel,Release}Builder()` signature to include the `releaseUrl` parameter. This parameter is unused in Docker/Bazel, but the `kubeRoot` param is unused in the ReleaseBuilder. I did this to simplify the diff so that `nameToImpl` could just add `"release"` as one more item in the switch statement.  Let me know if a larger refactor is preferred.
* I did not add any tests, as the rest of the package doesn't include any. I'd be happy to add tests if desired

### DIFF
--- a/pkg/build/nodeimage/build.go
+++ b/pkg/build/nodeimage/build.go
@@ -17,6 +17,7 @@ limitations under the License.
 package nodeimage
 
 import (
+	"net/url"
 	"runtime"
 
 	"sigs.k8s.io/kind/pkg/build/nodeimage/internal/kube"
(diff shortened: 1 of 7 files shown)


BenTheElder:
First: thanks for the PR! 😄 

---

> I modified the New{Docker,Bazel,Release}Builder() signature to include the releaseUrl parameter. This parameter is unused in Docker/Bazel, but the kubeRoot param is unused in the ReleaseBuilder. I did this to simplify the diff so that nameToImpl could just add "release" as one more item in the switch statement. Let me know if a larger refactor is preferred.

I've been thinking about this for v0.11.0 and I think we want to support this but not quite like this:

- I think we should deprecate the `--kube-root` flag and take the source spec as an argument instead which seems more natural and can be passed to any of the builders (also we prefer not to have disjoint flags in the CLI, you won't really find any currently)
- We should be considering multi-platform (can wait until after this PR)
- We should support another kind of spec that allows more control over the node image than just using a release URL, I think a small YAML spec specifing the sources for built-in manifests, images, etc., I've been drafting a design for this.

Most of these are not a blocker for adding this, but I think instead of adding a new flag we should go ahead and deprecate kube-root in favor of passing through the arguments to leave room for these additional mode without a proliferation of mode-specific flags.

Most of the modes should be able to translate a string specifying the source location (kubernetes source dir, yaml nodeImage spec filepath, or remote release URI), with some code to detect which the argument is.

The flag controlling build mode for source builds can go away with https://github.com/kubernetes/kubernetes/issues/88553 (I haven't mentioned this yet, but I'm actively drafting a concrete proposal for this before circling back to #166 / #381  ...)

aside: we prefer to discuss design for features in issues before moving to pull requests, check out our detailed new contributor guide for this and more 🙃  https://kind.sigs.k8s.io/docs/contributing/getting-started/

> I did not add any tests, as the rest of the package doesn't include any. I'd be happy to add tests if desired

I'm not sure how we want to approach this, we've been able to get pretty good coverage of these because we build from source in CI, I think perhaps at least one of the github actions jobs should move to using this build from release to ensure coverage.

BenTheElder:
this currently fails `make verify` which will need to be fixed. https://kind.sigs.k8s.io/docs/contributing/development/#linting

micahhausler:
I'm happy to close this out and come up with a better design first. This was a quick POC, and I'm glad to hear you're already thinking about grabbing pre-built artifacts from somewhere else.

PR #2028, opened 2021-01-22
Add docs for  "Couldn't find an alternative telinit implementation to…

OhMyMndy:
When trying to run `kind` I ran into the following error:

`Couldn't find an alternative telinit implementation to spawn`

Which was because I had `"init":true` in my `/etc/docker/daemon.json` which causes problems with `/sbin/init`

### DIFF
--- a/site/content/docs/user/known-issues.md
+++ b/site/content/docs/user/known-issues.md
@@ -231,9 +231,10 @@ or try to use `sudo` before your commands (if you get `command not found` please
 
 ## Docker init daemon config
 
-Please make sure that when you use `kind`, do not have `"init": true` in your `/etc/docker/daemon.json` because that will
-causes `/sbin/init` to show the following cryptic message *Couldn't find an alternative telinit implementation to spawn*.
-This has to to with `/sbin/init` not running as process id 1. 
+Please make sure that when you use `kind`, you can't have `"init": true` in your `/etc/docker/daemon.json` because that will
+cause `/sbin/init` to show the following cryptic message *Couldn't find an alternative telinit implementation to spawn*.
+This has to to with `/sbin/init` not running as process id 1.
+
 ## Windows Containers
 
 [Docker Desktop for Windows][docker desktop for windows] supports running both Linux (the default) and Windows Docker containers.
--- a/site/content/docs/user/known-issues.md
+++ b/site/content/docs/user/known-issues.md
@@ -37,6 +37,7 @@ description: |-
 * [AppArmor](#apparmor) (may break things, consider disabling)
 * [IPv6 Port Forwarding](#ipv6-port-forwarding) (docker doesn't seem to implement this correctly)
 * [Fedora 32 Firewalld](#fedora32-firewalld) (nftables + docker broken, switch to iptables)
+* [Couldn't find an alternative telinit implementation to spawn](#docker-init-daemon-config)
 
 ## Kubectl Version Skew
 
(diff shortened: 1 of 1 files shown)


site/content/docs/user/known-issues.md
```diff
@@ -227,6 +228,12 @@ open /home/user/.docker/config.json: permission denied
 To fix this problem, either follow the docker's docs [manage docker as a non root user][manage docker as a non root user],
 or try to use `sudo` before your commands (if you get `command not found` please check [this comment about sudo with kind][sudo with kind]).
 
+
+## Docker init daemon config
+
+Please make sure that when you use `kind`, do not have `"init": true` in your `/etc/docker/daemon.json` because that will
+causes `/sbin/init` to show the following cryptic message *Couldn't find an alternative telinit implementation to spawn*.
+This has to to with `/sbin/init` not running as process id 1.
```
aojea, line 236: `to to`

site/content/docs/user/known-issues.md
```diff
@@ -227,6 +228,12 @@ open /home/user/.docker/config.json: permission denied
 To fix this problem, either follow the docker's docs [manage docker as a non root user][manage docker as a non root user],
 or try to use `sudo` before your commands (if you get `command not found` please check [this comment about sudo with kind][sudo with kind]).
 
+
+## Docker init daemon config
+
+Please make sure that when you use `kind`, do not have `"init": true` in your `/etc/docker/daemon.json` because that will
```
BenTheElder, line 234: TIL you can even do this, I'm looking at the `docker run` docs and I don't see a way to opt back out of this when running a particular container (though maybe an explicit `--init=false`?). If it is possible to disable this per-container, then we should update kind to do this and avoid the issue altogether.

BenTheElder:
Thanks for the PR! TIL you can enable this at the daemon level instead of `docker run --init`.
docker seems to accept `docker run --init=false`, but I'm not sure if that would disable this feature for a particular container. If it does we can add this to how we create the nodes.

site/content/docs/user/known-issues.md
OhMyMndy: I tested this and that seems to work!

- Enable `"init": true` in `/etc/docker/daemon.json`
- Run `docker run --rm -it --init alpine:3 ps aux ` -> should be pid 7 or something
- Run `docker run --rm -it --init=false alpine:3 ps aux ` -> should be pid 1

OhMyMndy: Screenshot_2021-02-04_16-55-33

aojea: @OhMyMndy I think that what Ben means is, that instead of fixing the documentation, you add the `--init=false` flag here https://github.com/kubernetes-sigs/kind/blob/e595a85bc454e88c5a0c12201bbeaab9fba51b19/pkg/cluster/internal/providers/docker/provision.go#L216-L241 
If the flag overrides the config, KIND will always create the containers the right way

OhMyMndy: I understand, therefore I tried that solution and it is working for me.

aojea: don´t you want to submit it to kind ?
this way, we don´t need the doc change

BenTheElder: > @OhMyMndy I think that what Ben means is, that instead of fixing the documentation, you add the --init=false flag here

Well, perhaps both. It would be nice to fix this so users don't have to change this setting in their docker settings, but it will take some time for the upgraded kind to filter out so we may still want to add a known-issues entry with the workaround.

We can also add this patch to kind ourselves, I just wanted to know if we can fix it without continuing to require you to change the docker settings (which seems to be so 👍 ).

site/content/docs/user/known-issues.md
aojea: This has a typo @OhMyMndy

OhMyMndy: Fixed

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/2064 will fix it so that this is not necessary in future kind versions, thanks so much!

PR #2032, opened 2021-01-25
Add script to update release version in documentation

sestegra:
To ease the update of the documentation, I suggest to add `site` rule on Makefile with related script.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,6 +97,9 @@ generate:
 # gofmt
 gofmt:
 	hack/make-rules/update/gofmt.sh
+# update documentation site
+site:
+	hack/make-rules/update/site.sh
 ################################################################################
 # ================================== Linting ===================================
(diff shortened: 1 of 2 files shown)


BenTheElder:
Thanks for the PR! ❤️ 
I appreciate the help automating this, we should have done that already 🙃 

I think what we should _probably_ take a slightly different approach though using hugo data + a shortcode, or a shortcode reading from the site's git metadata to do this, `sed` of semver in the site is probably going to collide with other versions eventually. At minimum just a simple shortcoode with the version defined there would let us target a well known location.

I also think we should probably keep the site in it's own makefile for now, `make -C site build` for example, so if we do add a make target to do this it should probably be in `site/Makefile`.

BenTheElder:
(see prev comment, thanks again!)

PR #2033, opened 2021-01-26
Added SHA256 checksums to node base image downloads

### LINKED ISSUE #2030: Support configurability in base image sources
## What would you like to be added

The kind node base image today trusts multiple independent supply chains. Currently the [`Dockerfile`](https://github.com/kubernetes-sigs/kind/blob/v0.10.0/images/base/Dockerfile) is based on Docker hub's Ubuntu container image ([`FROM ubuntu:20.10`](https://github.com/kubernetes-sigs/kind/blob/v0.10.0/images/base/Dockerfile#L22)), uses [Canonical's packages](https://github.com/kubernetes-sigs/kind/blob/v0.10.0/images/base/Dockerfile#L67-L72), and pulls binaries of [containerd](https://github.com/kubernetes-sigs/kind/blob/v0.10.0/images/base/Dockerfile#L86), [crictl](https://github.com/kubernetes-sigs/kind/blob/v0.10.0/images/base/Dockerfile#L97), and [cni](https://github.com/kubernetes-sigs/kind/blob/v0.10.0/images/base/Dockerfile#L103) from GitHub releases. 

### deb packages

I am not proposing any change today. Future changes could allow specifying an alternate package repository.

### base image
I'd like to make the `FROM` argument configurable

```Dockerfile
ARG BASE_IMAGE=ubuntu:20.10
FROM $BASE_IMAGE
```

### binaries
I'd like to make the the CNI, crictl, and containerd tarball URLs configurable. So something roughly like

```Dockerfile
ARG CONTAINERD_SHA256SUM=f42a4a943ef387914a66b8839ee0ab885513f2ea65cbf30b75bff4ccbb4ec72a
ARG CONTAINERD_URL=https://github.com/kind-ci/containerd-nightlies/releases/download/containerd-${CONTAINERD_VERSION#v}
```

## Why is this needed

I'd like to be able to build a node base image, but be able to specify sources that I trust and not need to maintain a fork of this repo. Note that this is only for the base image, not the final node image

All defaults would remain the same, but builders could specify alternate sources or base images with container build time arguments. Would the maintainers be open to a PR to the Dockerfile with these changes?

micahhausler:
Resolves #2030

* Add checksums and verification to base node image Dockerfile
* Add script to generate checksums Dockerfile args for new version

I ran `make build` locally and all 3 architectures built successfully. I did not build kind clusters with these images, but there aren't any functional changes to test.

If builders were relying on a configurable `CONTAINERD_VERSION` build arg and expecting the tag to include a `v` prefix, they'll need to drop the `v`. Docker build args don't perform bash variable manipulation (ex: `${CONTAINERD_VERSION#v}`).

The kind-ci/containerd-nightlies build artifacts did not contain sha256 checksums for `runc`, so I used the checksums from a download I did myself.  Would generating checksums for `runc` that require a change in the github action file in kind-ci/containerd-nightlies?

Other than that, consumers will not see any difference with this change.

### DIFF
--- a/hack/build/update-base-image-shasums.sh
+++ b/hack/build/update-base-image-shasums.sh
@@ -1,0 +1,76 @@
+#!/usr/bin/env bash
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit -o nounset -o pipefail
+
+CONTAINERD_VERSION="1.4.0-106-gce4439a8"
+
+ARCHITECTURES=(
+    "amd64"
+    "arm64"
+    "ppc64le"
+)
+
+CONTAINERD_DEFAULT_VERSION="1.4.0-106-gce4439a8"
+read -r -p "What version of containerd? (defaults to $CONTAINERD_DEFAULT_VERSION) " CONTAINERD_VERSION
+CONTAINERD_VERSION=${CONTAINERD_VERSION:-${CONTAINERD_DEFAULT_VERSION}}
+
+CNI_DEFAULT_VERSION="v0.9.0"
+read -r -p "What version of CNI? (defaults to $CNI_DEFAULT_VERSION) " CNI_VERSION
+CNI_VERSION=${CNI_VERSION:-${CNI_DEFAULT_VERSION}}
+
+CRICTL_DEFAULT_VERSION="v1.19.0"
+read -r -p "What version of crictl: (defaults to $CRICTL_DEFAULT_VERSION) " CRICTL_VERSION
+CRICTL_VERSION=${CRICTL_VERSION:-${CRICTL_DEFAULT_VERSION}}
+
+echo
+CONTAINERD_BASE_URL="https://github.com/kind-ci/containerd-nightlies/releases/download/containerd-${CONTAINERD_VERSION}"
+for ARCH in "${ARCHITECTURES[@]}"; do
+    CONTAINERD_URL="${CONTAINERD_BASE_URL}/containerd-${CONTAINERD_VERSION}.linux-${ARCH}.tar.gz.sha256sum"
+    SHASUM=$(curl -sSL --retry 5 "${CONTAINERD_URL}" | awk '{print $1}')
+    ARCH_UPPER=$(echo "$ARCH" | tr '[:lower:]' '[:upper:]')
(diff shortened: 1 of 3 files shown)


images/base/Dockerfile
```diff
@@ -65,11 +91,24 @@ RUN echo "Ensuring scripts are executable ..." \
     && chmod +x /usr/local/bin/clean-install /usr/local/bin/entrypoint \
  && echo "Installing Packages ..." \
     && DEBIAN_FRONTEND=noninteractive clean-install \
-      systemd \
```
BenTheElder, line 68: FWIW these were previously organized into "classes" of dependency to help track what we need
- systemd is a base requirement
- conntrack etc are requirements for kubernetes / kubelet
- pigz line is for containerd
- bash etc. line is more or less for kind to use itself

(with the goal being to periodically re-evaluate the set of packages as kubernetes/containerd/... evolve).
If we're going to change this to one per line, we should at least update the comment in the source above (which outlines this)

BenTheElder (commented):
> The kind-ci/containerd-nightlies build artifacts did not contain sha256 checksums for runc, so I used the checksums from a download I did myself. Would generating checksums for runc that require a change in the github action file in kind-ci/containerd-nightlies?

I think so, previously we were using travis-ci but @aojea ported it to github actions.
It looks to me like we're failing to publish artifacts at _all_ currently.

@aojea can you take a look?

I took a quick pass at this PR and generally +1, thanks!

images/base/Dockerfile
micahhausler: Ah thanks, I missed that. I'll revert

hack/build/update-base-image-shasums.sh
```diff
@@ -0,0 +1,76 @@
...
+CONTAINERD_BASE_URL="https://github.com/kind-ci/containerd-nightlies/releases/download/containerd-${CONTAINERD_VERSION}"
+for ARCH in "${ARCHITECTURES[@]}"; do
+    CONTAINERD_URL="${CONTAINERD_BASE_URL}/containerd-${CONTAINERD_VERSION}.linux-${ARCH}.tar.gz.sha256sum"
+    SHASUM=$(curl -sSL --retry 5 "${CONTAINERD_URL}" | awk '{print $1}')
+    ARCH_UPPER=$(echo "$ARCH" | tr '[:lower:]' '[:upper:]')
+    echo "ARG CONTAINERD_${ARCH_UPPER}_SHA256SUM=${SHASUM}"
+done
+
+echo
+# TODO (micahhausler): Once kind-ci/containerd-nightlies adds .sha256sum files, just fetch those
```
aojea, line 48: I´m working on this, I hope to get it working today/tomorrow

aojea: and here we go, we have shadsums for runc now https://github.com/kind-ci/containerd-nightlies/releases/tag/containerd-1.5.0-beta.0-36-g19ee068f9

micahhausler: I'm hesitant to change the version of containerd with this PR, would you like me to update the script once a newer containerd is used in kind?

aojea: ahh, yeah, sorry I didn´t mean to change it, just  wanted to check that the new sha5sum are right and works so we don´t have to change anything later

BenTheElder:
/lgtm
/approve
Thank you!

PR #2043, opened 2021-01-29
kindnetd don't wait for kube-proxy

aojea:
We can override the apiserver address used for incluster config, that is the service address, with the internal apiserver address, so we don´t have to wait for  kube-proxy to install the iptables//ipvs rules

### DIFF
--- a/images/kindnetd/cmd/kindnetd/main.go
+++ b/images/kindnetd/cmd/kindnetd/main.go
@@ -20,7 +20,9 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"net"
 	"os"
+	"syscall"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
@@ -28,7 +30,11 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
-	"k8s.io/utils/net"
+	utilsnet "k8s.io/utils/net"
+)
+
+const (
+	probeTCPtimeout = 1 * time.Second
 )
 
 // kindnetd is a simple networking daemon to complete kind's CNI implementation
@@ -40,6 +46,7 @@ import (
 // - HOST_IP: should be populated by downward API
 // - POD_IP: should be populated by downward API
 // - CNI_CONFIG_TEMPLATE: the cni .conflist template, run with {{ .PodCIDR }}
+// - CONTROL_PLANE_ENDPOINT: control-plane endpoint format host:port
 
 // TODO: improve logging & error handling
 
@@ -54,10 +61,31 @@ func main() {
 	if err != nil {
 		panic(err.Error())
 	}
+
+	// override the internal apiserver endpoint to avoid
+	// waiting for kube-proxy to install the services rules.
+	// If the endpoint is not reachable, fallback the internal endpoint
+	controlPlaneEndpoint := os.Getenv("CONTROL_PLANE_ENDPOINT")
+	if controlPlaneEndpoint != "" {
+		// check that the apiserver is reachable before continue
+		// to fail fast and avoid waiting until the client operations timeout
+		var ok bool
+		for i := 0; i < 5; i++ {
+			ok = probeTCP(controlPlaneEndpoint, probeTCPtimeout)
+			if ok {
+				config.Host = "https://" + controlPlaneEndpoint
+				break
+			}
+			klog.Infof("apiserver not reachable, attempt %d ... retrying", i)
+			time.Sleep(time.Second * time.Duration(i))
+		}
+	}
+	// create the clientset to connect the apiserver
 	clientset, err := kubernetes.NewForConfig(config)
 	if err != nil {
 		panic(err.Error())
 	}
+	klog.Infof("connected to apiserver: %s", config.Host)
 
 	// obtain the host and pod ip addresses
 	// if both ips are different we are not using the host network
@@ -83,7 +111,7 @@ func main() {
 
 	// enforce ip masquerade rules
 	// TODO: dual stack...?
-	masqAgent, err := NewIPMasqAgent(net.IsIPv6String(hostIP), []string{os.Getenv("POD_SUBNET")})
+	masqAgent, err := NewIPMasqAgent(utilsnet.IsIPv6String(hostIP), []string{os.Getenv("POD_SUBNET")})
 	if err != nil {
 		panic(err.Error())
 	}
@@ -193,4 +221,33 @@ func internalIP(node corev1.Node) string {
 		}
 	}
 	return ""
+}
+
+// Modified from agnhost connect command in k/k
+// https://github.com/kubernetes/kubernetes/blob/c241a237f9a635286c76c20d07b103a663b1cfa4/test/images/agnhost/connect/connect.go#L66
+func probeTCP(address string, timeout time.Duration) bool {
+	klog.Infof("probe TCP address %s", address)
+	if _, err := net.ResolveTCPAddr("tcp", address); err != nil {
+		klog.Warningf("DNS problem %s: %v", address, err)
+		return false
+	}
+
+	conn, err := net.DialTimeout("tcp", address, timeout)
+	if err == nil {
+		conn.Close()
+		return true
+	}
+	if opErr, ok := err.(*net.OpError); ok {
+		if opErr.Timeout() {
+			klog.Warningf("TIMEOUT %s", address)
+		} else if syscallErr, ok := opErr.Err.(*os.SyscallError); ok {
+			if syscallErr.Err == syscall.ECONNREFUSED {
+				klog.Warningf("REFUSED %s", address)
+			}
+		}
+		return false
+	}
+
+	klog.Warningf("OTHER %s: %v", address, err)
+	return false
 }
--- a/pkg/build/nodeimage/const_cni.go
+++ b/pkg/build/nodeimage/const_cni.go
@@ -22,9 +22,11 @@ The default CNI manifest and images are our own tiny kindnet
 
 var defaultCNIImages = []string{"kindest/kindnetd:v20210119-d5ef916d"}
 
+// TODO: migrate to fully patching and deprecate the template
 const defaultCNIManifest = `
 # kindnetd networking manifest
 # would you kindly template this file
+# would you kindly patch this file
 ---
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
--- a/pkg/build/nodeimage/const_cni.go
+++ b/pkg/build/nodeimage/const_cni.go
@@ -20,7 +20,7 @@ package nodeimage
 The default CNI manifest and images are our own tiny kindnet
 */
 
-var defaultCNIImages = []string{"kindest/kindnetd:v20210119-d5ef916d"}
+var defaultCNIImages = []string{"kindest/kindnetd:v20210220-5b7e6d01"}
 
 // TODO: migrate to fully patching and deprecate the template
 const defaultCNIManifest = `
@@ -95,7 +95,7 @@ spec:
       serviceAccountName: kindnet
       containers:
       - name: kindnet-cni
-        image: kindest/kindnetd:v20210119-d5ef916d
+        image: kindest/kindnetd:v20210220-5b7e6d01
         env:
         - name: HOST_IP
           valueFrom:
--- a/pkg/cluster/internal/create/actions/installcni/cni.go
+++ b/pkg/cluster/internal/create/actions/installcni/cni.go
@@ -23,8 +23,10 @@ import (
 	"text/template"
 
 	"sigs.k8s.io/kind/pkg/errors"
+	"sigs.k8s.io/kind/pkg/internal/apis/config"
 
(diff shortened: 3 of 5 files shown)


aojea:
/hold
I have to test it, just open it for discussion @BenTheElder 
The other option I see is to mount the kubeconfig file, but this seems easier and we are already passing env variables

munnerz:
Will this cause restarting a cluster after host reboot to stop working? 👀 (ie IP of the control plane container(s) changing)

aojea:
> Will this cause restarting a cluster after host reboot to stop working?  (ie IP of the control plane container(s) changing)

this is a good question, and I honestly didn´t think about it.
We are using the hostname, so it will not be a problem 

```
	}
	// NOTE: we're using the nodes's hostnames which are their names
	return net.JoinHostPort(n.String(), fmt.Sprintf("%d", common.APIServerInternalPort)), nil
}
```

I will add a check to avoid possible issues and make this more reliable

pkg/build/nodeimage/const_cni.go
```diff
@@ -105,6 +105,8 @@ spec:
               fieldPath: status.podIP
         - name: POD_SUBNET
           value: {{ .PodSubnet }}
+        - name: CONTROL_PLANE_ENDPOINT
+          value: {{ .ControlPlaneEndpoint }}
```
BenTheElder, line 109: Note: this is backwards incompatible.

BenTheElder:
Shouldn't the downward API have the node name etc?

aojea:
> Shouldn't the downward API have the node name etc?

we need the controlplane endpoint, or at least one node with the apiserver running

pkg/build/nodeimage/const_cni.go
aojea: in which sense, the ci is passing with this change ... (I´m a bit slow today)

BenTheElder: older kind won't be able to run this image

aojea: :thinking: what will go template do?

BenTheElder: Explode on indexing a field that doesn't exist

aojea:
podman job flake on 
```
ERROR: Untarring "/var/log": error writing to /tmp/kind/logs/kind-control-plane/pods/kube-system_kube-apiserver-kind-control-plane_996ad038fbf1d97d6a715e6947b79946/kube-apiserver/0.log: unexpected EOF: error writing to /tmp/kind/logs/kind-control-plane/pods/kube-system_kube-apiserver-kind-control-plane_996ad038fbf1d97d6a715e6947b79946/kube-apiserver/0.log: unexpected EOF
```

@amwat I vaguely remember that you've solved something similar with podman before, or was it another issue?

pkg/cluster/internal/create/actions/installcni/cni.go
```diff
@@ -122,23 +122,6 @@ func (a *action) Execute(ctx *actions.ActionContext) error {
...
-  "cniVersion": "0.3.1",
-  "name": "kindnet",
-  "plugins": [
-	{ "type": "portmap" }
-  ]
-}`
-
-	if err := node.Command(
-		"cp", "/dev/stdin", "/etc/cni/net.d/10-kindnet.conflist",
-	).SetStdin(strings.NewReader(cniPlaceholder)).Run(); err != nil {
```
aojea: this is interesting @BenTheElder , adding the placeholder the kubelet is ready soon but the pods take more time, etcd takes a lot of time to be ready ... I don't know if there is some logic to give priority to hostNetwork pods or something if there are no network?

aojea:
:thinking: 
```
2021-01-31T15:45:07.399783554Z stderr F W0131 15:45:07.399321       1 main.go:269] REFUSED :
2021-01-31T15:45:07.402689774Z stderr F I0131 15:45:07.402468       1 main.go:97] hostIP = 172.18.0.3
2021-01-31T15:45:07.402726102Z stderr F podIP = 172.18.0.3
2021-01-31T15:45:07.40296515Z stderr F I0131 15:45:07.402774       1 main.go:106] setting mtu 1500 for CNI 
2021-01-31T15:45:07.897770946Z stderr F I0131 15:45:07.897545       1 main.go:189] handling current node
```

images/kindnetd/cmd/kindnetd/main.go
```diff
@@ -54,6 +62,30 @@ func main() {
 	if err != nil {
 		panic(err.Error())
 	}
+
+	// override the internal apiserver endpoint to avoid
+	// waiting for kube-proxy to install the services rules.
+	// If the endpoint is not reachable, fallback the internal endpoint
+	controlPlaneEndpoint := os.Getenv("CONTROL_PLANE_ENDPOINT")
+	if controlPlaneEndpoint != "" && probeTCP(controlPlaneEndpoint, probeTCPtimeout) {
```
BenTheElder, line 70: i'm not sure we should bother probing TCP if we have the env, since we know the env is what we want?

pkg/cluster/internal/create/actions/installcni/cni.go
```diff
@@ -83,11 +86,39 @@ func (a *action) Execute(ctx *actions.ActionContext) error {
...
+	if err != nil {
+		return err
+	}
+
+	patchValue := fmt.Sprintf(`
+- op: add
+  path: /spec/template/spec/containers/0/env/-
+  value:
+    name: CONTROL_PLANE_ENDPOINT
+    value: %s
```
BenTheElder, line 100: this could just be string concat.

images/kindnetd/cmd/kindnetd/main.go
aojea: this is much faster, you don't have to construct the client twice

pkg/cluster/internal/create/actions/installcni/cni.go
aojea: I was trying to make it more readable :shrug:

BenTheElder: IMO it's harder to parse a string format in your head than recognize `"foo" + bar`

BenTheElder: (it's also inefficient)

pkg/cluster/internal/create/actions/installcni/cni.go
```diff
@@ -83,11 +86,39 @@ func (a *action) Execute(ctx *actions.ActionContext) error {
...
+`, controlPlaneEndpoint)
+
+	controlPlanePatch6902 := config.PatchJSON6902{
+		Group:   "apps",
+		Version: "v1",
+		Kind:    "DaemonSet",
+		Patch:   patchValue,
+	}
+
+	patchedConfig, err := patch.KubeYAML(manifest, nil, []config.PatchJSON6902{controlPlanePatch6902})
```
BenTheElder, line 110: TODO: only patch this if it is kindnetd / wants to be patched (similar to the approach with templating previously with the sentinel comment in the yaml)

images/kindnetd/cmd/kindnetd/main.go
BenTheElder: Under what circumstance would you expect the probe to fail and desire falling back?

aojea: - wacky iptables/nftables distros/users
- control plane endpoint being a loadbalancer
- HA deployment and the control plane endpoint is removed
- ...
networking :shrug:

aojea:
ok, this works
```
2021-02-13T21:17:41.592337078Z stderr F I0213 21:17:41.592035       1 main.go:229] probe TCP address kind-control-plane:6443
2021-02-13T21:17:41.685711929Z stderr F I0213 21:17:41.685516       1 main.go:88] connected to apiserver: https://kind-control-plane:6443
```

kind ended at
```
I0213 21:17:49.744334     997 round_trippers.go:454] PATCH https://kind-control-plane:6443/api/v1/nodes/kind-worker?timeout=10s 200 OK in 6 milliseconds
This node has joined the cluster:
* Certificate signing request was sent to apiserver and a response was received.
* The Kubelet was informed of the new secure connection details.
```

and kube-proxy was still not running (I think that is because we path it to increase the verbosity and it has to restar)
```
Feb 13 21:18:03.144: INFO: The status of Pod kube-proxy-h62zb is Pending (Ready = false), waiting for it to be either Running (with Ready = true) or Failed
Feb 13 21:18:03.144: INFO: The status of Pod kube-proxy-tjcnm is Pending (Ready = false), waiting for it to be either Running (with Ready = true) or Failed
```

pkg/build/nodeimage/const_cni.go
```diff
@@ -20,11 +20,12 @@ package nodeimage
 The default CNI manifest and images are our own tiny kindnet
 */
 
-var defaultCNIImages = []string{"kindest/kindnetd:v20210119-d5ef916d"}
+var defaultCNIImages = []string{"aojea/kindnetd:fast"}
```
BenTheElder, line 23: @aojea you can push to the real registry now, just be careful :-)

pkg/build/nodeimage/const_cni.go
```diff
@@ -20,11 +20,12 @@ package nodeimage
...
 
 const defaultCNIManifest = `
 # kindnetd networking manifest
 # would you kindly template this file
```
BenTheElder, line 27: TODO: we should migrate fully to patching in the future and do away with the template crud.
This will break support for older releases without the patching so we should wait some time.

pkg/cluster/internal/create/actions/installcni/cni.go
```diff
@@ -83,6 +85,40 @@ func (a *action) Execute(ctx *actions.ActionContext) error {
...
+		}
+
+		patchedConfig, err := patch.KubeYAML(manifest, nil, []config.PatchJSON6902{controlPlanePatch6902})
+		if err != nil {
+			return err
+		}
+		manifest = patchedConfig
+	}
+
+	ctx.Logger.V(5).Infof("Using the following Kindnetd config on node %s:\n%s", node.String(), manifest)
```
BenTheElder, line 120: the node is pretty irrelevant? we're installing a daemonset.

BenTheElder (commented):
lgtm with nits

images/kindnetd/cmd/kindnetd/main.go
BenTheElder: I mean that's going to break other stuff anyhow 🤔 
I remain skeptical that this complexity is desirable here, I wouldn't expect a production CNI to do this.

aojea: I see CONTROL_PLANE_ENDPOINT as an input variable from the user that we can't trust.
probeTCP is validation, if is invalid it fall-back to the value that we know that works, that is the apiserver service KUBERNETES_HOST:KUBERNETES_PORT

pkg/cluster/internal/create/actions/installcni/cni.go
aojea: heh, what I was thinking

images/kindnetd/cmd/kindnetd/main.go
aojea: and now that I think it, if we don't fall back ,if the cluster restart and the containers change their ips, kindnet will keep trying to connect to the hardcoded value and we'll not be able to connect.

aojea:
/retest

561/external/debian-iptables-amd64/image/000.tar.gz: read tcp 10.61.15.241:32856->172.217.212.128:443: read: connection reset by peer

images/kindnetd/cmd/kindnetd/main.go
BenTheElder: The "user" is us.

It should be receiving a nodename:port, NOT IP.

aojea: > It should be receiving a nodename:port, NOT IP.

oh, yeah, I was thinking it was the external IP, it uses the internalIP that uses dns

aojea: but I don't trust dns either :upside_down_face:

BenTheElder: The docker Edns for the container name should be fine.

BenTheElder: Whereas the fallback is the in cluster VIP from kube-proxy

aojea: this is an old commit by the way, is not the way it is now

images/kindnetd/cmd/kindnetd/main.go
```diff
@@ -54,10 +61,31 @@ func main() {
 	if err != nil {
 		panic(err.Error())
 	}
+
+	// override the internal apiserver endpoint to avoid
+	// waiting for kube-proxy to install the services rules.
+	// If the endpoint is not reachable, fallback the internal endpoint
+	controlPlaneEndpoint := os.Getenv("CONTROL_PLANE_ENDPOINT")
+	if controlPlaneEndpoint != "" {
```
BenTheElder, line 69: this block still seems wrong. if CONTROL_PLANE_ENDPOINT is set we should trust it, kindnetd is not for external usage.
this polling seems likely to wind up the buggy thing blocking startup more than anything.

PR #2044, opened 2021-01-29
WIP: Support for overriding the loadbalancer config for HA clusters

pwittrock:
The load balancing strategy can impact the apiserver performance -- for instance the default loadbalancing approach creates optimistic locking contention on ResourceQuota objects.


pwittrock:
@BenTheElder LMK if I should file an issue or something.  I didn't see an obvious place to test this.  What would you recommend?

aojea:
what is the goal? replacing default haproxy configuration?

pwittrock:
@aojea

> what is the goal? replacing default haproxy configuration?

Yes, the default haproxy doesn't play nice with ResourceQuota under high-contention.  It would be useful to have more control over how traffic is sent to the apiservers.

pwittrock:
Closing this while I make some updates.

PR #2048, opened 2021-01-31
Add a cni placeholder when installing cni

llhuii:
Default kubelet would wait for container runtime network ready to mark
node ready status.

When DisableDefaultCNI is false, kindnet is used for network, then
kubelet would wait kindnet to be ready. Here add a cni placeholder to
avoid this.

### DIFF
--- a/pkg/cluster/internal/create/actions/installcni/cni.go
+++ b/pkg/cluster/internal/create/actions/installcni/cni.go
@@ -91,6 +91,31 @@ func (a *action) Execute(ctx *actions.ActionContext) error {
 		return errors.Wrap(err, "failed to apply overlay network")
 	}
 
+	// To avoid for kubelet waiting for container runtime network ready,
+	// here add a cni placeholder file to the control node to just tell the
+	// container runtime network ready.
+	// at least 30 seconds faster for control node to be ready status.
+	// see https://github.com/kubernetes-sigs/kind/issues/2036.
+
+	cniConflistPlaceHolder := `
+		{
+			"cniVersion": "0.3.0",
+			"name": "kindnet",
+			"plugins": [
+				{
+					"type": "portmap"
+				}
+			]
+		}
+	`
+	cniConflistPlaceHolderFile := "/etc/cni/net.d/kindnet.conflist"
+	// is needed to do this for all ControlPlaneNodes?
+	if err := node.Command(
+		"tee", cniConflistPlaceHolderFile,
+	).SetStdin(strings.NewReader(cniConflistPlaceHolder)).Run(); err != nil {
+		return errors.Wrap(err, "failed to write CNI placeholder file")
+	}
+
 	// mark success
 	ctx.Status.End(true)
 	return nil


llhuii:
with this script
```sh
export PATH=/root/go/src/github.com/kubernetes-sigs/kind/bin:$PATH
name=cni-ph-test
kind delete cluster --name $name
time kind create cluster --name $name
kubectl cluster-info --context kind-$name
time kubectl wait --for=condition=Ready node/$name-control-plane --timeout=90s
```

this can speed up at least 45 seconds for node to be ready status.

aojea:
@llhuii I was doing some tests with this too today, and I've found that the network takes longer, check the jobs, all of them times out to get the pods running with this change

```
Run # wait network is ready
  # wait network is ready
  kubectl wait --for=condition=ready pods --namespace=kube-system -l k8s-app=kube-dns
  kubectl get nodes -o wide
  kubectl get pods -A
  shell: /bin/bash -e {0}
  env:
    JOB_NAME: docker-singleNode-ipv4
    IP_FAMILY: ipv4
error: no matching resources found
Error: Process completed with exit code 1.
```

llhuii:
ok, wait for your analysis about the failing checks.

BenTheElder:
/ok-to-test
I'm pretty sure this will just force all the core pods using CNI to be recreated which is not what we actually want, versus preventing the hang on startup.

aojea:
> /ok-to-test
> I'm pretty sure this will just force all the core pods using CNI to be recreated which is not what we actually want, versus preventing the hang on startup.

I couldn´t investigate deeper, my observation is that the `etcd` pod is running but not ready for a long time (minutes)

llhuii:
Hi, here are my analysis:
#### for `no matching resources found` error cases:
in `Docker (ipv4, singleNode)` test-case.
```
2021-01-31T17:42:27.7167382Z ##[group]Run # wait network is ready
2021-01-31T17:42:27.7167974Z # wait network is ready 
2021-01-31T17:42:27.7168714Z sudo kubectl wait --for=condition=ready pods --namespace=kube-system -l k8s-app=kube-dns
2021-01-31T17:42:27.7169455Z sudo kubectl get nodes -o wide 
2021-01-31T17:42:27.7169954Z sudo kubectl get pods -A 
2021-01-31T17:42:27.7211859Z shell: /bin/bash -e {0}
2021-01-31T17:42:27.7212264Z env:
2021-01-31T17:42:27.7212784Z   JOB_NAME: podman-singleNode-ipv4
2021-01-31T17:42:27.7213910Z   KIND_EXPERIMENTAL_PROVIDER: podman
2021-01-31T17:42:27.7214625Z   IP_FAMILY: ipv4
2021-01-31T17:42:27.7215060Z   PODMAN_VERSION: stable
2021-01-31T17:42:27.7215524Z ##[endgroup]
2021-01-31T17:42:27.9477472Z error: no matching resources found
2021-01-31T17:42:27.9507058Z ##[error]Process completed with exit code 1.
```
**error: no matching resources found** was reported by cmd ` kubectl wait --for=condition=ready pods --namespace=kube-system -l k8s-app=kube-dns`.
this means no label `k8s-app=kube-dns` pod existed, i.e. no coredns pod was created when kubectl waited.

this happend when node became ready before coredns pod was created by kube-controller-manager.

#### for `timed-out` failed cases:
in e.g. `Cgroup v2 (docker)` cases
```
2021-01-31T17:49:11.5568350Z pod/coredns-74ff55c5b-q5nqb condition met
2021-01-31T17:49:11.5571220Z error: timed out waiting for the condition on pods/coredns-74ff55c5b-88zcb
```
was also reported by cmd ` kubectl wait --for=condition=ready pods --namespace=kube-system -l k8s-app=kube-dns`.

this said one coredns pod was ok, another timed out.

this happened when node became ready too early leading to 45 seconds extra to kubectl-wait comparing the case without this pr.

llhuii:
So in my opinion, this kubelet-wait just after node-ready is not seriously ok.
https://github.com/kubernetes-sigs/kind/blob/3947e5b0923980f7d7a70707a79d870cedd46233/.github/workflows/docker.yaml#L64-L67

llhuii:
FYI, only with this cni placeholder, kubelet logs `failed to setup network for sandbox...: must be called as chained plugin` when creating coredns-pod.

detailed log from kubelet:






```
Feb 03 12:22:00 o-control-plane kubelet[651]: E0203 12:22:00.641898     651 remote_runtime.go:116] RunPodSandbox from runtime service failed: rpc error: code = Unknown desc = failed to setup network for sandbox "0b001f9ec48f71f6decc774cf8e7e17a38ad70fe149ac5469c0442b554f8c3be": must be called as chained plugin
Feb 03 12:22:00 o-control-plane kubelet[651]: E0203 12:22:00.641938     651 kuberuntime_sandbox.go:70] CreatePodSandbox for pod "coredns-74ff55c5b-rd4ft_kube-system(01b61c7d-9d36-4912-a4e3-834c4e244ddf)" failed: rpc error: code = Unknown desc = failed to setup network for sandbox "0b001f9ec48f71f6decc774cf8e7e17a38ad70fe149ac5469c0442b554f8c3be": must be called as chained plugin
Feb 03 12:22:00 o-control-plane kubelet[651]: E0203 12:22:00.641951     651 kuberuntime_manager.go:755] createPodSandbox for pod "coredns-74ff55c5b-rd4ft_kube-system(01b61c7d-9d36-4912-a4e3-834c4e244ddf)" failed: rpc error: code = Unknown desc = failed to setup network for sandbox "0b001f9ec48f71f6decc774cf8e7e17a38ad70fe149ac5469c0442b554f8c3be": must be called as chained plugin
Feb 03 12:22:00 o-control-plane kubelet[651]: E0203 12:22:00.641991     651 pod_workers.go:191] Error syncing pod 01b61c7d-9d36-4912-a4e3-834c4e244ddf ("coredns-74ff55c5b-rd4ft_kube-system(01b61c7d-9d36-4912-a4e3-834c4e244ddf)"), skipping: failed to "CreatePodSandbox" for "coredns-74ff55c5b-rd4ft_kube-system(01b61c7d-9d36-4912-a4e3-834c4e244ddf)" with CreatePodSandboxError: "CreatePodSandbox for pod \"coredns-74ff55c5b-rd4ft_kube-system(01b61c7d-9d36-4912-a4e3-834c4e244ddf)\" failed: rpc error: code = Unknown desc = failed to setup network for sandbox \"0b001f9ec48f71f6decc774cf8e7e17a38ad70fe149ac5469c0442b554f8c3be\": must be called as chained plugin"
```

aojea:
> this happened when node became ready too early leading to 45 seconds extra to kubectl-wait comparing the case without this pr.

if the pods for dns are not ready the cluster is not ready :) , why do you think is wrong that check?



> FYI, only with this cni placeholder, kubelet logs `failed to setup network for sandbox...: must be called as chained plugin` when creating coredns-pod.

ah, so the problem is that kubelet tries to create pods because of the placeholder, and it fails because the cni is not really ready (is the placeholder) ...  that seems that the placeholder is not a good idea, having kubelet reporting that the node is ready and trying to create pods when it can´t doesn´t seem a good idea to me

llhuii:
> ah, so the problem is that kubelet tries to create pods because of the placeholder, and it fails because the cni is not really ready (is the placeholder) ... that seems that the placeholder is not a good idea, having kubelet reporting that the node is ready and trying to create pods when it can´t doesn´t seem a good idea to me

Agree as I said in https://github.com/kubernetes-sigs/kind/pull/2048#issuecomment-773135798

llhuii:
> if the pods for dns are not ready the cluster is not ready :) , why do you think is wrong that check?

sorry to mislead you, this check is ok. what's I want to say is that check immediately after node-ready is not ok.

aojea:
my apologies, was reading in my email and I miss some of the comment
Thanks @llhuii for the great analysis

llhuii:
It doesn't matter😄

PR #2059, opened 2021-02-06
set container env for podman nodes

BenTheElder:
this is mostly aesthetic, it may make debugging slightly saner someday, but it shouldn't functionally change anything.

### DIFF
--- a/pkg/cluster/internal/providers/podman/provision.go
+++ b/pkg/cluster/internal/providers/podman/provision.go
@@ -143,6 +143,8 @@ func commonArgs(cfg *config.Cluster, networkName string) ([]string, error) {
 		"--net", networkName, // attach to its own network
 		// label the node with the cluster ID
 		"--label", fmt.Sprintf("%s=%s", clusterLabelKey, cfg.Name),
+		// specify container implementation to systemd
+		"-e", "container=podman",
 	}
 
 	// enable IPv6 if necessary


pkg/cluster/internal/providers/podman/provision.go
```diff
@@ -143,6 +143,8 @@ func commonArgs(cfg *config.Cluster, networkName string) ([]string, error) {
 		"--net", networkName, // attach to its own network
 		// label the node with the cluster ID
 		"--label", fmt.Sprintf("%s=%s", clusterLabelKey, cfg.Name),
+		// specify container implementation to systemd
+		"-e", "container=podman",
```
aojea, line 147: I couldn't find this documented, what is the use of this environment variable `containre`?

BenTheElder: systemd, https://systemd.io/CONTAINER_INTERFACE/
we set `container=docker` in the dockerfile, which is a fine fallback (we're just required to set _something_ descriptive), but it makes sense to override this in podman so if you debug this it shows podman instead of docker.

as mentioned in the PR body, the value isn't actually important.

BenTheElder:
podman seems to be flaking a lot on exporting logs in CI.

pkg/cluster/internal/providers/podman/provision.go
aojea: TIL, it also fixes the macine-uuid thing

aojea:
> podman seems to be flaking a lot on exporting logs in CI.

didn't amwat fixed something related some time ago?

BenTheElder:
we had some issues with how we did log collecting and parallelism etc., but I don't remember a podman specific fix, and I'm seeing it happen the most in the podman CI like https://github.com/kubernetes-sigs/kind/pull/2059/checks?check_run_id=1844418440

PR #2060, opened 2021-02-06
retry vagrant up

BenTheElder:
basically https://github.com/containerd/containerd/pull/4999

saw a similar issue here https://github.com/kubernetes-sigs/kind/pull/2059/checks?check_run_id=1844399073 (not surprising)


aojea:
I'm going to override since 1.19 job is totally in unrelated to the patch

BenTheElder:
1.19 failure is also bazel image puller flaking.

PR #2064, opened 2021-02-09
explicitly opt nodes out of tini

BenTheElder:
fixes the problem documented in https://github.com/kubernetes-sigs/kind/pull/2028


BenTheElder:
wow all green CI on the first try! it's a miracle! 🎉

PR #2066, opened 2021-02-10
Fix init-buildx.sh

houfangdong:
The first bash being overwritten does not work, just leave one.
See discussion for details: [bash compare](https://stackoverflow.com/questions/16365130/what-is-the-difference-between-usr-bin-env-bash-and-usr-bin-bash)

### DIFF
--- a/hack/build/init-buildx.sh
+++ b/hack/build/init-buildx.sh
@@ -1,4 +1,3 @@
-#!/bin/bash
 #!/usr/bin/env bash
 # Copyright 2020 The Kubernetes Authors.
 #


BenTheElder:
nice catch!
/approve
/lgtm

houfangdong:
Hi, @BenTheElder
The podman of ci seems to have a problem, but it has nothing to do with this pr, and the **retest**  doesn't work anymore. I don't care about it, right?

aojea:
> Hi, @BenTheElder
> The podman of ci seems to have a problem, but it has nothing to do with this pr, and the **retest** doesn't work anymore. I don't care about it, right?

yeah, is a known issue, don't worry about ut

aojea:
Overriding podman job

Issue #2067, opened 2021-02-11
Clarification on whether Docker will always be required


mattysweeps:
The Kind homepage contains this banner.
```
Dockershim deprecation does NOT impact kind.  🐳
If you already use kind you've actually been testing your workloads on containerd!

While kind uses docker or podman on your host, it uses CRI / containerd "inside" the nodes and does not use dockershim.
```

However, creating a cluster without the docker cli results in this error message
```
$ kind create cluster                    
ERROR: failed to create cluster: failed to list clusters: command "docker ps -a --filter label=io.x-k8s.kind.cluster=kind --format '{{.Names}}'" failed with error: exec: "docker": executable file not found in $PATH               
Command Output:
```
Which comes from https://github.com/kubernetes-sigs/kind/blob/master/pkg/cluster/internal/providers/docker/provider.go#L97

I believe I have a decent understanding of what dockershim is and why it's being deprecated.
https://kubernetes.io/blog/2020/12/02/dockershim-faq/

My question is:

If kind is performing docker commands, then isn't it still reliant on Docker? I know that's slightly different from dockershim, but couldn't Kind use crictl or containerd's libraries directly to remove the need for docker altogether?

mattysweeps:
Additionally, I think an error should be thrown here instead of defaulting to docker.
https://github.com/kubernetes-sigs/kind/blob/master/pkg/cluster/provider.go#L79

Maybe Kind could have a crictl provider?

BenTheElder:
Dockershim is a component in Kubernetes, kind doesn't need *dockershim*, it does need docker or podman currently.

Kind runs containerd inside the nodes (which are created with podman or docker) and Kubernetes runs against that.

#1369 for a crictl provider.

The error is thrown when podman nor docker are present. Podman is still experimental.

The kind docs specify that docker must be installed. There's no relationship to dockershim though.

BenTheElder:
There hasn't been much interest in a crictl based provider because developers use docker or podman much more than they use say containerd in production. But as mentioned in the note if you run kind you've already run your app "in production" against containerd / CRI instead of dockershim.

Kubernetes is not trying to "kill docker" or anything like that, we have nothing against docker. It's just in production environments for Kubernetes's usage CRI is a more appropriate target and is implemented lower level than docker, and Kubernetes does not want to continue implementing dockershim as a special case built in when containerd can be used instead.

Whereas kind is: not for production and not using dockershim in any way. We are however working to GA the podman support so users have options.

[...]

PR #2069, opened 2021-02-13
add public API to detect node providers

BenTheElder:
see: #2067

### DIFF
--- a/pkg/cluster/provider.go
+++ b/pkg/cluster/provider.go
@@ -22,6 +22,7 @@ import (
 	"sigs.k8s.io/kind/pkg/cluster/constants"
 	"sigs.k8s.io/kind/pkg/cluster/nodes"
 	"sigs.k8s.io/kind/pkg/cluster/nodeutils"
+	"sigs.k8s.io/kind/pkg/errors"
 	"sigs.k8s.io/kind/pkg/log"
 
 	internalcreate "sigs.k8s.io/kind/pkg/cluster/internal/create"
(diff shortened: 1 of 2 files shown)


BenTheElder:
Not having people re-implement this to fail out early when there doesn't seem to be a provider available will be more important in the future when we have detection for some provider that is not a mere "$cli_binary is in path", which we'll likely want to do at some point.

pkg/cluster/provider.go
```diff
@@ -68,20 +69,54 @@ func NewProvider(options ...ProviderOption) *Provider {
...
+			providerOpt = ProviderWithDocker()
 		}
+		providerOpt.apply(p)
 	}
 	return p
 }
 
+// NoNodeProviderDetectedError indicates that we could not autolocate an available
+// NodeProvider backend on the host
+var NoNodeProviderDetectedError = errors.NewWithoutStack("failed to detect any supported node provider")
```
BenTheElder, line 93: we haven't actually had much in the way of client-handle-able errors yet (usually stuff like "cluster bringup failed in some unexpected way"), so I had to update the error utilities to handle this appropriately.

pkg/errors/errors.go
```diff
@@ -26,6 +28,12 @@ func New(message string) error {
 	return pkgerrors.New(message)
 }
 
+// NewWithoutStack is like new but does NOT wrap with a stack
+// This is useful for exported errors
+func NewWithoutStack(message string) error {
+	return stderrors.New(message)
```
BenTheElder, line 34: since this will necessarily be for making errors at the bottom of the stack that should not themselves contain a stack, this should be appropriate (a plain error with the message and nothing else).

BenTheElder:
installing podman is failing in the CI pipeline :|

neolit123 (commented):
/lgtm

BenTheElder:
hmm it seems to get github actions to run with the latest config you must manually rebase the PR.
that's ... not great.

aojea:
/retest
/lgtm

> hmm it seems to get github actions to run with the latest config you must manually rebase the PR.
> that's ... not great.

I´m sufferin gthis every day :upside_down_face:

PR #2071, opened 2021-02-13
try using preinstalled podman

BenTheElder:
/hold
xref: https://github.com/kubernetes-sigs/kind/issues/2070

PR #2073, opened 2021-02-13
use preinstalled podman

aojea:
Following up on https://github.com/kubernetes-sigs/kind/pull/2071/

If podman comes preinstalled, it should have the networking dependencies installed too, so no need to install them again.

### DIFF
--- a/.github/workflows/podman.yml
+++ b/.github/workflows/podman.yml
@@ -46,12 +46,6 @@ jobs:
       - name: Setup podman
         run: |
           podman version
-          # Install network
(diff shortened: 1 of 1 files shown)


aojea:
it went through in multinode and failed the known export logs issue in singl-node, so it seems safe to avoid installing the network
/assign @BenTheElder

BenTheElder:
/lgtm
/approve
SGTM

I think at some point we might want to be able to test against the latest podman since it seems not so stable, but we'll also want to be able to support the podman in github actions in particular anyhow ...

PR #2074, opened 2021-02-13
compress exported logs

aojea:
kind export logs uses tar to pipe all the /var/log folder and untar
them in the host.

docker always work, but podman flakes exporting the logs because of:
unexpected EOF: error writing to ...

it seems podman uses conmon to the pipe operations, to avoid that
any issues due to conmon not handling correctly some characters, we
can compress the bytes stream.

### DIFF
--- a/pkg/cluster/internal/logs/logs.go
+++ b/pkg/cluster/internal/logs/logs.go
@@ -18,6 +18,7 @@ package logs
 
 import (
 	"archive/tar"
+	"compress/gzip"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -43,7 +44,7 @@ func DumpDir(logger log.Logger, node nodes.Node, nodeDir, hostDir string) (err e
 		// Fatal errors will return exit code 2.
 		// http://man7.org/linux/man-pages/man1/tar.1.html#RETURN_VALUE
 		fmt.Sprintf(
-			`tar --hard-dereference -C %s -chf - . || (r=$?; [ $r -eq 1 ] || exit $r)`,
+			`tar --hard-dereference -C %s -zchf - . || (r=$?; [ $r -eq 1 ] || exit $r)`,
 			shellescape.Quote(path.Clean(nodeDir)+"/"),
 		),
 	)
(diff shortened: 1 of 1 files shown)


BenTheElder:
Typically moving data over local connection this actually makes performance worse.

BenTheElder:
Why would a compressed steam not have arbitrary bytes?
I'm not sure I'm parsing your comment correctly but that sounds like a bad bug in podman.

aojea:
> Why would a compressed steam not have arbitrary bytes?
> I'm not sure I'm parsing your comment correctly but that (inability to pipe certain characters) sounds like a bad bug in podman.

compression removes redundancies and increases entropy, it is a wild guess, but can't find another explanation that some sequence of characters (ie xffffff ) triggers something in conmon, https://github.com/containers/conmon/pull/131

> Typically moving data over local connection this actually makes performance worse.

I'm not sure if the penalty very significant, is just compress a stream of bytes, I used to do 100s of gigas of backups compressing because it reduced plain text by more than 100x

aojea:
1/1 no failures
retriggering

aojea:
2/2

aojea:
3/3
I can't say if it is because it is less amount of data, if this makes more difficult to trigger the unexpected eof or why , but this was pretty flaky

aojea:
4/4

BenTheElder:
> I'm not sure if the penalty very significant, is just compress a stream of bytes, I used to do 100s of gigas of backups compressing because it reduced plain text by more than 100x

It depends on the data, but the performance of this is something I spent time on early in kind (compression when moving stuff over exec & `kind export logs` speed).

I wonder if we can update podman in CI by moving from `ubuntu-latest` in the github actions config which is still ~= 18.04 => 20.04 (which supposedly it's switching to at some point).

aojea:
> I wonder if we can update podman in CI by moving from `ubuntu-latest` in the github actions config which is still ~= 18.04 => 20.04 (which supposedly it's switching to at some point).

https://github.com/kubernetes-sigs/kind/pull/2075

aojea:
/close
in favor of 
https://github.com/kubernetes-sigs/kind/pull/2074

PR #2075, opened 2021-02-14
switch ci runners to ubuntu 20.04

aojea:
instead of depending on the ubuntnu-latest tag used by Github, we can specify the version directly so we have more control on the CI environment

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -9,7 +9,7 @@ on:
 jobs:
   docker:
     name: Docker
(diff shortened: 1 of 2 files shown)


aojea:
missing dependencies in ubuntu 20.04 https://github.com/actions/virtual-environments/issues/2708 :(

aojea:
1/1

aojea:
2/2
ok this seems to make it too

aojea:
3/3
@BenTheElder I think we can go ahead

PR #2088, opened 2021-02-24
Add possibility to disable kube-proxy

brb:
This commit introduces the new kube-proxy mode "none" which is used
to disable kube-proxy when provisioning (via kubeadm init's
skip-phase=addon/kube-proxy).

The motivation for the change is to use Kind for testing Cilium's
kube-proxy replacement \[1\].

\[1\]: https://docs.cilium.io/en/v1.9/gettingstarted/kubeproxy-free/

### DIFF
--- a/pkg/internal/apis/config/types.go
+++ b/pkg/internal/apis/config/types.go
@@ -163,12 +163,12 @@ const (
 type ProxyMode string
 
 const (
-	// IPTablesMode sets ProxyMode to iptables
-	IPTablesMode ProxyMode = "iptables"
-	// IPVSMode sets ProxyMode to iptables
-	IPVSMode ProxyMode = "ipvs"
-	// NoneMode disables kube-proxy
-	NoneMode ProxyMode = "none"
+	// IPTablesProxyMode sets ProxyMode to iptables
+	IPTablesProxyMode ProxyMode = "iptables"
+	// IPVSProxyMode sets ProxyMode to iptables
+	IPVSProxyMode ProxyMode = "ipvs"
+	// NoneProxyMode disables kube-proxy
+	NoneProxyMode ProxyMode = "none"
 )
 
 // PatchJSON6902 represents an inline kustomize json 6902 patch
--- a/pkg/internal/apis/config/types.go
+++ b/pkg/internal/apis/config/types.go
@@ -167,6 +167,8 @@ const (
 	IPTablesMode ProxyMode = "iptables"
 	// IPVSMode sets ProxyMode to iptables
 	IPVSMode ProxyMode = "ipvs"
+	// NoneMode disables kube-proxy
+	NoneMode ProxyMode = "none"
 )
 
 // PatchJSON6902 represents an inline kustomize json 6902 patch
--- a/pkg/apis/config/v1alpha4/default.go
+++ b/pkg/apis/config/v1alpha4/default.go
@@ -68,7 +68,7 @@ func SetDefaultsCluster(obj *Cluster) {
 	}
 	// default the KubeProxyMode using iptables as it's already the default
 	if obj.Networking.KubeProxyMode == "" {
-		obj.Networking.KubeProxyMode = IPTablesMode
+		obj.Networking.KubeProxyMode = IPTablesProxyMode
 	}
 }
 
--- a/pkg/apis/config/v1alpha4/types.go
+++ b/pkg/apis/config/v1alpha4/types.go
@@ -202,10 +202,10 @@ const (
 type ProxyMode string
 
 const (
-	// IPTablesMode sets ProxyMode to iptables
-	IPTablesMode ProxyMode = "iptables"
-	// IPVSMode sets ProxyMode to iptables
-	IPVSMode ProxyMode = "ipvs"
+	// IPTablesProxyMode sets ProxyMode to iptables
+	IPTablesProxyMode ProxyMode = "iptables"
+	// IPVSProxyMode sets ProxyMode to iptables
+	IPVSProxyMode ProxyMode = "ipvs"
 )
 
 // PatchJSON6902 represents an inline kustomize json 6902 patch
--- a/pkg/cluster/internal/create/actions/kubeadminit/init.go
+++ b/pkg/cluster/internal/create/actions/kubeadminit/init.go
@@ -26,16 +26,19 @@ import (
 	"sigs.k8s.io/kind/pkg/cluster/nodeutils"
 
 	"sigs.k8s.io/kind/pkg/cluster/internal/create/actions"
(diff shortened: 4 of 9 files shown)


aojea:
/ok-to-test
I'm +1 on this, 
defer to Ben to review the mechanics of introducing the skip phases changed

pkg/internal/apis/config/types.go
```diff
@@ -167,6 +167,8 @@ const (
 	IPTablesMode ProxyMode = "iptables"
 	// IPVSMode sets ProxyMode to iptables
 	IPVSMode ProxyMode = "ipvs"
+	// NoneMode disables kube-proxy
+	NoneMode ProxyMode = "none"
```
neolit123, line 171: what is the alternative value of this field?
if `""` implies kube-proxy mode this should be reflected in a comment above.

for the API change, as is, this could be useful if one day kind decides to implement it's own proxy ALA Cilium.
adding another ProxyMode.

but if there are no plans for that and if the demand to skip kube-proxy is not high, the better alternative would have been to configure this via kubeadm's v1beta3 ClusterConfiguration.

of course, a couple of problems with that would be:
- kubeadm v1beta3 is not ready yet, but this exact skip phases feature is tracked [here](https://github.com/kubernetes/kubeadm/issues/2261)
- it would only work with kubeadm / k8s latest

a more flexible feature in kind would be to allow passing extra args to the `kubeadm init/join` commands the same way kubeadm allows passing such to the kubelet / control-plane. as per how cobra, go's CLI works, subsequent flags with the same key would override prior ones.

brb: > what is the alternative value of this field?
> if "" implies kube-proxy mode this should be reflected in a comment above.

"" implies iptables mode.

> but if there are no plans for that and if the demand to skip kube-proxy is not high, the better alternative would have been to configure this via kubeadm's v1beta3 ClusterConfiguration.

Once it's ready, we can change the way we disable kube-proxy.

> but if there are no plans for that and if the demand to skip kube-proxy is not high, the better alternative would have been to configure this via kubeadm's v1beta3 ClusterConfiguration.

If we need to pass more flags to kubeadm, then it makes sense. At the moment, it's only one flag.

brb:
[The CI failure](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/2088/pull-kind-e2e-kubernetes-1-20/1364642810329829376) seems to be unrelated:
```
ERROR: An error occurred during the fetch of repository 'debian-iptables-amd64':
   Pull command failed: 2021/02/24 18:31:24 Running the Image Puller to pull images from a Docker Registry...
2021/02/24 18:31:25 Image pull was unsuccessful: unable to save remote image k8s.gcr.io/build-image/debian-iptables@sha256:e30919918299988b318f0208e7fd264dee21a6be9d74bbd9f7fc15e78eade9b4: unable to write image layers: unable to write image layer: unable to write the contents of layer 0 to /bazel-scratch/.cache/bazel/_bazel_root/cae228f2a89ef5ee47c2085e441a3561/external/debian-iptables-amd64/image/000.tar.gz: read tcp 10.60.213.98:43602->172.217.214.128:443: read: connection reset by peer
 (/bazel-scratch/.cache/bazel/_bazel_root/cae228f2a89ef5ee47c2085e441a3561/external/go_puller_linux/file/downloaded -directory /bazel-scratch/.cache/bazel/_bazel_root/cae228f2a89ef5ee47c2085e441a3561/external/debian-iptables-amd64/image -os linux -os-version  -os-features  -architecture amd64 -variant  -features  -name k8s.gcr.io/build-image/debian-iptables@sha256:e30919918299988b318f0208e7fd264dee21a6be9d74bbd9f7fc15e78eade9b4) 
```

pkg/internal/apis/config/types.go
aojea: for the API of this field I think it should be
- "iptables" : default value
- "ipvs"
- "none"

if there are other proxies they should provide their installation, same we are doing with the cni

> a more flexible feature in kind would be to allow passing extra args to the kubeadm init/join commands the same way kubeadm allows passing such to the kubelet / control-plane. as per how cobra, go's CLI works, subsequent flags with the same key would override prior ones.

_I think we already discussed this in another issue_ :thinking:

BenTheElder (commented):
Ordinarily I'd ask that a feature request issue be filed first per our contributing guide [1]

This feature seems relatively straightforward and well contained within the existing API surface though.

In concept this seems reasonable enough to support, and difficult to work around cleanly without it 👍 

Some nits (see comments)

[1]: https://kind.sigs.k8s.io/docs/contributing/getting-started/#4-reaching-out

pkg/internal/apis/config/types.go
neolit123: flags are indeed the riskier option.

brb:
@BenTheElder I addressed your comment. PTAL.

brb:
Both CI failures above look unrelated to my changes.

BenTheElder (approved):
/lgtm
/approve
thanks for your patience, it's been super busy with code freeze in kubernetes and https://github.com/kubernetes/kubernetes/pull/99561

aojea:
dockerhub rate limits on github actions
```
"kindest/node:v1.20.2@sha256:8f7ea6e7642c0da54f04a7ee10431549c0257315b3a634f6ef2fecaaedb19bab": command "docker pull kindest/node:v1.20.2@sha256:8f7ea6e7642c0da54f04a7ee10431549c0257315b3a634f6ef2fecaaedb19bab" failed with error: exit status 1

Command Output: Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
```

flake on 1.19
connection reset on 1.20  read tcp 10.61.33.216:53166->108.177.112.128:443: read: connection reset by peer
/retest

aojea:
@BenTheElder 

```

Kubernetes e2e suite: [sig-network] Service endpoints latency should not be very high [Conformance] expand_less | 1m22s
-- | --
/home/prow/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/test/e2e/framework/framework.go:597 Mar  3 10:03:07.425: Tail (99 percentile) latency should be less than 50s 50, 90, 99 percentiles: 755.062239ms 3.879176932s 50.4752086s

```

50seconds latency for endpoints  is not a healthy environment to run e2e tests
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/2088/pull-kind-e2e-kubernetes-1-19/1367045340850556928
/retest

PR #2090, opened 2021-02-25
Exit with non-zero codes when user input is invalid

### LINKED ISSUE #2087: Running invalid command causes kind CLI to exit with 0
**What happened**:
For example `kind load typo` will exit with 0

Kind to exit with non-zero (Very useful when used in ci or script etc)

```
kind load typo
echo $?
```

This is due to cobra (https://github.com/spf13/cobra/issues/1156) and affects all commands that have sub-commands in kind (build, completion, create, delete, export, get, load)

The specified `NoArgs`or any of the cobra validators don't seem to have any effect unless Run or RunE has been defined for the command. One fix might be to just specify something like:
```
+               Run: func(cmd *cobra.Command, args []string) {
+                       cmd.Help()
+                       os.Exit(1)
+               },
``` 
to every command (or also have them return 0 in case called without arguments). 
I implemented this locally and it seemed to work as it should. Although it's very far from an elegant solution since that would need to duplicated to all of the commands that have sub-commands. 

If my (probably rather bad fix) seems reasonable I'm happy to create a PR, but I think this need a more elegant solution.

**Environment:**

- kind version: (use `kind version`): `kind v0.10.0 go1.15.7 darwin/amd64`

vainu-lauri:
This pr contains my attempt of fixing the exit codes of invalid CLI commands (as defined in the issue #2087).

As you can probably see from the submission I'm no golang expert so this will probably need a bit more or less tuning.

One thing that comes to mind is the philosophical question whether we should consider an incomplete but correct command (like `kind delete`) as a success or a failure. The current implementation in this PR considers it as a failure (I think this way just seemed right).

Closes #2087

### DIFF
--- a/cmd/kind/app/main_test.go
+++ b/cmd/kind/app/main_test.go
@@ -18,6 +18,8 @@ package app
 
 import (
 	"testing"
+
+	"sigs.k8s.io/kind/pkg/cmd"
 )
 
 func TestCheckQuiet(t *testing.T) {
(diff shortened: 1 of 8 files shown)


BenTheElder:
> One thing that comes to mind is the philosophical question whether we should consider an incomplete but correct command (like kind delete) as a success or a failure. The current implementation in this PR considers it as a failure (I think this way just seemed right).

I think if we have to show help and the user did not explicitly request help we should consider it a failure, +1
We can make it not a failure in the future if the command starts to do something useful without further subcommand.

cmd/kind/app/main_test.go
```diff
@@ -78,3 +80,59 @@ func TestCheckQuiet(t *testing.T) {
 		})
 	}
 }
+
+func Test_CommandErrReturn(t *testing.T) {
+	t.Parallel()
+	cases := []struct {
+		Name       string
+		Command    string
+		Subcommand string
```
aojea, line 89: should we add a `expectErr` field and do some positive testing? i..e commands that we know can't fail

BenTheElder: only if we turn this into an integration test, which can be in a follow-up (let's not be running real commands in unit tests)

BenTheElder:
thanks so much!
sorry for the delay getting to this, I've been tied up with some big changes in kubernetes/kubernetes before code freeze hits there 😅 
/lgtm
/approve

aojea:
/retest 
tests timing out in the 1.20 job
job timed out for the ipv6 job

BenTheElder:
we should probably up the limit, looks like it only took 40m

vainu-lauri:
No problem at all, thanks for having a look. Happy to help!

What comes to the positive testing we could have the `kind version`and `kind help` tested without needing integrations.
That would add a little bit of coverage on that side of the commands as well.

BenTheElder:
if you're up for another PR that sounds reasonable to me, good point!

PR #2091, opened 2021-02-25
Improve entrypoint cgroup handling

BenTheElder:
We don't need to depend on the systemd named heirarchy for this. Also we can more robustly match the target controller (in this case I chose `cpu`.

### DIFF
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -104,10 +104,17 @@ fix_cgroup() {
   fi
   echo 'INFO: detected cgroup v1'
   echo 'INFO: fix cgroup mounts for all subsystems'
-  # see: https://d2iq.com/blog/running-kind-inside-a-kubernetes-cluster-for-continuous-integration
-  # capture initial state before modifying
+  # See: https://d2iq.com/blog/running-kind-inside-a-kubernetes-cluster-for-continuous-integration
+  # Capture initial state before modifying
+  #
+  # Basically we're looking for the cgroup-path for the cpu controller for the
+  # current process. this tells us what cgroup-path the container is in.
+  # Then we collect the subsystems that are active on this path.
+  # We assume the cpu controller is in use on all node containers.
+  #
+  # See: https://man7.org/linux/man-pages/man7/cgroups.7.html
   local current_cgroup
-  current_cgroup=$(grep systemd /proc/self/cgroup | cut -d: -f3)
+  current_cgroup=$(grep -E '^[^:]*:([^:]*,)?cpu(,[^,:]*)?:.*' /proc/self/cgroup | cut -d: -f3)
   local cgroup_subsystems
   cgroup_subsystems=$(findmnt -lun -o source,target -t cgroup | grep "${current_cgroup}" | awk '{print $2}')
   # For each cgroup subsystem, Docker does a bind mount from the current
--- a/pkg/build/nodeimage/defaults.go
+++ b/pkg/build/nodeimage/defaults.go
@@ -20,7 +20,7 @@ package nodeimage
 const DefaultImage = "kindest/node:latest"
 
 // DefaultBaseImage is the default base image used
-const DefaultBaseImage = "kindest/base:v20210223-609a941a"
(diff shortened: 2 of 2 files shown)


images/base/files/usr/local/bin/entrypoint
```diff
@@ -104,10 +104,17 @@ fix_cgroup() {
...
+  #
+  # Basically we're looking for the cgroup-path for the cpu controller for the
+  # current process. this tells us what cgroup-path the container is in.
+  # Then we collect the subsystems that are active on this path.
+  # We assume the cpu controller is in use on all node containers.
+  #
+  # See: https://man7.org/linux/man-pages/man7/cgroups.7.html
   local current_cgroup
-  current_cgroup=$(grep systemd /proc/self/cgroup | cut -d: -f3)
+  current_cgroup=$(grep -E '^[^:]*:([^:]*,)?cpu(,[^,:]*)?:.*' /proc/self/cgroup | cut -d: -f3)
```
BenTheElder, line 117:
```
/proc/[pid]/cgroup (since Linux 2.6.24)
              This file describes control groups to which the process
              with the corresponding PID belongs.  The displayed
              information differs for cgroups version 1 and version 2
              hierarchies.

              For each cgroup hierarchy of which the process is a
              member, there is one entry containing three colon-
              separated fields:

                  hierarchy-ID:controller-list:cgroup-path

              For example:

                  5:cpuacct,cpu,cpuset:/daemons

              The colon-separated fields are, from left to right:

              1. For cgroups version 1 hierarchies, this field contains
                 a unique hierarchy ID number that can be matched to a
                 hierarchy ID in /proc/cgroups.  For the cgroups version
                 2 hierarchy, this field contains the value 0.

              2. For cgroups version 1 hierarchies, this field contains
                 a comma-separated list of the controllers bound to the
                 hierarchy.  For the cgroups version 2 hierarchy, this
                 field is empty.

              3. This field contains the pathname of the control group
                 in the hierarchy to which the process belongs.  This
                 pathname is relative to the mount point of the
                 hierarchy.
```

BenTheElder:
@afbjorklund FYI

images/base/files/usr/local/bin/entrypoint
BenTheElder: (we are matching the middle field, specifically one of the comma delimited controller entries)

BenTheElder, line 117: arguably the first and last `*` should be `+` since these fields can't be empty AIUI, but there also shouldn't be more than 2 `:` anyhow and we match both non-optionally.

BenTheElder: sample of this:
```
root@175aed878411:/# cat /proc/self/cgroup
11:devices:/docker/175aed878411df93ce0cbd75904376da6e18318197872e27f1563491f41ffb58
10:pids:/docker/175aed878411df93ce0cbd75904376da6e18318197872e27f1563491f41ffb58
9:cpuset:/docker/175aed878411df93ce0cbd75904376da6e18318197872e27f1563491f41ffb58
8:rdma:/
7:freezer:/docker/175aed878411df93ce0cbd75904376da6e18318197872e27f1563491f41ffb58
6:memory:/docker/175aed878411df93ce0cbd75904376da6e18318197872e27f1563491f41ffb58
5:net_cls,net_prio:/docker/175aed878411df93ce0cbd75904376da6e18318197872e27f1563491f41ffb58
4:perf_event:/docker/175aed878411df93ce0cbd75904376da6e18318197872e27f1563491f41ffb58
3:blkio:/docker/175aed878411df93ce0cbd75904376da6e18318197872e27f1563491f41ffb58
2:cpu,cpuacct:/docker/175aed878411df93ce0cbd75904376da6e18318197872e27f1563491f41ffb58
1:name=systemd:/docker/175aed878411df93ce0cbd75904376da6e18318197872e27f1563491f41ffb58
```

Before we were naively matching on the `name=systemd` named heirarchy, which is a special alternative to a controller (it has no controllers, just for keeping track of stuff). This particular name is systemd specific.

BenTheElder:
flaky old test
/restest

PR #2092, opened 2021-02-25
explicitly set cgroupDriver

BenTheElder:
unblocks https://github.com/kubernetes/kubernetes/pull/99471

will likely follow up switching everything to systemd, but for now keeping status quo, as the minimal patch to unblock.


amwat (commented):
/lgtm

neolit123:
/lgtm

the k/k PR change, as is, should not set it to `systemd` in that case.

BenTheElder:
yeah, as-is because we're not setting it the PR fails (containerd and kubelet out of sync on driver)
this should fix CI on the PR without changing anything, and then we can follow up with flipping over to systemd default anyhow.

PR #2097, opened 2021-02-28
speed up kubernetes builds

BenTheElder:
xref: https://github.com/kubernetes/kubernetes/issues/99517

iterating towards what would make sense to have in a make target

### DIFF
--- a/pkg/build/nodeimage/internal/kube/builder_docker.go
+++ b/pkg/build/nodeimage/internal/kube/builder_docker.go
@@ -24,6 +24,8 @@ import (
 	"sigs.k8s.io/kind/pkg/errors"
 	"sigs.k8s.io/kind/pkg/exec"
 	"sigs.k8s.io/kind/pkg/log"
+
+	"k8s.io/apimachinery/pkg/util/version"
 )
 
 // TODO(bentheelder): plumb through arch
@@ -69,6 +71,11 @@ func (b *dockerBuilder) Build() (Bits, error) {
 		return nil, err
 	}
 
+	kubeVersion, err := version.ParseSemantic(sourceVersionRaw)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to parse source version")
+	}
+
 	// we will pass through the environment variables, prepending defaults
 	// NOTE: if env are specified multiple times the last one wins
 	env := append(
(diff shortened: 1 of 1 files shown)


pkg/build/nodeimage/internal/kube/builder_docker.go
```diff
@@ -101,29 +100,33 @@ func (b *dockerBuilder) Build() (Bits, error) {
...
 	}
 
+	// build binaries
+	if err := exec.InheritOutput(
+		exec.Command(
+			// build dockerized
+			"build/run.sh",
+			// bypass make targets to skip running codegen again, call build
+			// rule implementation directly
+			"hack/make-rules/build.sh",
```
BenTheElder, line 120: the "all" (default) make target is just:
```
all: generated_files
	hack/make-rules/build.sh $(WHAT)
```

and has been for over four years. https://github.com/kubernetes/kubernetes/blame/184c6d348c80c1f529b742c8959677817689a182/build/root/Makefile#L91

BenTheElder: at least five years: https://github.com/kubernetes/kubernetes/blame/682f3a39a0a0a05e2290606a3df26eac575df669/Makefile

BenTheElder: _seven_ years ago it was `hack/build-go.sh $(WHAT)`. around 5 years ago things switched more to make and it went to `hack/make-rules/build.sh` and then the dependency on `generated_files` was introduced to automatically generate those.

BenTheElder:
booo, somehow it depends on this (generated files not being synced?)

aojea:
> booo, somehow it depends on this (generated files not being synced?)

yeah, some tests depend on the generated files

BenTheElder:
> yeah, some tests depend on the generated files

yes, but:
- we're not building tests at that point we're building cli binaries
- we _just_ ran `make quick-release-images` which ran the codegen, so something lost the codegen between that and building the binaries :/

BenTheElder:
The core problem is upstream is still trying to support running the make build against a remote docker which is:
a) unnecessarily complicated, the user can deal with source syncing / ssh themselves
b) slowing down the local build, because we clobber generated files while attempting to save bandwidth

Arguably the entire rsync thing should be done away-with in favor of a cached volume mount and no more special treatment for mac's performance issues here. See background: https://github.com/kubernetes/kubernetes/pull/50717

BenTheElder:
https://github.com/kubernetes/kubernetes/pull/100046 instead (will update the PR if/when that lands)

BenTheElder:
e2e working well. need to fix a spelling error

aojea:
/lgtm
/hold
unhold once you consider necessary

pkg/build/nodeimage/internal/kube/builder_docker.go
```diff
@@ -83,29 +90,35 @@ func (b *dockerBuilder) Build() (Bits, error) {
...
-	).SetEnv(env...)
-	exec.InheritOutput(cmd)
-	if err := cmd.Run(); err != nil {
-		return nil, errors.Wrap(err, "failed to build binaries")
-	}
 
-	// build images
-	cmd = exec.Command("make", "quick-release-images").SetEnv(env...)
+	// build images + binaries (binaries only on 1.21+)
+	cmd := exec.Command("make", "quick-release-images", "KUBE_EXTRA_WHAT="+strings.Join(what, " ")).SetEnv(env...)
```
aojea, line 102: why added here and not in env?

BenTheElder: because it's not common across commands and make supports it either way.

BenTheElder:
upstream test flakes

PR #2103, opened 2021-03-03
Add Documentation for Cluster Name via Yaml Configuration

jsievenpiper:
👋🏾 I love this project -- it's been exactly what I needed for a recent project of my own.

I saw that PRs were welcome for some undocumented config file options, so thought I'd lend a hand!

### DIFF
--- a/site/content/docs/user/configuration.md
+++ b/site/content/docs/user/configuration.md
@@ -36,6 +36,18 @@ To use this config, place the contents in a file `config.yaml` and then run
 
 You can also include a full file path like `kind create cluster --config=/foo/bar/config.yaml`.
 
+### A Note On CLI Parameters and Configuration Files
+
+Unless otherwise noted, parameters passed to the CLI take precedence over their
+equivalents in a config file. For example, if you invoke:
+
+{{< codeFromInline lang="bash" >}}
+kind create cluster --name my-cluster
+{{< /codeFromInline >}}
+
+The name `my-cluster` will be used regardless of the presence of that value in
+your config file.
+
 ## Cluster-Wide Options
 
 The following high level options are available.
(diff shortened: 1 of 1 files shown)


jsievenpiper:
/assign @amwat 

Hope I'm doing this correctly 🙂

site/content/docs/user/configuration.md
```diff
@@ -42,6 +42,20 @@ The following high level options are available.
...
+
+{{< codeFromInline lang="yaml" >}}
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+name: app-1-cluster
+{{< /codeFromInline >}}
+
+This name will be used unless you invoke `kind create cluster` with the `--name`
+option as well (for example `kind create cluster --config=config.yml --name other`),
+in which case `other` will be used since CLI args take precedence.
```
amwat, line 57: I think this is generally true for every parameter. i.e. CLI flags will overwrite whatever is specific in the config file.
Since this page specifically deals with the config, I think just the config example should suffice here in this subsection.

Even better if we can move this note as a general note at the top somewhere :)

amwat (commented):
Thanks for adding this!
/approve

minor nits

site/content/docs/user/configuration.md
BenTheElder: +1 for a general note and also the CLI usage documenting this.

jsievenpiper: @amwat @BenTheElder thanks for the feedback! I pulled that CLI precedence note up and trimmed down the `name` section specifically. Good call out!

BenTheElder:
/lgtm
thank you!

Issue #1895, opened 2020-10-16
Move dockerhub kindest/node to a non rate limited registry


howardjohn:
**What would you like to be added**:
A mirror of kindest/node to GCR or another registry

**Why is this needed**:
On Nov 1, dockerhub will introduce rate limiting on pulls. I am fairly sure this will break our CI, since we pull a lot of kind images.
This is not at all a blocker for us, as we can mirror them into our own registries without issue, but may be useful for the broader community.

BenTheElder:
I think we may even want to primarily move off of it going forward, TBD.

We've been discussing dockerhub mitigations at the SIG Testing level but haven't managed to move on much yet.

This is a higher priority for me though*, other subprojects we don't own can themselves move away from dockerhub and if any projects are using it Kubernetes doesn't officially support them in that and they would have received notice from dockerhub themselves.

\* as sig-testing we also need to mitigate dockerhub usage in e2e.test in kubernetes itself, and will still try to provide mitigations and guidance for CI / subprojects ..

BenTheElder:
Kubernetes provides k8s.gcr.io to subprojects but it's a bit problematic for us:
- not available in china (we have contributors and a sizeable userbase relying on the kindest/node image being on dockerhub and kind otherwise working offline without gcr)
- immutable tags via a promotion system, no direct pushes, no tag mutation ... we'll need to stop using mutable tags

the latter may be a good plan anyhow but the former is ... problematic. the promotion system also makes image pushes a bit onerous, even if you have automated pushes you must manually craft a yaml PR to request that an image be promoted from the staging registry to production.

https://github.com/kubernetes/k8s.io/tree/master/k8s.gcr.io

quay.io and the github package registry are obvious alternatives to look into.

BenTheElder:
https://github.com/kubernetes/test-infra/issues/19477#issuecomment-720782177 this may be the simple route 🙃

BenTheElder:
Mirroring and/or switching registries going forward is still on the table. We can't do a mirror of existing imagds to k8s.gcr.io due to the promotion process and one tag one image though.

I've applied to the docker OSS program.

BenTheElder:
I totally misread this and then went OOO. I'm not clear if we should move forward on this approach, particularly these constraints:

```
While the Publisher retains the Open Source project status, the Publisher agrees to -

Become a Docker public reference for press releases, blogs, webinars, etc
Create joint blogs, webinars and other marketing content 
[...]

PR #2104, opened 2021-03-04
switch all CI to make

BenTheElder:
xref: https://github.com/kubernetes/enhancements/issues/2420

building should be fine on previous branches, we still have plenty of bazel coverage for those in other jobs, after this goes through we can wind down our own bazel tech-debt here.

bazel already doesn't work for some older branches with kind

### DIFF
--- a/hack/ci/e2e-k8s.sh
+++ b/hack/ci/e2e-k8s.sh
@@ -265,11 +265,11 @@ main() {
   kind version
 
   # build kubernetes
-  if [ "${BUILD_TYPE:-}" = "bazel" ]; then
-    build_with_bazel
-  else
+  #if [ "${BUILD_TYPE:-}" = "bazel" ]; then
+  #  build_with_bazel
+  #else
     build
-  fi
+  #fi
 
   # in CI attempt to release some memory after building
   if [ -n "${KUBETEST_IN_DOCKER:-}" ]; then


BenTheElder:
uh oh: 
> Command Output: Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit

fyi @aojea we're hitting rate limits in github actions now: #1895

hack/ci/e2e-k8s.sh
```diff
@@ -265,11 +265,11 @@ main() {
   kind version
 
   # build kubernetes
-  if [ "${BUILD_TYPE:-}" = "bazel" ]; then
-    build_with_bazel
-  else
+  #if [ "${BUILD_TYPE:-}" = "bazel" ]; then
```
amwat, line 268: older branches? and/or delete completely?

amwat (commented):
/lgtm
/approve
/hold

hack/ci/e2e-k8s.sh
BenTheElder: leaving the most trivial revertable state, will come back and fully delete these code paths from more than just the script once I get a chance to confirm the state of CI.

PR #2105, opened 2021-03-04
mirror docker images

aojea:
Kind is publishing the official in dockerhub, however, since
the docker decision to rate limit the images, this causes issues
with CIs.

Switching completely the images is complicated, and can be very
disruptive.

Meanwhile we don't have a final decision we can mirror the images
to different public registries.

Fixes: https://github.com/kubernetes-sigs/kind/issues/1895

### DIFF
--- a/.github/workflows/mirror.yml
+++ b/.github/workflows/mirror.yml
@@ -1,0 +1,24 @@
+name: Mirror
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
+  release:
+    types: [published]
+
+  
+jobs:
+  mirror:
+    name: Mirror images
+    runs-on: ubuntu-20.04
+    timeout-minutes: 30
+
+    steps:
+      - name: Mirror docker images to Github Registry
+        run: |
+          skopeo sync --src docker --dest docker docker.io/kindest/node ghcr.io/kind-ci
+
+      - name: Mirror docker images to quay.io
+        run: |
+          skopeo sync --src docker --dest docker docker.io/kindest/node quay.io/kind
+


aojea:
/hold
/assign @BenTheElder 
Open for discussion, this still needs to handle the authentication
- I was thinking in using a dns domain to hide the registries, so we can geodistribute the requests and we no longer depend on the registry provider.
- Is this the right repo or do we want to use github.com/kind-ci as we do with the containerd nigthly builds
- ....

.github/workflows/mirror.yml
```diff
@@ -0,0 +1,24 @@
+name: Mirror
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
+  release:
```
aojea, line 6: this runs when a release is created

.github/workflows/mirror.yml
```diff
@@ -0,0 +1,24 @@
...
+jobs:
+  mirror:
+    name: Mirror images
+    runs-on: ubuntu-20.04
+    timeout-minutes: 30
+
+    steps:
+      - name: Mirror docker images to Github Registry
+        run: |
+          skopeo sync --src docker --dest docker docker.io/kindest/node ghcr.io/kind-ci
```
aojea, line 19: not familiar with the feature but if it handles the sync ala rsync that seems exactly what we need

neolit123 (commented):
LGTM

tao12345666333:
The authentication information can be completed through the configuration of `--dest-authfile` and `--src-authfile` combined with [GitHub Action encrypted-secrets](https://docs.github.com/en/actions/reference/encrypted-secrets).

Docker Inc. provides support for open source projects, we can consider using it. https://www.docker.com/community/open-source/application

.github/workflows/mirror.yml
AkihiroSuda, line 19: End users will want this mirrored image as well :+1:, so maybe `-ci` suffix can be omitted.

BenTheElder (changes requested):
I'm NACK on mirroring to random registries here without deciding on the solution over in the issue.

Any mirroring we run from here is "official" and should be decided and supported. Also because dockerd has poor support for mirrors we should really just pick 1-2 registries to use going forward (maybe k8s official GCR and something else for china, unless we can pick one alternative for both), and mirroring solutions should be done in discussion with wg-k8s-infra.

.github/workflows/mirror.yml
```diff
@@ -0,0 +1,24 @@
...
+
+      - name: Mirror docker images to quay.io
+        run: |
+          skopeo sync --src docker --dest docker docker.io/kindest/node quay.io/kind
```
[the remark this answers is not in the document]
aojea: I made up the names, the pr is to discuss if this make sense.
If affirmative fill the gaps with whatever we decide

BenTheElder:
[thanks for the PR, let's continue discussion elsewhere, closing to minimize issues from #2120 / thundering CI on open PRs with ok-to-test]

[also received contact from docker today about kindest, need to follow up with them]

PR #2116, opened 2021-03-11
Upgrade base image

BenTheElder:
picks up latest containerd, picks up #1935


BenTheElder:
e2e-kubernetes timed out.
while we want this job to be faster, it may go slower sometimes and there's no reason to not allow it up to an hour like other kubernetes presubmits
I'll try to send a PR to test-infra soon.
/retest

PR #2117, opened 2021-03-11
upgrade to go1.16.1

BenTheElder:
this is a security patch release. none of it looks relevant to us but we should still keep up to date


BenTheElder:
unrelated flakes / timeouts

PR #2121, opened 2021-03-11
reduce node CRI disk usage

BenTheElder:
Enable `discard_unpacked_layers`. You don't want this enabled if you intend to share layers under multiple snapshotters. That's not the case for kind by default. If someone is enabling multiple snapshotters somehow they can also disable this.

### DIFF
--- a/images/base/files/etc/containerd/config.toml
+++ b/images/base/files/etc/containerd/config.toml
@@ -1,10 +1,13 @@
 # explicitly use v2 config format
 version = 2
 
-# set default runtime handler to v2, which has a per-pod shim
 [plugins."io.containerd.grpc.v1.cri".containerd]
+  # save disk space when using a single snapshotter
+  discard_unpacked_layers = true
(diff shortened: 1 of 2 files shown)


aojea:
do we have some cri guy that can advice here? or did you already confirmed with them?

BenTheElder:
I'm the guy :P 

It's documented.

So basically the reason you wouldn't just have this always on is it's a breaking change for someone somehow using N snapshotters (which would then share the layers in that state they are in prior to unpacking to the snapshotter).
Otherwise the upside is once you get a layer unpacked to a snapshotter it can cleanup the layer.

We only use one snapshotter after startup (overlay normally, but maybe the native/naive one on certain filesystems, which we select before startup).

Kubernetes is totally unaware of all of this, so using multiple snapshotters as an end user would already involve patching containerd config.

aojea:
> I'm the guy :P

:100: 
/lgtm
/hold
unhold when you are satisfied with the testing :)

Issue #2113, opened 2021-03-09
podman driver does not handle filesystems using dev mapper


papanito:
*Basically the same issue as #1999 but podman instead docker*

**What happened**:

I run `kind create cluster` on Manjaro linux using podman, but the control plane failed to start.

**What you expected to happen**:

A kind k8s cluster to be created successfully

**How to reproduce it (as minimally and precisely as possible)**:

- Install podman (instead of docker)
- Run `kind create cluster` on Majaro

**Anything else we need to know?**:

- I have firewalld enabled and backend is `FirewallBackend=nftables`
- Filesystem is btrfs and podman is configured accordingly
  `/etc/containers/storage.conf`
   ```
   [storage]
   
   # Default Storage Driver
   driver = "btrfs"
   mountopt = "nodev"
   ...
   ```

**Environment:**

- kind version: (use `kind version`): `kind v0.10.0 go1.15.7 linux/amd64`
- Kubernetes version: (use `kubectl version`): 
  `Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.2", GitCommit:"faecb196815e248d3ecfb03c680a4507229c2a56", GitTreeState:"archive", BuildDate:"2021-01-13T22:07:54Z", GoVersion:"go1.15.6", Compiler:"gc", Platform:"linux/amd64"}`
- Docker version: (use `sudo podman info`):
    ```
    host:
    arch: amd64
    buildahVersion: 1.19.4
    cgroupManager: systemd
    cgroupVersion: v2
    conmon:
        package: /usr/bin/conmon is owned by conmon 1:2.0.26-1
        path: /usr/bin/conmon
        version: 'conmon version 2.0.26, commit: 0e155c83aa739ef0a0540ec9f9d265f57f68038b'
    cpus: 8
    distribution:
        distribution: manjaro
        version: unknown
    eventLogger: journald
    hostname: clawfinger
    idMappings:
        gidmap: null
        uidmap: null
    kernel: 5.10.18-1-MANJARO
    linkmode: dynamic
    memFree: 779718656
    memTotal: 16644816896
    ociRuntime:
        name: crun
        package: /usr/bin/crun is owned by crun 0.18-1
        path: /usr/bin/crun
        version: |-
        crun version 0.18
        commit: 808420efe3dc2b44d6db9f1a3fac8361dde42a95
        spec: 1.0.0
[...]

PR #2122, opened 2021-03-11
podman: mount dev mapper for btrfs and zfs storage

aojea:
Replicate the logic of the docker provider in podman

Fixes https://github.com/kubernetes-sigs/kind/issues/2113

### DIFF
--- a/pkg/cluster/internal/providers/podman/util.go
+++ b/pkg/cluster/internal/providers/podman/util.go
@@ -115,3 +115,19 @@ func deleteVolumes(names []string) error {
 	cmd := exec.Command("podman", args...)
 	return cmd.Run()
 }
+
+// mountDevMapper checks if the podman storage driver is Btrfs or ZFS
+func mountDevMapper() bool {
+	storage := ""
+	cmd := exec.Command("podman", "info", "-f",
+		`{{ index .Store.GraphStatus "Backing Filesystem"}}`)
+	lines, err := exec.OutputLines(cmd)
+	if err != nil {
+		return false
+	}
+
+	if len(lines) > 0 {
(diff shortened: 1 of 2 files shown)


pkg/cluster/internal/providers/podman/util.go
```diff
@@ -115,3 +115,19 @@ func deleteVolumes(names []string) error {
 	cmd := exec.Command("podman", args...)
 	return cmd.Run()
 }
+
+// mountDevMapper checks if the Docker storage driver is Btrfs or ZFS
```
amwat, line 119: nit: s/Docker/podman

amwat (commented):
/lgtm
/approve
/hold

assuming it all works the same, no setup to test :)

pkg/cluster/internal/providers/podman/util.go
aojea: thanks :)

pkg/cluster/internal/providers/podman/util.go
```diff
@@ -115,3 +115,19 @@ func deleteVolumes(names []string) error {
 	cmd := exec.Command("podman", args...)
 	return cmd.Run()
 }
+
+// mountDevMapper checks if the podman storage driver is Btrfs or ZFS
+func mountDevMapper() bool {
+	storage := ""
+	cmd := exec.Command("podman", "info", "-f",
+		`{{ index .Store.GraphStatus "Backing Filesystem"}}`)
```
BenTheElder, line 123: we need to get the docker driver checking the backing filesystem too (different PR)

BenTheElder:
/lgtm
/approve
thanks!

Issue #2120, opened 2021-03-11
tracking: switch default branch to main


BenTheElder:
see: https://k8s.dev/rename, https://github.com/kubernetes/org/issues/2222

Checklist below added from https://github.com/kubernetes/community/commit/198374dd9147981e561ed99d0e1453189343999f

## Prerequisites

- [x] Ensure that your repo has low PR volume (<20 open PRs) and
less number of periodic jobs. The branch rename will re-trigger
prow on _all_ open PRs, which will cause a huge spike in the CI load.
  - We have zero periodic jobs directly consuming this repo
  - We have < 10 PRs open now (17:14 PST 2021-03-15)

- [x] Create an issue in your repo to track the branch rename.
You can paste this checklist in the issue body.
  - This is the issue

- [x] If you are not a root approver for the repo, assign a root
approver for approval.
  - I am a root approver

- [x] Once the issue has been approved, send a notice to your SIG's
mailing list about the potential branch rename.
  - Notice here: https://groups.google.com/g/kubernetes-sig-testing/c/Gi0A_TJsP7A

## Changes pre-rename

Make the following changes _before_ renaming the branch the `master` branch.

Note: There might be additional changes required that have not been
covered in this checklist.

### Anytime

These changes are **non-disruptive**  and can be made anytime before
renaming the branch.

- [x] If a prowjob triggers on the `master` branch (`branches` field
of the prowjob), add the `main` branch to the list
(see [kubernetes/test-infra#20665] for an example).

  - N/A: we only use one branch and don't specify it wherever possible

- [x] If the [`milestone_applier`] prow config references the `master` branch,
add the `main` branch to the config (see [kubernetes/test-infra#20675] for an example).

  - N/A: we don't use milestone applier

- [x] If the [`branch_protection`] prow config references the `master` branch,
add the `main` branch to the config.

  - N/A: we only use the automatic branch protection based on prowjobs

### Just before rename

These changes are **disruptive** and should be made just before
renaming the branch.

- [x] If a prowjob mentions the `master` branch in `base_ref`,
update it to the `main` branch. For a periodic job, ensure that
the branch is renamed between periodic job runs.
    - N/A we have no jobs that mention branches of this repo.

- [x] If a prowjob mentions `master` in its name, rename the job to
to not include the branch name. [`status-reconciler`] should automatically
migrate the PR status contexts to the new job name but this has not been tested yet.
The job with the new name will also appear as a different job in Testgrid.
[...]

PR #2125, opened 2021-03-12
Add a shortcode for inserting the stable version

### LINKED ISSUE #2118: dedupe stable install version in site content
We should use a hugo shortcode and data under site/ to allow inserting the stable version everywhere.

NOTE: We do *not* want to use `@latest` because we want to encourage people to use *specific* versions and perform controlled upgrades, not a floating version.

rudeigerc:
Signed-off-by: rudeigerc <rudeigerc@gmail.com>

Closes #2118 

Add a shortcode for inserting the stable version, borrowed from the implementation of kubernetes/website.

I have tried fetching the latest tag through GitHub API,
```html
{{- $url := "https://api.github.com/repos/kubernetes-sigs/kind/releases/latest" -}}
{{- with getJSON $url -}}{{- .tag_name -}}{{- end -}}
```
but this may reach the request limit of API with intensive requests.

As a result, I think simply defining a parameter in `config.toml` would be better.

### DIFF
--- a/site/config.toml
+++ b/site/config.toml
@@ -66,6 +66,9 @@ baseName = "_redirects"
 [outputs]
 home = ["HTML", "REDIRECTS"]
 
+[params]
+stable = "v0.10.0"
+
 # privacy settings
 [privacy]
   [privacy.youtube]
--- a/site/content/_index.md
+++ b/site/content/_index.md
@@ -6,7 +6,7 @@ title: kind
 [kind] is a tool for running local Kubernetes clusters using Docker container "nodes".  
 kind was primarily designed for testing Kubernetes itself, but may be used for local development or CI.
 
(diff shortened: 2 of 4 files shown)


site/config.toml
```diff
@@ -66,6 +66,9 @@ baseName = "_redirects"
 [outputs]
 home = ["HTML", "REDIRECTS"]
 
+[params]
+latest = "v0.10.0"
```
BenTheElder, line 70: nit: s/latest/stable ?
latest in go conjures up @latest, which we are explicitly not using.

BenTheElder (commented):
thank you for the PR! one suggestion

site/config.toml
rudeigerc: Sure. Updated 😃

amwat (commented):
/lgtm
/approve

Thank you!

BenTheElder:
(not worried about code tests here, and getting PR backlog relatively cleared for #2120)

BenTheElder:
thank you!

PR #2127, opened 2021-03-13
rootless: fall back to fuse-overlayfs snapshotter when overlayfs is not available

AkihiroSuda:
Mounting overlayfs inside UserNS requires kernel >= 5.11, or Ubuntu kernel (since circa 2015), or Debian kernel (since Debian 10).

On older kernel, the kind entrypoint now changes the snapshotter from "overlayfs" to "fuse-overlayfs".

Prior to this commit, the snapshotter was changed from "overlayfs" to "native" in such a condition.

New binaries:
```console
# ls -lh /usr/bin/fuse-overlayfs /usr/local/bin/containerd-fuse-overlayfs-grpc 
-rwxr-xr-x. 1 root root  91K Apr 18  2020 /usr/bin/fuse-overlayfs
-rwxr-xr-x. 1 root root 9.6M Mar  8 08:10 /usr/local/bin/containerd-fuse-overlayfs-grpc
```

Test node image: `akihirosuda/tmp-kind-node:g661491bc-v1.20.5`

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -55,6 +55,14 @@ ARG CNI_PLUGINS_AMD64_SHA256SUM="962100bbc4baeaaa5748cdbfce941f756b1531c2eadb290
 ARG CNI_PLUGINS_ARM64_SHA256SUM="ef17764ffd6cdcb16d76401bac1db6acc050c9b088f1be5efa0e094ea3b01df0"
 ARG CNI_PLUGINS_PPC64LE_SHA256SUM="5bd3c82ef248e5c6cc388f25545aa5a7d318778e5f9bc0a31475361bb27acefe"
 
+# Configure containerd-fuse-overlayfs snapshotter binary from upstream
+ARG CONTAINERD_FUSE_OVERLAYFS_VERSION="1.0.2"
+ARG CONTAINERD_FUSE_OVERLAYFS_TARBALL="v${CONTAINERD_FUSE_OVERLAYFS_VERSION}/containerd-fuse-overlayfs-${CONTAINERD_FUSE_OVERLAYFS_VERSION}-linux-${TARGETARCH}.tar.gz"
+ARG CONTAINERD_FUSE_OVERLAYFS_URL="https://github.com/containerd/fuse-overlayfs-snapshotter/releases/download/${CONTAINERD_FUSE_OVERLAYFS_TARBALL}"
+ARG CONTAINERD_FUSE_OVERLAYFS_AMD64_SHA256SUM="1f1e69f71b5ea568e93e40059af1b02a377ac0966d2acd27e4cce388a27af218"
+ARG CONTAINERD_FUSE_OVERLAYFS_ARM64_SHA256SUM="7ade1a44d880b3fb8eaa3c5ff7d3890a43b777d06ec80439c9a51ae35626c83c"
+ARG CONTAINERD_FUSE_OVERLAYFS_PPC64LE_SHA256SUM="eaf9bdd3de4514546945ea93119acea2b7bfa55ced43766e20adabddd5d20978"
+
 # copy in static files
 # all scripts are 0755: http://www.filepermissions.com/file-permission/0755
 COPY --chmod=0755 files/usr/local/bin/* /usr/local/bin/
@@ -102,7 +110,7 @@ RUN echo "Installing Packages ..." \
       conntrack iptables iproute2 ethtool socat util-linux mount ebtables udev kmod \
       libseccomp2 pigz \
       bash ca-certificates curl rsync \
-      nfs-common \
+      nfs-common fuse-overlayfs \
     && find /lib/systemd/system/sysinit.target.wants/ -name "systemd-tmpfiles-setup.service" -delete \
     && rm -f /lib/systemd/system/multi-user.target.wants/* \
     && rm -f /etc/systemd/system/*.wants/* \
@@ -160,6 +168,15 @@ RUN echo "Installing Packages ..." \
          -o -iname loopback \
       \) \
       -delete \
+ && echo "Installing containerd-fuse-overlayfs ..." \
+    && curl -sSL --retry 5 --output /tmp/containerd-fuse-overlayfs.${TARGETARCH}.tgz "${CONTAINERD_FUSE_OVERLAYFS_URL}" \
+    && echo "${CONTAINERD_FUSE_OVERLAYFS_AMD64_SHA256SUM}  /tmp/containerd-fuse-overlayfs.amd64.tgz" | tee /tmp/containerd-fuse-overlayfs.sha256 \
+    && echo "${CONTAINERD_FUSE_OVERLAYFS_ARM64_SHA256SUM}  /tmp/containerd-fuse-overlayfs.arm64.tgz" | tee -a /tmp/containerd-fuse-overlayfs.sha256 \
+    && echo "${CONTAINERD_FUSE_OVERLAYFS_PPC64LE_SHA256SUM}  /tmp/containerd-fuse-overlayfs.ppc64le.tgz" | tee -a /tmp/containerd-fuse-overlayfs.sha256 \
+    && sha256sum --ignore-missing -c /tmp/containerd-fuse-overlayfs.sha256 \
+    && rm -f /tmp/containerd-fuse-overlayfs.sha256 \
+    && tar -C /usr/local/bin -xzvf /tmp/containerd-fuse-overlayfs.${TARGETARCH}.tgz \
+    && rm -rf /tmp/containerd-fuse-overlayfs.${TARGETARCH}.tgz \
  && echo "Ensuring /etc/kubernetes/manifests" \
     && mkdir -p /etc/kubernetes/manifests \
  && echo "Adjusting systemd-tmpfiles timer" \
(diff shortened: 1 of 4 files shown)


images/base/Dockerfile
```diff
@@ -55,6 +55,14 @@ ARG CNI_PLUGINS_AMD64_SHA256SUM="962100bbc4baeaaa5748cdbfce941f756b1531c2eadb290
...
 ARG CNI_PLUGINS_PPC64LE_SHA256SUM="5bd3c82ef248e5c6cc388f25545aa5a7d318778e5f9bc0a31475361bb27acefe"
 
+# Configure containerd-fuse-overlayfs snapshotter binary from upstream
+ARG CONTAINERD_FUSE_OVERLAYFS_VERSION="1.0.1"
+ARG CONTAINERD_FUSE_OVERLAYFS_TARBALL="v${CONTAINERD_FUSE_OVERLAYFS_VERSION}/containerd-fuse-overlayfs-${CONTAINERD_FUSE_OVERLAYFS_VERSION}-linux-${TARGETARCH}.tar.gz"
+ARG CONTAINERD_FUSE_OVERLAYFS_URL="https://github.com/AkihiroSuda/containerd-fuse-overlayfs/releases/download/${CONTAINERD_FUSE_OVERLAYFS_TARBALL}"
+ARG CONTAINERD_FUSE_OVERLAYFS_AMD64_SHA256SUM="ffeccbf101895193717f5cce8fa5197070af86f277783e22809f7d767ec5034f"
+ARG CONTAINERD_FUSE_OVERLAYFS_ARM64_SHA256SUM="f20e8a8fb812a0341eb8f63a3747efe7bdb4f2ee4be280e4e2a9d2fe42be5f8f"
+ARG CONTAINERD_FUSE_OVERLAYFS_PPC64LE_SHA256SUM="e6a67d58f5b31c8ae2ab7a67cb87992b40d91a5686b60ca5350cc3197f648b08"
+
```
AkihiroSuda, line 65: FYI these SHA256SUMS can be verified here https://github.com/AkihiroSuda/containerd-fuse-overlayfs/runs/2054687860?check_suite_focus=true

images/base/files/etc/systemd/system/containerd-fuse-overlayfs.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=containerd fuse-overlayfs snapshotter
+PartOf=containerd.service
+
+[Service]
+ExecStart=/usr/local/bin/containerd-fuse-overlayfs-grpc /run/containerd-fuse-overlayfs.sock /var/lib/containerd-fuse-overlayfs
+Type=notify
```
aojea, line 7: does this means this is always running?
is socket based activation an option?

AkihiroSuda: The `containerd-fuse-overlayfs.service` is disabled by default. The service is enabled when `entrypoint` detects that kernel-mode overlayfs is unavailable.

The `Type=notify` line just means that the `containerd-fuse-overlayfs.service` can talk `sd_notify` API.
Socket-based activation is unlikely to be an option, as the service needs to be always running when there is an active container that uses `fuse-overlayfs`.

aojea: :+1:

images/base/Dockerfile
```diff
@@ -55,6 +55,14 @@ ARG CNI_PLUGINS_AMD64_SHA256SUM="962100bbc4baeaaa5748cdbfce941f756b1531c2eadb290
 ARG CNI_PLUGINS_ARM64_SHA256SUM="ef17764ffd6cdcb16d76401bac1db6acc050c9b088f1be5efa0e094ea3b01df0"
 ARG CNI_PLUGINS_PPC64LE_SHA256SUM="5bd3c82ef248e5c6cc388f25545aa5a7d318778e5f9bc0a31475361bb27acefe"
 
+# Configure containerd-fuse-overlayfs snapshotter binary from upstream
+ARG CONTAINERD_FUSE_OVERLAYFS_VERSION="1.0.1"
+ARG CONTAINERD_FUSE_OVERLAYFS_TARBALL="v${CONTAINERD_FUSE_OVERLAYFS_VERSION}/containerd-fuse-overlayfs-${CONTAINERD_FUSE_OVERLAYFS_VERSION}-linux-${TARGETARCH}.tar.gz"
+ARG CONTAINERD_FUSE_OVERLAYFS_URL="https://github.com/AkihiroSuda/containerd-fuse-overlayfs/releases/download/${CONTAINERD_FUSE_OVERLAYFS_TARBALL}"
```
BenTheElder, line 61: we're a little adverse to shipping dependencies from user repos.
do you think containerd would be willing to host this?

BenTheElder:
Given how new packaged docker rootless is, can we not just require kernel 5.11?

I'm hesitant to add more branching and a new dependency to support something experimental that is already fixed if the user upgrades.

images/base/Dockerfile
AkihiroSuda: Discussing with containerd maintainers

AkihiroSuda:
> Given how new packaged docker rootless is, can we not just require kernel 5.11?

It will take a few years until users use kernel >= 5.11. 
Especially users of RHEL8-derived distros.

BenTheElder:
> Especially users of RHEL8-derived distros.

I'd still be inclined to suggest that this is a feature or RHEL and if users want the latest features they probably shouldn't use RHEL / force KIND to work around RHEL for this feature.

How long do you think it will take in other common less frozen distros? We usually aren't reliant on bleeding edge kernel features so I haven't really looked into this before.

We have users that need to account for all binaries (see: https://github.com/kubernetes-sigs/kind/issues/2030) in order to trust kind (which is somewhat understandable, given what running kind entails), so we also have to consider that angle when we add something like this.

aojea:
I see that podman uses same fuse overlayfs , can this be a workardound for the RHEL derivatives?
```
    overlay.mount_program:
      Executable: /usr/bin/fuse-overlayfs
      Package: fuse-overlayfs-1.2.0-1.fc32.x86_64
      Version: |-
```

if not, I agree with Ben, and I prefer to postpone this until we have more evidence/demand for this feature on this distros

images/base/Dockerfile
AkihiroSuda: Moved the repo to https://github.com/containerd/fuse-overlayfs-snapshotter

images/base/Dockerfile
AkihiroSuda: Moved the repo: https://github.com/containerd/fuse-overlayfs-snapshotter/runs/2153593746?check_suite_focus=true#step:6:7

AkihiroSuda:
Moved  `github.com/AkihiroSuda/containerd-fuse-snapshotter` repo to https://github.com/containerd/fuse-overlayfs-snapshotter .

New node image: `akihirosuda/tmp-kind-node:g661491bc-v1.20.5`

> How long do you think it will take in other common less frozen distros?

### Debian
Debian 11 is planned to be released with kernel 5.10 this year, so Debian users will not be able to use kernel >= 5.11 until Debian 12 (ETA: circa. 2023-2024).

Although Debian has been already supporting rootless overlayfs since Debian 10 by patching their kernel, it requires modprobbing `overlay.ko` with `permit_mounts_in_userns=1` option.
And it seems to have stability issue compared to genuine rootless overlayfs in kernel 5.11: https://github.com/docker/for-linux/issues/1055

So Debian 10 and 11 users are likely to prefer `fuse-overlayfs`.

### Ubuntu
Ubuntu 21.04 is likely to be released with kernel 5.11, but typical users prefer LTS release, so they won't be able to use kernel >=5.11 until the release of Ubuntu 22.04 LTS.

Ubuntu has been already supporting rootless overlayfs since circa. 2015-2016 by patching their kernel, but users may prefer to  use `fuse-overlayfs` for stability, until they can get kernel >= 5.11.

### SUSE
openSUSE 15.3 (ETA July) and SLES 15 SP3 plan to use kernel 5.3: https://en.opensuse.org/Features_15.3#Linux_kernel

It is likely that (non-tumbleweed) SUSE users cannot use kernel >= 5.11 until 2022-2023.

### Fedora
Fedora 34 (ETA April) is planned to be released with kernel 5.11.

### Arch
Already shipped with kernel 5.11

### Gentoo
Already shipped with kernel 5.11

- - -

> I see that podman uses same fuse overlayfs , can this be a workardound for the RHEL derivatives?

containerd requires https://github.com/containerd/stargz-snapshotter to use fuse-overlayfs.

AkihiroSuda:
@BenTheElder @aojea WDYT? ↑

aojea:
but users still have to enable cgroupsv2 in order to have rootless, right?

AkihiroSuda:
> but users still have to enable cgroupsv2 in order to have rootless, right?

Yes (w.r.t. kind)

aojea:
> > but users still have to enable cgroupsv2 in order to have rootless, right?
> 
> Yes (w.r.t. kind)

buff, that sounds like the biggest blocker for users on the "more stable" distros.
I think we should go first with the Ubuntu/Fedora/Arch/Gentoo people and revisit later when we have more evidence

AkihiroSuda:
It turned out that Fedora 34 (kernel 5.11) still requires fuse-overlayfs when SELinux is enabled: https://github.com/moby/moby/issues/42333

Any chance to get this PR merged? :pray:

jesperpedersen:
Fedora 34 is on Linux 5.12.x now

AkihiroSuda:
Probably it is better to use fuse-overlayfs by default for rootless, regardless to the kernel version.
Opened https://github.com/kubernetes-sigs/kind/issues/2275 for discussion.

PR #2128, opened 2021-03-13
create control plane containers with `-e KUBECONFIG=/etc/kubernetes/admin.conf`

### LINKED ISSUE #2101: Add 'export KUBECONFIG=/etc/kubernetes/admin.conf` to `/root/.bashrc`

Add `export KUBECONFIG=/etc/kubernetes/admin.conf` to `/root/.bashrc`

**Why is this needed**:

So that `kubectl` works inside `docker exec -it <node> bash` shell (for ease of debugging kind itself).

AkihiroSuda:
For ease of running `kubectl` inside `docker exec -it <CONTROLPLANE>` for debugging.

Fix #2101

### DIFF
--- a/pkg/cluster/internal/providers/docker/provision.go
+++ b/pkg/cluster/internal/providers/docker/provision.go
@@ -254,6 +254,11 @@ func runArgsForNode(node *config.Node, clusterIPFamily config.ClusterIPFamily, n
 	}
 	args = append(args, mappingArgs...)
 
+	switch node.Role {
+	case config.ControlPlaneRole:
+		args = append(args, "-e", "KUBECONFIG=/etc/kubernetes/admin.conf")
+	}
+
 	// finally, specify the image to run
 	return append(args, node.Image), nil
 }
--- a/pkg/cluster/internal/providers/podman/provision.go
+++ b/pkg/cluster/internal/providers/podman/provision.go
@@ -217,6 +217,11 @@ func runArgsForNode(node *config.Node, clusterIPFamily config.ClusterIPFamily, n
 	}
 	args = append(args, mappingArgs...)
 
+	switch node.Role {
+	case config.ControlPlaneRole:
+		args = append(args, "-e", "KUBECONFIG=/etc/kubernetes/admin.conf")
+	}
+
 	// finally, specify the image to run
 	_, image := sanitizeImage(node.Image)
 	return append(args, image), nil


aojea:
/lgtm
/ok-to-test
This is really nice

aojea:
@BenTheElder do you prefer to do it in the node image or during the node provisioning?

BenTheElder:
If we're going to enable this as an "api" why only under bash?

AkihiroSuda:
Changed to add  `ENV KUBECONFIG=/etc/kubernetes/admin.conf` to Dockerfile

aojea:
/lgtm
This file is a kubeadm default https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-config/ so I think that it won't break

BenTheElder:
if it does break we can move adding the env to node creation.

BenTheElder:
> @BenTheElder do you prefer to do it in the node image or during the node provisioning?

node provisioning probably makes sense tbh, it also means we get this across existing images.
it's slightly awkward layering either way but 🤷‍♂️

images/base/Dockerfile
```diff
@@ -163,6 +163,9 @@ RUN echo "Ensuring scripts are executable ..." \
 # tell systemd that it is in docker (it will check for the container env)
 # https://systemd.io/CONTAINER_INTERFACE/
 ENV container docker
+# set KUBECONFIG so that kubectl can be executed inside the control plane containers.
+# this is invalid for node containers (by design).
```
amwat, line 167:
```suggestion
# this is invalid for worker node containers (by design).
```

BenTheElder:
also this file does not exist on worker nodes (and we don't plan for it to exist), so it makes more sense to decorate control plane nodes with this at runtime

AkihiroSuda:
> also this file does not exist on worker nodes (and we don't plan for it to exist), so it makes more sense to decorate control plane nodes with this at runtime

How can we do this?

images/base/Dockerfile
AkihiroSuda: fixed

BenTheElder:
In https://github.com/kubernetes-sigs/kind/blob/7eaed6d7cc5361471c82fc07fb4fd98c5e639ec5/pkg/cluster/internal/providers/docker/provision.go#L99 I guess.
There's not a super great way to avoid assumptions as it conflates the nodes (or currently the images) with the provisioning details (kubeadm + kind logic at runtime), but we can at least only export it on control plane nodes (which will have the config).

AkihiroSuda:
Updated. Didn't know that `docker run -e` envs are also propagated to `docker exec`

BenTheElder:
/lgtm
/approve
this seems good to me, thanks!

PR #2129, opened 2021-03-15
rootless: fail early if prerequiresites are not satisfied

AkihiroSuda:
`kind create cluster` now prints errors before running the containers.

### DIFF
--- a/pkg/cluster/internal/create/create.go
+++ b/pkg/cluster/internal/create/create.go
@@ -66,6 +66,11 @@ type ClusterOptions struct {
 
 // Cluster creates a cluster
 func Cluster(logger log.Logger, p providers.Provider, opts *ClusterOptions) error {
+	// validate provider first
+	if err := validateProvider(p); err != nil {
+		return err
+	}
+
 	// default / process options (namely config)
 	if err := fixupOptions(opts); err != nil {
 		return err
@@ -232,5 +237,21 @@ func fixupOptions(opts *ClusterOptions) error {
 	// may be constructed in memory rather than from disk)
 	config.SetDefaultsCluster(opts.Config)
 
+	return nil
+}
+
+func validateProvider(p providers.Provider) error {
+	info, err := p.Info()
+	if err != nil {
+		return err
+	}
+	if info.Rootless {
+		if !info.Cgroup2 {
+			return errors.New("running kind with rootless provider requires cgroup v2, see https://kind.sigs.k8s.io/docs/user/rootless/")
+		}
+		if !info.SupportsMemoryLimit || !info.SupportsPidsLimit || !info.SupportsCPUShares {
+			return errors.New("running kind with rootless provider requires setting systemd property \"Delegate=yes\", see https://kind.sigs.k8s.io/docs/user/rootless/")
+		}
+	}
 	return nil
 }
--- a/pkg/cluster/internal/providers/podman/provider.go
+++ b/pkg/cluster/internal/providers/podman/provider.go
@@ -19,6 +19,7 @@ package podman
 import (
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"net"
 	"os"
 	"path/filepath"
@@ -53,6 +54,7 @@ func NewProvider(logger log.Logger) providers.Provider {
 // see NewProvider
 type provider struct {
 	logger log.Logger
+	info   *providers.ProviderInfo
 }
 
 // String implements fmt.Stringer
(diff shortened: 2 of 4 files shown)


pkg/cluster/internal/create/create.go
```diff
@@ -66,6 +66,11 @@ type ClusterOptions struct {
 
 // Cluster creates a cluster
 func Cluster(logger log.Logger, p providers.Provider, opts *ClusterOptions) error {
+	// validate provider first
+	if err := validateProvider(p); err != nil {
```
aojea, line 70: @BenTheElder I don't know if we should validate it only on `create cluster` or always in `DetectNodeProvider()` , 
we have `<provider>.IsAvailable()`
Should it be `<provider>.IsAvailableAndValid()` ?

AkihiroSuda: I added this validation to `create.go` because deletion should not require this validation.

aojea: my comment is more about the long term solution, Ben always wanted to model the providers API (as you can see in the comment that I pasted above)

So I'm wondering if this is the time to do it, to avoid start to grow it organically, :+1: 

> When a user booted the host with cgroup v2, created a rootless kind cluster, and then rebooted with cgroup v1 for running some other apps that do not support cgroup v2, the user still want to be able to remove the kind cluster.

we should also start to think in what is supported , that is an interesting use case

BenTheElder: I failed to click submit on my comment earlier it seems ...

Delete should not require validation indeed, the validation is wether the tagged resource exists.

It's fine for this to grow somewhat organically internally for now, we can create a better API in the future and consider it exporting it once we have a better idea what we need. It's the public APIs that we need to be more careful with (because people already depend on them and we can't refactor them easily, so we need to make any changes minimally difficult to deal with / not really remove APIs etc.). We can completely rewrite our own internal usage. I'm going to take a sledgehammer to the "actions" thing when I get some freetime someday, and the node build code ...

pkg/cluster/internal/providers/podman/provider.go
```diff
@@ -347,8 +349,46 @@ func (p *provider) CollectLogs(dir string, nodes []nodes.Node) error {
...
 // Info returns the provider info.
 func (p *provider) Info() (*providers.ProviderInfo, error) {
+	var err error
+	if p.info == nil {
+		p.info, err = info(p.logger)
+	}
+	return p.info, err
+}
+
+func info(logger log.Logger) (*providers.ProviderInfo, error) {
```
aojea, line 359: the linter complains here
```
pkg/cluster/internal/providers/podman/provider.go:359:56: info - result 1 (error) is always nil (unparam)
func info(logger log.Logger) (*providers.ProviderInfo, error) {
```

AkihiroSuda: fixed, thanks

AkihiroSuda:
--

AkihiroSuda:
@BenTheElder PTAL?

BenTheElder (approved):
thanks, sorry I've been prioritizing bug fixes & test coverage (k8s release cycle spilling over everywhere)

pkg/cluster/internal/providers/podman/provider.go
```diff
@@ -347,8 +349,45 @@ func (p *provider) CollectLogs(dir string, nodes []nodes.Node) error {
 
 // Info returns the provider info.
 func (p *provider) Info() (*providers.ProviderInfo, error) {
```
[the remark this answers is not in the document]
aojea: this can not change on runtime, right?

aojea:
something is going on with 
pull-kind-e2e-kubernetes

lots of timeout and failures lately

pkg/cluster/internal/providers/podman/provider.go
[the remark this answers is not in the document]
AkihiroSuda: Added a comment line about that.

> this can not change on runtime, right?

Right

pkg/cluster/internal/providers/podman/provider.go
```diff
@@ -347,8 +349,45 @@ func (p *provider) CollectLogs(dir string, nodes []nodes.Node) error {
...
+	return p.info, nil
+}
+
+func info(logger log.Logger) *providers.ProviderInfo {
+	euid := os.Geteuid()
 	info := &providers.ProviderInfo{
-		Rootless: os.Geteuid() != 0,
+		Rootless: euid != 0,
+	}
+	if _, err := os.Stat("/sys/fs/cgroup/cgroup.controllers"); err == nil {
```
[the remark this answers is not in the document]
AkihiroSuda: I think that will be a future work, and probably needs some work on Podman side.

PR #2134, opened 2021-03-16
Update pause container

troy0820:
After kubernetes/k8s.io#1793 is promoted, update the pause image in KinD


BenTheElder:
watching https://github.com/containerd/containerd/pull/5211 first cc @dims

BenTheElder:
this fell off my radar and we've had enough changes in flight close to the much awaited v0.11 so I'm going to say this should be a very early merge after we finish publishing v0.11

I'm still not sure kubernetes is handling this right overall but that's a different discussion ...

BenTheElder:
/hold cancel
the release is out now, thanks again!

PR #2137, opened 2021-03-16
update references to kind master branch -> main branch

BenTheElder:
updates docs and github actions, xref: #2120


hack/ci/cache-wrapper.sh
```diff
@@ -20,7 +20,7 @@ set -o errexit -o nounset -o pipefail
 
 # options for where the cache is stored
 BUCKET="${BUCKET:-bentheelder-kind-ci-builds}"
```
BenTheElder, line 22: this script is not actually in use currently, as evidence by the demo bucket here, but updating this aspect of it anyhow.

PR #2139, opened 2021-03-17
Copy scripts and configs separately during base image build

### LINKED ISSUE #2138: Copying files/ to / before clean-install, fails base image build
**What happened**:
```
$ make -C images/base quick
...<redacted>...
#7 0.150 Installing Packages ...
#7 0.231 Err:1 http://archive.ubuntu.com/ubuntu groovy InRelease
#7 0.231   Something wicked happened resolving 'archive.ubuntu.com:80' (-5 - No address associated with hostname)
#7 0.231 Err:2 http://archive.ubuntu.com/ubuntu groovy-updates InRelease
#7 0.231   Something wicked happened resolving 'archive.ubuntu.com:80' (-5 - No address associated with hostname)
#7 0.231 Err:3 http://security.ubuntu.com/ubuntu groovy-security InRelease
#7 0.231   Something wicked happened resolving 'security.ubuntu.com:80' (-5 - No address associated with hostname)
#7 0.232 Err:4 http://archive.ubuntu.com/ubuntu groovy-backports InRelease
#7 0.232   Something wicked happened resolving 'archive.ubuntu.com:80' (-5 - No address associated with hostname)
...<redacted>...
```

base image is built successfully

This failed for me on my Google corp workstation, somehow works fine on Mac.

This is most likely coming from us overwriting some of the `/etc` system files that we copy before running `clean-install`

https://github.com/kubernetes-sigs/kind/blob/426e03350adbe02b65cea8b4ddcfecd8b8119093/images/base/Dockerfile#L59

~wasn't able to pinpoint which one exactly, best guess is https://github.com/kubernetes-sigs/kind/blob/main/images/base/files/etc/sysctl.d/10-network-security.conf affecting the networking somehow?~

Some weird interaction with `/` being the destination ?

**Environment:**

- kind version: (use `kind version`): HEAD
- Kubernetes version: (use `kubectl version`): N/A
- Docker version: (use `docker info`): 20.10.2
- OS (e.g. from `/etc/os-release`):  Debian GNU/Linux

/cc @BenTheElder @aojea

amwat:
fixes #2138

Tested this fix locally and it works on my workstation now, and still works on Mac.

Not sure if we want to reorganize this in some other way to reduce layers

/cc @BenTheElder @aojea

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -55,8 +55,17 @@ ARG CNI_PLUGINS_AMD64_SHA256SUM="962100bbc4baeaaa5748cdbfce941f756b1531c2eadb290
 ARG CNI_PLUGINS_ARM64_SHA256SUM="ef17764ffd6cdcb16d76401bac1db6acc050c9b088f1be5efa0e094ea3b01df0"
 ARG CNI_PLUGINS_PPC64LE_SHA256SUM="5bd3c82ef248e5c6cc388f25545aa5a7d318778e5f9bc0a31475361bb27acefe"
 
-# copy in static files (configs, scripts)
-COPY files/ /
+# copy in static files
+# all scripts are 0755: http://www.filepermissions.com/file-permission/0755
+COPY --chmod=0755 files/usr/local/bin/* /usr/local/bin/
+
+# all configs are 0644: http://www.filepermissions.com/file-permission/0644
+COPY --chmod=0644 files/etc/* /etc/
+COPY --chmod=0644 files/etc/containerd/* /etc/containerd/
+COPY --chmod=0644 files/etc/default/* /etc/default/
+COPY --chmod=0644 files/etc/sysctl.d/* /etc/sysctl.d/
+COPY --chmod=0644 files/etc/systemd/system/* /etc/systemd/system/
+COPY --chmod=0644 files/etc/systemd/system/kubelet.service.d/* /etc/systemd/system/kubelet.service.d/
 
 # Install dependencies, first from apt, then from release tarballs.
 # NOTE: we use one RUN to minimize layers.
@@ -87,9 +96,7 @@ COPY files/ /
 # Finally we adjust tempfiles cleanup to be 1 minute after "boot" instead of 15m
 # This is plenty after we've done initial setup for a node, but before we are
 # likely to try to export logs etc.
-RUN echo "Ensuring scripts are executable ..." \
-    && chmod +x /usr/local/bin/clean-install /usr/local/bin/entrypoint \
- && echo "Installing Packages ..." \
+RUN echo "Installing Packages ..." \
     && DEBIAN_FRONTEND=noninteractive clean-install \
       systemd \
       conntrack iptables iproute2 ethtool socat util-linux mount ebtables udev kmod \
--- a/pkg/build/nodeimage/defaults.go
+++ b/pkg/build/nodeimage/defaults.go
@@ -20,4 +20,4 @@ package nodeimage
 const DefaultImage = "kindest/node:latest"
 
 // DefaultBaseImage is the default base image used
-const DefaultBaseImage = "kindest/base:v20210315-64ac2e7f@sha256:d93111ce508dd3134141449d0658de71b6789c01ad73aa340c98b7e640e74ed8"
+const DefaultBaseImage = "kindest/base:v20210318-4009e6fd@sha256:c3c71ab9ad9c4f0e5fd8e3909854a4b8a3b7f50d39f3b814771d4e35f09a0e8f"


images/base/Dockerfile
```diff
@@ -95,8 +95,12 @@ RUN echo "Ensuring scripts are executable ..." \
...
       libseccomp2 pigz \
       bash ca-certificates curl rsync \
-      nfs-common \
-    && find /lib/systemd/system/sysinit.target.wants/ -name "systemd-tmpfiles-setup.service" -delete \
+      nfs-common
+
+# copy in static files (configs)
+COPY files/etc/ /etc/
+
+RUN echo "Cleanup temporary files ..." && find /lib/systemd/system/sysinit.target.wants/ -name "systemd-tmpfiles-setup.service" -delete \
```
BenTheElder, line 103: this step cannot be separate, otherwise the deleted files are still in the layer below.

amwat: Ack, I guess I can change on the next step instead cause we need the configs before enabling kubelet.

aojea:
I can't see any other way to reduce layers than using ADD with a tarball or multistage, the squash flag is still "experimental"

amwat:
updated to naively match 
https://github.com/kubernetes-sigs/kind/issues/2138#issuecomment-802367108

maybe we can reduce some of these?

images/base/Dockerfile
```diff
@@ -88,8 +102,7 @@ COPY files/ /
 # This is plenty after we've done initial setup for a node, but before we are
 # likely to try to export logs etc.
 RUN echo "Ensuring scripts are executable ..." \
```
BenTheElder, line 104: this log should be dropped

images/base/Dockerfile
```diff
@@ -55,8 +55,22 @@ ARG CNI_PLUGINS_AMD64_SHA256SUM="962100bbc4baeaaa5748cdbfce941f756b1531c2eadb290
...
+COPY --chmod=0755 files/etc/systemd/system  /etc/systemd/
+COPY --chmod=0755 files/etc/systemd/system/kubelet.service.d  /etc/systemd/system/
+
+# all configs are 0644: http://www.filepermissions.com/file-permission/0644
+COPY --chmod=0644 files/etc/* /etc/
+COPY --chmod=0644 files/etc/containerd/* /etc/containerd/
+COPY --chmod=0644 files/etc/default/* /etc/default/
+COPY --chmod=0644 files/etc/sysctl.d/* /etc/sysctl.d/
+COPY --chmod=0644 files/etc/systemd/system/* /etc/systemd/system/
+COPY --chmod=0644 files/etc/systemd/system/kubelet.service.d/* /etc/systemd/system/kubelet.service.d/
```
BenTheElder, line 73: this wound up really verbose.
maybe we should script creating a tarball in the makefile and then ADD it :/
at least then we could maybe programattically say "directories are X, files are Y, scripts are Z"

images/base/Dockerfile
amwat: done

images/base/Dockerfile
aojea: :)

BenTheElder: there's some issues that approach too. discussed offline wtih amit, with correct usage of `COPY` (non recursive) directories will default to 755 like we want and it's not more verbose then.

BenTheElder:
/lgtm
/hold
rebase this to HEAD, `make -C images/base push`, then bump the default.

images/base/Dockerfile
aojea: know I was wrong because I was assuming tar from a host, not from files in a git repo, but was surprisingly nice to see you reconsider it XD

amwat:
done bumped to https://hub.docker.com/layers/kindest/base/v20210318-4009e6fd/images/sha256-f9026ddd33da088a9f81a41393189fc541485f8e1d4caad82a4d36ebfb3620d4?context=explore

pkg/build/nodeimage/defaults.go
```diff
@@ -20,4 +20,4 @@ package nodeimage
 const DefaultImage = "kindest/node:latest"
 
 // DefaultBaseImage is the default base image used
-const DefaultBaseImage = "kindest/base:v20210315-64ac2e7f@sha256:d93111ce508dd3134141449d0658de71b6789c01ad73aa340c98b7e640e74ed8"
+const DefaultBaseImage = "kindest/base:v20210318-4009e6fd@sha256:f9026ddd33da088a9f81a41393189fc541485f8e1d4caad82a4d36ebfb3620d4"
```
BenTheElder, line 23: this should be sha256:c3c71ab9ad9c4f0e5fd8e3909854a4b8a3b7f50d39f3b814771d4e35f09a0e8f
```
crane digest kindest/base:v20210318-4009e6fd
sha256:c3c71ab9ad9c4f0e5fd8e3909854a4b8a3b7f50d39f3b814771d4e35f09a0e8f
```

the reason is multi-arch manifest digest vs the digest for just the amd64 image

amwat: done, this isn't visible on dockerhub anywhere?

BenTheElder: yeah, it isn't, someone at docker was going to try get that re-prioritized recently.

aojea:
this was an impressive invetigation :muscle:

Issue #2142, opened 2021-03-19
Podman error with IP/Port format


christianh814:
**What happened**:

When running `KIND_EXPERIMENTAL_PROVIDER=podman kind create cluster --config=kind.yaml` I get the following error

```shell
enabling experimental podman provider
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.20.2) 🖼 
 ✓ Preparing nodes 📦  
 ✓ Writing configuration 📜 
 ✓ Starting control-plane 🕹️ 
 ✓ Installing CNI 🔌 
 ✓ Installing StorageClass 💾 
ERROR: failed to create cluster: network details should be in the format IP/Port, received: [ 443127.0.0.1 42521 80]
```

This is the config I'm using

```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
  extraPortMappings:
  - containerPort: 80
    hostPort: 80
    listenAddress: 0.0.0.0
  - containerPort: 443
    hostPort: 443
    listenAddress: 0.0.0.0
```

**What you expected to happen**:

The `kind` cluster to come up with the portmappings using `podman` as the runtime.

**How to reproduce it (as minimally and precisely as possible)**:

1. Install podman v3
2. Install kind
3. Create a config with portmappings
4. run kind

**Anything else we need to know?**:

This same config works fine using `docker`

**Environment:**

- kind version: (use `kind version`): 

```shell
$ kind --version
kind version 0.10.0
```

Also broken using alpha

```shell
$ kind version
kind v0.11.0-alpha go1.15.6 linux/amd64
```

- Kubernetes version: (use `kubectl version`):

```shell
kubectl version
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.4", GitCommit:"e87da0bd6e03ec3fea7933c4b5263d151aafd07c", GitTreeState:"clean", BuildDate:"2021-02-18T16:12:00Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}
```
Kindest image

```
kindest/node:v1.20.2
```

- Podman version: (use `podman info`):

```shell
$ podman info
host:
  arch: amd64
  buildahVersion: 1.19.4
  cgroupManager: systemd
  cgroupVersion: v1
  conmon:
    package: conmon-2.0.26-1.fc33.x86_64
    path: /usr/bin/conmon
[...]

PR #2144, opened 2021-03-19
restore podman portmap check for 3.0.0+

aojea:
podman inspect was broken between 2.2.0 and 3.0.0, newer releases
doesn't have the problem so we don't have to use the hack to check
the portmapped ports.

This keeps podman versions between 2.2.0 and 3.0.0 broken for custom portmaps, but it is really a podman bug, we can always ask people to use a podman version without the bug

Fixes: https://github.com/kubernetes-sigs/kind/issues/2142

### DIFF
--- a/pkg/cluster/internal/providers/podman/provider.go
+++ b/pkg/cluster/internal/providers/podman/provider.go
@@ -186,13 +186,19 @@ func (p *provider) GetAPIServerEndpoint(cluster string) (string, error) {
 	if err != nil {
 		return "", errors.Wrap(err, "failed to check podman version")
 	}
-	if v.LessThan(version.MustParseSemantic("2.2.0")) {
+	// podman inspect was broken between 2.2.0 and 3.0.0
+	// https://github.com/containers/podman/issues/8444
+	if v.AtLeast(version.MustParseSemantic("2.2.0")) &&
+		v.LessThan(version.MustParseSemantic("3.0.0")) {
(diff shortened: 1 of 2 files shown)


pkg/cluster/internal/providers/podman/provider.go
```diff
@@ -186,7 +186,10 @@ func (p *provider) GetAPIServerEndpoint(cluster string) (string, error) {
 	if err != nil {
 		return "", errors.Wrap(err, "failed to check podman version")
 	}
-	if v.LessThan(version.MustParseSemantic("2.2.0")) {
+	// podman inspect was broken between 2.2.0 and 3.0.0
+	// https://github.com/containers/podman/issues/8444
+	if v.LessThan(version.MustParseSemantic("2.2.0")) ||
```
amwat, line 191: nit: let's invert the logic and have the "hack" inside the `if` and the actual flow to be the normal flow.

do we also want to output a warning?

BenTheElder: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow +1

amwat (commented):
GitHub diff is messed up, the only change is the 3.0.0 check 
and podman CI looks good on this

/lgtm

aojea:
I'm going to override, the only job failing is totally orthogonal to the PR and will be fixed by https://github.com/kubernetes/test-infra/pull/21482

PR #2145, opened 2021-03-19
actually support merge patches with document streams

BenTheElder:
this explains https://github.com/kubernetes/kubernetes/issues/99802#issuecomment-802997868

PR #2146, opened 2021-03-20
correctly generate newline in e2e-k8s.sh

BenTheElder:
follow up to https://github.com/kubernetes-sigs/kind/pull/2145

shell does not support `\n`. bash can do it with strings prefixed with `$` but that's bash specific. `printf` is POSIX.


BenTheElder:
see: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-kind-e2e-json-logging/1373131865153605632

Issue #2050, opened 2021-02-03
[xfs?] kind create cluster fails on centos8 docker 20.10.3 with timeout


maxant:
**What happened**:
I installed kind v0.10.0 go1.15.7 linux/amd64 on a centos8 server running containerd and docker daemon.
I run `./kind -v 9 create cluster --wait 10m` but after 2 minutes I get a timeout error, see below.

**What you expected to happen**:
No timeout error.

**How to reproduce it (as minimally and precisely as possible)**:

    ./kind -v 9 create cluster --wait 10m

wait... and then:

    Creating cluster "kind" ...
    DEBUG: docker/images.go:58] Image: 
    kindest/node:v1.20.2@sha256:8f7ea6e7642c0da54f04a7ee10431549c0257315b3a634f6ef2fecaaedb19bab present locally
     ✓ Ensuring node image (kindest/node:v1.20.2) 🖼 
     ✓ Preparing nodes 📦  
    DEBUG: config/config.go:90] Using the following kubeadm config for node kind-control-plane:

...

    I0203 21:59:19.491111     212 loader.go:379] Config loaded from file:  /etc/kubernetes/admin.conf
    [wait-control-plane] Waiting for the kubelet to boot up the control plane as static Pods from directory "/etc/kubernetes/manifests". 
    This can take up to 4m0s

... the following is then repeated a lot:

    I0203 22:01:13.999076     212 round_trippers.go:445] GET https://kind-control-plane:6443/healthz?timeout=10s  in 2 milliseconds
    I0203 22:01:14.498120     212 round_trippers.go:445] GET https://kind-control-plane:6443/healthz?timeout=10s  in 1 milliseconds
    [kubelet-check] It seems like the kubelet isn't running or healthy.
    [kubelet-check] The HTTP call equal to 'curl -sSL http://localhost:10248/healthz' failed with error: Get 
    "http://localhost:10248/healthz": dial tcp [::1]:10248: connect: connection refused.

	Unfortunately, an error has occurred:
		timed out waiting for the condition

	This error is likely caused by:
		- The kubelet is not running
		- The kubelet is unhealthy due to a misconfiguration of the node in some way (required cgroups disabled)

	If you are on a systemd-powered system, you can try to troubleshoot the error with the following commands:
		- 'systemctl status kubelet'
		- 'journalctl -xeu kubelet'

	Additionally, a control plane component may have crashed or exited when started by the container runtime.
[...]

PR #2149, opened 2021-03-22
detect docker btrfs backend filesystem

aojea:
Fixes: #1945,#2050

### DIFF
--- a/pkg/cluster/internal/providers/docker/util.go
+++ b/pkg/cluster/internal/providers/docker/util.go
@@ -17,6 +17,7 @@ limitations under the License.
 package docker
 
 import (
+	"encoding/json"
 	"strings"
 
 	"sigs.k8s.io/kind/pkg/exec"
@@ -48,15 +49,39 @@ func usernsRemap() bool {
 }
 
 // mountDevMapper checks if the Docker storage driver is Btrfs or ZFS
+// or if the backing filesystem is Btrfs
 func mountDevMapper() bool {
 	storage := ""
+	// check the docker storage driver
 	cmd := exec.Command("docker", "info", "-f", "{{.Driver}}")
 	lines, err := exec.OutputLines(cmd)
-	if err != nil {
+	if err != nil || len(lines) != 1 {
 		return false
 	}
-	if len(lines) > 0 {
-		storage = strings.ToLower(strings.TrimSpace(lines[0]))
+
+	storage = strings.ToLower(strings.TrimSpace(lines[0]))
+	if storage == "btrfs" || storage == "zfs" || storage == "devicemapper" {
+		return true
 	}
-	return storage == "btrfs" || storage == "zfs"
+
+	// check the backing file system
+	// docker info -f '{{json .DriverStatus  }}'
+	// [["Backing Filesystem","extfs"],["Supports d_type","true"],["Native Overlay Diff","true"]]
+	cmd = exec.Command("docker", "info", "-f", "{{json .DriverStatus }}")
+	lines, err = exec.OutputLines(cmd)
+	if err != nil || len(lines) != 1 {
+		return false
+	}
+	var dat [][]string
+	if err := json.Unmarshal([]byte(lines[0]), &dat); err != nil {
+		return false
+	}
+	for _, item := range dat {
+		if item[0] == "Backing Filesystem" {
+			storage = strings.ToLower(item[1])
+			break
+		}
+	}
+
+	return storage == "btrfs" || storage == "zfs" || storage == "xfs"
 }


aojea:
/assign @BenTheElder 
@anfredette can you please verify this solves the issue for fedora33?

pkg/cluster/internal/providers/docker/util.go
```diff
@@ -48,15 +48,29 @@ func usernsRemap() bool {
...
+	// docker info -f '{{json .DriverStatus  }}'
+	// [["Backing Filesystem","extfs"],["Supports d_type","true"],["Native Overlay Diff","true"]]
+	cmd = exec.Command("docker", "info", "-f", "{{ index (index .DriverStatus 0) 1 }}")
+	lines, err = exec.OutputLines(cmd)
+	if err != nil || len(lines) != 1 {
+		return false
 	}
-	return storage == "btrfs" || storage == "zfs"
+	storage = strings.ToLower(strings.TrimSpace(lines[0]))
+	return storage == "btrfs"
```
BenTheElder, line 75: We also should do this for xfs, zfs.

aojea: in the driver too?, we weren´t doing it for xfs

pkg/cluster/internal/providers/docker/util.go
```diff
@@ -48,15 +48,29 @@ func usernsRemap() bool {
...
 	lines, err := exec.OutputLines(cmd)
-	if err != nil {
+	if err != nil || len(lines) != 1 {
 		return false
 	}
-	if len(lines) > 0 {
-		storage = strings.ToLower(strings.TrimSpace(lines[0]))
+
+	storage = strings.ToLower(strings.TrimSpace(lines[0]))
+	if storage == "btrfs" || storage == "zfs" {
```
aojea, line 62: do I need to change this too @BenTheElder ?

anfredette (approved):
I'm running on a fresh install of Fedora 33 Workstation in which btrfs replaced ext4 as the default filesystem.

kind was failing to start with the following error:
ERROR: failed to create cluster: failed to init node with kubeadm: command "docker exec --privileged ovn-control-plane kubeadm init --skip-phases=preflight --config=/kind/kubeadm.conf --skip-token-print --v=6" failed with error: exit status 1

For reference:
$ docker info --format='{{ index (index .DriverStatus 0) 1 }}'
btrfs

This change fixes my issue.

pkg/cluster/internal/providers/docker/util.go
BenTheElder: we should add `devicemapper` here https://github.com/kubernetes-sigs/kind/issues/2050#issue-800750520

aojea: done

pkg/cluster/internal/providers/docker/util.go
aojea: done

pkg/cluster/internal/providers/docker/util.go
```diff
@@ -48,15 +48,29 @@ func usernsRemap() bool {
...
+
+	storage = strings.ToLower(strings.TrimSpace(lines[0]))
+	if storage == "btrfs" || storage == "zfs" || storage == "devicemapper" {
+		return true
+	}
+
+	// check the backing file system
+	// docker info -f '{{json .DriverStatus  }}'
+	// [["Backing Filesystem","extfs"],["Supports d_type","true"],["Native Overlay Diff","true"]]
+	cmd = exec.Command("docker", "info", "-f", "{{ index (index .DriverStatus 0) 1 }}")
```
BenTheElder, line 69: do we have reason to believe this is always index 0 of DriverStatus?

aojea: I tested it in 3 different deployments, maybe @AkihiroSuda can help us here ^^^

BenTheElder: This seems unnecessarily fragile we can decode the structure and inspect the tuple with "Backing Filesystem"

pkg/cluster/internal/providers/docker/util.go
```diff
@@ -48,15 +49,39 @@ func usernsRemap() bool {
...
+	if err != nil || len(lines) != 1 {
+		return false
+	}
+	var dat [][]string
+	if err := json.Unmarshal([]byte(lines[0]), &dat); err != nil {
+		return false
+	}
+	for _, item := range dat {
+		if item[0] == "Backing Filesystem" {
+			storage = strings.ToLower(strings.TrimSpace(item[1]))
```
BenTheElder, line 81: i'm not sure we should be trimspace etc. here, but that can be tweaked in a follow-up

aojea: let's do it now

BenTheElder:
/lgtm
/approve
thanks!

PR #2153, opened 2021-03-23
use docker0 mtu for kind network

aojea:
Use the docker default network mtu for the kind network

### DIFF
--- a/pkg/cluster/internal/providers/docker/network.go
+++ b/pkg/cluster/internal/providers/docker/network.go
@@ -21,10 +21,12 @@ import (
 	"crypto/sha1"
 	"encoding/binary"
 	"encoding/json"
+	"fmt"
 	"io"
 	"net"
 	"regexp"
 	"sort"
+	"strconv"
 	"strings"
 
 	"sigs.k8s.io/kind/pkg/errors"
@@ -61,9 +63,11 @@ func ensureNetwork(name string) error {
 
 	// Generate unique subnet per network based on the name
 	// obtained from the ULA fc00::/8 range
+	// Use the MTU configured for the docker default network
 	// Make N attempts with "probing" in case we happen to collide
 	subnet := generateULASubnetFromName(name, 0)
-	err = createNetworkNoDuplicates(name, subnet)
+	mtu := getDefaultNetworkMTU()
+	err = createNetworkNoDuplicates(name, subnet, mtu)
 	if err == nil {
 		// Success!
 		return nil
@@ -75,7 +79,7 @@ func ensureNetwork(name string) error {
 	// If it is, make more attempts below
 	if isIPv6UnavailableError(err) {
 		// only one attempt, IPAM is automatic in ipv4 only
-		return createNetworkNoDuplicates(name, "")
+		return createNetworkNoDuplicates(name, "", mtu)
 	}
 	if isPoolOverlapError(err) {
 		// pool overlap suggests perhaps another process created the network
@@ -97,7 +101,7 @@ func ensureNetwork(name string) error {
 	const maxAttempts = 5
 	for attempt := int32(1); attempt < maxAttempts; attempt++ {
 		subnet := generateULASubnetFromName(name, attempt)
-		err = createNetworkNoDuplicates(name, subnet)
+		err = createNetworkNoDuplicates(name, subnet, mtu)
 		if err == nil {
 			// success!
 			return nil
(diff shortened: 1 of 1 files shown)


aojea:
I tested changing the mtu in the /etc/docker/daemon.json

```
$ docker network inspect bridge -f '{{ index .Options "com.docker.network.driver.mtu" }}'
1500
$ docker network rm kind
kind
$ sudo vi /etc/docker/daemon.json 
$ sudo systemctl restart docker
$ docker network inspect bridge -f '{{ index .Options "com.docker.network.driver.mtu" }}'
1450
$ ./kind create cluster
$ docker network inspect kind -f '{{ index .Options "com.docker.network.driver.mtu" }}'
1450
```

and inside the cluster both the node and the pods have the correct mtu
```
docker exec -it kind-control-plane ip link show
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
2: vethf10e3ec8@if3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1450 qdisc noqueue state UP mode DEFAULT group default 
    link/ether f6:9c:bf:98:8c:10 brd ff:ff:ff:ff:ff:ff link-netns cni-4add33c0-01e5-e622-e1bf-66fff8a9d608
3: veth2a7467bf@if3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1450 qdisc noqueue state UP mode DEFAULT group default 
    link/ether 22:a5:f1:58:57:27 brd ff:ff:ff:ff:ff:ff link-netns cni-2e60c737-3233-a3f7-626f-81421fff352c
4: veth74c4d25a@if3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1450 qdisc noqueue state UP mode DEFAULT group default 
    link/ether 4e:e9:2a:82:0e:ce brd ff:ff:ff:ff:ff:ff link-netns cni-3f8fe429-0acc-b62f-be57-c9c68241cef3
94: eth0@if95: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1450 qdisc noqueue state UP mode DEFAULT group default 
    link/ether 02:42:ac:13:00:02 brd ff:ff:ff:ff:ff:ff link-netnsid 0
```

aojea:
hmm, this error is new 
```
DEBUG: docker/pull.go:44] Pulling image: rancher/local-path-provisioner:v0.0.14 ...
ctr: failed to dial "/run/containerd/containerd.sock": connection error: desc = "transport: error while dialing: dial unix /run/containerd/containerd.sock: connect: connection refused"
ctr: failed to dial "/run/containerd/containerd.sock": connection error: desc = "transport: error while dialing: dial unix /run/containerd/containerd.sock: connect: connection refused"
read unix @->/var/run/docker.sock: read: connection reset by peer
Image build Failed! Failed to load images [command "docker exec --privileged -i kind-build-1616536576-1580108784 ctr --namespace=k8s.io images import --all-platforms --no-unpack -" failed with error: exit status 1, write |1: broken pipe]
ctr: transport is closing: unavailable
```

hellt:
Hi @aojea 
I think checking the `bridge` network for it's MTU is not reliable, since by default the MTU option is not visible:

```
docker network create mtu1500
268f782b9b1955ea70044e90c392848c1049282f38f7b144f4cbed8f0c3dc129


❯ docker network inspect 268f782b9b1955ea70044e90c392848c1049282f38f7b144f4cbed8f0c3dc129
[
    {
        "Name": "mtu1500",
        "Id": "268f782b9b1955ea70044e90c392848c1049282f38f7b144f4cbed8f0c3dc129",
        "Created": "2021-03-24T06:44:03.595323038Z",
        "Scope": "local",
        "Driver": "bridge",
        "EnableIPv6": false,
        "IPAM": {
            "Driver": "default",
            "Options": {},
            "Config": [
                {
                    "Subnet": "172.25.0.0/16",
                    "Gateway": "172.25.0.1"
                }
            ]
        },
        "Internal": false,
        "Attachable": false,
        "Ingress": false,
        "ConfigFrom": {
            "Network": ""
        },
        "ConfigOnly": false,
        "Containers": {},
        "Options": {},
        "Labels": {}
    }
]


❯ docker network inspect 268f782b9b1955ea70044e90c392848c1049282f38f7b144f4cbed8f0c3dc129 -f '{{ index .Options "com.docker.network.driver.mtu" }}'


```

I think a better option would be to check the MTU of the `docker0` interface, which will always report its value.

aojea:
> I think checking the `bridge` network for it's MTU is not reliable, since by default the MTU option is not visible for custom networks:

the `bridge` network is the default, is kind of special, it is always created (it always showed the MTU to me in different installation, but that is not enough evidence, I know.) Also, if there is no mtu reported it means default, and what we do here is just not simply set the option so docker will use its default.
I think that the question here is who is the source of truth, docker or the network interface, and I personally think that we should keep having docker as source of truth.
If we look for the interface mtu, people can change the interface MTU without changing docker, and then weird things start to happen XD

BenTheElder:
the default network is unlikely to change in a breaking way. `bridge` is special not just in being the default but in that the behavior changes of custom networks is not backported to bridge.

after this PR the default bridge MTU still requires configuring by the user in docker options, but in a docker host that needs custom MTU I think it's reasonable to expect the user to configure this. and docker exposes a system config option for this one.

/lgtm

Issue #2063, opened 2021-02-09
How to build a node image with preloaded docker images 


hatzhang:
Is there a way to build a node image with some preloaded docker images/tars, then when the cluster is created with ``` kind create cluster --image ```, those docker images will need not be pulled inside or loaded into kind.

BenTheElder:
This is not supported currently. It's possible to reproduce what kind does internally, but you'd have to do it yourself, and it's going to be clunky and brittle (you can't do this in a normal `docker build` for example).

I recommend just using a local registry, or saving all the images you want to a tarball (e.g. `docker save`) so you can load them in one-shot with `kind load image-archive`.

dkoshkin:
I have the same use-case, for air-gapped installations it would be great to be able to build a custom image with additional images pre-loaded.
I would be happy to work on this, please let me know.

BenTheElder:
see: https://github.com/kubernetes-sigs/kind/issues/2160 for current discussion 😅

nitishm:
I noticed this PR as well - https://github.com/kubernetes-sigs/kind/pull/2303. Wonder if this might be something that I could try? I essentially want to prime by kind node with some docker images to speed up my CI pipeline (since pulling images takes a lot of time at the moment leading to slow tests)

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/2303 is the current in-review PR for issue #2160.

BenTheElder:
You will still have to pull down the content of course, the total download size is not easily decreased, but it can be done at a different point in the process (arguably one that is less concurrent and not necessarily net improvement in performance, unless you are managing to cache images at the CI level, in which case it may be a big win)

PR #2159, opened 2021-03-25
add additional-images flag to build command

dkoshkin:
A solution for https://github.com/kubernetes-sigs/kind/issues/2063

Add a way to specify additional images to pull when running `kind build node-image` with a new flag `--additional-images`.

My use-case is to have a `kind` image that is used in an air-gapped cluster that includes a few additional images.
There is no easy way to extend the existing `kindest/node` images (please correct me here if there is some process), `kind load` is an option but its less desirable for my use case since it would require to package and ship around multiple images instead of just the one `kindest/node` image.

### DIFF
--- a/pkg/build/nodeimage/build.go
+++ b/pkg/build/nodeimage/build.go
@@ -83,9 +83,10 @@ func supportedArch(arch string) bool {
 // build configuration
 type buildContext struct {
 	// option fields
-	image     string
-	baseImage string
-	logger    log.Logger
+	image                  string
+	baseImage              string
+	logger                 log.Logger
+	additionalImagesToPull []string
 	// non-option fields
 	arch     string // TODO(bentheelder): this should be an option
 	kubeRoot string
--- a/pkg/build/nodeimage/build_impl.go
+++ b/pkg/build/nodeimage/build_impl.go
@@ -255,6 +255,9 @@ func (c *buildContext) prePullImages(bits kube.Bits, dir, containerID string) ([
 	// all builds should install the default storage driver images currently
 	requiredImages = append(requiredImages, defaultStorageImages...)
 
+	// include additional images passed in by the user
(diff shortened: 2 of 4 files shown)


BenTheElder:
Please file a feature request and discuss a design first: https://kind.sigs.k8s.io/docs/contributing/getting-started/#4-reaching-out
(we are looking at instituting a better process for recording proposals, but in the meantime we use feature issues)

Some questions to consider:
- should we pull images if they're not present locally?
  - if we're pulling images, should we just pull them directly to containerd instead of to the host?
- what about podman?
- what about tarballs / image archives?
- what about image eviction in the future? should kind be responsible for somehow preventing these images from being evicted if we re-enable eviction once we have a solution for the core images?
- could we instead support this in a way that involved extending an existing node image without having to build from source? would that be better?

dkoshkin:
Hi @BenTheElder thanks for the quick reply!

Sorry should have dug around the docs more to find the contributing process, as you may have guessed this is my first attempt at a contribution to kind 😰

I created a feature ticket https://github.com/kubernetes-sigs/kind/issues/2160 so I can close this and move the discussion there?

BenTheElder:
No worries! We need to put more effort into surfacing this and improving the process 😅 

Apologies in advance for the scattered response latency, we're trying to wrap up some things for the next release and all of us work on other projects as well (upstream kubernetes in particular). I know Antonio and I also have a little bit of time out / vacation intermittently in the next two weeks.

BenTheElder:
I think https://github.com/kubernetes-sigs/kind/pull/2303 will be a more powerful alternative based on the discussion in the https://github.com/kubernetes-sigs/kind/issues/2160. (and save people the trouble of having to build a node image again)

BenTheElder:
thanks again for the PR! tentatively closing in favor of the other approach

PR #2161, opened 2021-03-26
fix kind export logs

BenTheElder:
- export kind-version.txt at the top level, instead of per node. it's not a property of the node
- always print the exported path so any logs that are exported can be debugged even if an error is encountered

### DIFF
--- a/pkg/cmd/kind/export/logs/logs.go
+++ b/pkg/cmd/kind/export/logs/logs.go
@@ -80,12 +80,11 @@ func runE(logger log.Logger, streams cmd.IOStreams, flags *flagpole, args []stri
 		dir = args[0]
 	}
 
-	// collect the logs
-	if err := provider.CollectLogs(flags.Name, dir); err != nil {
-		return err
-	}
-
-	logger.V(0).Infof("Exported logs for cluster %q to:", flags.Name)
+	// NOTE: the path is the output of this command to be captured by calling tools
+	// whereas "Exporting logs..." is info / debug (stderr)
+	logger.V(0).Infof("Exporting logs for cluster %q to:", flags.Name)
 	fmt.Fprintln(streams.Out, dir)
-	return nil
+
+	// collect the logs
+	return provider.CollectLogs(flags.Name, dir)
 }
--- a/pkg/cluster/internal/providers/common/logs.go
+++ b/pkg/cluster/internal/providers/common/logs.go
@@ -5,7 +5,6 @@ import (
 	"path/filepath"
 
 	"sigs.k8s.io/kind/pkg/cluster/nodes"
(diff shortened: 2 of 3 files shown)


BenTheElder:
working as expected:
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/2161/pull-kind-e2e-kubernetes-1-20/1375330992604057600/artifacts/logs/kind-version.txt

pkg/cmd/kind/export/logs/logs.go
```diff
@@ -80,12 +80,11 @@ func runE(logger log.Logger, streams cmd.IOStreams, flags *flagpole, args []stri
...
 
-	// collect the logs
-	if err := provider.CollectLogs(flags.Name, dir); err != nil {
-		return err
-	}
-
-	logger.V(0).Infof("Exported logs for cluster %q to:", flags.Name)
+	// NOTE: the path is the output of this command to be captured by calling tools
+	// whereas "Exporting logs..." is info / debug (stderr)
+	logger.V(0).Infof("Exporting logs for cluster %q to:", flags.Name)
```
aojea, line 85: to: %s", dir)  ??

BenTheElder: no, we print the path to stdout for capture by the calling program.
we print the message to stderr.
(see the comment above)

BenTheElder: e.g.:
```console
$ kind export logs
Exporting logs for cluster "kind" to: 
/private/var/folders/rt/xdpktcmj0p7fx3kvwmrgdqyh00fsfv/T/609940926
ERROR: [command "docker exec --privileged kind-control-plane sh -c 'tar --hard-dereference -C /var/log/ -chf - . || (r=$?; [ $r -eq 1 ] || exit $r)'" failed with error: exit status 1, [command "docker exec --privileged kind-control-plane cat /kind/version" failed with error: exit status 1, command "docker exec --privileged kind-control-plane journalctl --no-pager -u kubelet.service" failed with error: exit status 1, command "docker exec --privileged kind-control-plane journalctl --no-pager -u containerd.service" failed with error: exit status 1, command "docker exec --privileged kind-control-plane journalctl --no-pager" failed with error: exit status 1]]
```

the stdout part is `/private/var/folders/rt/xdpktcmj0p7fx3kvwmrgdqyh00fsfv/T/609940926`

BenTheElder: note: in this commit I basically just moved this up before we actually export (so if we hit an error you can still see where the logs we did collect are) and s/Exported/Exporting/

Issue #1975, opened 2020-12-12
coredns CrashLoopBackOff on Ubuntu 20.04


paolomainardi:
**What happened**:

```
❯ k get pods
NAME                                                       READY   STATUS             RESTARTS   AGE
coredns-66bff467f8-j5cf4                                   0/1     CrashLoopBackOff   1          54s
coredns-66bff467f8-l6gtz                                   0/1     CrashLoopBackOff   1          54s
etcd-retrogames-k8s-dev-control-plane                      1/1     Running            0          66s
kindnet-wgxw8                                              1/1     Running            0          54s
kube-apiserver-retrogames-k8s-dev-control-plane            1/1     Running            0          66s
kube-controller-manager-retrogames-k8s-dev-control-plane   1/1     Running            0          66s
kube-proxy-nnkwz                                           1/1     Running            0          54s
kube-scheduler-retrogames-k8s-dev-control-plane            1/1     Running            0          66s
```

**What you expected to happen**:

Coredns should work.

**How to reproduce it (as minimally and precisely as possible)**:

```
❯ cat /etc/issue
Ubuntu 20.04.1 LTS \n \l
```
```
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
  image: kindest/node:v1.19.1@sha256:98cf5288864662e37115e362b23e4369c8c4a408f99cbc06e58ac30ddc721600
  kubeadmConfigPatches:
  - |
    kind: InitConfiguration
    nodeRegistration:
      kubeletExtraArgs:
        node-labels: "ingress-ready=true"
  extraPortMappings:
  - containerPort: 80
    hostPort: 80
    protocol: TCP
  - containerPort: 443
    hostPort: 443
    protocol: TCP
```
```
❯ docker exec -it retrogames-k8s-dev-control-plane cat /etc/resolv.conf
search homenet.telecomitalia.it
nameserver 127.0.0.1
options ndots:0
```

**Environment:**

- kind version: `kind v0.9.0 go1.15.2 linux/amd64`
- Kubernetes version:
```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.4", GitCommit:"d360454c9bcd1634cf4cc52d1867af5491dc9c5f", GitTreeState:"clean", BuildDate:"2020-11-11T13:17:17Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #2165, opened 2021-03-27
don't use loopback addresses on the nodes resolv.conf file

aojea:
we were using the well known name host.docker.internal to get the
host IP address, however, it seems that in some systems it resolves
to a loopback address. We use this address to replace the embedded
docker dns IP address in the resolv.conf.
We can't use a loopback address, because coredns obtains the upstream
dns resolvers from the resolv.conf, and evidentelly, it crashes if the
upstream DNS resolver is a loopback, because it meains that is itself,
and that is a loop.

Fixes: #1975

### DIFF
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -337,7 +337,8 @@ enable_network_magic(){
   # first we need to detect an IP to use for reaching the docker host
   local docker_host_ip
   docker_host_ip="$( (getent ahostsv4 'host.docker.internal' | head -n1 | cut -d' ' -f1) || true)"
-  if [[ -z "${docker_host_ip}" ]]; then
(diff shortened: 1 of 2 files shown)


BenTheElder:
PR title seems inaccurate.

aojea:
> PR title seems inaccurate.

oh yeah, because at first I though that we could use directly that, until I realise window and mac needs to resolve

BenTheElder:
That test sounds bogus with concurrent watches 🤔
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/2165/pull-kind-conformance-parallel-ga-only/1376070430464413696
/retest

aojea:
/retest
> Kubernetes e2e suite: [sig-apps] ReplicaSet Replace and Patch tests [Conformance] expand_less

Issue #2172, opened 2021-04-01
dualstack should automatically enable feature gate for < 1.21


BenTheElder:
see: https://kubernetes.slack.com/archives/CEKK1KTN2/p1617232113200400

note: we can be smart and enable this _precisely_ before it comes on by default 🙃, see e.g.  https://github.com/kubernetes-sigs/kind/commit/a4df15b6f77e65fa3598e87f139071f8697ec625#diff-1d8da887aaff57615ed7bc578730aff0c89074cb92ba7f5ef2bcb6536f9ae521R110

filing so we don't lose track, I can do this in the morning though, been working on #166.

cc @aojea

BenTheElder:
also: we need to get a kind CI job. maybe just a small subset of tests on kubernetes master in kind presubmit for now (we need to avoid excessively expanding presubmit in this repo and we already subject ourselves to equivalents of all kubernetes/kubernetes presubmits using kind).

aojea:
> also: we need to get a kind CI job. maybe just a small subset of tests on kubernetes master in kind presubmit for now (we need to avoid excessively expanding presubmit in this repo and we already subject ourselves to equivalents of all kubernetes/kubernetes presubmits using kind).

There are 2 periodic KIND dual-stack jobs (ipvs and iptables proxy) and 2 optional presubmits already
https://testgrid.k8s.io/sig-network-kind#sig-network-kind,%20dual,%20master&width=20
https://testgrid.k8s.io/sig-network-kind#pr-sig-network-kind,%20dual&width=20

PR #2173, opened 2021-04-01
Dual stack follow-ups

aojea:
Fix documentation and workaround 1.20 versions to enable the feature gate

This needs an image bump for having the kindnet dualstack versions, so pods can get dual-stack addresses

Fixes: #2172

### DIFF
--- a/pkg/cluster/internal/kubeadm/config.go
+++ b/pkg/cluster/internal/kubeadm/config.go
@@ -25,6 +25,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/util/version"
 	"sigs.k8s.io/kind/pkg/errors"
+	"sigs.k8s.io/kind/pkg/internal/apis/config"
 )
 
 // ConfigData is supplied to the kubeadm config template, with values populated
@@ -67,8 +68,8 @@ type ConfigData struct {
 	// Kubernetes API Server RuntimeConfig
 	RuntimeConfig map[string]string
 
-	// IPv4 values take precedence over IPv6 by default, if true set IPv6 default values
-	IPv6 bool
+	// IPFamily of the cluster, it can be IPv4, IPv6 or DualStack
+	IPFamily config.ClusterIPFamily
 
 	// Labels are the labels, in the format "key1=val1,key2=val2", with which the respective node will be labeled
 	NodeLabels string
@@ -96,6 +97,10 @@ type DerivedConfigData struct {
 	FeatureGatesString string
 	// RuntimeConfigString is of the form `Foo=true,Baz=false`
 	RuntimeConfigString string
+	// KubeadmFeatureGates contains Kubeadm only feature gates
+	KubeadmFeatureGates map[string]bool
+	// IPv4 values take precedence over IPv6 by default, if true set IPv6 default values
+	IPv6 bool
 }
 
 // Derive automatically derives DockerStableTag if not specified
@@ -106,6 +111,9 @@ func (c *ConfigData) Derive() {
 	if c.DockerStableTag == "" {
 		c.DockerStableTag = strings.Replace(c.KubernetesVersion, "+", "_", -1)
 	}
+
+	// get the IP addresses family for defaulting components
+	c.IPv6 = c.IPFamily == config.IPv6Family
 
 	// get sorted list of FeatureGate keys
 	featureGateKeys := make([]string, 0, len(c.FeatureGates))
@@ -281,6 +289,10 @@ metadata:
   name: config
 kubernetesVersion: {{.KubernetesVersion}}
 clusterName: "{{.ClusterName}}"
+{{ if .KubeadmFeatureGates}}featureGates:
+{{ range $key, $value := .KubeadmFeatureGates }}
+  "{{ $key }}": {{ $value }}
+{{end}}{{end}}
 controlPlaneEndpoint: "{{ .ControlPlaneEndpoint }}"
 # on docker for mac we have to expose the api server via port forward,
 # so we need to ensure the cert is valid for localhost so we can talk
@@ -438,6 +450,17 @@ func Config(data ConfigData) (config string, err error) {
 
 	// derive any automatic fields if not supplied
 	data.Derive()
+
+	// Kubeadm has its own feature-gate for dual stack
+	// we need to enable it for Kubernetes version 1.20 only
+	// dual-stack is only supported in 1.20+
+	// TODO: remove this when 1.20 is EOL or we no longer support
+	// dual-stack for 1.20 in KIND
+	if ver.LessThan(version.MustParseSemantic("v1.21.0")) &&
+		ver.AtLeast(version.MustParseSemantic("v1.20.0")) {
+		data.KubeadmFeatureGates = make(map[string]bool)
+		data.KubeadmFeatureGates["IPv6DualStack"] = true
+	}
 
 	// execute the template
 	var buff bytes.Buffer
--- a/site/content/docs/user/configuration.md
+++ b/site/content/docs/user/configuration.md
@@ -102,9 +102,35 @@ Multiple details of the cluster's networking can be customized under the
 
 #### IP Family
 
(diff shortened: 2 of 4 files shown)


aojea:
/hold 
it needs images bumps

pkg/cluster/internal/create/create.go
```diff
@@ -255,3 +262,64 @@ func validateProvider(p providers.Provider) error {
...
+	dualPatch := `
+apiVersion: kubeadm.k8s.io/v1beta2
+kind: ClusterConfiguration
+metadata:
+  name: config
+featureGates:
+  IPv6DualStack: true
+`
+	if len(cfg.KubeadmConfigPatches) > 0 {
+		cfg.KubeadmConfigPatches = append(cfg.KubeadmConfigPatches, dualPatch)
```
aojea, line 320: is this ok, or should we use Json6902?

pkg/cluster/internal/create/create.go
```diff
@@ -255,3 +262,64 @@ func validateProvider(p providers.Provider) error {
...
+		return errors.New("failed to create cluster, dual-stack is only supported on Kubernetes versions 1.20+")
+	}
+
+	// dual-stack feature gate is enabled by default since 1.21
+	if ver.AtLeast(version.MustParseSemantic("v1.21.0")) {
+		return nil
+	}
+
+	// for 1.20 we have to explicitly enable the feature gate
+	// we can patch kubeadm here so we avoid to plumb this hack further
```
aojea, line 310: kubeadm adds the corresponding feature gates flags to the other components, controller-manager, apiserver, kube-proxy ... and validates that podsubnet and serviceSubnet are correct ... so we don't need to set the feature flags

aojea:
after updated the kindnet image
```
kubectl set image ds/kindnet kindnet-cni=kindest/kindnetd:v20210326-1e038dc5 -n kube-system
```

and delete the old kindnet pods that should be crashlooping because it can't handle dualstack correctly
```
I0401 13:49:37.797835       1 main.go:185] handling current node
panic: Can't synchronize rules after 3 attempts: running [/usr/sbin/iptables -t nat -C KIND-MASQ-AGENT -d 10.244.0.0/16,fd00:10:244::/56 -j RETURN -m comment --comment kind-masq-agent: local traffic is not subject to MASQUERADE --wait]: exit status 2: iptables v1.8.5 (legacy): invalid mask `56' specified
Try `iptables -h' or 'iptables --help' for more information.
```

I have some problems because the apiserver fails to validate everything after this error
```
E0401 12:44:27.426813       1 controller.go:152] Unable to remove old endpoints from kubernetes service: StorageError: key not found, Code: 1, Key: /registry/masterleases/172.18.0.2, ResourceVersion: 0, AdditionalErrorMsg: 
E0401 12:44:27.435678       1 fieldmanager.go:186] [SHOULD NOT HAPPEN] failed to update managedFields for /, Kind=: failed to convert new object (/v1, Kind=Node) to smd typed: .status.addresses: duplicate entries for key [type="InternalIP"]
```

@liggitt can you give me some hints to debug this further?

liggitt:
> ```
> E0401 12:44:27.435678       1 fieldmanager.go:186] [SHOULD NOT HAPPEN] failed to update managedFields for /, Kind=: failed to convert new object (/v1, Kind=Node) to smd typed: .status.addresses: duplicate entries for key [type="InternalIP"]
> ```
> 
> @liggitt can you give me some hints to debug this further?

sounds like https://github.com/kubernetes/kubernetes/issues/88182

cc @apelisse

liggitt:
> duplicate entries for key [type="InternalIP"]

Looks like a retreat of https://github.com/kubernetes/kubernetes/pull/79391#issue-291747312, but now with server-side apply.

liggitt:
If we can have existing objects with duplicate items here (and it sounds like we can), the managedFields update *cannot* break update of those objects. Complaining on Apply requests is in-bounds (though it sounds like the merge key selected for the field may be wrong in the first place if we expect to have multiple entries of the same address type with different IP families)

apelisse:
Yeah, I'd be very surprised if it does, I think that's a red-herring.

pkg/cluster/internal/create/create.go
BenTheElder: There's a much easier way to default this when we generate the base kind config which also respects the user patch applied on top. See how we handled it for the PV feature gatr

site/content/docs/user/configuration.md
```diff
@@ -102,22 +102,52 @@ Multiple details of the cluster's networking can be customized under the
...
-is not yet supported in docker for mac.
+{{< codeFromInline lang="yaml" >}}
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+networking:
+  ipFamily: ipv6
+{{< /codeFromInline >}}
+
+##### Dual Stack clusters
+You can run dual stack clusters using `kind`, on kubernetes versions 1.20+.
```
BenTheElder, line 143: Please note that this requires 0.11.0+

aojea:
@liggitt @apelisse indeed it was a redherring, thanks for your quick response, I can't explain why by it was caused by a wrong kubeadm configuration, there was also rbac errors on the scheduler and the controller manager.

@BenTheElder PTAL, it needs the new kindnet image in order to work
```
kubectl set image ds/kindnet kindnet-cni=kindest/kindnetd:v20210326-1e038dc5 -n kube-system
```

```sh
$ kubectl apply -f https://gist.githubusercontent.com/aojea/90768935ab71cb31950b6a13078a7e92/raw/99ceac308f2b2658c7313198a39fbe24b155ae68/dual-stack.yaml
$ kubectl get endpointslices -o wide
NAME                            ADDRESSTYPE   PORTS   ENDPOINTS                       AGE
kubernetes                      IPv4          6443    172.18.0.2                      4m24s
my-service-prefer-dual-jswjr    IPv4          80      10.244.0.6,10.244.0.5           45s
my-service-prefer-dual-vvkdt    IPv6          80      fd00:10:244::3,fd00:10:244::2   45s
my-service-require-dual-4hqq2   IPv6          80      fd00:10:244::3,fd00:10:244::2   45s
my-service-require-dual-75hk8   IPv4          80      10.244.0.6,10.244.0.5           45s
my-service-v4-429wd             IPv4          80      10.244.0.6,10.244.0.5           45s
my-service-v6-crtqs             IPv6          80      fd00:10:244::3,fd00:10:244::2   45s
```

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -439,6 +451,17 @@ func Config(data ConfigData) (config string, err error) {
 	// derive any automatic fields if not supplied
 	data.Derive()
 
+	// Kubeadm has its own feature-gate for dual stack
+	// we need to enable it for Kubernetes version 1.20 only
+	// dual-stack is only supported in 1.20+
+	// TODO: remove this when 1.20 is EOL or we no longer support
+	// dual-stack for 1.20 in KIND
+	if ver.LessThan(version.MustParseSemantic("v1.21.0")) &&
```
BenTheElder, line 459: we should really use the version for the PR that enabled it by default so it works with arbitrary commits 🙃 

if you checkout kubernetes master then the commit you can grab the version.

BenTheElder: can be a follow-up

BenTheElder:
@aojea upgrade the kindnetd image / manifest? you have access 🙃

BenTheElder:
/lgtm
/approve
can be handled in another PR, along with refining the versions

pkg/cluster/internal/kubeadm/config.go
aojea: this was the kubeadm one https://github.com/kubernetes/kubernetes/pull/99294

and this the kubernetes one https://github.com/kubernetes/kubernetes/pull/98969

aojea: the must parse semantic doesn't accept versions with letters IIRC

PR #2176, opened 2021-04-02
support cross-compiling node-image

### LINKED ISSUE #166: Support arm64

BenTheElder:
specifically intended to resolve #166 

changes:
- deprecate `--kube-root` in favor of making it a require argument (this is not strictly related, but I am reworking node image build bit by bit and started there)
- switch to pulling images with containerd in the running base, instead of to the host. this breaks offline kubernetes development but the previous approach with docker pull / save / export cannot be used reliably for multi-arch
  - fully qualify images to prepull (needed by ctr)
- fix image architecture in upstream built images, some of which are have the wrong metadata
- move architecture support to warning (not strictly necessary, but when people ask for new architectures to be possible to build we can allow it without explicitly supporting them, initially)
- add `--arch` flag to `kind build node-image`
- update push-build.sh to push a multi-arch image using this cross compile mechanism
- update the default node image to a multi-arch image

### DIFF
--- a/pkg/build/nodeimage/buildcontext.go
+++ b/pkg/build/nodeimage/buildcontext.go
@@ -295,13 +295,17 @@ func (c *buildContext) prePullImages(bits kube.Bits, dir, containerID string) ([
 		image := image // https://golang.org/doc/faq#closures_and_goroutines
 		fns = append(fns, func() error {
 			if !builtImages.Has(image) {
-				err := importer.Pull(image, dockerBuildOsAndArch(c.arch))
-				if err != nil {
-					c.logger.Warnf("Failed to pull %s with error: %v", image, err)
-					runE := exec.RunErrorForError(err)
-					c.logger.Warn(string(runE.Output))
-				}
-				// TODO(bentheelder): generate a friendlier name
+				/*
+					TODO: show errors when we have real errors. See comments in
+					importer implementation
+					err := importer.Pull(image, dockerBuildOsAndArch(c.arch))
+					if err != nil {
+						c.logger.Warnf("Failed to pull %s with error: %v", image, err)
+						runE := exec.RunErrorForError(err)
+						c.logger.Warn(string(runE.Output))
+					}
+				*/
+				_ = importer.Pull(image, dockerBuildOsAndArch(c.arch))
 			}
 			return nil
 		})
--- a/pkg/build/nodeimage/buildcontext.go
+++ b/pkg/build/nodeimage/buildcontext.go
@@ -33,7 +33,21 @@ import (
 	"sigs.k8s.io/kind/pkg/errors"
 	"sigs.k8s.io/kind/pkg/exec"
 	"sigs.k8s.io/kind/pkg/fs"
+	"sigs.k8s.io/kind/pkg/log"
 )
+
+// buildContext is used to build the kind node image, and contains
+// build configuration
+type buildContext struct {
+	// option fields
+	image     string
+	baseImage string
+	logger    log.Logger
+	arch      string
+	kubeRoot  string
+	// non-option fields
+	builder kube.Builder
+}
 
 // Build builds the cluster node image, the sourcedir must be set on
 // the buildContext
@@ -188,7 +202,7 @@ func (c *buildContext) prePullImages(bits kube.Bits, dir, containerID string) ([
 	fixRepository := func(repository string) string {
 		if strings.HasSuffix(repository, archSuffix) {
 			fixed := strings.TrimSuffix(repository, archSuffix)
-			fmt.Println("fixed: " + repository + " -> " + fixed)
+			c.logger.V(1).Info("fixed: " + repository + " -> " + fixed)
 			repository = fixed
 		}
 		return repository
@@ -206,7 +220,7 @@ func (c *buildContext) prePullImages(bits kube.Bits, dir, containerID string) ([
 		fixedImages.Insert(registry + ":" + tag)
 	}
 	builtImages = fixedImages
-	c.logger.V(0).Info("Detected built images: " + strings.Join(builtImages.List(), ", "))
+	c.logger.V(1).Info("Detected built images: " + strings.Join(builtImages.List(), ", "))
 
 	// gets the list of images required by kubeadm
 	requiredImages, err := exec.OutputLines(cmder.Command(
@@ -262,38 +276,6 @@ func (c *buildContext) prePullImages(bits kube.Bits, dir, containerID string) ([
 		return nil, errors.Wrap(err, "failed to make images dir")
 	}
 
-	fns := []func() error{}
-	pulledImages := make(chan string, len(requiredImages))
-	for i, image := range requiredImages {
-		i, image := i, image // https://golang.org/doc/faq#closures_and_goroutines
-		fns = append(fns, func() error {
-			if !builtImages.Has(image) {
-				fmt.Printf("Pulling: %s\n", image)
-				err := docker.Pull(c.logger, image, 2)
-				if err != nil {
-					c.logger.Warnf("Failed to pull %s with error: %v", image, err)
-				}
-				// TODO(bentheelder): generate a friendlier name
-				pullName := fmt.Sprintf("%d.tar", i)
-				pullTo := path.Join(imagesDir, pullName)
-				err = docker.Save(image, pullTo)
-				if err != nil {
-					return err
-				}
-				pulledImages <- pullTo
-			}
-			return nil
-		})
-	}
-	if err := errors.AggregateConcurrent(fns); err != nil {
-		return nil, err
-	}
-	close(pulledImages)
-	pulled := []string{}
-	for image := range pulledImages {
-		pulled = append(pulled, image)
-	}
-
 	// setup image importer
 	importer := newContainerdImporter(cmder)
 	if err := importer.Prepare(); err != nil {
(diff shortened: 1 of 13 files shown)


BenTheElder:
I've been maintaining a test image at `bentheelder/kind-node:arm64-test`. 
I do not have arm hardware so I can only do pretty minimal testing easily.

https://github.com/kubernetes/release/issues/1842 is relevant, we just patch the image config to workaround this which seems to be fine so far.

Switching away from pulling images to the host docker will cause two regressions:
- users building node images and leveraging a mirror / proxy
- users building node images offline (currently possible if you build once and none of the pulled / non-built images change)

However without switching away we are blocked by https://github.com/moby/moby/issues/36552#issuecomment-582051871

`docker save` does not even have a `--platform` option, so we'll need a different approach one way or another.

we could re-introduce the pull-to-host approach when we're confident that we're building for the same platform as the host, but I'm not sure this is worth prioritizing right now, these only affect building node images in specific cases, not consuming built images.

because manifest lists can only exist on a registry and not in local docker, `kind build node-image` is *not* building multiple architectures at once, instead it can be invoked for each architecture and `docker manifest` used to construct a multi-arch combined manifest image in the registry.

I will update our push-node.sh to do this.

BenTheElder:
This is failing in CI because `docker run `--platform` is technically experimental in the daemon still, at least on the version of docker we have in CI. My local daemon has experimental enabled currently. I'm checking if it's still experimental gated in the latest version.

BenTheElder:
1.15/1.16 e2e images have a docker version that is too old for the multi-arch --run without enabling experimental daemon options.

BenTheElder:
We can probably just spin down these old jobs.

One other annoying aspect remaining besides code cleanup is extremely noisy output from the image pulling now. I'll get back to that tomorrow, as well as updating `push-node.sh` to produce a multi-arch image using this.

This is done enough though that you can do `kind build node-image --arch=arm64` on a non-arm64 host and it should still produce an image for arm64, so it's ready for testing. I've pre-built one for v1.20.2 at `bentheelder/kind-node:arm-test`.

pkg/build/nodeimage/buildcontext.go
```diff
@@ -308,19 +290,28 @@ func (c *buildContext) prePullImages(bits kube.Bits, dir, containerID string) ([
...
+				if err != nil {
+					c.logger.Warnf("Failed to pull %s with error: %v", image, err)
+					runE := exec.RunErrorForError(err)
+					c.logger.Warn(string(runE.Output))
+				}
+				// TODO(bentheelder): generate a friendlier name
 			}
-			defer f.Close()
-			return importer.LoadCommand().SetStdout(os.Stdout).SetStderr(os.Stdout).SetStdin(f).Run()
+			return nil
```
aojea, line 306: this always returns nil

pkg/build/nodeimage/buildcontext.go
```diff
@@ -308,19 +290,28 @@ func (c *buildContext) prePullImages(bits kube.Bits, dir, containerID string) ([
...
 		})
 	}
+	if err := errors.AggregateConcurrent(fns); err != nil {
```
aojea, line 309: if never returns an error :shrug:

pkg/build/nodeimage/const_cni.go
```diff
@@ -20,7 +20,7 @@ package nodeimage
 The default CNI manifest and images are our own tiny kindnet
 */
 
-var defaultCNIImages = []string{"kindest/kindnetd:v20210326-1e038dc5"}
+var defaultCNIImages = []string{"docker.io/kindest/kindnetd:v20210326-1e038dc5"}
```
aojea, line 23: should we use a constant?
 in case we need/have to switch to other registry that will make it easier

aojea: ups, it is embedded into the manifests, forget it

pkg/build/nodeimage/internal/container/docker/archive.go
```diff
@@ -117,6 +117,15 @@ func EditArchiveRepositories(reader io.Reader, writer io.Writer, editRepositorie
 				return err
 			}
 			hdr.Size = int64(len(b))
+			// edit image config when we find that
+		} else if strings.HasSuffix(hdr.Name, ".json") {
```
aojea, line 121: this is very obscure, so every json but manifest.json is an image config manifest?

BenTheElder: yes, they will be named `<hash>.json`. we could get the names from the manifest.json, but then we need to do two passes over the tar or depend on ordering or ...

we could do this, but it doesn't seem necessary. images following the spac should only contain `manifest.json` and `<hash>.json` .json files where the latter are image config files.

links to the spec are in the comments above https://github.com/moby/moby/blob/master/image/spec/v1.2.md

pkg/build/nodeimage/internal/container/docker/archive.go
```diff
@@ -87,7 +87,7 @@ func GetArchiveTags(path string) ([]string, error) {
 // https://github.com/moby/moby/blob/master/image/spec/v1.md
 // https://github.com/moby/moby/blob/master/image/spec/v1.1.md
 // https://github.com/moby/moby/blob/master/image/spec/v1.2.md
-func EditArchiveRepositories(reader io.Reader, writer io.Writer, editRepositories func(string) string) error {
+func EditArchive(reader io.Reader, writer io.Writer, editRepositories func(string) string, architectureOverride string) error {
```
BenTheElder, line 90: we have to do this because kubernetes has frequently produced images with the wrong architecture metadata when cross compiling images, which prevents success importing / unning them under containerd which is actually checking for the matching image. dockerd is pretty loose with using these but does throw warnings.

pkg/build/nodeimage/internal/container/docker/archive.go
BenTheElder: a typical image archive only has two .json files in it, but it's possible to create a tarball with multiple images, in which case there is one `manifest.json` and one config hash.json. 

The manifest contains an array of structs each with the config file name for an image, it's tags, and it's layers. There is also the legacy repositories file for the tags.

The config file specifies the non-layer data like architecture/os, entrypoint, env, etc. for one image. We are patching the architecture field to avoid kind only working on highly patched kubernetes builds.

pkg/build/nodeimage/const_cni.go
BenTheElder: this is just the fully qualified value, it's already implicitly `docker.io/` for images that do not specify a host.
except on podman, which is often configured to try a redhat registry first IIRC. but kubernetes etc. will resolve this to `docker.io/kindest/kindnetd`.

pkg/build/nodeimage/buildcontext.go
BenTheElder: this line cannot know that, `AggregateConcurrent` can return it's own error for example.

aojea: That is true

pkg/build/nodeimage/internal/container/docker/archive.go
aojea: That is  a lot of information, do you think future us can figure it out?

BenTheElder: There are comments and there are links to the spec, which is not very lengthy :-)

pkg/build/nodeimage/buildcontext.go
BenTheElder: Yes? Because the contract is a method that returns an error type but we are only running one thing and it will error and that's fine. We check later that the images are loaded.

Ctr pull does not have the --no-unpack option unlike the option I added to import.

pkg/build/nodeimage/internal/container/docker/archive.go
aojea: :+1:

BenTheElder:
I will finish this up when I'm back Monday (at minimum: reduce excessive terminal / log spam from image loading, rebase).

In the meantime I think we should hold off v0.11 until this is good and landed @amwat.

egandro:
Following here. Just notice me here - I have a  Pi400 and a Pi4/8gb for testing.

egandro:
Any progress here or help required?

BenTheElder:
rebased and "fixed" the excessive noise issue (silenced output from images pulled to node by importer).

should still add multi-arch to push-node.sh

BenTheElder:
this is ~good enough:
```console
$ DOCKER_CLI_EXPERIMENTAL=enabled docker manifest inspect kindest/node:v1.21.1
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 2821,
         "digest": "sha256:21a1f4a2ff8e62abab5286db9a6d13734b62b702a61247a2500ea3289d2276dd",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 2821,
         "digest": "sha256:0e501b88eda8c7618ae9308c07f9d14dd6bb7835dfb83836a3ef0cfec50ba190",
         "platform": {
            "architecture": "arm64",
            "os": "linux"
         }
      }
   ]
}
```

BenTheElder:
note: 1.15 and 1.16 will fail because the docker installed in those environments is simply too old. we will need to retire these jobs, Kubernetes just did the last 1.18 patch release ...

spiffxp:
I'm guessing this means kind will no longer support 1.15 and 1.16 in its next release? Do we call out supported versions somewhere?

aojea:
> I'm guessing this means kind will no longer support 1.15 and 1.16 in its next release? Do we call out supported versions somewhere?

There at least 2 places in the docs, "principles" and "roadmap" that talk about 

> Support all currently upstream-supported Kubernetes versions

BenTheElder:
It doesn't mean 1.15/1.16 are totally unsupported it just means we won't be able to run e2e. They will still receive best effort support.

PR #2179, opened 2021-04-02
avoid head sigpipe

### LINKED ISSUE #2178: kindest/node entrypoint sometimes fails with exit code 141 (SIGPIPE)
**What happened**:

Some context: 
* we're using kind for e2e tests in cluster-api
* tests are using CAPD (ClusterAPI Provider Docker) to spin up other Docker-based clusters 
* tests are executed in Prow
* tests are running in parallel
* we're using the kindest/node image to spin up our own Docker-based clusters 

Most of it shouldn't matter, but it might be helpful context. From time to time we're seeing that the kindest/node image fails on startup. According to `docker inspect` it fails with exit code `141` ([logs](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api/4416/pull-cluster-api-e2e-full-main/1377665252484714496/artifacts/clusters/bootstrap/controllers/capd-controller-manager/capd-controller-manager-845bbfdf7d-fmbcx/manager.log)). `docker logs` shows the following:
* regular log:
```bash
"INFO: ensuring we can execute mount/umount even with userns-remap"
"INFO: remounting /sys read-only"
"INFO: making mounts shared"
"INFO: fix cgroup mounts for all subsystems"
```
* log with set -x:
```bash
+ local docker_cgroup cgroup_subsystems subsystem
++ head -n 1
++ cut '-d ' -f 4
++ echo '6217 6216 0:29 /docker/61633ea3cf604679d3cc4292524d1d37dfbcdcce1a35b0ae6f487bd32a86e556 /sys/fs/cgroup/systemd rw,nosuid,nodev,noexec,relatime shared:923 - cgroup cgroup rw,xattr,name=systemd
6287 6216 0:32 /docker/61633ea3cf604679d3cc4292524d1d37dfbcdcce1a35b0ae6f487bd32a86e556 /sys/fs/cgroup/cpu,cpuacct rw,nosuid,nodev,noexec,relatime shared:924 - cgroup cgroup rw,cpu,cpuacct
6288 6216 0:33 /docker/61633ea3cf604679d3cc4292524d1d37dfbcdcce1a35b0ae6f487bd32a86e556 /sys/fs/cgroup/net_cls,net_prio rw,nosuid,nodev,noexec,relatime shared:925 - cgroup cgroup rw,net_cls,net_prio
6289 6216 0:34 /docker/61633ea3cf604679d3cc4292524d1d37dfbcdcce1a35b0ae6f487bd32a86e556 /sys/fs/cgroup/memory rw,nosuid,nodev,noexec,relatime shared:926 - cgroup cgroup rw,memory
6290 6216 0:35 /docker/61633ea3cf604679d3cc4292524d1d37dfbcdcce1a35b0ae6f487bd32a86e556 /sys/fs/cgroup/freezer rw,nosuid,nodev,noexec,relatime shared:927 - cgroup cgroup rw,freezer
6291 6216 0:36 /docker/61633ea3cf604679d3cc4292524d1d37dfbcdcce1a35b0ae6f487bd32a86e556 /sys/fs/cgroup/hugetlb rw,nosuid,nodev,noexec,relatime shared:928 - cgroup cgroup rw,hugetlb
6320 6216 0:37 /docker/61633ea3cf604679d3cc4292524d1d37dfbcdcce1a35b0ae6f487bd32a86e556 /sys/fs/cgroup/pids rw,nosuid,nodev,noexec,relatime shared:929 - cgroup cgroup rw,pids
6322 6216 0:39 /docker/61633ea3cf604679d3cc4292524d1d37dfbcdcce1a35b0ae6f487bd32a86e556 /sys/fs/cgroup/blkio rw,nosuid,nodev,noexec,relatime shared:931 - cgroup cgroup rw,blkio
6323 6216 0:40 /docker/61633ea3cf604679d3cc4292524d1d37dfbcdcce1a35b0ae6f487bd32a86e556 /sys/fs/cgroup/cpuset rw,nosuid,nodev,noexec,relatime shared:932 - cgroup cgroup rw,cpuset
6324 6216 0:41 /docker/61633ea3cf604679d3cc4292524d1d37dfbcdcce1a35b0ae6f487bd32a86e556 /sys/fs/cgroup/perf_event rw,nosuid,nodev,noexec,relatime shared:933 - cgroup cgroup rw,perf_event
6325 6216 0:42 /docker/61633ea3cf604679d3cc4292524d1d37dfbcdcce1a35b0ae6f487bd32a86e556 /sys/fs/cgroup/devices rw,nosuid,nodev,noexec,relatime shared:934 - cgroup cgroup rw,devices'
+ docker_cgroup=/docker/61633ea3cf604679d3cc4292524d1d37dfbcdcce1a35b0ae6f487bd32a86e556
```

To me it looks like it fails here: https://github.com/kubernetes-sigs/kind/blob/v0.9.0/images/base/files/usr/local/bin/entrypoint#L82

The only reasonable explanation I could come up with is:
* `echo` sends the `docker_cgroup_mounts` into the pipe (maybe line by line?)
* `head -n 1` exits after the first line
* the pipe (and thus the whole script) fails because `echo` cannot send the other lines

I suspect it happens rarely because this might be caused by a CPU context switch (or something like this?) at the wrong moment, because of the high concurrency we have during our tests.

kindest/node should not fail on startup.

Really hard to reproduce. In our case we can be sure it's gone after about 10-20 successful test runs in a row.

I changed the following and this seems to fix the issue in my case:
* old code:
```bash
docker_cgroup=$(echo "${docker_cgroup_mounts}" | head -n 1 | cut -d' ' -f 4)
```
* "fixed" code:
```bash
docker_cgroup=$( < <(echo "${docker_cgroup_mounts}") head -n 1 | cut -d' ' -f 4)
```
The idea is to move the echo out of the pipe so it should not be possible anymore to get a SIGPIPE/ pipefail.

The code in kind on the main branch changed a bit since the version 0.9 we're currently using, but there's one line left which should have the same issue:
https://github.com/kubernetes-sigs/kind/blob/main/images/base/files/usr/local/bin/entrypoint#L232

For reference, the respective cluster-api issue: https://github.com/kubernetes-sigs/cluster-api/issues/4405

**Environment:**

- kind version: (use `kind version`): 0.9
- Kubernetes version: (use `kubectl version`): 1.19.1
```bash
Client: Docker Engine - Community
 Version: 20.10.3
 API version: 1.40
 Go version: go1.13.15
 Git commit: 48d30b5
 Built: Fri Jan 29 14:33:25 2021
 OS / Arch: linux / amd64
 Context: default
 Experimental: true

 Engine:
  Version: 19.03.15
  API version: 1.40 (minimum version 1.12)
  Go version: go1.13.15
  Git commit: 99e3ed8919
  Built: Sat Jan 30 03:15:34 2021
  OS / Arch: linux / amd64
  Experimental: false
 containerd:
  Version: 1.4.3
  GitCommit: 269548fa27e0089a8b8278fc4fc781d7f65a939b
 runc:
  Version: 1.0.0-rc92
  GitCommit: ff819c7e9184c13b7c2607fe6c30ae19403a7aff
 docker-init:
  Version: 0.18.0
  GitCommit: fec3683
```
- OS (e.g. from `/etc/os-release`): not sure but the image we're using in the ProwJob is: `gcr.io/k8s-testimages/kubekins-e2e:v20210330-fadf59c-go-canary`

BenTheElder:
fixes #2178

### DIFF
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -229,7 +229,7 @@ fix_cgroup() {
   cgroup_mounts=$(grep -E -o '/[[:alnum:]].* /sys/fs/cgroup.*.*cgroup' /proc/self/mountinfo || true)
   if [[ -n "${cgroup_mounts}" ]]; then
     local mount_root
-    mount_root=$(echo "${cgroup_mounts}" | head -n 1 | cut -d' ' -f1)
+    mount_root=$(head -n 1 <<<"${cgroup_mounts}" | cut -d' ' -f1)
     for mount_point in $(echo "${cgroup_mounts}" | cut -d' ' -f 2); do
       # bind mount each mount_point to mount_point + mount_root
       # mount --bind /sys/fs/cgroup/cpu /sys/fs/cgroup/cpu/docker/fb07bb6daf7730a3cb14fc7ff3e345d1e47423756ce54409e66e01911bab2160
@@ -336,7 +336,7 @@ enable_network_magic(){
 
   # first we need to detect an IP to use for reaching the docker host
   local docker_host_ip
-  docker_host_ip="$( (getent ahostsv4 'host.docker.internal' | head -n1 | cut -d' ' -f1) || true)"
+  docker_host_ip="$( (head -n1 <(getent ahostsv4 'host.docker.internal') | cut -d' ' -f1) || true)"
   # if the ip doesn't exist or is a loopback address use the default gateway
   if [[ -z "${docker_host_ip}" ]] || [[ $docker_host_ip =~ ^127\.[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
     docker_host_ip=$(ip -4 route show default | cut -d' ' -f3)
@@ -358,7 +358,7 @@ enable_network_magic(){
   sed -e "s/${docker_embedded_dns_ip}/${docker_host_ip}/g" /etc/resolv.conf.original >/etc/resolv.conf
 
   # fixup IPs in manifests ...
-  curr_ipv4="$( (getent ahostsv4 "$(hostname)" | head -n1 | cut -d' ' -f1) || true)"
+  curr_ipv4="$( (head -n1 <(getent ahostsv4 "$(hostname)") | cut -d' ' -f1) || true)"
   echo "INFO: Detected IPv4 address: ${curr_ipv4}" >&2
   if [ -f /kind/old-ipv4 ]; then
       old_ipv4=$(cat /kind/old-ipv4)
@@ -378,7 +378,7 @@ enable_network_magic(){
   fi
 
   # do IPv6
-  curr_ipv6="$( (getent ahostsv6 "$(hostname)" | head -n1 | cut -d' ' -f1) || true)"
+  curr_ipv6="$( (head -n1 <(getent ahostsv6 "$(hostname)") | cut -d' ' -f1) || true)"
   echo "INFO: Detected IPv6 address: ${curr_ipv6}" >&2
   if [ -f /kind/old-ipv6 ]; then
       old_ipv6=$(cat /kind/old-ipv6)
--- a/pkg/build/nodeimage/defaults.go
+++ b/pkg/build/nodeimage/defaults.go
@@ -20,4 +20,4 @@ package nodeimage
 const DefaultImage = "kindest/node:latest"
 
 // DefaultBaseImage is the default base image used
-const DefaultBaseImage = "kindest/base:v20210328-c17ca167@sha256:0311870f4d35b0f68e2fedb5d703a552a8e7eb438acc67a3bd13982c2bda7487"
+const DefaultBaseImage = "docker.io/kindest/base:v20210402-3d9112b0"


pkg/build/nodeimage/defaults.go
```diff
@@ -20,4 +20,4 @@ package nodeimage
 const DefaultImage = "kindest/node:latest"
 
 // DefaultBaseImage is the default base image used
-const DefaultBaseImage = "kindest/base:v20210328-c17ca167@sha256:0311870f4d35b0f68e2fedb5d703a552a8e7eb438acc67a3bd13982c2bda7487"
+const DefaultBaseImage = "docker.io/kindest/base:v20210402-69a81dd6"
```
BenTheElder, line 23: note: we need to drop the digest here for #2176 anyhow.
this is fine-ish, I guess, as these are unique tags at least.

images/base/files/usr/local/bin/entrypoint
```diff
@@ -229,7 +229,7 @@ fix_cgroup() {
   cgroup_mounts=$(grep -E -o '/[[:alnum:]].* /sys/fs/cgroup.*.*cgroup' /proc/self/mountinfo || true)
   if [[ -n "${cgroup_mounts}" ]]; then
     local mount_root
-    mount_root=$(echo "${cgroup_mounts}" | head -n 1 | cut -d' ' -f1)
+    mount_root=$(head -n 1 <<<${cgroup_mounts}" | cut -d' ' -f1)
```
amwat, line 232:
```suggestion
    mount_root=$(head -n 1 <<<"${cgroup_mounts}" | cut -d' ' -f1)
```

BenTheElder: thanks! bad refactoring

amwat (commented):
/lgtm

PR #2185, opened 2021-04-06
docs: added contributor over time

juzhiyuan:
Hi, this PR aims to add Contributor Over Time for Kind since 2018-07-24.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,10 @@ start, or feel free to reach out to discuss. 🙂
 
 See also: our own [contributor guide] and the Kubernetes [community page].
 
+### Contributor Over Time
+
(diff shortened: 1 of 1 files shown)


BenTheElder:
What is the Purpose of this? Won't it go out of date? Why wouldn't people just check GitHub?

juzhiyuan:
ok, this graph shows that the trend for kind repo :) As for one link with dynamically generated data, this feature is still in development, once it's done, I would add it back. Does that make sense?

BenTheElder:
While this is cool to know we already have a lot of information for people to consume here. I don't think this is the right page for it and I'd rather not include external content we don't control, the Kubernetes documentation guidelines are against this.

juzhiyuan:
Got it, thanks!

PR #2190, opened 2021-04-09
add monitoring to kind

aojea:
Install a prometheus deployment that collect metrics from all
the kubernetes components.
Modify kube-proxy to listen for metrics in all addresses, it listen
by default only in localhost.
Once the cluster is destroyed, it dumps the prometheus snapshot
as a tarball, so it can be analyzed locally.

https://suraj.io/post/how-to-backup-and-restore-prometheus/

### DIFF
--- a/hack/ci/e2e-k8s.sh
+++ b/hack/ci/e2e-k8s.sh
@@ -35,6 +35,13 @@ cleanup() {
   # KIND_CREATE_ATTEMPTED is true once we: kind create
   if [ "${KIND_CREATE_ATTEMPTED:-}" = true ]; then
     kind "export" logs "${ARTIFACTS}/logs" || true
+    # get prometheus URL from the Service (servicePort = 8080)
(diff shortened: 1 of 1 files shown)


aojea:
/hold
I want to investigate more the CI performance from the cluster POV

aojea:
niiiiiiiiiiiiiiiiiiiiiiiice

image

BenTheElder:
I think we should discuss having a mode where e2e.test in k/k is responsible for installing monitoring and snapshotting it 🤔

aojea:
> I think we should discuss having a mode where e2e.test in k/k is responsible for installing monitoring and snapshotting it

let's fork this to k/k then, I really find it useful in the openshift CI, they do this metrics snapshoting 

> I think we should discuss having a mode where e2e.test in k/k is responsible for installing monitoring and snapshotting it

yeah. let's discuss it there
https://github.com/kubernetes/kubernetes/issues/103145

PR #2199, opened 2021-04-14
fix getSubnets for podman

flobz:
The command to getSubnets for  podman isn't the same as the one for docker.

### DIFF
--- a/pkg/cluster/internal/providers/podman/provision.go
+++ b/pkg/cluster/internal/providers/podman/provision.go
@@ -281,7 +281,7 @@ func getProxyEnv(cfg *config.Cluster) (map[string]string, error) {
 }
 
 func getSubnets(networkName string) ([]string, error) {
-	format := '{{ range (index (index (index (index . "plugins") 0 ) "ipam" ) "ranges")}}{{ index ( index . 0 ) "subnet" }} {{end}}'
+	format := `{{ range (index (index (index (index . "plugins") 0 ) "ipam" ) "ranges")}}{{ index ( index . 0 ) "subnet" }} {{end}}`
 	cmd := exec.Command("podman", "network", "inspect", "-f", format, networkName)
 	lines, err := exec.OutputLines(cmd)
 	if err != nil {
--- a/pkg/cluster/internal/providers/podman/provision.go
+++ b/pkg/cluster/internal/providers/podman/provision.go
@@ -281,7 +281,7 @@ func getProxyEnv(cfg *config.Cluster) (map[string]string, error) {
 }
 
 func getSubnets(networkName string) ([]string, error) {
-	format := `{{index (index  .plugins  0).ipam.ranges 0 0 "subnet"}}`
+	format := '{{ range (index (index (index (index . "plugins") 0 ) "ipam" ) "ranges")}}{{ index ( index . 0 ) "subnet" }} {{end}}'
 	cmd := exec.Command("podman", "network", "inspect", "-f", format, networkName)
 	lines, err := exec.OutputLines(cmd)
 	if err != nil {
--- a/pkg/cluster/internal/providers/podman/provision.go
+++ b/pkg/cluster/internal/providers/podman/provision.go
@@ -281,7 +281,7 @@ func getProxyEnv(cfg *config.Cluster) (map[string]string, error) {
 }
 
 func getSubnets(networkName string) ([]string, error) {
-	format := `{{range (index (index . "IPAM") "Config")}}{{index . "Subnet"}} {{end}}`
+	format := `{{index (index  .plugins  0).ipam.ranges 0 0 "subnet"}}`
 	cmd := exec.Command("podman", "network", "inspect", "-f", format, networkName)
 	lines, err := exec.OutputLines(cmd)
 	if err != nil {


pkg/cluster/internal/providers/podman/provision.go
```diff
@@ -281,7 +281,7 @@ func getProxyEnv(cfg *config.Cluster) (map[string]string, error) {
 }
 
 func getSubnets(networkName string) ([]string, error) {
-	format := `{{range (index (index . "IPAM") "Config")}}{{index . "Subnet"}} {{end}}`
+	format := `{{index (index  .plugins  0).ipam.ranges 0 0 "subnet"}}`
```
aojea, line 284: has this changed again in podman?
it seems that the new command works for me in 3.0.0 but not the older one

aojea: we need to provide all the networks, this returns only the first subnet
```
sudo podman network inspect -f '{{index (index  .plugins 0 ).ipam.ranges 0 0 "subnet"}}' kind
fc00:8866:27d0:bd7e::/64
```

```
sudo podman network inspect -f json  kind
[
    {
        "cniVersion": "0.4.0",
        "name": "kind",
        "plugins": [
            {
                "bridge": "cni-podman4",
                "hairpinMode": true,
                "ipMasq": true,
                "ipam": {
                    "ranges": [
                        [
                            {
                                "subnet": "fc00:8866:27d0:bd7e::/64"
                            }
                        ],
                        [
                            {
                                "subnet": "10.89.2.0/24"
                            }
                        ]

```

flobz: Yes but I don't know how, I'm not good at templates.

aojea:
```
sudo podman network inspect kind -f '{{ range (index (index (index (index . "plugins") 0 ) "ipam" ) "ranges")}}{{ index ( index . 0 ) "subnet" }} {{end}} '
fc00:8866:27d0:bd7e::/64 10.89.0.0/24
```

aojea: @flobz can you update the PR with this template ^^^?

pkg/cluster/internal/providers/podman/provision.go
```diff
@@ -281,7 +281,7 @@ func getProxyEnv(cfg *config.Cluster) (map[string]string, error) {
 }
 
 func getSubnets(networkName string) ([]string, error) {
-	format := `{{index (index  .plugins  0).ipam.ranges 0 0 "subnet"}}`
+	format := '{{ range (index (index (index (index . "plugins") 0 ) "ipam" ) "ranges")}}{{ index ( index . 0 ) "subnet" }} {{end}}'
```
aojea: Keep the backticks, it doesn't work well without them

aojea:
@flobz you need to sign the CLA
/ok-to-test
/lgtm
/assign @amwat

flobz:
> @flobz you need to sign the CLA
Done

pkg/cluster/internal/providers/podman/provision.go
```diff
@@ -281,7 +281,7 @@ func getProxyEnv(cfg *config.Cluster) (map[string]string, error) {
 }
 
 func getSubnets(networkName string) ([]string, error) {
-	format := `{{range (index (index . "IPAM") "Config")}}{{index . "Subnet"}} {{end}}`
+	format := `{{ range (index (index (index (index . "plugins") 0 ) "ipam" ) "ranges")}}{{ index ( index . 0 ) "subnet" }} {{end}}`
```
amwat, line 284: This format is getting too complicated now. might be worth just unmarshalling the JSON ourselves.
but fine as a follow up.

Also, is this after podman 3.0 ?
do we want to warn/error if podman version is less than that.

@aojea how is our Github Actions working, is this path not exercised?

amwat (commented):
I'm assuming the format is working (don't currently have a setup to test)
So the changes lgtm.

But we might also need to fix the entire code path here as a follow up
even the default network name seems to have changed from "bridge" to "podman"?

pkg/cluster/internal/providers/podman/provision.go
aojea: > This format is getting too complicated now. might be worth just unmarshalling the JSON ourselves.
but fine as a follow up.

amen, it took me a big chunk of time to come up with the template

> Also, is this after podman 3.0 ?
> do we want to warn/error if podman version is less than that.

I can't know if this worked before because this is only used to populate the proxy environment variables ... let's keep it this way, I can't imagine we have such power users of podman until now

> @aojea how is our Github Actions working, is this path not exercised?

nope

BenTheElder: +1 to json in a follow up

aojea:
@flobz your CLA is not signed or the commit doesn't have the right info, can you please check?

aojea:
/close
@flobz I've added you as coauthor, I rushed it to get into the release
https://github.com/kubernetes-sigs/kind/pull/2257
I didn't get your email so it doesn't reflect it correct in github, I assume is the same issue that your cla error

Issue #2186, opened 2021-04-07
Fully specify image sources


garyedwards:
**What happened**:

```sh
$ kind create cluster
enabling experimental podman provider
Creating cluster "kind" ...
 ✗ Ensuring node image (kindest/node:v1.20.2) 🖼
ERROR: failed to create cluster: failed to pull image "kindest/node@sha256:8f7ea6e7642c0da54f04a7ee10431549c0257315b3a634f6ef2fecaaedb19bab": command "podman pull kindest/node@sha256:8f7ea6e7642c0da54f04a7ee10431549c0257315b3a634f6ef2fecaaedb19bab" failed with error: exit status 125
Command Output: Error: error getting default registries to try: short-name resolution enforced but cannot prompt without a TTY
```

**What you expected to happen**:

```sh
kind create cluster
enabling experimental podman provider
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.20.2) 🖼
 ✓ Preparing nodes 📦
 ✓ Writing configuration 📜
 ✓ Starting control-plane 🕹️
 ✓ Installing CNI 🔌
 ✓ Installing StorageClass 💾
Set kubectl context to "kind-kind"
You can now use your cluster with:

kubectl cluster-info --context kind-kind

Have a nice day! 👋
```

**How to reproduce it (as minimally and precisely as possible)**:

Use a distro with >=Podman 3.0 with `short-name-mode="enforcing"` enabled in `/etc/containers/registries.conf`

Install `kind` as per the documentation.  Try to create a cluster.

**Anything else we need to know?**:

[Container short name mode](https://www.redhat.com/sysadmin/container-image-short-names) for Podman has been set to `enforcing` by default for Fedora 34 and newer.  This brakes the current scripts.

To fix this you can set the setting to `short-name-mode="permissive"`.  This should be added to the [known issues page](https://kind.sigs.k8s.io/docs/user/known-issues).

As a more robust fix, `kind` should fully specify image sources in its scripts.

**Environment:**

- kind version: (use `kind version`): `kind v0.10.0 go1.15.7 linux/amd64`
- Kubernetes version: (use `kubectl version`): N/A
- Docker version: (use `docker info`): Podman `Version: 3.1.0`
- OS (e.g. from `/etc/os-release`): Fedora IoT 34 Beta

BenTheElder:
@aojea podman broke compat with docker again 🤷‍♂️

aojea:
[...]

PR #2220, opened 2021-04-28
CI: update Fedora to 34

AkihiroSuda:
(Marked as a draft until https://github.com/kubernetes-sigs/kind/pull/2230 `make podman resolve short names ` gets merged)

PR #2230, opened 2021-05-06
make podman resolve short names

tao12345666333:
fix: #2186

xref: #2220 

Signed-off-by: Jintao Zhang <zhangjintao9020@gmail.com>

### DIFF
--- a/pkg/cluster/internal/providers/podman/images.go
+++ b/pkg/cluster/internal/providers/podman/images.go
@@ -83,10 +83,33 @@ func pull(logger log.Logger, image string, retries int) error {
 
 // sanitizeImage is a helper to return human readable image name and
 // the podman pullable image name from the provided image
-func sanitizeImage(image string) (string, string) {
+func sanitizeImage(image string) (friendlyImageName, pullImageName string) {
+	const (
+		defaultDomain    = "docker.io/"
+		officialRepoName = "library"
+	)
+
+	var remainder string
+
 	if strings.Contains(image, "@sha256:") {
 		splits := strings.Split(image, "@sha256:")
-		return splits[0], strings.Split(splits[0], ":")[0] + "@sha256:" + splits[1]
+		friendlyImageName = splits[0]
+		remainder = strings.Split(splits[0], ":")[0] + "@sha256:" + splits[1]
+	} else {
+		friendlyImageName = image
+		remainder = image
 	}
-	return image, image
+
+	if !strings.ContainsRune(remainder, '/') {
+		remainder = officialRepoName + "/" + remainder
+	}
+
+	i := strings.IndexRune(friendlyImageName, '/')
+	if i == -1 || (!strings.ContainsAny(friendlyImageName[:i], ".:") && friendlyImageName[:i] != "localhost") {
+		pullImageName = defaultDomain + remainder
+	} else {
+		pullImageName = remainder
+	}
+
+	return
 }
--- a/pkg/cluster/internal/providers/podman/images_test.go
+++ b/pkg/cluster/internal/providers/podman/images_test.go
@@ -1,0 +1,105 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package podman
+
+import (
+	"testing"
+)
+
+func Test_sanitizeImage(t *testing.T) {
+
+	cases := []struct {
+		image             string
+		friendlyImageName string
+		pullImageName     string
+	}{
+		{
+			image:             "kindest/node:v1.21.1@sha256:69860bda5563ac81e3c0057d654b5253219618a22ec3a346306239bba8cfa1a6",
+			friendlyImageName: "kindest/node:v1.21.1",
+			pullImageName:     "docker.io/kindest/node@sha256:69860bda5563ac81e3c0057d654b5253219618a22ec3a346306239bba8cfa1a6",
+		},
+		{
+			image:             "kindest/node:v1.21.1",
+			friendlyImageName: "kindest/node:v1.21.1",
+			pullImageName:     "docker.io/kindest/node:v1.21.1",
+		},
+		{
+			image:             "kindest/node@sha256:69860bda5563ac81e3c0057d654b5253219618a22ec3a346306239bba8cfa1a6",
+			friendlyImageName: "kindest/node",
+			pullImageName:     "docker.io/kindest/node@sha256:69860bda5563ac81e3c0057d654b5253219618a22ec3a346306239bba8cfa1a6",
+		},
+		{
+			image:             "foo.bar/kindest/node:v1.21.1@sha256:69860bda5563ac81e3c0057d654b5253219618a22ec3a346306239bba8cfa1a6",
+			friendlyImageName: "foo.bar/kindest/node:v1.21.1",
+			pullImageName:     "foo.bar/kindest/node@sha256:69860bda5563ac81e3c0057d654b5253219618a22ec3a346306239bba8cfa1a6",
+		},
+		{
+			image:             "foo.bar/kindest/node:v1.21.1",
+			friendlyImageName: "foo.bar/kindest/node:v1.21.1",
+			pullImageName:     "foo.bar/kindest/node:v1.21.1",
+		},
+		{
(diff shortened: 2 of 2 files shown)


AkihiroSuda (approved):

go.mod
```diff
@@ -5,8 +5,10 @@ go 1.14
 require (
 	github.com/BurntSushi/toml v0.3.1
 	github.com/alessio/shellescape v1.4.1
+	github.com/docker/distribution v2.7.1+incompatible
```
aojea, line 8: Ben is not fan of importing new things , can we just append docker.io if is a shortname?

BenTheElder: We actually already have a robust image name parser + normalizer for extracting the tag in our own sources that could easy handle this without an additional dependency 😅
It's slightly more complex than just add docker.io if missing but not much more.

tao12345666333: I can modify it. :smile_cat: 

The introduction of external dependencies is because I also want to discuss whether we should use external dependencies or continue to use our own implementation.

BenTheElder: our other dependencies are pretty self contained, mostly things like "is-atty" which has no dependencies itself and is somewhat pointless to reimplement.
this one has a comparatively massive set in comparison https://github.com/distribution/distribution/blob/main/go.mod

we have to take care to keep kind easy to embed in other projects.

tao12345666333: ok， I will submit the fix today

aojea:
checking in @tao12345666333 , something I can help?

tao12345666333:
Sorry for the delay, I forgot it. :sweat_smile:  I will modify it ASAP!

pkg/cluster/internal/providers/podman/images.go
```diff
@@ -83,10 +83,25 @@ func pull(logger log.Logger, image string, retries int) error {
 
 // sanitizeImage is a helper to return human readable image name and
 // the podman pullable image name from the provided image
-func sanitizeImage(image string) (string, string) {
+func sanitizeImage(image string) (friendlyImageName, pullImageName string) {
```
tao12345666333, line 86: Only dealt with our actual situation, and users may upload the node image to their own private registry.

Did not judge scenarios such as users using `docker.io/library/xxx:vvv`

BenTheElder: we should probably leave a TODO then, because in the case of someone using library someday (??) this will break

BenTheElder: also what about a custom registry? `foo.bar/baz:quux` ?

pkg/cluster/internal/providers/podman/images.go
```diff
@@ -83,10 +83,25 @@ func pull(logger log.Logger, image string, retries int) error {
...
 		splits := strings.Split(image, "@sha256:")
-		return splits[0], strings.Split(splits[0], ":")[0] + "@sha256:" + splits[1]
+		friendlyImageName = splits[0]
+		remainder = strings.Split(splits[0], ":")[0] + "@sha256:" + splits[1]
+	} else {
+		friendlyImageName = image
+		remainder = image
+	}
+
+	s := strings.Count(remainder, "/")
```
BenTheElder, line 99: I think instead we want to look for a period before the tag.
Even then that may not be safe actually, due to custom hostnames. Hmm.

tao12345666333: Addressed.

pkg/cluster/internal/providers/podman/images.go
tao12345666333: Addressed, please see test cases.

pkg/cluster/internal/providers/podman/images.go
```diff
@@ -83,10 +83,33 @@ func pull(logger log.Logger, image string, retries int) error {
 
 // sanitizeImage is a helper to return human readable image name and
 // the podman pullable image name from the provided image
-func sanitizeImage(image string) (string, string) {
+func sanitizeImage(image string) (friendlyImageName, pullImageName string) {
+	const (
+		defaulDomain     = "docker.io/"
```
aojea, line 88:
```suggestion
		defautlDomain     = "docker.io/"
```

tao12345666333: Thanks!  typo

aojea: super nit, forget about it

aojea:
maybe we can add an additional commit to update the CI to fedora 34 so we test this PR against it?
```
  config.vm.box = "fedora/34-cloud-base"
```

pkg/cluster/internal/providers/podman/images.go
```diff
@@ -83,10 +83,33 @@ func pull(logger log.Logger, image string, retries int) error {
...
+		remainder = image
+	}
+
+	if !strings.ContainsRune(remainder, '/') {
+		remainder = officialRepoName + "/" + remainder
 	}
-	return image, image
+
+	i := strings.IndexRune(friendlyImageName, '/')
+	if i == -1 || (!strings.ContainsAny(friendlyImageName[:i], ".:") && friendlyImageName[:i] != "localhost") {
```
aojea, line 108: what is the localhost scenario?
do we need a testcase for it?

tao12345666333: just for cover user using `localhost` as registry domain case.

BenTheElder: I'm not sure this is sufficient to be safe, how does docker actually behave here?

Also: sanitizing at pull-time alone is insufficient, the node container creation needs to reference the resolved name as well.

tao12345666333: both docker and containerd using the same logic.

https://github.com/distribution/distribution/blob/d80a63f1eaa215246eee566f4426d2f2ceecc06e/reference/normalize.go#L87-L104

https://github.com/containerd/containerd/blob/f0d2f89437af17c549f5757e148bc74f8712ceea/reference/docker/reference.go#L664-L681

tao12345666333: > Also: sanitizing at pull-time alone is insufficient, the node container creation needs to reference the resolved name as well.

@BenTheElder  I think it is enough. Podman will do this automatically.


```
➜  podman pull docker.io/library/redis:alpine       
Trying to pull docker.io/library/redis:alpine...                      
Getting image source signatures                                                            
Copying blob 540db60ca938 done                                                             
Copying blob 3f34a000c6b3 done                                                             
Copying blob 8173c12df40f done                                                             
Copying blob a77b7ddf4978 done  
Copying blob 275dfaedaf41 done  
Copying blob 29712d301e8c done                                                             
Copying config efb4fa30f1 done                                                             
Writing manifest to image destination                                                                                                                                                 
Storing signatures                                                                         
efb4fa30f1cfbbf7a637818d1e8de048e3c3d67421c0e2fdae4ca951c3944fed
➜  ~ podman image ls 
REPOSITORY               TAG     IMAGE ID      CREATED      SIZE
docker.io/library/redis  alpine  efb4fa30f1cf  3 weeks ago  33.5 MB
➜  ~ podman run --rm -d redis:alpine
d421acf004ec593b53c2b96396aed2937f3d909dd09f80894f23c826ef1400d1
➜  ~ podman ps                                                                       
CONTAINER ID  IMAGE                           COMMAND       CREATED         STATUS             PORTS   NAMES
d421acf004ec  docker.io/library/redis:alpine  redis-server  12 seconds ago  Up 13 seconds ago          sharp_cartwright

➜  ~ podman image pull docker.io/taobeier/echo:go 
Trying to pull docker.io/taobeier/echo:go...
Getting image source signatures
Copying blob df20fa9351a1 done  
Copying blob 1c6ed02afb58 done  
Copying config 7119caa4d1 done  
Writing manifest to image destination
Storing signatures
7119caa4d13fe93db35a6c4403ff3b474c3ebfc84a1661d09adcf9a8df13817c
➜  ~ podman image ls
REPOSITORY               TAG     IMAGE ID      CREATED       SIZE
docker.io/library/redis  alpine  efb4fa30f1cf  3 weeks ago   33.5 MB
docker.io/taobeier/echo  go      7119caa4d13f  9 months ago  8.84 MB
➜  ~ podman run --rm -d taobeier/echo:go
615b16d1df049f5e63360f6010e3c6e89aeef7b0355698cd7e28c1f3a25690a9
➜  ~ podman ps
CONTAINER ID  IMAGE                           COMMAND       CREATED             STATUS                 PORTS   NAMES
d421acf004ec  docker.io/library/redis:alpine  redis-server  About a minute ago  Up About a minute ago          sharp_cartwright
615b16d1df04  docker.io/taobeier/echo:go                    7 seconds ago       Up 7 seconds ago               boring_napier
➜  ~ podman version
Version:      3.1.2
API Version:  3.1.2
Go Version:   go1.16.3
Built:        Thu May 13 03:27:59 2021
OS/Arch:      linux/amd64

```

tao12345666333: I used the current code and updated the fedora 34 version to trigger CI https://github.com/kubernetes-sigs/kind/pull/2285, which is the original reason for creating this PR.

It works fine.

job link: https://github.com/kubernetes-sigs/kind/pull/2285/checks?check_run_id=2681813882#step:5:12

tao12345666333:
using #2285 replaced.

/close

PR #2231, opened 2021-05-06
bump to go1.16.4

tao12345666333:
We have just released Go versions 1.16.4 and 1.15.12, minor point releases.

This minor release includes a security fix according to the new security policy (#44918).

ReadRequest and ReadResponse in net/http can hit an unrecoverable panic when reading a very large header (over 7MB on 64-bit architectures, or over 4MB on 32-bit ones). Transport and Client are vulnerable and the program can be made to crash by a malicious server.  Server is not vulnerable by default, but can be if the default max header of 1MB is overridden by setting Server.MaxHeaderBytes to a higher value, in which case the program can be made to crash by a malicious client.

This also affects golang.org/x/net/http2/h2c and HeaderValuesContainsToken in golang.org/x/net/http/httpguts, and is fixed in golang.org/x/n...@v0.0.0-20210428140749-89ef3d95e781.

This is issue [#45710](https://github.com/golang/go/issues/45710) and CVE-2021-31525.

Thanks to Guido Vranken who reported the crash as part of the Ethereum 2.0 bounty program.

View the release notes for more information:
    https://golang.org/doc/devel/release.html#go1.16.minor


aojea:
/approve
seems kubernets/kubernetes didn't bumped it yet :thinking:

tao12345666333:
@aojea  https://github.com/kubernetes/release/pull/2059 release team has been bumped it.

need a lgtm label 😸

BenTheElder:
note for anyone reading along: there's no HTTP server using this go version.

thanks for updating anyhow!

PR #2235, opened 2021-05-07
[WIP] Support remote podman

afrittoli:
Gather info about the provider using `podman info` instead of
looking on the host directly. This allows starting a cluster
when the host is remote, for instance running in a VM on
a MacOS or Windows host.

Not all the info available via `docker info` is available in
`podman info`, so logging a warning when using podman.

Signed-off-by: Andrea Frittoli <andrea.frittoli@gmail.com>

### DIFF
--- a/pkg/cluster/internal/providers/podman/provider.go
+++ b/pkg/cluster/internal/providers/podman/provider.go
@@ -19,7 +19,6 @@ package podman
 import (
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"net"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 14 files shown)


pkg/cluster/internal/providers/podman/provider.go
```diff
@@ -358,45 +357,46 @@ func (p *provider) CollectLogs(dir string, nodes []nodes.Node) error {
...
-	return p.info, nil
+	return p.info, err
 }
 
-func info(logger log.Logger) *providers.ProviderInfo {
-	euid := os.Geteuid()
-	info := &providers.ProviderInfo{
-		Rootless: euid != 0,
-	}
-	if _, err := os.Stat("/sys/fs/cgroup/cgroup.controllers"); err == nil {
```
aojea, line 372: and .... do we stop handling the local podman case?

aojea:
is there a way to know if podman is working in remote or local mode?
I rather prefer to branch on that that removing current code

pkg/cluster/internal/providers/podman/provider.go
aojea, line 372: maybe, we just need to skip this check if we can detect is a remote podman

```
if !podman_remote(U && _, err := os.Stat("/sys/fs/cgroup/cgroup.controllers"); err == nil {
```

aojea: let me summon the expert, 
@vrothberg can I know from the podman cli if is running remote or local?

vrothberg: Not yet but I opened https://github.com/containers/podman/issues/10289. @baude, do you know a workaround?

aojea: @afrittoli can we try with this approach?
https://github.com/containers/podman/issues/10289#issuecomment-838302394
something like
```go
// podman_remote returns true if podman is using a remote client
// xref: https://github.com/containers/podman/issues/10289
func podman_remote() bool{
  ... check https://github.com/containers/podman/issues/10289#issuecomment-838302394 ...
}
```

BenTheElder:
This change needs to be compatible with older podman versions.

I suspect instead of trying to detect if podman is remote or not it needs to check the podman version anyhow before selecting the format string to request, since go text templates explode on unknown fields.

aojea:
> This change needs to be compatible with older podman versions.
> 
> I suspect instead of trying to detect if podman is remote or not it needs to check the podman version anyhow before selecting the format string to request, since go text templates explode on unknown fields.

?

```
if _, err = os.exec(podman --remote info); err != nil {
 podman is local
}
```

BenTheElder:
I mean the fields available to the format string depend on the podman version. So we should just read the podman version and if it's new enough we request the new fields and if not we collect the other fields from podman and the rest from reading local as we already support?

You don't need to check if it's local or not, either it's new enough to get this info for remote and we do that, or it's too old and we should still get the info for local. That's purely based on version.

afrittoli:
> I mean the fields available to the format string depend on the podman version. So we should just read the podman version and if it's new enough we request the new fields and if not we collect the other fields from podman and the rest from reading local as we already support?
> 
> You don't need to check if it's local or not, either it's new enough to get this info for remote and we do that, or it's too old and we should still get the info for local. That's purely based on version.

If we want to support older versions of podman we do indeed check the version. 
The alternative could be - since support is experimental anyways - to only support newer versions. I suspect that might break someone's workflow if they upgrade to the latest `kind`, but given the experimental tag on the feature, it feel like it would be acceptable to require a podman upgrade along with a kind upgrade? 
In any case I'm happy to oblige is the `kind` team prefers to continue supporting the older versions of podman.

In terms of local vs. remote, the `podman info` command does not provide all the info that `kind` expects. It is possible to work with the info only (like I've done in this PR), but we could still detect the local case and collect the extra info there.

To sum up, based on the various comments, my proposal would be to:

- detect podman version
- parse the output of `podman info` according to the version
- detect if running local of remote
- if local collect the info not available from the `podman info` command from the local system
- if not enough info is available to run, fail with a specific error message

Does this make sense?
This week I have no bandwidth to work on this, but I'd be happy to come back to it next week.

aojea:
podman info will provide the "remote" info, it has merged in master https://github.com/containers/podman/pull/10300

I honestly will not try to support older podman versions, we have too many `if` branches already and the remote case seems relatively new

BenTheElder:
I would suggest a slightly more specific order:
- detect podman version (and error if too old due to difficult to deal with bugs)
- collect as much info as we can for the podman version, including an attempt to detect local vs remote
- if we do not detect remote, assume local
- if we are local and do not have enough info, detect from local host details (probe vfs etc.)

@aojea I think time has shown that we will need to switch on the version anyhow as we cannot depend on format structs to be stable. We already do this for e.g. port mapping.

aojea:
> I would suggest a slightly more specific order:
> 
> * detect podman version (and error if too old due to difficult to deal with bugs)

THIS ^^^

> * collect as much info as we can for the podman version, including an attempt to detect local vs remote
> * if we do not detect remote, assume local
> * if we are local and do not have enough info, detect from local host details (probe vfs etc.)
> 
> @aojea I think time has shown that we will need to switch on the version anyhow as we cannot depend on format structs to be stable. We already do this for e.g. port mapping.

sounds great

BenTheElder:
Does anyone know if this general approach is still required? I think podman has made some improvements here 🤞

afrittoli:
I tried recently kind with podman on an M1 Mac and it all worked well, so this can be closed now.
I'm sorry o never followed up on this PR.

BenTheElder:
not at all, thank you!

PR #2239, opened 2021-05-08
add nosetup flag for create cluster subcommand

zhanghe9702:
when user don't need setup cluster, this flag is very useful


BenTheElder:
I would prefer that you use the API for this. This is not a use case we're targeting with the tool, this API is available for kinder https://github.com/kubernetes/kubeadm/tree/master/kinder

Also, when adding a feature, please see our contributing guide first: https://kind.sigs.k8s.io/docs/contributing/getting-started/#4-reaching-out

zhanghe9702:
thx! got it

Issue #2240, opened 2021-05-11
CrashLoopBackOff Error in kube-proxy with kernel versions 5.12.2.arch1-1 and 5.10.35-1-lts


hyutota:
**What happened:** After creating the cluster with `kind create cluster`, the kube-proxy pod have a CrashLoopBackOff Error. This happens at the kernel versions 5.12.2.arch1-1 and 5.10.35-1-lts. With kernel versions 5.12.1.arch1-1 and 5.10.34-1-lts I didn't had the issue.

**What you expected to happen**: All pods in the cluster should start without problems.

**How to reproduce it (as minimally and precisely as possible)**: On a Arch Linux install with kernel version 5.12.2.arch1-1 or 5.10.35-1-lts with docker installed download the latest version of kind and run `kind create cluster`.

**Anything else we need to know?**: 

- Log of kube-proxy pod:
```
I0511 11:47:28.906526       1 node.go:172] Successfully retrieved node IP: 172.18.0.2                                                                                
I0511 11:47:28.906613       1 server_others.go:142] kube-proxy node IP is an IPv4 address (172.18.0.2), assume IPv4 operation
I0511 11:47:28.953210       1 server_others.go:185] Using iptables Proxier.                                                                                          
I0511 11:47:28.953346       1 server_others.go:192] creating dualStackProxier for iptables.
W0511 11:47:28.960804       1 server_others.go:492] detect-local-mode set to ClusterCIDR, but no IPv6 cluster CIDR defined, , defaulting to no-op detect-local for I
I0511 11:47:28.962804       1 server.go:650] Version: v1.20.2                                                                                                        
I0511 11:47:28.965997       1 conntrack.go:100] Set sysctl 'net/netfilter/nf_conntrack_max' to 131072                                                                
F0511 11:47:28.966114       1 server.go:495] open /proc/sys/net/netfilter/nf_conntrack_max: permission denied
```

- Events from pod:
```
Events:                                                                                                                                                              
Type     Reason     Age               From               Message                                                                                                   
----     ------     ----              ----               -------                                                                                                   
Normal   Scheduled  48s               default-scheduler  Successfully assigned kube-system/kube-proxy-s7w5w to kind-control-plane
Normal   Pulled     2s (x4 over 48s)  kubelet            Container image "k8s.gcr.io/kube-proxy:v1.20.2" already present on machine
Normal   Created    2s (x4 over 45s)  kubelet            Created container kube-proxy                                                                              
[...]

PR #2241, opened 2021-05-11
don't set conntrack parameters in kube-proxy

aojea:
It seems the kernel doesn't allow to set some conntrack fields
from non-init netns:

netfilter: conntrack: Make global sysctls readonly in non-init netns
https://github.com/torvalds/linux/commit/671c54ea8c7ff47bd88444f3fffb65bf9799ce43

By default kube-proxy tries to set them, hence failing and the pods
crashlooping.

We can configure kube-proxy to net try to set these values in kubeadm.

Fixes: #2240

### DIFF
--- a/pkg/cluster/internal/kubeadm/config.go
+++ b/pkg/cluster/internal/kubeadm/config.go
@@ -278,6 +278,10 @@ mode: "{{ .KubeProxyMode }}"
 {{end}}{{end}}
 iptables:
   minSyncPeriod: 1s
+conntrack:
+# Skip setting sysctl value "net.netfilter.nf_conntrack_max"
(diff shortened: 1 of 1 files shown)


aojea:
/hold
the 3 jobs fail the conntrack test
> May 11 20:33:40.355: no valid conntrack entry for port 11302 on node 172.18.0.3: wrong TCP CLOSE_WAIT timeout: 48 expected: 3600

aojea:
/hold cancel
it seems that only nf_conntrack_max is global, we can set the tcp close wait

BenTheElder:
/lgtm
/approve
thank you!

PR #2255, opened 2021-05-18
possibly fix cross build

BenTheElder:
still not fully root caused but some debugging shows that 

`$ KUBE_BUILD_PLATFORMS="linux/arm64" KUBE_BUILD_CONFORMANCE=n KUBE_BUILD_HYPERKUBE=n KUBE_VERBOSE=0 build/run.sh make all WHAT="cmd/kubeadm cmd/kubectl cmd/kubelet"`

does not work as expected

while

`$ build/run.sh make all WHAT="cmd/kubeadm cmd/kubectl cmd/kubelet" KUBE_BUILD_PLATFORMS="linux/arm64" KUBE_BUILD_CONFORMANCE=n KUBE_BUILD_HYPERKUBE=n KUBE_VERBOSE=0 `

does, on 1.14

### DIFF
--- a/hack/release/build/push-node.sh
+++ b/hack/release/build/push-node.sh
@@ -44,6 +44,8 @@ ARCHES="${ARCHES:-amd64 arm64}"
 IFS=" " read -r -a __arches__ <<< "$ARCHES"
 
 set -x
+# ensure clean build
+(cd "${KUBEROOT}" && make clean)
 # get kubernetes version
 version_line="$(cd "${KUBEROOT}"; ./hack/print-workspace-status.sh | grep 'gitVersion')"
 kube_version="${version_line#"gitVersion "}"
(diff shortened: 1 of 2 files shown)


BenTheElder:
testing currently on 1.14

BenTheElder:
this fixed 1.14.10 at least (was working from the back to speed up between the two of us)
/assign @amwat

BenTheElder:
there's a *different* issue with 1.20.X (and 1.19.X? possibly more?) that is upstream in kubernetes where cross compiling some things just doesn't work properly in our environments AFAICT, specifically when cross building some of the core images it fails to use the correct platform's image. 

that will need a different solution / isn't related to this.

BenTheElder:
the issue mentioned above is in 1.19 and 1.20 but not 1.18.19, so we should be able to fix upstream in Kubernetes still potentially.

PR #2256, opened 2021-05-18
podman: make Info() robust

### LINKED ISSUE #2254: [Podman] ERROR: failed to create cluster: running kind with rootless provider requires setting systemd property "Delegate=yes", see https://kind.sigs.k8s.io/docs/user/rootless/
**What happened**:

kind + rootless podman fails, even when the systemd property "Delegate=yes" is set.

```console
ERROR: failed to create cluster: running kind with rootless provider requires setting systemd property "Delegate=yes", see https://kind.sigs.k8s.io/docs/user/rootless/
```

The issue does not happen with Rootless Docker.

Podman itself can handle cgroup delegation as expected
```console
$ podman run  --rm --cpus=0.42 --memory=42m --pids-limit 42 -w /sys/fs/cgroup docker.io/library/alpine cat cpu.max memory.max pids.max
42000 100000
44040192
42
```

It should work.

Set up a vagrant VM with the following Vagrantfile:
```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

# Vagrant box for testing kind with cgroup v2
Vagrant.configure("2") do |config|
  config.vm.box = "fedora/33-cloud-base"
  memory = 2048
  cpus = 2
  config.vm.provider :virtualbox do |v|
    v.memory = memory
    v.cpus = cpus
  end
  config.vm.provider :libvirt do |v|
    v.memory = memory
    v.cpus = cpus
  end
  config.vm.provision "install-packages", type: "shell", run: "once" do |sh|
    sh.inline = <<~SHELL
    set -eux -o pipefail
    # Ensure network-related modules to be loaded
    modprobe tap iptables_nat ip6tables_nat

    # The moby-engine package included in Fedora lacks support for rootless,
    # So we need to install docker-ce and docker-ce-rootless-extras from the upstream.
    curl -fsSL https://get.docker.com | sh
    dnf install -y golang-go make kubernetes-client podman docker-ce-rootless-extras
    systemctl enable --now docker

    # Configuration for rootless: https://kind.sigs.k8s.io/docs/user/rootless/
    loginctl enable-linger vagrant
    mkdir -p "/etc/systemd/system/user@.service.d"
    cat <<EOF >"/etc/systemd/system/user@.service.d/delegate.conf"
    [Service]
    Delegate=yes
    EOF
    systemctl daemon-reload
    SHELL
  end
  config.vm.provision "install-kind", type: "shell", run: "once" do |sh|
    sh.inline = <<~SHELL
    set -eux -o pipefail
    make -C /vagrant install INSTALL_DIR=/usr/local/bin
    SHELL
  end
end
```

Enter into the VM via `vagrant ssh`, and run `KIND_EXPERIMENTAL_PROVIDER=podman kind create cluster` there.

No
**Environment:**

- kind version: (use `kind version`): kind v0.11.0-alpha+7b166ac2b6a94f go1.16.4 linux/amd64
- Kubernetes version: (use `kubectl version`): unrelated
- Podman version: 3.1.2
- OS (e.g. from `/etc/os-release`): Fedora 33

AkihiroSuda:
The previous implementation was not robust for unexpected cgroup paths (issue #2254), and did not support remote Podman hosts.

Fix #2254

### DIFF
--- a/pkg/cluster/internal/providers/podman/provider.go
+++ b/pkg/cluster/internal/providers/podman/provider.go
@@ -19,7 +19,6 @@ package podman
 import (
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"net"
 	"os"
 	"path/filepath"
@@ -359,44 +358,58 @@ func (p *provider) CollectLogs(dir string, nodes []nodes.Node) error {
 // The info is cached on the first time of the execution.
 func (p *provider) Info() (*providers.ProviderInfo, error) {
 	if p.info == nil {
-		p.info = info(p.logger)
+		var err error
+		p.info, err = info(p.logger)
+		if err != nil {
+			return p.info, err
+		}
 	}
 	return p.info, nil
 }
 
-func info(logger log.Logger) *providers.ProviderInfo {
-	euid := os.Geteuid()
+// podmanInfo corresponds to `podman info --format 'json`.
+// The structure is different from `docker info --format '{{json .}}'`,
+// and lacks information about the availability of the cgroup controllers.
+type podmanInfo struct {
+	Host struct {
+		CgroupVersion string `json:"cgroupVersion,omitempty"` // "v2"
+		Security      struct {
+			Rootless bool `json:"rootless,omitempty"`
+		} `json:"security"`
+	} `json:"host"`
+}
+
+// info detects ProviderInfo by executing `podman info --format json`.
+func info(logger log.Logger) (*providers.ProviderInfo, error) {
+	const podman = "podman"
+	args := []string{"info", "--format", "json"}
+	cmd := exec.Command(podman, args...)
+	out, err := exec.Output(cmd)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get podman info (%s %s): %q",
+			podman, strings.Join(args, " "), string(out))
+	}
+	var pInfo podmanInfo
+	if err := json.Unmarshal(out, &pInfo); err != nil {
+		return nil, err
+	}
 	info := &providers.ProviderInfo{
-		Rootless: euid != 0,
+		Rootless: pInfo.Host.Security.Rootless,
+		Cgroup2:  pInfo.Host.CgroupVersion == "v2",
+		// We assume all the cgroup controllers to be available.
+		//
+		// For rootless, this assumption is not always correct,
+		// so we print the warning below.
+		//
+		// TODO: We wiil be able to implement proper cgroup controller detection
+		// after the GA of Podman 3.2.x: https://github.com/containers/podman/pull/10387
+		SupportsMemoryLimit: true, // not guaranteed to be correct
+		SupportsPidsLimit:   true, // not guaranteed to be correct
+		SupportsCPUShares:   true, // not guaranteed to be correct
 	}
-	if _, err := os.Stat("/sys/fs/cgroup/cgroup.controllers"); err == nil {
-		info.Cgroup2 = true
-		// Unlike `docker info`, `podman info` does not print available cgroup controllers.
-		// So we parse "cgroup.subtree_control" file by ourselves.
-		subtreeControl := "/sys/fs/cgroup/cgroup.subtree_control"
-		if info.Rootless {
-			// Change subtreeControl to the path of the systemd user-instance.
-			// Non-systemd hosts are not supported.
-			subtreeControl = fmt.Sprintf("/sys/fs/cgroup/user.slice/user-%d.slice/user@%d.service/cgroup.subtree_control", euid, euid)
-		}
-		if subtreeControlBytes, err := ioutil.ReadFile(subtreeControl); err != nil {
-			logger.Warnf("failed to read %q: %+v", subtreeControl, err)
-		} else {
-			for _, controllerName := range strings.Fields(string(subtreeControlBytes)) {
-				switch controllerName {
-				case "cpu":
-					info.SupportsCPUShares = true
-				case "memory":
-					info.SupportsMemoryLimit = true
-				case "pids":
-					info.SupportsPidsLimit = true
-				}
-			}
-		}
-	} else if !info.Rootless {
-		info.SupportsCPUShares = true
-		info.SupportsMemoryLimit = true
-		info.SupportsPidsLimit = true
+	if info.Rootless {
+		logger.Warn("Cgroup controller detection is not implemented for Podman. " +
+			"If you see cgroup-related errors, you might need to set systemd property \"Delegate=yes\", see https://kind.sigs.k8s.io/docs/user/rootless/")
 	}
-	return info
+	return info, nil
 }


AkihiroSuda:
CI failure is unrelated

```
Reading package lists...
E: Failed to fetch https://packages.microsoft.com/ubuntu/20.04/prod/dists/focal/main/binary-amd64/Packages.bz2  File has unexpected size (73402 != 73160). Mirror sync in progress? [IP: 104.42.185.173 443]
   Hashes of expected file:
    - Filesize:73160 [weak]
    - SHA512:782bf0315c8de4314049e599009430053b75a5015491b3d383b09239fcf31857d52bac3399cef8e813f7b2131dce09b03eaf89d79c5ebd2849402cee7aa9d228
    - SHA256:dbf4bf81b4a0a17dae53f2aff3edd4cf5bb1dcd5dcd216ce5c5c8ab7c507a26a
    - SHA1:0e1b47276cae4168a08d8fd981891d4773f053b9 [weak]
    - MD5Sum:0fcf3ed335cd4efde65300bb48c05afe [weak]
   Release file created at: Sat, 15 May 2021 00:38:08 +0000
E: Some index files failed to download. They have been ignored, or old ones used instead.
```

pkg/cluster/internal/providers/podman/provider.go
```diff
@@ -359,44 +358,80 @@ func (p *provider) CollectLogs(dir string, nodes []nodes.Node) error {
...
+	// Unlike `docker info`, `podman info` lacks availability of
+	// cgroup controllers, so we have to detect them by running a container.
+	args = []string{"run",
+		"--rm",
+		"--workdir=/sys/fs/cgroup",
+		"--cgroupns=private",
+		"--cpus=0.1",
+		"--memory=32m",
+		"--pids-limit=32",
+		"quay.io/libpod/busybox:1.30.1",
```
BenTheElder, line 415: This is not great. we have another host that is depended on at runtime :/
I think some orgs also are not able to use busybox due to licensing...?

AkihiroSuda: What's the problem of licensing?

AkihiroSuda: Would it be better to just skip cgroup detection, and print `logrus.Warn("Cgroup controller detection is not implemented for Podman. If you see cgroup-related errors, you might need to set systemd property "Delegate=yes", see https://kind.sigs.k8s.io/docs/user/rootless/)`?

AkihiroSuda: done

flouthoc:
@AkihiroSuda Could you confirm if this is any helpful https://github.com/containers/podman/pull/10387

AkihiroSuda:
> @AkihiroSuda Could you confirm if this is any helpful https://github.com/containers/podman/pull/10387

Thanks, but we need to support existing version of Podman here.

aojea:
> @AkihiroSuda Could you confirm if this is any helpful [containers/podman#10387](https://github.com/containers/podman/pull/10387)

we can do it in a followup once it merges in podman

aojea:
/ok-to-test
/lgtm
I trust Akihiro here, since I really don't have clear the implications of setting those controller values always to true :smile:

pkg/cluster/internal/providers/podman/provider.go
```diff
@@ -359,44 +358,50 @@ func (p *provider) CollectLogs(dir string, nodes []nodes.Node) error {
...
+type podmanInfo struct {
+	Host struct {
+		CgroupVersion string `json:"cgroupVersion,omitempty"` // "v2"
+		Security      struct {
+			Rootless bool `json:"rootless,omitempty"`
+		} `json:"security"`
+	} `json:"host"`
+}
+
+func info(logger log.Logger) (*providers.ProviderInfo, error) {
```
dougsland, line 382: how about add a godoc header for the new version of info? :)

AkihiroSuda: covered by the godoc for `func (p *provider) Info() (*providers.ProviderInfo, error)`

tao12345666333 (commented):
/lgtm

pkg/cluster/internal/providers/podman/provider.go
dougsland: Would still be valid IMO, but it's up to the maintainers of the project.

aojea:
@amwat can you chime in, I'd like a second opinion and you played a lot with podman

pkg/cluster/internal/providers/podman/provider.go
```diff
@@ -359,44 +358,50 @@ func (p *provider) CollectLogs(dir string, nodes []nodes.Node) error {
...
-		}
-	} else if !info.Rootless {
-		info.SupportsCPUShares = true
-		info.SupportsMemoryLimit = true
-		info.SupportsPidsLimit = true
+		Rootless:            pInfo.Host.Security.Rootless,
+		Cgroup2:             pInfo.Host.CgroupVersion == "v2",
+		SupportsMemoryLimit: true, // not guaranteed to be correct
+		SupportsPidsLimit:   true, // not guaranteed to be correct
+		SupportsCPUShares:   true, // not guaranteed to be correct
```
amwat, line 400: probably worth adding an additional comment as to why we default to `true`?

(IIUC rootfull is always `true` and for rootless we are assuming `true` for best effort proceeding with rootless (i.e. not failing `validateProvider`) since we can't detect the actual values)

pkg/cluster/internal/providers/podman/provider.go
```diff
@@ -359,44 +358,50 @@ func (p *provider) CollectLogs(dir string, nodes []nodes.Node) error {
...
+		CgroupVersion string `json:"cgroupVersion,omitempty"` // "v2"
+		Security      struct {
+			Rootless bool `json:"rootless,omitempty"`
+		} `json:"security"`
+	} `json:"host"`
+}
+
+func info(logger log.Logger) (*providers.ProviderInfo, error) {
+	const podman = "podman"
+	args := []string{"info", "--format", "{{json .}}"}
```
amwat, line 384: nit: maybe use `json` directly instead of the go template version of json.

amwat (commented):
Looks good overall,

If our custom detection isn't working as expected relying on whatever info podman is providing and warning as such sgtm.

+1 on following up with actual detection once https://github.com/containers/podman/pull/10387 is in (for newer versions atleast).

pkg/cluster/internal/providers/podman/provider.go
AkihiroSuda: updated

pkg/cluster/internal/providers/podman/provider.go
AkihiroSuda: updated

pkg/cluster/internal/providers/podman/provider.go
AkihiroSuda: updated

aojea:
/lgtm
/approve
Thanks

PR #2257, opened 2021-05-18
Fix podman getSubnets()

aojea:
Fix podman getSubnets()

Since we moved to custom networks, it has to get the information from the kind network
It also carries the fix in https://github.com/kubernetes-sigs/kind/pull/2199 that got stuck on CLA signing

### DIFF
--- a/pkg/cluster/internal/providers/podman/provision.go
+++ b/pkg/cluster/internal/providers/podman/provision.go
@@ -154,7 +154,7 @@ func commonArgs(cfg *config.Cluster, networkName string) ([]string, error) {
 	}
 
 	// pass proxy environment variables
-	proxyEnv, err := getProxyEnv(cfg)
+	proxyEnv, err := getProxyEnv(cfg, networkName)
 	if err != nil {
 		return nil, errors.Wrap(err, "proxy setup error")
 	}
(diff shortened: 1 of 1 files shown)


pkg/cluster/internal/providers/podman/provider.go
```diff
@@ -369,6 +369,14 @@ func info(logger log.Logger) *providers.ProviderInfo {
 	info := &providers.ProviderInfo{
 		Rootless: euid != 0,
 	}
+
+	// check if is a local or remote podman
+	// xref: https://github.com/containers/podman/issues/10289
+	err := exec.Command("podman", "--remote", "info").Run()
+	if err == nil {
+		// podman is remote we can't check cgroups information so we exit now
```
AkihiroSuda, line 377: Alternative PR: https://github.com/kubernetes-sigs/kind/pull/2256

AkihiroSuda (changes requested):

pkg/cluster/internal/providers/podman/provider.go
aojea: Thanks Akihiro, that is much better

BenTheElder:
what's the context? did the previous one ever work? did podman change something?

aojea:
> what's the context? did the previous one ever work? did podman change something?

that is stuck https://github.com/kubernetes-sigs/kind/pull/2199
and when we moved to custom networks I forget to pass the kind network

> error unrelated
E: Failed to fetch https://packages.microsoft.com/ubuntu/20.04/prod/dists/focal/main/binary-amd64/Packages.bz2  File has unexpected size (73402 != 73160). Mirror sync in progress? [IP: 40.74.238.15 443]
   Hashes of expected file:
    - Filesize:73160 [weak]

BenTheElder:
/lgtm
/approve
I still think json should be high on the list, but /shrug

aojea:
> I still think json should be high on the list, but /shrug

I know I know ... too many things sorry

Issue #2262, opened 2021-05-19
race restarting containerd with config patches


igorcalabria:
**What happened**:

Cluster is unable to pull images on `localhost:5000` after following the local registry guide https://kind.sigs.k8s.io/docs/user/local-registry/. It fails with dial tcp 127.0.0.1:5000: connect: connection refused

**What you expected to happen**:

Cluster is able to pull images on `localhost:5000` registry

**How to reproduce it (as minimally and precisely as possible)**:

On host machine 

```
./kind-with-registry.sh 
docker pull busybox:latest
docker tag busybox:latest localhost:5000/busybox:latest 
docker push  localhost:5000/busybox:latest
```
inside control pane(`docker exec -it kind-control-pane bash`):

```
root@kind-control-plane:/# crictl --debug pull localhost:5000/busybox:latest
```

It should fail with the error above(in my setup at least)

**Anything else we need to know?**:

From my limited investigation, it seems that containerd config patch was applied after the process started. The patch had no effect on the running process and this may explain why it tried to pull the image from 127.0.0.1 instead of the configured endpoint. To fix this I simply ran `service containerd restart` inside the control plane container and everything started working. Hope this helps

**Environment:**

- kind version: (use `kind version`): kind v0.11.0 go1.16 linux/amd64
- Kubernetes version: (use `kubectl version`):
```
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.2", GitCommit:"faecb196815e248d3ecfb03c680a4507229c2a56", GitTreeState:"clean", BuildDate:"2021-01-13T13:28:09Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.1", GitCommit:"5e58841cce77d4bc13713ad2b91fa0d961e69192", GitTreeState:"clean", BuildDate:"2021-05-18T01:10:20Z", GoVersion:"go1.16.4", Compiler:"gc", Platform:"linux/amd64"}
```
- Docker version: (use `docker info`):
```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Build with BuildKit (Docker Inc., v0.5.1-docker)
  scan: Docker Scan (Docker Inc., v0.7.0)

Server:
 Containers: 2
  Running: 2
  Paused: 0
  Stopped: 0
[...]

PR #2274, opened 2021-05-24
reload containerd on config changes

aojea:
Wait until the system finish to boot to restart containerd, after we patch it
Fixes: https://github.com/kubernetes-sigs/kind/issues/2262

### DIFF
--- a/pkg/cluster/internal/create/actions/config/config.go
+++ b/pkg/cluster/internal/create/actions/config/config.go
@@ -185,8 +185,8 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 					return errors.Wrap(err, "failed to write patched containerd config")
 				}
 				// restart containerd now that we've re-configured it
-				// skip if the systemd (also the containerd) is not running
-				if err := node.Command("bash", "-c", `! systemctl is-system-running || systemctl restart containerd`).Run(); err != nil {
+				// wait until boot has finished
+				if err := node.Command("bash", "-c", `systemctl is-system-running --wait && systemctl restart containerd`).Run(); err != nil {
 					return errors.Wrap(err, "failed to restart containerd after patching config")
 				}
 				return nil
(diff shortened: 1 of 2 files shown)


BenTheElder (changes requested):
the way we're doing image loading is considered a bug to fix.
we should not mimic it here. the tools are available to do this correctly.

aojea:
TIL `systemctl is-system-running --wait`

aojea:
fedora images seems to not be available
```

URL: ["https://vagrantcloud.com/fedora/33-cloud-base"]
Error: The requested URL returned error: 403
Bringing machine 'default' up with 'virtualbox' provider..
```

AkihiroSuda:
> @AkihiroSuda should we switch to fedora34?

I have a PR to switch the default to 34: https://github.com/kubernetes-sigs/kind/pull/2220
But depends on https://github.com/kubernetes-sigs/kind/pull/2230.

Opened https://github.com/kubernetes-sigs/kind/pull/2279 as an attempt to deflake 33.

pkg/cluster/internal/create/actions/config/config.go
```diff
@@ -185,8 +185,8 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 					return errors.Wrap(err, "failed to write patched containerd config")
 				}
 				// restart containerd now that we've re-configured it
-				// skip if the systemd (also the containerd) is not running
-				if err := node.Command("bash", "-c", `! systemctl is-system-running || systemctl restart containerd`).Run(); err != nil {
+				// wait until boot has finished
```
BenTheElder, line 188: we want to only restart it if we observed it was already running after completing the patching, if it wasn't running waiting and restarting it is:
a) slow
b) unnecessary (as it never read the old unpatched config)

there's still a race as to wether or not it starts before we check to restart it, but if we check to restart it and it wasn't running we're fine and if it was running we restart it and wether or not it read the old config it's fine and we don't waste much time.

the problem previously is an ineffective check for it it is running.

BenTheElder:
filed https://github.com/kubernetes-sigs/kind/pull/2382 https://github.com/kubernetes-sigs/kind/pull/2274#discussion_r675212707

aojea:
Duplicate

PR #2278, opened 2021-05-25
rootless: use fuse-overlayfs snapshotter by default

### LINKED ISSUE #2275: rootless: use fuse-overlayfs by default

Use `fuse-overlayfs` containerd snapshotter by default when running with Rootless Docker/Podman.

**Why is this needed**:

kind v0.11 uses `overlayfs` containerd snapshotter for rootless, but rootless overlayfs has very complex requirement: 
```
(debian kernel && modprobe opton permit_mounts_in_userns==1) || (ubuntu kernel) || (kernel >= 5.11 && !selinux)
```

And yet Debian variant is known to be broken (while Ubuntu variant is not): https://github.com/moby/moby/issues/42302

OTOH, fuse-overlayfs just requires kernel >= 4.18.

AkihiroSuda:
kind v0.11 uses `overlayfs` containerd snapshotter for rootless, but rootless overlayfs has very complex requirement: 
```
(debian kernel && modprobe opton permit_mounts_in_userns==1) || (ubuntu kernel) || (kernel >= 5.11 && !selinux)
```

And yet Debian variant is known to be broken (while Ubuntu variant is not): https://github.com/moby/moby/issues/42302

OTOH, fuse-overlayfs just requires kernel >= 4.18. 

The snapshotter can be still rolled back to `overlayfs` by setting `KIND_EXPERIMENTAL_CONTAINERD_SNAPSHOTTER=overlayfs`.

Close #2275

- - -
Node image: akihirosuda/tmp-kind-node:g42e6ce83-v1.21.1

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -55,6 +55,14 @@ ARG CNI_PLUGINS_AMD64_SHA256SUM="962100bbc4baeaaa5748cdbfce941f756b1531c2eadb290
 ARG CNI_PLUGINS_ARM64_SHA256SUM="ef17764ffd6cdcb16d76401bac1db6acc050c9b088f1be5efa0e094ea3b01df0"
 ARG CNI_PLUGINS_PPC64LE_SHA256SUM="5bd3c82ef248e5c6cc388f25545aa5a7d318778e5f9bc0a31475361bb27acefe"
 
+# Configure containerd-fuse-overlayfs snapshotter binary from upstream
+ARG CONTAINERD_FUSE_OVERLAYFS_VERSION="1.0.2"
+ARG CONTAINERD_FUSE_OVERLAYFS_TARBALL="v${CONTAINERD_FUSE_OVERLAYFS_VERSION}/containerd-fuse-overlayfs-${CONTAINERD_FUSE_OVERLAYFS_VERSION}-linux-${TARGETARCH}.tar.gz"
+ARG CONTAINERD_FUSE_OVERLAYFS_URL="https://github.com/containerd/fuse-overlayfs-snapshotter/releases/download/${CONTAINERD_FUSE_OVERLAYFS_TARBALL}"
+ARG CONTAINERD_FUSE_OVERLAYFS_AMD64_SHA256SUM="1f1e69f71b5ea568e93e40059af1b02a377ac0966d2acd27e4cce388a27af218"
+ARG CONTAINERD_FUSE_OVERLAYFS_ARM64_SHA256SUM="7ade1a44d880b3fb8eaa3c5ff7d3890a43b777d06ec80439c9a51ae35626c83c"
+ARG CONTAINERD_FUSE_OVERLAYFS_PPC64LE_SHA256SUM="eaf9bdd3de4514546945ea93119acea2b7bfa55ced43766e20adabddd5d20978"
+
 # copy in static files
 # all scripts are 0755: http://www.filepermissions.com/file-permission/0755
 COPY --chmod=0755 files/usr/local/bin/* /usr/local/bin/
@@ -102,7 +110,7 @@ RUN echo "Installing Packages ..." \
       conntrack iptables iproute2 ethtool socat util-linux mount ebtables udev kmod \
       libseccomp2 pigz \
       bash ca-certificates curl rsync \
-      nfs-common \
+      nfs-common fuse-overlayfs \
     && find /lib/systemd/system/sysinit.target.wants/ -name "systemd-tmpfiles-setup.service" -delete \
     && rm -f /lib/systemd/system/multi-user.target.wants/* \
     && rm -f /etc/systemd/system/*.wants/* \
@@ -160,6 +168,15 @@ RUN echo "Installing Packages ..." \
          -o -iname loopback \
       \) \
       -delete \
+ && echo "Installing containerd-fuse-overlayfs ..." \
+    && curl -sSL --retry 5 --output /tmp/containerd-fuse-overlayfs.${TARGETARCH}.tgz "${CONTAINERD_FUSE_OVERLAYFS_URL}" \
+    && echo "${CONTAINERD_FUSE_OVERLAYFS_AMD64_SHA256SUM}  /tmp/containerd-fuse-overlayfs.amd64.tgz" | tee /tmp/containerd-fuse-overlayfs.sha256 \
+    && echo "${CONTAINERD_FUSE_OVERLAYFS_ARM64_SHA256SUM}  /tmp/containerd-fuse-overlayfs.arm64.tgz" | tee -a /tmp/containerd-fuse-overlayfs.sha256 \
+    && echo "${CONTAINERD_FUSE_OVERLAYFS_PPC64LE_SHA256SUM}  /tmp/containerd-fuse-overlayfs.ppc64le.tgz" | tee -a /tmp/containerd-fuse-overlayfs.sha256 \
+    && sha256sum --ignore-missing -c /tmp/containerd-fuse-overlayfs.sha256 \
+    && rm -f /tmp/containerd-fuse-overlayfs.sha256 \
+    && tar -C /usr/local/bin -xzvf /tmp/containerd-fuse-overlayfs.${TARGETARCH}.tgz \
+    && rm -rf /tmp/containerd-fuse-overlayfs.${TARGETARCH}.tgz \
  && echo "Ensuring /etc/kubernetes/manifests" \
     && mkdir -p /etc/kubernetes/manifests \
  && echo "Adjusting systemd-tmpfiles timer" \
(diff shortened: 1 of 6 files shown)


AkihiroSuda:
CI failure is unrelated

```
URL: ["https://vagrantcloud.com/fedora/33-cloud-base"]
Error: The requested URL returned error: 403
```

aojea:
> CI failure is unrelated
> 
> ```
> URL: ["https://vagrantcloud.com/fedora/33-cloud-base"]
> Error: The requested URL returned error: 403
> ```

oh,you already detected it

images/base/Dockerfile
```diff
@@ -55,6 +55,14 @@ ARG CNI_PLUGINS_AMD64_SHA256SUM="962100bbc4baeaaa5748cdbfce941f756b1531c2eadb290
 ARG CNI_PLUGINS_ARM64_SHA256SUM="ef17764ffd6cdcb16d76401bac1db6acc050c9b088f1be5efa0e094ea3b01df0"
 ARG CNI_PLUGINS_PPC64LE_SHA256SUM="5bd3c82ef248e5c6cc388f25545aa5a7d318778e5f9bc0a31475361bb27acefe"
 
+# Configure containerd-fuse-overlayfs snapshotter binary from upstream
+ARG CONTAINERD_FUSE_OVERLAYFS_VERSION="1.0.2"
+ARG CONTAINERD_FUSE_OVERLAYFS_TARBALL="v${CONTAINERD_FUSE_OVERLAYFS_VERSION}/containerd-fuse-overlayfs-${CONTAINERD_FUSE_OVERLAYFS_VERSION}-linux-${TARGETARCH}.tar.gz"
+ARG CONTAINERD_FUSE_OVERLAYFS_URL="https://github.com/containerd/fuse-overlayfs-snapshotter/releases/download/${CONTAINERD_FUSE_OVERLAYFS_TARBALL}"
```
BenTheElder, line 61: Will this eventually be merged in? It seems like this is crucial enough to bake in 🤔

pkg/cluster/internal/providers/docker/provision.go
```diff
@@ -243,6 +243,11 @@ func runArgsForNode(node *config.Node, clusterIPFamily config.ClusterIPFamily, n
 		"--volume", "/var",
 		// some k8s things want to read /lib/modules
 		"--volume", "/lib/modules:/lib/modules:ro",
+		// propagate KIND_EXPERIMENTAL_CONTAINERD_SNAPSHOTTER to the entrypoint script
+		"-e", "KIND_EXPERIMENTAL_CONTAINERD_SNAPSHOTTER",
```
BenTheElder, line 247: Is there any reason to not just enable this when rootless and forget the env?

pkg/cluster/internal/providers/podman/provision.go
```diff
@@ -206,6 +206,11 @@ func runArgsForNode(node *config.Node, clusterIPFamily config.ClusterIPFamily, n
 		"--volume", fmt.Sprintf("%s:/var:suid,exec,dev", varVolume),
 		// some k8s things want to read /lib/modules
 		"--volume", "/lib/modules:/lib/modules:ro",
+		// propagate KIND_EXPERIMENTAL_CONTAINERD_SNAPSHOTTER to the entrypoint script
+		"-e", "KIND_EXPERIMENTAL_CONTAINERD_SNAPSHOTTER",
+		// enable /dev/fuse explicitly for fuse-overlayfs
+		// (Rootless Podman does not automatically mount /dev/fuse with --privileged)
```
BenTheElder, line 212: What's the reason for that? I'm only familiar with the rough concept of fuse 😅

images/base/Dockerfile
AkihiroSuda: No, the containerd/containerd repo does not accept a new built-in snapshotter.

images/base/files/etc/systemd/system/containerd-fuse-overlayfs.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
```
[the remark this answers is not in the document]
AkihiroSuda: Yes, this is already conditionally enabled in the entrypoint (the ` systemctl enable containerd-fuse-overlayfs` line).

The service file is always added, but it is not enabled by default.

pkg/cluster/internal/providers/docker/provision.go
AkihiroSuda: This env var is for users who want to opt-in to `KIND_EXPERIMENTAL_CONTAINERD_SNAPSHOTTER=overlayfs` (kernel-mode real overlayfs) for better performance.
This env var is not needed for most users.

pkg/cluster/internal/providers/podman/provision.go
AkihiroSuda: Because `fuse-overlayfs` needs `/dev/fuse` to mount FUSE.

pkg/cluster/internal/providers/docker/provision.go
BenTheElder: Ah, hopefully in the future there some way we can check if real overlayfs is viable and automate selection.

images/base/files/etc/systemd/system/containerd-fuse-overlayfs.service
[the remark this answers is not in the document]
BenTheElder: thanks 👍

images/base/Dockerfile
BenTheElder: > No, the containerd/containerd repo does not accept a new built-in snapshotter.

That seems unfortunate, especially with it being essentially required for rootless. I hope they may reconsider for this case.

> If having a separate binary is not comfortable, maybe we can consider patching kind-ci/containerd-nightlies. (Just adding import _ "github.com/containerd/fuse-overlayfs-snapshotter/plugin" line)

We probably don't want to patch containerd sources, but maybe worth considering if we can do it with our own stub source importing or something.

Users building images for compliance (e.g. EKS AIUI) will still need to update their builds to fix this though. It's less that we need to add a binary to the image and more that there's another unverified binary they may need to notice and start overriding + building. If we keep doing that it makes updating kind a riskier proposition.

It's also a little annoying to need to run more services though / extra entrypoint complexity though and I imagine the binary size is smaller when merged in. But it is *not* a deal breaker. It's still more important to make sure kind works.

pkg/cluster/internal/providers/podman/provision.go
BenTheElder: > Because `fuse-overlayfs` needs `/dev/fuse` to mount FUSE.

Sorry, that part I get, what I meant ways: What's the reason for (the comment):

> Rootless Podman does not automatically mount /dev/fuse with --privileged

Why this deviation?

AkihiroSuda: Kinda bug / TODO (of runc) due to complexity

BenTheElder:
btw I'd be happy to sponsor your kubernetes organization membership, which would remove the `needs-ok-to-test` from all future PRs immediately, if nothing else 😅

pkg/cluster/internal/providers/podman/provision.go
BenTheElder: thanks!

BenTheElder:
/ok-to-test

https://github.com/kubernetes/community/blob/master/community-membership.md#member

I'm sure @aojea would be happy to be a second sponsor

aojea:
>  I'm sure @aojea would be happy to be a second sponsor

yep, absolutely

AkihiroSuda:
Thanks, opened a request in https://github.com/kubernetes/org/issues/2740

I should have requested this a few years ago... 😅

BenTheElder:
#2287 we're going to cut a release with the runc update / CVE patch, but after that's done I intend to update the base image and move forward with this PR. thank you for your patience.

BenTheElder (approved):
/lgtm
/approve

pkg/cluster/internal/providers/docker/provision.go
```diff
@@ -243,6 +243,11 @@ func runArgsForNode(node *config.Node, clusterIPFamily config.ClusterIPFamily, n
...
+		// enable /dev/fuse explicitly for fuse-overlayfs
+		// (Rootless Docker does not automatically mount /dev/fuse with --privileged)
+		"--device", "/dev/fuse",
```
BenTheElder, line 250: what if fuse isn't enabled / mounted?

pkg/cluster/internal/providers/podman/provision.go
BenTheElder: it sounds like they don't because it may not be present see e.g. https://github.com/openshift/enhancements/issues/362

Issue #2287, opened 2021-05-27
v0.11.1


BenTheElder:
**What would you like to be added**:

Release v0.11.1.

**Why is this needed**:

We've merged very few changes since v0.11, but importantly there was a runc CVE and we've upgraded runc/containerd in the base image. This also may improve performance, as on a similar timeframe to upgrading this Kubernetes's CI improved.

NOTE:
>  Applying this attack requires the attacker to start containers with a slightly

https://github.com/opencontainers/runc/security/advisories/GHSA-c3xm-pvg7-gh7r

Untrusted / suspicious configuration should *not* be run with kind, but we should still patch this.

We should push new binaries with the already updated 1.21.1 default image, and a list of new images.

amwat:
I can start a patch release for v0.11.1

BenTheElder:
Started pushing images back from 1.20, amit started from 1.14. Pushing from https://github.com/kubernetes-sigs/kind/pull/2288

see also https://github.com/kubernetes-sigs/kind/pull/2288#issuecomment-850007782

BenTheElder:
`v1.20.7@sha256:cbeaf907fc78ac97ce7b625e4bf0de16e3ea725daf6b04f930bd14c67c671ff9`
`v1.19.11@sha256:07db187ae84b4b7de440a73886f008cf903fcf5764ba8106a9fd5243d6f32729`

BenTheElder:
`kindest/node:v1.18.19@sha256:7af1492e19b3192a79f606e43c35fb741e520d195f96399284515f077b3b622c`

BenTheElder:
`kindest/node:v1.17.17@sha256:66f1d0d91a88b8a001811e2f1054af60eef3b669a9a74f9b6db871f2f1eeed00`

amwat:
https://github.com/kubernetes-sigs/kind/releases/tag/v0.11.1 is now out

/close

PR #2288, opened 2021-05-27
Release 0.11.1

amwat:
https://github.com/kubernetes-sigs/kind/issues/2287


BenTheElder:
let's merge this so we mark the commit we want to be the release but *not* push the tag until the images / relnotes are ready.

then when we do push the tag it doesn't matter i some more PRs have merged after this, but the tag won't exist and trigger attempts to use this release until it's ready.

Issue #2271, opened 2021-05-24
Error rootless-cni-infra: modprobe: can't change directory to '/lib/modules': No such file or director


bagnaram:
**What happened**:
Preparing Nodes stage fails under rootless Podman.

**What you expected to happen**:
It should work.

**How to reproduce it (as minimally and precisely as possible)**:

Install podman 3.1.2-2
Install kind v0.11.0 go1.16.4 linux/amd64

Follow steps https://wiki.archlinux.org/title/Podman#Rootless_Podman to enable rootless Podman
Follow  steps https://kind.sigs.k8s.io/docs/user/rootless/ to enable rootless Kind

```sh
KIND_EXPERIMENTAL_PROVIDER=podman kind create cluster
using podman due to KIND_EXPERIMENTAL_PROVIDER
enabling experimental podman provider
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.21.1) 🖼
 ✗ Preparing nodes 📦
ERROR: failed to create cluster: podman run error: command "podman run --hostname kind-control-plane --name kind-control-plane --label io.x-k8s.kind.role=control-plane --privileged --tmpfs /tmp --tmpfs /run --volume b553212b8e4f27ffdc4913d05e9fa094741bace1b2b64c2b6e31958b4ef27678:/var:suid,exec,dev --volume /lib/modules:/lib/modules:ro --detach --tty --net kind --label io.x-k8s.kind.cluster=kind -e container=podman --publish=127.0.0.1:39369:6443/tcp -e KUBECONFIG=/etc/kubernetes/admin.conf kindest/node@sha256:fae9a58f17f18f06aeac9772ca8b5ac680ebbed985e266f711d936e91d113bad" failed with error: exit status 127
Command Output: Error: command rootless-cni-infra [alloc 3dca4d05101ca0d3e9eb0f86c095f8ec2ec4d14d0570c5b44fc630f65c544f5d kind kind-control-plane   ] in container eb38361e52803d8166430e540033d44259e2364ff0c3a77b75f26dfd5b0e9ff7 failed with status 1, stdout="", stderr="failed to list chains: running [/sbin/ip6tables -t nat -S --wait]: exit status 3: modprobe: can't change directory to '/lib/modules': No such file or directory\nip6tables v1.8.4 (legacy): can't initialize ip6tables table `nat': Table does not exist (do you need to insmod?)\nPerhaps ip6tables or your kernel needs to be upgraded.\n\n"
[...]

PR #2289, opened 2021-05-28
match https://github.com/kubernetes-sigs/iptables-wrappers/blob/maste…

BenTheElder:
…r/iptables-wrapper-installer.sh selection logic

see: https://github.com/kubernetes-sigs/kind/issues/2271#issuecomment-850256442

this is one of the only deviations in functionality from the wrapper there.

### DIFF
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -312,18 +312,15 @@ fix_kmsg() {
 }
 
 select_iptables() {
-  # based on: https://github.com/kubernetes/kubernetes/blob/ffe93b3979486feb41a0f85191bdd189cbd56ccc/build/debian-iptables/iptables-wrapper
-  local mode=nft
+  # based on: https://github.com/kubernetes-sigs/iptables-wrappers/blob/97b01f43a8e8db07840fc4b95e833a37c0d36b12/iptables-wrapper-installer.sh
+  local mode num_legacy_lines num_nft_lines
   num_legacy_lines=$( (iptables-legacy-save || true; ip6tables-legacy-save || true) 2>/dev/null | grep '^-' | wc -l || true)
-  if [ "${num_legacy_lines}" -ge 10 ]; then
+  num_nft_lines=$( (timeout 5 sh -c "iptables-nft-save; ip6tables-nft-save" || true) 2>/dev/null | grep '^-' | wc -l || true)
+  if [ "${num_legacy_lines}" -ge "${num_nft_lines}" ]; then
     mode=legacy
   else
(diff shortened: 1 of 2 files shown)


BenTheElder:
note: since I haven't bumped the base this doesn't change anything in CI, this is just for review. namely @aojea 🙃

aojea:
yeah, Dan Winship make it work for the case of users that have both rules on the system

BenTheElder:
the logic we had previously was the same as the source at https://github.com/kubernetes/kubernetes/blob/ffe93b3979486feb41a0f85191bdd189cbd56ccc/build/debian-iptables/iptables-wrapper

but that's changed now in https://github.com/kubernetes-sigs/iptables-wrappers/blob/97b01f43a8e8db07840fc4b95e833a37c0d36b12/iptables-wrapper-installer.sh

I haven't checked yet if kube-proxy was also updated

aojea:
> https://github.com/kubernetes/release/blob/0539f9b737ef056d6007b991bd25b92a8706eb7e/images/build/debian-iptables/buster/iptables-wrapper#L28-L38 it has not been updated in kube-proxy

I wonder if that may explain recent bugs open about kube-proxy delays on endpoints :thinking:

aojea:
@danwinship it seems that there are 2 detection modes:

1. # lines legacy > 10
2- # lines legacy > # lines nft

which one do you think is better?
https://github.com/kubernetes-sigs/iptables-wrappers/blob/97b01f43a8e8db07840fc4b95e833a37c0d36b12/iptables-wrapper-installer.sh#L112-L141

danwinship:
The version of the script in the iptables-wrappers repo is best. You ideally want to check `num lines legacy > num lines nft` because that's more accurate, but you don't want to do that if you have the buggy version of iptables-nft that will hang and need to be timed out and killed. So the installer in iptables-wrappers does the good way if you have a good iptables-nft binary, and the less good way if you have a bad binary

BenTheElder:
I filed https://github.com/kubernetes/release/pull/2106 to correct the kube-proxy base image to the second version, we should be on iptables >= 1.8.5 there so we don't need the bug workaround version.

BenTheElder:
This PR also updates kind to that variant.

BenTheElder:
we should push new image(s) anyhow
/hold

BenTheElder:
need to fix https://github.com/kubernetes-sigs/kind/issues/2313 first

BenTheElder:
will rebase on https://github.com/kubernetes-sigs/kind/pull/2320

BenTheElder:
thought we'd be landing other prioritized changes but since we're not yet, rebased this one.

BenTheElder:
re-applying as this is just a rebase

PR #2297, opened 2021-06-03
bump to go 1.16.5

BenTheElder:
couple CVEs
https://groups.google.com/g/golang-announce/c/RgCMkAEQjSI


tao12345666333 (commented):
/lgtm

/retest

BenTheElder:
the cgroup v2 CI is fialing but shouldn't be related.

we have too many github actions jobs now, they're hitting the dockerhub limits (I had more things come up and have not been able to follow up on the docker side).

BenTheElder:
@aojea I don't think we need multi and single node ipv6 and ipv4 for starters. We have docker multinode in all the prow jobs.

aojea:
> @aojea I don't think we need multi and single node ipv6 and ipv4 for starters. We have docker multinode in all the prow jobs.

I've added them for podman, there was a bug that they only worked in single node and this is the only coverage we have.
I think that for docker we don't need multinode except in rootless 
Anyway, you are right and this is starting to grow exponential, I need to think in a better testing matrix

PR #2301, opened 2021-06-09
Adds SCTP to comment of protocol in PortMapping

atoato88:
This PR updates comments of protocol in PortMapping, because protocol supports "SCTP" also.

### DIFF
--- a/pkg/apis/config/v1alpha4/types.go
+++ b/pkg/apis/config/v1alpha4/types.go
@@ -277,7 +277,7 @@ type PortMapping struct {
 	HostPort int32 `yaml:"hostPort,omitempty"`
 	// TODO: add protocol (tcp/udp) and port-ranges
 	ListenAddress string `yaml:"listenAddress,omitempty"`
-	// Protocol (TCP/UDP)
+	// Protocol (TCP/UDP/SCTP)
 	Protocol PortMappingProtocol `yaml:"protocol,omitempty"`
 }
 
--- a/pkg/internal/apis/config/types.go
+++ b/pkg/internal/apis/config/types.go
@@ -235,7 +235,7 @@ type PortMapping struct {
 	HostPort int32
 	// TODO: add protocol (tcp/udp) and port-ranges
 	ListenAddress string
-	// Protocol (TCP/UDP)
+	// Protocol (TCP/UDP/SCTP)
 	Protocol PortMappingProtocol
 }
 


atoato88:
/assign @amwat
It's small updates, could you reveiw this?

aojea:
this seems not true, at least for docker using userlandproxy, however the docs says docker supports it :thinking: 

```
kind create cluster --config kind.conf 
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.20.2) 🖼
 ✗ Preparing nodes 📦  
ERROR: failed to create cluster: docker run error: command "docker run --hostname kind-control-plane --name kind-control-plane --label io.x-k8s.kind.role=control-plane --privileged --security-opt seccomp=unconfined --security-opt apparmor=unconfined --tmpfs /tmp --tmpfs /run --volume /var --volume /lib/modules:/lib/modules:ro --detach --tty --label io.x-k8s.kind.cluster=kind --net kind --restart=on-failure:1 --init=false --publish=127.0.0.1:80:80/SCTP --publish=127.0.0.1:37393:6443/TCP -e KUBECONFIG=/etc/kubernetes/admin.conf kindest/node:v1.20.2@sha256:15d3b5c4f521a84896ed1ead1b14e4774d02202d5c65ab68f30eeaf310a3b1a7" failed with error: exit status 125
Command Output: cc6e7c21715acadbea10bd73543bed7d589d388d80ffa5c3f7124ee6d9746932
docker: Error response from daemon: driver failed programming external connectivity on endpoint kind-control-plane (ee4d58a0fb861fff87e96e72b7224cc66c2759d75ea41c547a5ad1fe83f651e3): Error starting userland proxy: protocol not supported.
```

aojea:
it seems that userland proxy in docker can be disabled, and it will be disabled by default soon
https://github.com/moby/moby/issues/14856
should we add a comment mentioning that SCTP doesn't work with docker userland proxy?

atoato88:
I recently start to read codes of this repository.
As you say, I found [docs here](https://kind.sigs.k8s.io/docs/user/configuration/#extra-port-mappings) which say `optional: set the protocol to one of TCP, UDP, SCTP.`.

On the other hand, [here comments](https://github.com/kubernetes-sigs/kind/blob/main/pkg/apis/config/v1alpha4/types.go#L280) `TCP/UDP` only.
That's why I created this PR.

aojea:
Yeah, it is ok, I just discovered the userland proxy issue thanks to you.

I think we should go ahead with your pr, but just adding the comment that it doesn't work with docker userland proxy

pkg/apis/config/v1alpha4/types.go
```diff
@@ -277,7 +277,7 @@ type PortMapping struct {
 	HostPort int32 `yaml:"hostPort,omitempty"`
 	// TODO: add protocol (tcp/udp) and port-ranges
```
atoato88, line 278: Maybe this comment is target of this PR? :thinking:
like as follows: 
```suggestion
	// TODO: add port-ranges
```

pkg/internal/apis/config/types.go
```diff
@@ -235,7 +235,7 @@ type PortMapping struct {
 	HostPort int32
 	// TODO: add protocol (tcp/udp) and port-ranges
```
atoato88, line 236: Same with my comment above.

aojea:
/lgtm
/approve
The comments are correct, docs says sctp is supported
Thanks

atoato88:
Thank you to approve this! :smile:

Issue #2160, opened 2021-03-25
ability to include additional images in the kindest/node image


dkoshkin:
**What would you like to be added**:
Currently the kind node images comes with the core kubernetes images and a few other images that it needs to start.
It would be great if it was possible to include additional images into the kind node image that it can have ready to be used.

**Why is this needed**:
In an air-gapped cluster where you may only need to have a few additional images at bootstrap, having pre-baked images in a `kind` image could be a simple way to get access to a Kubernetes API.

Some options here could be:

* Document a way to use the existing `kindest/node` image as a base image (I tried for a little bit but didn't find anything online and `crictl pull` wasn't working, but would love feedback if it is possible).
* Extend `kind build node-image` CLI command to add additional images.
* Rely on existing `kind load` command. More complicated since you would need to ship around multiple images, but an option that already works.
* Rely on a local registry. This adds an external dependency that complicates the setup.

BenTheElder:
> Document a way to use the existing kindest/node image as a base image (I tried for a little bit but didn't find anything online and crictl pull wasn't working, but would love feedback if it is possible).

It won't work for the same reason that `kind build node-image` isn't just a `docker build`, you need a privileged container to run container such that you can pull / import images. You can accomplish this external to kind but not via a normal image build.

Another approach might be trying to build a tool / command for this instead of doing it while building the rest of the node image (so users could start from a standard node image).

> Rely on existing kind load command. More complicated since you would need to ship around multiple images, but an option that already works.

Did you know that you can have multiple images into a single tarball? `docker save image1 image2 image3`.
So if this is your only concern, you can do `docker save kindest/node app-image1 app-image2 > images.tar` and copy a single tarball around.

> Rely on a local registry. This adds an external dependency that complicates the setup.

This one also persists image loading across clusters though. And is something we've already been tracking / documenting.

https://github.com/kubernetes-sigs/kind/pull/2159#issuecomment-807499777 has some more questions for considration

joshrosso:
+1 to this desire of loading additional images to create a custom node image.

I've been [working off a fork that has very similar modifications](https://github.com/kubernetes-sigs/kind/compare/main...joshrosso:prepull-custom) to @dkoshkin's work in https://github.com/kubernetes-sigs/kind/pull/2159#issuecomment-807499777. The only difference is I input my images from a `.yaml` file, e.g:

```yaml
preload:
  images:
[...]

PR #2303, opened 2021-06-09
Feature 2160 add images subcommand

curtbushko:
This PR addresses [Feature 2160 - ability to include additional images in the kindest/node image](https://github.com/kubernetes-sigs/kind/issues/2160) and hopefully addresses the concerns in the discussion.

The PR:

- adds a subcommand to the build target `kind build add-image IMAGE [IMAGES...]` in order to keep the command separate from `kind build node-image` 
- pulls images locally before they are added into the resulting image to speed up future builds
- tars those images temporarily (similar to what kind load does)
- starts up the base image and use critl to copy/stream the images into it
- saves the image using docker save

Extras: 
- follows the same patterns as node image when it comes to code layout and args
- has defaults for the base image used (kindest/node:latest) and the resulting image (kindest/custom-node:latest)
- can handle adding multiple images at the same time

An example of the command being used:
```
>> ./bin/kind build add-image --base-image=docker.io/kindest/node:v1.19.11 --image=myrepo/custom-node:latest alpine:3.7
Starting to add images to base image
Creating build container based on docker.io/kindest/node:v1.19.11
Building in kind-build-1623261338-1249758035
Saving images into tar file at /private/var/folders/81/myw19h2108xdx31prlv3qsd00000gn/T/images-tar466634204/images.tar
Importing images into build container kind-build-1623261338-1249758035
Saving new image myrepo/custom-node:latest
sha256:bc17c0f40f826574adb8aa886c29404b02c5e1fd89f2ff8d58f7cb55f59dce1e
Add image build completed.

>> cat <<EOF | kind create cluster --config=-
kind: Cluster
nodes:
- role: control-plane
  image: myrepo/custom-node:latest
EOF

Creating cluster "kind" ...
 ✓ Ensuring node image (myrepo/custom-node:latest) 🖼
 ✓ Preparing nodes 📦
 ✓ Writing configuration 📜
 ✓ Starting control-plane 🕹️
 ✓ Installing CNI 🔌
 ✓ Installing StorageClass 💾

Have a question, bug, or feature request? Let us know! https://kind.sigs.k8s.io/#community 🙂

>> docker exec -it kind-control-plane crictl images
IMAGE                                      TAG                  IMAGE ID            SIZE
docker.io/kindest/kindnetd                 v20210326-1e038dc5   6de166512aa22       54MB
docker.io/library/alpine                   3.7                  51af6a90e1cee       4.47MB
docker.io/rancher/local-path-provisioner   v0.0.14              e422121c9c5f9       13.4MB
k8s.gcr.io/build-image/debian-base         v2.1.0               c7c6c86897b63       21.1MB
k8s.gcr.io/coredns                         1.7.0                bfe3a36ebd252       14MB
k8s.gcr.io/etcd                            3.4.13-0             0369cf4303ffd       86.7MB
k8s.gcr.io/kube-apiserver                  v1.19.11             2fb26b7ebd23a       120MB
k8s.gcr.io/kube-controller-manager         v1.19.11             1af1564e1890d       112MB
k8s.gcr.io/kube-proxy                      v1.19.11             d2c2a8b355f56       120MB
k8s.gcr.io/kube-scheduler                  v1.19.11             120d339f7d67a       47.7MB
k8s.gcr.io/pause                           3.5                  ed210e3e4a5ba       301kB

```

Please let me know if there is something you would like to see changed or fixed.


BenTheElder:
We may need to consider `--arch` support similar to `kind build node-image`. Unfortunately when doing that ...

> pulls images locally before they are added into the resulting image to speed up future builds

That will not work once this supports multi-arch, perhaps best to head that off now. (see: https://github.com/kubernetes-sigs/kind/pull/2176)

curtbushko:
> We may need to consider `--arch` support similar to `kind build node-image`. Unfortunately when doing that ...
> 
> > pulls images locally before they are added into the resulting image to speed up future builds
> 
> That will not work once this supports multi-arch, perhaps best to head that off now. (see: #2176)

Is there a way I could use `docker pull image:tag@sha` and parse out the manifest through docker manifest inspect and still use the local docker cache?

What I am aiming for is being being able to build our product images locally and build a node image for all developers to use.

BenTheElder:
> Is there a way I could use docker pull image:tag@sha and parse out the manifest through docker manifest inspect and still use the local docker cache?

The problem I found is that if you need to ask docker to pull an image for a platform other than the host and then save it for loading you run into issues (e.g. try pulling the image for each platform you're building for and loading it, it will just save the host platform one, there's no way to save --platform), so the only solution I found was to pull for the platform directly to the node container we're going to commit.

I think there's more on this in the linked PR? But without a solution for that multi-arch support is untenable, and multi-arch support is very much needed now.

BenTheElder:
Also AFAIK there is no way to ask docker for layers locally, the only way docker exposes layers is through pushing to registries.

curtbushko:
Thanks for the explanation and insight. I did go through your PR to see how you did it and I even took a dive into docker engine to see how it was doing save. It's too bad that they didn't add --platform to save even though they added it to other commands.

I found something very interesting! It looks like save does save the correct local image but it changes the manifest as it is writing it.

I pulled an odd platform version and built my custom image. I am on a mac.
```
>> docker pull --platform=s390x alpine:3.6

>> docker run alpine:3.6
WARNING: The requested image's platform (linux/s390x) does not match the detected host platform (linux/amd64) and no specific platform was requested
qemu-s390x: warning: 'msa5-base' requires 'kimd-sha-512'.
qemu-s390x: warning: 'msa5-base' requires 'klmd-sha-512'.

>> ./bin/kind build add-image alpine:3.6
Starting to add images to base image
Creating build container based on docker.io/kindest/node:latest
Building in kind-build-1623283038-727782895
Saving images into tar file at /private/var/folders/81/myw19h2108xdx31prlv3qsd00000gn/T/images-tar588984380/images.tar
Importing images into build container kind-build-1623283038-727782895
Saving new image kindest/custom-node:latest
sha256:a362f1ce00e277e34acb9dc2423039f168cd76c6e48678725278d56bbee6de20
Add image build completed.

>> cat <<EOF | kind create cluster --config=-
kind: Cluster
nodes:
- role: control-plane
  image: myrepo/custom-node:latest
EOF

Creating cluster "kind" ...
....
```

Inspect the images locally and in the custom node. The rootfs layers match. 
```
>> docker inspect alpine:3.6 |grep -A5 'RootFS'
        "RootFS": {
            "Type": "layers",
            "Layers": [
                "sha256:f129ff7d977bdbcb6eb0ec9e2e24c6cbe68e60e46d9620e79e09e73ab2e545e5"
            ]
        },

>> docker exec -it kind-control-plane crictl inspecti docker.io/library/alpine:3.6 |grep -A5 'rootfs'
      "rootfs": {
        "type": "layers",
        "diff_ids": [
          "sha256:f129ff7d977bdbcb6eb0ec9e2e24c6cbe68e60e46d9620e79e09e73ab2e545e5"
        ]
      },
```
Even the creation dates are the same!
```
>> docker inspect alpine:3.6 |grep 'Created'
        "Created": "2019-03-08T03:35:59.716135296Z",

>> docker exec -it kind-control-plane crictl inspecti docker.io/library/alpine:3.6 | grep 'created'
        "created": "2019-03-08T03:35:59.716135296Z",

>> docker pull --platform=amd64 alpine:3.6 && docker inspect alpine:3.6 |grep 'Created'
3.6: Pulling from library/alpine
Digest: sha256:66790a2b79e1ea3e1dabac43990c54aca5d1ddf268d9a5a0285e4167c8b24475
Status: Image is up to date for alpine:3.6
docker.io/library/alpine:3.6
        "Created": "2019-03-07T22:20:00.563496859Z",
```

Bonus: If I run an alpine pod in my custom node image, the logs give me:
```
> kubectl logs -f alpine
qemu-s390x: warning: 'msa5-base' requires 'kimd-sha-512'.
qemu-s390x: warning: 'msa5-base' requires 'klmd-sha-512'.
qemu-s390x: warning: 'msa5-base' requires 'kimd-sha-512'.
qemu-s390x: warning: 'msa5-base' requires 'klmd-sha-512'
```

Just for fun, I pulled a mips version of redis, built it into my node and tried running it
```
kubeclt logs -f redis
qemu-mips64el: unable to find CPU model 'max'
```

Digging into this was quite fun!

pkg/build/addimage/buildcontext.go
```diff
@@ -0,0 +1,175 @@
...
+	// pull images to local docker
+	for _, imageName := range c.additionalImages {
+		// Check to see if the image exists and if not pull it
+		_, err := docker.ImageID(imageName)
+		if err != nil {
+			err = docker.Pull(c.logger, imageName, dockerBuildOsAndArch(c.arch), 3)
+			if err != nil {
+				c.logger.Errorf("Add image build Failed! Failed to pull image %v: %v", imageName, err)
+			}
+		}
```
aojea, line 64: nit, this reads better unnesting
```suggestion
              // continue if image already exist, no need to pull
		if err == nil {
		        continue
		}
              err = docker.Pull(c.logger, imageName, dockerBuildOsAndArch(c.arch), 3)
		if err != nil {
			c.logger.Errorf("Add image build Failed! Failed to pull image %v: %v", imageName, err)
		}
	}
```

aojea:
@BenTheElder does it makes sense to duplicate the code for the docker commands `pkg/build/nodeimage/internal/container/docker` ?

pkg/build/addimage/defaults.go
```diff
@@ -0,0 +1,23 @@
...
+limitations under the License.
+*/
+
+package addimage
+
+// DefaultImage is the default name:tag for the built image
+const DefaultImage = "kindest/custom-node:latest"
+
+// DefaultBaseImage is the default base image used
+const DefaultBaseImage = "docker.io/kindest/node:latest"
```
aojea, line 23: do we want to use different defaults than nodeimage https://github.com/kubernetes-sigs/kind/blob/9fe964443fe34eb3bb59080d513eadeaa35370cf/pkg/build/nodeimage/defaults.go#L23

curtbushko:
> @BenTheElder does it makes sense to duplicate the code for the docker commands `pkg/build/nodeimage/internal/container/docker` ?

In general, I would say no but should it also go to a shared location?

pkg/build/addimage/defaults.go
curtbushko: I will push a change for this.

curtbushko: Now, I am using the default node image that gets built.

https://github.com/kubernetes-sigs/kind/blob/e3864c0be8cc3b17380d7d98d7f2632212ef79b6/pkg/apis/config/defaults/image.go#L21

BenTheElder:
> @BenTheElder does it makes sense to duplicate the code for the docker commands pkg/build/nodeimage/internal/container/docker ?

eh, I prescribe to WET: Write Everything Thrice. Once or twice is not enough for an abstraction. but more importantly the code in any `internal/` is not meant to be a publicly importable API (We did not put enough work into making it a good stable API, they are just implementation details of the good APIs), so it will still have to be private, and so far we have made the rule that `pkg/cmd/` is doable using only it's own internal details or the public APIs, the CLI does not import any shared internals so we know the CLI commands are vetting the public APIs.

raphink:
Is there any news on this PR? This would be extremely useful!

batistein:
Any progress on this PR? This is extremly useful and is already used as a fork in multiple projekt for round about a year. No issues so far, so I don't see any reason to defer a merge. Clean-up tasks could be done in a follow-up PR!

pkg/build/addimage/buildcontext.go
```diff
@@ -0,0 +1,177 @@
...
+
+		err = docker.Pull(c.logger, imageName, dockerBuildOsAndArch(c.arch), 3)
+		if err != nil {
+			c.logger.Errorf("Add image build Failed! Failed to pull image %v: %v", imageName, err)
+		}
+	}
+
+	// create build container
+	c.logger.V(0).Info("Creating build container based on " + c.baseImage)
+	// pull images to local docker
```
BenTheElder, line 71: we shouldn't do this anymore, it's too problematic for multi-arch, as discussed in previous review comments.

we pull straight to containerd in the current version of the node image build

BenTheElder: this is also a UX issue that is not thought out here, we should probably have `--pull` vs no `--pull`.

without `--pull` we should use the local docker stored image or tarball, which makes a predictable experience for locally built images

with `--pull` + `--arch` we can support multi-arch builds

`--arch` without `--pull` should possibly be prohibited or throw a warning, unless the inputs are tarballs not image names.

BenTheElder: supporting tarballs would make it easier to deal with architectures locally without a registry

BenTheElder (commented):
I've personally been overwhelmed earlier this year, and then the past few weeks on vacation.

Will try to be more responsive here.

pkg/build/addimage/internal/container/docker/archive.go
```diff
@@ -0,0 +1,167 @@
...
+	if err := json.Unmarshal(raw, &entries); err != nil {
+		return nil, err
+	}
+
+	for i, entry := range entries {
+		fixed := make([]string, len(entry.RepoTags))
+		for i, tag := range entry.RepoTags {
+			parts := strings.Split(tag, ":")
+			if len(parts) > 2 {
+				return nil, fmt.Errorf("invalid repotag: %s", entry)
```
anhpngt, line 148: This check would prevent images stored in a local registry I believe. For example:
```sh
localhost:5001/asia.gcr.io/my-gcp-project/my-docker-image:my-tag
```

anhpngt: not sure if it would be an intended behavior

pjotre86:
This is a great feature. I just successfully tested it on my machine, great job!
I'm using the fork currently to build custom node images containing a bunch of images I need to have in place in order to run a big application at my job. This saves me a lot of time and traffic.
Great job, I really hope this makes it into the main repo soon!
Cheers!

raphink:
Is there anything I can do to help with this feature?

stmcginnis:
Looks like there are some comments to be addressed yet. Commits should also be squashed.

curtbushko:
Closing in favour of the more recent https://github.com/kubernetes-sigs/kind/pull/3634

BenTheElder:
Thanks @curtbushko, I think we're close on that one, followed up on the discussion there again.

PR #2306, opened 2021-06-12
Update ingress from v1beta1 to v1

juniorjbn:
Update ingress from v1bet1 to v1 and update `Ingress.spec.rules[0].http.paths[0].backend` changes.
https://kubernetes.io/docs/concepts/services-networking/ingress/

### DIFF
--- a/site/static/examples/ingress/usage.yaml
+++ b/site/static/examples/ingress/usage.yaml
@@ -46,7 +46,7 @@ spec:
   # Default port used by the image
   - port: 5678
 ---
-apiVersion: networking.k8s.io/v1beta1
+apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   name: example-ingress
@@ -54,12 +54,18 @@ spec:
   rules:
   - http:
       paths:
-      - path: /foo
+      - pathType: Prefix
+        path: "/foo"
         backend:
-          serviceName: foo-service
-          servicePort: 5678
-      - path: /bar
+          service:
+            name: foo-service
+            port:
+              number: 5678
+      - pathType: Prefix
+        path: "/bar"
         backend:
-          serviceName: bar-service
-          servicePort: 5678
+          service:
+            name: bar-service
+            port:
+              number: 5678
 ---


aojea:
we discussed this some time ago, the problem is that this we are always going to break someone
ingress/v1 is 1.19 IIRC, it won't work for previous versions
ingress/v1beta1 will stop to work since 1.22 :shrug:

BenTheElder:
> ingress/v1 is 1.19 IIRC, it won't work for previous versions
> ingress/v1beta1 will stop to work since 1.22 🤷

with 1.22 coming it might be time to switch though, we can add a note about this to the docs.
older releases have to eventually be best effort I think.

site/static/examples/ingress/usage.yaml
```diff
@@ -46,20 +46,26 @@ spec:
...
-          serviceName: foo-service
-          servicePort: 5678
-      - path: /bar
+          service:
+            name: foo-service
+            port:
+              number: 5678
+      - pathType: Prefix
+        path: "/bar"
         backend:
```
aojea, line 66:
```suggestion
        defaultBackend:
```

aojea: ups , this is ok, sorry

site/static/examples/ingress/usage.yaml
```diff
@@ -46,20 +46,26 @@ spec:
...
+apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   name: example-ingress
 spec:
   rules:
   - http:
       paths:
-      - path: /foo
+      - pathType: Prefix
```
aojea, line 57: we enter here on the compatiblity problem, some ingresses need
`ImplementationSpecific`

aojea:
> with 1.22 coming it might be time to switch though, we can add a note about this to the docs.
> older releases have to eventually be best effort I think.

this is going to be a drama XD, it took me a week to update k/k to v1, 
but this is clear the way
/ok-to-test
@juniorjbn we don't have CI , have you tried it with at least one of the ingresses of the docs?

site/static/examples/ingress/usage.yaml
aojea: I can't say which is better though

aojea:
cc @rikatz

aojea:
at least nginx ingress is not going to work https://github.com/kubernetes/ingress-nginx/issues/7145

rikatz:
Wanna release a beta next week :)

BenTheElder:
Ingress Impl can't support both for some time?

aojea:
/hold
until we have some positive feedback that the examples work

aojea:
related https://github.com/kubernetes/ingress-nginx/issues/7305 , right now the examples doesn't work for nginx

BenTheElder:
I think we should probably roll forward with this, people are confused about the ingress-nginx failures and kind is nominally shipping v1.21, with 1.19+ having v1 AIUI. Yes some users are still on <= v1.18, but, similarly we cannot patch bugs in Kubernetes for those versions anymore, the last patch release is done upstream since we first started talking about this issue.

site/static/examples/ingress/usage.yaml
BenTheElder: we don't have to support *all* ingresses here, I'd prefer to use ingresses that can use portable upstream specs for the purposes of kind anyhow.

BenTheElder (approved):
/lgtm
/approve

site/static/examples/ingress/usage.yaml
aojea: I can't say which pathType is more portabl XD

rikatz: Prefix is more portable :)

With Fixed you have all the traffic directed to a specific path only catched :)

BenTheElder:
bypassing github actions flakes.

PR #2311, opened 2021-06-16
upgrade base images

BenTheElder:
drops kindnetd size 👍 

| image | size (compressed/dockerhub) | size (local) |
|------|----------------|-------|
| `kindest/kindnetd:v20210326-1e038dc5` | 51.46 MB | 118MB |
|`kindest/kindnetd:v20210616-5aea9f9e` | 37.84 MB | 99.1MB |


aojea:
but we need to also bump the base image with these new images, don't we?

BenTheElder:
No they're not included in the base image.

aojea:
/lgtm
/approve
ahh right, release image != base image

Issue #2313, opened 2021-06-17
Starting control-plane error with kindest/node:v1.21.1


dkoshkin:
**What happened**:

```
$ kind create cluster --image kindest/node:v1.21.1
I0617 18:18:40.122102     208 round_trippers.go:454] GET https://kind-control-plane:6443/healthz?timeout=10s  in 0 milliseconds
I0617 18:18:40.622409     208 round_trippers.go:454] GET https://kind-control-plane:6443/healthz?timeout=10s  in 0 milliseconds
couldn't initialize a Kubernetes cluster
k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/init.runWaitControlPlanePhase
	/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/init/waitcontrolplane.go:114
k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/workflow.(*Runner).Run.func1
	/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/workflow/runner.go:234
k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/workflow.(*Runner).visitAll
	/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/workflow/runner.go:421
k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/workflow.(*Runner).Run
	/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/workflow/runner.go:207
k8s.io/kubernetes/cmd/kubeadm/app/cmd.newCmdInit.func1
	/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/cmd/kubeadm/app/cmd/init.go:152
k8s.io/kubernetes/vendor/github.com/spf13/cobra.(*Command).execute
	/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/vendor/github.com/spf13/cobra/command.go:850
k8s.io/kubernetes/vendor/github.com/spf13/cobra.(*Command).ExecuteC
[...]

PR #2320, opened 2021-06-22
Revert bad entrypoint bash

BenTheElder:
fixes https://github.com/kubernetes-sigs/kind/issues/2313


aojea:
/lgtm

What is the deal with cut Vs ask?

BenTheElder:
cut takes a specific delimeter
awk with print like that is reading the args and selecting the nth arg.
if your input is not just say space delimited, but perhaps varying whitespace, they are not equivilant.

I didn't look too deeply at what happened yet, I just tested reverting the entrypoint because it was the smallest and most trivial thing to test of the two suspect changes. Then I looked at the diff and it dawned on me that they are not the same for all inputs so 🙃

Issue #2331, opened 2021-06-25
sha256sums for kind binaries


cokegen:
**What would you like to be added**: sha256sums / gpg signatures for kind binaries

**Why is this needed**: some basic form of verification at least for the kind binaries would be good to have, hashes, gpg signatures, etc.

BenTheElder:
We can and should definitely add shasums to the binaries so download integrity can be checked. 

I don't think the Kubernetes project currently has a solution for signing, if/when it does we'll follow suit here.

kind binary builds should be reproducible at least though, so anyone can vet that the binary + sha reflects the sources fairly easily.

---

This should make a pretty good starter issue for anyone interested, basically update this script which we run as part of the release process to produce a `${file_name}.sha256sum` file containing the hash of `${file_name}` for each build output file after we build:

https://github.com/kubernetes-sigs/kind/blob/c64d7644516f7214c063b4cee5c324fd14e6dd1e/hack/release/build/cross.sh 

Then when we release these will be alongside the binary outputs and we'll upload them as well

/help
/good-first-issue
/sig testing

asher-lab:
hello @viveksyngh is there a readme or some place where the sha256sum/sha512sum are published? like from kubernetes where the sha512 sum are in there readme file? https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.27.md thanks

BenTheElder:
They're alongside the release binaries. See "assets" at the bottom of e.g. https://github.com/kubernetes-sigs/kind/releases/tag/v0.20.0, which is standard for github hosted releases.

PR #2336, opened 2021-06-27
feat: add sha256sums for kind binaries

viveksyngh:
- Add sha256sum file for kind binaries build as a part of release process

Fixes: #2331

### DIFF
--- a/hack/release/build/cross.sh
+++ b/hack/release/build/cross.sh
@@ -48,3 +48,9 @@ export GOOS=linux GOARCH=arm64
 export GOOS=linux GOARCH=ppc64le
 EOF
 )
+
+# add sha256 for binaries
+cd "${REPO_ROOT}"/bin
+for f in kind-*; do
+    shasum -a 256 "$f" > "$f".sha256sum;
+done


hack/release/build/cross.sh
```diff
@@ -48,3 +48,6 @@ export GOOS=linux GOARCH=arm64
 export GOOS=linux GOARCH=ppc64le
 EOF
 )
+
+# add sha256 for binaries
+for f in ${REPO_ROOT}/bin/kind-*; do shasum -a 256 $f > $f.sha256sum; done
```
xinydev, line 53: the original output for `shasum` is something like `hashValue  fileName`, I think it only needs the hash value here?  

https://github.com/kubernetes/kompose/blob/0b331d9e5dede975d62ca382b2dd7dff89019f6e/script/release.sh#L243

aojea: how do you know the filaname to check it with the command?

```
[aojea@aojea-laptop ~]$ shasum -a 256 ifac.go > ifac.sha
[aojea@aojea-laptop ~]$ sha256sum -c ifac.sha 
ifac.go: OK
[aojea@aojea-laptop ~]$ cp ifac.sha ifac.shaonly
[aojea@aojea-laptop ~]$ vi ifac.shaonly #removed the filename
[aojea@aojea-laptop ~]$ sha256sum -c ifac.shaonly 
sha256sum: ifac.shaonly: no properly formatted SHA256 checksum lines found
```

aojea:
the verify job fails 
```
In ./hack/release/build/cross.sh line 53:
for f in ${REPO_ROOT}/bin/kind-*; do shasum -a 256 $f > $f.sha256sum; done
         ^----------^ SC2231: Quote expansions in this for loop glob to prevent wordsplitting, e.g. "$dir"/*.txt .
                                                   ^-- SC2086: Double quote to prevent globbing and word splitting.
                                                        ^-- SC2086: Double quote to prevent globbing and word splitting.
```

hack/release/build/cross.sh
xinydev: You are right, it is not easy to check.  

however,I found that we have a lot of current tar files are only have the hash value,I'm just curious what is the reason for this.

`
➜  ~ gsutil cat gs://kubernetes-release/release/v1.21.0-alpha.0/kubernetes-test-windows-amd64.tar.gz.sha256   
e3b8a57d0bba4665f54874e4478bf2b360254314a482f4ba36f15ce228c561b7%     

➜  ~ gsutil cat gs://kubernetes-release/release/v1.21.0-alpha.0/kubernetes-src.tar.gz.sha256   
7eaf8fa1aa03afead4292d1e10dbc839439b4cd45a5a749bac242f8cf0040f45%    
`

aojea: not an expert here, I always used the hash filename format, that is what shasum reports directly

hack/release/build/cross.sh
```diff
@@ -48,3 +48,6 @@ export GOOS=linux GOARCH=arm64
 export GOOS=linux GOARCH=ppc64le
 EOF
 )
+
+# add sha256 for binaries
+for f in "${REPO_ROOT}"/bin/kind-*; do shasum -a 256 "$f" > "$f".sha256sum; done
```
aojea, line 53: don't use the oneliner and keep the last line, is easier to read
```suggestion
for f in "${REPO_ROOT}"/bin/kind-*; do 
    shasum -a 256 "$f" > "$f".sha256sum
done

```

viveksyngh: Updated

aojea: hmm this adds the whole path to the generated file, I think that we have to cd first in the folder
```
cd "${REPO_ROOT}"/bin
for f in kind-*; do
      shasum -a 256 "$f" > "$f".sha256sum
done
```

viveksyngh: yup it was adding full path. I have updated that to change to the directory and then run sha command.

hack/release/build/cross.sh
viveksyngh: I have also used hash filename format for one of the CLI in the past. Not sure what benefit other approach has.

aojea:
/lgtm
thanks
/assign @BenTheElder

hack/release/build/cross.sh
```diff
@@ -48,3 +48,9 @@ export GOOS=linux GOARCH=arm64
 export GOOS=linux GOARCH=ppc64le
 EOF
 )
+
+# add sha256 for binaries
+cd "${REPO_ROOT}"/bin
```
BenTheElder, line 53: TODO: refactor to not need `cd`
TODO: update loop to ignore files with .sha256sum extension.

BenTheElder (approved):
/lgtm
/approve
/hold
thanks!

BenTheElder:
/hold cancel
we can iterate later, thank you!

Issue #2296, opened 2021-06-03
create cluster fails with permission denied error on /dev/dma_heap


elmiko:
**What happened**:

After updating my Fedora Linux to the latest kernel (`5.11.12-300.fc34.x86_64`) i tried to run `kind create cluster` and hit this error:

```
Creating cluster "mgmt" ...                                                                                                               
 ✓ Ensuring node image (kindest/node:v1.21.1) 🖼                                                                                           
 ✗ Preparing nodes 📦                                                                                                                     
ERROR: failed to create cluster: docker run error: command "docker run --hostname mgmt-control-plane --name mgmt-control-plane --label io.
x-k8s.kind.role=control-plane --privileged --security-opt seccomp=unconfined --security-opt apparmor=unconfined --tmpfs /tmp --tmpfs /run 
--volume /var --volume /lib/modules:/lib/modules:ro -e KIND_EXPERIMENTAL_CONTAINERD_SNAPSHOTTER --device /dev/fuse --detach --tty --label 
io.x-k8s.kind.cluster=mgmt --net kind --restart=on-failure:1 --init=false --volume=/var/run/docker.sock:/var/run/docker.sock --publish=127
.0.0.1:40963:6443/TCP -e KUBECONFIG=/etc/kubernetes/admin.conf kindest/node:v1.21.1@sha256:80773e2069dd4a80a4929fdef050c1e463e1d5578c2cd9a
f3962cfbc230e1500" failed with error: exit status 126                                                                                     
Command Output: 6a36fe59f40c4d4338b75c0b357d0ceb4f789319550fe61196c3f76dbdb8ec3b                                                          
docker: Error response from daemon: open /dev/dma_heap: permission denied.
```

**What you expected to happen**:

Cluster to be created as normal.

**How to reproduce it (as minimally and precisely as possible)**:

1. Install Fedora 34
2. run `dnf --refresh upgrade -y` to update to latest kernel
3. run `kind create cluster`

**Anything else we need to know?**:

**Environment:**

- kind version: `kind v0.12.0-alpha+1188d9bd86afbf go1.16.4 linux/amd64`
- Kubernetes version: `1.21.1`
- Docker version: `20.10.6`
- OS (e.g. from `/etc/os-release`): Fedora 34

BenTheElder:
Is this rootless docker by any chance?
https://kind.sigs.k8s.io/docs/user/rootless/

Can you share the rest of `docker info`?

[...]

PR #2340, opened 2021-06-29
add known issue for Fedora 33 SELinux

dlipovetsky:
Describes temporary workaround for anyone running into #2296 on Fedora 33.


site/content/docs/user/known-issues.md
```diff
@@ -322,6 +323,16 @@ systemctl restart firewalld
 See [#1547 (comment)](https://github.com/kubernetes-sigs/kind/issues/1547#issuecomment-623756313)
 and [Docker and Fedora 32 article](https://fedoramagazine.org/docker-and-fedora-32/)
 
+## Fedora 33 SELinux
```
BenTheElder, line 326: Is this one fixed on Fedora 44?
Can we add a note about that?

Last time we added a note it was a mistake to include the fedora version, at the time it didn't apply to older versions but it has applied to newer ones.

BenTheElder (approved):
/lgtm
/approve
thanks!

tao12345666333 (commented):
/lgtm

PR #2345, opened 2021-07-01
upgrade direct dependencies, drop unnecessary requirement on x/sys

BenTheElder:
we don't depend on anything new in x/sys, we're just using it as part of go-isatty.


go.mod
```diff
@@ -5,14 +5,13 @@ go 1.14
...
+	github.com/mattn/go-isatty v0.0.13
+	github.com/pelletier/go-toml v1.9.3
 	github.com/pkg/errors v0.9.1
-	github.com/spf13/cobra v1.1.1
+	github.com/spf13/cobra v1.1.3
 	github.com/spf13/pflag v1.0.5
-	golang.org/x/sys v0.0.0-20210124154548-22da62e12c0c // indirect
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
-	k8s.io/apimachinery v0.20.2
+	k8s.io/apimachinery v0.21.2
```
BenTheElder, line 15: we should really look at dropping this dependency. staging repos are not fun to have as transitive deps and we're not using much from here.

go.mod
```diff
@@ -5,14 +5,13 @@ go 1.14
...
 	github.com/alessio/shellescape v1.4.1
-	github.com/evanphx/json-patch/v5 v5.2.0
-	github.com/mattn/go-isatty v0.0.12
-	github.com/pelletier/go-toml v1.8.1
+	github.com/evanphx/json-patch/v5 v5.5.0
+	github.com/mattn/go-isatty v0.0.13
+	github.com/pelletier/go-toml v1.9.3
 	github.com/pkg/errors v0.9.1
-	github.com/spf13/cobra v1.1.1
+	github.com/spf13/cobra v1.1.3
```
BenTheElder, line 12: v1.2.0 just released per #k8s-code-organization but looking at the release notes require some changes to auto-completion, will follow up with that as the PR is still good and already tested / minimal changes required.

PR #2346, opened 2021-07-01
api version v1beta1 deprecated

kundan2707:
Ingrsss api version vebeta1 deprecated for version 1.19 and higher.
kind latest version is using 1.21 so it needs to be updated to v1


BenTheElder:
Please see existing PR and accompanying discussion https://github.com/kubernetes-sigs/kind/pull/2306

The assertion that we need to change is also wrong. It is not removed until 1.22 in Kubernetes.

BenTheElder:
/triage duplicate
/close
the prior PR is merging soon.

PR #2347, opened 2021-07-01
Change Ingress NGINX script to use stable version

rikatz:
Signed-off-by: Ricardo Pchevuzinske Katz <ricardo.katz@gmail.com>

We have a bunch of users complaining (with reason) about the new non-released deploy manifest.

Examples: 
https://github.com/kubernetes/ingress-nginx/issues/7306
https://github.com/kubernetes/ingress-nginx/issues/7305

This change adds a stable.txt file that might be used as part of the install script in KinD

### DIFF
--- a/site/content/docs/user/ingress.md
+++ b/site/content/docs/user/ingress.md
@@ -116,7 +116,8 @@ Additional information about Contour can be found at: [projectcontour.io](https:
 ### Ingress NGINX
 
 {{< codeFromInline lang="bash" >}}
-kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/static/provider/kind/deploy.yaml
+VERSION=$(curl https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/stable.txt)
+kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/${VERSION}/deploy/static/provider/kind/deploy.yaml
 {{< /codeFromInline >}}
 
 The manifests contains kind specific patches to forward the hostPorts to the


BenTheElder (approved):
/lgtm
/approve
thanks!

site/content/docs/user/ingress.md
```diff
@@ -116,7 +116,8 @@ Additional information about Contour can be found at: [projectcontour.io](https:
 ### Ingress NGINX
 
 {{< codeFromInline lang="bash" >}}
-kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/static/provider/kind/deploy.yaml
+VERSION=$(curl https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/stable.txt)
```
BenTheElder, line 119: one little nit: this now only works on mac/linux, this doc did work on windows too
when v1 is released we can revisit this perhaps

rikatz: Completely agreed. Will check with @cpanato the best way actually to have like a "stable" branch and always merge there

BenTheElder: maybe a copy of the stable version's yaml somewhere? 🤔

BenTheElder:
(does not affect CI, just the netlify build/preview which passed)

site/content/docs/user/ingress.md
rikatz: hum, will think. Does sig-release have some storage bucket? Or maybe I can ask one for k8s.io workgroup

cpanato: k8s.io wg is the best place to we setup this, maybe we can setup a staging account, not sure

PR #2357, opened 2021-07-08
Clean up extra spaces

jayonlau:
Clean up extra spaces, although these errors are not important, they affect the code specification.


BenTheElder:
Thank you for the PR, but these are intended.

PR #2358, opened 2021-07-09
Update api doc redirects

xinydev:
https://kind.sigs.k8s.io/v1alpha3 works
https://kind.sigs.k8s.io/v1alpha4 not works

### DIFF
--- a/site/data/apiVersions.yaml
+++ b/site/data/apiVersions.yaml
@@ -1,1 +1,1 @@
-- v1alpha3
+- v1alpha4


site/data/apiVersions.yaml
```diff
@@ -1 +1 @@
-- v1alpha3
+- v1alpha4
```
neolit123, line 1: https://github.com/kubernetes-sigs/kind/blob/b6bc112522651d98c81823df56b7afa511459a3b/site/layouts/index.redirects#L1-L3

looks like both versions should be added here, not only one of them?

xinydev: I don’t know if we still need v1alpha3,  The [master branch](https://github.com/kubernetes-sigs/kind/tree/main/pkg/apis/config) only has v1alpha4. and the [latest version](https://pkg.go.dev/sigs.k8s.io/kind/pkg/apis/config/v1alpha3?utm_source=godoc) of v1alpha3 was published on May 1, 2020

neolit123: /lgtm

BenTheElder:
/ok-to-test
/approve
thanks

BenTheElder:
#2361

PR #2360, opened 2021-07-09
Refactor the pull function

xinydev:
Using `wait.ExponentialBackoff`   for retiring to pull the image 

I’m not sure why so many lines have been added to `go.sum`, I just ran `go mod tidy`

/kind cleanup


pkg/build/nodeimage/internal/container/docker/pull.go
```diff
@@ -19,25 +19,29 @@ package docker
 import (
 	"time"
 
+	"k8s.io/apimachinery/pkg/util/wait"
```
BenTheElder, line 22: We would prefer to actually drop this dependency not add more dependency on it.

Pulling in dependencies staged from Kubernetes makes kind harder to embed. This is the last remaining one of those, which we currently only use for two small things.

pkg/build/nodeimage/internal/container/docker/pull.go
```diff
@@ -19,25 +19,29 @@ package docker
...
-			if err == nil {
-				break
-			}
+	backoff := wait.Backoff{
+		Duration: time.Second,
+		Factor:   2,
+		Steps:    retries,
+	}
+	var lastErr error
+	err := wait.ExponentialBackoff(backoff, func() (bool, error) {
```
BenTheElder, line 35: I don't think exponential backoff makes sense here, we're expecting some small network flake, users do not want to wait exponential time for an image pull, we're just *slightly* optimizing vs a plain `docker run`.

pkg/build/nodeimage/internal/container/docker/pull.go
xinydev: Thanks for the explanation！

PR #2363, opened 2021-07-12
bump go to 1.16.6

BenTheElder:
main fix is for TLS which shouldn't affect the kind binary, but we stay up to date anyhow.


BenTheElder:
@aojea we should probably build a new kindnetd binary (no need to change the base image, just make sure it is freshly pulled).

aojea:
```
Cgroup v2 / Cgroup v2 (podman, rootless) (pull_request) Failing after 35m — Cgroup v2 (podman, rootless)
```
knonw issue

PR #2365, opened 2021-07-13
[WIP] build kind images automatically

aojea:
let the machines work for us

### DIFF
--- a/.github/workflows/imagebuilder.yaml
+++ b/.github/workflows/imagebuilder.yaml
@@ -1,0 +1,32 @@
+name: Build images
+
+on:
+  workflow_dispatch:
+
+jobs:
+  build-kindnetd:
+    name: Kindnet image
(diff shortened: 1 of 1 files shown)


aojea:
@BenTheElder this can be triggered by an event or manually (right now is set to only manual), 
we can also push to an intermediate registry if we don't want to publish it to dockerhub directly
You have a run example here https://github.com/aojea/kind/runs/3055037619?check_suite_focus=true
thoughts?

BenTheElder:
> we can also push to an intermediate registry if we don't want to publish it to dockerhub directly


https://github.com/kubernetes/k8s.io/tree/main/k8s.gcr.io

https://github.com/kubernetes/test-infra/tree/master/images/builder

aojea:
I don't have clear how substitutions work @BenTheElder 
https://github.com/kubernetes/test-infra/pull/22986

PR #2368, opened 2021-07-14
Use main branch as the stable branch in Ingress demo

rikatz:
We've renamed master to main (actually created a new one, over the stable branch and not the development branch).

So now we can use again the file present directly in main branch, assuming main is going to be the stable and that as soon as we release Ingress v1 that branch is gonna get merged into this one

### DIFF
--- a/site/content/docs/user/ingress.md
+++ b/site/content/docs/user/ingress.md
@@ -116,8 +116,7 @@ Additional information about Contour can be found at: [projectcontour.io](https:
 ### Ingress NGINX
 
 {{< codeFromInline lang="bash" >}}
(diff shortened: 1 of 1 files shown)


neolit123 (commented):
/lgtm

aojea:
> So now we can use again the file present directly in main branch, assuming main is going to be the stable and that as soon as we release Ingress v1 that branch is gonna get merged into this one

I can't follow, once you release v1 do we need to change again this file?

rikatz:
Nope, we are going to use the main branch as the stable once we release v1.

So I'm going to merge the v1 branch into main branch, and we can add a note to people using k8s < v1.19 to point to the deprecated branch (which is called release-v1beta1 but we are going to find a better naming)

aojea:
Thanks Ricardo for the hard work

PR #2372, opened 2021-07-15
[WIP] Switch from Alpine to Ubuntu based haproxy image

dims:
should we get off of alpine? looks like there's a ubuntu based image we could go to

(since https://github.com/kubernetes/enhancements/blob/master/keps/sig-release/1729-rebase-images-to-distroless/README.md is preaching for debian-base and we don't have one of those)

### DIFF
--- a/images/haproxy/Dockerfile
+++ b/images/haproxy/Dockerfile
@@ -13,6 +13,6 @@
 # limitations under the License.
 
 # standard haproxy image + minimal config so the container will not exit
-ARG BASE="haproxy:2.2.0-alpine"
+ARG BASE="haproxy:2.2.0"
 FROM ${BASE}
 COPY haproxy.cfg /usr/local/etc/haproxy/haproxy.cfg


images/haproxy/Dockerfile
```diff
@@ -13,6 +13,6 @@
 # limitations under the License.
 
 # standard haproxy image + minimal config so the container will not exit
-ARG BASE="haproxy:2.2.0-alpine"
```
BenTheElder, line 16: what's the size difference?
we just need an image with haproxy IIRC (and maybe `cat` ?). we could also do some COPY / multi-stage onto distroless or something

dims:
```
haproxy         2.2.0                d27dbbd0e8a9   11 months ago   93.3MB
haproxy         2.2.0-alpine         0b8abcf7bfdb   12 months ago   24.5MB
```

BenTheElder: slightly unfortunate, but this image is also the one we're least sensitive to.

I think we really just need:
- haproxy
- kill
- cat

It's too bad there isn't a base image closer to just haproxy.

dims: @BenTheElder so start with the ubuntu image and COPY to distroless with those 3 binaries? is there a optional presubmit CI job that would be useful in testing a change this?

BenTheElder: I don't think we have one currently, there's not been much demand for the HA functionality.

Issue #2326, opened 2021-06-24
Kind clusters using ipvs kubeProxyMode doesn't work On v1.21.1


mnencia:
**What happened**:

Clusters using image `kindest/node:v1.21.1` cannot resolve DNS names.
It works, for example, with `kindest/node:v1.20.7`

**What you expected to happen**:

DNS should work

**How to reproduce it (as minimally and precisely as possible)**:

```
mnencia@sirio  ➤  kind create cluster --image=kindest/node:v1.21.1
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.21.1) 🖼
 ✓ Preparing nodes 📦
 ✓ Writing configuration 📜
 ✓ Starting control-plane 🕹️
 ✓ Installing CNI 🔌
 ✓ Installing StorageClass 💾
Set kubectl context to "kind-kind"
You can now use your cluster with:

kubectl cluster-info --context kind-kind

Have a nice day! 👋
mnencia@sirio  ➤  kubectl run -i -t busybox --image=busybox --restart=Never --rm
If you don't see a command prompt, try pressing enter.
/ # ping wwww.google.com
ping: bad address 'wwww.google.com'
/ #
```

**Anything else we need to know?**:

As a workaround, you can patch the CoreDNS configuration and restart the CoreDNS deployment

```diff
--- coredns.yaml.orig	2021-06-24 11:40:42.000000000 +0200
+++ coredns.yaml	2021-06-24 11:41:00.000000000 +0200
@@ -1,30 +1,30 @@
 apiVersion: v1
 data:
   Corefile: |
     .:53 {
         errors
         health {
            lameduck 5s
         }
         ready
         kubernetes cluster.local in-addr.arpa ip6.arpa {
            pods insecure
            fallthrough in-addr.arpa ip6.arpa
            ttl 30
         }
         prometheus :9153
-        forward . /etc/resolv.conf {
+        forward . dns://1.1.1.1 {
            max_concurrent 1000
         }
         cache 30
         loop
         reload
         loadbalance
     }
 kind: ConfigMap
 metadata:
   creationTimestamp: "2021-06-24T09:28:09Z"
   name: coredns
   namespace: kube-system
   resourceVersion: "238"
   uid: c2def70f-5890-41cb-8184-b4dd33e8cece
```

```
mnencia@sirio  ➤  kubectl apply -f coredns.yaml
[...]

PR #2375, opened 2021-07-18
enable sysctl route_localnet

aojea:
Pods dns traffic is rewritten to localhost, hence we need to
enable route_localnet option.

docker for custom networks installs a custom dns server on the ip 172.17.0.11 and adds some iptables rules to redirect the traffic.
```
$ docker run -it ubuntu bash --network kind
root@e33b86fa05dd:/# iptables-save
# Generated by iptables-save v1.8.4 on Sun Jul 18 22:57:49 2021
*filter
:INPUT ACCEPT [1542:19090817]
:FORWARD ACCEPT [0:0]
:OUTPUT ACCEPT [1416:83722]
COMMIT
# Completed on Sun Jul 18 22:57:49 2021
# Generated by iptables-save v1.8.4 on Sun Jul 18 22:57:49 2021
*nat
:PREROUTING ACCEPT [19:1715]
:INPUT ACCEPT [0:0]
:OUTPUT ACCEPT [5:300]
:POSTROUTING ACCEPT [10:677]
:DOCKER_OUTPUT - [0:0]
:DOCKER_POSTROUTING - [0:0]
-A OUTPUT -d 127.0.0.11/32 -j DOCKER_OUTPUT
-A POSTROUTING -d 127.0.0.11/32 -j DOCKER_POSTROUTING
-A DOCKER_OUTPUT -d 127.0.0.11/32 -p tcp -m tcp --dport 53 -j DNAT --to-destination 127.0.0.11:39405
-A DOCKER_OUTPUT -d 127.0.0.11/32 -p udp -m udp --dport 53 -j DNAT --to-destination 127.0.0.11:44107
-A DOCKER_POSTROUTING -s 127.0.0.11/32 -p tcp -m tcp --sport 39405 -j SNAT --to-source :53
-A DOCKER_POSTROUTING -s 127.0.0.11/32 -p udp -m udp --sport 44107 -j SNAT --to-source :53
COMMIT

root@e33b86fa05dd:/# sysctl -a | grep localnet
net.ipv4.conf.all.route_localnet = 0
net.ipv4.conf.default.route_localnet = 0
net.ipv4.conf.eth0.route_localnet = 0
net.ipv4.conf.lo.route_localnet = 0

root@e33b86fa05dd:/# cat /etc/resolv.conf 
nameserver 127.0.0.11
nameserver 2001:4860:4860::8888
nameserver 2001:4860:4860::8844
```

Kubernetes can't use a localhost address as resolv.conf so kind rewrites the docker dns address by the gateway of the kind network, and uses it as nameserver on the resolv.conf
```
root@kind-control-plane:/# iptables-save | grep DOCKER
:DOCKER_OUTPUT - [0:0]
:DOCKER_POSTROUTING - [0:0]
-A PREROUTING -d 172.18.0.1/32 -j DOCKER_OUTPUT
-A OUTPUT -d 172.18.0.1/32 -j DOCKER_OUTPUT
-A POSTROUTING -d 172.18.0.1/32 -j DOCKER_POSTROUTING
-A DOCKER_OUTPUT -d 172.18.0.1/32 -p tcp -m tcp --dport 53 -j DNAT --to-destination 127.0.0.11:40457
-A DOCKER_OUTPUT -d 172.18.0.1/32 -p udp -m udp --dport 53 -j DNAT --to-destination 127.0.0.11:51490
-A DOCKER_POSTROUTING -s 127.0.0.11/32 -p tcp -m tcp --sport 40457 -j SNAT --to-source 172.18.0.1:53
-A DOCKER_POSTROUTING -s 127.0.0.11/32 -p udp -m udp --sport 51490 -j SNAT --to-source 172.18.0.1:53
root@kind-control-plane:/# cat /etc/resolv.conf
nameserver 172.18.0.1
nameserver 2001:4860:4860::8888
nameserver 2001:4860:4860::8844

```

then the traffic comes from a pod, it has to DNAT to a localhost, hence route_localnet is needed so pods can access the DNS

Fixes: #2326

### DIFF
--- a/images/base/files/etc/sysctl.d/10-network-magic.conf
+++ b/images/base/files/etc/sysctl.d/10-network-magic.conf
@@ -1,0 +1,11 @@
+# Do not consider loopback addresses as martian source or destination while routing
+# - Docker with custom networks uses an embedded DNS server with address 172.0.0.11
+# - Kubernetes pods mount the node resolv.conf, so they can't use a loopack address
+# that is only reachable from the node.
+#
+# KIND rewrites the well-known docker DNS address 127.0.0.11 by a non-loopback address.
+# The DNS traffic coming from a pod will have to route to a localhost address to be NATed,
+# hence we have to enable route_localnet or pods DNS will not work.
+# Kubernetes mitigates the possible security issue caused by enabling this option.
+# ref: https://nvd.nist.gov/vuln/detail/CVE-2020-8558
+net.ipv4.conf.all.route_localnet=1
--- a/pkg/build/nodeimage/defaults.go
+++ b/pkg/build/nodeimage/defaults.go
@@ -20,4 +20,4 @@ package nodeimage
 const DefaultImage = "kindest/node:latest"
 
 // DefaultBaseImage is the default base image used
-const DefaultBaseImage = "docker.io/kindest/base:v20210729-fb821a4c"
+const DefaultBaseImage = "docker.io/kindest/base:v20210729-302b42d2"


BenTheElder:
I'm not observing the linked issue with iptables backend. Further wasn't this sysctl disabled upstream for security reasons?

aojea:
> I'm not observing the linked issue with iptables backend. 

:thinking: I observe it with iptables if I set `net.ipv4.conf.all.route_localnet = 0` , bear in mind is dns resolution from a "networking" pod, hostNetwork pods and node resolution works.

> Further wasn't this sysctl disabled upstream for security reasons?

yeah, but I would consider it very low, ever more for kind clusters ... actually I think that the fix implemented doesn't solve the supposed security problem https://github.com/kubernetes/kubernetes/pull/91666#issuecomment-849416083

BenTheElder:
I think we should consider setting the flag in IPVS mode then instead of baking it into the image. This also fixes more usage?

BenTheElder:
> is a problem of how kind implements custom dns, if people use other proxies and don't set the flag, pods will not be able to reach dns too

not really concerned about other proxies, we're maintaining compatibility with kube-proxy specifically, for other proxies *nothing* has changed.

IIRC kube-proxy had a new flag when this change was made in IPVS, to preserve the old behavior.

aojea:
> IIRC kube-proxy had a new flag when this change was made in IPVS, to preserve the old behavior.

no, ipvs never required that flag, that is why it doesn't need to keep the behavior 

> not really concerned about other proxies, we're maintaining compatibility with kube-proxy specifically, for other proxies nothing has changed.

the way KIND implements dns requires net.ipv4.conf.all.route_localnet = 1, it happens that kube-proxy iptables requires it too, but there is nothing about compatibility here, is just more  a fortunate coincidence

images/base/files/etc/sysctl.d/10-network-magic.conf
```diff
@@ -0,0 +1,7 @@
+# Do not consider loopback addresses as martian source or destination while routing
+# Since pods mount the node resolv.conf, they can't use a loopack address that is only
+# reachable from the node, so KIND rewrites the well-known docker DNS address 127.0.0.11 
+# by a non-loopback address.
+# The DNS traffic coming from a pod will have to route to a localhost address to be NATed
+# hence we have to enable route_localnet or pods DNS will not work.
+net.ipv4.conf.all.route_localnet=1
```
AkihiroSuda, line 7: Seems to cause https://nvd.nist.gov/vuln/detail/CVE-2020-8558

Maybe fine for `kind` usecases, but I think at least the CVE should be explained here.

aojea: heh, thanks for the comment
I forgot that kubelet install iptables rules to mitigate the issue, so this doesn't suppose a risk for kind
cc @BenTheElder

aojea:
@AkihiroSuda does this break rootless in some way? I remember now that you had to add some fake sysctls

images/base/files/etc/sysctl.d/10-network-magic.conf
BenTheElder: I don't think this is fully mitigated https://github.com/kubernetes/kubernetes/issues/92315#issuecomment-656506304

aojea: I disagree that is a real issue https://github.com/kubernetes/kubernetes/pull/91666#issuecomment-849416083

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/2375#issuecomment-882745180

a flag was discussed for ipvs but AFAICT never happened. so I guess this is the best we can do for now.

I dislike that this applies to even non-kube-proxy users and kube-proxy users not using iptables or ipvs (userspace?) but
1) we don't _really_ support not using kube-proxy
2) pure userspace mode is a bit LOL at this point and going away https://github.com/kubernetes/kubernetes/issues/103860

> the userspace proxy is essentially basically an untested easter egg

BenTheElder (approved):
/lgtm
/approve

PR #2380, opened 2021-07-22
Add doc for optional parameters on extraMounts

sestegra:
After diving into the source code, I updated the extraMounts example to show all parameters.

### DIFF
--- a/site/static/examples/config-with-mounts.yaml
+++ b/site/static/examples/config-with-mounts.yaml
@@ -6,3 +6,13 @@ nodes:
   extraMounts:
   - hostPath: /path/to/my/files/
     containerPath: /files
+    # optional: if set, the mount is read-only.
+    # false is the current default
+    readOnly: true
+    # optional: if set, the mount needs SELinux relabeling.
+    # false is the current default
+    selinuxRelabel: false
+    # optional: set propagation mode (None, HostToContainer or Bidirectional)
+    # see https://kubernetes.io/docs/concepts/storage/volumes/#mount-propagation
+    # None is the current default
+    propagation: HostToContainer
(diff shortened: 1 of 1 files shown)


aojea:
I think we should match the API docs
https://pkg.go.dev/sigs.k8s.io/kind@v0.11.1/pkg/apis/config/v1alpha4#Mount

sestegra:
As a user of this documentation, I would see this documentation as an end-user reference guide.
End-user probably doesn't understand a GO struct.

aojea:
> As a user of this documentation, I would see this documentation as an end-user reference guide.
> End-user probably doesn't understand a GO struct.

I mean the text on the comments :smile:

aojea:
> > As a user of this documentation, I would see this documentation as an end-user reference guide.
> > End-user probably doesn't understand a GO struct.
> 
> I mean the text on the comments

oh, they already match , sorry
/ok-to-test
/lgtm

site/static/examples/config-with-mounts.yaml
```diff
@@ -5,4 +5,14 @@ nodes:
   # add a mount from /path/to/my/files on the host to /files on the node
   extraMounts:
   - hostPath: /path/to/my/files/
-    containerPath: /files
\ No newline at end of file
+    containerPath: /files
+    # optional: if set, the mount is read-only.
+    # false is the current default
```
aojea, line 10: does this reads better this way?
```suggestion
    # default false
```

sestegra: Update done in last commit

aojea:
/lgtm
/approve
Thanks

Issue #2352, opened 2021-07-05
patched containerd config cannot take effect as the containerd restart operation is skipped


tokers:
**What happened**:

I have some custom configurations for the container to mirror some image registries. But the patched config was not taken effect.

**What you expected to happen**:

The patched config should take effect.

**How to reproduce it (as minimally and precisely as possible)**:

1. prepare the config:

```
# test.yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
- role: worker
- role: worker
- role: worker
containerdConfigPatches:
- |-
  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."localhost:5000"]
    endpoint = ["https://docker.io"]
```

2. create the cluster:

```
kind create cluster --config test.yaml
```

3. enter a container and try to pull an image

```
# crictl pull localhost:5000/nginx
FATA[0000] pulling image: rpc error: code = Unknown desc = failed to pull and unpack image "localhost:5000/nginx:latest": failed to resolve reference "localhost:5000/nginx:latest": failed to do request: Head "http://localhost:5000/v2/nginx/manifests/latest": dial tcp [::1]:5000: connect: connection refused
```

**Anything else we need to know?**:

* Croup version: v2
* docker is run in rootless mode

After some inspections, I found the containerd restart is conditional, it firstly checks the running status of systems.

```
if err := node.Command("bash", "-c", `! systemctl is-system-running || systemctl restart containerd`).Run(); err != nil {
	return errors.Wrap(err, "failed to restart containerd after patching config")
}
```

https://github.com/kubernetes-sigs/kind/blob/main/pkg/cluster/internal/create/actions/config/config.go#L189-L191

Indeed there are two failed units:

```
# systemctl --failed
  UNIT                     LOAD   ACTIVE SUB    DESCRIPTION
● sys-kernel-debug.mount   loaded failed failed Kernel Debug File System
● sys-kernel-tracing.mount loaded failed failed Kernel Trace File System

LOAD   = Reflects whether the unit definition was properly loaded.
ACTIVE = The high-level unit activation state, i.e. generalization of SUB.
SUB    = The low-level unit activation state, values depend on unit type.
2 loaded units listed.
```

So I think the check logic for the status of systemd can be more precise, for instance, we may be permissive for the `degrade`, `running`.

**Environment:**

- kind version: (use `kind version`): kind v0.11.1 go1.16.4 linux/amd64
[...]

PR #2382, opened 2021-07-22
check if containerd is running not if system is running

BenTheElder:
fixes: #2262

### DIFF
--- a/pkg/cluster/internal/create/actions/config/config.go
+++ b/pkg/cluster/internal/create/actions/config/config.go
@@ -185,8 +185,8 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 					return errors.Wrap(err, "failed to write patched containerd config")
 				}
 				// restart containerd now that we've re-configured it
-				// skip if the systemd (also the containerd) is not running
-				if err := node.Command("bash", "-c", `! systemctl is-system-running || systemctl restart containerd`).Run(); err != nil {
+				// skip if containerd is not running
+				if err := node.Command("bash", "-c", `! pgrep --exact containerd || systemctl restart containerd`).Run(); err != nil {
 					return errors.Wrap(err, "failed to restart containerd after patching config")
 				}
 				return nil


pkg/cluster/internal/create/actions/config/config.go
```diff
@@ -185,8 +185,8 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 					return errors.Wrap(err, "failed to write patched containerd config")
 				}
 				// restart containerd now that we've re-configured it
-				// skip if the systemd (also the containerd) is not running
-				if err := node.Command("bash", "-c", `! systemctl is-system-running || systemctl restart containerd`).Run(); err != nil {
+				// skip if containerd is not running
+				if err := node.Command("bash", "-c", `! pgrep --exact containerd || systemctl restart containerd`).Run(); err != nil {
```
BenTheElder, line 189: `pgrep` returns `0` on a match and `1` on no match.

```
EXIT STATUS
0      One or more processes matched the criteria. For pkill and
              pidwait, one or more processes must also have been
              successfully signalled or waited for.
1      No processes matched or none of them could be signalled.
2      Syntax error in the command line.
3      Fatal error: out of memory etc.
```
https://man7.org/linux/man-pages/man1/pgrep.1.html

tao12345666333:
LGTM！

It seems more appropriate to check the process directly

fixes #2352

pkg/cluster/internal/create/actions/config/config.go
BenTheElder: containerd when running has a process tree rooted by `containerd`, the child processes are the shim which is `containerd-shim`. `pgrep --exact [pattern]` matches only an exact pattern match.

aojea: Why not pkill -HUP then?

cpanato (commented):
Nice

LGTM

pkg/cluster/internal/create/actions/config/config.go
BenTheElder: This shell command we crafted exits success when successful in any case and only non zero when something doesn't work. We can correctly distinguish success.

aojea:
```suggestion
				if err := node.Command("bash", "-c", `pgrep --exact containerd || systemctl restart containerd`).Run(); err != nil {
```

aojea: we want to restart containerd if the pgrep fails, right? we don't need to negate then

aojea:
```
$ bash -c '! pgrep --exact containerd || echo "notfound"'
490904
notfound
$ bash -c 'pgrep --exact containerd || echo "notfound"
490904
$ bash -c 'pgrep --exact container || echo "notfound"'
notfound
```

BenTheElder: If the pgrep fails there is no containerd process found, so what are we restarting? 🙃

aojea: lol, what I was thinking ...

aojea:
/lgtm
/approve
doooooh

PR #2387, opened 2021-07-25
Add ability to specify log format for componentManager and apiServer.

pchan:
This PR adds the ability to specify the log format for control-plane components, componentManager and apiServer. 

Reference: [kubernetes/kubernetes#100748 ](https://github.com/kubernetes/kubernetes/issues/100748)

### DIFF
--- a/hack/ci/e2e-k8s.sh
+++ b/hack/ci/e2e-k8s.sh
@@ -90,27 +90,16 @@ create_cluster() {
   fi
 
   scheduler_extra_args="      \"v\": \"${KIND_CLUSTER_LOG_LEVEL}\""
-  SCHEDULER_LOG_FORMAT=${SCHEDULER_LOG_FORMAT:-$CLUSTER_LOG_FORMAT}
-  if [ -n "$SCHEDULER_LOG_FORMAT" ]; then
+  controllerManager_extra_args="      \"v\": \"${KIND_CLUSTER_LOG_LEVEL}\""
+  apiServer_extra_args="      \"v\": \"${KIND_CLUSTER_LOG_LEVEL}\""
+  if [ -n "$CLUSTER_LOG_FORMAT" ]; then
       check_log_format
       scheduler_extra_args="${scheduler_extra_args}
-      \"logging-format\": \"${SCHEDULER_LOG_FORMAT}\""
-  fi
-
-  controllerManager_extra_args="      \"v\": \"${KIND_CLUSTER_LOG_LEVEL}\""
-  CONTROLLERMANAGER_LOG_FORMAT=${CONTROLLERMANAGER_LOG_FORMAT:-$CLUSTER_LOG_FORMAT}
-  if [ -n "$CONTROLLERMANAGER_LOG_FORMAT" ]; then
-      check_log_format
+      \"logging-format\": \"${CLUSTER_LOG_FORMAT}\""
       controllerManager_extra_args="${controllerManager_extra_args}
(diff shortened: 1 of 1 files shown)


BenTheElder:
Please see the discussion on the last related PR: I'd like to wind down the per component options in this script to keep the "API surface" low. We started with kubelet but are moving towards one option to turn on structured logs for the structured logs CI.

At some point we will move this to a command instead of bash and having less env interpolated as config strings to support will make that a smoother transition.

serathius:
@pchan, from Bens comment I take that we should minimize number of *_LOG_FORMAT envs we use. We need CLUSTER_LOG_FORMAT as the main way for configuration and KUBELET_LOG_FORMAT for backward compatibility as it's already used by CI scripts. Let's remove the rest

pchan:
> @pchan, from Bens comment I take that we should minimize number of *_LOG_FORMAT envs we use. We need CLUSTER_LOG_FORMAT as the main way for configuration and KUBELET_LOG_FORMAT for backward compatibility as it's already used by CI scripts. Let's remove the rest

Done.

pchan:
> At some point we will move this to a command instead of bash and having less env interpolated as config strings to support will make that a smoother transition.

Do you mean a kind command or something else ? If the command is in kind, I would think the the candidate will look something like this ?

`
  kind create cluster 
....                                                                                                                                                                                                                                                                                                                                                                         
   "--log=apiServer:json,scheduler:text"
`

This probably deserves to be on slack. But I was curious to see the direction for the long-term solution.

serathius:
/lgtm
ping @BenTheElder

BenTheElder:
> Do you mean a kind command or something else ? If the command is in kind, I would think the the candidate will look something like this ?


I mean a command that replaces the shell script, like https://github.com/kubernetes-sigs/kind/pull/1269

hack/ci/e2e-k8s.sh
```diff
@@ -62,6 +62,15 @@ build() {
   make all WHAT='cmd/kubectl test/e2e/e2e.test vendor/github.com/onsi/ginkgo/ginkgo'
 }
 
+check_log_format() {
```
BenTheElder, line 65: nit: this method name doesn't seem very clear, it's only checking the Kubernetes version and erroring and only used for KUBELET_LOG_FORMAT

hack/ci/e2e-k8s.sh
```diff
@@ -73,33 +82,24 @@ create_cluster() {
...
-     v1.1[0-8].*)
-      echo "SCHEDULER_LOG_FORMAT is only supported on versions >= v1.19, got ${KUBE_VERSION}"
-      exit 1
-      ;;
-    *)
-      # NOTE: the indendation on the next line is meaningful!
+  controllerManager_extra_args="      \"v\": \"${KIND_CLUSTER_LOG_LEVEL}\""
+  apiServer_extra_args="      \"v\": \"${KIND_CLUSTER_LOG_LEVEL}\""
+  if [ -n "$CLUSTER_LOG_FORMAT" ]; then
+      check_log_format
```
BenTheElder, line 96: this shouldn't be throwing an error message about KUBELET_LOG_FORMAT

hack/ci/e2e-k8s.sh
pchan: I have renamed it as check_structured_log_support to better reflect the operation.

hack/ci/e2e-k8s.sh
pchan: Fixed in the latest commit.

PR #2388, opened 2021-07-25
podman rootless doc

bagnaram:
Added `ip6_table` kernel module for rootless Podman.

Formatting and grammar fixes


site/content/docs/user/rootless.md
```diff
@@ -31,15 +32,17 @@ Delegate=yes
...
 
 ## Restrictions
 
 The restrictions of Rootless Docker apply to kind clusters as well.
 
 e.g.
-- OverlayFS cannot be used unless the host is using kernel >= 5.11, or Ubuntu/Debian kernel
-- Cannot mount block storages
+- OverlayFS cannot be used unless the host is using kernel >= 5.11, or
+  Ubuntu/Debian kernel
```
BenTheElder, line 44: What is the purpose of splitting this line?
Unless we're going to require a particular auto formatter I'd rather not endlessly reformat lines like this arbitrarily. This doesn't seem to have any content change.

BenTheElder:
This page should really be renamed to reflect it's just generally rootless not docker or podman specific.

site/content/docs/user/rootless.md
bagnaram: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md#line-length

bagnaram (commented):
Page has been renamed to address both runtimes.

site/content/docs/user/rootless.md
BenTheElder: That style guide is specifically for that repo owned by SIG Contributor Experience and is not agreed upon by the rest of the project 🙃 

> This style guide is for content in the Kubernetes github community repository. It is an extension of the Kubernetes documentation style-guide.

I would really prefer to not have this re-formatted by competing settings. If any auto-formatting is to be used here it should be enforced by a linter in presubmit.

https://kubernetes.io/docs/contribute/style/style-guide/ is the guide used for kubernetes.io, which is a closer analogue of site/ here.

site/content/docs/user/rootless.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: "Running kind with Rootless Docker"
+title: "Running kind with Rootless Docker or Podman"
```
BenTheElder, line 2: Let's just call this "Rootless" similar to "Ingress" "Local Registry" etc., this is the only entry > 2 words.

PR #2390, opened 2021-07-27
Added WSL2 workaround for kubelet

### LINKED ISSUE #2323: WSL2 ERROR: failed to create cluster
**ERROR: failed to create cluster**:
ERROR: failed to create cluster: failed to init node with kubeadm: command "docker exec --privileged kind-control-plane kubeadm init --skip-phases=preflight --config=/kind/kubeadm.conf --skip-token-print --v=6" failed with error: exit status 1
Command Output: I0622 15:16:13.468494     216 initconfiguration.go:246] loading configuration from "/kind/kubeadm.conf"		

**What is Expected**:
cluster should be created without any error.

**How to reproduce it**:
run below command to reproduce it :
$ kind create cluster

I have recently install ubuntu as virtual machine on windows 10 as **wsl 2.**
running ubuntu on windows terminal as admin user, also installed docker and set it as non root user.
below I am providing environment related information.

**ENVIRONMENT:**

**Ubuntu**
command used `$lsb_release -a`

No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.2 LTS
Release:        20.04
Codename:       focal

**Kubectl Installation**
https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/#install-using-native-package-management

**Kubectl Version** 
command used `kubectl version --client`

Client Version: 
	version.Info{
		Major:"1", Minor:"21", 
		GitVersion:"v1.21.2", 
		GitCommit:"092fbfbf53427de67cac1e9fa54aaa09a28371d7", 
		GitTreeState:"clean", 
		BuildDate:"2021-06-16T12:59:11Z", 
		GoVersion:"go1.16.5", 
		Compiler:"gc", 
		Platform:"linux/amd64"
		}

**kind Installtion :** 

curl -Lo ./kind https://kind.sigs.k8s.io/dl/v0.11.1/kind-linux-amd64
chmod +x ./kind
sudo mv ./kind /usr/local/bin/kind

**kind version: :**
command used `$kind version`

kind v0.11.1 go1.16.4 linux/amd64

**docker info**

command used `$docker info`

Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Build with BuildKit (Docker Inc., v0.5.1-docker)
  scan: Docker Scan (Docker Inc., v0.8.0)

Server:
 Containers: 1
  Running: 0
  Paused: 0
  Stopped: 1
 Images: 2
 Server Version: 20.10.7
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
 Swarm: inactive
 Runtimes: runc io.containerd.runc.v2 io.containerd.runtime.v1.linux
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: d71fcd7d8303cbf684402823e425e9dd2e99285d
 runc version: b9ee9c6314599f1b4a7f497e1f1f856fe433d3b7
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
 Kernel Version: 5.4.72-microsoft-standard-WSL2
 Operating System: Ubuntu 20.04.2 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 12
 Total Memory: 6.133GiB
 Name: LAPTOP-TN6NO0LS
 ID: JDCK:NRQ2:ML5P:EUMK:OBYG:76PM:5SXD:FMYK:KHCX:NDTB:IQ4R:KIBJ
 Debug Mode: false
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false

WARNING: No blkio throttle.read_bps_device support
WARNING: No blkio throttle.write_bps_device support
WARNING: No blkio throttle.read_iops_device support
WARNING: No blkio throttle.write_iops_device support

networkop:
This PR fixes #2323.


networkop:
signed the CLA

images/base/files/usr/local/bin/entrypoint
```diff
@@ -204,6 +204,10 @@ fix_cgroup() {
   # See: https://man7.org/linux/man-pages/man7/cgroups.7.html
   local current_cgroup
   current_cgroup=$(grep -E '^[^:]*:([^:]*,)?cpu(,[^,:]*)?:.*' /proc/self/cgroup | cut -d: -f3)
+  # on WSL2 systems /sys/fs/cgroup/systemd dir is missing which eventually leads to kubelet failing to start.
```
BenTheElder, line 207: Is it WSL2 or just distros without systemd?
Are there WSL2 distros with systemd?

images/base/files/usr/local/bin/entrypoint
```diff
@@ -204,6 +204,10 @@ fix_cgroup() {
...
+  # see: https://github.com/kubernetes-sigs/kind/issues/2323
+  # the following mkdir command is a common workaround, see: https://github.com/microsoft/WSL/issues/4189#issuecomment-758439957
+  mkdir -p /sys/fs/cgroup/systemd/
```
BenTheElder, line 210: Technically I don't think this (cgroup controller heirarchy) has to be mounted at `/sys/fs/cgroup` though I've not ever encountered it mounted elsewhere.

see: https://man7.org/linux/man-pages/man7/cgroups.7.html

This directory would also normally be configured like `mount -t cgroup -o none,name=systemd cgroup /sys/fs/cgroup/systemd`

BenTheElder: We actually already depend on it being under `/sys/fs/cgroup` below though, regarding the first point.

images/base/files/usr/local/bin/entrypoint
networkop: At this stage I'm only certain about WSL2. I can reword this to say  "and potentially other distros without systemd"?

images/base/files/usr/local/bin/entrypoint
BenTheElder: ... And it's not just kind assuming this in the ecosystem https://kubernetes.slack.com/archives/C0BP8PW9G/p1627412176214900

images/base/files/usr/local/bin/entrypoint
BenTheElder: SGTM

images/base/files/usr/local/bin/entrypoint
networkop: awesome, that's good to know.

networkop: as for the mounting, I believe this will be done by [this line](https://github.com/kubernetes-sigs/kind/pull/2390/files/12a8deeb38f5f11074cdc558b2a7f809a5242c67#diff-3c55751d83af635109cece495ee2ff38206764a8b95f4cb8f11fc08a5c0ea8dcR171) 
The current placement of this mkdir command is deliberate so that it gets picked up by the `cgroup_subsystems` on the next line and then passed to the `mount_kubelet_cgroup_root "/kubelet" "${subsystem}"`

At least, this is how it works for me. I've added just this one line and this allowed for the cluster to start up and results in the following mount being created by the entrypoint 
```
root@k8s-guide-control-plane:/# mount | grep kubelet | head -n 1
cgroup on /sys/fs/cgroup/cpuset/kubelet type cgroup (rw,nosuid,nodev,noexec,relatime,cpuset)
```

networkop:
something's not right, investigating

networkop:
I think I've finally found a stable workaround which works across reboots and docker restarts. 
The reason why it didn't quite work previously was that any modifications made by the entrypoint (e.g. `mkdir -p /sys/fs/cgroup/systemd/kubelet`) were getting wiped out when the systemd started. Mounting to self like it's done [here](https://github.com/kubernetes-sigs/kind/blob/12a8deeb38f5f11074cdc558b2a7f809a5242c67/images/base/files/usr/local/bin/entrypoint#L171) didn't work either.

The conclusion is that any changes done by the entrypoint to any path under `/sys/fs/cgroup/systemd` would not survive `/sbin/init` on WSL2. 

So the alternative is to add an extra `ExecStartPre` to kubelet's drop-in unit and create `/sys/fs/cgroup/systemd/kubelet`  when kubelet starts.

aojea:
/ok-to-test
/lgtm
Thanks

stealthybox:
I was able to reproduce the failed log message on WSL2.
I can confirm running the ExecStartPre during cluster creation fixes my problem:
```shell
docker exec -it kind-control-plane /bin/sh -euc "if [ ! -f /sys/fs/cgroup/cgroup.controllers ] && [ ! -d /sys/fs/cgroup/systemd/kubelet ]; then mkdir -p /sys/fs/cgroup/systemd/kubelet; fi"
```
/lgtm

BenTheElder:
I'm taking a break from some other things I've been on and doing some catchup today, https://github.com/kubernetes-sigs/kind/pull/2408 will go in first as it's an earlier PR

powerman:
Can someone test it using environment from #2440, to make sure it actually fixes it? Or tell me how to test it myself…

PR #2391, opened 2021-07-27
Update node image build source reference

stmcginnis:
Change 950a31918721daa3c636c91e2dd706b5b1a701c3 renamed `pkg/build/nodeimage/build_impl.go` to `buildcontext.go`. The README doc for node images had a link to the old doc that was missed. This updates the README to point to the current source file name for reference.

### DIFF
--- a/images/node/README.md
+++ b/images/node/README.md
@@ -1,7 +1,7 @@
 ## images/node
 
 See: [`pkg/build/nodeimage/build.go`][pkg/build/nodeimage/build.go] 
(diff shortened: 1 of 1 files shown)


BenTheElder (approved):

BenTheElder:
This readme does also note that it's based on the initial design and *not* kept up to date currently, but I think it's reasonable to update this link. We should consider a permalink to an earlier implementation in the future.

Thanks for the PR!

PR #2393, opened 2021-07-28
change http to https websites

vdjagilev:
Some links are http.

Notes:

* `http://slack.k8s.io/` does not redirect to https
* Not sure about license file which has links with http schema
* SVG's also have http for dtd


vdjagilev:
cla signed

BenTheElder:
The License file should be verbatim from the original copy without alteration.

vdjagilev:
> The License file should be verbatim from the original copy without alteration.

@BenTheElder 

I see. No change there then.

BenTheElder (approved):
/lgtm
/approve
/ok-to-test

PR #2408, opened 2021-08-08
rootless: replace procfs workaround with KubeletInUserNamespace feature gate

AkihiroSuda:
KubeletInUserNamespace feature gate was merged into Kubernetes v1.22, so we can drop the workaround to fake procfs.

See:
- https://github.com/kubernetes/kubernetes/pull/92863/commits
- https://kubernetes.io/docs/tasks/administer-cluster/kubelet-in-userns/

- - -
- Base: `docker.io/akihirosuda/tmp-kind-base:g34e9a52a`
- Node: `docker.io/akihirosuda/tmp-kind-node:g34e9a52a-v1.22.0`


neolit123 (commented):
/lgtm

Issue #2409, opened 2021-08-08
CI failing for cgroup2+docker+rootless


AkihiroSuda:
CI is recently failing for cgroup2+docker+rootless.

CI isn't failing for cgroup2+docker+rootful and cgroup2+podman+rootless.

Also the failure is not reproducible on my local Vagrant.

https://github.com/kubernetes-sigs/kind/runs/3255789909

`kind-logs-cgroup2-docker-rootless-1102670304.zip:kind-control-plane/kubelet.log`

```
-- Journal begins at Thu 2021-08-05 19:38:30 UTC, ends at Thu 2021-08-05 19:40:32 UTC. --
Aug 05 19:38:31 kind-control-plane systemd[1]: Condition check resulted in kubelet: The Kubernetes Node Agent being skipped.
Aug 05 19:38:36 kind-control-plane systemd[1]: Starting kubelet: The Kubernetes Node Agent...
Aug 05 19:38:36 kind-control-plane sh[217]: ERROR: this script needs /sys/fs/cgroup/cgroup.procs to be empty (for writing the top-level cgroup.subtree_control)
Aug 05 19:38:36 kind-control-plane systemd[1]: kubelet.service: Control process exited, code=exited, status=1/FAILURE
Aug 05 19:38:36 kind-control-plane systemd[1]: kubelet.service: Failed with result 'exit-code'.
Aug 05 19:38:36 kind-control-plane systemd[1]: Failed to start kubelet: The Kubernetes Node Agent.
Aug 05 19:38:37 kind-control-plane systemd[1]: kubelet.service: Scheduled restart job, restart counter is at 1.
Aug 05 19:38:37 kind-control-plane systemd[1]: Stopped kubelet: The Kubernetes Node Agent.
Aug 05 19:38:37 kind-control-plane systemd[1]: Starting kubelet: The Kubernetes Node Agent...
Aug 05 19:38:37 kind-control-plane sh[220]: ERROR: this script needs /sys/fs/cgroup/cgroup.procs to be empty (for writing the top-level cgroup.subtree_control)
Aug 05 19:38:37 kind-control-plane systemd[1]: kubelet.service: Control process exited, code=exited, status=1/FAILURE
Aug 05 19:38:37 kind-control-plane systemd[1]: kubelet.service: Failed with result 'exit-code'.
Aug 05 19:38:37 kind-control-plane systemd[1]: Failed to start kubelet: The Kubernetes Node Agent.
...
```

[...]

PR #2354, opened 2021-07-06
Fix `container` envvar and StopSignal are not set in kindest/node:v1.21.2

ktock:
In a recently released node image, `container` environment variable and `StopSignal` configuration are disabled.

```console
$ docker inspect -f 'Env: {{ .Config.Env }}, StopSignal: {{ .Config.StopSignal }}' kindest/node:v1.21.2@sha256:de73899439c77a4bb6609b10d4eb8b8c22c2d483fa569b112039e652fe4bf81f
Env: [PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin], StopSignal: 
```

In the older version, these configurations were enabled.

```console
$ docker inspect -f 'Env: {{ .Config.Env }}, StopSignal: {{ .Config.StopSignal }}' kindest/node:v1.20.7@sha256:4c68efafa97d278a75a5b4677fb27e4986b03301739f1f87575af32202da9cfe
Env: [PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin container=docker], StopSignal: SIGRTMIN+3
```

Because of this, `kind create cluster` will fail when using kind with the `main` version of containerd:

```
Jul 06 02:06:34 7eceeb9-control-plane containerd[188]: time="2021-07-06T02:06:34.012977745Z" level=error msg="CreateContainer within sandbox \"5b7f060e5088d97bfb09c6423d40be3828743cacff13daa2117137c16dd0a22d\" for &ContainerMetadata{Name:kube-apiserver,Attempt:0,} failed" error="failed to create containerd container: get apparmor_parser version: exec: \"apparmor_parser\": executable file not found in $PATH"
```

(current `kindest/node:v1.21.1` uses containerd v1.5.2 so this doesn't encounter this failure because of [a bug in containerd which is fixed recently](https://github.com/containerd/containerd/pull/5519))

This is because `ENV` and `STOPSIGNAL` are not set in the final stage of `images/base/Dockerfile`.
This commit fixes this issue.

PR #2410, opened 2021-08-08
Debug "CI failing for cgroup2+docker+rootless"

AkihiroSuda:
Debug #2409

### DIFF
--- a/pkg/cluster/internal/providers/docker/provision.go
+++ b/pkg/cluster/internal/providers/docker/provision.go
@@ -21,6 +21,7 @@ import (
 	"net"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"sigs.k8s.io/kind/pkg/cluster/constants"
 	"sigs.k8s.io/kind/pkg/errors"
@@ -131,6 +132,7 @@ func createContainer(args []string) error {
 	if err := exec.Command("docker", args...).Run(); err != nil {
 		return errors.Wrap(err, "docker run error")
 	}
+	time.Sleep(10 * time.Second) // debug
 	return nil
 }
 
--- a/.github/workflows/cgroup2.yaml
+++ b/.github/workflows/cgroup2.yaml
@@ -74,6 +74,25 @@ jobs:
         run: |
           "$HELPER" kind create cluster -v7 --wait 10m --retain
 
+      - name: Debug https://github.com/kubernetes-sigs/kind/issues/2409
+        if: always()
+        run: |
+          cat <<EOF | "$HELPER" "$KIND_EXPERIMENTAL_PROVIDER" exec -i kind-control-plane bash
+          set -x
+          {
+          ls -l /sys/fs/cgroup
+          cat /proc/1/cgroup
+          cat /proc/self/cgroup
+          cat /sys/fs/cgroup/cgroup.procs
+          ps auxw
+          cat /sys/fs/cgroup/cgroup.controllers
+          cat /sys/fs/cgroup/cgroup.subtree_control
+          grep cgroup2 /proc/mounts
+          systemctl status kubelet.service
+          journalctl -u kubelet.service
+          } 2>&1
+          EOF
+
       - name: Get Cluster status
         run: |
           "$HELPER" kubectl wait --for=condition=ready pods --namespace=kube-system -l k8s-app=kube-dns


aojea:
are these restars of kubelet normal?

```
Aug 08 15:20:37 kind-control-plane systemd[1]: Starting kubelet: The Kubernetes Node Agent...
Aug 08 15:20:37 kind-control-plane sh[484]: ERROR: this script needs /sys/fs/cgroup/cgroup.procs to be empty (for writing the top-level cgroup.subtree_control)
Aug 08 15:20:37 kind-control-plane systemd[1]: kubelet.service: Control process exited, code=exited, status=1/FAILURE
Aug 08 15:20:37 kind-control-plane systemd[1]: kubelet.service: Failed with result 'exit-code'.
Aug 08 15:20:37 kind-control-plane systemd[1]: Failed to start kubelet: The Kubernetes Node Agent.
Aug 08 15:20:38 kind-control-plane systemd[1]: kubelet.service: Scheduled restart job, restart counter is at 89.
Aug 08 15:20:38 kind-control-plane systemd[1]: Stopped kubelet: The Kubernetes Node Agent.
Aug 08 15:20:38 kind-control-plane systemd[1]: Starting kubelet: The Kubernetes Node Agent...
Aug 08 15:20:38 kind-control-plane systemd[1]: Started kubelet: The Kubernetes Node Agent.
```

the podman job doesn't have them

pkg/cluster/internal/providers/docker/provision.go
```diff
@@ -245,6 +245,8 @@ func runArgsForNode(node *config.Node, clusterIPFamily config.ClusterIPFamily, n
 		"--volume", "/lib/modules:/lib/modules:ro",
 		// propagate KIND_EXPERIMENTAL_CONTAINERD_SNAPSHOTTER to the entrypoint script
 		"-e", "KIND_EXPERIMENTAL_CONTAINERD_SNAPSHOTTER",
+		// unset container=docker and fake the value, for debugging https://github.com/kubernetes-sigs/kind/issues/2409
+		"-e", "container=podman",
```
AkihiroSuda, line 249: This seems to work, but not sure why.

cc @ktock 
https://github.com/kubernetes-sigs/kind/pull/2354 might be related?

pkg/cluster/internal/providers/docker/provision.go
```diff
@@ -131,6 +132,7 @@ func createContainer(args []string) error {
 	if err := exec.Command("docker", args...).Run(); err != nil {
 		return errors.Wrap(err, "docker run error")
 	}
+	time.Sleep(10 * time.Second) // debug
```
AkihiroSuda, line 135: The “ ERROR: this script needs /sys/fs/cgroup/cgroup.procs to be empty” issue was happening when `docker exec -i` was executed before systemd initializes `/sys fs/cgroup`.

So this sleep works as a workaround.

aojea: do we have some way to "detect it" instead of sleeping?

BenTheElder: I guess we could start watching for a log line in the node output if this is the case.

AkihiroSuda: I'm not sure.

Grepping `Reached target .*Multi-User System.*` is fine?

BenTheElder: I guess. It's unfortunate that this is the case, we have codepaths that should be fine without a running system (e.g. catting the version file, building kubeadm config), and speed is definitely a priority even if we haven't improved on it much lately.

Maybe we can conditionally wait on this when rootless?

BenTheElder:
Did we ever determine what happens if we roll back the image https://github.com/kubernetes-sigs/kind/issues/2409#issuecomment-894803863? Or is it some docker change and the exec timing?

Sorry, still haven't had time to poke at this one

pkg/cluster/internal/providers/docker/provision.go
AkihiroSuda: Opened https://github.com/kubernetes-sigs/kind/pull/2421

> Maybe we can conditionally wait on this when rootless?

I don't think this is specific to rootless, though specific to cgroup v2.

AkihiroSuda:
Closing this and opened #2421 

> Did we ever determine what happens if we roll back the image #2409 (comment)? Or is it some docker change and the exec timing?


I don't think this is a "regression" in the base image nor in docker.
This is a timing issue (on cgroup v2 hosts).

Issue #2413, opened 2021-08-13
regression in performance of kind load docker-image if all images are already present


Alfa-Alex:
**What would you like to be added**:
Currently, when you run `kind load docker-image` and all images are already present on all nodes, `kind` takes 2.5 secs on my machine to return after printing `Image: ... with ID "..." found to be already present on all nodes.`

Looking into the code, I found that the reason for this is simple: 
The code in ll. 144-155 in [docker-image.go](https://github.com/kubernetes-sigs/kind/blob/c78bd3f581a1da72f514c3c6f8ef59f530d13b7d/pkg/cmd/kind/load/docker-image/docker-image.go#L144) is run even if `len(selectedNodes)` is zero.

**Why is this needed**:
In my use case, I load a lot of images via a script by individual `kind load docker-image` invocations (each with exactly one image). Some of these images are already available on the desired node. But currently loading those images slows down the whole process because the above-mentioned part of code (that doesn't have any notable effect for the already-present images) takes 2.5 secs on my machine per `kind load` (I benchmarked this).

My suggestion is to add a simple:
```go
	if len(selectedNodes) == 0 {
		return nil
	}
```
after the loop ending at Line 142. I tested this locally and it allowed me to save the 2.5 seconds.

BenTheElder:
Thanks, this is probably a regression from when support for multiple images was addes, please do send a PR.

Alfa-Alex:
Thanks for the fast response! Closing this because #2414 is merged.

PR #2414, opened 2021-08-13
Return early if all images are already present on all nodes

Alfa-Alex:
Solves #2413.

### DIFF
--- a/pkg/cmd/kind/load/docker-image/docker-image.go
+++ b/pkg/cmd/kind/load/docker-image/docker-image.go
@@ -141,6 +141,11 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
 		}
 	}
 
+	// return early if no node needs the image
+	if len(selectedNodes) == 0 {
+		return nil
+	}
+
 	// Setup the tar path where the images will be saved
 	dir, err := fs.TempDir("", "images-tar")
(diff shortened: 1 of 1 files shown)


Alfa-Alex:
I signed it

Alfa-Alex:
I signed it.

pkg/cmd/kind/load/docker-image/docker-image.go
```diff
@@ -141,6 +141,11 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
 		}
 	}
 
+	// return early if no node needs the image
```
BenTheElder, line 144: unrelated to this PR but the code above also looks suspect ... where we check `len(selectedNodes)` when logging about a particular image even though nodes may have been selected for another image, will address in a follow-up.

BenTheElder: selectedNodes also really ought to be a set, clearly should have reviewed this code more carefully.

BenTheElder (approved):
/lgtm
/approve
thank you!

BenTheElder:
rootless docker at least in github actions is currently known broken https://github.com/kubernetes-sigs/kind/issues/2409

PR #2421, opened 2021-08-16
Block `docker exec` until systemd in the node container initializes /sys/fs/cgroup (Fix docker+cgroup2+rootless CI failure)

### LINKED ISSUE #2409: CI failing for cgroup2+docker+rootless

AkihiroSuda:
Wait until systemd initializes the cgroup hier so that `docker exec` can be executed safely without breaking cgroup v2 hierarchy.

This is implemented by grepping `docker logs` with "Reached target .*Multi-User System.*" message from systemd.
(we can't use `docker exec` to check whether we are allowed to run `docker exec`, obviously.)

Needed for avoiding `ERROR: this script needs /sys/fs/cgroup/cgroup.procs to be empty (for writing the top-level cgroup.subtree_control)`

Fix  #2409
Replace #2410

### DIFF
--- a/pkg/cluster/internal/providers/common/provision.go
+++ b/pkg/cluster/internal/providers/common/provision.go
@@ -1,0 +1,123 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"io"
+	osexec "os/exec"
+	"regexp"
+	"time"
+
+	"sigs.k8s.io/kind/pkg/exec"
+)
+
+type runContainerOpts struct {
+	waitUntilLogRegexpMatches *regexp.Regexp
+}
+
+type RunContainerOpt func(*runContainerOpts) error
+
+// WithWaitUntilLogRegexpMatches waits until the container log matches the regexp.
+func WithWaitUntilLogRegexpMatches(re *regexp.Regexp) RunContainerOpt {
+	return func(o *runContainerOpts) error {
+		o.waitUntilLogRegexpMatches = re
+		return nil
+	}
+}
+
+// WithWaitUntilSystemdReachesMultiUserSystem waits until the systemd in the container
+// reaches "Multi-User System" target, so that `docker exec` can be executed safely without
+// breaking cgroup v2 hierarchy.
+//
+// This is implemented by grepping `docker logs` with "Reached target .*Multi-User System.*"
+// message from systemd.
+// (we can't use `docker exec` to check whether we are allowed to run `docker exec`, obviously.)
+//
+// Needed for avoiding "ERROR: this script needs /sys/fs/cgroup/cgroup.procs to be empty (for writing the top-level cgroup.subtree_control)"
+// See https://github.com/kubernetes-sigs/kind/issues/2409
+func WithWaitUntilSystemdReachesMultiUserSystem() RunContainerOpt {
+	re, err := regexp.Compile("Reached target .*Multi-User System.*")
+	if err != nil {
+		panic(err)
+	}
+	return WithWaitUntilLogRegexpMatches(re)
+}
+
+// RunContainer runs a container.
+// engine is either "docker" or "podman".
+// name is the name of the container.
+// args is to be appended to {engine, "run", "--name", name}.
+func RunContainer(engine, name string, args []string, opts ...RunContainerOpt) error {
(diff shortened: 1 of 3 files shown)


AkihiroSuda:
@aojea @BenTheElder PTAL

aojea:
> Wait until systemd initializes the cgroup hier so that docker exec can be executed safely without breaking cgroup v2 hierarchy.

is one specific `docker exec` who breaks the cgroup? os is the entrypoint script or one of the systemd units?

i.e, if this is because of the `create-kubelet-cgroup-v2`script we can just delay it with something like
```
diff --git a/images/base/files/etc/systemd/system/kubelet.service.d/10-kubeadm.conf b/images/base/files/etc/systemd/system/kubelet.service.d/10-kubeadm.conf
index 04fd9853..804b7850 100644
--- a/images/base/files/etc/systemd/system/kubelet.service.d/10-kubeadm.conf
+++ b/images/base/files/etc/systemd/system/kubelet.service.d/10-kubeadm.conf
@@ -13,3 +13,6 @@ EnvironmentFile=-/etc/default/kubelet
 ExecStartPre=/bin/sh -euc "if [ -f /sys/fs/cgroup/cgroup.controllers ]; then create-kubelet-cgroup-v2; fi"
 ExecStart=
 ExecStart=/usr/bin/kubelet $KUBELET_KUBECONFIG_ARGS $KUBELET_CONFIG_ARGS $KUBELET_KUBEADM_ARGS $KUBELET_EXTRA_ARGS --cgroup-root=/kubelet
+[Install]
+Requires=multi-user.target
```

AkihiroSuda:
I don't think delaying `create-kubelet-cgroup-v2` make sense.
The problem is that running `docker exec` before systemd initializes cgroups breaks cgroups hier.

AkihiroSuda:
@BenTheElder Could you take a look

pkg/cluster/internal/providers/common/provision.go
```diff
@@ -0,0 +1,123 @@
...
+	fullArgs := append([]string{"run", "--name", name}, args...)
+	if err := exec.Command(engine, fullArgs...).Run(); err != nil {
+		return fmt.Errorf("%s run error: %w", engine, err)
+	}
+
+	if o.waitUntilLogRegexpMatches != nil {
+		logCtx := context.Background()
+		logCtx, logCancel := context.WithTimeout(logCtx, 30*time.Second)
+		defer logCancel()
+		// use os/exec.CommandContext directly, as kind/pkg/exec.CommandContext lacks support for killing
```
BenTheElder, line 83: IMO this should be fixed instead in the future. The abstraction exists for purposes such as testing and composition with running commands in nested environments.

BenTheElder: It's also not clear that we need to kill processes directly at all? We should be able to kill indirectly by telling docker/podman to delete the container, which is what the rest of the tool relies on.

BenTheElder (approved):
For a bugfix like this I would prefer to unblock and @aojea you can approve at your own discretion.

I'm available today doing some catchup but I will once again be out with an ongoing family matter ..

I don't think this is implemented quite ideally but given that it's internal implementation details we can always alter that in a future PR. If it were adding a new user facing feature I might try to be more immediately involved up-front.

Issue #2419, opened 2021-08-15
Node labels on being moved to other nodes


fergalk:
Kind appears to be applying node labels defined in v1alpha4.Node.Labels onto other nodes in the kind configuration. Other configuration is not moved between nodes, which presents an issue for my use-case (using labels & taints to control pod placement in daemonsets). This behaviour has been observed on mac os & linux - not tested on Windows.

**What happened**:

Labels defined on one worker node were applied to another worker node. The other configuration defined for the node was applied to the correct node.

**What you expected to happen**:

Labels to be applied to correct nodes as per v1alpha4 cluster configuration file.

**How to reproduce it (as minimally and precisely as possible)**:

Create kind config file `config.yaml` with below contents:

```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
- role: worker
- role: worker
  labels:
    myLabel: someValue
  extraPortMappings:
  - containerPort: 80
    hostPort: 2000
```

Run `kind create cluster --config config.yaml --name test`

Validate misconfiguration (sample output shown):

```
$ kubectl  get node test-worker -ojsonpath={.metadata.labels}
{"beta.kubernetes.io/arch":"amd64","beta.kubernetes.io/os":"linux","kubernetes.io/arch":"amd64","kubernetes.io/hostname":"test-worker","kubernetes.io/os":"linux","myLabel":"someValue"}
$ kubectl  get node test-worker2 -ojsonpath={.metadata.labels}
{"beta.kubernetes.io/arch":"amd64","beta.kubernetes.io/os":"linux","kubernetes.io/arch":"amd64","kubernetes.io/hostname":"test-worker2","kubernetes.io/os":"linux"}
$ docker inspect test-worker -f {{.NetworkSettings.Ports}}
map[]
$ kubecdocker inspect test-worker2 -f {{.NetworkSettings.Ports}}
map[80/tcp:[{0.0.0.0 2000}]]
```

**Anything else we need to know?**: N/A

**Environment:**

- kind version: (use `kind version`): kind v0.11.1 go1.16.4 linux/amd64
[...]

PR #2423, opened 2021-08-18
assign node labels correctly

aojea:
The labels assigned by the users in the configuration file were
assigned randomly to the nodes.
Since nodes are names deterministically respecting the configuration
file order, we can map the labels to the corresponding node.

Also, do some cleanup and reduce duplicate code.

Fixes: #2419


BenTheElder:
Did you confirm the actual labels on the running nodes? I forget how we're plumbing them through and another kind gotcha right now is the way kubeadm handles kubelet config (cluster wide from first node)

aojea:
/hold
sorry got distracted, I need to test this

neolit123:
kind seems to be using the kubelet flags to add labels:
https://github.com/kubernetes-sigs/kind/blob/515fb3aeb2ae4af6d8229754a152eccd016d1d15/pkg/cluster/internal/kubeadm/config.go#L353

which is actually the only option from the kubelet side, since the kubelet v1beta1 currently lacks taints and labels, they are only possible via flags:
https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/

should be noted that the kubelet would fail with user provided k8s.io labels as these restricted, so you might see some reports from user about it, if not already:
https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#noderestriction
could be documented in the kind docs (if not already). disabling noderestriction is a less desired workaround.

i think ideally the safest contract for kind would be to take the labels the users wants and use a privileged client to apply them as blocking step during `kind create`, but this means kind would have to import client-go or bootleg a repo-local HTTPS client.

aojea:
/hold cancel
3/3 the label is in the right node

BenTheElder:
> should be noted that the kubelet would fail with user provided k8s.io labels as these restricted, so you might see some reports from user about it, if not already:

Yes, I don't think we should do any work ourselves to support applying restricted labels. Users should pick a label in their own namespace.

PR #2429, opened 2021-08-20
feat(cmd/list):  label images with imported label and list them according to that label

### LINKED ISSUE #2425: list all installed images via kind load docker-image

I would like to add a new subcommand like `list load-images` that will help us to list all installed docker images via the command `kind load docker-image` and this will also help us to check whether the image is loaded successfully or not.

```shell
$ kind load docker-image docker.io/librar/alpine:latest
Image: "docker.io/library/alpine:latest" with ID "sha256:021b3423115ff662225e83d7e2606475217de7b55fde83ce3447a54019a77aa2" found to be already present on all nodes.
$ kind list load-images
IMAGE                                           ID
docker.io/library/alpine:latest     sha256:021b3423115ff662225e83d7e2606475217de7b55fde83ce3447a54019a77aa2
```

**Why is this needed**:

It provides better control over checking whether the image is loaded successfully or not.

cc: @Dentrax

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>
Co-authored-by: Furkan <furkan.turkal@trendyol.com>
Fixes #2425 

Screen Shot 2021-08-20 at 14 41 20

Screen Shot 2021-08-20 at 15 13 47

### DIFF
--- a/pkg/cmd/kind/list/list.go
+++ b/pkg/cmd/kind/list/list.go
@@ -1,0 +1,48 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package list implements the `list` command
+package list
+
+import (
+	"errors"
+	loadimage "sigs.k8s.io/kind/pkg/cmd/kind/list/load-image"
+
+	"github.com/spf13/cobra"
+
+	"sigs.k8s.io/kind/pkg/cmd"
+	"sigs.k8s.io/kind/pkg/log"
+)
+
+// NewCommand returns a new cobra.Command for get
+func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
+	cmd := &cobra.Command{
+		Args:  cobra.NoArgs,
+		Use:   "list",
+		Short: "List loaded images into nodes",
+		Long:  "List loaded images into node from an archive or image on host",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			err := cmd.Help()
+			if err != nil {
+				return err
+			}
+			return errors.New("Subcommand is required")
+		},
+	}
+	// add subcommands
+	cmd.AddCommand(loadimage.NewCommand(logger, streams))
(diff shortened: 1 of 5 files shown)


pkg/cmd/kind/list/list.go
```diff
@@ -0,0 +1,48 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
neolit123, line 2: new files should have 2021

neolit123 (commented):
looks like https://github.com/kubernetes-sigs/kind/issues/2425 has not received maintainer feedback whether this feature would be accepted.

i'm giving a tentative +1 for this, but ideally the PR should have been filed after discussion.

pkg/cmd/kind/list/list.go
developer-guy: fixed @neolit123 🙏🤝

developer-guy:
> i'm giving a tentative +1 for this

yes you are right @neolit123 but we couldn't wait for it because we are so thrilled about this issue, so we wanted to do it ASAP. 🙏🤩

pkg/cmd/kind/list/load-image/load-image.go
```diff
@@ -0,0 +1,134 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
neolit123, line 2: this file as well.

BenTheElder:
does containerd have guidance about these labels? possibly we should be using a namespaced label?
what happens if an image was imported at one point but then later pulled?

I think we should discuss in #2425, I have some other points to add there.

pkg/cmd/kind/list/load-image/load-image.go
developer-guy: fixed @neolit123 🙏

developer-guy:
> does containerd have guidance about these labels?

Yeah, we found old documentation about filtering labels, you can see it from [here](https://github.com/containerd/containerd/blob/main/reports/2017-06-23.md).

> possibly we should be using a namespaced label?

We realized that KinD is using `k8s.io` as a namespace by default while [loading](https://github.com/kubernetes-sigs/kind/blob/b6bc112522651d98c81823df56b7afa511459a3b/pkg/cluster/nodeutils/util.go#L79) docker-images, so we obey the convention while [labeling](https://github.com/kubernetes-sigs/kind/pull/2429/files#diff-f27efd6b0e4f4b09a6bcf127aa4681d2c9bc33e9ae689344c31efd7e55fc441cR93) them.

> what happens if an image was imported at one point but then later pulled?

I'm not sure about this one. Because there is a mismatch here, right? If an image can later be pulled, why do we want to import it? I think also if an image can later be pulled, this means that this one is not our image right now, so we should not add it to the list while listing the loaded images. 

WDYT @Dentrax?

BenTheElder:
https://github.com/kubernetes-sigs/kind/issues/2425#issuecomment-1512461067

PR #2367, opened 2021-07-14
CI: fix Fedora 34 VM

AkihiroSuda:
Fix #2361 , by upgrading packages and rebooting

https://github.com/containers/podman/issues/10928#issuecomment-879662331

PR #2432, opened 2021-08-24
Revert workarounds for CI(rootless podman)

xinydev:
Revert this PR https://github.com/kubernetes-sigs/kind/pull/2367  

Because podman v3.3.0 is released, we don't need these workarounds anymore, and I think it might cover up some problems?

cc @AkihiroSuda


aojea:
/lgtm
/approve
Great, thanks

PR #2434, opened 2021-08-25
bump go to 1.17

AkihiroSuda:
Required by  `pull-kind-e2e-kubernetes` due to kubernetes/kubernetes#103692
```
!!! [0825 06:57:07]  1: hack/run-in-gopath.sh:31 kube::golang::setup_env(...)
Detected go version: go version go1.16.7 linux/amd64.
Kubernetes requires go1.17.0 or greater.
Please install go1.17.0 or later.
```

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/2421/pull-kind-e2e-kubernetes/1430420896627036160

PR #2435, opened 2021-08-25
bump golang to 1.17

aojea:


aojea:
kind jobs are failing on ci https://testgrid.k8s.io/sig-testing-kind#kind-master-parallel&width=20

BenTheElder:
@AkihiroSuda already filed #2434

aojea:
> @AkihiroSuda already filed #2434

PR #2438, opened 2021-08-25
PR 2408 rebase + base image bump

BenTheElder:
so this can run in CI prior to merge. I know the workflow is not great but we also don't really have the bandwidth to work on that right now

xref https://github.com/kubernetes-sigs/kind/pull/2408


BenTheElder:
ah we also have to bump the node image for the github actions CI

PR #2443, opened 2021-09-01
refactor: move from io/ioutil to io and os package

Juneezee:
The `io/ioutil` package has been deprecated in Go 1.16 (See https://golang.org/doc/go1.16#ioutil). This PR replaces the existing `io/ioutil` functions with their new definitions in `io` and `os` packages.

This PR also upgrades the go directive in go.mod to 1.17.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module sigs.k8s.io/kind
 
-go 1.14
+go 1.17
 
 require (
 	github.com/BurntSushi/toml v0.3.1
@@ -14,4 +14,10 @@ require (
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
 	k8s.io/apimachinery v0.21.2
 	sigs.k8s.io/yaml v1.2.0
+)
+
+require (
+	github.com/inconshreveable/mousetrap v1.0.0 // indirect
+	golang.org/x/sys v0.0.0-20210426230700-d19ff857e887 // indirect
+	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
--- a/pkg/fs/fs.go
+++ b/pkg/fs/fs.go
@@ -21,7 +21,6 @@ package fs
 
 import (
 	"io"
-	"io/ioutil"
 	"os"
 	"path"
 	"path/filepath"
@@ -29,10 +28,10 @@ import (
 	"strings"
 )
 
-// TempDir is like ioutil.TempDir, but more docker friendly
+// TempDir is like os.MkdirTemp, but more docker friendly
 func TempDir(dir, prefix string) (name string, err error) {
 	// create a tempdir as normal
-	name, err = ioutil.TempDir(dir, prefix)
+	name, err = os.MkdirTemp(dir, prefix)
 	if err != nil {
 		return "", err
 	}
@@ -138,14 +137,19 @@ func copyDir(src, dst string, info os.FileInfo) error {
 		return err
 	}
 	// copy every source dir entry
-	entries, err := ioutil.ReadDir(src)
+	entries, err := os.ReadDir(src)
 	if err != nil {
 		return err
 	}
 	for _, entry := range entries {
 		entrySrc := filepath.Join(src, entry.Name())
 		entryDst := filepath.Join(dst, entry.Name())
-		if err := copy(entrySrc, entryDst, entry); err != nil {
+		info, err := entry.Info()
+		if err != nil {
+			return err
+		}
+
+		if err := copy(entrySrc, entryDst, info); err != nil {
 			return err
 		}
 	}
--- a/cmd/kind/app/main.go
+++ b/cmd/kind/app/main.go
@@ -17,7 +17,7 @@ limitations under the License.
 package app
 
 import (
-	"io/ioutil"
+	"io"
 	"os"
 
 	"github.com/spf13/pflag"
@@ -45,7 +45,7 @@ func Run(logger log.Logger, streams cmd.IOStreams, args []string) error {
 		// if we are in quiet mode, we want to suppress all status output
 		// only streams.Out should be written to (program output)
 		logger = log.NoopLogger{}
-		streams.ErrOut = ioutil.Discard
+		streams.ErrOut = io.Discard
 	}
 	// actually run the command
 	c := kind.NewCommand(logger, streams)
--- a/pkg/build/nodeimage/imageimporter.go
+++ b/pkg/build/nodeimage/imageimporter.go
@@ -17,7 +17,7 @@ limitations under the License.
 package nodeimage
 
 import (
-	"io/ioutil"
+	"io"
 
 	"sigs.k8s.io/kind/pkg/exec"
 )
@@ -51,7 +51,7 @@ func (c *containerdImporter) Pull(image, platform string) error {
 	// without discarding output
 	return c.containerCmder.Command(
 		"ctr", "--namespace=k8s.io", "images", "pull", "--platform="+platform, image,
-	).SetStdout(ioutil.Discard).SetStderr(ioutil.Discard).Run()
+	).SetStdout(io.Discard).SetStderr(io.Discard).Run()
 }
 
 func (c *containerdImporter) LoadCommand() exec.Cmd {
--- a/pkg/build/nodeimage/internal/container/docker/archive.go
+++ b/pkg/build/nodeimage/internal/container/docker/archive.go
@@ -23,7 +23,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"os"
 	"strings"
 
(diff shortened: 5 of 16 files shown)


go.mod
```diff
@@ -15,3 +15,9 @@ require (
 	k8s.io/apimachinery v0.21.2
 	sigs.k8s.io/yaml v1.2.0
 )
+
+require (
```
aojea, line 19: why are there 2 require statements?

aojea:
/ok-to-test
/hold
is the go.mod right?

go.mod
Juneezee: This is because Go 1.17 now has pruned module graph. You can see the full details in the changelog here: https://golang.org/doc/go1.17#go-command

> For the go command to correctly resolve transitive imports using the pruned module graph, the go.mod file for each module needs to include more detail about the transitive dependencies relevant to that module. If a module specifies go 1.17 or higher in its go.mod file, **its go.mod file now contains an explicit require directive for every module that provides a transitively-imported package**. (In previous versions, the go.mod file typically only included explicit requirements for directly-imported packages.)

aojea: lovely, thanks TIL

pkg/fs/fs.go
```diff
@@ -138,14 +137,19 @@ func copyDir(src, dst string, info os.FileInfo) error {
...
 	for _, entry := range entries {
 		entrySrc := filepath.Join(src, entry.Name())
 		entryDst := filepath.Join(dst, entry.Name())
-		if err := copy(entrySrc, entryDst, entry); err != nil {
+		info, err := entry.Info()
+		if err != nil {
+			return err
+		}
+
+		if err := copy(entrySrc, entryDst, info); err != nil {
```
aojea, line 152: this seems the only change that is not a 1to1 replacement :+1:

aojea:
/lgtm
/assign @BenTheElder 
Ben you should be more exposed to this new golang1.7 things, feel free to unhold

pkg/fs/fs.go
Juneezee: Yes, `os.ReadDir` returns a slice of `os.DirEntry` rather than a slice of `fs.FileInfo`. This is the only function that is not a direct replacement.

BenTheElder:
What is the benefit to requiring newer go here?
See the same discussion in kubernetes/kubernetes

Juneezee:
> What is the benefit to requiring newer go here?
> See the same discussion in kubernetes/kubernetes

Upgrading to Go 1.17 allows the module to take advantage of [Lazy Module Loading](https://golang.org/ref/mod#lazy-loading). Since this doesn't introduce any breaking changes, I think it is a good time to upgrade the Go version to possibly reduce the technical debt.

BenTheElder:
> @BenTheElder Upgrading to Go 1.17 allows the module to take advantage of Lazy Module Loading. 

That seems all but completely irrelevant with the standard library? The module graph is not going to substantially change, the standard library cannot depend on any external dependencies.

> Since this doesn't introduce any breaking changes, I think it is a good time to upgrade the Go version to possibly reduce the technical debt.

That's incorrect, at a quick glance I see this uses `io.ReadAll` which was introduced in go1.16. KIND does not use any features that would be required in go1.16, hence the go1.14 in the `go.mod` (which tells go which version to require). We upgrade go regularly ourselves but are not in the habit of requiring users to upgrade. Some users are embedded kind as a library in their own packages.

https://pkg.go.dev/io#ReadAll

Juneezee:
@BenTheElder Once again, thank you for taking the time to review this PR.

> > @BenTheElder Upgrading to Go 1.17 allows the module to take advantage of Lazy Module Loading.
> 
> That seems all but completely irrelevant with the standard library? The module graph is not going to substantially change, the standard library cannot depend on any external dependencies.

Yes, the standard library does not depend on any external dependencies. The new Module Graph Pruning and Lazy Module Loading features are about the dependencies needed to build any package in the main module (in our case, the KIND module, sigs.k8s.io/kind).

According to the [Go Modules Reference: Glossary](https://golang.org/ref/mod#glossary),
> **lazy module loading:** A change in Go 1.17 that avoids loading the module graph for commands that do not need it in modules that specify go 1.17 or higher. See Lazy module loading.
**module graph pruning:** A change in Go 1.17 that reduces the size of the module graph by omitting transitive dependencies of modules that specify go 1.17 or higher. See Module graph pruning.

Correct me if I am wrong, it seems that the new Module Graph Pruning and Lazy Module Loading features are available to the main module and its dependencies. 


> > Since this doesn't introduce any breaking changes, I think it is a good time to upgrade the Go version to possibly reduce the technical debt.
> 
> That's incorrect, at a quick glance I see this uses `io.ReadAll` which was introduced in go1.16. KIND does not use any features that would be required in go1.16, hence the go1.14 in the `go.mod` (which tells go which version to require). We upgrade go regularly ourselves but are not in the habit of requiring users to upgrade. Some users are embedded kind as a library in their own packages.
> 
> https://pkg.go.dev/io#ReadAll

You are right actually :sweat_smile:. What I meant for the "_no breaking changes_" was actually no breaking API changes, i.e. assume that another "Module A" is using KIND as a dependency and has upgraded to Go 1.17, that "Module A" is expected to work.

My primary intention of creating this PR is to follow the [Go Release Policy](https://golang.org/doc/devel/release#policy), since Go 1.14 has reached end-of-life.

Image source: https://endoflife.date/go

BenTheElder:
> Correct me if I am wrong, it seems that the new Module Graph Pruning and Lazy Module Loading features are available to the main module and its dependencies.

Sure, but there's not much upside to avoiding loading ioutil, it's not that big, and it's also not affecting the module graph in any meaningful way. It's one package with a single version. We've been pretty aggressive with binary size, kind is around 6mb uncompressed which is minuscule in the domain and yet this doesn't seem impactful.



> My primary intention of creating this PR is to follow the Go Release Policy, since Go 1.14 has reached end-of-life.

I appreciate the intent, but I don't think even if the go project has EOL 1.14 developers are done using 1.14. Similarly Kubernetes has EOL 1.19 IIRC but kind supports back to 1.14 at the moment and as many versions as we can reasonably manage without major hinderance because it is hard for our users to upgrade (they need to match their production target versions). They also tend to lock in the go version or `go get` from machines with e.g. go installed via the debian package or similar.

We're well aware of the go version policy due to supporting builds in Kubernetes. Kubernetes often winds up on technically out of support versions and developers using kind as a library or building with go get tend to be pinned to whatever go version matches the (potentially out of support) kubernetes version in use due to client-go etc. We support kubernetes back to 1.14+, which uses go 1.12.12 actually.

I forget what motivated us to go to 1.14, but I don't think we want to go to 1.16 required yet, it's not even a year old. Debian stable is packaging go 1.15, kubernetes is only on at least 1.16 for v1.21+ (released April 2021), 1.20 is still officially supported with go 1.15.

Unless go causes an actual problem with these packages I don't think we should rush to migrate unless we carefully do so in a way that does not break users with older golang. We basically just require modules-enabled golang because that was a significant development change in the community early in the project's lifecycle.

stmcginnis:
FWIW, I'm in favor of moving to a more recent version of go. From what I've seen, I _think_ developers are generally staying with newer minor releases (though I'll admit no actual data to back up that claim).

There are other benefits to newer releases that aren't explicitly called out in release notes, like minor bug fixes and improvements.

There isn't a strongly compelling reason to move up though. So maybe at a minimum, an outcome of this discussion could be to get the `kind` version policy written down somewhere so it's understood.

BenTheElder:
> FWIW, I'm in favor of moving to a more recent version of go. From what I've seen, I think developers are generally staying with newer minor releases (though I'll admit no actual data to back up that claim).
>
> There are other benefits to newer releases that aren't explicitly called out in release notes, like minor bug fixes and improvements.
>
> There isn't a strongly compelling reason to move up though. So maybe at a minimum, an outcome of this discussion could be to get the kind version policy written down somewhere so it's understood.

We upgrade to the latest version at-will, best effort from maintainers. kind *does* use the latest version, however the latest version is able to use the old packages/functions from the standard library just fine due to go1 compatibility.

I don't see the benefit currently to _requiring_ users of this package to use the latest go. Some users are embedding kind as a library and kind works fine with many go versions, we aren't using any special language features / incompatible standard library additions and don't really need to.

As a binary project, we are currently on go1.17.3
https://github.com/kubernetes-sigs/kind/blob/50c4dce467a4cd11ab8e756ae79032998dceb63d/.go-version#L1

Details on that file are covered in the contributor guide at https://kind.sigs.k8s.io/docs/contributing/development/#building

stmcginnis:
> Unless go causes an actual problem with these packages I don't think we should rush to migrate unless we carefully do so in a way that does not break users with older golang.

Should this be closed for now?

BenTheElder:
Yes, let's revisit when there's a clear benefit from switching or we otherwise have reason to require go >= 1.16 instead of 1.14.
The methods we are using appear to be stable. And similarly in kubernetes kind supports versions that are no longer in upstream support.

Issue #2455, opened 2021-09-08
remove the need for cp/mkdir/kill from haproxy image


jaxesn:
**What should be cleaned up or changed**:
By using docker/podman cp/kill directly we can avoid needing [cp/mkdir/kill](https://github.com/kubernetes-sigs/kind/blob/main/images/haproxy/Dockerfile#L46) in the final haproxy container image.  Files could be tar'd and passed via stdin to docker cp directly. docker kill can be used to trigger config reload instead of calling kill inside the container.

**Why is this needed**:
This would be a pretty small optimization, but wanted to see if there was any interest in coming up with a solution to remove these so the final haproxy image really only has haproxy (plus distroless cc) in it.

BenTheElder:
It probably makes sense at some point but as you said the optimization is pretty small.

The downside is it will increase the surface area of the otherwise still relatively minimal node provider, and we may need to consider what that looks like for future providers. Currently nearly everything is drop-in portable to any container runtime with support for exec.

We do have some code comments noting that this may make sense to investigate anyhow. I'm not sure it will actually be more clean though.

BenTheElder:
I think we decided not to go forward with this, it's such a small optimization and we're not actually worried about e.g. code exec in this container. We may have to revert the whole distroless version of this image anyhow due to issues upgrading versions. (lost track of it, but there's discussion on a PR where we ran into this).

thanks so much for taking a stab at this and filing the issues and PRs anyhow! 🙏

PR #2456, opened 2021-09-08
WIP: remove the need for cp/mkdir/kill from haproxy image

jaxesn:
I haven't test podman yet, but according to the docs it should function the same way. We could also avoid using cat here, if that was desirable, by use docker cp with stdout instead, but that is not necessary for the haproxy use case.

TODO:

- [ ]  Support via podman as well as docker
- [ ]  Change nodeimage helper to use a tar file instead of calling cp

Issue: https://github.com/kubernetes-sigs/kind/issues/2455

### DIFF
--- a/pkg/cluster/internal/loadbalancer/const.go
+++ b/pkg/cluster/internal/loadbalancer/const.go
@@ -17,7 +17,7 @@ limitations under the License.
 package loadbalancer
 
 // Image defines the loadbalancer image:tag
-const Image = "kindest/haproxy:v20210715-a6da3463"
+const Image = "kindest/haproxy:latest"
 
 // ConfigPath defines the path to the config file in the image
 const ConfigPath = "/usr/local/etc/haproxy/haproxy.cfg"
--- a/pkg/cluster/internal/providers/docker/node.go
+++ b/pkg/cluster/internal/providers/docker/node.go
@@ -100,34 +100,19 @@ type nodeCmd struct {
 }
 
 func (c *nodeCmd) Run() error {
-	args := []string{
-		"exec",
-		// run with privileges so we can remount etc..
-		// this might not make sense in the most general sense, but it is
-		// important to many kind commands
-		"--privileged",
+	var args []string
+	switch c.command {
+	case "cp":
+		if c.stdin == nil {
+			return errors.Errorf("cp not supported without stdin") 
+		}
+		args = c.cpArgs()		
+	case "kill":
+		args = c.killArgs()
+	default:
+		args = c.execArgs()
 	}
-	if c.stdin != nil {
-		args = append(args,
-			"-i", // interactive so we can supply input
-		)
-	}
-	// set env
-	for _, env := range c.env {
-		args = append(args, "-e", env)
-	}
-	// specify the container and command, after this everything will be
-	// args the command in the container rather than to docker
-	args = append(
-		args,
-		c.nameOrID, // ... against the container
-		c.command,  // with the command specified
-	)
-	args = append(
-		args,
-		// finally, with the caller args
-		c.args...,
-	)
+	
 	var cmd exec.Cmd
 	if c.ctx != nil {
 		cmd = exec.CommandContext(c.ctx, "docker", args...)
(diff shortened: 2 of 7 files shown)


jaxesn:
CLA signed

pkg/cluster/internal/providers/docker/node.go
```diff
@@ -169,3 +154,51 @@ func (c *nodeCmd) SetStderr(w io.Writer) exec.Cmd {
 func (n *node) SerialLogs(w io.Writer) error {
 	return exec.Command("docker", "logs", n.name).SetStdout(w).SetStderr(w).Run()
 }
+
+func (c *nodeCmd) cpArgs() []string {
+	return []string {
+		"cp",
```
BenTheElder, line 160: Doesn't docker cp depend on a tar binary? That's probably *larger* than cp/mkdir and certainly more complex.

pkg/cluster/internal/providers/docker/node.go
```diff
@@ -100,34 +100,19 @@ type nodeCmd struct {
...
-	if c.stdin != nil {
-		args = append(args,
-			"-i", // interactive so we can supply input
-		)
-	}
-	// set env
-	for _, env := range c.env {
-		args = append(args, "-e", env)
+	var args []string
+	switch c.command {
```
BenTheElder, line 104: This seems like a fragile hack. The cp replacement for example is *not* identical to running posix cp commands on the nodes. I don't think we should magically switch to a different mode of operation based on argv[0].

If we do go forward with this change probably it should call some new node methods.

pkg/cluster/nodeutils/util.go
```diff
@@ -44,34 +43,42 @@ func KubeVersion(n nodes.Node) (version string, err error) {
 
 // WriteFile writes content to dest on the node
 func WriteFile(n nodes.Node, dest, content string) error {
-	// create destination directory
-	err := n.Command("mkdir", "-p", path.Dir(dest)).Run()
+	// destination folder exists in tar and is created during cp from stdin
```
BenTheElder, line 46: Without knowing the hack above this snippet on its own doesn't make sense. It should reference that. But also probably we should not have multiple meanings for command. It is inconsistent with the other command implementations and breaks the potential for testing with a fake which is much of the point of the exec wrapper besides the node command abstraction utility built from it.

BenTheElder (commented):
Thanks for the PR!
I'm not sure this is quite the right approach currently. cc @aojea.
In the future we tend to prefer to discuss approach on issues first https://kind.sigs.k8s.io/docs/contributing/getting-started/ 😅

pkg/cluster/internal/loadbalancer/const.go
```diff
@@ -17,7 +17,7 @@ limitations under the License.
 package loadbalancer
 
 // Image defines the loadbalancer image:tag
-const Image = "kindest/haproxy:v20210715-a6da3463"
+const Image = "kindest/haproxy:latest"
```
aojea, line 20: pinned versions please, otherwise we can't detect problems caused by different versions

aojea:
> Thanks for the PR!
> I'm not sure this is quite the right approach currently. cc @aojea.
> In the future we tend to prefer to discuss approach on issues first https://kind.sigs.k8s.io/docs/contributing/getting-started/

I'm going to be super conservative here, it seems too much risk to change these core methods that were pretty reliable just for  haproxy , is this haproxy thing causing some problem or is just cosmetic?

pkg/cluster/internal/providers/docker/node.go
jaxesn: I don't think so, I was able to create clusters using the haproxy image I built without cp/mkdir/kill in the image and had no issues copying the cfg into the container.  I believe docker is using tools from the host to tar/untar input and output.

pkg/cluster/internal/providers/docker/node.go
jaxesn: Totally agree.  I was hoping to tease the approach and start the conversation before going through and introducing new methods on the Note interface, something that I am pretty sure I'd need some direction from ya'll on.

jaxesn:
> > Thanks for the PR!
> > I'm not sure this is quite the right approach currently. cc @aojea.
> > In the future we tend to prefer to discuss approach on issues first https://kind.sigs.k8s.io/docs/contributing/getting-started/
> 
> I'm going to be super conservative here, it seems too much risk to change these core methods that were pretty reliable just for haproxy , is this haproxy thing causing some problem or is just cosmetic?

I think that is totally fair.  I was going through the dockerfile while building a AL2 based version for use in EKS Anywhere and noticed the note about potentially refactoring to avoid mkdir/cp.  Being a bit picky about what makes it in the final image, even if it is admittedly super minor, I figured I'd try an approach to potentially remove these dependencies just to see if there was interested.

Even still, I would be interested in taking this further, even if there are other approaches you'd prefer me try out.  As said above, there is not a ton of value add here, these are minor dependencies running in a container that is really only meant for local development, so I totally understand if this is not something worth pursuing.

pkg/cluster/internal/loadbalancer/const.go
aojea: ahh, got it

BenTheElder:
https://github.com/kubernetes-sigs/kind/issues/2455#issuecomment-1030401606 🙏

Issue #2460, opened 2021-09-14
could not find a line that matches "Reached target .*Multi-User System.*"


ryanjbaxter:
We started receiving this error recently when creating a Kind cluster on CircleCI.

https://app.circleci.com/pipelines/github/spring-cloud/spring-cloud-kubernetes/1008/workflows/c6d1d226-2028-41ed-a5d8-d0c912f3e0d1/jobs/2090?invite=true#step-108-220

I am pretty sure it has to do with this change

https://github.com/kubernetes-sigs/kind/pull/2421

Any ideas what we can do to make Kind work on CircleCI?

Right now I have reverted to using the 0.11.1 release for now.

AkihiroSuda:
> https://app.circleci.com/pipelines/github/spring-cloud/spring-cloud-kubernetes/1008/workflows/c6d1d226-2028-41ed-a5d8-d0c912f3e0d1/jobs/2090?invite=true#step-108-220

Can't see the logs.

Could you post the full logs of `docker logs kind-control-plane`?

ryanjbaxter:
Sorry about that.

Here are the logs from the build on CircleCI

```
+ main
+ install_latest_kind
+ local tmp_dir
++ TMPDIR=/tmp/tmp.4HBinlkgtF
++ mktemp -d /tmp/tmp.4HBinlkgtF/kind-source.XXXXX
+ tmp_dir=/tmp/tmp.4HBinlkgtF/kind-source.Qm5mm
+ cd /tmp/tmp.4HBinlkgtF/kind-source.Qm5mm
+ git clone https://github.com/kubernetes-sigs/kind
Cloning into 'kind'...
Warning: Permanently added the RSA host key for IP address '140.82.113.4' to the list of known hosts.
remote: Enumerating objects: 23489, done.
remote: Counting objects: 100% (635/635), done.
remote: Compressing objects: 100% (329/329), done.
remote: Total 23489 (delta 312), reused 530 (delta 255), pack-reused 22854
Receiving objects: 100% (23489/23489), 11.69 MiB | 17.30 MiB/s, done.
Resolving deltas: 100% (13297/13297), done.
+ cd ./kind
+ make install INSTALL_DIR=/tmp/tmp.4HBinlkgtF
[...]

PR #2466, opened 2021-09-18
wait 90sec for systemd to finish booting

aojea:
Free CIs use to have very bad performance, so it is common operations to take time.
Wait for 90seconds until systemd finish, instead of 30


AkihiroSuda:
https://github.com/kubernetes-sigs/kind/issues/2460#issuecomment-921020265

“exec format error” implies that the timeout is probably unrelated.

aojea:
> [#2460 (comment)](https://github.com/kubernetes-sigs/kind/issues/2460#issuecomment-921020265)
> 
> “exec format error” implies that the timeout is probably unrelated.

/hold

Issue #2468, opened 2021-09-21
Validation: ipv6 only with ipv4 pod subnet


fedepaol:
**What should be cleaned up or changed**:

It is currently possible to create an ipv6 only cluster specifying an ipv4 pod subnet.
I am not sure if this combination should be blocked during the validation phase, in case I can try to provide a PR. 

**Why is this needed**:

With this combination, pods won't be able to access the api server which is under a v6 address, making the cluster unusable.
I.e: 
```
kubectl -n local-path-storage logs local-path-provisioner-547f784dff-lvmss
time="2021-09-21T07:07:33Z" level=fatal msg="Error starting daemon: Cannot start Provisioner: failed to get Kubernetes server version: Get https://[fd00:10:96::1]:443/version?timeout=32s: dial tcp [fd00:10:96::1]:443: connect: network is unreachable" 
```

fedepaol:
cc @aojea

aojea:
I was surprised when Fede reported this but is true, I can't remember why I've added validation for dual but not for ip families :thinking:  , @BenTheElder do you remember something about this topci?


If you want to contribute to this, it should be done here

https://github.com/kubernetes-sigs/kind/blob/4910c3e221a858e68e29f9494170a38e1c4e8b80/pkg/internal/apis/config/validate.go#L54-L63

fedepaol:
> If you want to contribute to this, it should be done here
> 
> https://github.com/kubernetes-sigs/kind/blob/4910c3e221a858e68e29f9494170a38e1c4e8b80/pkg/internal/apis/config/validate.go#L54-L63

This is removing half of the fun :P

BenTheElder:
this makes sense to me, thanks for catching this!

fedepaol:
> this makes sense to me, thanks for catching this!

It's really the least I can do! Without kind my life would be way more complex :-)

PR #2473, opened 2021-09-22
Validate: check conflicting ipfamily and pods / services subnet

fedepaol:
When choosing a single stack ip family, we must ensure that the pod subnet / service subnet belong to the same family.

Fixes https://github.com/kubernetes-sigs/kind/issues/2468

### DIFF
--- a/pkg/internal/apis/config/validate.go
+++ b/pkg/internal/apis/config/validate.go
@@ -51,14 +51,13 @@ func (c *Cluster) Validate() error {
 		}
 	}
 
-	isDualStack := c.Networking.IPFamily == DualStackFamily
 	// podSubnet should be a valid CIDR
-	if err := validateSubnets(c.Networking.PodSubnet, isDualStack); err != nil {
+	if err := validateSubnets(c.Networking.PodSubnet, c.Networking.IPFamily); err != nil {
 		errs = append(errs, errors.Errorf("invalid pod subnet %v", err))
 	}
 
 	// serviceSubnet should be a valid CIDR
-	if err := validateSubnets(c.Networking.ServiceSubnet, isDualStack); err != nil {
+	if err := validateSubnets(c.Networking.ServiceSubnet, c.Networking.IPFamily); err != nil {
 		errs = append(errs, errors.Errorf("invalid service subnet %v", err))
 	}
 
@@ -140,7 +139,7 @@ func validatePort(port int32) error {
 	return nil
 }
 
-func validateSubnets(subnetStr string, dualstack bool) error {
+func validateSubnets(subnetStr string, ipFamily ClusterIPFamily) error {
 	allErrs := []error{}
 
 	cidrsString := strings.Split(subnetStr, ",")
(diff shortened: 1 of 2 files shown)


pkg/internal/apis/config/validate.go
```diff
@@ -168,6 +168,10 @@ func validateSubnets(subnetStr string, dualstack bool) error {
 	// if not DualStack only one CIDR allowed
 	case !dualstack && len(subnets) > 1:
 		allErrs = append(allErrs, errors.New("only one CIDR allowed for single-stack networking"))
+	case ipFamily == IPv4Family && subnets[0].IP.To4() == nil:
+		allErrs = append(allErrs, errors.New("IPv4 family but IPv6 CIDR"))
+	case ipFamily == IPv6Family && subnets[0].IP.To4() != nil:
+		allErrs = append(allErrs, errors.New("IPv6 family but IPv4 CIDR"))
```
neolit123, line 174: I would reword the errors to:
> expected IPvX CIDR for IPvX family

fedepaol: done, thanks!

pkg/internal/apis/config/validate.go
```diff
@@ -168,6 +168,10 @@ func validateSubnets(subnetStr string, dualstack bool) error {
 	// if not DualStack only one CIDR allowed
 	case !dualstack && len(subnets) > 1:
 		allErrs = append(allErrs, errors.New("only one CIDR allowed for single-stack networking"))
+	case ipFamily == IPv4Family && subnets[0].IP.To4() == nil:
```
aojea, line 171: can subnets be 0?
maybe we should be paranoid and consider that to avoid a panic here

fedepaol: Fair enough, I added a case len == 0 on top of the switch

pkg/internal/apis/config/validate.go
```diff
@@ -153,7 +152,11 @@ func validateSubnets(subnetStr string, dualstack bool) error {
 		subnets = append(subnets, cidr)
 	}
 
+	dualstack := ipFamily == DualStackFamily
 	switch {
+	// if no subnets are defined
+	case len(subnets) == 0:
+		allErrs = append(allErrs, errors.New("no subnets defined"))
```
aojea, line 159: IIRC defaulting is before validation, but in case of having zero here, is this an error or is a "no need validation"  @BenTheElder ?

aojea:
/ok-to-test
just one question about the case with no subnets

pkg/internal/apis/config/validate.go
BenTheElder: we only added this case because https://github.com/kubernetes-sigs/kind/pull/2473/files#r714772513 :-)

IIRC defaulting does come first so maybe we should be throwing a "this can't happen" type error (maybe even panic?) /shrug

aojea: We can have an error and add a test for protecting of future changes that may remove the defaulting, if we panic and we don't test it we never now until someone hits the panic

@fedepaol can you add a test that it fails without subnets with a comment explaining that this should never happen for future us?

fedepaol: Will do

aojea:
/lgtm
/approve
Thank you very much

BenTheElder:
thanks!

Issue #2484, opened 2021-10-07
IPv6 Primary Dual Stack Clusters Error


adobley:
**What happened**:
The `local-path-provisioner` pod crashes after the cluster starts. It is unable to reach the apiserver.

```
level=fatal msg="Error starting daemon: Cannot start Provisioner: failed to get Kubernetes server version: Get https://[fd00:100:64::1]:443/version?timeout=32s: dial tcp [fd00:100:64::1]:443: connect: connection refused"
```
`fd00:100:64::1` is the IP for the service.

**What you expected to happen**:
The pods would be able to reach the kubeapi and would not crash.

**How to reproduce it (as minimally and precisely as possible)**:
Create a kind.yaml with:
```
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
networking:
  ipFamily: dual
  podSubnet: "fd00:100:96::/48,100.96.0.0/11"
  serviceSubnet: "fd00:100:64::/108,100.64.0.0/13"
```

`kind create cluster --config ./kind.yaml`

It will create successfully but given some time the pod will start to crash.

**Anything else we need to know?**:
I believe this is due to the kubeapi getting an IPv4 address for its `advertise-address` even though the first subnet is IPv6, but have not extensively tested that theory yet.

**Environment:**

- kind version: (use `kind version`): `kind v0.11.1 go1.15.2 darwin/amd64`
- Kubernetes version: (use `kubectl version`): 
```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.1+wcp.2", GitCommit:"c78c56047324e817e4bd29b08848a13e3623e263", GitTreeState:"clean", BuildDate:"2020-12-12T17:23:54Z", GoVersion:"go1.15", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.1", GitCommit:"5e58841cce77d4bc13713ad2b91fa0d961e69192", GitTreeState:"clean", BuildDate:"2021-05-21T23:01:33Z", GoVersion:"go1.16.4", Compiler:"gc", Platform:"linux/amd64"}
```
- Docker version: (use `docker info`):
```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Build with BuildKit (Docker Inc., v0.5.1-docker)
[...]

PR #2487, opened 2021-10-07
dual-stack order matters to configure apiserver advertise address

aojea:
Fixes: #2484


aojea:
/hold 

needs testing, who knows if the order breaks more things

PR #2497, opened 2021-10-13
Support getting all kind nodes with -A flag

YuviGold:
Usage example

```bash
➜  kind git:(get-all-nodes) ./bin/kind get nodes
No kind nodes found for cluster "kind".
➜  kind git:(get-all-nodes) ./bin/kind get nodes --help
Lists existing kind nodes by their name

Usage:
  kind get nodes [flags]

Flags:
  -A, --all-clusters     -A, --all-clusters   If present, list all the available nodes across all cluster contexts. Current context is ignored even if specified with --name.
  -h, --help           help for nodes
      --name string    the cluster context name (default "kind")

Global Flags:
      --loglevel string   DEPRECATED: see -v instead
  -q, --quiet             silence all stderr output
  -v, --verbosity int32   info log verbosity, higher value produces more output
➜  kind git:(get-all-nodes) ./bin/kind get nodes -A    
road-to-secure-kubernetes-worker
road-to-secure-kubernetes-control-plane
road-to-secure-kubernetes-worker2
```

### DIFF
--- a/pkg/cmd/kind/get/nodes/nodes.go
+++ b/pkg/cmd/kind/get/nodes/nodes.go
@@ -23,6 +23,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"sigs.k8s.io/kind/pkg/cluster"
+	"sigs.k8s.io/kind/pkg/cluster/nodes"
 	"sigs.k8s.io/kind/pkg/cmd"
 	"sigs.k8s.io/kind/pkg/log"
 
@@ -31,7 +32,8 @@ import (
 )
 
 type flagpole struct {
-	Name string
+	Name        string
+	AllClusters bool
 }
 
 // NewCommand returns a new cobra.Command for getting the list of nodes for a given cluster
@@ -53,6 +55,13 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 		cluster.DefaultName,
 		"the cluster context name",
 	)
+	cmd.Flags().BoolVarP(
+		&flags.AllClusters,
+		"all-clusters",
+		"A",
+		false,
+		"If present, list all the available nodes across all cluster contexts. Current context is ignored even if specified with --name.",
+	)
 	return cmd
 }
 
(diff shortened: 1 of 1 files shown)


YuviGold:
> It should probably be invalid to supply `--name` and `-A`?
> 
> Usage probably needs updating and this case needs handling. We should prefer to match how kubectl handles with with -A and --namespace

@BenTheElder 
Actually, `kubectl` allows providing both flags
https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/get/get.go#L195
I will update the usage according to it, but in case we want to be aligned, then providing both should be allowed while `-A` overrides the other.
```
➜  kind git:(get-all-nodes) ./bin/kind get nodes --name asd -A
road-to-secure-kubernetes-worker
road-to-secure-kubernetes-control-plane
road-to-secure-kubernetes-worker2

➜  kind git:(get-all-nodes) ✗ ./bin/kind get nodes --help
...
Flags:
  -A, --all-clusters   If present, list all the available nodes across all cluster contexts. Current context is ignored even if specified with --name.
...
➜  kind git:(get-all-nodes) kubectl get pods -A -n asd | head
NAMESPACE            NAME                                                              READY   STATUS      RESTARTS   AGE
default              redis-0                                                           1/1     Running     5          23d
default              road-to-secure-kubernetes-7b458649cc-9sv7b                        1/1     Running     5          23d
default              road-to-secure-kubernetes-7b458649cc-gmzcd                        1/1     Running     5          23d
default              road-to-secure-kubernetes-7b458649cc-rxnnm                        1/1     Running     5          23d
ingress-nginx        ingress-nginx-admission-create-79kf7                              0/1     Completed   0          23d
ingress-nginx        ingress-nginx-admission-patch-h6sx5                               0/1     Completed   0          23d
ingress-nginx        ingress-nginx-controller-ddf6c9bfc-t99gk                          1/1     Running     5          23d
kube-system          cilium-node-init-d7hqk                                            1/1     Running     11         23d
kube-system          cilium-node-init-n2g78                                            1/1     Running     11         23d

```

pkg/cluster/internal/providers/provider.go
```diff
@@ -32,6 +32,9 @@ type Provider interface {
 	// ListClusters discovers the clusters that currently have resources
 	// under this providers
 	ListClusters() ([]string, error)
+	// ListAllNodes discovers the nodes that currently have resources
+	// under this providers
+	ListAllNodes() ([]nodes.Node, error)
```
BenTheElder, line 37: I think we should just call list clusters and then call list nodes for each cluster.

Sure it's a bit slower, but this method has a single actual use case, complicating adding more providers for one niche command. Users should not have a super high number of clusters / nodes realistically anyhow, and they can always reach around and use ps themselves.

BenTheElder (commented):
/lgtm
/approve

BenTheElder:
thanks! 😄

PR #123, opened 2018-11-21
add `kind export logs`

BenTheElder:
adds snappily fast native log dumping functionality. we should dump more logs later on, but this should be a decent start. fixes #5 

/kind enhancement

PR #2501, opened 2021-10-14
Fixes issue with overlapping ipv6 addresses when using Docker

marcosnils:
This commit just adds a condition to let kind know to pick another ipv6
address in case docker reports that's ovelapping with any of the system
interfaces

Fixes 2496

Signed-off-by: Marcos Lilljedahl <marcosnils@gmail.com>

### DIFF
--- a/pkg/cluster/internal/providers/docker/network.go
+++ b/pkg/cluster/internal/providers/docker/network.go
@@ -266,7 +266,7 @@ func isIPv6UnavailableError(err error) bool {
 
 func isPoolOverlapError(err error) bool {
 	rerr := exec.RunErrorForError(err)
-	return rerr != nil && strings.HasPrefix(string(rerr.Output), "Error response from daemon: Pool overlaps with other one on this address space")
+	return rerr != nil && strings.HasPrefix(string(rerr.Output), "Error response from daemon: Pool overlaps with other one on this address space") || strings.Contains(string(rerr.Output), "networks have overlapping")
 }
 
 func isNetworkAlreadyExistsError(err error) bool {


aojea:
> > you need to sign the CLA with the email associated to your github account
> 
> I've done this already.. however, seems like the bot didn't pick it up. I've opened a PR that a I had to close, then went ahead and signed the CLA and opened this one. However, seems like the CLA bot never updated the status, is there a way to re-trigger the check? ~Tried `/check-cla` but didn't do anything~ Seems like it picked the CLA now.
> 
> > Also, use the prefix # for the issue number so the bot closes the issue once the PR merges
> 
> Also did this but I had to remove it because the bot complained that I can't use `@` or `#` prefixes in commits. Check the previous PR here (#2500) image

it is working now, the check is correct

BenTheElder:
FWIW you can fix that with `git commit --amend`, update the commit message, `git push -f` (force push).

The bot blocks these because github has unfortunate behavior if you don't: It will ping the issue or user anytime the commit is merged in any fork (!), see for fun an example like: https://github.com/kubernetes/kubernetes/issues/1 🙃 
Even worse: if your commit says something like `fixes kubernetes-sigs/kind#123` it will close the issue repeatedly too 🔥 

Thanks for the PR!

marcosnils:
> FWIW you can fix that with `git commit --amend`, update the commit message, `git push -f` (force push).

Yep, that what I did in the previous PR but the CLA bot was still not updating the label.

BenTheElder:
> Yep, that what I did in the previous PR but the CLA bot was still not updating the label.

hmm, if you see that again that is a bug we should report to the CNCF 😅

PR #2502, opened 2021-10-14
bump default node image

aojea:


aojea:
I have the feeling that podman rootless broke with the new podman 3.4.0 release

AkihiroSuda:
https://github.com/kubernetes-sigs/kind/suites/4057989849/artifacts/103222385

`kind-control-plane/serial.log`:
> WARN: UserNS: SELinux might be Enforcing. If you see an error related to overlayfs, try setting `KIND_EXPERIMENTAL_CONTAINERD_SNAPSHOTTER=fuse-overlayfs` .

`podman-info.txt`: kernel: 5.11.12-300.fc34.x86_64


---

Workarounds:

- Modify the `entrypoint` to use `fuse-overlayfs` by default, when SELinux *seems* to be enabled.
  This is the easiest fix, but decreases performance.

https://github.com/kubernetes-sigs/kind/blob/220bd84b187986aa4d2629cf820a77cdf9a11d0a/images/base/files/usr/local/bin/entrypoint#L82-L88

- Just run the CI with `KIND_EXPERIMENTAL_CONTAINERD_SNAPSHOTTER=fuse-overlayfs` and call it a day

- Just run the CI without SELinux and call it a day.

- Upgrade the kernel to 5.13 or later, which contains fixes for overlayfs on SELinux.
  Fedora 35 will be released soon with kernel 5.14.

Issue #2524, opened 2021-11-03
Control Plane Fails to Start on Fedora 35 with rootless podman


rverenich:
**What happened**:

When using the podman provider, the KinD control plane fails to come up on Fedora 35

**What you expected to happen**:

Control plane to start using rootless Podman

**How to reproduce it (as minimally and precisely as possible)**:

1. Install Fedora 35
2. Install KinD

**Anything else we need to know?**:

configured to comply https://kind.sigs.k8s.io/docs/user/known-issues/#fedora
https://kind.sigs.k8s.io/docs/user/rootless/

tried both drivers in /etc/containers/storage.conf
```
# Default Storage Driver, Must be set for proper operation.
# driver = "overlay"
driver = "btrfs"
```

**Environment:**

- kind version: (use `kind version`):

kind v0.11.1 go1.16.8 linux/amd64
kind v0.12.0-alpha go1.16.8 linux/amd64

- Kubernetes version: (use `kubectl version`):

N/A

- Docker version: (use `docker info`):
```
$ podman info
host:
  arch: amd64
  buildahVersion: 1.23.1
  cgroupControllers:
  - cpu
  - io
  - memory
  - pids
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.30-2.fc35.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.30, commit: '
  cpus: 8
  distribution:
    distribution: fedora
    variant: workstation
    version: "35"
  eventLogger: journald
  hostname: rvx1
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.14.14-300.fc35.x86_64
  linkmode: dynamic
  logDriver: journald
  memFree: 18337759232
  memTotal: 33361170432
  ociRuntime:
    name: crun
    package: crun-1.2-1.fc35.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 1.2
      commit: 4f6c8e0583c679bfee6a899c05ac6b916022561b
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +CRIU +YAJL
  os: linux
  remoteSocket:
[...]

PR #2527, opened 2021-11-04
podman: use devicemapper for some filesystems

aojea:
use the same logic used for docker in podman

### DIFF
--- a/pkg/cluster/internal/providers/podman/util.go
+++ b/pkg/cluster/internal/providers/podman/util.go
@@ -17,6 +17,7 @@ limitations under the License.
 package podman
 
 import (
+	"encoding/json"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
I was hitting https://github.com/kubernetes-sigs/kind/issues/2524 with Rootless Docker 20.10.10 on Fedora 35 with btrfs.

So this PR doesn't seem enough to fix #2524

aojea:
> I was hitting #2524 with Rootless Docker 20.10.10 on Fedora 35 with btrfs.
> 
> So this PR doesn't seem enough to fix #2524

/hold
indeed

aojea:
/hold cancel

I think that at least is worth to have the same logic in both providers

BenTheElder (approved):
/lgtm
/approve

PR #2531, opened 2021-11-10
Bumps kindnetd image and go version

PushkarJ:
To keep things up to date with latest go version and k8s upstream base images:

- Bumps base build image for haproxy to debian-base:buster-v1.9.0
- Bumps base image for kindnetd to debian-iptables:buster-v1.7.0
- Bumps go version for main kind to v1.17.3

/kind cleanup
/area dependency
/assign @BenTheElder 
/sig security testing


BenTheElder:
/lgtm
/approve
/hold

the images will need to be build and pushed to ensure there's no breaking changes, and haproxy manually tested (HA mode is tbqh barely supported still, there's been so little demand, understandably).

images/haproxy/Dockerfile
```diff
@@ -15,7 +15,7 @@
 # This image is a haproxy image + minimal config so the container will not exit
 # while we rewrite the config at runtime and signal haproxy to reload.
 
-ARG BASE="k8s.gcr.io/build-image/debian-base:buster-v1.8.0"
+ARG BASE="k8s.gcr.io/build-image/debian-base:buster-v1.9.0"
```
BenTheElder, line 18: something broke with out image with this change (just built and tested it) so this won't be so simple.

BenTheElder:
```
haproxy: error while loading shared libraries: libcrypt.so.1: cannot open shared object file: No such file or directory
haproxy: error while loading shared libraries: libcrypt.so.1: cannot open shared object file: No such file or directory
```

BenTheElder: probably a bug in `stage-binary-and-deps.sh` or something changed with the env in the base image

BenTheElder:
/approve cancel
/lgtm cancel
due to https://github.com/kubernetes-sigs/kind/pull/2531#discussion_r749737824

images/haproxy/Dockerfile
PushkarJ: @BenTheElder I will try to investigate this but I am close to long vacation next month, so may not get around to providing a fix. Happy for someone else to pick this up if this is urgent.

BenTheElder: no worries! this is going to be a fun one and I think we're OK in the meantime, I think we can circle back to haproxy and just bump the rest for now.

Most of this image is only used at build time anyhow.

PushkarJ: @BenTheElder Thanks! Should I break the PR into two then? i.e. Remove the ha-proxy into a new PR to be investigated and keep the rest here so we can merge that and get it bumped?

BenTheElder: yes please!

PushkarJ: @BenTheElder removed haproxy image update as requested!

BenTheElder:
turns out we have a critical base image change to push shortly so I'm going to roll this up #2551

BenTheElder:
Building those images atm.

Issue #2535, opened 2021-11-14
Could not load docker image in ubuntu in latest kind version


rpiotrow:
**What happened**:
```
$ kind load docker-image 'postgres:13.4' 'postgres:13.4'
Image: "postgres:13.4" with ID "sha256:113197da0347a82925486217beece0386a4de3b4aa3a34e53a20c988606ffa96" not yet present on node "kind-control-plane", loading...
Image: "postgres:13.4" with ID "sha256:113197da0347a82925486217beece0386a4de3b4aa3a34e53a20c988606ffa96" not yet present on node "kind-control-plane", loading...
ERROR: command "docker save -o /tmp/images-tar389269045/images.tar postgres:13.4 postgres:13.4" failed with error: exit status 1
Command Output: failed to save image: invalid output path: directory "/tmp/images-tar389269045" does not exist
```

**What you expected to happen**:
Kind loads docker image without any error.

**How to reproduce it (as minimally and precisely as possible)**:
```
kind create cluster
docker pull postgres:13.4
kind load docker-image 'postgres:13.4' 'postgres:13.4'
```

**Anything else we need to know?**:

**Environment:**

- kind version: (use `kind version`):
```
kind v0.11.1 go1.16.4 linux/amd64
```
- Kubernetes version: (use `kubectl version`):
```
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.3", GitCommit:"c92036820499fedefec0f847e2054d824aea6cd1", GitTreeState:"clean", BuildDate:"2021-10-29T02:41:56Z", GoVersion:"go1.16.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.1", GitCommit:"5e58841cce77d4bc13713ad2b91fa0d961e69192", GitTreeState:"clean", BuildDate:"2021-05-21T23:01:33Z", GoVersion:"go1.16.4", Compiler:"gc", Platform:"linux/amd64"}
```
- Docker version: (use `docker info`):
```
Client:
 Context:    default
 Debug Mode: false

Server:
 Containers: 4
  Running: 1
  Paused: 0
  Stopped: 3
 Images: 41
 Server Version: 20.10.8
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
[...]

PR #2536, opened 2021-11-14
kind load remove duplicate images names

aojea:
Fixes: https://github.com/kubernetes-sigs/kind/issues/2535

### DIFF
--- a/pkg/cmd/kind/load/docker-image/docker-image.go
+++ b/pkg/cmd/kind/load/docker-image/docker-image.go
@@ -203,18 +203,15 @@ func imageID(containerNameOrID string) (string, error) {
 	return lines[0], nil
 }
 
-// remove duplicates from a slice
+// removeDuplicates removes duplicates from a string slice
 func removeDuplicates(slice []string) []string {
 	result := []string{}
-	if len(slice) == 0 {
-		return result
-	}
-	seen := make(map[string]struct{})
-	for _, s := range slice {
-		seen[s] = struct{}{}
-	}
-	for k, _ := range seen {
-		result = append(result, k)
+	seenKeys := make(map[string]struct{})
+	for _, k := range slice {
+		if _, seen := seenKeys[k]; !seen {
+			result = append(result, k)
+			seenKeys[k] = struct{}{}
+		}
 	}
 	return result
 }
--- a/pkg/cmd/kind/load/docker-image/docker-image.go
+++ b/pkg/cmd/kind/load/docker-image/docker-image.go
@@ -82,7 +82,7 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
 	)
 
 	// Check that the image exists locally and gets its ID, if not return error
-	imageNames := args
+	imageNames := removeDuplicates(args)
 	var imageIDs []string
 	for _, imageName := range imageNames {
 		imageID, err := imageID(imageName)
@@ -201,4 +201,20 @@ func imageID(containerNameOrID string) (string, error) {
 		return "", errors.Errorf("Docker image ID should only be one line, got %d lines", len(lines))
 	}
 	return lines[0], nil
+}
+
+// remove duplicates from a slice
+func removeDuplicates(slice []string) []string {
+	result := []string{}
+	if len(slice) == 0 {
+		return result
+	}
+	seen := make(map[string]struct{})
+	for _, s := range slice {
+		seen[s] = struct{}{}
+	}
+	for k, _ := range seen {
+		result = append(result, k)
+	}
+	return result
 }
--- a/pkg/cmd/kind/load/docker-image/docker-image_test.go
+++ b/pkg/cmd/kind/load/docker-image/docker-image_test.go
@@ -1,0 +1,62 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package load
+
+import (
+	"reflect"
+	"sort"
+	"testing"
+)
+
+func Test_removeDuplicates(t *testing.T) {
+	tests := []struct {
+		name  string
+		slice []string
+		want  []string
+	}{
+		{
+			name:  "empty",
+			slice: []string{},
+			want:  []string{},
+		},
+		{
+			name:  "all different",
+			slice: []string{"one", "two"},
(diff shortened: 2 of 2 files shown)


pkg/cmd/kind/load/docker-image/docker-image.go
```diff
@@ -24,6 +24,8 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"k8s.io/apimachinery/pkg/util/sets"
```
BenTheElder, line 27: ughhh we need to get rid of this dependency

BenTheElder (commented):
this is surprising, docker should handle this?
let me do some testing

BenTheElder:
I'm not sure this optimization is necessary, the check for presense is cheap (also the user can just dedupe themselves?) and docker etc. dedupe layers when save/load.

pkg/cmd/kind/load/docker-image/docker-image.go
aojea: > grep -r machinery pkg/ | grep sets
> pkg/build/nodeimage/buildcontext.go:    "k8s.io/apimachinery/pkg/util/sets"
> pkg/cluster/internal/kubeconfig/internal/kubeconfig/paths.go:   "k8s.io/apimachinery/pkg/util/sets"
> pkg/cluster/internal/providers/docker/provider.go:      "k8s.io/apimachinery/pkg/util/sets"
> pkg/cluster/internal/providers/podman/provider.go:      "k8s.io/apimachinery/pkg/util/sets"
> pkg/cluster/internal/providers/common/images_test.go:   "k8s.io/apimachinery/pkg/util/sets"
> pkg/cluster/internal/providers/common/images.go:        "k8s.io/apimachinery/pkg/util/sets"
> pkg/cmd/kind/load/docker-image/docker-image.go: "k8s.io/apimachinery/pkg/util/sets"

aojea:
> I'm not sure this optimization is necessary, the check for presense is cheap (also the user can just dedupe themselves?) and docker etc. dedupe layers when save/load.

the bug is legit and reproducible :shrug:

BenTheElder:
> the bug is legit and reproducible 🤷

it does not reproduce for me? what are you seeing?

we print two messages about loading the image (that just means we detected it wasn't loaded before loading it), we still only load the image once because we make one docker save call and docker dedupes (as does containerd), this seems to be WAI to me. if the user asks to load it twice we check both names.

besides, the user can have two slightly different names for the same image and the same behavior happens. it could be `postgres:13` and `postgres:13.4`.

aojea:
```
 kind load docker-image busybox busybox -v7
Image: "busybox" with ID "sha256:018c9d7b792b4be80095d957533667279843acf9a46c973067c8d1dff31ea8b4" not yet present on node "kind-control-plane", loading...
Image: "busybox" with ID "sha256:018c9d7b792b4be80095d957533667279843acf9a46c973067c8d1dff31ea8b4" not yet present on node "kind-control-plane", loading...
ERROR: failed to load image: command "docker exec --privileged -i kind-control-plane ctr --namespace=k8s.io images import -" failed with error: exit status 1
Command Output: ctr: image "docker.io/library/busybox:1.29": already exists
Stack Trace: 
        /home/aojea/go/pkg/mod/sigs.k8s.io/kind@v0.11.2-0.20210826184821-f8e6aa668edd/pkg/errors/errors.go:59
        /home/aojea/go/pkg/mod/sigs.k8s.io/kind@v0.11.2-0.20210826184821-f8e6aa668edd/pkg/exec/local.go:124
sigs.k8s.io/kind/pkg/cluster/internal/providers/docker.(*nodeCmd).Run
        /home/aojea/go/pkg/mod/sigs.k8s.io/kind@v0.11.2-0.20210826184821-f8e6aa668edd/pkg/cluster/internal/providers/docker/node.go:146
sigs.k8s.io/kind/pkg/cluster/nodeutils.LoadImageArchive
        /home/aojea/go/pkg/mod/sigs.k8s.io/kind@v0.11.2-0.20210826184821-f8e6aa668edd/pkg/cluster/nodeutils/util.go:80
sigs.k8s.io/kind/pkg/cmd/kind/load/docker-image.loadImage
        /home/aojea/go/pkg/mod/sigs.k8s.io/kind@v0.11.2-0.20210826184821-f8e6aa668edd/pkg/cmd/kind/load/docker-image/docker-image.go:181
sigs.k8s.io/kind/pkg/cmd/kind/load/docker-image.runE.func1
        /home/aojea/go/pkg/mod/sigs.k8s.io/kind@v0.11.2-0.20210826184821-f8e6aa668edd/pkg/cmd/kind/load/docker-image/docker-image.go:166
        /home/aojea/go/pkg/mod/sigs.k8s.io/kind@v0.11.2-0.20210826184821-f8e6aa668edd/pkg/errors/concurrent.go:30
runtime.goexit
        /usr/local/go/src/runtime/asm_amd64.s:1581
```

BenTheElder:
https://github.com/kubernetes-sigs/kind/commit/f8e6aa668edd is a bit old, perhaps I can't reproduce because this was fixed in containerd at some point?

pkg/cmd/kind/load/docker-image/docker-image.go
BenTheElder: pkg/internal/sets after #2539

BenTheElder (commented):
needs a litte update after my mega PR, otherwise lgtm

pkg/cmd/kind/load/docker-image/docker-image.go
BenTheElder: (please group internal imports on their own, at the bottom, I had hoped to avoid internal imports in cmd/ but I think we can make an exception here, people re-implementing the cmd/ logic using the kind public APIs can just import k8s.io/apimachinery sets themselves if they need to implement this bit of logic)

aojea:
/hold
there is a bug

aojea:
> /hold there is a bug

/hold cancel
false alarm

BenTheElder (approved):
/lgtm
/approve

pkg/cmd/kind/load/docker-image/docker-image.go
```diff
@@ -202,3 +202,19 @@ func imageID(containerNameOrID string) (string, error) {
...
 }
+
+// remove duplicates from a slice
+func removeDuplicates(slice []string) []string {
+	result := []string{}
+	if len(slice) == 0 {
+		return result
+	}
+	seen := make(map[string]struct{})
+	for _, s := range slice {
```
BenTheElder, line 213: 1) the loop below is not go formatted properly (should be k := ...)
2) since we're rolling our own:
```go
func removeDuplicates(slice []string) []string {
	result := []string{}
	seenKeys := make(map[string]struct{})
	for _, s := range slice {
		if _, seen := seenKeys[s]; !seen {
			result = append(result, s)
			seenKeys[s] = struct{}{}
		}
	}
	return result
}
```
(faster with one loop, simpler, no need to optimize for empty input, we require an argument)

BenTheElder:
s390x core dumped https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/2536/pull-kind-build/1461135516324859904

PR #2539, opened 2021-11-15
assorted overdue cleanup

BenTheElder:


pkg/cluster/internal/providers/common/provision.go
```diff
@@ -80,18 +79,10 @@ func RunContainer(engine, name string, args []string, opts ...RunContainerOpt) e
 		logCtx := context.Background()
 		logCtx, logCancel := context.WithTimeout(logCtx, 30*time.Second)
 		defer logCancel()
-		// use os/exec.CommandContext directly, as kind/pkg/exec.CommandContext lacks support for killing
```
BenTheElder, line 83: NOTE: ctx.Cancel() for a CommandContext already kills the process. We don't need to call killl directly, we should leverage cancelling the context instead (and we do).

go.mod
```diff
@@ -12,7 +12,5 @@ require (
 	github.com/spf13/cobra v1.2.1
 	github.com/spf13/pflag v1.0.5
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
-	// TODO: drop this requirement
-	k8s.io/apimachinery v0.22.3
```
aojea, line 16: :man_dancing:

pkg/cluster/internal/providers/common/provision.go
```diff
@@ -1,114 +0,0 @@
...
-// WithWaitUntilSystemdReachesMultiUserSystem waits until the systemd in the container
-// reaches "Multi-User System" target if is using cgroups v2, so that `docker exec` can be
-// executed safely without breaking cgroup v2 hierarchy.
-//
-// This is implemented by grepping `docker logs` with "Reached target .*Multi-User System.*"
-// message from systemd.
-// (we can't use `docker exec` to check whether we are allowed to run `docker exec`, obviously.)
-//
-// Needed for avoiding "ERROR: this script needs /sys/fs/cgroup/cgroup.procs to be empty (for writing the top-level cgroup.subtree_control)"
-// See https://github.com/kubernetes-sigs/kind/issues/2409
```
aojea, line 54: we should keep the comment

BenTheElder: I renamed this to be even clearer about the purpose and commented this in detail in the latest commit.

BenTheElder:
Test flake https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/2539/pull-kind-conformance-parallel-ipv6/1461044809455112192

> Kubernetes e2e suite: [sig-node] Events should be sent by kubelets and the scheduler about pods scheduling and running [Conformance]

TIL we let an events based test get into conformance 🤨 

/override pull-kind-conformance-parallel-ipv6

PR #2546, opened 2021-11-19
set --snapshotter when loading image archives

### LINKED ISSUE #2543: `kind load` fails with "failed to extract layer" on debian bookworm with rootless docker
Split off from #2541. Has some similarity to #2402.

**What happened**: `kind load` fails with error. See trace below. Image still shows up in internal containerd despite error (`docker exec -it kind-control-plane crictl images`)

**What you expected to happen**: Image loaded into Kind successfully.

1. Create default cluster with `kind create cluster`.
2. Attempt to load image archive using `kind load image-archive` OR load image into Docker and use `kind load docker-image`

* Sample run with trace:
```
$ kind load image-archive bazel-container-bin/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/server/async_computation_control_server_image.tar -v 1
ERROR: failed to load image: command "docker exec --privileged -i kind-control-plane ctr --namespace=k8s.io images import -" failed with error: exit status 1

Command Output: unpacking docker.io/bazel/src/main/kotlin/org/wfanet/measurement/duchy/deploy/common/server:async_computation_control_server_image (sha256:72adfdb270f0a3a69b425fe725c3028ec72b1d6227046ce31f2ff3baf0f2e443)...time="2021-11-17T19:11:25Z" level=info msg="apply failure, attempting cleanup" error="failed to extract layer sha256:e3d24823466584f8fcc6fe0d7d5bb09d2360b86ef9b655e21e8002ee04dd386b: failed to mount /var/lib/containerd/tmpmounts/containerd-mount562134243: operation not permitted: unknown" key="extract-583305635-lKcw sha256:5c30f10f05a6636f4b4a9fe393b48215dd35657cf972784e74027e488f9cbd0f"
ctr: failed to extract layer sha256:e3d24823466584f8fcc6fe0d7d5bb09d2360b86ef9b655e21e8002ee04dd386b: failed to mount /var/lib/containerd/tmpmounts/containerd-mount562134243: operation not permitted: unknown

Stack Trace: 
	/usr/local/google/home/sanjayvas/go/pkg/mod/sigs.k8s.io/kind@v0.11.2-0.20211115230152-60143cb4e28c/pkg/errors/errors.go:59
	/usr/local/google/home/sanjayvas/go/pkg/mod/sigs.k8s.io/kind@v0.11.2-0.20211115230152-60143cb4e28c/pkg/exec/local.go:124
sigs.k8s.io/kind/pkg/cluster/internal/providers/docker.(*nodeCmd).Run
	/usr/local/google/home/sanjayvas/go/pkg/mod/sigs.k8s.io/kind@v0.11.2-0.20211115230152-60143cb4e28c/pkg/cluster/internal/providers/docker/node.go:146
sigs.k8s.io/kind/pkg/cluster/nodeutils.LoadImageArchive
	/usr/local/google/home/sanjayvas/go/pkg/mod/sigs.k8s.io/kind@v0.11.2-0.20211115230152-60143cb4e28c/pkg/cluster/nodeutils/util.go:80
sigs.k8s.io/kind/pkg/cmd/kind/load/image-archive.loadImage
	/usr/local/google/home/sanjayvas/go/pkg/mod/sigs.k8s.io/kind@v0.11.2-0.20211115230152-60143cb4e28c/pkg/cmd/kind/load/image-archive/image-archive.go:136
sigs.k8s.io/kind/pkg/cmd/kind/load/image-archive.runE.func1
	/usr/local/google/home/sanjayvas/go/pkg/mod/sigs.k8s.io/kind@v0.11.2-0.20211115230152-60143cb4e28c/pkg/cmd/kind/load/image-archive/image-archive.go:123
	/usr/local/google/home/sanjayvas/go/pkg/mod/sigs.k8s.io/kind@v0.11.2-0.20211115230152-60143cb4e28c/pkg/errors/concurrent.go:30
runtime.goexit
	/usr/lib/google-golang/src/runtime/asm_amd64.s:1581
```

* containerd log: [containerd.log](https://github.com/kubernetes-sigs/kind/files/7557357/containerd.log)

**Environment:**

- kind version: (use `kind version`): v0.12.0-alpha devel +b7a85e0003 linux/amd64
```
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.3", GitCommit:"c92036820499fedefec0f847e2054d824aea6cd1", GitTreeState:"clean", BuildDate:"2021-10-27T18:41:28Z", GoVersion:"go1.16.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.2", GitCommit:"8b5a19147530eaac9476b0ab82980b4088bbc1b2", GitTreeState:"clean", BuildDate:"2021-11-04T09:36:56Z", GoVersion:"go1.16.8", Compiler:"gc", Platform:"linux/amd64"}
```
- Docker version: (use `docker info`): [edited to remove some unnecessary/potentially sensitive info]
```
Client:
 Context:    default
 Debug Mode: false

Server:
 Containers: 1
  Running: 1
  Paused: 0
  Stopped: 0
 Images: 2
 Server Version: 20.10.10
 Storage Driver: fuse-overlayfs
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 5b46e404f6b9f661a205e28d59c982d3634148f8
 runc version: v1.0.2-0-g52b36a2d
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
  rootless
  cgroupns
 Kernel Version: 5.10 amd64
 Operating System: Debian GNU/Linux
 OSType: linux
 Architecture: x86_64
 CPUs: 12
 Total Memory: 62.41GiB
 Debug Mode: false
 Username: sanjayvas
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
```
- OS (e.g. from `/etc/os-release`): Debian Bookworm

BenTheElder:
When loading images, get the current snapshotter in user in CRI by parsing `containerd config dump` and then pass that to `ctr images import`.

Doing it this way ensures even if say the user uses containerdConfigPatches or otherwise edits the containerd config at runtime, we will always load into the currently configured snapshotter.

fixes #2543

### DIFF
--- a/pkg/cluster/nodeutils/util.go
+++ b/pkg/cluster/nodeutils/util.go
@@ -23,6 +23,8 @@ import (
 	"path"
 	"strings"
 
+	"github.com/pelletier/go-toml"
+
 	"sigs.k8s.io/kind/pkg/cluster/nodes"
 	"sigs.k8s.io/kind/pkg/errors"
 	"sigs.k8s.io/kind/pkg/exec"
@@ -76,11 +78,35 @@ func CopyNodeToNode(a, b nodes.Node, file string) error {
 
 // LoadImageArchive loads image onto the node, where image is a Reader over an image archive
 func LoadImageArchive(n nodes.Node, image io.Reader) error {
-	cmd := n.Command("ctr", "--namespace=k8s.io", "images", "import", "-").SetStdin(image)
+	snapshotter, err := getSnapshotter(n)
+	if err != nil {
+		return err
+	}
+	cmd := n.Command("ctr", "--namespace=k8s.io", "images", "import", "--snapshotter", snapshotter, "-").SetStdin(image)
 	if err := cmd.Run(); err != nil {
 		return errors.Wrap(err, "failed to load image")
 	}
 	return nil
+}
+
+func getSnapshotter(n nodes.Node) (string, error) {
+	out, err := exec.Output(n.Command("containerd", "config", "dump"))
+	if err != nil {
+		return "", errors.Wrap(err, "failed to detect containerd snapshotter")
+	}
+	return parseSnapshotter(string(out))
+}
+
+func parseSnapshotter(config string) (string, error) {
+	parsed, err := toml.Load(config)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to detect containerd snapshotter")
+	}
+	snapshotter, ok := parsed.GetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "containerd", "snapshotter"}).(string)
+	if !ok {
+		return "", errors.New("failed to detect containerd snapshotter")
+	}
+	return snapshotter, nil
 }
 
 // ImageID returns ID of image on the node with the given image name if present
--- a/pkg/cluster/nodeutils/util_test.go
+++ b/pkg/cluster/nodeutils/util_test.go
@@ -1,0 +1,264 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package nodeutils
+
+import (
+	"testing"
+)
+
+func TestParseSnapshotter(t *testing.T) {
+	// a real, known kind node containerd config
+	config := `disabled_plugins = []
+	imports = ["/etc/containerd/config.toml"]
+	oom_score = 0
+	plugin_dir = ""
+	required_plugins = []
+	root = "/var/lib/containerd"
+	state = "/run/containerd"
+	version = 2
+	
+	[cgroup]
+	  path = ""
+	
+	[debug]
+	  address = ""
+	  format = ""
+	  gid = 0
+	  level = ""
+	  uid = 0
+	
+	[grpc]
+	  address = "/run/containerd/containerd.sock"
+	  gid = 0
+	  max_recv_message_size = 16777216
+	  max_send_message_size = 16777216
+	  tcp_address = ""
+	  tcp_tls_cert = ""
+	  tcp_tls_key = ""
+	  uid = 0
+	
+	[metrics]
+	  address = ""
+	  grpc_histogram = false
+	
+	[plugins]
+	
+	  [plugins."io.containerd.gc.v1.scheduler"]
+		deletion_threshold = 0
+		mutation_threshold = 100
+		pause_threshold = 0.02
+		schedule_delay = "0s"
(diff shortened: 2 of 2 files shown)


pkg/cluster/nodeutils/util.go
```diff
@@ -76,13 +78,37 @@ func CopyNodeToNode(a, b nodes.Node, file string) error {
...
+	}
+	return parseSnapshotter(string(out))
+}
+
+func parseSnapshotter(config string) (string, error) {
+	parsed, err := toml.Load(config)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to detect containerd snapshotter")
+	}
+	snapshotter, ok := parsed.GetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "containerd", "snapshotter"}).(string)
```
BenTheElder, line 105: worth noting: `io.containerd.grpc.v1.cri` is a key in containerd config, *not* a dot delimited path, this doesn't play well with `Get` which just splits on `.`, so we need to use `GetPath` and write out each key (which, is fine)

we'll need to be careful of this if we ever switch libraries, but we have a unit test here with an actual known config.

go-toml has a v2 now, which is faster, but only does marshal/unmarshal, at some point we should figure out what maintained toml library to standardize on and rework the toml patching code we have as well ... but that's not for today 🙃

amwat (commented):
/lgtm

pkg/cluster/nodeutils/util.go
```diff
@@ -76,13 +78,37 @@ func CopyNodeToNode(a, b nodes.Node, file string) error {
 
 // LoadImageArchive loads image onto the node, where image is a Reader over an image archive
 func LoadImageArchive(n nodes.Node, image io.Reader) error {
-	cmd := n.Command("ctr", "--namespace=k8s.io", "images", "import", "-").SetStdin(image)
+	snapshotter, err := getSnapshotter(n)
+	if err != nil {
+		return err
+	}
+	cmd := n.Command("ctr", "--namespace=k8s.io", "images", "import", "--snapshotter", snapshotter, "-").SetStdin(image)
```
aojea, line 85: I think  it will be more robust if we don't use the snaphotter flag if we fail to find it or there is an error parsing the conffig file, instead of returning

BenTheElder: We fail if we can't read the snapshotter because that shouldn't happen and that's an error that will cause subtle issues if we don't pass the snapshotter.

aojea: well, my concern is that before we didn't and I really don't know if is valid not having a snapshotter defined :smile:

BenTheElder: It's valid but you get the default and if the CRI is not on the default this is a problem and we get into a confusing failure mode 🙃

Not being able to detect the snapshotter either means we failed to exec, and therefore load will fail anyhow, or parsing is broken (which shouldn't happen anytime soon, the containerd config is versioned and kind explicitly used V2 for some time now). In the future this method might need to handle multiple versions of config but we can parse that first and then select the correct field.

PR #2547, opened 2021-11-19
upgrade to containerd 1.5.8

BenTheElder:
there's a stable tag with the fixes we needed now, upgrading to that.


aojea:
/hold
1.22.4 ???

pkg/apis/config/defaults/image.go
```diff
@@ -18,4 +18,4 @@ limitations under the License.
 package defaults
 
 // Image is the default for the Config.Image field, aka the default node image.
-const Image = "kindest/node:v1.22.2@sha256:9af3ab3e36fb59890b2fb3a18000930b4792d62d10d2060e0ca701e2c392d487"
+const Image = "kindest/node:v1.22.4@sha256:9871c0646a7c1d82d3a2c0d38a64ab5985b25a7716985d2ca116251efe456be8"
```
aojea, line 21: 1.22.3?

aojea: never mind https://github.com/kubernetes/kubernetes/releases/tag/v1.22.4

aojea:
> /hold 1.22.4 ???

/hold cancel

https://github.com/kubernetes/kubernetes/releases/tag/v1.22.4

pkg/apis/config/defaults/image.go
BenTheElder: Heh, I built the packages for this yesterday to be fair 🙃

Issue #2551, opened 2021-11-22
Pods with non-standard PATH can fail to start


jeffmhastings:
**What happened**:

Pods with a non-standard `PATH` can fail to start with errors from `kubelet`. `kubectl describe pod` shows an error creating the OCI runtime in `/usr/local/bin/mount-product-files`.

```
Error: failed to create containerd task: failed to create shim: OCI runtime create failed: container_linux.go:380: starting container process caused: process_linux.go:545: container init caused: Running hook #0:: error running hook: exit status 127, stdout: , stderr: /usr/local/bin/mount-product-files: line 30: jq: command not found: unknown
```

**What you expected to happen**:

Container `PATH` shouldn't affect the ability of kubelet to start the Pod.

**How to reproduce it (as minimally and precisely as possible)**:

```
kind create cluster --image=kindest/node:v1.22.4@sha256:9871c0646a7c1d82d3a2c0d38a64ab5985b25a7716985d2ca116251efe456be8
kubectl run nginx --image=nginx --env PATH=/usr/sbin
kubectl get pod nginx -o jsonpath='{.status.containerStatuses[0].lastState.terminated}'
```

Example output:

```
# kind create cluster --image=kindest/node:v1.22.4@sha256:9871c0646a7c1d82d3a2c0d38a64ab5985b25a7716985d2ca116251efe456be8

Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.22.4) 🖼
 ✓ Preparing nodes 📦
 ✓ Writing configuration 📜
 ✓ Starting control-plane 🕹️
 ✓ Installing CNI 🔌
 ✓ Installing StorageClass 💾
Set kubectl context to "kind-kind"
You can now use your cluster with:

kubectl cluster-info --context kind-kind

Not sure what to do next? 😅  Check out https://kind.sigs.k8s.io/docs/user/quick-start/

# kubectl run nginx --image=nginx --env PATH=/usr/sbin

pod/nginx created

# kubectl get pod nginx -o jsonpath='{.status.containerStatuses[0].lastState.terminated}'

[...]

PR #2553, opened 2021-11-23
fix mount-product-files, bump deps / images

BenTheElder:
picks up #2531, fix for #2551


images/base/files/usr/local/bin/mount-product-files
```diff
@@ -22,22 +22,33 @@ set -o errexit
...
+
 # The bundle represents the dir path to container filesystem, container runtime state [1] is
 # passed to the hook's stdin
 #
 # [1] https://github.com/opencontainers/runtime-spec/blob/master/runtime.md#state
 #
 bundle=$(jq -r .bundle)
 
-cp /kind/product_* "$bundle/rootfs/"
+cp /kind/product_* "${bundle:?}/rootfs/"
```
BenTheElder, line 43: inconsistent quote location here but /shrug

at some point I will find a linter for this and normalize the whole repo

aojea: TIL
```
echo ${test:?}
bash: test: parameter null or not set
```

PR #2558, opened 2021-11-29
fix dual-stack order detection

aojea:
the logic was parsing a CIDR as an IP, hence always detecting the
order as IPv6 - IPv4.

https://go.dev/play/p/50xM3zCrWDe

Fixes: https://github.com/kubernetes/kubernetes/issues/106709

### DIFF
--- a/pkg/cluster/internal/create/actions/config/config.go
+++ b/pkg/cluster/internal/create/actions/config/config.go
@@ -196,7 +196,11 @@ func getKubeadmConfig(cfg *config.Cluster, data kubeadm.ConfigData, node nodes.N
 			// order matters since the nodeAddress will be used later to configure the apiserver advertise address
 			// Ref: #2484
 			primaryServiceSubnet := strings.Split(cfg.Networking.ServiceSubnet, ",")[0]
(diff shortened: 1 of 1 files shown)


pkg/cluster/internal/create/actions/config/config.go
```diff
@@ -196,7 +196,11 @@ func getKubeadmConfig(cfg *config.Cluster, data kubeadm.ConfigData, node nodes.N
 			// order matters since the nodeAddress will be used later to configure the apiserver advertise address
 			// Ref: #2484
 			primaryServiceSubnet := strings.Split(cfg.Networking.ServiceSubnet, ",")[0]
```
BenTheElder, line 198: we should probably split this out into a helper w/ tests at some point

BenTheElder (approved):
/lgtm
/approve

pkg/cluster/internal/create/actions/config/config.go
aojea: yeah, I started to do that, and ended moving too much code for something I'd like to get working fast XD

BenTheElder:
prow had an outage I think?
/lgtm
/approve

PR #2525, opened 2021-11-04
rootless: set LocalStorageCapacityIsolation=false to avoid "could not find device..." error

AkihiroSuda:
For avoiding `err="failed to get rootfs info: failed to get device for dir \"/var/lib/kubelet\": could not find device with major: 0, minor: 41 in cached partitions map"`

Fix #2524

See https://github.com/kubernetes/kubernetes/blob/v1.22.2/pkg/kubelet/cm/container_manager_linux.go#L641-L649

---

Tested with Rootless Docker 20.10.10 on Fedora 35

PR #2559, opened 2021-12-02
CI: update Fedora to 35

AkihiroSuda:
Contains
- https://github.com/kubernetes-sigs/kind/pull/2525
for passing CI

### DIFF
--- a/pkg/cluster/internal/kubeadm/config.go
+++ b/pkg/cluster/internal/kubeadm/config.go
@@ -448,6 +448,10 @@ func Config(data ConfigData) (config string, err error) {
 			return "", errors.Errorf("version %q is not compatible with rootless provider (hint: kind v0.11.x may work with this version)", ver)
 		}
 		data.FeatureGates["KubeletInUserNamespace"] = true
+
+		// For avoiding err="failed to get rootfs info: failed to get device for dir \"/var/lib/kubelet\": could not find device with major: 0, minor: 41 in cached partitions map"
+		// https://github.com/kubernetes-sigs/kind/issues/2524
+		data.FeatureGates["LocalStorageCapacityIsolation"] = false
 	}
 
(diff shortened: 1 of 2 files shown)


aojea:
does this mean that most fedora users has to enable this feature gate for podman rootless?
https://github.com/kubernetes-sigs/kind/issues/2524#issuecomment-965642432

AkihiroSuda:
> does this mean that most fedora users has to enable this feature gate for podman rootless?

Yes, probably

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -448,6 +448,10 @@ func Config(data ConfigData) (config string, err error) {
 			return "", errors.Errorf("version %q is not compatible with rootless provider (hint: kind v0.11.x may work with this version)", ver)
 		}
 		data.FeatureGates["KubeletInUserNamespace"] = true
+
+		// For avoiding err="failed to get rootfs info: failed to get device for dir \"/var/lib/kubelet\": could not find device with major: 0, minor: 41 in cached partitions map"
+		// https://github.com/kubernetes-sigs/kind/issues/2524
+		data.FeatureGates["LocalStorageCapacityIsolation"] = false
```
aojea, line 454: docker too?

AkihiroSuda: What do you mean?

aojea: sorry, I mean if this is a general problem for rootless or podman and rootless specific

AkihiroSuda: General for all rootless containers

aojea:
/lgtm
@BenTheElder for approval, I can't find a better solution than this, no ideal , but at least is constrained to rootless that we mark as experimental

aojea:
/approve
This is contained under a flag for rootless

PR #648, opened 2019-06-24
[WIP] Test PR DO NOT MERGE

BenTheElder:
/hold

PR #2561, opened 2021-12-02
[WIP] Fedora 35 debug

BenTheElder:
would prefer to avoid #2525 if we can help it, debugging the fedora 35 issue

### DIFF
--- a/.github/workflows/cgroup2.yaml
+++ b/.github/workflows/cgroup2.yaml
@@ -52,6 +52,8 @@ jobs:
           "$HELPER" dockerd-rootless-setuptool.sh install
           # Modify the client config to use the rootless daemon by default
           "$HELPER" docker context use rootless
+          # TEMP: debug
+          "$HELPER" ls /dev/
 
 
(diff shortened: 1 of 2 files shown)


aojea:
have you found something?

stmcginnis:
Is this still needed with https://github.com/kubernetes-sigs/kind/pull/2559?

BenTheElder:
I haven't had time recently, I've been out on bereavement. I would still like to finish getting an understanding of the fedora 35 default filesystem and what we can do to handle the actual root cause instead of enabling a non-standard feature gate and disabling a core feature.

It is important for kind as a project to test default-kubernetes, if in the future CI wanted to go rootless then we have a problem. Plus if we try to change something like this back, it can be a breaking change for users.

This PR is not going to merge, it is just for debugging, like #648, [WIP] prevents the bot from merging it.

BenTheElder:
I still think this is something we need to fix someday so we can use rootless to test kubernetes itself, but I just don't have the time to debug this currently.

PR #2568, opened 2021-12-09
refactor: remove deprecated flag --address from scheduler config

knight42:
Signed-off-by: Jian Zeng <anonymousknight96@gmail.com>

The flag `--address` in `kube-scheduler` has been deprecated and will be removed.

### DIFF
--- a/pkg/cluster/internal/kubeadm/config.go
+++ b/pkg/cluster/internal/kubeadm/config.go
@@ -326,7 +326,6 @@ scheduler:
 {{ end }}
     # configure ipv6 default addresses for IPv6 clusters
     {{ if .IPv6 -}}
-    address: "::"
     bind-address: "::1"
     {{- end }}
 networking:


aojea:
these are template for the kubeadm API, is it a direct match against the flag @neolit123 ?

bear in mind we have to keep compatibility with older versions of kubernetes, if kubeadm is doing the conversion in their API we shouldn't have to touch anything, right?

knight42:
Hi @aojea, as [the comment](https://github.com/kubernetes/kubernetes/blob/ba200841fddcf8fb5edb413546ce9574f2a14d82/cmd/kubeadm/app/apis/kubeadm/types.go#L138-L140) says:
> // ExtraArgs is an extra set of flags to pass to the control plane component.
	// A key in this map is the flag name as it appears on the
	// command line except without leading dash(es).

These args are simply the flags passed to the scheduler. IIUC kubeadm is not touching this field during different versions, but I would like to cc @neolit123 for confirmation.

BenTheElder:
I don't recall why we were setting this *and* bind-address, but this change should only be in `ConfigTemplateBetaV2` and not `ConfigTemplateBetaV1`, as we auto-select the config template based on the Kubernetes version and this change only applies to recent versions, so we should update the most recent template only.

knight42:
> I don't recall why we were setting this _and_ bind-address, but this change should only be in `ConfigTemplateBetaV2` and not `ConfigTemplateBetaV1`, as we auto-select the config template based on the Kubernetes version and this change only applies to recent versions, so we should update the most recent template only.

@BenTheElder Hi! I have reverted the change of `ConfigTemplateBetaV1`, PTAL.

jonyhy96:
ping @neolit123 @aojea 
Is this ok to merge? This cause block on pr https://github.com/kubernetes/kubernetes/pull/106865

BenTheElder:
Sorry I have been out and prioritizing things assigned to me when catching up intermittently.

Thinking over this now ... this may still be a breaking change for users of the insecure port for kubernetes versions new enough to use this template but old enough to still have the flag.

We could switch to the latest kubeadm API only for this upcoming kubernetes version and enable it there 🤔, as kind has not yet switched to kubeadm v1beta3.

Or we can just add some conditional to the template input based on kubernetes version.

BenTheElder:
also unrelated to this PR: but we get to "cheat" in hack/cluster-up.sh, in kubernetes, this sort of problem is going to get worse if we ever switch to only-out-of-tree provisioners .... 🤔

jonyhy96:
What if we add a template function called compareVersion to determine whether should add some args base on k8s version?

```golang
func getTemplate() *template.Template {
	return template.New("kubeadm-config").Funcs(template.FuncMap{
		// -1 when version1 is less than version2
		// 0 when version1 is equal to version2
		// 1 when version1 is larger than version2
		"compareVersion": func(version1, version2 string) int {
			result, err := version.MustParseSemantic(version1).Compare(version2)
			if err != nil {
				panic(err)
			}
			return result
		},
	})
}
```
could use 
```text
    {{ $result := compareVersion .KubernetesVersion "v1.23.0" }}
    # when k8s version is less than v1.23.0, insecure config could be used.
    {{ if ( eq $result -1 ) }} 
    address: "::"
    {{- end }}
```

aojea:
> I don't recall why we were setting this _and_ bind-address, but this change should only be in `ConfigTemplateBetaV2` and not `ConfigTemplateBetaV1`, as we auto-select the config template based on the Kubernetes version and this change only applies to recent versions, so we should update the most recent template only.

because those were needed for ipv6 only clusters

We had to use address because kubeadm used that option to choose the probe address, but I didn't remember I fixed that :sweat_smile:  and now use the `bind-address` field

https://github.com/kubernetes/kubernetes/pull/86493

This change is correct and address is no needed anymore, CI seems to be happy on all versions too

/lgtm
/approve

BenTheElder:
thanks all!

PR #2578, opened 2022-01-05
Fix start in container

msciabarra:
If you try to run kind in a containerized environment (with docker access), like those generated by vs-code, it fails because of timing problems.

This patch fixes the problem.

### DIFF
--- a/pkg/cluster/internal/create/actions/waitforready/waitforready.go
+++ b/pkg/cluster/internal/create/actions/waitforready/waitforready.go
@@ -97,7 +97,7 @@ func waitForReady(node nodes.Node, until time.Time) bool {
 		if err != nil {
 			return false
 		}
-		//fmt.Println(lines)
+		// If the server is not ready the answer is empty and this situation must be taken in account
 		if len(lines) == 0 {
 			return false
 		}
--- a/pkg/cluster/internal/create/actions/waitforready/waitforready.go
+++ b/pkg/cluster/internal/create/actions/waitforready/waitforready.go
@@ -97,6 +97,10 @@ func waitForReady(node nodes.Node, until time.Time) bool {
 		if err != nil {
 			return false
 		}
+		//fmt.Println(lines)
+		if len(lines) == 0 {
+			return false
+		}
 
 		// 'lines' will return the status of all nodes labeled as master. For
 		// example, if we have three control plane nodes, and all are ready,
--- a/pkg/cluster/internal/create/actions/kubeadminit/init.go
+++ b/pkg/cluster/internal/create/actions/kubeadminit/init.go
@@ -18,7 +18,9 @@ limitations under the License.
 package kubeadminit
 
 import (
+	"fmt"
 	"strings"
+	"time"
 
 	"sigs.k8s.io/kind/pkg/errors"
 	"sigs.k8s.io/kind/pkg/exec"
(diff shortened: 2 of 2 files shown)


msciabarra:
I signed the CLA

pkg/cluster/internal/create/actions/waitforready/waitforready.go
```diff
@@ -97,6 +97,10 @@ func waitForReady(node nodes.Node, until time.Time) bool {
 		if err != nil {
 			return false
 		}
+		// If the server is not ready the answer is empty and this situation must be taken in account
```
BenTheElder, line 100: The API Server *must* be ready by the point this call is used. This is a prerequisite of so many other things.

BenTheElder (commented):
What kubernetes version are you seeing this on?

`kubeadm init` shouldn't be exiting successfully with an unready API-server, because it waits for kubelet to be healthy which requires registering with the API-server.

pkg/cluster/internal/create/actions/waitforready/waitforready.go
msciabarra: The problem it is not. Or at least even if it is ready I receive sometimes when I run it an empty command Output to the subsequent lines[0] gives an index out of bound exception.

BenTheElder: Hmm, can you reply to https://github.com/kubernetes-sigs/kind/pull/2578#pullrequestreview-844900787?

I mean, with an unready api-server we also can't do things like install the CNI / storage daemonsets, remove the taint (this one you appear to have added a special wait to), etc.

If the api-server is unhealthy in the environment, that's a problem, but it *shouldn't* be happening once we reach waitForReady here, where we're waiting for node readiness and have long since been relying on api-server readiness. I don't think we're prepared to remove that assumption nor should we need to. The wait before removing the taint might be reasonable but smells like a kubeadm bug that we should probably be fixing upstream https://github.com/kubernetes-sigs/kind/pull/2578#pullrequestreview-844900787

msciabarra (commented):
You can see the problem I am trying to solve this way:
```
docker run -d --name=demokindbug --mount=source=/var/run/docker.sock,target=/var/run/docker-host.sock,type=bind ghcr.io/nuvolaris/nuvolaris-devkit:neo-22.0103.09
docker exec -ti demokindbug /home/nuvolaris/.local/bin/kind create cluster
```

From the output it appears to be running 1.21.1

This is exactly what I see:

```
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.21.1) 🖼
 ✓ Preparing nodes 📦
 ✓ Writing configuration 📜
 ✗ Starting control-plane 🕹️
ERROR: failed to create cluster: failed to remove master taint: command "docker exec --privileged kind-control-plane kubectl --kubeconfig=/etc/kubernetes/admin.conf taint nodes --all node-role.kubernetes.io/master-" failed with error: exit status 1
Command Output: The connection to the server kind-control-plane:6443 was refused - did you specify the right host or port?                               exit:1
```

Note however that in this environment (the one created VSCcode) you access docker.sock through a proxy to give access to it to non-root users.

msciabarra:
I am closing this PR as I figured out the problem is actually in the socat proxy that VSCode creates to let use docker from non-root users. If I run kind with
```
export DOCKER_HOST=/var/run/docker-host.sock
sudo kind create cluster
```
it works.
And I believe it is a better solution than patching kind.

BenTheElder:
Interesting! Perhaps this was making the connection flaky?

This might be worth documenting somewhere 👀 , perhaps https://kind.sigs.k8s.io/docs/user/known-issues/

yongxiu:
>  kubeadm init shouldn't be exiting successfully with an unready API-server, because it waits for kubelet to be healthy which requires registering with the API-server.

Based on the tests, seems like that's not the case? I also hit this issue.

BenTheElder:
> Based on the tests, seems like that's not the case?

Which tests?

> I also hit this issue.

Please file a bug report with all the details asked in the bug template. We'll most likely need the cluster's logs (`kind export logs`), it could be a networking issue on the host, a resource exhaustion issue on the host, or some bug in kubelet, api-server, etcd ...

PR #2582, opened 2022-01-08
upgrade to kubeadm v1beta3 config format

BenTheElder:
As far as I can tell all fields we use are compatible.

This will be a required upgrade in kubernetes 1.24.

Arbitrarily cutting this off at 1.23 for now, but I think 1.22 even has support.

We should revisit the cutoff and leveraging new functionality cutting off earlier / later has a tradeoff (breaking patches potentially / removed + added fields ... compatibility with kind users). This is a straightforward migration from v1beta2 currently for k8s 1.23.0+

There are no actual config changes. all fields we use should be compatible.

https://kubernetes.io/docs/reference/config-api/kubeadm-config.v1beta3/

https://groups.google.com/g/kubernetes-sig-cluster-lifecycle/c/XJhn8bpvHac/m/ZP1sN_20BwAJ


pkg/cluster/internal/kubeadm/config.go
```diff
@@ -427,6 +427,147 @@ conntrack:
 {{end}}{{end}}
 `
 
+// ConfigTemplateBetaV2 is the kubeadm config template for API version v1beta2
+const ConfigTemplateBetaV3 = `# config generated by kind
+apiVersion: kubeadm.k8s.io/v1beta2
```
neolit123, line 432: i guess, apiVersion and code comments should be for V3 / v1beta3.

neolit123:
> There are no actual config changes. all fields we use should be compatible.

sounds accurate.

pkg/cluster/internal/kubeadm/config.go
BenTheElder: LOL sigh, fixed 🙃

Issue #2411, opened 2021-08-11
Can not create a cluster when running on BTRFS + LUKS encryption


bergmannf:
**What happened**:

When starting a kind cluster on an encrypted `btrfs` root partition the `control-plane` won't start up, because of an error in the `kubelet`:

```
Aug 11 07:33:59 kind-control-plane kubelet[833]: W0811 07:33:59.653820     833 fs.go:588] stat failed on /dev/mapper/luks-a389c146-db36-4c96-bcbc-0fa3f5f3fcd1 with error: no such file or directory
Aug 11 07:33:59 kind-control-plane kubelet[833]: E0811 07:33:59.653846     833 kubelet.go:1423] "Failed to start ContainerManager" err="failed to get rootfs info: failed to get device for dir \"/var/lib/kubelet\": could not find device with major: 0, minor: 40 in cached partitions map"
```

On the host the `luks` path is a symlink:

```
ls -la /dev/mapper
total 0
drwxr-xr-x.  2 root root      80 Aug 11 08:43 .
drwxr-xr-x. 21 root root    4600 Aug 11 08:44 ..
crw-------.  1 root root 10, 236 Aug 11 08:43 control
lrwxrwxrwx.  1 root root       7 Aug 11 08:43 luks-a389c146-db36-4c96-bcbc-0fa3f5f3fcd1 -> ../dm-0
```

As this path is not available in the container it fails.

**What you expected to happen**:

All paths required inside kind should be mapped into the node.

**How to reproduce it (as minimally and precisely as possible)**:

Attempt to create a cluster on an encrypted root partition - in my case I simply installed Fedora and chose to encrypt the system in the installer.

**Anything else we need to know?**:

The issue is quite simple to fix, by just also mounting the missing path into the container.

With the following configuration it will work:

```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
  extraMounts:
    - hostPath: /dev/dm-0
      containerPath: /dev/dm-0
      propagation: HostToContainer
```

**Environment:**

- kind version: (use `kind version`):
kind v0.11.1 go1.16.4 linux/amd64

- Kubernetes version: (use `kubectl version`):
```
[...]

PR #2584, opened 2022-01-11
Document known issue around BTRFS

simon-geard:
Following discussions under issue #2411, documenting problem with finding rootfs device with BTRFS (and maybe other unrecognised filesystems), along with the workaround of adding devices as extra mounts.

Also threw in a quick reminder at the top of the page about how to obtain logs if cluster creation fails.

### DIFF
--- a/site/content/docs/user/known-issues.md
+++ b/site/content/docs/user/known-issues.md
@@ -322,7 +322,7 @@ stat failed on /dev/nvme0n1p3 with error: no such file or directory
 "Failed to start ContainerManager" err="failed to get rootfs info: failed to get device for dir \"/var/lib/kubelet\": could not find device with major: 0, minor: 40 in cached partitions map"
 ```
 
-Kubernetes needs access to storage device nodes in order to do some stuff, e.g. tracking free disk space. Therefore, Kind needs to mount the necessary device nodes from the host into the control-plane container — however, it cannot always determine which device Kubernetes requires, since this varies with the host filesystem. For example, Kind doesn't handle BTRFS, which is the default for modern Fedora.
+Kubernetes needs access to storage device nodes in order to do some stuff, e.g. tracking free disk space. Therefore, Kind needs to mount the necessary device nodes from the host into the control-plane container — however, it cannot always determine which device Kubernetes requires, since this varies with the host OS and filesystem. For example, the error above occurred with a BTRFS filesystem on Fedora Desktop 35.
 
 This can be worked around by including the necessary device as an extra mount in the cluster configuration file.
 
@@ -337,7 +337,9 @@ nodes:
       propagation: HostToContainer
 ```
 
-The expected device is named in the error message, but will typically be the location where container volumes are stored — for rootless Docker or Podman, this will usually be $HOME.
+To identify the device that must be listed, two variations have been observed.
+* The device reported in the error message is a symlink (e.g. `/dev/mapper/luks-903aad3d-...`) — in this case, the config file should refer to the target of that symlink (e.g. `/dev/dm-0`).
+* The device reported in the error message is a regular block device (e.g. `/dev/nvme0n1p3`) — in this case, use the device reported.
 
 See Previous Discussion: [kind#2411]
 
--- a/site/content/docs/user/known-issues.md
+++ b/site/content/docs/user/known-issues.md
@@ -23,6 +23,7 @@ description: |-
 
 
 ## Contents
+* [Troubleshooting Kind](#troubleshooting-kind)
 * [Kubectl Version Skew](#kubectl-version-skew) (Kubernetes limits supported version skew)
 * [Older Docker Installations](#older-docker-installations) (untested, known to have bugs)
 * [Docker Installed With Snap](#docker-installed-with-snap) (snap filesystem restrictions problematic)
@@ -39,6 +40,12 @@ description: |-
 * [IPv6 Port Forwarding](#ipv6-port-forwarding) (docker doesn't seem to implement this correctly)
 * [Couldn't find an alternative telinit implementation to spawn](#docker-init-daemon-config)
 * [Fedora](#fedora) (various)
+* [Failed to get rootfs info](#failed-to-get-rootfs-info--stat-failed-on-dev)
+
+## Troubleshooting Kind
+
+If the cluster fails to create, try again with the `--retain` option (preserving the failed container),
+then run `kind export logs` to export the logs from the container to a temporary directory on the host.
 
 ## Kubectl Version Skew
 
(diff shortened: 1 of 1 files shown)


site/content/docs/user/known-issues.md
```diff
@@ -39,6 +40,12 @@ description: |-
 * [IPv6 Port Forwarding](#ipv6-port-forwarding) (docker doesn't seem to implement this correctly)
 * [Couldn't find an alternative telinit implementation to spawn](#docker-init-daemon-config)
 * [Fedora](#fedora) (various)
+* [Failed to get rootfs info](#failed-to-get-rootfs-info--stat-failed-on-dev)
+
+## Troubleshooting Kind
+
+If the cluster fails to create, try again with the `--retain` option (preserving the failed container),
+then run `kind export logs` to export the logs from the container to a temporary directory on the host.
```
aojea, line 48: do you mind address one thing only and focus on the rootfs problem by now?

site/content/docs/user/known-issues.md
```diff
@@ -306,6 +313,34 @@ your workloads inside the cluster via the nodes IPv6 addresses.
...
+This can be worked around by including the necessary device as an extra mount in the cluster configuration file.
+
+```yaml
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+nodes:
+- role: control-plane
+  extraMounts:
+    - hostPath: /dev/nvme0n1p3
+      containerPath: /dev/nvme0n1p3
```
aojea, line 336: should we add some instructions for users on how to obtain the device path?

site/content/docs/user/known-issues.md
```diff
@@ -306,6 +313,34 @@ your workloads inside the cluster via the nodes IPv6 addresses.
...
+## Failed to get rootfs info / "stat failed on /dev/..."
+
+On some systems, creating a cluster times out with these errors in kubelet.log (device varies):
+
+```
+stat failed on /dev/nvme0n1p3 with error: no such file or directory
+"Failed to start ContainerManager" err="failed to get rootfs info: failed to get device for dir \"/var/lib/kubelet\": could not find device with major: 0, minor: 40 in cached partitions map"
+```
+
+Kubernetes needs access to storage device nodes in order to do some stuff, e.g. tracking free disk space. Therefore, Kind needs to mount the necessary device nodes from the host into the control-plane container — however, it cannot always determine which device Kubernetes requires, since this varies with the host filesystem. For example, Kind doesn't handle BTRFS, which is the default for modern Fedora.
```
AkihiroSuda, line 325: `modern Fedora` -> `modern Fedora on Desktop`, IIUC?

site/content/docs/user/known-issues.md
```diff
@@ -306,6 +313,34 @@ your workloads inside the cluster via the nodes IPv6 addresses.
...
+apiVersion: kind.x-k8s.io/v1alpha4
+nodes:
+- role: control-plane
+  extraMounts:
+    - hostPath: /dev/nvme0n1p3
+      containerPath: /dev/nvme0n1p3
+      propagation: HostToContainer
+```
+
+The expected device is named in the error message, but will typically be the location where container volumes are stored — for rootless Docker or Podman, this will usually be $HOME.
```
AkihiroSuda, line 340: Does this workaround really work on rootless?
I also guess rootless may not require this workaround, as it uses fuse-overlayfs?

site/content/docs/user/known-issues.md
```diff
@@ -306,6 +313,34 @@ your workloads inside the cluster via the nodes IPv6 addresses.
...
+On some systems, creating a cluster times out with these errors in kubelet.log (device varies):
+
+```
+stat failed on /dev/nvme0n1p3 with error: no such file or directory
+"Failed to start ContainerManager" err="failed to get rootfs info: failed to get device for dir \"/var/lib/kubelet\": could not find device with major: 0, minor: 40 in cached partitions map"
+```
+
+Kubernetes needs access to storage device nodes in order to do some stuff, e.g. tracking free disk space. Therefore, Kind needs to mount the necessary device nodes from the host into the control-plane container — however, it cannot always determine which device Kubernetes requires, since this varies with the host filesystem. For example, Kind doesn't handle BTRFS, which is the default for modern Fedora.
+
+This can be worked around by including the necessary device as an extra mount in the cluster configuration file.
```
AkihiroSuda, line 327: I guess, setting `$KIND_EXPERIMENTAL_CONTAINERD_SNAPSHOTTER` to `native` or `fuse-overlafys` may work too?

site/content/docs/user/known-issues.md
simon-geard: Sure, if you prefer it to be split. I just figured that adding a short paragraph to the top of the page contributed information that I wanted to include in my answer (i.e. where to find the error) but was also common to many of the other known issues in the page.

site/content/docs/user/known-issues.md
simon-geard: That's in the following paragraph, where I state "the expected device is named in the error message".

site/content/docs/user/known-issues.md
simon-geard: It certainly works for Podman on Fedora - and yes, the workaround is definitely required; otherwise I wouldn't have known this issue existed.

site/content/docs/user/known-issues.md
simon-geard: That I don't know, and can't test right now. Is that documented somewhere? I've been using Kind for about 2 days now, so not an expert by any means.

site/content/docs/user/known-issues.md
simon-geard: Desktop, certainly... I don't know about other variants. But I don't think Fedora itself is relevant to the issue... I mention it only as an example of a common distro where the problem will occur on a stock configuration. I can adjust the wording if you have any suggestions?

site/content/docs/user/known-issues.md
e-minguez: Just in case, in my Fedora 35 it complained about `/dev/mapper/luks-903aad3d-...` and using `hostPath/containerPath: /dev/mapper/luks-903aad3d-..` didn't worked because it is a symlink to `/dev/dm-0`. Using `/dev/dm-0` worked.

```
ls -l /dev/mapper/
total 0
crw-------. 1 root root 10, 236 ene 11 08:06 control
lrwxrwxrwx. 1 root root       7 ene 11 08:06 luks-903aad3d-... -> ../dm-0
```

site/content/docs/user/known-issues.md
simon-geard: Ok, did some quick experimenting today. Assuming you're expecting something like:
```sh
KIND_EXPERIMENTAL_CONTAINERD_SNAPSHOTTER=native kind create cluster --retain
```

...then neither `native` nor `fuse-overlayfs` has any apparent effect. Both fail with the original problem, "stat failed on /dev/nvme0n1p3 with error".

site/content/docs/user/known-issues.md
simon-geard: So, two variations - if the device cited in the error message is under `/dev/mapper`, the extra mount should be the /dev/dm-* device it points to; otherwise it should be the device from the message?

AkihiroSuda (approved):

site/content/docs/user/known-issues.md
e-minguez: I guess it needs to be the real device, not a symlink. So if the error complains about `/dev/foo/bar`, the config needs to have the real device such as `readlink -f /dev/foo/bar`

```
MYDEVICE=$(readlink -f /dev/mapper/luks-903aad3d-...)
cat << EOF | kind create cluster --config=-
kind: Cluster
nodes:
- role: control-plane
  extraMounts:
  - hostPath: ${MYDEVICE}
    containerPath: ${MYDEVICE}
    propagation: HostToContainer
EOF
```

site/content/docs/user/known-issues.md
aojea: my concern is that it doesn't look like a section in its own
Maybe we can reword the description block in lines 8-21 of this file ?

simon-geard:
A question for whomever it might concern — what's still needed in order to get this documentation change merged?

I believe I've satisfied the feedback around the change itself, but as someone entirely unfamiliar with the processes surrounding this project, it's not clear if there's anything else I need to do? There's an automated comment advising of a build failure, but this does not seem to have an obvious connection to my change...

aojea:
> A question for whomever it might concern — what's still needed in order to get this documentation change merged?

sorry, we have been busy and some reviews slip 

/lgtm
/approve

aojea:
easycla pass but cla/linuxfoundation doesn't 

@mrbobbytables is it ok to override?

BenTheElder:
Both CLA are passing now

simon-geard:
Yeah, the CLA thing was my mistake, I think... I thought I'd done it, but there are actually two of them?

PR #2590, opened 2022-01-12
Point broken links to live ad-free file permissions calculator

gotzmann:
Replace two broken documentation links to http://www.filepermissions.com returning HTTP 404 errors with live ad-free alternative on http://permissions-calculator.org


aojea:
@BenTheElder why do we have those links?
I assume the people touching those files will have tha knowledge 😅

aojea:
@gotzmann you need to sign the cla :smile:

gotzmann:
@aojea I've signed "CNCF Individual Contributor License Agreement" with HelloSign before opening this PR - not sure, what else do I need to confirm it? 

I see "EasyCLA — Missing CLA Authorization" link here, but it is not allowed to proceed for individual contributors.

aojea:
`cla/linuxfoundation — commit missing GitHub user` 

don't worry, this things use to be tricky but we'll figure it out, check first the error about missing github user in commit, I think that the tool uses that to double check the CLA

PR #2591, opened 2022-01-12
Point broken links to live ad-free file permissions calculator

gotzmann:
Replace two broken documentation links to http://www.filepermissions.com returning HTTP 404 errors with live ad-free alternative on http://permissions-calculator.org


aojea:
> the new URLs seem valid, but instead i'd remove the external URLs to avoid the binding to websites and include in parentheses the `rwx` notation - e.g. `644 (rw- r-- r--)`

/ok-to-test
I agree with that, I prefer to no have links too, @gotzmann do you mind change the PR for doing that?

gotzmann:
@aojea done (with --amend previous changes)

aojea:
/lgtm
/approve
Thanks

Issue #1699, opened 2020-06-30
migrate kubeadm control-plane taint


BenTheElder:
See: https://github.com/kubernetes/kubeadm/issues/2200

We'll need to keep in sync with the planning there and ensure kind's handling of this does not block work there.

Currently off the top of my head we only reference the `node-role.kubernetes.io/master:NoSchedule` taint to remove it from single-node clusters. We'll want to handle removing the correct taint based on version.

BenTheElder:
This is blocked on upstream changes, probably in Kubernetes 1.20?
When the upstream project is ready we'll coordinate on this.
Frozen to avoid fejta-bot noise / accidental closure before then.

BenTheElder:
https://github.com/kubernetes/kubernetes/pull/95382

BenTheElder:
Updates:
- wg-naming was concluded
- I think this is closing out in kubeadm in 1.24

There are still various bits in this repo referencing the old label, we'll need to carefully handle new + old depending on version.

Tentatively flagging as a blocker for kind v0.13

BenTheElder:
In 1.24:

> [*] continue the rename of the kubeadm "master" label / taint to
"control-plane".
in 1.24 we are adding a new taint
"node-role.kubernetes.io/control-plane" with effect "NoSchedule" and
removing the "node-role.kubernetes.io/master" label on nodes. in 1.20
we instructed all users to preemptively add toleration for this taint
in their workloads and to start using the newly added label
"node-role.kubernetes.io/control-plane".
this plan is detailed in the KEP linked here:
https://github.com/kubernetes/kubeadm/issues/2200

See: https://groups.google.com/d/msgid/kubernetes-sig-cluster-lifecycle/CAGDbWi_5tQ83H_bKYKn2G1O_UWQviUsHmnspfboB1%3DHZ-tVv7A%40mail.gmail.com

We will need to handle removing the taint for single node clusters, and we will need to make sure waitForReady handles the updated label as well. Docs for workload samples pinning to the control plane node may also need updating (I _think_ we already used our own label everywhere for that but ...)

neolit123:
PR with the label / taint adjustments for 1.24 is here:
https://github.com/kubernetes-sigs/kind/issues/1699

BenTheElder:
the kind changes are now done thanks to @neolit123, but we still need to review the docs / samples

wolf99:
Having searched for incidences of "master" across the repo, it seems like all occasions are either 

[...]

PR #2595, opened 2022-01-17
handle "master" label/taint changes for kubeadm 1.24

neolit123:
As per the kubeadm KEP for "master" label/taint changes:
- kubeadm 1.24 nodes will only have the "control-plane" label.
- kubeadm 1.24 nodes will have both the "master"
and "control-plane" taints.
- kubeadm 1.25 nodes will have only the "control-plane" taint

Adapt kind to these changes and add TODOs with links
for the tracking issue in k-sigs/kind.

this is blocking https://github.com/kubernetes/kubernetes/pull/107533
tracking issue in kind repository https://github.com/kubernetes-sigs/kind/issues/1699

### DIFF
--- a/pkg/cluster/internal/create/actions/kubeadminit/init.go
+++ b/pkg/cluster/internal/create/actions/kubeadminit/init.go
@@ -27,6 +27,7 @@ import (
 
 	"sigs.k8s.io/kind/pkg/cluster/internal/create/actions"
 	"sigs.k8s.io/kind/pkg/internal/apis/config"
+	"sigs.k8s.io/kind/pkg/internal/version"
 )
 
 // kubeadmInitAction implements action for executing the kubeadm init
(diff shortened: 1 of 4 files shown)


neolit123:
/kind deprecation
/priority important-soon

/cc @BenTheElder @aojea 
PTAL for review. if you think this can be improved, i can update it.

aojea:
```
- First - 1.20
- Second - Minimum deprecation period for GA features is 1 year.
Estimated 1.24, but may depend on user feedback.
```

@neolit123 IIUIC 

before 1.20 only master
1.20 to 1.24 master and control-plane
1.25+ only control-plane

we try to support as much as possible old versions, I think we should consider another branch in the logic for the clusters 1.20-

neolit123:
@aojea yes, your understanding is correct.
because of the wide range of old versions supported in kind i went for this:
```
	selectorLabel := "node-role.kubernetes.io/control-plane"
	if kubeVersion.LessThan(version.MustParseSemantic("v1.24.0")) {
		selectorLabel = "node-role.kubernetes.io/master"
	}
```

which means that the "master" label will be used for 1.23 and older.

pkg/cluster/internal/create/actions/kubeadminit/init.go
```diff
@@ -106,14 +107,31 @@ func (a *action) Execute(ctx *actions.ActionContext) error {
...
+			return errors.Wrap(err, "failed to get Kubernetes version from node")
+		}
+		kubeVersion, err := version.ParseSemantic(rawVersion)
+		if err != nil {
+			return errors.Wrap(err, "could not parse Kubernetes version")
+		}
+		taints := []string{"node-role.kubernetes.io/control-plane-", "node-role.kubernetes.io/master-"}
+		if kubeVersion.LessThan(version.MustParseSemantic("v1.24.0")) {
+			taints = []string{"node-role.kubernetes.io/master-"}
+		}
```
aojea, line 127: here is my confusion

if kubernetes > 1.24
it removes `"node-role.kubernetes.io/control-plane-", "node-role.kubernetes.io/master-` , should not remove `node-role.kubernetes.io/control-plane-` only?

the way I understand it is that we should have 3 cases,
1. > 1.24 remove `node-role.kubernetes.io/control-plane-` 
2. 1.20-1.24 remove `"node-role.kubernetes.io/control-plane-", "node-role.kubernetes.io/master-"`
3. < 1.20 remove `"node-role.kubernetes.io/master-"`

neolit123: note that taints and labels are staged differently.
the description of https://github.com/kubernetes/kubernetes/pull/107533 has the details.

my response here is about labels:
https://github.com/kubernetes-sigs/kind/pull/2595#issuecomment-1017451427

for taints and single CP nodes:
- in < 1.24 we want to remove only the `master` taint, because old versions of kubeadm only add the `master` taint.
- in >= 1.24 we want to remove the `master` and `control-plane` taints because kubeadm 1.24+ will start adding both taints.
- in 1.25 we have to touch kind again to remove the removal of the `master` taint, because kubeadm 1.25 will stop applying it.

aojea: ic my confusion now, thanks

BenTheElder (approved):
/lgtm
/approve
thanks so much and sorry for the delay!

Issue #2600, opened 2022-01-21
Kind local registry: Error response from daemon: endpoint with name kind-registry already exists in network kind


iamtodor:
Hello,

I run the script from the page https://kind.sigs.k8s.io/docs/user/local-registry/ here is the console output:
```
usr@mcbook  ~  sh kind-local-registry.sh
Creating cluster "kind-registry" ...
 ✓ Ensuring node image (kindest/node:v1.21.1) 🖼
 ✓ Preparing nodes 📦
 ✓ Writing configuration 📜
 ✓ Starting control-plane 🕹️
 ✓ Installing CNI 🔌
 ✓ Installing StorageClass 💾
Set kubectl context to "kind-kind-registry"
You can now use your cluster with:

kubectl cluster-info --context kind-kind-registry

Have a nice day! 👋
Error response from daemon: endpoint with name kind-registry already exists in network kind
configmap/local-registry-hosting created
```

And it says to me the error: `Error response from daemon: endpoint with name kind-registry already exists in network kind`. What should I do to handle it properly? I've searched google for this kind of issue but no luck.

I've walked thru https://github.com/kubernetes-sigs/kind/issues/1213 but there is no mention regarding my topic.

If I miss something or any other details has to be provided from my side please let me know

aojea:
most probably you already have a container running as registry `kind-registry`

iamtodor:
@aojea seems like you are right. However, it's quite interesting: I delete the cluster by running `kind delete cluster --name kind-registry` but it turned out the container is still running 🤔 

Here is the full console output:
```
itodorenko@mcbook  ~  docker ps
CONTAINER ID   IMAGE                                 COMMAND                  CREATED              STATUS              PORTS                                                                                                                                  NAMES
c4f49cd54e35   registry:2                            "/entrypoint.sh /etc…"   About a minute ago   Up About a minute   127.0.0.1:5000->5000/tcp                                                                                                               kind-registry
14b08c33aeb3   gcr.io/k8s-minikube/kicbase:v0.0.28   "/usr/local/bin/entr…"   7 days ago           Up 2 hours          127.0.0.1:55853->22/tcp, 127.0.0.1:55854->2376/tcp, 127.0.0.1:55856->5000/tcp, 127.0.0.1:55857->8443/tcp, 127.0.0.1:55855->32443/tcp   minikube
 itodorenko@mcbook  ~  kind get clusters

No kind clusters found.
 itodorenko@mcbook  ~ 
[...]

PR #2601, opened 2022-01-24
Local registry: only attach network if missing

stmcginnis:
The example script for creating a kind cluster with a local registry
will connect the container to the "kind" network. The first time the
local registry is created, this is fine and is what is needed. On
subsequent runs to create new clusters, if the local registry container
has already been set up, the command to attach to the "kind" network
will emit an error message to the create output that can be confusing.

This updates the example script to do a basic check of the registry
containers network configuration. If the kind network appears to be
already connected, then the script will not attempt to connect it again.

Fixed: #2600

### DIFF
--- a/site/static/examples/kind-with-registry.sh
+++ b/site/static/examples/kind-with-registry.sh
@@ -23,7 +23,9 @@ EOF
 
 # connect the registry to the cluster network
 # (the network may already be connected)
-docker network connect "kind" "${reg_name}" || true
+if [ "$(docker inspect --format='{{json .NetworkSettings.Networks.kind}}' "${reg_name}")" = 'null' ]; then
+    docker network connect "kind" "${reg_name}" || true
+fi
 
 # Document the local registry
 # https://github.com/kubernetes/enhancements/tree/master/keps/sig-cluster-lifecycle/generic/1755-communicating-a-local-registry


site/static/examples/kind-with-registry.sh
```diff
@@ -23,7 +23,11 @@ EOF
 
 # connect the registry to the cluster network
 # (the network may already be connected)
-docker network connect "kind" "${reg_name}" || true
+networks=$(docker inspect --format='{{.NetworkSettings.Networks}}' "$reg_name")
+kindnetwork=$(echo "${networks}" | grep 'kind:')
+if [ "${networks}" != "${kindnetwork}" ]; then
+    docker network connect "kind" "${reg_name}" || true
+fi
```
neolit123, line 30: i think it is possible to do something like:

```
kindnetwork=$(docker inspect --format='{{json .NetworkSettings.Networks.kind}}' "$reg_name")
if [ "${kindnetwork}" != "null" ]; then
    docker network connect "kind" "${reg_name}" || true
fi
```

(untested)

stmcginnis: I don't think you can use a variable network name in the `docker inspect --format` specifier. I was not able to find a way to make that work.

neolit123: hm, this article here references a network like that:
https://www.baeldung.com/ops/docker-network-information

`{{.NetworkSettings.Networks.network1....`

stmcginnis: Hmm, I tried it again and can only assume I typo'd something when I was trying to get it before. This works, and returns the string "null" if `.NetworkSettings.Networks.kind` can't be found, otherwise all the details of the network. So this would be a much simpler check.

I'll get this updated to use that approach instead. Thanks!

neolit123:
thanks
/lgtm

site/static/examples/kind-with-registry.sh
```diff
@@ -23,7 +23,10 @@ EOF
 
 # connect the registry to the cluster network
 # (the network may already be connected)
-docker network connect "kind" "${reg_name}" || true
+kindnetwork=$(docker inspect --format='{{json .NetworkSettings.Networks.kind}}' "${reg_name}")
```
BenTheElder, line 26: nit: we're only going to compare this value, so we should just compare the result directly.
```suggestion
if [ "$(docker inspect --format='{{json .NetworkSettings.Networks.kind}}' "${reg_name}")" = 'null' ]; then
```

also, we should single-quote literals (`'null'`)

BenTheElder (commented):
thanks for this! 😅 
much cleaner than I hoped, some nits 😅

site/static/examples/kind-with-registry.sh
BenTheElder: actually, let's just do the single quote nit for now, I don't know why we didn't compare directly above. perhaps trying to avoid scrolling in the docs presentation?

BenTheElder: we can always revisit both conditionals later

stmcginnis: Oops, pushed update while you were still looking. Let me know if the latest rev needs any changes.

BenTheElder:
https://deploy-preview-2601--k8s-kind.netlify.app/docs/user/local-registry/ looks like it was about the scrollbox, but we can follow-up, this is a good change to land.

thank you!
/lgtm
/approve

PR #2607, opened 2022-01-26
fix: typo and a wrong comment

wanlerong:
It looks like when `err` is nil, it will return false

### DIFF
--- a/pkg/cluster/internal/providers/docker/network.go
+++ b/pkg/cluster/internal/providers/docker/network.go
@@ -291,7 +291,7 @@ func isOnlyErrorNoSuchNetwork(err error) bool {
 		} else if err != nil {
 			return false
 		}
-		// if the line begins with Eror: No such network: it's fine
+		// if the line begins with Error: No such network: it's fine
 		s := string(l)
 		if strings.HasPrefix(s, "Error: No such network:") {
 			continue
--- a/pkg/cluster/internal/providers/docker/network.go
+++ b/pkg/cluster/internal/providers/docker/network.go
@@ -275,7 +275,6 @@ func isNetworkAlreadyExistsError(err error) bool {
 }
 
 // returns true if:
-// - err is nil
 // - err only contains no such network errors
 func isOnlyErrorNoSuchNetwork(err error) bool {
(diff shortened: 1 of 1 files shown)


aojea:
/lgtm
/ok-to-test
/approve
Thanks

pkg/cluster/internal/providers/docker/network.go
```diff
@@ -275,7 +275,6 @@ func isNetworkAlreadyExistsError(err error) bool {
 }
 
 // returns true if:
-// - err is nil
```
aojea, line 278: upps :upside_down_face:

aojea:
/lgtm
Thanks

aojea:
hmm, should it return true and the implementation is wrong?

aojea:
it seems it never gets nil 
```
pkg/cluster/internal/providers/docker/network.go:               if err := deleteNetworks(networks[1:]...); err != nil && !isOnlyErrorNoSuchNetwork(err) {
pkg/cluster/internal/providers/docker/network.go:       if err != nil && !isOnlyErrorNoSuchNetwork(err) {
pkg/cluster/internal/providers/docker/network.go:func isOnlyErrorNoSuchNetwork(err error) bool {
```

@wanlerong can you recover the comment and fix the function ?

```
+++ b/pkg/cluster/internal/providers/docker/network.go
@@ -278,6 +278,9 @@ func isNetworkAlreadyExistsError(err error) bool {
 // - err is nil
 // - err only contains no such network errors
 func isOnlyErrorNoSuchNetwork(err error) bool {
+       if err == nil {
+               return true
+       }
        rerr := exec.RunErrorForError(err)
        if rerr == nil {
                return false
```

wanlerong:
@aojea thanks, please see this [https://github.com/kubernetes-sigs/kind/pull/2608](pr)

pkg/cluster/internal/providers/docker/network.go
BenTheElder: I don't know why we included this in the comment, but IMHO the caller should be handling nil error vs non-nil error.

PR #2608, opened 2022-01-26
fix: isOnlyErrorNoSuchNetwork return true when err is nil

wanlerong:


aojea:
/ok-to-test
/lgtm
/hold

@BenTheElder for more eyes, I think that now is right
https://github.com/kubernetes-sigs/kind/pull/2607#issuecomment-1022096959

pkg/cluster/internal/providers/docker/network.go
```diff
@@ -275,8 +275,12 @@ func isNetworkAlreadyExistsError(err error) bool {
 }
 
 // returns true if:
+// - err is nil
```
BenTheElder, line 278: why? this seems wrong? it can't be a particular kind of error if it's nil / not even an error

BenTheElder:
@aojea I think https://github.com/kubernetes-sigs/kind/pull/2607 was correct.

aojea:
> @aojea I think #2607 was correct.

Thanks @wanlerong for submitting the PR and sorry for the noise

Issue #2540, opened 2021-11-15
Graceful degradation if fuse device is missing


jawnsy:
**What happened**:

Following #2278, kind will always try to mount `/dev/fuse` to the node image, which is useful when running with a rootless docker daemon. However, if the fuse device is missing on the host, then kind will fail to start, even if running under a rootful docker daemon.

At present, [the sysbox container runtime](https://github.com/nestybox/sysbox) does not implicitly mount any devices - such as `/dev/fuse` - from the host, and so the node image will fail to start.

**What you expected to happen**:

If we run the command that kind runs, but remove the --device and --detach flags, then we get the following output (i.e. the node appears to start up correctly):

```shell-session
$ docker run --name kind-control-plane --hostname kind-control-plane --label io.x-k8s.kind.role=control-plane --privileged --security-opt seccomp=unconfined --security-opt apparmor=unconfined --tmpfs /tmp --tmpfs /run --volume /var --volume /lib/modules:/lib/modules:ro -e KIND_EXPERIMENTAL_CONTAINERD_SNAPSHOTTER  --tty --label io.x-k8s.kind.cluster=kind --net kind --restart=on-failure:1 --init=false --publish=127.0.0.1:36507:6443/TCP -e KUBECONFIG=/etc/kubernetes/admin.conf kindest/node:v1.22.2@sha256:9af3ab3e36fb59890b2fb3a18000930b4792d62d10d2060e0ca701e2c392d487
INFO: running in a user namespace (experimental)
WARN: /dev/kmsg does not exist, symlinking /dev/console
INFO: ensuring we can execute mount/umount even with userns-remap
INFO: remounting /sys read-only
INFO: making mounts shared
INFO: detected cgroup v1
INFO: fix cgroup mounts for all subsystems
INFO: clearing and regenerating /etc/machine-id
Initializing machine ID from random generator.
INFO: faking /sys/class/dmi/id/product_name to be "kind"
INFO: faking /sys/class/dmi/id/product_uuid to be random
INFO: faking /sys/devices/virtual/dmi/id/product_uuid as well
INFO: setting iptables to detected mode: legacy
INFO: Detected IPv4 address: 172.17.0.2
INFO: Detected IPv6 address: fc00:f853:ccd:e793::2
[...]

PR #2619, opened 2022-02-04
only mount /dev/fuse for rootless

aojea:
Fixes: #2540

### DIFF
--- a/pkg/cluster/internal/providers/docker/provision.go
+++ b/pkg/cluster/internal/providers/docker/provision.go
@@ -194,6 +194,14 @@ func commonArgs(cluster string, cfg *config.Cluster, networkName string, nodeNam
 		args = append(args, "--volume", "/dev/mapper:/dev/mapper")
 	}
 
+	// rootless: use fuse-overlayfs by default
+	// https://github.com/kubernetes-sigs/kind/issues/2275
+	i, _ := info()
+	if i != nil && i.Rootless {
+		// enable /dev/fuse explicitly for fuse-overlayfs
+		// (Rootless Docker does not automatically mount /dev/fuse with --privileged)
+		args = append(args, "--device", "/dev/fuse")
+	}
 	return args, nil
 }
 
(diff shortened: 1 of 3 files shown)


pkg/cluster/internal/providers/docker/provision.go
```diff
@@ -194,6 +194,14 @@ func commonArgs(cluster string, cfg *config.Cluster, networkName string, nodeNam
 		args = append(args, "--volume", "/dev/mapper:/dev/mapper")
 	}
 
+	// rootless: use fuse-overlayfs by default
+	// https://github.com/kubernetes-sigs/kind/issues/2275
+	i, _ := info()
```
aojea, line 199: this is consistent with `mountDevMapper` and `usernsRemap` that exec and parse `docker ...` commands and swallow the errors

pkg/cluster/internal/providers/podman/provider.go
```diff
@@ -430,8 +430,10 @@ func info(logger log.Logger) (*providers.ProviderInfo, error) {
 		SupportsCPUShares:   cgroupSupportsCPUShares,
 	}
 	if info.Rootless && !v.AtLeast(version.MustParseSemantic("4.0.0")) {
-		logger.Warn("Cgroup controller detection is not implemented for Podman. " +
-			"If you see cgroup-related errors, you might need to set systemd property \"Delegate=yes\", see https://kind.sigs.k8s.io/docs/user/rootless/")
+		if logger != nil {
```
aojea, line 433: the logger is only used, here, since we need to use the `info(logger log.Logger)` function to detect if it is rootless, better to make the logging optional, when podman < 4.0 is deprecated we don't need this logger anymore

pkg/cluster/internal/providers/docker/provision.go
BenTheElder: it would be nice if we could capture all this once but the cost is low enough that prefer getting the fix.

we should probably revisit error swallowing later ...

pkg/cluster/internal/providers/podman/provider.go
BenTheElder: logger is a standard part of the provider /shrug

BenTheElder:
you know it's probably better to just do this than plumbing through the info and capturing it once, we can revisit later. thank you for fixing this.

pkg/cluster/internal/providers/docker/provision.go
BenTheElder: TODO: should we make this `if mountFuse() {` for consistency? 🙃

BenTheElder:
/lgtm
better to do nits later, and get the fix in.

pkg/cluster/internal/providers/docker/provision.go
aojea: not TODO, it is less ugly 
let me change it

PR #2620, opened 2022-02-04
only mount /dev/fuse for rootless

aojea:
follow up on https://github.com/kubernetes-sigs/kind/pull/2619/files#r799838896

### DIFF
--- a/pkg/cluster/internal/providers/docker/util.go
+++ b/pkg/cluster/internal/providers/docker/util.go
@@ -85,3 +85,16 @@ func mountDevMapper() bool {
 
 	return storage == "btrfs" || storage == "zfs" || storage == "xfs"
 }
+
+// rootless: use fuse-overlayfs by default
+// https://github.com/kubernetes-sigs/kind/issues/2275
+func mountFuse() bool {
+	i, err := info()
+	if err != nil {
+		return false
+	}
+	if i != nil && i.Rootless {
+		return true
+	}
+	return false
+}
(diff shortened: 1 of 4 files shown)


pkg/cluster/internal/providers/docker/util.go
```diff
@@ -85,3 +86,34 @@ func mountDevMapper() bool {
 
 	return storage == "btrfs" || storage == "zfs" || storage == "xfs"
 }
+
+// rootless: use fuse-overlayfs by default
+// https://github.com/kubernetes-sigs/kind/issues/2275
+func mountFuse() bool {
+	cmd := exec.Command("docker", "info", "--format", "{{json .}}")
```
BenTheElder, line 93: should we just wrap the info method like we did with podman?

aojea: what do you mean exactly? I see both the same :thinking:

BenTheElder: I mean one of them actually re-uses the `info` method without reimplementing the contents

aojea: I've removed those dependencies on this PR in both places, but I get your point now ... since we depend on the rootless variable of info() seems good to add that dependency

aojea: updated

BenTheElder (approved):
/lgtm
/approve
thanks!

PR #2650, opened 2022-03-03
pkg/build: improve source finding message

stevekuznetsov:
Previously, users were not told where `kind` was looking for the
Kubernetes source, making the error message not actionable:

```
$ kind build node-image
ERROR: error building node image: error finding kuberoot: could not find kubernetes source
```

Now, we expose that information so the user can ensure we locate it.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

### DIFF
--- a/pkg/build/nodeimage/internal/kube/source.go
+++ b/pkg/build/nodeimage/internal/kube/source.go
@@ -38,7 +38,7 @@ func FindSource() (root string, err error) {
 	if err == nil && maybeKubeDir(pkg.Dir) {
 		return pkg.Dir, nil
 	}
-	return "", errors.New("could not find kubernetes source")
+	return "", fmt.Errorf("could not find %s module source under GOPATH=%s: %w", ImportPath, build.Default.GOPATH, err)
 }
 
 // maybeKubeDir returns true if the dir looks plausibly like a kubernetes


pkg/build/nodeimage/internal/kube/source.go
```diff
@@ -38,7 +38,7 @@ func FindSource() (root string, err error) {
 	if err == nil && maybeKubeDir(pkg.Dir) {
 		return pkg.Dir, nil
 	}
-	return "", errors.New("could not find kubernetes source")
+	return "", fmt.Errorf("could not find %s module source under GOPATH=%s", ImportPath, build.Default.GOPATH)
```
aojea, line 41: I think that is better to surface the original error, it contains more meaningful messages

tao12345666333: +1

stevekuznetsov:
@aojea ah! Didn't even realize that there was an error getting ignored. Thanks, updated the message with it.

BenTheElder:
thanks!


Looks like we hit timeouts again which are obviously unrelated. 

@aojea At some point these got configured to only request 2 cpu on the default old google.com prow cluster which is not enforcing resource requests still, probably got stomped by a neighboring job. We should look at moving it to the same prow cluster we're running k/k presubmits with so we're actually testing in the _same_ environment ... (e.g. think if the CNCF cluster gets to a newer kernel or something ...)

/retest

aojea:
> thanks!
> 
> Looks like we hit timeouts again which are obviously unrelated.
> 
> @aojea At some point these got configured to only request 2 cpu on the default old google.com prow cluster which is not enforcing resource requests still, probably got stomped by a neighboring job. We should look at moving it to the same prow cluster we're running k/k presubmits with so we're actually testing in the _same_ environment ... (e.g. think if the CNCF cluster gets to a newer kernel or something ...)
> 
> /retest

yeah, I'm seeing a lot of timeouts lately on kind jobs, in k/k too, I have to review those config to check everything is as you say

aojea:
/retest
jobs may be flaky due to resources constraints, moving them to cncf infra and setting resource limits
bear in mind, master has increased memory allocation due to go 1.18
https://github.com/kubernetes/test-infra/pull/25527

aojea:
> /retest
> jobs may be flaky due to resources constraints, moving them to cncf infra and setting resource limits
> bear in mind, master has increased memory allocation due to go 1.18
> [kubernetes/test-infra#25527](https://github.com/kubernetes/test-infra/pull/25527)

PR #2653, opened 2022-03-04
bump go to 1.17.8

BenTheElder:
we shouldn't be impacted as we don't have non-trivial regexp but pickup the CVE fix anyhow ahead of release


aojea:
/retest
:thinking:

PR #2655, opened 2022-03-04
exec: show command error output

stevekuznetsov:
Without the output, often the error that would be printed would simply
not be actionable for the user.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

Before:

```
ERROR: failed to delete cluster "kind": command "podman volume rm --force ''" failed with error: exit status 125
```

After:

```
ERROR: failed to delete cluster "kind": command "podman volume rm --force ''" failed with error: exit status 125: Error: name or ID cannot be empty
```

### DIFF
--- a/pkg/exec/types.go
+++ b/pkg/exec/types.go
@@ -52,7 +52,7 @@ var _ error = &RunError{}
 
 func (e *RunError) Error() string {
 	// TODO(BenTheElder): implement formatter, and show output for %+v ?
-	return fmt.Sprintf("command \"%s\" failed with error: %v", e.PrettyCommand(), e.Inner)
+	return fmt.Sprintf("command \"%s\" failed with error: %v: %s", e.PrettyCommand(), e.Inner, e.Output)
 }
 
 // PrettyCommand pretty prints the command in a way that could be pasted


BenTheElder:
FWIW: output should printed from the top level error handling in app main when verbosity is increased.

aojea:
/hold
to check the impact on verbosity

stevekuznetsov:
I was previously running with `-v 10` and never saw any output, so I'm not sure what you'd like me to check here, specifically?

aojea:
> I was previously running with `-v 10` and never saw any output, so I'm not sure what you'd like me to check here, specifically?

the output should be appended to the error

https://github.com/stevekuznetsov/kind/blob/dece9fae9539d691a15d2f0ec5a61c28ae799f2c/pkg/exec/local.go#L124-L128

stevekuznetsov:
Ah sorry I think I had `kinder` but not `kind` at high verbosity. In any case, I don't really think it's good UX for someone to see "error 135" and have to re-run the command to see why it failed, so ouput at all log levels on error seems appropriate IMHO

BenTheElder:
> Ah sorry I think I had kinder but not kind at high verbosity. In any case, I don't really think it's good UX for someone to see "error 135" and have to re-run the command to see why it failed, so ouput at all log levels on error seems appropriate IMHO

That's fair, though when we output the full error frequently in broken environments the exec output would be the full `kubeadm init` logs which users would then paste an arbitrary "error" line from that just told us that the api-server wasn't running and nothing useful in particular. It's currently gated behind setting `-v` for debugging attempts given that the usefulness of the command output varies widely.

I see an argument for printing it at default verbosity, especially now that the output is a bit better formatted at least and we more often catch and handle / translate specific errors. We should just do more of that when we know what the output means (like https://github.com/kubernetes-sigs/kind/blob/91cc28782d7343dc1547c3d40e7136fd24ed5ddc/pkg/cluster/internal/providers/docker/network.go#L109)  and otherwise display it.

stevekuznetsov:
@BenTheElder I'm tempted to allow-list all `docker` and `podman` outputs - while making sense of the output to a specific error message sounds good, I imagine any and all errors coming from the simple container driver commands on localhost would be good to display at the least. WDYT? Agreed that enormous logs coming from `kubeadm` likely should not be spat out to stdout.

stevekuznetsov:
This needs more thought

/close

PR #2666, opened 2022-03-07
update banner to support humanitarian aid for ukraine

BenTheElder:
As discussed with co-maintainers.

Meant to ship this with v0.12 release, at least it hasn't been out long / announced widely yet.


aojea:
/lgtm
/hold
unhold once you are ok

PR #2670, opened 2022-03-10
Add KUBE_PROXY_DETECT_LOCAL_MODE option in e2e-k8s.sh

wedaly:
Allow running E2E tests with different configurations for kube-proxy "--detect-local-mode".

Currently, the two supported options are ClusterCIDR (default) and NodeCIDR. [PR 95400](https://github.com/kubernetes/kubernetes/pull/95400) is adding two additional options BridgeInterfaceName and InterfacePrefix, which we'd like to be able to validate end-to-end.

Related issue: https://github.com/kubernetes/kubernetes/issues/108525

### DIFF
--- a/hack/ci/e2e-k8s.sh
+++ b/hack/ci/e2e-k8s.sh
@@ -173,6 +173,9 @@ ${kubelet_extra_args}
   nodeRegistration:
     kubeletExtraArgs:
 ${kubelet_extra_args}
+  ---
+  kind: KubeProxyConfiguration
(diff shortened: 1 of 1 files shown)


aojea:
I prefer to start on experiment mode before adding more knobs here , just the same we did with multizone, see https://github.com/kubernetes/test-infra/blob/master/experiment/kind-multizone-e2e.sh an example
/hold

wedaly:
@aojea that makes sense, thanks! Closing this PR

Issue #1685, opened 2020-06-23
When after restart docker, kind cluster could't connect


tsubasaxZZZ:
I created kind cluster with following YAML:
```yaml
# a cluster with 3 control-plane nodes and 3 workers
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
- role: control-plane
- role: control-plane
- role: worker
- role: worker
- role: worker
```

```console
tsunomur@VM:~$ kind create cluster --config kind-example-config.yaml
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.18.2) 🖼
 ✓ Preparing nodes 📦 📦 📦 📦 📦 📦
 ✓ Configuring the external load balancer ⚖️
 ✓ Writing configuration 📜
 ✓ Starting control-plane 🕹️
 ✓ Installing CNI 🔌
 ✓ Installing StorageClass 💾
 ✓ Joining more control-plane nodes 🎮
 ✓ Joining worker nodes 🚜
Set kubectl context to "kind-kind"
You can now use your cluster with:

kubectl cluster-info --context kind-kind

Thanks for using kind! 😊
tsunomur@VM:~$ kubectl cluster-info --context kind-kind
Kubernetes master is running at https://127.0.0.1:43185
KubeDNS is running at https://127.0.0.1:43185/api/v1/namespaces/kube-system/services/kube-dns:dns/proxy

To further debug and diagnose cluster problems, use 'kubectl cluster-info dump'.
tsunomur@VM:~$ docker ps -a
CONTAINER ID        IMAGE                          COMMAND                  CREATED             STATUS              PORTS                       NAMES
03dad9ed89f2        kindest/node:v1.18.2           "/usr/local/bin/entr…"   8 minutes ago       Up 6 minutes                                    kind-worker2
cbd3f2c279a8        kindest/node:v1.18.2           "/usr/local/bin/entr…"   8 minutes ago       Up 6 minutes        127.0.0.1:44681->6443/tcp   kind-control-plane3
1531621e9806        kindest/node:v1.18.2           "/usr/local/bin/entr…"   8 minutes ago       Up 6 minutes                                    kind-worker3
1ceaa76b5149        kindest/haproxy:2.1.1-alpine   "/docker-entrypoint.…"   8 minutes ago       Up 8 minutes        127.0.0.1:43185->6443/tcp   kind-external-load-balancer
[...]

Issue #1689, opened 2020-06-25
HA clusters don't reboot properly


BenTheElder:
first reported in https://github.com/kubernetes-sigs/kind/issues/1685
tracking in an updated bug.

reproduce with:
```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
- role: control-plane
- role: control-plane
```
\+ restart docker.

BenTheElder:
Still needs root causing, but multiple user reports. We should fix this.

ozbillwang:
yes, I hit the same issue today. 

any workaround I can manually fix it? I spent a little bit long time to set up the test KIND environment, I don't want to recreate it. 

Any way to restore it back?

BenTheElder:
I haven't looked into this issue yet.

Regarding the node versions, please read the release notes about the changes, and see the usage and user guide for how to change it.

BenTheElder:
This has never worked. 0.7 and down did not survive reboots for *any*
configuration. 0.8+ apparently doesn't survive reboots for "HA" clusters.

BenTheElder:
cc @aojea you were recently looking at the loadbalancer networking

aojea:
it is more complicated than the load balancer, the control plane nodes has different ips and the cluster does not come up

```
2020-08-27 07:58:58.054356 E | etcdserver: publish error: etcdserver: request timed out
2020-08-27 07:58:58.061687 W | rafthttp: health check for peer 6dd029603bf5e797 could not connect: x509: certificate is valid for 172.18.0.7, 127.0.0.1, ::1, not 172.18.0.5
2020-08-27 07:58:58.061717 W | rafthttp: health check for peer 6dd029603bf5e797 could not connect: x509: certificate is valid for 172.18.0.7, 127.0.0.1, ::1, not 172.18.0.5
2020-08-27 07:58:58.063416 W | rafthttp: health check for peer 2b4992c658e42934 could not connect: dial tcp 172.18.0.7:2380: connect: no route to host
2020-08-27 07:58:58.063454 W | rafthttp: health check for peer 2b4992c658e42934 could not connect: dial tcp 172.18.0.7:2380: connect: no route to host
```

seems we should use hostnames to sign certificates to avoid this

BenTheElder:
We do where we can already. IIRC etcd won't use hostnames.

shlomibendavid:
Same issue here.
my environment:
kind-control-plane
[...]

PR #2671, opened 2022-03-11
Fix multi-node cluster not working after restarting docker

tnqn:
In a multi-node cluster with single controlplane node, if the
controlplane node's IP changes, kube-controller-manager and
kube-scheduler would fail to connect kube-apiserver. Updating the
server address to new IP doesn't work because the API server's
certificate isn't valid for it.

This patch uses "patches" option of kubeadm to replace the server
address in the kubeconfig files of kube-controller-manager and
kube-scheduler with loopback address, which is an alternative address
of the API server's certificate.

Signed-off-by: Quan Tian <qtian@vmware.com>

For #1685

Depends on https://github.com/kubernetes/kubernetes/pull/109886

### DIFF
--- a/pkg/cluster/internal/create/actions/config/config.go
+++ b/pkg/cluster/internal/create/actions/config/config.go
@@ -33,6 +33,7 @@ import (
 	"sigs.k8s.io/kind/pkg/cluster/internal/providers/common"
 	"sigs.k8s.io/kind/pkg/cluster/nodeutils"
 	"sigs.k8s.io/kind/pkg/internal/apis/config"
+	"sigs.k8s.io/kind/pkg/internal/version"
 )
 
 // Action implements action for creating the node config files
@@ -87,6 +88,23 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 	kubeadmConfigPlusPatches := func(node nodes.Node, data kubeadm.ConfigData) func() error {
 		return func() error {
 			data.NodeName = node.String()
+			kubeVersion, err := nodeutils.KubeVersion(node)
+			if err != nil {
+				// TODO(bentheelder): logging here
+				return errors.Wrap(err, "failed to get kubernetes version from node")
+			}
+			data.KubernetesVersion = kubeVersion
+
+			patches, err := getKubeadmPatches(data)
+			if err != nil {
+				return errors.Wrap(err, "failed to generate kubeadm patches content")
+			}
+
+			ctx.Logger.V(2).Infof("Using the following kubeadm patches for node %s:\n%s", node.String(), patches)
+			if err := writeKubeadmPatches(patches, node); err != nil {
+				return err
+			}
+
 			kubeadmConfig, err := getKubeadmConfig(ctx.Config, data, node, provider)
 			if err != nil {
 				// TODO(bentheelder): logging here
@@ -155,13 +173,6 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 // getKubeadmConfig generates the kubeadm config contents for the cluster
 // by running data through the template and applying patches as needed.
 func getKubeadmConfig(cfg *config.Cluster, data kubeadm.ConfigData, node nodes.Node, provider string) (path string, err error) {
-	kubeVersion, err := nodeutils.KubeVersion(node)
-	if err != nil {
-		// TODO(bentheelder): logging here
-		return "", errors.Wrap(err, "failed to get kubernetes version from node")
-	}
-	data.KubernetesVersion = kubeVersion
-
 	// TODO: gross hack!
 	// identify node in config by matching name (since these are named in order)
 	// we should really just streamline the bootstrap code and maintain
@@ -241,6 +252,31 @@ func getKubeadmConfig(cfg *config.Cluster, data kubeadm.ConfigData, node nodes.N
 	return removeMetadata(patchedConfig), nil
 }
 
+// getKubeadmPatches generates the kubeadm patch contents. It returns a map of patch file name to patch content.
+func getKubeadmPatches(data kubeadm.ConfigData) (map[string]string, error) {
+	ver, err := version.ParseGeneric(data.KubernetesVersion)
+	if err != nil {
+		return nil, err
+	}
+
+	patches := map[string]string{}
+	// Kubernetes older than v1.25 don't support patching kubeconfig files.
+	if ver.AtLeast(version.MustParseSemantic("v1.25.0")) {
+		// controller-manager and scheduler connect to local API endpoint, which defaults to the advertise address of the
+		// API server. If the Node's IP changes (which could happen after docker restarts), the server address in KubeConfig
+		// should be updated. However, the server certificate isn't valid for the new IP. To resolve it, we update the
+		// address to loopback address which is an alternative address of the certificate.
+		loopbackAddress := "127.0.0.1"
+		if data.IPFamily == config.IPv6Family {
+			loopbackAddress = "[::1]"
+		}
+		jsonPatch := fmt.Sprintf(`[{"op": "replace", "path": "/clusters/0/cluster/server", "value": "https://%s:%d"}]`, loopbackAddress, data.APIBindPort)
+		patches["controller-manager.conf+json.json"] = jsonPatch
+		patches["scheduler.conf+json.json"] = jsonPatch
+	}
+	return patches, nil
+}
+
 // trims out the metadata.name we put in the config for kustomize matching,
 // kubeadm will complain about this otherwise
 func removeMetadata(kustomized string) string {
@@ -264,6 +300,22 @@ func writeKubeadmConfig(kubeadmConfig string, node nodes.Node) error {
 	if err := nodeutils.WriteFile(node, "/kind/kubeadm.conf", kubeadmConfig); err != nil {
 		// TODO(bentheelder): logging here
 		return errors.Wrap(err, "failed to copy kubeadm config to node")
+	}
+
+	return nil
+}
+
+// writeKubeadmPatches writes the kubeadm patches in the specified node
+func writeKubeadmPatches(patches map[string]string, node nodes.Node) error {
+	patchesDir := "/kind/patches/"
+	if err := node.Command("mkdir", "-p", patchesDir).Run(); err != nil {
+		return errors.Wrapf(err, "failed to create directory %s", patchesDir)
+	}
+
+	for file, patch := range patches {
+		if err := nodeutils.WriteFile(node, patchesDir+file, patch); err != nil {
+			return errors.Wrapf(err, "failed to copy patch file %s to node", file)
+		}
 	}
 
 	return nil
--- a/pkg/cluster/internal/kubeadm/config.go
+++ b/pkg/cluster/internal/kubeadm/config.go
@@ -477,6 +477,8 @@ apiVersion: kubeadm.k8s.io/v1beta3
 kind: InitConfiguration
 metadata:
   name: config
+patches:
+  directory: /kind/patches
 # we use a well know token for TLS bootstrap
 bootstrapTokens:
 - token: "{{ .Token }}"
@@ -498,6 +500,8 @@ apiVersion: kubeadm.k8s.io/v1beta3
 kind: JoinConfiguration
 metadata:
   name: config
+patches:
+  directory: /kind/patches
 {{ if .ControlPlane -}}
 controlPlane:
   localAPIEndpoint:


images/base/files/usr/local/bin/entrypoint
```diff
@@ -372,6 +372,13 @@ enable_network_magic(){
       sed -i "s#${old_ipv4}#${curr_ipv4}#" /etc/kubernetes/manifests/*.yaml || true
       # this is no longer required with autodiscovery
       sed -i "s#${old_ipv4}#${curr_ipv4}#" /var/lib/kubelet/kubeadm-flags.env || true
+      # controller-manager and scheduler connect to local API endpoint. If the Node's IP changes, server address in
+      # kubeconfig should be updated. However, the server certificate isn't valid for the new IP. We update it to
+      # loopback address as it's an alternative address of the certificate.
+      if [[ "${old_ipv4}" != "${curr_ipv4}" ]]; then
+        sed -i "s#${old_ipv4}#127.0.0.1#" /etc/kubernetes/controller-manager.conf || true
```
aojea, line 379: hmm,  @neolit123 how are these files generated?
can we influence them?

it seems easier to generate with the localhost IP instead of the node IP, they will only contact the local apiserver ... in HA we have one of these in each CP

aojea: what I'm saying is to modify the code so the scheduler and the controller-manager always use API server ip address = `localhost` , maybe there is a knob or a flag to do it

tnqn: Thanks @aojea. I agree with you and I actually tried it but it seems more complicated.
The address in controller-manager.conf and scheduler.conf comes from `localAPIEndpoint.advertiseAddress`:
https://github.com/kubernetes-sigs/kind/blob/b4ab02cffbc8aca096ee7e2bc0f105654f1572c6/pkg/cluster/internal/kubeadm/config.go#L210-L211
https://github.com/kubernetes/kubernetes/blob/cdee77a4a9ba77a01bab6b462c1780d7a52af8c3/cmd/kubeadm/app/phases/kubeconfig/kubeconfig.go#L436

But `localAPIEndpoint.advertiseAddress` is used in many places in kubeadm so we can't simply replace it with localhost IP.

aojea: the commit says https://github.com/kubernetes/kubernetes/commit/d9441906c4155173ce1a75421d8fcd1d2f79c471

```
Instead of using the CPE for these components, use the LocalAPIEndpoint.
This guarantees that the components would talk to the local
kube-apiserver, which should be the same version, unless the user
explicitly patched manifests.
```

wonder if we can patch it

tnqn: I guess it uses the local node IP instead of localhost because kubeadm doesn't add localhost to alternative addresses of the server certificate by default.
kind adds it for specific reason: https://github.com/kubernetes-sigs/kind/blob/b4ab02cffbc8aca096ee7e2bc0f105654f1572c6/pkg/cluster/internal/kubeadm/config.go#L167-L171.

aojea:
/ok-to-test

I would prefer to do less magic
https://github.com/kubernetes-sigs/kind/pull/2671/files#r824831420
@neolit123 if the goal is to talk to localhost why not be able to indicate `localhost` directly? do you have some context?

images/base/files/usr/local/bin/entrypoint
```diff
@@ -372,6 +372,13 @@ enable_network_magic(){
       sed -i "s#${old_ipv4}#${curr_ipv4}#" /etc/kubernetes/manifests/*.yaml || true
       # this is no longer required with autodiscovery
       sed -i "s#${old_ipv4}#${curr_ipv4}#" /var/lib/kubelet/kubeadm-flags.env || true
+      # controller-manager and scheduler connect to local API endpoint. If the Node's IP changes, server address in
+      # kubeconfig should be updated. However, the server certificate isn't valid for the new IP. We update it to
+      # loopback address as it's an alternative address of the certificate.
+      if [[ "${old_ipv4}" != "${curr_ipv4}" ]]; then
```
aojea, line 378: we are unconditionally changing the IP, do we need the if?

tnqn: I added this check because if docker restart doesn't change the Node's IP, it can continue working with the original IP. I'm not sure which is better: use the trick when it has to, or use it unconditionally as long as the container is restarted. Would like to hear your opinion.

aojea: yeah, that sounds safer, we don't change the current behavior

aojea:
/lgtm
/assign @BenTheElder 

I can't think in a better solution , we are already doing "magic" for reboots, this is building on that so :shrug:

images/base/files/usr/local/bin/entrypoint
neolit123: localAPIEndpoint.advertiseAddress is an IP because that's what kube-apiserver supports. Advertise / bind addresses for k8s components are IPs...they cannot be hostnames. So it must be 127.0.0.1 and not localhost. Just wanted to be clear about that.

I think I've seen users binding everything to loopback and it should work. I haven't tried it. If something in kubeadm is blocking you let me know. But also check if the manifest patches work for you ALA initconfiguration.patches

BenTheElder:
Thank you for this!

I think we should probably be altering the config during the initial bringup if at all possible, rather than on-reboot, where we have static changes like this (substituting loopback). If kubeadm config doesn't enable this directly, we may have to move into using the phases and generating, which will be trickier but probably the better longterm approach 🤔

tnqn:
> I think we should probably be altering the config during the initial bringup if at all possible, rather than on-reboot, where we have static changes like this (substituting loopback). If kubeadm config doesn't enable this directly, we may have to move into using the phases and generating, which will be trickier but probably the better longterm approach 🤔

@BenTheElder Thanks for the suggestion. I investigated several approaches to alter the config in the beginning but found they seem not ideal:
1. Set `--master` in `extraArgs` of kubeadm ConfigTemplate to override server address in kubeconfig files. However, it only overrides the server address in the main kubeconfig but not `--authentication-kubeconfig` and `--authorization-kubeconfig`, there would be continuous error logs like the following after Node IP change:
```
W0401 09:08:40.383960       1 reflector.go:324] k8s.io/apiserver/pkg/server/dynamiccertificates/configmap_cafile_content.go:205: failed to list *v1.ConfigMap: Get "https://[fc00:f853:ccd:e793::3]:6443/api/v1/namespaces/kube-system/configmaps?fieldSelector=metadata.name%3Dextension-apiserver-authentication&limit=500&resourceVersion=0": dial tcp [fc00:f853:ccd:e793::3]:6443: connect: connection refused
E0401 09:08:40.384763       1 reflector.go:138] k8s.io/apiserver/pkg/server/dynamiccertificates/configmap_cafile_content.go:205: Failed to watch *v1.ConfigMap: failed to list *v1.ConfigMap: Get "https://[fc00:f853:ccd:e793::3]:6443/api/v1/namespaces/kube-system/configmaps?fieldSelector=metadata.name%3Dextension-apiserver-authentication&limit=500&resourceVersion=0": dial tcp [fc00:f853:ccd:e793::3]:6443: connect: connection refused
```
2. initconfiguration.patches doesn't help much as it patches the static pod manifest, not the kubeconfig file. It's not much different from the 1st approach.
3. Use phases: However, even the `kubeconfig` phase command doesn't accept loopback address as `--apiserver-advertise-address` with the following error. So we will have to use substituting command in-between phase commands. Is it what you suggested?
```
# kubeadm init phase kubeconfig controller-manager --apiserver-advertise-address 127.0.0.1
unable to select an IP from lo network interface
```

```
for _, cmd := range []exec.Cmd {
	phaseCommand("kubelet-start"),
	phaseCommand("certs", "all"),
	phaseCommand("kubeconfig", "all"),
	replaceServerCommand(nodeAddress, "127.0.0.1", "/etc/kubernetes/scheduler.conf"),
	phaseCommand("control-plane", "all"),
	...
} {
	lines, err := exec.CombinedOutputLines(cmd)
	ctx.Logger.V(3).Info(strings.Join(lines, "\n"))
	if err != nil {
		return errors.Wrap(err, "failed to init node with kubeadm")
	}
}
```

BenTheElder:
> Use phases: However, even the kubeconfig phase command doesn't accept loopback address as --apiserver-advertise-address with the following error. So we will have to use substituting command in-between phase commands. Is it what you suggested?


yes, sorry.

this change will be more involved, alternatively if we can make it possible to specify everything to kubeadm between patches and other inputs that would be reasonable.

BenTheElder:
[I just realized I never replied to this, I don't know how it slipped through ... so many notifications, I apologize]

I think most ideal is one of:
- kuebeadm makes it possible to specify this directly
- kind uses phases to work around kubeadm

Doing it only upon reboot will be trickier to debug / surprising and be more likely to leave this broken in the future as we won't be doing that in CI.

tnqn:
@BenTheElder @neolit123 @aojea I didn't find a suitable option name to specify the server address for kube-controller-manager and kube-scheduler separately in kubeadm, and the change may sound too specific to kind. So I followed @neolit123's suggestion at https://github.com/kubernetes-sigs/kind/pull/2671#discussion_r829017499 and extended kubeadm's patching support to kubeconfig files (https://github.com/kubernetes/kubernetes/pull/109886). This PR has been updated accordingly. Please let me know your opinions.

neolit123:
The kubeadm patches feature has a KEP so ideally we should document the change and rationale before adding it. Personally i do not see a problem with adding kubeconfig patches but i need to double check.

Phases *should* already support what is required here:
- Call kubeconfig phase on demand
- modify the generated files
- call kubeadm but skip kubeconfig phase

Then again, one of the reasons we added patches was because users did not like phases for some customizations.

I can comment more on https://github.com/kubernetes/kubernetes/pull/109886

Trade off - If it merges, only kubeadm 1.25 will support this, while using phases is supported for a number of old kubeadm versions already.

tnqn:
@neolit123 thanks for your comment. I thought about using phases but found:

- For init phases, we would need to call `certs` first then `kubeconfig`, otherwise the latter would fail due of missing cert, so it would become calling two phases first, modify the generated files, call kubeadm but skip the two phases

- For join phases, it would depend on whether the node is a controlplane Node to decide whether to call control-plane-prepare and modify the generated files.

It seems it would become a bit coupled with the content of phases and may need some adjustment when kubeadm phases change, not sure if it's a suitable long term solution.

For https://github.com/kubernetes/kubernetes/pull/109886, if we agree to this approach, I would be happy to follow your guidance to add documentation for the change.

neolit123:
@tnqn 

can you clarify what is the blocker for using this in kind via `Init/JoinConfiguration.{controlPlane}.localApiEndpoint`?
```
localAPIEndpoint: 
   advertiseAddress: 127.0.0.1
```

https://github.com/kubernetes-sigs/kind/pull/2671#discussion_r824825700

errors are not returned for 127.0.0.1,
https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/util/config/common.go#L142-L161

so unless i'm missing something kubeadm *should* allow you to bind all components and kubeconfigs to loopback.

neolit123:
> - For init phases, we would need to call certs first then kubeconfig, otherwise the latter would fail due of missing cert, so it would become calling two phases first, modify the generated files, call kubeadm but skip the two phases
> - For join phases, it would depend on whether the node is a controlplane Node to decide whether to call control-plane-prepare and modify the generated files.

these seem tolerable. patches also have complexity as kubeadm would need to load files from disk (i.e. from the kind nodes).

> It seems it would become a bit coupled with the content of phases and may need some adjustment when kubeadm phases change, not sure if it's a suitable long term solution.

phases are CLI and have a GA deprecation policy in kubeadm (> 1 year change if it happens).

> For https://github.com/kubernetes/kubernetes/pull/109886, if we agree to this approach, I would be happy to follow your guidance to add documentation for the change.

i think the biggest issue is this trade off that i've mentioned:

>> Trade off - If it merges, only kubeadm 1.25 will support this, while using phases is supported for a number of old kubeadm versions already.

tnqn:
> @tnqn
> 
> can you clarify what is the blocker for using this in kind via `{Init|Join}Configuration[.controlPlane].localApiEndpoint`?
> 
> ```
> localAPIEndpoint: 
>    advertiseAddress: 127.0.0.1 # or v6
> ```
> 
> [#2671 (comment)](https://github.com/kubernetes-sigs/kind/pull/2671#discussion_r824825700)
> 
> errors are not returned for `ip.IsLoopback()`, https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/util/config/common.go#L142-L161
> 
> so unless i'm missing something kubeadm _should_ allow you to bind all components and kubeconfigs (except kubelets) to loopback by setting it in `advertiseAddress`.

It would return the error "unable to select an IP from lo network interface" because setting loopback address has a special meaning as explained by the below comment and it expects another global IP on loopback device.
https://github.com/kubernetes/kubernetes/blob/4717a59097d7365ddcea578dae9aa29ade78398a/cmd/kubeadm/app/util/config/initconfiguration.go#L138-L141

neolit123:
> It would return the error "unable to select an IP from lo network interface" because setting loopback address has a special meaning as explained by the below comment and it expects another global IP on loopback device.
https://github.com/kubernetes/kubernetes/blob/4717a59097d7365ddcea578dae9aa29ade78398a/cmd/kubeadm/app/util/config/initconfiguration.go#L138-L141

i vaguely recall when that logic was added and also recall the "-1"s from people when "just allowing loopback" was discussed.

so one question here would be whether we want to change the logic and just allow loopback addresses.
but if we do that we have to somehow preserve the current bahavior "if loopback address -> ask LO interface -> pick global address".

tnqn:
> i vaguely recall when that logic was added and also recall the "-1"s from people when "just allowing loopback" was discussed.
> 
> so one question here would be whether we want to change the logic and just allow loopback addresses. but if we do that we have to somehow preserve the current bahavior "if loopback address -> ask LO interface -> pick global address".

It seems even kubeadm allows loopback address, kube-apiserver validation doesn't allow it (when it tried to sync `kubernetes` service: https://github.com/kubernetes/kubernetes/blob/543893cbb0948bf5e4fad5eb71d8e70537453698/pkg/apis/core/validation/validation.go#L6056-L6076
```
2022-05-09T15:10:24.123430406Z stderr F W0509 15:10:24.123176       1 lease.go:250] Resetting endpoints for master service "kubernetes" to [127.0.0.1]
2022-05-09T15:10:24.128487062Z stderr F E0509 15:10:24.128276       1 controller.go:241] unable to sync kubernetes service: Endpoints "kubernetes" is invalid: subsets[0].addresses[0].ip: Invalid value: "127.0.0.1": may not be in the loopback range (127.0.0.0/8, ::1/128)
```

neolit123:
> It seems even kubeadm allows loopback address, kube-apiserver validation doesn't allow it (when it tries to sync kubernetes service):

Does this mean that even if we use kubeadm  phases or patches, apiserver will just fail if we use loopback address?

aojea:
I lost track and could not read all messages

> , kube-apiserver validation doesn't allow it (when it tries to sync kubernetes service):

that is for services and endpoints, those IPs should be reachable, it's orthogonal to what we want to do , that is tell controller-manager and scheduler to use the 127.0.0.1 to communicate with the apiserver.

The thing is that the apiserver address is obtained from `localAPIEndpoint.advertiseAddress` ... but we should not touch that, just add the way of telling those components to use localhost to reach apiserver, because we know that those components can communicate through localhost because are in the same host

BenTheElder:
> I lost track and could not read all messages, so I may miss some context

Incidentally, this is why our contributing guide asks that features first be discussed in issues prior to implementation, it's hard to keep up with review + approach discussion on PRs.

I think this is also the case with the kubeadm OWNERS.

I really appreciate the effort on this though.

pkg/cluster/internal/create/actions/config/config.go
```diff
@@ -241,6 +252,31 @@ func getKubeadmConfig(cfg *config.Cluster, data kubeadm.ConfigData, node nodes.N
...
+	if err != nil {
+		return nil, err
+	}
+
+	patches := map[string]string{}
+	// Kubernetes older than v1.25 don't support patching kubeconfig files.
+	if ver.AtLeast(version.MustParseSemantic("v1.25.0")) {
+		// controller-manager and scheduler connect to local API endpoint, which defaults to the advertise address of the
+		// API server. If the Node's IP changes (which could happen after docker restarts), the server address in KubeConfig
+		// should be updated. However, the server certificate isn't valid for the new IP. To resolve it, we update the
```
BenTheElder, line 267: we should probably fix *this*, perhaps in the entrypoint we can re-roll the cert if it exists?

pkg/cluster/internal/create/actions/config/config.go
```diff
@@ -241,6 +252,31 @@ func getKubeadmConfig(cfg *config.Cluster, data kubeadm.ConfigData, node nodes.N
...
+		// address to loopback address which is an alternative address of the certificate.
+		loopbackAddress := "127.0.0.1"
+		if data.IPFamily == config.IPv6Family {
+			loopbackAddress = "[::1]"
+		}
+		jsonPatch := fmt.Sprintf(`[{"op": "replace", "path": "/clusters/0/cluster/server", "value": "https://%s:%d"}]`, loopbackAddress, data.APIBindPort)
```
aojea, line 273:
```suggestion
		loopbackAddress := "127.0.0.1"
		if data.IPFamily == config.IPv6Family {
			loopbackAddress = "::1"
		}
		jsonPatch := fmt.Sprintf(`[{"op": "replace", "path": "/clusters/0/cluster/server", "value": "https://%s"}]`, net.JoinHostPort(loopbackAddress, strconv.Itoa(data.APIBindPort)))
```

pkg/cluster/internal/create/actions/config/config.go
tnqn: Thanks for the suggestion. It reminds me that since we have the CA cert and key, we could just re-roll the cert and replace the server address in kubeconfig with the new Node IP like the existing magic: https://github.com/kubernetes-sigs/kind/blob/1a6bfa7576ad06903f72fd7c2e7f5d510296bef0/images/base/files/usr/local/bin/entrypoint#L413-L414
Then everything should just work in theory. Do you think this is a better solution?

tnqn: I tried the above approach and it worked as expected, created issue https://github.com/kubernetes-sigs/kind/issues/2759 to discuss the approaches.

BenTheElder:
#2775, thanks!

SiegfredRodriguez:
This seems to still occur on v0.18.0, is there a specific usage change needed to use #2775  ?

BenTheElder:
@SiegfredRodriguez please file a bug with cluster logs etc unless it's #1689 which is a known limitation.

PR #2674, opened 2022-03-14
Switch to registry.k8s.io

ameukam:
Related:
  - https://github.com/kubernetes/k8s.io/issues/3411

`registry.k8s.io` is for the moment a proxy in front `k8s.gcr.io`.
There should be no impact.

Signed-off-by: Arnaud Meukam <ameukam@gmail.com>


ameukam:
> The thing is that this not only changes k8s ci, changes all the ecosystem using kind , that is considerable

@aojea we don't have to merge this right now. i don't really know what are the criteria for `kind` to make this mergeable.

/hold

aojea:
> > The thing is that this not only changes k8s ci, changes all the ecosystem using kind , that is considerable
> 
> @aojea we don't have to merge this right now. i don't really know what are the criteria for `kind` to make this mergeable.
> 
> /hold

ahh cool :smile:

BenTheElder:
maybe we should have some env to override the source registry. Not sure how it should apply to kindest and rancher though.

If we're ready to roll this out for real we can hard code it, but definitely some users will have to update proxies / mirrors.

It would also be possible to do this in CI by just changing the containerd config in the base image we use to use registry.k8s.io as a mirror and just changing our CI to use this. All images are pulled within a container from this image except that image itself which is on kindest.

dims:
> It would also be possible to do this in CI by just changing the containerd config in the base image we use to use registry.k8s.io as a mirror and just changing our CI to use this. All images are pulled within a container from this image except that image itself which is on kindest.

I like this one ben. @ameukam wanna poke at this option?

aojea:
> maybe we should have some env to override the source registry. Not sure how it should apply to kindest and rancher though.

+1 EXPERIMENTAL_KUBE_REGISTRY or similar

pkg/build/nodeimage/const_storage.go
```diff
@@ -25,7 +25,7 @@ NOTE: we have customized it in the following ways:
 - install as the default storage class
 */
 
-var defaultStorageImages = []string{"docker.io/rancher/local-path-provisioner:v0.0.14", "k8s.gcr.io/build-image/debian-base:buster-v1.7.2"}
+var defaultStorageImages = []string{"docker.io/rancher/local-path-provisioner:v0.0.14", "registry.k8s.io/build-image/debian-base:buster-v1.7.2"}
```
aojea, line 28: this is hardcoded in the manifest below

aojea: @BenTheElder we may fork the storage provisionser https://github.com/rancher/local-path-provisioner, it seems similar to kindnet, is only 3 go files

images/base/files/etc/containerd/config.toml
```diff
@@ -27,7 +27,7 @@ version = 2
 
 [plugins."io.containerd.grpc.v1.cri"]
   # use fixed sandbox image
-  sandbox_image = "k8s.gcr.io/pause:3.6"
+  sandbox_image = "registry.k8s.io/pause:3.6"
```
aojea, line 30: @BenTheElder we should be able to patch this for CI from the config, right?

```
kind: Cluster
nodes:
  - role: control-plane
containerdConfigPatches:
  - |-
    [plugins."io.containerd.grpc.v1.cri".]
      sandbox_image = "registry.k8s.io/pause:3.6"
```

aojea: https://github.com/kubernetes-sigs/kind/blob/4cd81b46b9b6e3015ca64331a6e24e0c301c7e7b/hack/ci/e2e-k8s.sh#L138-L152

pkg/build/nodeimage/const_storage.go
aojea: WDYT?

images/base/files/etc/containerd/config.toml
BenTheElder: > @BenTheElder we should be able to patch this for CI from the config, right?

yes, but that won't apply to image builds.

pkg/build/nodeimage/const_storage.go
aojea: > docker.io/rancher/local-path-provisioner:v0.0.14

they are in 0.21 release and we have 0.14, also, it doesn't seem to update golang periodically

https://github.com/rancher/local-path-provisioner/commit/3436e3cf83ed0c43f3a2f92f53a5f698718d640a

it is a bit more for having a better control on this dependency we are deliverying ,

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/2688 implements the approach we discussed

ameukam:
> #2688 implements the approach we discussed most recently

/close
/honk

PR #2676, opened 2022-03-16
docs(READEME): update kind version

zaunist:
Currently, we release 0.12.0, so we should update the `READEME.md`

### DIFF
--- a/site/content/docs/user/configuration.md
+++ b/site/content/docs/user/configuration.md
@@ -140,7 +140,7 @@ networking:
 {{< /codeFromInline >}}
 
 ##### Dual Stack clusters
-You can run dual stack clusters using `kind` 0.11+, on kubernetes versions 1.20+.
+You can run dual stack clusters using `kind` 0.12+, on kubernetes versions 1.20+.
 
 {{< codeFromInline lang="yaml" >}}
 kind: Cluster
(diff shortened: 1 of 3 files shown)


zaunist:
Hi, community.

I noticed our `site` docs need be update too, do I update it together in this PR or in a new PR?

tao12345666333:
Please complete the update in this PR, thanks.

zaunist:
I have updated the docs, please take a look, thanks.

aojea:
/ok-to-test
/lgtm
/approve
Thanks

site/content/docs/user/configuration.md
```diff
@@ -140,7 +140,7 @@ networking:
 {{< /codeFromInline >}}
 
 ##### Dual Stack clusters
-You can run dual stack clusters using `kind` 0.11+, on kubernetes versions 1.20+.
+You can run dual stack clusters using `kind` 0.12+, on kubernetes versions 1.20+.
```
BenTheElder, line 143: this change is wrong, this feature was not introduced in v0.12.0

aojea: I overlooked this , sorry,

Ben is right , is not about being semantically correct , this section talks about the dual stack feature in kind, that is 0.11 ... just revert this change to 0.11+

zaunist: Sorrry, I will revert this change later.

PR #2677, opened 2022-03-16
Generate a list of images in a log file

dims:
If we log the images, we can peek to ensure that they are coming from where we think they are should be from!


dims:
hmm, this logs images outside of kind...

PR #2688, opened 2022-03-21
enable registry.k8s.io

BenTheElder:
This will cause kind to pull from registry.k8s.io first before k8s.gcr.io directly if that fails.

We use this configuration mechanism for the local registry doc.

See: https://github.com/containerd/containerd/blob/main/docs/cri/registry.md#configure-registry-endpoint


images/base/files/etc/containerd/config.toml
```diff
@@ -33,3 +33,8 @@ version = 2
   tolerate_missing_hugepages_controller = true
   # restrict_oom_score_adj needs to be true when running inside UserNS (rootless)
   restrict_oom_score_adj = false
+
+# Enable registry.k8s.io as the primary mirror for k8s.gcr.io
+# See: https://github.com/kubernetes/k8s.io/issues/3411
+[plugins."io.containerd.grpc.v1.cri".registry.mirrors."k8s.gcr.io"]
+  endpoint = ["https://registry.k8s.io", "https://k8s.gcr.io.io",]
```
neolit123, line 40: i guess it should be `https://k8s.gcr.io`.

neolit123 (commented):
/lgtm

images/base/files/etc/containerd/config.toml
BenTheElder: thanks, corrected, re-pushing

BenTheElder:
/hold cancel
re-pushed

dims:
@BenTheElder if you use `crictl images`  would we see actual or the mirror?

BenTheElder:
> @BenTheElder if you use crictl images would we see actual or the mirror?

You'll still see images reported as `k8s.gcr.io` regardless of where they're pulled.

The flip side is *any* image specifying to be from k8s.gcr.io should hit the mirror first and only k8s.gcr.io directly if registry.k8s.io fails, including e.g. manifests deploying ingress-nginx etc.
 
To verify where images are pulled, containerd debug logs should suffice (not sure if we currently have default logging turned high enough).

BenTheElder:
/retest

flaky e2e?

> Kubernetes e2e suite: [sig-network] Proxy version v1 A set of valid responses are returned for both pod and service ProxyWithPath [Conformance] | Kubernetes e2e suite: [sig-network] Proxy version v1 A set of valid responses are returned for both pod and service ProxyWithPath [Conformance]

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/2688/pull-kind-conformance-parallel-ga-only/1506027322203967488

BenTheElder:
/hold
let's also bump debugging so we can see this.

images/base/files/etc/containerd/config.toml
```diff
@@ -1,6 +1,10 @@
 # explicitly use v2 config format
 version = 2
 
+# log lots of detail
+[debug]
+  level = "trace"
```
BenTheElder, line 6: https://github.com/containerd/containerd/blob/77d53d2d230c3bcd3f02e6f493019a72905c875b/cmd/containerd/command/main.go#L337

BenTheElder:
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/2688/pull-kind-e2e-kubernetes/1506042308267610112/artifacts/logs/kind-worker2/containerd.log

we can see from trace level logging that it is indeed pulling via registry.k8s.io

I think we should probably go back to debug level logging or lower though, AFAICT most of CI is using info level for containerd, and trace level logs are very chatty, my browser craps out trying to load that log 🙃 

/hold

aojea:
> I think we should probably go back to debug level logging or lower though

yes please, most sure that also decrease the performance

dims:
@BenTheElder Nice! +1 to go back to previous level

BenTheElder:
reverted log level change, re-bumped base image

ameukam:
/lgtm

Ref: https://github.com/kubernetes/k8s.io/issues/3411

PR #2690, opened 2022-03-23
Add workarounds for Docker Desktop cgroupv2

stmcginnis:
This adds some detail to the known issues for those using Docker Desktop
and trying to deploy a k8s version prior to 1.19. Newer Docker Desktop
releases instantiate a VM that defaults to cgroupv2, which was not
supported by those k8s releases.

This gives the user the options of either downgrading their Docker
Desktop to an older release that still used cgroupv1, or using the newer
configuration option introduced in Docker Desktop 4.4.2 for forcing the
VM back to cgroupv1.


BenTheElder (approved):
/lgtm
/approve
thanks!

PR #2696, opened 2022-03-28
add CreateWithName func

yxxhero:
Signed-off-by: yxxhero <aiopsclub@163.com>

add CreateWithName func


aojea:
is this solving any problem, do we need this?

yxxhero:
TODO: move name to an option to override config  @aojea

aojea:
> TODO: move name to an option to override config @aojea

:+1: 
/ok-to-test

yxxhero:
@aojea

pkg/cluster/createoption_test.go
```diff
@@ -0,0 +1,37 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
stmcginnis, line 2: Should probably say 2022 since you are adding a new file.

pkg/cluster/provider.go
```diff
@@ -169,12 +169,9 @@ func ProviderWithPodman() ProviderOption {
 }
 
 // Create provisions and starts a kubernetes-in-docker cluster
-// TODO: move name to an option to override config
-func (p *Provider) Create(name string, options ...CreateOption) error {
+func (p *Provider) Create(options ...CreateOption) error {
```
stmcginnis, line 172: This breaks an API contract, and there are some folks that use `kind` as a library to create clusters, so I don't think this is a good idea. We need to keep backward compatibility so we don't break any consumers of this, so we either need to deprecate the current `Create` call, and introduce something new like `CreateWithOpts` or something along those lines. Or I suppose basically the same, but don't deprecate the `Create` call and just provide two different ways to do this.

yxxhero: @stmcginnis you are right. I will fix it.

yxxhero:
@stmcginnis fixed.

pkg/cluster/provider.go
```diff
@@ -169,7 +169,7 @@ func ProviderWithPodman() ProviderOption {
 }
 
 // Create provisions and starts a kubernetes-in-docker cluster
-// TODO: move name to an option to override config
+// TODO: Replace by CreateWithOpts in the next major version
 func (p *Provider) Create(name string, options ...CreateOption) error {
 	// apply options
 	opts := &internalcreate.ClusterOptions{
```
stmcginnis, line 175: Just commenting this as an option to think about if you need to do any further updates - it might be nice to now make `Create` just be a light wrapper that takes the separate `name` argument and use `CreateWithName` to just call out to `CreateWithOpts` and force everything through the one common code path.

Not a big deal at all though, so feel free to ignore this. ;)

stmcginnis (commented):
/ok-to-test

pkg/cluster/provider.go
yxxhero: ok. Thanks very much.

pkg/cluster/provider.go
BenTheElder: Why are we breaking this?

pkg/cluster/provider.go
```diff
@@ -169,7 +169,7 @@ func ProviderWithPodman() ProviderOption {
 }
 
 // Create provisions and starts a kubernetes-in-docker cluster
-// TODO: move name to an option to override config
+// TODO: Replace by CreateWithOpts in the next major version
```
BenTheElder, line 172: We don't have major releases yet, kind is still in alpha.

We should be striving to avoid this though, we should probably just remove this TODO instead.

`name` is the primary key for the cluster, and API callers really ought to be keeping track of a `name` so they can handle deletion etc., I think we should just drop this TODO, avoid breaking callers, and acknowledge that all cluster calls must supply name.

BenTheElder:
Thank you for the PR!

I see that we left a TODO for this, but I think at this point we're going to break a lot of callers for no real gain and worse moving name into an option obscures that you really need to keep track of name as the caller of this API in order to do any further operations with the cluster. Many applications using this API should probably *not* be relying on the default name (and e.g. potentially deleting a cluster they didn't actually create).

I think we should just delete the TODO and leave the code as-is.

yxxhero:
> Thank you for the PR!
> 
> I see that we left a TODO for this, but I think at this point we're going to break a lot of callers for no real gain and worse moving name into an option obscures that you really need to keep track of name as the caller of this API in order to do any further operations with the cluster. Many applications using this API should probably _not_ be relying on the default name (and e.g. potentially deleting a cluster they didn't actually create).
> 
> I think we should just delete the TODO and leave the code as-is.

ok. I will fix it.

go.sum
```diff
@@ -10,11 +10,6 @@ github.com/evanphx/json-patch/v5 v5.6.0/go.mod h1:G79N1coSVB93tBe7j6PhzjmR3/2Vvl
 github.com/inconshreveable/mousetrap v1.0.0 h1:Z8tu5sraLXCXIcARxBp/8cbvlwVa7Z1NHg9XEKhtSvM=
 github.com/inconshreveable/mousetrap v1.0.0/go.mod h1:PxqpIevigyE2G7u3NXJIT2ANytuPF1OarO4DADm73n8=
 github.com/jessevdk/go-flags v1.4.0/go.mod h1:4FA24M0QyGHXBuZZK/XkWh8h0e1EYbRYJSGM75WSRxI=
-github.com/kr/pretty v0.1.0 h1:L/CwN0zerZDmRFUapSPitk6f+Q3+0za1rQkzVuMiMFI=
-github.com/kr/pretty v0.1.0/go.mod h1:dAy3ld7l9f0ibDNOQOHHMYYIIbhfbHSm3C4ZsoJORNo=
-github.com/kr/pty v1.1.1/go.mod h1:pFQYn66WHrOpPYNljwOMqo10TkYh1fy3cYio2l3bCsQ=
-github.com/kr/text v0.1.0 h1:45sCR5RtlFHMR4UwH9sdQ5TC8v0qDQCHnXt+kaKSTVE=
-github.com/kr/text v0.1.0/go.mod h1:4Jbv+DJW3UT/LiOwJeYQe1efqtUx/iVham/4vfdArNI=
```
aojea, line 17: the go.sum doesn't have to be modified

yxxhero: changed by go mod tidy. I will fix it.

aojea:
/lgtm
/approve
Thanks

PR #2699, opened 2022-03-29
use switch case in DetectNodeProvider

yxxhero:
Signed-off-by: yxxhero <aiopsclub@163.com>

use switch case in DetectNodeProvider

### DIFF
--- a/pkg/cluster/provider.go
+++ b/pkg/cluster/provider.go
@@ -113,13 +113,14 @@ var NoNodeProviderDetectedError = errors.NewWithoutStack("failed to detect any s
 // that logic will be in a public API as well.
 func DetectNodeProvider() (ProviderOption, error) {
 	// auto-detect based on each node provider's IsAvailable() function
-	if docker.IsAvailable() {
+	switch {
+	case docker.IsAvailable():
 		return ProviderWithDocker(), nil
-	}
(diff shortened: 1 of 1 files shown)


yxxhero:
@aojea ping

BenTheElder (commented):
What improvement is this change expected to bring?

I think this is less clear than the original code, which explicitly checks in a particular order (and should continue to do so for backwards compatibility).

yxxhero:
@BenTheElder 
1.I think multi branch is clearer, and the priority is from top to bottom
2.It is convenient to add more nodeproviders and reduce repeated if
3.This code has no compatibility problem and is backward compatible

WDTF?

BenTheElder:
1. I don't think this is any clearer than imperative if { return}.

2.  I don't expect to be adding many (realistically any in the immediate future) providers anyhow, but also it is not really any less typing.

3. The key part -- Yes, but I think it is less obvious once switching to switch that the ordering must be preserved by future PRs to this code. Switch case is typically used for exhaustive-ness with ordering as an afterthought, bit these method calls are not necessarily mutually exclusive.

yxxhero:
@BenTheElder 
ok. I think you are right. Thanks for your reply.

PR #2709, opened 2022-04-06
entrypoint: fix chicken-and-egg runtime problem

kolyshkin:
In case the runtime used to run the KIND container is not aware of some
cgroup subsystems, those will be exposed to the container without proper
scoping (note the rdma and misc):

	kir@ubu2110:~$ sudo docker run -i --rm --privileged ubuntu sh -xc 'cat /proc/self/cgroup; grep cgroup /proc/self/mountinfo'
	+ cat /proc/self/cgroup
	13:pids:/docker/c1f3fc37b0d6e5a109c62e861feb4d6fd4ef381bf5a9576e5e7c56da4eca841b
	12:net_cls,net_prio:/docker/c1f3fc37b0d6e5a109c62e861feb4d6fd4ef381bf5a9576e5e7c56da4eca841b
	11:hugetlb:/docker/c1f3fc37b0d6e5a109c62e861feb4d6fd4ef381bf5a9576e5e7c56da4eca841b
	10:misc:/
	9:freezer:/docker/c1f3fc37b0d6e5a109c62e861feb4d6fd4ef381bf5a9576e5e7c56da4eca841b
	8:devices:/docker/c1f3fc37b0d6e5a109c62e861feb4d6fd4ef381bf5a9576e5e7c56da4eca841b
	7:cpu,cpuacct:/docker/c1f3fc37b0d6e5a109c62e861feb4d6fd4ef381bf5a9576e5e7c56da4eca841b
	6:perf_event:/docker/c1f3fc37b0d6e5a109c62e861feb4d6fd4ef381bf5a9576e5e7c56da4eca841b
	5:memory:/docker/c1f3fc37b0d6e5a109c62e861feb4d6fd4ef381bf5a9576e5e7c56da4eca841b
	4:blkio:/docker/c1f3fc37b0d6e5a109c62e861feb4d6fd4ef381bf5a9576e5e7c56da4eca841b
	3:rdma:/
	2:cpuset:/docker/c1f3fc37b0d6e5a109c62e861feb4d6fd4ef381bf5a9576e5e7c56da4eca841b
	1:name=systemd:/docker/c1f3fc37b0d6e5a109c62e861feb4d6fd4ef381bf5a9576e5e7c56da4eca841b
	0::/system.slice/containerd.service
	+ grep cgroup /proc/self/mountinfo
	666 665 0:65 / /sys/fs/cgroup rw,nosuid,nodev,noexec,relatime - tmpfs tmpfs rw,mode=755,inode64
	667 666 0:32 /docker/c1f3fc37b0d6e5a109c62e861feb4d6fd4ef381bf5a9576e5e7c56da4eca841b /sys/fs/cgroup/systemd rw,nosuid,nodev,noexec,relatime master:11 - cgroup cgroup rw,xattr,name=systemd
	668 666 0:35 /docker/c1f3fc37b0d6e5a109c62e861feb4d6fd4ef381bf5a9576e5e7c56da4eca841b /sys/fs/cgroup/cpuset rw,nosuid,nodev,noexec,relatime master:15 - cgroup cgroup rw,cpuset
	669 666 0:36 / /sys/fs/cgroup/rdma rw,nosuid,nodev,noexec,relatime master:16 - cgroup cgroup rw,rdma
	670 666 0:37 /docker/c1f3fc37b0d6e5a109c62e861feb4d6fd4ef381bf5a9576e5e7c56da4eca841b /sys/fs/cgroup/blkio rw,nosuid,nodev,noexec,relatime master:17 - cgroup cgroup rw,blkio
	671 666 0:38 /docker/c1f3fc37b0d6e5a109c62e861feb4d6fd4ef381bf5a9576e5e7c56da4eca841b /sys/fs/cgroup/memory rw,nosuid,nodev,noexec,relatime master:18 - cgroup cgroup rw,memory
	672 666 0:39 /docker/c1f3fc37b0d6e5a109c62e861feb4d6fd4ef381bf5a9576e5e7c56da4eca841b /sys/fs/cgroup/perf_event rw,nosuid,nodev,noexec,relatime master:19 - cgroup cgroup rw,perf_event
	673 666 0:40 /docker/c1f3fc37b0d6e5a109c62e861feb4d6fd4ef381bf5a9576e5e7c56da4eca841b /sys/fs/cgroup/cpu,cpuacct rw,nosuid,nodev,noexec,relatime master:20 - cgroup cgroup rw,cpu,cpuacct
	674 666 0:41 /docker/c1f3fc37b0d6e5a109c62e861feb4d6fd4ef381bf5a9576e5e7c56da4eca841b /sys/fs/cgroup/devices rw,nosuid,nodev,noexec,relatime master:21 - cgroup cgroup rw,devices
	675 666 0:42 /docker/c1f3fc37b0d6e5a109c62e861feb4d6fd4ef381bf5a9576e5e7c56da4eca841b /sys/fs/cgroup/freezer rw,nosuid,nodev,noexec,relatime master:22 - cgroup cgroup rw,freezer
	676 666 0:43 / /sys/fs/cgroup/misc rw,nosuid,nodev,noexec,relatime master:23 - cgroup cgroup rw,misc
	677 666 0:44 /docker/c1f3fc37b0d6e5a109c62e861feb4d6fd4ef381bf5a9576e5e7c56da4eca841b /sys/fs/cgroup/hugetlb rw,nosuid,nodev,noexec,relatime master:24 - cgroup cgroup rw,hugetlb
	678 666 0:45 /docker/c1f3fc37b0d6e5a109c62e861feb4d6fd4ef381bf5a9576e5e7c56da4eca841b /sys/fs/cgroup/net_cls,net_prio rw,nosuid,nodev,noexec,relatime master:25 - cgroup cgroup rw,net_cls,net_prio
	679 666 0:46 /docker/c1f3fc37b0d6e5a109c62e861feb4d6fd4ef381bf5a9576e5e7c56da4eca841b /sys/fs/cgroup/pids rw,nosuid,nodev,noexec,relatime master:26 - cgroup cgroup rw,pids

Now, if a newer runtime (the one that is aware of e.g. rdma subsystem)
will be used inside this container, it may create cgroups under those
subsystems. Since those are not properly scoped, they will leak to the
host and thus will become non-removable (EBUSY on rmdir).

The workaround, as implemented here, is to hide (unmount and remove)
such unscoped subsystemd.

Fixes https://github.com/kubernetes/kubernetes/issues/109182

### DIFF
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -217,6 +217,28 @@ fix_cgroup() {
   current_cgroup=$(grep -E '^[^:]*:([^:]*,)?cpu(,[^,:]*)?:.*' /proc/self/cgroup | cut -d: -f3)
   local cgroup_subsystems
   cgroup_subsystems=$(findmnt -lun -o source,target -t cgroup | grep "${current_cgroup}" | awk '{print $2}')
+  # Unmount the cgroup subsystems that are not known to runtime used to
+  # run the container we are in. Those subsystems are not properly scoped
+  # (i.e. the root cgroup is exposed, rather than something like docker/xxxx).
+  # In case a runtime (which is aware of more subsystems -- such as rdma,
+  # misc, or unified) is used inside the container, it may create cgroups for
+  # these subsystems, and as they are not scoped, they will leak to the host
+  # and thus will become non-removable.
+  #
+  # See https://github.com/kubernetes/kubernetes/issues/109182
+  local unsupported_cgroups
+  unsupported_cgroups=$(findmnt -lun -o source,target -t cgroup | grep -v "${current_cgroup}" | awk '{print $2}')
+  if [ -n "$unsupported_cgroups" ]; then
+    local mnt
+    echo "$unsupported_cgroups" |
+    while IFS= read -r mnt; do
+      echo "INFO: unmounting and removing $mnt"
+      umount "$mnt" || true
+      rmdir "$mnt" || true
+    done
+  fi
+
+
   # For each cgroup subsystem, Docker does a bind mount from the current
   # cgroup to the root of the cgroup subsystem. For instance:
   #   /sys/fs/cgroup/memory/docker/<cid> -> /sys/fs/cgroup/memory
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -318,8 +318,8 @@ fix_kmsg() {
 select_iptables() {
   # based on: https://github.com/kubernetes-sigs/iptables-wrappers/blob/97b01f43a8e8db07840fc4b95e833a37c0d36b12/iptables-wrapper-installer.sh
   local mode num_legacy_lines num_nft_lines
-  num_legacy_lines=$( (iptables-legacy-save || true; ip6tables-legacy-save || true) 2>/dev/null | grep '^-' | wc -l || true)
-  num_nft_lines=$( (timeout 5 sh -c "iptables-nft-save; ip6tables-nft-save" || true) 2>/dev/null | grep '^-' | wc -l || true)
+  num_legacy_lines=$( (iptables-legacy-save || true; ip6tables-legacy-save || true) 2>/dev/null | grep -c '^-' || true)
+  num_nft_lines=$( (timeout 5 sh -c "iptables-nft-save; ip6tables-nft-save" || true) 2>/dev/null | grep -c '^-' || true)
   if [ "${num_legacy_lines}" -ge "${num_nft_lines}" ]; then
     mode=legacy
   else
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -203,18 +203,31 @@ fix_cgroup() {
     return
   fi
   echo 'INFO: detected cgroup v1'
-  echo 'INFO: fix cgroup mounts for all subsystems'
+  local current_cgroup
+  current_cgroup=$(grep -E '^[^:]*:([^:]*,)?cpu(,[^,:]*)?:.*' /proc/self/cgroup | cut -d: -f3)
+  if [ "$current_cgroup" = "/" ]; then
+    echo "INFO: cgroupns detected, no need to fix cgroups"
+    return
+  fi
+
+  # NOTE The rest of this function deals with the unfortunate situation of
+  # cgroup v1 with no cgroupns enabled. One fine day every user will have
+  # cgroupns enabled (or switch or cgroup v2 which has it enabled by default).
+  # Once that happens, this function can be removed completely.
+
+  echo 'WARN: cgroupns not enabled! Please use cgroup v2, or cgroup v1 with cgroupns enabled.'
+
   # See: https://d2iq.com/blog/running-kind-inside-a-kubernetes-cluster-for-continuous-integration
   # Capture initial state before modifying
   #
   # Basically we're looking for the cgroup-path for the cpu controller for the
   # current process. this tells us what cgroup-path the container is in.
   # Then we collect the subsystems that are active on this path.
-  # We assume the cpu controller is in use on all node containers.
+  # We assume the cpu controller is in use on all node containers,
+  # and other controllers use the same sub-path.
   #
   # See: https://man7.org/linux/man-pages/man7/cgroups.7.html
-  local current_cgroup
-  current_cgroup=$(grep -E '^[^:]*:([^:]*,)?cpu(,[^,:]*)?:.*' /proc/self/cgroup | cut -d: -f3)
+  echo 'INFO: fix cgroup mounts for all subsystems'
   local cgroup_subsystems
   cgroup_subsystems=$(findmnt -lun -o source,target -t cgroup | grep "${current_cgroup}" | awk '{print $2}')
   # Unmount the cgroup subsystems that are not known to runtime used to
--- a/pkg/build/nodeimage/defaults.go
+++ b/pkg/build/nodeimage/defaults.go
@@ -20,4 +20,4 @@ package nodeimage
 const DefaultImage = "kindest/node:latest"
 
 // DefaultBaseImage is the default base image used
-const DefaultBaseImage = "docker.io/kindest/base:v20220324-837a3da8"
+const DefaultBaseImage = "docker.io/kindest/base:v20220413-d2354dce"


kolyshkin:
I briefly tested the code locally:
```console
kir@ubu2110:~$ sudo docker info
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.8.1-docker)
  scan: Docker Scan (Docker Inc., v0.17.0)

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 1
 Server Version: 20.10.14
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 3df54a852345ae127d1fa3092b95168e4a88e2f8
 runc version: v1.0.3-0-gf46b6ba
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: default
 Kernel Version: 5.13.0-39-generic
 Operating System: Ubuntu 21.10
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 3.832GiB
 Name: ubu2110
 ID: 4DH7:JJYA:QRUK:3M6K:764C:N5XI:RXZL:3PK3:RZFV:LLSJ:KHL6:D4PQ
 Debug Mode: false
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false

kir@ubu2110:~$ cat xxx
function fix_cgroup() {
  local current_cgroup
  current_cgroup=$(grep -E '^[^:]*:([^:]*,)?cpu(,[^,:]*)?:.*' /proc/self/cgroup | cut -d: -f3)
  local cgroup_subsystems
  cgroup_subsystems=$(findmnt -lun -o source,target -t cgroup | grep "${current_cgroup}" | awk '{print $2}')
  # Unmount the cgroup subsystems that are not known to runtime used to
  # run the container we are in. Those subsystems are not properly scoped
  # (i.e. the root cgroup is exposed, rather than something like docker/xxxx).
  # In case a runtime (which is aware of more subsystems -- such as rdma,
  # misc, or unified) is used inside the container, it may create cgroups for
  # these subsystems, and as they are not scoped, they will leak to the host
  # and thus will become non-removable.
  #
  # See https://github.com/kubernetes/kubernetes/issues/109182
  local unsupported_cgroups
  unsupported_cgroups=$(findmnt -lun -o source,target -t cgroup | grep -v "${current_cgroup}" | awk '{print $2}')
  local mnt
  echo "$unsupported_cgroups" |
  while IFS= read -r mnt; do
    echo "INFO: unmounting and removing $mnt"
    umount "$mnt" || true
    rmdir "$mnt" || true
  done
}

fix_cgroup
kir@ubu2110:~$ cat xxx | sudo docker run -i --rm --privileged ubuntu bash
INFO: unmounting and removing /sys/fs/cgroup/rdma
INFO: unmounting and removing /sys/fs/cgroup/misc
```

kolyshkin:
A few important things worth noting/reiterating:

1. This workaround only makes sense for cgroup v1. For v2, we use cgroupns.

2. Cgroup v1 setups can probably also use cgroupns to eliminate the issue (and in this case the whole `fix_cgroup` won't be needed!), but I haven't looked into it.

3. This workaround only makes sense in a situation when the runtime used to create a KIND container is sufficiently newer than the runtime used inside the container (as in, runc 1.0.x on the host and runc 1.1.x inside a KIND container). If the inner/outer runtimes are the same, subsystems unknown to them are not used. If the outer runtime is newer, inner runtime still doesn't know about new subsystems. OTOH having it does no harm.

4. The existing issue with `rdma` and `unified` subsystems may repeat in the future with the `misc` subsystemd if/once runc starts to support it. Same would happen with any other new cgroup v1 subsystem that kernel might want to add (and runc -- to support).  So this workaround still makes sense to have, basically eliminating any non-scoped subsystems from the container.

5. For some reason, I do not see `/sys/fs/cgroup/unified` (aka hybrid hierarchy) being mounted in my setup, but it seems it is the case for kubernetes CI (as unified appears to be mounted). Thus, I haven't checked if my code fixes the issue with `unified`, but I believe it is.

6. I also checked it with podman+crun (as installed from the default ubuntu repo) on the same Ubuntu 21.10 machine. In there, all the cgroups (even misc) are properly scoped. So, yet another possible solution to this problem is to replace docker with podman (on the host). Might bring other issues, of course.

images/base/files/usr/local/bin/entrypoint
```diff
@@ -203,20 +203,47 @@ fix_cgroup() {
...
   # Then we collect the subsystems that are active on this path.
-  # We assume the cpu controller is in use on all node containers.
+  # We assume the cpu controller is in use on all node containers,
+  # and other controllers use the same sub-path.
   #
   # See: https://man7.org/linux/man-pages/man7/cgroups.7.html
   local current_cgroup
   current_cgroup=$(grep -E '^[^:]*:([^:]*,)?cpu(,[^,:]*)?:.*' /proc/self/cgroup | cut -d: -f3)
+  if [ "$current_cgroup" = "/" ]; then
+    echo "INFO: cgroupns detected, no need to fix cgroups"
```
AkihiroSuda, line 219: For simplicity I'd suggest always enabling `--cgroupns=private`.
This is not present in Docker 19.03 CLI, but 19.03 has already reached EOL.

kolyshkin:
Added another commit with detects cgroupns on cgroup v1 and skips the cgroup setup.

Indeed, using `cgroupns=private` should
 - eliminate the issue we're fixing here
 - eliminate the need for complex cgroup setup.

We could go one step further and **require cgroupns** instead.

images/base/files/usr/local/bin/entrypoint
kolyshkin: I concur (see https://github.com/kubernetes-sigs/kind/pull/2709#issuecomment-1092350339 which I wrote yesterday but forgot to click "Comment").

kolyshkin: So, we can document that KIND now requires `--cgroupns=private`, and replace the whole `fix_cgroup` with a check that we got a private cgroupns.

kolyshkin: I'm not sure though it if will work for other environments. It does work for podman (checked with a version bundled with Ubuntu 21.10). The script also mentions "Cloud Shell" (not sure what's that but seems to be k8s based) and GitHub Actions.

In fact, I think I know what to do. Please see the last commit (just updated).

images/base/files/usr/local/bin/entrypoint
```diff
@@ -203,20 +203,53 @@ fix_cgroup() {
...
   echo 'INFO: detected cgroup v1'
-  echo 'INFO: fix cgroup mounts for all subsystems'
+  local current_cgroup
+  current_cgroup=$(grep -E '^[^:]*:([^:]*,)?cpu(,[^,:]*)?:.*' /proc/self/cgroup | cut -d: -f3)
+  if [ "$current_cgroup" = "/" ]; then
+    echo "INFO: cgroupns detected, no need to fix cgroups"
+    return
+  fi
+
+  echo 'WARNING: cgroupns is not enabled! Please run with --cgroupns=private,'
```
AkihiroSuda, line 213: This might be confusing to end users because the `kind` CLI does not have `--cgroupns` flag.

AkihiroSuda: Maybe better to just suggest cgroup v2?

kolyshkin: As kind is used to test kubernetes, we need to support both v1 and v2, otherwise we won't test it.

What about something like this:
```go
  echo 'WARNING: cgroupns is not enabled! Please use cgroup v2 or cgroup v1 with cgroupns enabled'
```

kolyshkin: Modified accordingly, PTAL

images/base/files/usr/local/bin/entrypoint
```diff
@@ -203,18 +203,29 @@ fix_cgroup() {
...
-  echo 'INFO: fix cgroup mounts for all subsystems'
+  local current_cgroup
+  current_cgroup=$(grep -E '^[^:]*:([^:]*,)?cpu(,[^,:]*)?:.*' /proc/self/cgroup | cut -d: -f3)
+  if [ "$current_cgroup" = "/" ]; then
+    echo "INFO: cgroupns detected, no need to fix cgroups"
+    return
+  fi
+
+  echo 'WARN: cgroupns not enabled! Please use cgroup v2, or cgroup v1 with cgroupns enabled.'
+  echo 'WARN: Support for host cgroup namespace will be removed!'
```
BenTheElder, line 214: I'm not sure we're actually going to do that, hacky as it is.
The line above is good, I think this is my only nit with this PR.

BenTheElder:
Thank you very much for looking into this issue and for the PR 🙏 
I only have one complaint https://github.com/kubernetes-sigs/kind/pull/2709#discussion_r847601069, otherwise we should push a new image, test and merge this @aojea

BenTheElder:
[missed clicking "comment" on this earlier]

> We could go one step further and require cgroupns instead.

I would prefer not to break existing users where it is possible to workaround, AFAICT `--cgroupns=private` is a non-default dockerd setting, but at least with time cgroupsv2 will take over anyhow ... 🤞 

https://github.com/kubernetes-sigs/kind/pull/2709/commits/72433e5e0e4d772fd16744aa9c6896b47c51a862 seems like a pretty elegant workaround thanks!

images/base/files/usr/local/bin/entrypoint
kolyshkin: My thinking was that such a message should encourage more users to enable cgroupns, and not force us  to actually do anything (as the message does not specify when such support will be removed).

Anyway, message removed, and I rewrote the TODO item into the NOTE.

PR updated, PTAL @BenTheElder

kolyshkin:
> I would prefer not to break existing users where it is possible to workaround, AFAICT `--cgroupns=private` is a non-default dockerd setting, but at least with time cgroupsv2 will take over anyhow ... :crossed_fingers:

It is probably a non-default for the sake of such workarounds that we have here. In fact, this is yet another chicken and egg problem. I mean, nothing should break when cgroupns is enabled for cgroup v2, except some kludges like this.

In the ideal word, cgroupns should have been enabled by default, and the users would adopt (by either disabling it or removing the workarounds/kludges). I hope this will eventually happen, but remembering the time it took to fix a bad default for container's IPC mode, it can take a few years.

kolyshkin:
Rebased; addressed the review comments.

aojea:
neat, thanks for the great work
/lgtm
/approve

aojea:
> #30 [linux/s390x stage-0 4/6] RUN go mod download
> #0 0.540 Segmentation fault (core dumped)
> #30 ERROR: process "/bin/sh -c go mod download" did not complete successfully: exit code: 139


/test pull-kind-build

images/base/files/usr/local/bin/entrypoint
BenTheElder: +1

BTW, For the record: Kubernetes's CI is unfortunately one of these places still.

PR #2711, opened 2022-04-07
Add ppc64le support

mkumatag:
This PR addresses
- Add linux/ppc64le support
- Allow OUTPUT flag to override to use with build target e.g: `PLATFORMS=linux/ppc64le OUTPUT=--load make build`

### DIFF
--- a/images/Makefile.common.in
+++ b/images/Makefile.common.in
@@ -13,8 +13,8 @@ IMAGE?=$(REGISTRY)/$(IMAGE_NAME):$(TAG)
 export DOCKER_CLI_EXPERIMENTAL=enabled
 
 # build with buildx
-PLATFORMS?=linux/amd64,linux/arm64,linux/s390x
-OUTPUT=
+PLATFORMS?=linux/amd64,linux/arm64,linux/s390x,linux/ppc64le
+OUTPUT?=
 PROGRESS=auto
 build: ensure-buildx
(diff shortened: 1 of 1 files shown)


BenTheElder:
> Allow OUTPUT flag to override to use with build target e.g: PLATFORMS=linux/ppc64le OUTPUT=--load make build

This should be make quick with PLATFORMS set, --load target is already in the makefile.

mayurwaghmode:
@BenTheElder are we going to add the ppc64le base image digest [here ](https://hub.docker.com/r/kindest/base/tags)?

mkumatag:
> > Allow OUTPUT flag to override to use with build target e.g: PLATFORMS=linux/ppc64le OUTPUT=--load make build
> 
> This should be make quick with PLATFORMS set, --load target is already in the makefile.

we are overriding the PLATFORMS in the quick target, its not working

```
% PLATFORMS=linux/ppc64le make quick
./../../hack/build/init-buildx.sh
kind-builder
docker buildx build --platform=linux/amd64 --load --progress=auto -t kindest/base:v20220411-8cc0c0b5-dirty --pull .
```

mkumatag:
> `make` syntax for setting a variable is a little different than setting an environment variable. Can you try `make quick PLATFORMS=linux/ppc64le`

oops, true that..

PR #2712, opened 2022-04-07
Added multi-arch support for quay quick

mayurwaghmode:
quick allows building image into the local docker


mkumatag:
The whole philosophy behind using quick is to not build all the possible platforms, https://github.com/kubernetes-sigs/kind/pull/2711 will fix the actual issue

mayurwaghmode:
> The whole philosophy behind using quick is to not build all the possible platforms, #2711 will fix the actual issue

Got it! Thank you so much :)

aojea:
> The whole philosophy behind using quick is to not build all the possible platforms, #2711 will fix the actual issue

Exactly, I ll close this, thanks

PR #2716, opened 2022-04-13
consider prereleases on comparisons

aojea:
subtle bug, but it seems that the comparison functions are correct

http://masterminds.github.io/sprig/semver.html#:~:text=A%20pre%2Drelease%20version%20indicates,comparator%20will%20skip%20prerelease%20versions.

it should append the `-0` to the version to consider prereleases on comparison

### DIFF
--- a/pkg/internal/version/version_test.go
+++ b/pkg/internal/version/version_test.go
@@ -346,3 +346,72 @@ func TestComponents(t *testing.T) {
 		}
 	}
 }
+
+func TestVersion_LessThan_AtLeast(t *testing.T) {
+	tests := []struct {
+		name     string
+		version  string
+		lessThan string
+		want     bool
+	}{
+		{
+			name:     "same version",
+			version:  "1.21.1",
+			lessThan: "1.21.1",
+			want:     false,
+		},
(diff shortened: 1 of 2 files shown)


aojea:
as always, thanks to @howardjohn  for reporting

pkg/internal/version/version_test.go
```diff
@@ -346,3 +346,72 @@ func TestComponents(t *testing.T) {
...
+		{
+			name:     "prerelease against same version",
+			version:  "v1.24.0-beta.0.115+65178fec72df62",
+			lessThan: "v1.24.0",
+			want:     true,
+		},
+		{
+			name:     "prerelease against same version considering prereleases",
+			version:  "v1.24.0-beta.0.115+65178fec72df62",
+			lessThan: "v1.24.0-0",
```
neolit123, line 402: Does the same happen in the api machinery version parser package?
IIRC prelease is compared alpha numerically, thus -0 is less than -beta.

neolit123: Nvm, this is correct.

/lgtm

PR #2721, opened 2022-04-19
#2717 Exit on startup if the host is out of space

### LINKED ISSUE #2717: Kind should note when Docker is out of disk space
Kind should note when Docker is out of disk space on startup and refuse to start.

**Why is this needed**:
Spent a lot of time trying to figure out why a deployment was failing to be able to create a directory. Finally found the problem when I went to install Minikube and it refused to start due to Docker being out of disk space. `docker prune` fixed the problem and allowed the deployment to function properly.

IoannisMatzaris:
There is a new function `CheckFreeDiskSpace` under `pkg/cluster/internal/providers/common/`. In this way, it can be checked on the `kind create custer`, if there is enough disk space available.

Fixes #2717

### DIFF
--- a/pkg/cluster/internal/providers/podman/provider.go
+++ b/pkg/cluster/internal/providers/podman/provider.go
@@ -17,7 +17,6 @@ limitations under the License.
 package podman
 
 import (
-	"bytes"
 	"encoding/json"
 	"fmt"
 	"net"
@@ -26,7 +25,6 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/shirou/gopsutil/disk"
 	"sigs.k8s.io/kind/pkg/cluster/nodes"
 	"sigs.k8s.io/kind/pkg/cluster/nodeutils"
 	"sigs.k8s.io/kind/pkg/errors"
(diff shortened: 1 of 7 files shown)


BenTheElder:
Thank you for the PR, however I think we should discuss the approach on #2717 first.

See: our contributor guide https://kind.sigs.k8s.io/docs/contributing/getting-started/

I'll comment over there, but also note that we maintain a very small dependency set so as to ensure kind is easily embedded in e.g. test frameworks.

aojea:
I don't think kind should do free space checking, people using normal docker containers can run out of disk too and kind is not special in this regard, it creates containers ...

pkg/cluster/internal/providers/podman/provider.go
```diff
@@ -396,6 +399,12 @@ func info(logger log.Logger) (*providers.ProviderInfo, error) {
 	if err := json.Unmarshal(out, &pInfo); err != nil {
 		return nil, err
 	}
+
+	// Check if there is enough disk space available
+	if err := common.CheckFreeDiskSpace(pInfo.Store.GraphRoot); err != nil {
+		return nil, err
```
flouthoc, line 405: I think it would be better if kind does not takes on the system management task like checking free disk space but if it still happens I think instead of `err` this should just print a `warn` here and actual error message from provider can be relayed back to user that would be really helpful.

BenTheElder:
Thank you, but we won’t be taking this approach at this time.

PR #2728, opened 2022-04-26
load: support loading images with different providers (`podman`)

flouthoc:
As of now `kind` hardcodes looking and loading images from `docker`
as a result `kind load docker-image <image>:<tag>` does not works well when
`Provider` is configured as `podman` instead of `docker`.

Following PR makes sure that `kind` supports loading images irrespective of
the providers configured and works for both `podman` and `docker`

Closes: https://github.com/kubernetes-sigs/kind/issues/2729

### DIFF
--- a/pkg/cluster/internal/providers/docker/provider.go
+++ b/pkg/cluster/internal/providers/docker/provider.go
@@ -133,6 +133,28 @@ func (p *provider) ListNodes(cluster string) ([]nodes.Node, error) {
 	return ret, nil
 }
 
+// SaveImage saves images to dest, as in `docker save`
+func (p *provider) SaveImage(images []string, dest string) error {
+	commandArgs := append([]string{"save", "-o", dest}, images...)
+	return exec.Command("docker", commandArgs...).Run()
+}
+
+// ImageID return the Id of the container image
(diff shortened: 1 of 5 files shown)


flouthoc:
Created a bug so it can be tracked even if PR is closed: https://github.com/kubernetes-sigs/kind/issues/2729

BenTheElder:
https://github.com/kubernetes-sigs/kind/issues/2729#issuecomment-1110152680

BenTheElder:
1. this is a breaking change, some projects like kubernetes require docker to build images, but the user may opt to create the cluster with podman (it's possible to install both), after this change loading the images would no longer work. there's an outline for a new unified image load command in #2038 
2. image saving isn't necessarily a property of node providers, we should not move it there as discussed in #2038, you could imagine a provider using say ingnite, kubectl pods or other where we can still run containers but not obtain images, and I should really be able to load across docker image x podman nodes and podman build x docker nodes (it's admittedly likely niche, but we might as well make it simple to handle because someone is going to want this and it shouldn't be difficult to implement or maintain).

flouthoc (commented):
Closing this until there is a final design discussed in upstream and this is not a blocker for my use-case so this can wait.

Issue #1726, opened 2020-07-13
containerd: use systemd cgroup driver by default?


AkihiroSuda:
**What would you like to be added**:
Use systemd cgroup driver by default

**Why is this needed**:
Because it is recommended on systemd hosts: https://kubernetes.io/docs/setup/production-environment/container-runtimes/

aojea:
this seems a simple change, @AkihiroSuda do you know there can be issues due to be running in nested containers?
I mean, let's says docker runs in the host with cgroupfs, and kubelet inside the kind node (docker container) with systemd driver (this systemd process is runnig inside a cgroup already)

tao12345666333:
Maybe there is no problem?

Currently, my docker is using systemd as the cgroup driver, and cgroupfs is used in KIND.

I will test locally and give the results as soon as possible

BenTheElder:
That page talks about having two controllers, but if it intends to recommend it, the language there could do with some updating, I wouldn't really call that "recommends". Nowhere does it say "you should do this".

I'd want to know that we're not going to cause a regression here.

We also want to be doing something like the changes in https://d2iq.com/blog/running-kind-inside-a-kubernetes-cluster-for-continuous-integration with custom kubepods path etc. sooner rather than later, cc @jieyu xref: #1614.

AkihiroSuda:
For cgroup v2, runc explicitly recommends systemd driver, though it is still opt-in.
https://github.com/opencontainers/runc/blob/master/docs/cgroup-v2.md#systemd

Podman and Docker already switched the default on v2 to systemd.

BenTheElder:
cc @neolit123
/assign

BenTheElder:
We should strongly consider this in v0.11.0, see also https://github.com/kubernetes/kubernetes/pull/96594 for some excitement around kind's (poor) behavior with cgroups v1 currently ...

BenTheElder:
Noting that there's currently some issues upstream with runc upgrade and systemd
https://github.com/kubernetes/kubernetes/issues/88197 https://github.com/kubernetes/kubernetes/pull/102094

BenTheElder:
There is some cri-o CI in Kubernetes related to this now but still this driver continues to have problems https://github.com/kubernetes/kubernetes/pull/102508#issuecomment-873546456

BenTheElder:
This driver is getting more testing upstream but still seems to have major issues https://github.com/kubernetes/kubernetes/issues/104280 😬

[...]

PR #2737, opened 2022-05-03
Use systemd cgroup driver for v1.24.0+

BenTheElder:
see: #1726


images/base/files/etc/containerd/config.toml
```diff
@@ -19,6 +19,9 @@ version = 2
   runtime_type = "io.containerd.runc.v2"
   # Generated by "ctr oci spec" and modified at base container to mount poduct_uuid
   base_runtime_spec = "/etc/containerd/cri-base.json"
+  [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
+    # use systemd cgroup by default
+    SystemdCgroup = true
```
BenTheElder, line 24: may also need to do this for test-handler.

pkg/build/nodeimage/containerd.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"sigs.k8s.io/kind/pkg/errors"
+	"sigs.k8s.io/kind/pkg/exec"
+
+	"sigs.k8s.io/kind/pkg/internal/patch"
+)
+
+const containerdConfigPath = "/etc/containerd/config.toml"
+
+const containerdConfigPatchSystemdCgroupFalse = `
```
BenTheElder, line 30: should also patch test-handler here

BenTheElder: done.

images/base/files/etc/containerd/config.toml
```diff
@@ -19,11 +19,18 @@ version = 2
...
 
 # Setup a runtime with the magic name ("test-handler") used for Kubernetes
 # runtime class tests ...
 [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.test-handler]
+  # same settings as runc
```
aojea, line 29: What is this test-handler used for?

BenTheElder: see comment two lines above in file? :-)

BenTheElder: https://github.com/kubernetes/kubernetes/blob/e045e5c875b2f51c4c4b1bc14d4e717123649b17/test/e2e/framework/node/runtimeclass.go

it's a tricky problem in k/k e2e testing. to test that runtimeclass works, but need to know that a non-default runtime is configured.

BenTheElder:
hmm 1.24 kubelet unhealthy 
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/2737/pull-kind-e2e-kubernetes/1521573229309726720

https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/2737/pull-kind-e2e-kubernetes/1521573229309726720/artifacts/logs/kind-control-plane/kubelet.log

> May 03 19:52:45 kind-control-plane kubelet[1749]: Error: failed to run Kubelet: invalid configuration: cgroup ["kubelet"] has some missing paths: /sys/fs/cgroup/pids/kubelet.slice, /sys/fs/cgroup/cpu,cpuacct/kubelet.slice, /sys/fs/cgroup/cpu,cpuacct/kubelet.slice, /sys/fs/cgroup/cpuset/kubelet.slice, /sys/fs/cgroup/memory/kubelet.slice, /sys/fs/cgroup/systemd/kubelet.slice, /sys/fs/cgroup/hugetlb/kubelet.slice

BenTheElder:
right ... all the entrypoint tricks are aimed at cgroupfs.

images/base/files/etc/containerd/config.toml
aojea: I was wondering why do we need to fill this too, but seems that is for coherence,

aojea: found it in the 5th commit

BenTheElder:
https://github.com/opencontainers/runc/blob/main/docs/systemd.md

BenTheElder:
Got back to this a bit today:
- kubelet appends .slice internally and you should not add it yourself
- ... but the cgroup `--cgroup-root` + .slice must exist
- if we use the current `/kubelet` cgroup root approach but populate `/kubelet.slice` instead, everything seems fine

Debating if we should plumb through a switch or just always setup both, leaning towards the latter so it's simpler to ensure either can be used.

BenTheElder:
s390x build flake https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/2737/pull-kind-build/1522383838079094784

images/base/files/etc/containerd/config.toml
BenTheElder: in e2e we will actually create pod(s) with this runtime, we need that to match kubelet wrt cgroup driver

BenTheElder:
s390x flake again
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/2737/pull-kind-build/1522405658077433856
/retest

BenTheElder:
logs still show failing to remove rmda cgroup but that's not new to this PR / driver https://github.com/kubernetes/kubernetes/issues/109182

BenTheElder:
We need to test this in more environments before moving forward, just to be on the safe side that we haven't missed a quirk.

I'm pushing a node image from this, will update the PR with it so that github actions environments pick it up, and then see about getting it tested in some additional environments.

BenTheElder:
podman the node only got this far:
```
INFO: ensuring we can execute mount/umount even with userns-remap
INFO: remounting /sys read-only
INFO: making mounts shared
INFO: detected cgroup v1
WARN: cgroupns not enabled! Please use cgroup v2, or cgroup v1 with cgroupns enabled.
INFO: fix cgroup mounts for all subsystems
```

exit status 1 https://github.com/kubernetes-sigs/kind/actions/runs/2296780194

BenTheElder:
Across the board broken in podman, not in docker. https://github.com/kubernetes-sigs/kind/pull/2737/checks

/hold

BenTheElder:
Also cgroup v2 across the board in CI https://github.com/kubernetes-sigs/kind/runs/6359585961?check_suite_focus=true
... but not locally on my machine 🤔

BenTheElder:
ah, cgroupv2 failed on setting up the vagrant env.

BenTheElder:
Figured as much, containerd has the same issue. https://github.com/containerd/containerd/pull/6915

https://github.blog/2022-04-12-git-security-vulnerability-announced/

BenTheElder:
OK, so it's now just (podman, cgroup v1) that is broken, all of cgroup v2 is passing after #2742

BenTheElder:
trace from https://github.com/kubernetes-sigs/kind/pull/2737/commits/9e14207e657296ca10946745ca024ffdd212b72c
```
+ source /usr/local/bin/entrypoint
++ set -o errexit
++ set -o nounset
++ set -o pipefail
++ userns=
++ grep -Eqv '0[[:space:]]+0[[:space:]]+4294967295' /proc/self/uid_map
++ validate_userns
++ [[ -z '' ]]
++ return
++ configure_containerd
++ local snapshotter=
++ [[ -n '' ]]
++ [[ -z '' ]]
+++ stat -f -c %T /kind
++ [[ overlayfs == \z\f\s ]]
++ [[ -n '' ]]
++ configure_proxy
++ mkdir -p /etc/systemd/system.conf.d/
++ cat
++ fix_kmsg
++ [[ ! -e /dev/kmsg ]]
++ fix_mount
++ echo 'INFO: ensuring we can execute mount/umount even with userns-remap'
INFO: ensuring we can execute mount/umount even with userns-remap
+++ which mount
+++ which umount
++ chown root:root /usr/bin/mount /usr/bin/umount
+++ which mount
+++ which umount
++ chmod -s /usr/bin/mount /usr/bin/umount
++++ which mount
+++ stat -f -c %T /usr/bin/mount
++ [[ overlayfs == \a\u\f\s ]]
++ echo 'INFO: remounting /sys read-only'
INFO: remounting /sys read-only
++ mount -o remount,ro /sys
++ echo 'INFO: making mounts shared'
INFO: making mounts shared
++ mount --make-rshared /
++ fix_cgroup
++ [[ -f /sys/fs/cgroup/cgroup.controllers ]]
++ echo 'INFO: detected cgroup v1'
INFO: detected cgroup v1
++ local current_cgroup
+++ grep -E '^[^:]*:([^:]*,)?cpu(,[^,:]*)?:.*' /proc/self/cgroup
+++ cut -d: -f3
++ current_cgroup=/machine.slice/libpod-956ff18413c2d67f2a684aab03727625784586aa3398f2b03da84763df5155e8.scope
++ '[' /machine.slice/libpod-956ff18413c2d67f2a684aab03727625784586aa3398f2b03da84763df5155e8.scope = / ']'
++ echo 'WARN: cgroupns not enabled! Please use cgroup v2, or cgroup v1 with cgroupns enabled.'
WARN: cgroupns not enabled! Please use cgroup v2, or cgroup v1 with cgroupns enabled.
++ echo 'INFO: fix cgroup mounts for all subsystems'
INFO: fix cgroup mounts for all subsystems
++ local cgroup_subsystems
+++ findmnt -lun -o source,target -t cgroup
+++ awk '{print $2}'
+++ grep /machine.slice/libpod-956ff18413c2d67f2a684aab03727625784586aa3398f2b03da84763df5155e8.scope
++ cgroup_subsystems='/sys/fs/cgroup/blkio
/sys/fs/cgroup/memory
/sys/fs/cgroup/cpu,cpuacct
/sys/fs/cgroup/rdma
/sys/fs/cgroup/freezer
/sys/fs/cgroup/cpuset
/sys/fs/cgroup/perf_event
/sys/fs/cgroup/pids
/sys/fs/cgroup/devices
/sys/fs/cgroup/misc
/sys/fs/cgroup/hugetlb
/sys/fs/cgroup/net_cls,net_prio
/sys/fs/cgroup/systemd'
++ local unsupported_cgroups
+++ grep -v /machine.slice/libpod-956ff18413c2d67f2a684aab03727625784586aa3398f2b03da84763df5155e8.scope
+++ findmnt -lun -o source,target -t cgroup
+++ awk '{print $2}'
++ unsupported_cgroups=
```

BenTheElder:
So this is just podman runtime being updated and  `unsupported_cgroups=$(findmnt -lun -o source,target -t cgroup | grep -v "${current_cgroup}" | awk '{print $2}')` failing due to their being no match.

That's not related to this PR. It's a minor bug in https://github.com/kubernetes-sigs/kind/pull/2709 which hasn't been picked up yet due to not previously bumping the node image + podman CI uses default node image.

I'll send a fix for that, then we can rebase this PR and try again.

BenTheElder:
rebasing on fix in https://github.com/kubernetes-sigs/kind/pull/2743

BenTheElder:
And now podman is passing https://github.com/kubernetes-sigs/kind/runs/6362603916?check_suite_focus=true

BenTheElder:
> May 10 02:34:53.338: INFO: externalsvc-72bcl[services-5001].container[externalsvc]=failed to create containerd task: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: unable to apply cgroup configuration: failed to write 84042: write /sys/fs/cgroup/rdma/kubelet.slice/kubelet-kubepods.slice/kubelet-kubepods-besteffort.slice/kubelet-kubepods-besteffort-podbd0b7b96_aa32_4e00_8323_6f20dd1b85a4.slice/cri-containerd-6e1a4e9c43aba1945e9a28e410939e0801c180fa69bb1850018c9c0b2956a91a.scope/cgroup.procs: no such device: unknown

we still have the rdma issue https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/2737/pull-kind-conformance-parallel-ipv6/1523846989832261632

but that's not new
/retest

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -621,6 +627,13 @@ func Config(data ConfigData) (config string, err error) {
 		data.KubeadmFeatureGates["IPv6DualStack"] = true
 	}
 
+	// before 1.24 kind uses cgroupfs
+	// after 1.24 kind uses systemd starting in kind v0.13.0
+	// before kind v0.13.0 kubernetes 1.24 wasn't released yet
+	if ver.LessThan(version.MustParseSemantic("v1.24.0")) {
+		data.CgroupDriver = "cgroupfs"
```
aojea, line 634: this effectively means that  `ConfigTemplateBetaV1` and `ConfigTemplateBetaV2` are going to use data.CgroupDriver = "cgroupfs", maybe for consistency is better to replace in all templates?

images/base/files/usr/local/bin/entrypoint
```diff
@@ -294,10 +294,12 @@ fix_cgroup() {
   # "nesting" clusters, unless we instruct it to use a different cgroup root.
   # We do this, and when doing so we must fixup this alternative root
   # currently this is hardcoded to be /kubelet
+  # under systemd cgroup driver, kubelet appends .slice
```
aojea, line 297: is there any problem for mounting it twice?
is not possible to detect which one we want?

BenTheElder: No there's no problem. The cgroup controllers are enabled on many trees. Only one of these two will be used at runtime.

Doing it this way is robust in the face of future changes and user config patches.
We don't know what the node config will be until after we spawn the node and discover details like Kubernetes version.

pkg/cluster/internal/kubeadm/config.go
BenTheElder: Do you mean don't template it?

I don't think we should do that no, selecting based on Kubernetes version is the current approach and this is more explicit about how that is selected. Templating it everywhere ensures greater logical consistency.

pkg/build/nodeimage/containerd.go
```diff
@@ -0,0 +1,50 @@
...
+[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
+  SystemdCgroup = false
+
+[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.test-handler.options]
+  SystemdCgroup = false
+`
+
+func configureContainerdSystemdCgroupFalse(containerCmdr exec.Cmder, config string) error {
```
aojea, line 38: can this fail against an image that was built for 1.24.x with an old version of kind, without the containerd change?

BenTheElder: yes, users that upgrade to kind v0.13 will need to use an image built with a newer version
we have no official images that way.

for custom built images that are 1.24+ with kind < v0.13, users will need to build new images. we'll have a release note.

the opposite is also true, 1.24+ official images / images built going forward will require v0.13.

this already would have happened in 1.21 with kubeadm defaulting to systemd then, we just shifted it a few releases.

BenTheElder:
Working on v0.13.0 based on this, pushing images. Wrote release notes.

BenTheElder:
https://github.com/kubernetes-sigs/kind/releases/tag/v0.13.0

PR #2741, opened 2022-05-09
docs:update using-wsl2.md(more shorter,and up-to-date)

arukiidou:
WSL2 installation instructions are now up-to-date and more shorter

### DIFF
--- a/site/content/docs/user/using-wsl2.md
+++ b/site/content/docs/user/using-wsl2.md
@@ -47,16 +47,11 @@ Once your Windows machine is ready, you need to do a few more steps to set up WS
 1. Open a PowerShell window as an admin, then run
 
     {{< codeFromInline lang="powershell" >}}
-Enable-WindowsOptionalFeature -Online -FeatureName VirtualMachinePlatform, Microsoft-Windows-Subsystem-Linux
+wsl --install -d Ubuntu-20.04
 {{< /codeFromInline >}}
 
-1. Reboot when prompted.
-1. After the reboot, set WSL to default to WSL2. Open an admin PowerShell window and run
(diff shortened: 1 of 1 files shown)


site/content/docs/user/using-wsl2.md
```diff
@@ -47,16 +47,11 @@ Once your Windows machine is ready, you need to do a few more steps to set up WS
 1. Open a PowerShell window as an admin, then run
 
     {{< codeFromInline lang="powershell" >}}
-Enable-WindowsOptionalFeature -Online -FeatureName VirtualMachinePlatform, Microsoft-Windows-Subsystem-Linux
+wsl --install -d Ubuntu-20.04
```
stmcginnis, line 50: This is the command to run when WSL is already enabled to install a specific distro. The previous command was how to get WSL set up in the first place.

Not sure if this has changed with Windows 11, but at least for older versions I think we still need the WSL installation instructions.

BenTheElder: We also have instructions below about picking your favorite distro, but this command now does a specific distro.

Appreciate the PR, but I don't think we should make this change.

If WSL2 is pre-installed in some releases, we could add a note about checking if it's already installed.

arukiidou: Apologies for the lack of PR at the finish line.
Will the following be enough for the finish line?

- Add `wsl --install` Requirements (requires windows 2004 or later.)
- In case `wsl --install` are not available, keep the original instructions  they 
- Modify the procedure to select your favorite distribution

arukiidou: ok,thanks

PR #2743, opened 2022-05-09
allow no matches when grepping for unsupported_cgroups

BenTheElder:
see: https://github.com/kubernetes-sigs/kind/pull/2737#issuecomment-1121690676

https://linuxcommand.org/lc3_man_pages/grep1.html#:~:text=is%20not%20set.-,EXIT%20STATUS,-Normally%20the%20exit

### DIFF
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -27,6 +27,11 @@ if grep -Eqv "0[[:space:]]+0[[:space:]]+4294967295" /proc/self/uid_map; then
   echo 'INFO: running in a user namespace (experimental)'
 fi
 
+grep_allow_nomatch() {
+  # grep exits 0 on match, 1 on no match, 2 on error
+  grep "$@" || [[ $? == 1 ]]
+}
+
 validate_userns() {
   if [[ -z "${userns}" ]]; then
     return
@@ -240,7 +245,7 @@ fix_cgroup() {
   #
   # See https://github.com/kubernetes/kubernetes/issues/109182
   local unsupported_cgroups
-  unsupported_cgroups=$(findmnt -lun -o source,target -t cgroup | grep -v "${current_cgroup}" | awk '{print $2}')
+  unsupported_cgroups=$(findmnt -lun -o source,target -t cgroup | grep_allow_nomatch -v "${current_cgroup}" | awk '{print $2}')
   if [ -n "$unsupported_cgroups" ]; then
     local mnt
     echo "$unsupported_cgroups" |
--- a/pkg/apis/config/defaults/image.go
+++ b/pkg/apis/config/defaults/image.go
@@ -18,4 +18,4 @@ limitations under the License.
 package defaults
 
 // Image is the default for the Config.Image field, aka the default node image.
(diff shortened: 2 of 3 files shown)


BenTheElder:
pushing images

images/base/files/usr/local/bin/entrypoint
```diff
@@ -27,6 +27,11 @@ if grep -Eqv "0[[:space:]]+0[[:space:]]+4294967295" /proc/self/uid_map; then
   echo 'INFO: running in a user namespace (experimental)'
 fi
 
+grep_allow_nomatch() {
+  # grep exits 0 on match, 1 on no match, 2 on error
```
aojea, line 31: > > EXIT STATUS
>        Normally the exit status is 0 if a line is selected, 1 if no lines were
>        selected, and 2 if an error occurred.  However, if the -q or --quiet or
>        --silent  is  used and a line is selected, the exit status is 0 even if
>        an error occurred.
> 

is it better to use --silent?

BenTheElder: `-q`, `--quiet`, and `--silent` are the same flag, and no that's not a suitable option:

>   -q, --quiet, --silent
>               Quiet; do not write  anything  to  standard  output.   Exit
>               immediately with zero status if any match is found, even if
>               an error was detected.  Also see the  -s  or  --no-messages
>               option.

BenTheElder: also we're not trying to swallow *errors*, we're remapping from:

- match: 0
- no-match: 1
- error: 2

to:

- match: 0
- no-match: 0
- error: non-zero

aojea: ahh

images/base/files/usr/local/bin/entrypoint
```diff
@@ -27,6 +27,11 @@ if grep -Eqv "0[[:space:]]+0[[:space:]]+4294967295" /proc/self/uid_map; then
...
+  grep "$@" || [[ $? == 1 ]]
```
BenTheElder, line 32: TODO: We could consider using a less terse version and retain the exact exit code.
right now that's not important and I'd rather not have to rebase and re-build https://github.com/kubernetes-sigs/kind/pull/2737 again unless there's a seriours error here.

We probably should follow-up with preserving the exact status for future usage.

Or just switch to pure awk.

kolyshkin: Yeah, it seems that awk would look nicer. Something like

```bash
unsupported_cgroups=$(findmnt -lun -o source,target -t cgroup | awk '!/'"${current_cgroup}"'/ {print $2}')
```

BenTheElder: It feels slightly wrong that we're regex matching in either case instead of substring literal but :this-is-fine:
I suppose we could regex escape it with some more bash 🙈 

It looks like if we stuck with grep we could use `-F` / `--fix-strings`.

kolyshkin: For `awk`, one can use something `index(string, substring) == -1`. Some implementations treat substr as a regexp, but GNU awk does a fixed substring match.

Issue #2246, opened 2021-05-14
Improve `kind get kubeconfig` and `kind export kubeconfig` for multicluster


howardjohn:
**What would you like to be added**:

* A new flag, `kind export kubeconfig --internal`, matching `kind get kubeconfig --internal`. I don't personally care about this one, but it seems like we should be consistent.
* A new flag, like `kind export kubeconfig --docker` (with a better name). This would export the result of `docker inspect "${CLUSTER_NAME}`-control-plane" --format "{{ .NetworkSettings.Networks.kind.IPAddress }}" as the address

**Why is this needed**:

* Default of 127.0.0.1 allows only reaching from host
* --internal only allows reaching from internal
* This mode allows reaching from the host *and* internally

BenTheElder:
The first point has an open discussion / tracking issue and pull request that need following up on. @aojea has been active reviewing that.

The second point I brought up in that discussion.

BenTheElder:
> This mode allows reaching from the host and internally

* On Linux, or else it would be the default. Also it will not be the case if rebooting / restarting docker and getting a new IP, which otherwise works for single node clusters. I think John knows this, but for anyone else interested this is why.

Existing discussion: https://github.com/kubernetes-sigs/kind/issues/2217

aojea:
what if we use a env variable to modify the `internal` flag or `export kubeconfig` behaviour?
 at least with some EXPERIMENTAL_KUBECONFIG_INTERNAL_IP=true we can unblock this

gprossliner:
Having Environment-Variable based Flags without an corresponding CLI flag is no good UX, in my opinion. So I would consider implementing the `--address-type` flag, as discussed in #2217. 

From https://github.com/kubernetes-sigs/kind/issues/2217#issuecomment-832530967:

> My suggestion for --address-type would be:
> 
> | -address-type=  | URL                             | Host-Resolvable | Container-Resolvable   |
> |----------------|---------------------------------|-----------------|------------------------|
> | container-name | https://kind-control-plane:6443 | No              | Yes                    |
> | container-ip   | https://172.18.0.2:6443         | Linux only      | Yes, on 'kind' network |
> | host-port      | https://127.0.0.1:36995         | Yes             | No                     |
> 
> The current default would be "host-port". I would use "container-ip" as the default on Linux, because the URL is resolvable on the host, and internally.
> 
> A clean migration path could be like:
> 
> * Phase 1: Quit with error if one uses --internal and --address-type
[...]

PR #2745, opened 2022-05-10
Add a flag to override server address in the generated config

Shadi:
We are using Kind in one of our project integration tests suites, these tests run inside a docker container using github actions, when kind creates a cluster as part of the test run preparation it uses the default "127.0.0.1" for server address: https://github.com/kubernetes-sigs/kind/blob/main/pkg/internal/apis/config/default.go#L57

when the test runs a `kubectl` command it can't connect to the created cluster since `kubectl` is being ran from inside a container and `127.0.0.1` resolves to that container instead of the cluster on the host machine.

This change would allow us to set a different address when preparing for tests so it can execute `kubectl` commands and connect to the created cluster

### DIFF
--- a/pkg/cluster/createoption.go
+++ b/pkg/cluster/createoption.go
@@ -124,3 +124,11 @@ func CreateWithDisplaySalutation(displaySalutation bool) CreateOption {
 		return nil
 	})
 }
+
+// CreateWithApiServerAddress Sets server in kubeconfig to provided value
+func CreateWithApiServerAddress(apiServerAddress string) CreateOption {
+	return createOptionAdapter(func(o *internalcreate.ClusterOptions) error {
+		o.ApiServerAddressOverride = apiServerAddress
+		return nil
+	})
+}
--- a/pkg/cluster/internal/create/create.go
+++ b/pkg/cluster/internal/create/create.go
@@ -60,8 +60,9 @@ type ClusterOptions struct {
 	// see https://github.com/kubernetes-sigs/kind/issues/324
 	StopBeforeSettingUpKubernetes bool // if false kind should setup kubernetes after creating nodes
 	// Options to control output
-	DisplayUsage      bool
-	DisplaySalutation bool
+	DisplayUsage             bool
+	DisplaySalutation        bool
+	ApiServerAddressOverride string
 }
 
 // Cluster creates a cluster
@@ -231,6 +232,10 @@ func fixupOptions(opts *ClusterOptions) error {
 		for i := range opts.Config.Nodes {
 			opts.Config.Nodes[i].Image = opts.NodeImage
 		}
+	}
+
+	if opts.ApiServerAddressOverride != "" {
+		opts.Config.Networking.APIServerAddress = opts.ApiServerAddressOverride
 	}
 
 	// default config fields (important for usage as a library, where the config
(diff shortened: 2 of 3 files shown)


BenTheElder:
Thank you for your PR.
In the future, per the contributor guide, we'd like to discuss features, how they will work, etc. *before* reviewing PRs.

[CONTRIBUTING.md](https://github.com/kubernetes-sigs/kind/blob/main/CONTRIBUTING.md) => https://kind.sigs.k8s.io/docs/contributing/getting-started/

We do not currently require KEPs like the rest of kubernetes, but we do prefer to flesh out the feature in an issue.

For example on such an issue I would ask:
- Does this need to be functionality in kind? It should be relatively straightforward to alter the address in kubeconfig yourself if we're just substituting it in kubeconfig vs say https://kind.sigs.k8s.io/docs/user/configuration/#api-server
- Which commands? This PR only seems to implement the flag for `kind create cluster` but there are other kubeconfig related commands
- What flag name sounds reasonable? `apiServerAddress` is not the typical naming convention (camelCase vs hypen-case) for flags and is also a misleading name IMO.
- Have we considered the related existing feature request issues and their discussions https://github.com/kubernetes-sigs/kind/issues/1558 https://github.com/kubernetes-sigs/kind/issues/2246

Shadi:
Hello @BenTheElder, thanks for taking a look

> Thank you for your PR. In the future, per the contributor guide, we'd like to discuss features, how they will work, etc. _before_ reviewing PRs.
> 
> [CONTRIBUTING.md](https://github.com/kubernetes-sigs/kind/blob/main/CONTRIBUTING.md) => https://kind.sigs.k8s.io/docs/contributing/getting-started/
> 
> We do not currently require KEPs like the rest of kubernetes, but we do prefer to flesh out the feature in an issue.

> For example on such an issue I would ask:
> 
> * Does this need to be functionality in kind? It should be relatively straightforward to alter the address in kubeconfig yourself if we're just substituting it in kubeconfig vs say https://kind.sigs.k8s.io/docs/user/configuration/#api-server
> * Have we considered the related existing feature request issues and their discussions [Need a way to get kubeconfig with node IP #1558](https://github.com/kubernetes-sigs/kind/issues/1558) [Improve `kind get kubeconfig` and `kind export kubeconfig` for multicluster #2246](https://github.com/kubernetes-sigs/kind/issues/2246)

this indeed similar to  [Need a way to get kubeconfig with node IP #1558](https://github.com/kubernetes-sigs/kind/issues/1558), the same `sed` workaround can be used in our case too, but it seemed it can be changed quickly by using `--name` flag as a guide so I created this PR without looking at the older issues, feel free to close this PR.

> * Which commands? This PR only seems to implement the flag for `kind create cluster` but there are other kubeconfig related commands
I'm not very familiar with this project and my goal was to only override the server address for the generated file at `KubeconfigPath`, so I didn't look at the other cases, but I'm happy to make changes for them too if you think this PR can be a starting point to implement this feature properly.

> * What flag name sounds reasonable and clearest? `apiServerAddress` is not the typical naming convention (camelCase vs hypen-case) for flags and is also a misleading name IMO.
I used `name` flag as a guide and changed it now to just `address` and updated the description.


I decided to create this PR since it seemed to be not a lot of work to implement this change in a way the `name` flag is being used now when creating clusters, if you prefer to have a proper implementation based on the discussion in  [Need a way to get kubeconfig with node IP #1558](https://github.com/kubernetes-sigs/kind/issues/1558) then feel free to close it.

aojea:
> `kubectl` is being ran from inside a container and `127.0.0.1` resolves to that container instead of the cluster on the host machine.

you can use `kind get kubeconfig --internal` then, it will return a kubeconfig usable inside containers

BenTheElder:
> you can use kind get kubeconfig --internal then, it will return a kubeconfig usable inside containers

Or better yet: I think `kind create cluster && kind export kubeconfig --internal` is what you want (`get` returns it to stdout, `export` does the kubeconfig merging like `kind create cluster`)

We should really follow up on those issues and rethink these flags for some additional options though.

Issue #1331, opened 2020-02-13
containerd versus dockerd WORKDIR non-root permissions


michaelbannister:
**What happened**:
An image whose WORKDIR is set to a directory with permissions only for one user, run in a Pod with securityContext.runAsUser set to a _different_ UID. Kind runs the pod just fine, but Kubernetes fails with an error like `failed to create containerd task: OCI runtime create failed: container_linux.go:346: starting container process caused \"chdir to cwd (\\\"/home/nonroot\\\") set in config.json failed: permission denied\": unknown`

Kind only fails in the same way as "normal" Kubernetes if the securityContext is also configured to drop all capabilities.

**What you expected to happen**:
Kind should fail to run the container in the same way as Kubernetes.

**How to reproduce it (as minimally and precisely as possible)**:
See https://github.com/michaelbannister/distroless-permissions-test for a worked example.

**Anything else we need to know?**:
This came up while working on this Istio PR: https://github.com/istio/istio/pull/20854

**Environment:**

- kind version: kind v0.7.0 go1.13 darwin/amd64
- Kubernetes version: v1.15.5 (Docker for Desktop on macOS)
- Docker version: 19.03.5
- OS: macOS 10.15.3

However this also occurs on the testing infrastructure for the Istio project. I don't know the details for that other than that it uses Kind.

BenTheElder:
Can you define what "kubernetes" is when not KIND? KIND is also kubernetes :+)
It might be containerd vs docker as the node runtime, or something with KIND ...

will investigate O(soon)

michaelbannister:
In my case I tested this against the Kubernetes installed by Docker Desktop on macOS. I might be able to try it on GKE, will get back to you…

michaelbannister:
GKE v1.14.8-gke.33 with Docker as the container runtime, when running the job defined in [job.yaml](https://github.com/michaelbannister/distroless-permissions-test/blob/master/job.yaml):

`Error: failed to start container "distroless-permissions-test": Error response from daemon: OCI runtime create failed: container_linux.go:345: starting container process caused "chdir to cwd (\"/home/nonroot\") set in config.json failed: permission denied": unknown`

BenTheElder:
repro job.yaml working in kind locally on my linux workstation.

[...]

PR #2748, opened 2022-05-10
containerd: Enable enable_unprivileged_ports and enable_unprivileged_icmp by default

olljanat:
#### Proposed Changes ####
Make using hardened containers a bit easier by:
* Allowing non-root containers to listen ports < 1024
* Allowing ICMP on containers without any capabilities

#### Types of Changes ####
Containerd configuration change. Those settings was added on containerd 1.6.x version.

#### Verification ####
Deploy following test service:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: test
  name: test
spec:
  selector:
    matchLabels:
      app: test
  template:
    metadata:
      labels:
        app: test
    spec:
      containers:
      - name: test
        image: nginx
        ports:
        - containerPort: 80
          protocol: TCP
        volumeMounts:
        - name: nginx-tmp
          mountPath: /var/cache/nginx/
        securityContext:
          capabilities:
            drop:
            - ALL
      securityContext:
        runAsNonRoot: true
        runAsUser: 1001
      volumes:
      - name: nginx-tmp
        emptyDir: {}
```
Check that NGINX is running and listening port 80

On theory also ping with command like:
```bash
kubectl exec -it test-... -- ping 1.1.1.1
```
should works but at least on my env it does not.

However at least sysctl setting `net.ipv4.ping_group_range` got value value `0   2147483647` as expected. It might be configuration issue on my env or to be related the fact that kind run inside of docker.

Change can be tested without building kind with configuration like this:
```yaml
kind: Cluster
name: test
containerdConfigPatches:
- |-
  [plugins."io.containerd.grpc.v1.cri"]
    enable_unprivileged_ports = true
    enable_unprivileged_icmp = true
```

### DIFF
--- a/images/base/files/etc/containerd/config.toml
+++ b/images/base/files/etc/containerd/config.toml
@@ -40,6 +40,10 @@ version = 2
   tolerate_missing_hugepages_controller = true
   # restrict_oom_score_adj needs to be true when running inside UserNS (rootless)
   restrict_oom_score_adj = false
+  # allow listen ports < 1024 for non-root
+  enable_unprivileged_ports = true
+  # allow icmp without capabilities
+  enable_unprivileged_icmp = true
 
 # Enable registry.k8s.io as the primary mirror for k8s.gcr.io
 # See: https://github.com/kubernetes/k8s.io/issues/3411


BenTheElder:
Hmm, I don't know that we should be enabling non-default containerd settings that aren't strictly necessary or documented as part of https://kubernetes.io/docs/setup/production-environment/container-runtimes/

I appreciate the desire to have these settings enabled, but doing so only in kind makes kind an inaccurate way to test applications. I would like to see this enabled, but not so much as a special case in KIND only.

AFAICT https://github.com/kubernetes/kubernetes/issues/102612 did not reach consensus to enable it. https://github.com/containerd/containerd/pull/6170 certainly did not.

>. Change can be tested without building kind with configuration like this:

As you noted these settings can be enabled in your own clusters without being enabled by default by using a containerd config patch (which we need to document under https://kind.sigs.k8s.io/docs/user/configuration/, an example is in https://kind.sigs.k8s.io/docs/user/local-registry/).

> kubectl exec -it test-... -- ping 1.1.1.1

ICMP is generally not enabled by Kubernetes's networking stack like TCP and UDP are.

olljanat:
> AFAICT [kubernetes/kubernetes#102612](https://github.com/kubernetes/kubernetes/issues/102612) did not reach consensus to enable it. [containerd/containerd#6170](https://github.com/containerd/containerd/pull/6170) certainly did not.

Some containerd developers was worry about some possible corner cases / unknown unknowns which might came with this one. I assumed that kind would be good place to try find those so they can be fixed.

FYI. I have same proposal open also on k3s side https://github.com/k3s-io/k3s/pull/5538

BenTheElder:
> Some containerd developers was worry about some possible corner cases / unknown unknowns which might came with this one. I assumed that kind would be good place to try find those so they can be fixed.

Our first and foremost goal still is testing Kubernetes itself https://kind.sigs.k8s.io/docs/contributing/project-scope/

I think edge cases could be discovered by using containerdConfigPAtches to turn it on and explore for those cases, but I don't think we should find them by turning it on and waiting for users to complain. Instead we're more likely to have users complain that they tested their application in kind but then it didn't work on their "real" cluster since there's currently no plan to roll this out in containerd by default or in any of the major vendors, and we try to avoid that.

See e.g. https://github.com/kubernetes-sigs/kind/issues/1331

> FYI. I have same proposal open also on k3s side https://github.com/k3s-io/k3s/pull/5538

ACK, but K3s is very clearly a forked subset of Kubernetes, as a general rule we're not inclined to follow that project.

olljanat:
Ok. I noticed that containerd has added 2.0 milestone so opened this one now https://github.com/containerd/containerd/issues/6924

However it comes to details  `enable_unprivileged_ports` would be enought to handle https://github.com/kubernetes/kubernetes/issues/102612 and I think that it was `enable_unprivileged_icmp` which containerd developers was worry about and I just didn't really realized option to separate those two when worked with PR.

What you mean by saying:
> ICMP is generally not enabled by Kubernetes's networking stack like TCP and UDP are.

That ping test case works fine before dropping all capabilities so I would assume that ICMP need to be enabled at least on some level.

aojea:
does `securityContext` work in all kind environments?
does it work in rootless?

olljanat:
@aojea containerd run PR validations also against of rootless. On rootless `enable_unprivileged_icmp` is actually ignored by https://github.com/containerd/containerd/blob/v1.6.4/pkg/cri/server/sandbox_run_linux.go#L147 as other why it didn't passed those tests.

What comes to `securityContext` like on my example test case above I'm not actually sure if those works on rootless. Maybe someone who have access to that kind of environment should test? It is bit outside of this proposal but valuable information anyway.

aojea:
I prefer to not enable non-defaults options too, we already have a lot of exposure and we don't know the fallout ... people can always patch it in kind for their use cases...

BenTheElder:
as of yet there is no response to https://github.com/containerd/containerd/issues/6924 https://github.com/containerd/containerd/pull/6170#issuecomment-1126322995, I'd be curious to see those responses, but in the current state I think we should stick to:
```yaml
kind: Cluster
name: test
containerdConfigPatches:
- |-
  [plugins."io.containerd.grpc.v1.cri"]
    enable_unprivileged_ports = true
    enable_unprivileged_icmp = true
```

olljanat:
@BenTheElder btw I wonder that has there been yet any discussion about Kubernetes 2.0 and collecting requirements for it? There is so many of those "secure by default" changes which would be needed that perhaps it would make sense to agree that new major version is needed to be able to include those breaking changes and that those need to live side by side quite long time?

I'm thinking changes like that pods should run as non-root, without capabilities and using read only root filesystem by default unless user specify other why.

BenTheElder:
>  I wonder that has there been yet any discussion about Kubernetes 2.0 and collecting requirements for it?

I don't think there's currently any particularly serious discussion.

> There is so many of those "secure by default" changes which would be needed that perhaps it would make sense to agree that new major version is needed to be able to include those breaking changes and that those need to live side by side quite long time?

I think it would be difficult to reach agreement on to be honest, there's usually a lot of pushback against any breaking changes, they do happen though. Which is why I don't think there's any major 2.0 discussion happening, most users want less breakage and more portability, AFAICT.

PR #2749, opened 2022-05-10
Upgrade kindnetd

BenTheElder:
all dependencies, base images, go version


BenTheElder:
I think we want v7 once it's available https://github.com/tonistiigi/binfmt/issues/83

PR #2757, opened 2022-05-11
containerd configuration with drop-in files

sathieu:
See https://github.com/containerd/containerd/blob/main/docs/hosts.md
for `/etc/containerd/certs.d` usage,
and https://github.com/containerd/containerd/blob/main/docs/man/containerd-config.toml.5.md
for `imports`.

### DIFF
--- a/images/base/files/etc/containerd/config.toml
+++ b/images/base/files/etc/containerd/config.toml
@@ -1,6 +1,8 @@
 # explicitly use v2 config format
 version = 2
 
+imports = ["/etc/containerd/conf.d/*.toml"]
+
 [proxy_plugins]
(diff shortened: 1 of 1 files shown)


BenTheElder:
`imports` has pretty crummy behavior unfortunately, I'm not sure we want to encourage this vs `containerdConfigPatches` in kind's config https://github.com/containerd/containerd/issues/3776

IIRC configuring certs.d disables the previous approach with mirror config inline in the containerd config?
Which we have many users using here: https://kind.sigs.k8s.io/docs/user/local-registry/

sathieu:
Abandoning this PR. There are other (more complex IMO) ways to configure this.

Issue #2626, opened 2022-02-10
cmd/load: ability to load from oci layout


Dentrax:
**What would you like to be added**:

Currently, we can load images from tar files or directly a docker image as follows:

```
Available Commands:
  docker-image  Loads docker images from host into nodes
  image-archive Loads docker image from archive into nodes
```

We can also add an `oci-layout` command here to load an image directly from a folder. As a workaround, we have to do:

```
$ skopeo copy oci:<OCI_DIR> docker://<DOCKER_IMAGE>
# and then
$ kind load docker-image <DOCKER_IMAGE>
```

vs.

```
$ kind load oci-dir <OCI_DIR>
```

To export your image as an oci-layout, [ko](https://github.com/google/ko/blob/c67fb03b79b280435074e9c6544310056d697a28/doc/ko_build.md#options) have a support for that, for example.

**Why is this needed**:

To support [OCI image layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md).

cc @developer-guy @imjasonh

BenTheElder:
Is there some source that can only produce OCI? ko also supports docker tarballs instead of OCI.

Note that loading is performed in containerd / ctr. kind only streams the tarball and makes some minor optimizations.

It is important that we minimize the dependencies in the kind source code for embedding e.g. in cluster-api, this would probably have to be done as a feature to ctr.

estesp:
`ctr image import` already supports OCIv1 format:
```
USAGE:
   ctr images import [command options] [flags] <in>

DESCRIPTION:
   Import images from a tar stream.
Implemented formats:
- oci.v1
- docker.v1.1
- docker.v1.2
```

Is that sufficient for what you are looking for or did I misunderstand what you want to add to `ctr`?

Dentrax:
Oh, thanks! Missed that.

In this case, we can easily add a new `oci-dir` subcommand into kind by adding a new function in [nodeutils.go](https://github.com/kubernetes-sigs/kind/blob/master/pkg/cluster/nodeutils/util.go)

```go
func LoadImageFromDir(n nodes.Node, imageDir string) error {
 ...
 cmd := n.Command("ctr", "--namespace=k8s.io", "images", "import", imageDir)
 ...
}
```

As it already describes in [import.go](https://github.com/containerd/containerd/blob/887615c7d044af735f1d177e5afad595c976181c/cmd/ctr/commands/images/import.go#L49).

UX would be:

```
[...]

PR #2758, opened 2022-05-12
WIP: Adds OCI image support

claudiubelu:


BenTheElder:
What's the use case?
https://github.com/kubernetes-sigs/kind/issues/2626

Also we're generally loathe to add dependencies (due to kind being embedded in tools / test suites / ...), though these seem fairly light.

claudiubelu:
> What's the use case? #2626
> 
> Also we're generally loathe to add dependencies (due to kind being embedded in tools / test suites / ...), though these seem fairly light.

My main goal at this point is to be able to have official kube-proxy images for Windows (https://github.com/kubernetes/kubernetes/pull/109939). However, the current image building process for it won't work for it. In k/k, the images are first built with ``docker buildx``, and then exported with ``docker save``. However, that won't work with Windows images at all since they're not available / visible from docker itself (since you cannot have Windows images on Linux). Hence, my idea was to export the image directly from ``docker buildx``, but it seems that ``docker buildx`` exports OCI images, which then causes the CIs that use ``kind`` to fail with ``error building node image: could not find image metadata`` because OCI images are not supported in kind.

However, this might not help me accomplish my goal at this point. I see in ``kubernetes/release`` that when the images are to be released, they are loaded with ``docker load``, which would fail for Windows images because you cannot have Windows images on Linux. I was thinking of adding an option to k/k to allow the ``docker buildx build`` output to be ``registry``, which would mean that the images would be uploaded directly to the given registry, however, I see in ``kubernetes/release`` that the image tars are also uploaded to GCS buckets, which means that the tars are still needed (and ``docker buildx build`` doesn't support multiple outputs, although there's a draft for that feature).

BenTheElder:
> Hence, my idea was to export the image directly from docker buildx, but it seems that docker buildx exports OCI images, which then causes the CIs that use kind to fail with error building node image: could not find image metadata because OCI images are not supported in kind.


Ahhh.

We have also to consider Kubernetes' `hack/local-up-cluster.sh`, I think we've switched from docker to containerd there so we probably have OCI import support.


>  see in kubernetes/release that when the images are to be released, they are loaded with docker load, which would fail for Windows images because you cannot have Windows images on Linux. 

This is probably a remnant for re-tagging / pushing them?
We can probably rework those scripts to avoid actually using docker, https://github.com/google/go-containerregistry/tree/main/cmd/crane

> I was thinking of adding an option to k/k to allow the docker buildx build output to be registry, which would mean that the images would be uploaded directly to the given registry, however, I see in kubernetes/release that the image tars are also uploaded to GCS buckets, which means that the tars are still needed (and docker buildx build doesn't support multiple outputs, although there's a draft for that feature).

I think we could take advantage of caching in buildx and build repeatedly with different output setting to produce the tars, or else we could fetch tars (`crane pull`) from the pushed images, but release may not like the ordering of having the images pushed first.

claudiubelu:
> Also possible `ctr image import` doesnt support OCI yet ... [#2760 (comment)](https://github.com/kubernetes-sigs/kind/issues/2760#issuecomment-1125243694)

containerd supports OCI images:https://github.com/containerd/containerd/blob/b9bffd1f38c7e85f433c22d0968cffb196ede000/images/archive/importer.go#L126

wherka-ama:
> Also possible `ctr image import` doesnt support OCI yet ... [#2760 (comment)](https://github.com/kubernetes-sigs/kind/issues/2760#issuecomment-1125243694)

This issue is a bit old, so my comment may not be relevant. This is a standard flow I use in my daily tinkering:

```
podman build [--squash-all] --tag myimage:0.0.1  .
podman save --format oci-archive --output ../myimage.0.0.1.tar localhost/myimage:0.0.1
KIND_EXPERIMENTAL_PROVIDER=podman kind load image-archive ../myimage.0.0.1.tar
```

I very rarely have a problem with that. The OCI format works perfectly well. Sometimes(very rarely) I need to resort to squashing the image to a single layer(--squash-all) to sort the problems with permissions(I work in rootless context most of times) and/or reduce the overall size of the image. However, I do work on Linux, so that's not quite the same context.

As a side note: I would love to see kind being able to consume the images directly from the docker/podman cache without the need for an interim step of saving the image as tarball or pushing it to a repo.. even if I can easily set such repo up on the same machine and keep pushing the images there. I find it equally (in)convenient to using the image dump. No real improvement in usability.

What would help here a little is to enrich kind with the option to pull the images straight from cache. I guess it would be best to do it explicitly by expressing the intent with an extra parameter in the CLI. I would like to avoid, is to guess the source of an image. We do have such prompt on podman already when using the short image names and multiple repos in the config and it may become a source of confusion very quickly. 
It would be better to use something like --from-build-cache, sniff the OCI complaint layers' store and work with that in order to pump the image to the cluster somehow.

There was some work done on this PR and I think it could be used as a starting point. Alternatively  we could start from scratch and define a new set of requirements for such feature as I don't think it's only about OCI layout support. That support is already available I believe. It is more about taking out that extra interim translation (build -> cache -> **archive ->** cluster). It would be better still to get to the point where we have full transparency where the caching layer of the image builder is shared with the cluster i.e. it is mounted directly there so there is no transformation required. This could be pretty difficult to achieve though.

I could try to use my spare time and help in pushing it a bit further.. unless  somebody tried that before and analysed such flow and there is no point in going there. Just to be clear - I love simplicity, so if it's not simple, I'd rather shoot these extra commands and settle for a reliable flow rather than implementing the monster which would be impossible to manage.

@BenTheElder : do you think it makes sense what I'm suggesting here?

BenTheElder:
> consume the images directly from the docker/podman cache without the need for an interim step of saving the image as tarball or pushing it to a repo.

this is a very complex and brittle feature. at least in docker, e.g. the on-disk image storage is expressly not documented, before we even get into remote daemons. I'm pretty sure podman this is also the case (and podman also has nascent remote support now), not to mention builders like buildx.

image layers are only officially exposed in these tools by way of pushing to a registry, or by way of exporting a complete image or images.

> Alternatively we could start from scratch and define a new set of requirements for such feature as I don't think it's only about OCI layout support. That support is already available I believe. 

This PR is actually totally unrelated to loading images into a cluster and is about the images built by Kubernetes's build system, which are currently always docker tarballs. We need to handle them specially when building node images not quite in the same way as loading at runtime.


> It would be better still to get to the point where we have full transparency where the caching layer of the image builder is shared with the cluster i.e. it is mounted directly there so there is no transformation required. This could be pretty difficult to achieve though.

yes, difficult, non-portable, and depending on internal details of other tools 😬 

and somewhere we still have to create content ingestible by containerd, so we're still probably going to be converting to a tarball somewhere in the pipeline ...

BenTheElder:
I think this is defunct for now, we can revisit later if needed.

PR #2764, opened 2022-05-12
move some flags into kubelet config

BenTheElder:
We can't move node-address, labels, and provider-id, which is just as well because those need to be per-node.

The others we can, and need to, we get warning logs about deprecated flags.

In some future kind release we can stop setting `--cgroup-root` in the base image.
I'm not sure if we want to do that now, because it will break users attempting to use newer images since existing kind releases are not setting it in the config.

This breaks using kind with old images that don't have `/kubelet` cgroup setup, but it's been a few releases since we started using this cgroup root.


BenTheElder:
this requires at least kind v0.10.0 https://github.com/kubernetes-sigs/kind/commit/6ed8f56ac939e5632b326d6a835d79ca4032e4c0

Issue #2765, opened 2022-05-13
[rancher desktop] v0.13.0 not starting in Mac (Intel)


mariomac:
With v0.13.0 trying to create a cluster in Mac OS X fails due a timeout when trying to check the health of the control plane. I reverted back to v0.12.0 and I could create it without any problem.

See logs at the end of the message.

- kind version: (use `kind version`): v0.13.0
- Kubernetes version: (use `kubectl version`): Client Version:
```
version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.5", GitCommit:"c285e781331a3785a7f436042c65c5641ce8a9e9", GitTreeState:"clean", BuildDate:"2022-03-16T15:58:47Z", GoVersion:"go1.17.8", Compiler:"gc", Platform:"darwin/amd64"}
```
- Docker version: (use `docker info`): Using Rancher Desktop for Mac (Intel)
```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.7.1)
  compose: Docker Compose (Docker Inc., v2.2.3)

Server:
 Containers: 1
  Running: 1
  Paused: 0
  Stopped: 0
 Images: 1
 Server Version: 20.10.14
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runtime.v1.linux runc io.containerd.runc.v2
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 3df54a852345ae127d1fa3092b95168e4a88e2f8
 runc version: 52b36a2dd837e8462de8e01458bf02cf9eea47dd
 init version:
 Security Options:
  seccomp
   Profile: default
 Kernel Version: 5.15.32-0-virt
 Operating System: Alpine Linux v3.15
 OSType: linux
 Architecture: x86_64
 CPUs: 2
 Total Memory: 5.809GiB
 Name: lima-rancher-desktop
[...]

Issue #2766, opened 2022-05-13
Is kind w/ podman provisioner for installing k8s v1.24.0 in a WSL instance?


Unveraenderbar:
I know that kind+podman cannot run rootless in a WSL instance -- but installation when running as superuser works fine for k8s v1.23.6
(using podman as runtime).
However, when trying to install k8s v1.24.0 within the same environment, the kubelets in the podman containers fail w/ errors

```
$ sudo sh -c 'podman exec $(podman ps --filter label=io.x-k8s.kind.role=control-plane -q) journalctl -xeu kubelet' | grep -A 10 'kubeletVersion="v1.24.0"'
May 13 10:34:17 k8s-v1.24.0-control-plane kubelet[1044]: I0513 10:34:17.026601    1044 server.go:399] "Kubelet version" kubeletVersion="v1.24.0"
May 13 10:34:17 k8s-v1.24.0-control-plane kubelet[1044]: I0513 10:34:17.026627    1044 server.go:401] "Golang settings" GOGC="" GOMAXPROCS="" GOTRACEBACK=""
May 13 10:34:17 k8s-v1.24.0-control-plane kubelet[1044]: I0513 10:34:17.027340    1044 server.go:813] "Client rotation is on, will bootstrap in background"
May 13 10:34:17 k8s-v1.24.0-control-plane kubelet[1044]: I0513 10:34:17.028219    1044 certificate_store.go:130] Loading cert/key pair from "/var/lib/kubelet/pki/kubele
t-client-current.pem".
May 13 10:34:17 k8s-v1.24.0-control-plane kubelet[1044]: I0513 10:34:17.028665    1044 container_manager_linux.go:848] "CPUAccounting not enabled for process" pid=1044
May 13 10:34:17 k8s-v1.24.0-control-plane kubelet[1044]: I0513 10:34:17.028683    1044 container_manager_linux.go:851] "MemoryAccounting not enabled for process" pid=10
44
May 13 10:34:17 k8s-v1.24.0-control-plane kubelet[1044]: I0513 10:34:17.028733    1044 dynamic_cafile_content.go:157] "Starting controller" name="client-ca-bundle::/etc
/kubernetes/pki/ca.crt"
May 13 10:34:17 k8s-v1.24.0-control-plane kubelet[1044]: W0513 10:34:17.029821    1044 sysinfo.go:203] Nodes topology is not available, providing CPU topology
[...]

PR #2767, opened 2022-05-13
entrypoint cleanup + non-systemd-host fix

BenTheElder:
see https://github.com/kubernetes-sigs/kind/issues/2765, https://github.com/kubernetes-sigs/kind/pull/2743#discussion_r868931694

### DIFF
--- a/images/base/files/etc/systemd/system/kubelet.slice
+++ b/images/base/files/etc/systemd/system/kubelet.slice
@@ -1,0 +1,7 @@
+[Unit]
+Description=slice used to run Kubernetes / Kubelet
+Before=slices.target
+
+[Slice]
+MemoryAccounting=true
+CPUAccounting=true
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -298,9 +298,15 @@ fix_cgroup() {
   mount --make-rprivate /sys/fs/cgroup
   echo "${cgroup_subsystems}" |
   while IFS= read -r subsystem; do
-    mount_kubelet_cgroup_root "/kubelet" "${subsystem}"
-    mount_kubelet_cgroup_root "/kubelet.slice" "${subsystem}"
+    mount_kubelet_cgroup_root /kubelet "${subsystem}"
+    mount_kubelet_cgroup_root /kubelet.slice "${subsystem}"
   done
+  # workaround for hosts not running systemd
+  # we only do this for kubelet.slice because it's not relevant when not using
+  # the systemd cgroup driver
+  if [[ ! "${cgroup_subsystems}" = */sys/fs/cgroup/systemd* ]]; then
+    mount_kubelet_cgroup_root /kubelet.slice /sys/fs/cgroup/systemd
+  fi
 }
 
 fix_machine_id() {
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -234,7 +234,7 @@ fix_cgroup() {
   # See: https://man7.org/linux/man-pages/man7/cgroups.7.html
   echo 'INFO: fix cgroup mounts for all subsystems'
   local cgroup_subsystems
-  cgroup_subsystems=$(findmnt -lun -o source,target -t cgroup | grep "${current_cgroup}" | awk '{print $2}')
+  cgroup_subsystems=$(findmnt -lun -o source,target -t cgroup | grep -F "${current_cgroup}" | awk '{print $2}')
   # Unmount the cgroup subsystems that are not known to runtime used to
   # run the container we are in. Those subsystems are not properly scoped
   # (i.e. the root cgroup is exposed, rather than something like docker/xxxx).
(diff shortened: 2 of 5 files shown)


images/base/files/usr/local/bin/entrypoint
```diff
@@ -55,43 +55,43 @@ validate_userns() {
 }
 
 overlayfs_preferrable() {
-	if [[ -z "$userns" ]]; then
```
BenTheElder, line 58: tabs => double space (like the rest of the file)

images/base/files/usr/local/bin/entrypoint
```diff
@@ -225,9 +227,7 @@ fix_cgroup() {
   # See: https://d2iq.com/blog/running-kind-inside-a-kubernetes-cluster-for-continuous-integration
   # Capture initial state before modifying
   #
-  # Basically we're looking for the cgroup-path for the cpu controller for the
```
BenTheElder, line 228: things got re-arranged, and this comment got out of place.

images/base/files/usr/local/bin/entrypoint
```diff
@@ -298,9 +298,15 @@ fix_cgroup() {
...
   echo "${cgroup_subsystems}" |
   while IFS= read -r subsystem; do
-    mount_kubelet_cgroup_root "/kubelet" "${subsystem}"
-    mount_kubelet_cgroup_root "/kubelet.slice" "${subsystem}"
+    mount_kubelet_cgroup_root /kubelet "${subsystem}"
+    mount_kubelet_cgroup_root /kubelet.slice "${subsystem}"
   done
+  # workaround for hosts not running systemd
+  # we only do this for kubelet.slice because it's not relevant when not using
+  # the systemd cgroup driver
```
BenTheElder, line 306: functional change here

images/base/files/usr/local/bin/entrypoint
```diff
@@ -234,7 +234,7 @@ fix_cgroup() {
   # See: https://man7.org/linux/man-pages/man7/cgroups.7.html
   echo 'INFO: fix cgroup mounts for all subsystems'
   local cgroup_subsystems
-  cgroup_subsystems=$(findmnt -lun -o source,target -t cgroup | grep "${current_cgroup}" | awk '{print $2}')
+  cgroup_subsystems=$(findmnt -lun -o source,target -t cgroup | grep -F "${current_cgroup}" | awk '{print $2}')
```
BenTheElder, line 237: -F / --fixed-strings means match fixed strings, not patterns / regexes.
we're intending to match literal values

images/base/files/usr/local/bin/entrypoint
aojea: @AkihiroSuda @kolyshkin can you please chime in here, I'm really in deep waters here and I think that a another pair of eyes will be nice for this change ... the cgroup glue keeps growing ...

BenTheElder: This doesn't work, wishful thinking :-)

https://github.com/kubernetes-sigs/kind/issues/2766#issuecomment-1126524575

I need to get a v1 host going again and take a peek at what exactly systemd is up to with the pseudo-subsystem.

BenTheElder:
s390x flake https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/2767/pull-kind-build/1527143332810067968

BenTheElder:
`bentheelder/kind-node:v1.24.0@sha256:cf92db251fb3f420634ec46d9c222812fee4d75f1e0105a325e57679b0f57370` is built with this change

I've tested it in colima xref https://github.com/kubernetes-sigs/kind/issues/2778, and this works now.

images/base/files/usr/local/bin/entrypoint
BenTheElder: I setup colima on an old personal machine to debug, this _and_ a real slice works https://github.com/kubernetes-sigs/kind/pull/2767/commits/0ffcf8d6959f98ed038a925830b25f1f77480838.

BenTheElder:
Confirmed WSL 2 https://github.com/kubernetes-sigs/kind/issues/2766#issuecomment-1131477890

I'll bump the node image and then we should ship this to v0.14

images/base/files/etc/systemd/system/kubelet.slice
```diff
@@ -0,0 +1,7 @@
+[Unit]
+Description=slice used to run Kubernetes / Kubelet
+Before=slices.target
+
+[Slice]
+MemoryAccounting=true
```
aojea, line 6: Are these parameters ***Accounting needed twice?

BenTheElder: yes these parameters are for the slice. group vs specific unit.

https://www.freedesktop.org/software/systemd/man/systemd.slice.html

BenTheElder:
Passed across github actions with the new node image.

Tested locally by me in colima and cgroups v2 ~debian. 
Also tested by others on rancher desktop and WSL 2

Base image already tested in all the pull-kubernetes* jobs across k8s versions in prow.

aojea:
👌👍

BenTheElder:
full test runs in #2780, not sure why 1.22 flaked but it only flaked here and not on the 2 subsequent PRs.

PR #2768, opened 2022-05-13
make registry configurable in push-node.sh

BenTheElder:


hack/release/build/push-node.sh
```diff
@@ -46,7 +48,7 @@ kube_version="${version_line#"gitVersion "}"
 # kubernetes build option(s)
 GOFLAGS="${GOFLAGS:-}"
 if [ -z "${GOFLAGS}" ]; then
-    # TODO: add dockerless when 1.19 or greater
+    # TODO: dockerless only applies to < 1.24, the version selection here is brittle
```
aojea, line 51: brittle as it only matches 1.1[0-8] :smile:

BenTheElder: well, it actually matches `1.1[0-8].*`, so when k8s 1.100 comes ;-)

PR #2775, opened 2022-05-17
Fix multi-node cluster not working after restarting docker

### LINKED ISSUE #2759: Support docker restart for clusters with multiple nodes (single controlplane node)

For clusters with multiple nodes like below, once the machine or docker restarts, the cluster cannot recover like clusters with single node.
```
kind: Cluster
nodes:
- role: control-plane
- role: worker
- role: worker
```

The cluster is still reachable but kube-controller-manager and kube-scheduler are not working properly because of the IP address they use to connect kube-apiserver is stale. 
```
1 leaderelection.go:325] error retrieving resource lock kube-system/kube-controller-manager: Get "https://172.18.0.6:6443/apis/coordination.k8s.io/v1/namespaces/kube-system/leases/kube-controller-manager?timeout=5s": dial tcp 172.18.0.6:6443: connect: connection refused
1 reflector.go:138] k8s.io/client-go/informers/factory.go:134: Failed to watch *v1.Service: failed to list *v1.Service: Get "https://172.18.0.6:6443/api/v1/services?limit=500&resourceVersion=0": dial tcp 172.18.0.6:6443: connect: connection refused
```

This is because the IP allocation is not persistent, and there is a great chance the IP of the controlplane node will change after docker restarts. To recover the cluster, two things need to be fixed:
* The IP address in kubeconfigs of kube-controller-manager and kube-scheduler needs to be updated
* The new IP address must be a valid alternative name of kube-apiserver's certificate

There are two possible approaches that can achieve it:

#### 1. Use loopback address to connect kube-apiserver
This leverages the fact that loopback address is a valid alternative name of kube-apiserver's certificate: https://github.com/kubernetes-sigs/kind/blob/b4ab02cffbc8aca096ee7e2bc0f105654f1572c6/pkg/cluster/internal/kubeadm/config.go#L167-L171
However, setting loopback address to kubeconfig of kube-controller-manager and kube-scheduler is not simple. kubeadm uses `localAPIEndpoint.advertiseAddress` parameter to configure that address, but it's also used to determine which address kube-apiserver will listen to. So we cannot just set the parameter (and current kubeadm doesn't accept advertiseAddress set to loopback address).  @BenTheElder suggested the following options at https://github.com/kubernetes-sigs/kind/pull/2671#issuecomment-1116475575:

>  * kubeadm makes it possible to specify this directly
>  * kind uses phases to work around kubeadm
 
I haven't found a proper parameter to specify it in kubeadm directly, and currently kubeadm "patches" don't support patching kubeconfig files. I tried to extend "patches" to kubeconfig via https://github.com/kubernetes/kubernetes/pull/109886 and added patch files to kind via https://github.com/kubernetes-sigs/kind/pull/2671, it could resolve the problem. However, it would require a new Kubernetes release and the certificate of kube-apiserver would still not match the IP address it advertises, though the defect is worked around by making clients use loopback address.

#### 2. Regenerate certificate with new IP

Inspired by the comment in https://github.com/kubernetes-sigs/kind/pull/2671#discussion_r868563989, the entrypoint could regenerate a certificate for kube-apiserver with the new IP and replace the IP address used in kubeconfig files, given that CA cert and key are available on controlplane node.

I have verified it could resolve the problem, and it doesn't require a new Kubernetes release and the certificate matches the new IP. Required change can be found at https://github.com/tnqn/kind/commit/cf95727a5d35360c219604fb0a832395f775c3fb.

Personally I think the second approach is better, I would like hear your opinions before updating the PR. @BenTheElder @aojea @neolit123

tnqn:
In a multi-node cluster with single controlplane node, if the
controlplane node's IP changes, kube-controller-manager and
kube-scheduler would fail to connect kube-apiserver.

This patch fixes it by replacing the server address in kubeconfig files
of kube-controller-manager and kube-scheduler with the new IP and
re-generate a certificate with the new IP for kube-apiserver.

Signed-off-by: Quan Tian <qtian@vmware.com>

Fixes #2759

### DIFF
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -378,6 +378,22 @@ select_iptables() {
   update-alternatives --set ip6tables "/usr/sbin/ip6tables-${mode}" > /dev/null
 }
 
+fix_certificate() {
+  local apiserver_crt_file="/etc/kubernetes/pki/apiserver.crt"
+  local apiserver_key_file="/etc/kubernetes/pki/apiserver.key"
+
+  # Skip if this Node doesn't run kube-apiserver
+  if [[ ! -f ${apiserver_crt_file} ]] || [[ ! -f ${apiserver_key_file} ]]; then
+    return
+  fi
+
+  # Deletes the certificate for kube-apiserver and generates a new one.
+  # This is necessary because the old one doesn't match the current IP.
+  echo 'INFO: clearing and regenerating the certificate for serving the Kubernetes API' >&2
+  rm -f ${apiserver_crt_file} ${apiserver_key_file}
+  kubeadm init phase certs apiserver --config /kind/kubeadm.conf
+}
+
 enable_network_magic(){
   # well-known docker embedded DNS is at 127.0.0.11:53
   local docker_embedded_dns_ip='127.0.0.11'
@@ -416,10 +432,17 @@ enable_network_magic(){
         echo "ERROR: Have an old IPv4 address but no current IPv4 address (!)" >&2
         exit 1
       fi
-      # kubernetes manifests are only present on control-plane nodes
-      sed -i "s#${old_ipv4}#${curr_ipv4}#" /etc/kubernetes/manifests/*.yaml || true
-      # this is no longer required with autodiscovery
-      sed -i "s#${old_ipv4}#${curr_ipv4}#" /var/lib/kubelet/kubeadm-flags.env || true
+      if [[ "${old_ipv4}" != "${curr_ipv4}" ]]; then
+        # kubernetes manifests are only present on control-plane nodes
+        sed -i "s#${old_ipv4}#${curr_ipv4}#" /etc/kubernetes/manifests/*.yaml || true
+        sed -i "s#${old_ipv4}#${curr_ipv4}#" /etc/kubernetes/controller-manager.conf || true
+        sed -i "s#${old_ipv4}#${curr_ipv4}#" /etc/kubernetes/scheduler.conf || true
+        sed -i "s#${old_ipv4}#${curr_ipv4}#" /kind/kubeadm.conf || true
+        # this is no longer required with autodiscovery
+        sed -i "s#${old_ipv4}#${curr_ipv4}#" /var/lib/kubelet/kubeadm-flags.env || true
+        # certificate must match the new IP
+        fix_certificate || true
+      fi
   fi
   if [[ -n $curr_ipv4 ]]; then
     echo -n "${curr_ipv4}" >/kind/old-ipv4
@@ -435,10 +458,17 @@ enable_network_magic(){
       if [[ -z $curr_ipv6 ]]; then
         echo "ERROR: Have an old IPv6 address but no current IPv6 address (!)" >&2
       fi
-      # kubernetes manifests are only present on control-plane nodes
-      sed -i "s#${old_ipv6}#${curr_ipv6}#" /etc/kubernetes/manifests/*.yaml || true
-      # this is no longer required with autodiscovery
-      sed -i "s#${old_ipv6}#${curr_ipv6}#" /var/lib/kubelet/kubeadm-flags.env || true
+      if [[ "${old_ipv6}" != "${curr_ipv6}" ]]; then
+        # kubernetes manifests are only present on control-plane nodes
+        sed -i "s#${old_ipv6}#${curr_ipv6}#" /etc/kubernetes/manifests/*.yaml || true
+        sed -i "s#${old_ipv6}#${curr_ipv6}#" /etc/kubernetes/controller-manager.conf || true
+        sed -i "s#${old_ipv6}#${curr_ipv6}#" /etc/kubernetes/scheduler.conf || true
+        sed -i "s#${old_ipv6}#${curr_ipv6}#" /kind/kubeadm.conf || true
+        # this is no longer required with autodiscovery
+        sed -i "s#${old_ipv6}#${curr_ipv6}#" /var/lib/kubelet/kubeadm-flags.env || true
+        # certificate must match the new IP
+        fix_certificate || true
+      fi
   fi
   if [[ -n $curr_ipv6 ]]; then
     echo -n "${curr_ipv6}" >/kind/old-ipv6
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -29,7 +29,12 @@ fi
 
 grep_allow_nomatch() {
   # grep exits 0 on match, 1 on no match, 2 on error
-  grep "$@" || [[ $? == 1 ]]
+  grep "$@"|| [[ $? == 1 ]]
+}
+
+# regex_escape_ip converts IP address string $1 to a regex-escaped literal
+regex_escape_ip(){
+  sed -e 's#\.#\\.#g' -e 's#\[#\\[#g' -e 's#\]#\\]#g' <<<"$1"
 }
 
 validate_userns() {
@@ -434,12 +439,13 @@ enable_network_magic(){
       fi
       if [[ "${old_ipv4}" != "${curr_ipv4}" ]]; then
         # kubernetes manifests are only present on control-plane nodes
-        sed -i "s#${old_ipv4}#${curr_ipv4}#" /etc/kubernetes/manifests/*.yaml || true
-        sed -i "s#${old_ipv4}#${curr_ipv4}#" /etc/kubernetes/controller-manager.conf || true
-        sed -i "s#${old_ipv4}#${curr_ipv4}#" /etc/kubernetes/scheduler.conf || true
-        sed -i "s#${old_ipv4}#${curr_ipv4}#" /kind/kubeadm.conf || true
+        sed_ipv4_command="s#$(regex_escape_ip "${old_ipv4}")#${curr_ipv4}#g"
+        sed -i "${sed_ipv4_command}" /etc/kubernetes/manifests/*.yaml || true
+        sed -i "${sed_ipv4_command}" /etc/kubernetes/controller-manager.conf || true
+        sed -i "${sed_ipv4_command}" /etc/kubernetes/scheduler.conf || true
+        sed -i "${sed_ipv4_command}" /kind/kubeadm.conf || true
         # this is no longer required with autodiscovery
-        sed -i "s#${old_ipv4}#${curr_ipv4}#" /var/lib/kubelet/kubeadm-flags.env || true
+        sed -i "${sed_ipv4_command}" /var/lib/kubelet/kubeadm-flags.env || true
         # certificate must match the new IP
         fix_certificate || true
       fi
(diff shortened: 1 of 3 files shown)


images/base/files/usr/local/bin/entrypoint
```diff
@@ -330,6 +330,24 @@ select_iptables() {
...
+  # Skip if this Node doesn't run kube-apiserver
+  if [[ ! -f ${apiserver_crt_file} ]] || [[ ! -f ${apiserver_key_file} ]]; then
+    return
+  fi
+
+  # Deletes the certificate for kube-apiserver and generates a new one.
+  # This is necessary because the old one doesn't match the current IP.
+  echo 'INFO: clearing and regenerating the certificate for serving the Kubernetes API' >&2
+  rm -f ${apiserver_crt_file} ${apiserver_key_file}
+  kubeadm init phase certs apiserver --config /kind/kubeadm.conf
```
BenTheElder, line 348: what version added this command?

images/base/files/usr/local/bin/entrypoint
```diff
@@ -368,10 +386,17 @@ enable_network_magic(){
         echo "ERROR: Have an old IPv4 address but no current IPv4 address (!)" >&2
         exit 1
       fi
-      # kubernetes manifests are only present on control-plane nodes
-      sed -i "s#${old_ipv4}#${curr_ipv4}#" /etc/kubernetes/manifests/*.yaml || true
-      # this is no longer required with autodiscovery
-      sed -i "s#${old_ipv4}#${curr_ipv4}#" /var/lib/kubelet/kubeadm-flags.env || true
+      if [[ "${old_ipv4}" != "${curr_ipv4}" ]]; then
+        # kubernetes manifests are only present on control-plane nodes
+        sed -i "s#${old_ipv4}#${curr_ipv4}#" /etc/kubernetes/manifests/*.yaml || true
```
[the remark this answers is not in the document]
tnqn: It could work but it would be a change irrelevant to the PR's purpose as the script has been using sed from the beginning and `kubeadm.conf`, `kubeadm-flags.env` will still need sed. It may require a few more lines to check if this is control-plane node before invoking `kubeadm init phase control-plane`. 
Do we want to make the change in this PR?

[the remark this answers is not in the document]
BenTheElder: Sed means we retain any user changes performed manually out of band except the targeted edit we're making.

images/base/files/usr/local/bin/entrypoint
tnqn: Should be v1.8.0: https://github.com/kubernetes/kubernetes/pull/48196

BenTheElder: excellent! kind has never worked with lower than v1.10 even :-)

images/base/files/usr/local/bin/entrypoint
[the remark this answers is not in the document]
tnqn: thanks for commenting, then I suppose we could keep sed

BenTheElder:
Sorry, I am going to finish shipping a release to fix non-systemd hosts after the v0.13 changes before we ship any further potentially problematic changes https://github.com/kubernetes-sigs/kind/pull/2767

BenTheElder:
This is the first thing I'd like to see in v0.15.0, assigning to see this through next.

images/base/files/usr/local/bin/entrypoint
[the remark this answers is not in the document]
BenTheElder: we should however, do some fixes so we replace the actual literals instead of letting the IP address be mis-interpreted as a regex, though that bug is not new to this PR.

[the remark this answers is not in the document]
BenTheElder: rebased the PR (because it's based on out of date version of entrypoint script) and then fixed this, we regex escape the ip addresses now (I think ...)

images/base/files/usr/local/bin/entrypoint
```diff
@@ -29,7 +29,12 @@ fi
 
 grep_allow_nomatch() {
   # grep exits 0 on match, 1 on no match, 2 on error
-  grep "$@" || [[ $? == 1 ]]
+  grep "$@"|| [[ $? == 1 ]]
+}
+
+# regex_escape_ip converts IP address string $1 to a regex-escaped literal
+regex_escape_ip(){
+  sed -e 's#\.#\\.#g' -e 's#\[#\\[#g' -e 's#\]#\\]#g' <<<"$1"
```
aojea, line 37: is this valid for ipv6?

BenTheElder: IPv6 has `[` potentially right? but no `.`
the colons shouldn't need escaping. but square brackets and periods are meaningful in regex and need escaping.

BenTheElder: we're just replacing regex-meaningful characters with the character prefixed by `\`, and only handling the subset we expect to appear in ip addresses, because that's a simpler problem than doing it for _all_ possible characters

BenTheElder:
/hold
I have not tested building a node image and then rebooting a multi-node cluster with the latest changes, even though they should be "more correct" we should test that :-)

images/base/files/usr/local/bin/entrypoint
```diff
@@ -378,6 +383,22 @@ select_iptables() {
   update-alternatives --set ip6tables "/usr/sbin/ip6tables-${mode}" > /dev/null
 }
 
+fix_certificate() {
```
aojea, line 386: we are calling this twice if both ips, the ipv4 and ipv6 has changed, is that the expected?

another question, is this keeping the localhost address, that we added, on the certificate?

BenTheElder: > we are calling this twice if both ips, the ipv4 and ipv6 has changed, is that the expected?

good point, we should store if either changed and then run it once at the end.

> another question, is this keeping the localhost address, that we added, on the certificate?

we use the same kubeadm config (`kind/kubeadm.conf`) so it should be

images/base/files/usr/local/bin/entrypoint
```diff
@@ -416,10 +437,18 @@ enable_network_magic(){
...
+        # kubernetes manifests are only present on control-plane nodes
+        sed_ipv4_command="s#$(regex_escape_ip "${old_ipv4}")#${curr_ipv4}#g"
+        sed -i "${sed_ipv4_command}" /etc/kubernetes/manifests/*.yaml || true
+        sed -i "${sed_ipv4_command}" /etc/kubernetes/controller-manager.conf || true
+        sed -i "${sed_ipv4_command}" /etc/kubernetes/scheduler.conf || true
+        sed -i "${sed_ipv4_command}" /kind/kubeadm.conf || true
+        # this is no longer required with autodiscovery
+        sed -i "${sed_ipv4_command}" /var/lib/kubelet/kubeadm-flags.env || true
+        # certificate must match the new IP
+        fix_certificate || true
```
BenTheElder, line 450: lets store above `local should_fix_certificate=false` and then in both the ipv4 and ipv6 paths do `should_fix_certificate=true` here.

then run this only once below.

BenTheElder:
dropped the base image bump while we iterate more, @tnqn handing back to you.

very excited about this PR :-)

images/base/files/usr/local/bin/entrypoint
aojea: also, do we need to swallow the exit status of the function?
should not better do that inside and not have to do it here?

BenTheElder: i think that's just matching the existing pattern for ip change above.

we swallow the error replacing file contents currently. we probably should not, we should check for files and error if trying to replace in them fails.

images/base/files/usr/local/bin/entrypoint
```diff
@@ -416,10 +437,18 @@ enable_network_magic(){
         echo "ERROR: Have an old IPv4 address but no current IPv4 address (!)" >&2
         exit 1
       fi
-      # kubernetes manifests are only present on control-plane nodes
-      sed -i "s#${old_ipv4}#${curr_ipv4}#" /etc/kubernetes/manifests/*.yaml || true
-      # this is no longer required with autodiscovery
-      sed -i "s#${old_ipv4}#${curr_ipv4}#" /var/lib/kubelet/kubeadm-flags.env || true
+      if [[ "${old_ipv4}" != "${curr_ipv4}" ]]; then
+        # kubernetes manifests are only present on control-plane nodes
+        sed_ipv4_command="s#$(regex_escape_ip "${old_ipv4}")#${curr_ipv4}#g"
```
BenTheElder, line 442: note that when I did the regex fix I also:
- stopped computing this string repeatedly
- changed it to a global replace (!) instead of just replacing the first instance

the existing sed pattern carried forward from the existing script was not good :(
it should be reasonable now.

images/base/files/usr/local/bin/entrypoint
tnqn: thanks for pointing it out, updated to regenerate only once

images/base/files/usr/local/bin/entrypoint
tnqn: updated to check file existence first and not swallow error

images/base/files/usr/local/bin/entrypoint
tnqn: Thanks!
Reminded by it, I also added word boundary to sed pattern to avoid mismatch like
```
# sed "s/1\.1\.1\.2/1.1.1.3/g" <<< "1.1.1.22"
1.1.1.32
```

images/base/files/usr/local/bin/entrypoint
```diff
@@ -29,7 +29,12 @@ fi
 
 grep_allow_nomatch() {
   # grep exits 0 on match, 1 on no match, 2 on error
-  grep "$@" || [[ $? == 1 ]]
+  grep "$@"|| [[ $? == 1 ]]
```
tnqn: I guess this is an unintended change and reverted it in 3rd patch, please let me know if this is on purpose

tnqn:
I have tested the change by restarting IPv4, IPv6 and dualstack clusters and they worked as expected.

aojea:
/retest

> error: no builder "kind-builder" found

images/base/files/usr/local/bin/entrypoint
```diff
@@ -435,14 +464,25 @@ enable_network_magic(){
...
         echo "ERROR: Have an old IPv6 address but no current IPv6 address (!)" >&2
       fi
-      # kubernetes manifests are only present on control-plane nodes
-      sed -i "s#${old_ipv6}#${curr_ipv6}#" /etc/kubernetes/manifests/*.yaml || true
-      # this is no longer required with autodiscovery
-      sed -i "s#${old_ipv6}#${curr_ipv6}#" /var/lib/kubelet/kubeadm-flags.env || true
+      if [[ "${old_ipv6}" != "${curr_ipv6}" ]]; then
+        sed_ipv6_command="s#\b$(regex_escape_ip "${old_ipv6}")\b#${curr_ipv6}#g"
+        # kubernetes manifests are only present on control-plane nodes
+        sed -i "${sed_ipv6_command}" /etc/kubernetes/manifests/*.yaml || true
```
BenTheElder, line 470: we can loop over these as well instead, just add it to the loop below?

BenTheElder: er no, if we put the wildcard in the list of for loop items it will expand inline as individual loop items

BenTheElder: (and if nothing matches the wildcard there will be no additional loop items)

tnqn: Thanks for the tips. It seems when any file matches, it's expanded as expected, but it's not expanded when nothing matches, it will check file existence with wildcard, though the check will return false which will prevent further operation, do you think this is expected? is it legal to check file existence with wildcard?
```
# for f in /etc/emptydir/*.yaml /etc/kubernetes/controller-manager.conf; do echo "testing $f"; done
testing /etc/emptydir/*.yaml
testing /etc/kubernetes/controller-manager.conf
```

BenTheElder: 🤯 I totally forgot that no match => the wildcard pattern itself 🤯 

let's just keep the explicit file list + check for now, thanks!

BenTheElder:
Tentatively unavailable tomorrow, top of my list to revisit when I am back.

Still the first thing I intend to merge in v0.15.0, I think this is about good to go, but I want to double check everything again 😅 

Thanks again for working on this 🙏

BenTheElder:
it looks good, pushing images so we can soak all the CI

BenTheElder (approved):
/lgtm
/approve
/hold cancel
thank you!!! 😄

PR #2777, opened 2022-05-18
kubeadminit: handle the "control-plane" taint for 1.25-pre

neolit123:
The 1.25 prerelease of kubeadm applies only the
new "control-plane" taint. PR: https://github.com/kubernetes/kubernetes/pull/110095

For versions >= 1.25.0-0 only remove the new taint.
Remove both old and new taints for versions >= 1.24.0-0,
but < 1.25.0-0.
Continue removing only the old taint for versions <1.24.0-0.

xref https://github.com/kubernetes/kubeadm/issues/2200
https://github.com/kubernetes-sigs/kind/issues/1699

/kind cleanup

### DIFF
--- a/pkg/cluster/internal/create/actions/kubeadminit/init.go
+++ b/pkg/cluster/internal/create/actions/kubeadminit/init.go
@@ -111,7 +111,7 @@ func (a *action) Execute(ctx *actions.ActionContext) error {
 	// https://kubernetes.io/docs/setup/independent/create-cluster-kubeadm/#master-isolation
 	if len(allNodes) == 1 {
 		// TODO: Once kubeadm 1.23 is no longer supported remove the <1.24 handling.
-		// TODO: Remove only the "control-plane" taint for kubeadm >= 1.25.
+		// TODO: Once kubeadm 1.24 is no longer supported remove the <1.25 handling.
 		// https://github.com/kubernetes-sigs/kind/issues/1699
 		rawVersion, err := nodeutils.KubeVersion(node)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


BenTheElder:
this technically introduces a commit range where we'll be broken, right?

neolit123:
> this technically introduces a commit range where we'll be broken, right?

yes, it might be better to pin exact commits like you stated on the [related issue](https://github.com/kubernetes-sigs/kind/issues/1699#issuecomment-1048595301). a breakage can only happen for users that are building a node image from source for a commit that is not handled by the version ranges in this PR and if they deploy kind with a single node.

then again, they might as well just do:
> At HEAD of kind HEAD of Kubernetes should work, generally.

last time there was a single user report about it (for 1.24.0-alpha.3):
https://github.com/kubernetes-sigs/kind/issues/1699#issuecomment-1048269832

i guess the problem commit range right now is between the 1.25.0-alpha.0 tag and the commit from [this PR](https://github.com/kubernetes/kubernetes/pull/110095).

BenTheElder:
`v1.25.0-alpha.0.557+84c8afeba39ec9` is the precise version. (`git fetch upstream && git checkout [84c8afeba39ec9ebbb231c67daf9862077299e01](https://github.com/kubernetes/kubernetes/commit/84c8afeba39ec9ebbb231c67daf9862077299e01 && hack/print-workspace-status.sh`)

meant to let this merge and then follow-up with the precise version
/lgtm

I think it's worth supporting more than just HEAD with kind @ HEAD where feasible for the purposes of e.g. `git bisect`.

PR #2779, opened 2022-05-19
use precise versions for kubeadm label / taint changes

BenTheElder:
follow up to https://github.com/kubernetes-sigs/kind/pull/2777, https://github.com/kubernetes-sigs/kind/issues/1699#issuecomment-1048595301 

there are 3 versions here:
- when we removed the old label
- when we added the new taint
- when we removed the old taint

versions are from `hack/print-workspace-status.sh` when checked out to the commit in kubernetes, after fetching the latest upstream changes

### DIFF
--- a/pkg/cluster/internal/create/actions/kubeadminit/init.go
+++ b/pkg/cluster/internal/create/actions/kubeadminit/init.go
@@ -122,10 +122,10 @@ func (a *action) Execute(ctx *actions.ActionContext) error {
 			return errors.Wrap(err, "could not parse Kubernetes version")
 		}
 		var taints []string
-		if kubeVersion.LessThan(version.MustParseSemantic("v1.24.0-0")) {
+		if kubeVersion.LessThan(version.MustParseSemantic("v1.24.0-alpha.1.592+370031cadac624")) {
 			// for versions older than 1.24 prerelease remove only the old taint
 			taints = []string{"node-role.kubernetes.io/master-"}
-		} else if kubeVersion.LessThan(version.MustParseSemantic("v1.25.0-0")) {
+		} else if kubeVersion.LessThan(version.MustParseSemantic("v1.25.0-alpha.0.557+84c8afeba39ec9")) {
 			// for versions between 1.24 and 1.25 prerelease remove both the old and new taint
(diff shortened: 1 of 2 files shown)


BenTheElder:
I took a look around and other places where we select based on Kubernetes version remain release versions, because kind is just using them for consistency for when kind opts to switch to a new API etc, not because of a breaking change.

BenTheElder:
I'm going to cut a v0.14.0 release with the systemd fix, this change is trivial but I'd like to include it as it shouldn't be breaking but it should let us wrap up https://github.com/kubernetes-sigs/kind/issues/1699 in this release.

Since CI doesn't cover in this fine grained detail anyhow, I'm just going to merge it, and get about cutting the release.

BenTheElder:
full tests anyhow in https://github.com/kubernetes-sigs/kind/pull/2780

ckotzbauer:
Hi @BenTheElder,
just a quick question (maybe it is unrelated to this PR): When I create a 5 node cluster (1.24.0) with kind 0.14.0 all nodes except one worker-node are tainted with control-plane and master. I would expect, that there's only a control-plane taint. Did I miss something here?

Kind-Config:
```yaml
kind: Cluster
nodes:
- role: control-plane
  image: "kindest/node:v1.24.0"
- role: control-plane
  image: "kindest/node:v1.24.0"
- role: control-plane
  image: "kindest/node:v1.24.0"
- role: worker
  image: "kindest/node:v1.24.0"
- role: worker
  image: "kindest/node:v1.24.0"
```

BenTheElder:
We don't add taints to any nodes. KIND will remove the control-plane / master taint from a control plane node in a single node cluster in order to make "kind create cluster" have a schedulable node. It will not manipulate taints in any other situation.

This PR (hopefully...) makes it more precise which between-release Kubernetes versions the different taint and label changes are handled in kind.

I wouldn't expect kubeadm to be tainting either of the workers with this taint. Perhaps one of the workers was tainted not ready instead?

https://github.com/kubernetes/kubeadm/issues/2200

ckotzbauer:
Yeah, you are right. One of the workers was tainted with `not-ready`. Thanks for your quick response!

PR #2780, opened 2022-05-19
mark release v0.14.0

BenTheElder:
/hold

I'm still pushing the images. I have release notes written, binaries uploaded, just waiting to update the image list with the digests.

Will unhold and merge, then push the tag once that's done.


BenTheElder:
/hold cancel
last image is pushed

Issue #2576, opened 2021-12-30
Cannot use HostToContainer mount propagation on MacOS


fardarter:
https://github.com/kubernetes-sigs/kind/blob/bdfe777928f675ebcecc838511d173c36ba6d789/site/static/examples/config-with-mounts.yaml#L18

Cannot use HostToContainer on MacOS as per: https://github.com/kubernetes/kubernetes/issues/61058

aojea:
> https://github.com/kubernetes-sigs/kind/blob/bdfe777928f675ebcecc838511d173c36ba6d789/site/static/examples/config-with-mounts.yaml#L18
> 
> Cannot use HostToContainer on MacOS as per: [kubernetes/kubernetes#61058](https://github.com/kubernetes/kubernetes/issues/61058)

that is kind config not kubernetes config, that configuration works in Mac

fardarter:
These is discussion on the linked thread.

The error I get is identical: 

```
ERROR: failed to create cluster: docker run error: command "docker run --hostname kind-control-plane --name kind-control-plane --label io.x-k8s.kind.role=control-plane --privileged --security-opt seccomp=unconfined --security-opt apparmor=unconfined --tmpfs /tmp --tmpfs /run --volume /var --volume /lib/modules:/lib/modules:ro --detach --tty --label io.x-k8s.kind.cluster=kind --net kind --restart=on-failure:1 --init=false --volume=/path/WithTestData:/WithTestData:ro,rslave --publish=0.0.0.0:80:80/TCP --publish=0.0.0.0:443:443/TCP --publish=0.0.0.0:31433:31433/TCP --publish=127.0.0.1:64174:6443/TCP -e KUBECONFIG=/etc/kubernetes/admin.conf kindest/node:v1.21.1@sha256:69860bda5563ac81e3c0057d654b5253219618a22ec3a346306239bba8cfa1a6" failed with error: exit status 125
Command Output: 80f688eb0b45e1d5e8db561d1748a602581be5c6834a3a2df09a13cf52e43953
docker: Error response from daemon: path /host_mnt//path/WithTestData is mounted on /host_mnt but it is not a shared or slave mount.
make: *** [set-up-kind] Error 1
```

aojea:
The error may be identical , but in this case is docker who fails to mount the volume `docker run --hostname ... --volume=/path/WithTestData:/WithTestData:ro,rslave `

[...]

Issue #1966, opened 2020-12-08
Kind 0.9.0 does not support docker volumes


mtaivalj:
In kind tool codebase there is utility code making hostpaths absolute when external mounts are declared for kind nodes via configuration. This blocks using docker created volumes, which are easibly provisioned via e.g. docker-compose. Also it bounds the host for kind tool to the kind nodes making portability issues when host needs to be changed. Please implement support for docker volumes also in addition to hostpaths as this is provided by docker out-of-the box and kind is intended to be used on top of docker.

BenTheElder:
> In kind tool codebase there is utility code making hostpaths absolute when external mounts are declared for kind nodes via configuration. 

yes

> This blocks using docker created volumes, which are easibly provisioned via e.g. docker-compose. 

yes. This is not supported and was never supported.

> Also it bounds the host for kind tool to the kind nodes making portability issues when host needs to be changed. 

No? You should use generally relative paths in your configuration, and then when the particular *container* is running _at runtime_ kind will resolve it to an absolute path, _because docker requires this_.

> Please implement support for docker volumes also in addition to hostpaths as this is provided by docker out-of-the box and kind is intended to be used on top of docker.

No, kind is intended to _leverage_ docker to create clusters. It's not intended to compete with docker compose.
The only reason kind even has volume mounts is so you can pass through files from your host easily as a power user feature. 

Otherwise you should generally not be depending on filesystem within the kind nodes anyhow, this is non-portable usage of Kubernetes. Many cluster users won't even have host filesystem access.

BenTheElder:
They're called "[extra mounts](https://kind.sigs.k8s.io/docs/user/configuration/#extra-mounts)" for a reason, they are to support extra [bind mounts](https://docs.docker.com/storage/bind-mounts/).

You should not need volumes, and using volumes by absuing the way extraMounts is imiplemented is *not* supported in any version of kind.

BenTheElder:
The additional missing context: lack of support for relative paths in config in previous versions actually meant that your kind configs _were not_ portable, you had to do something like generate them with a script.

By adding the relative path resolution, this allowed you to have a static config in some repo for testing it that *is* portable because the paths in the config are relative.

see the original PR https://github.com/kubernetes-sigs/kind/pull/910

there should be release notes related to the relative path enhancement.

BenTheElder:
The relative path config is working as intended. However I'll keep an eye on this for comments as to why we might need a *different* feature with named volumes instead.

[...]

PR #2783, opened 2022-05-21
document unavailability of mount propogation

doehyunbaek:
Hi. I really did write small writeup concerning unavailability of mount propagation on Docker Desktop with a pointer to the issue thread(https://github.com/kubernetes-sigs/kind/issues/2576).

The thing I'm confused is, is the note below still relevant? As I read your explaniation on the issue thread, setting hostPath in Docker desktop configuration would not really help. Is there something I am missing? If that explaination is misleading, maybe I should delete that note as well? 

https://github.com/kubernetes-sigs/kind/blob/80836a7ecbadcc3864ae295ff8e1a06877e81674/site/content/docs/user/configuration.md?plain=1#L287-L290

### DIFF
--- a/site/content/docs/user/configuration.md
+++ b/site/content/docs/user/configuration.md
@@ -289,6 +289,11 @@ included in the Preferences -> Resources -> File Sharing.
 
 For more information see the [Docker file sharing guide.](https://docs.docker.com/docker-for-mac/#file-sharing)
 
+**NOTE**: Mount propagation is not possible with Mac or Windows Docker Desktop,
+due to limitation of Docker Desktop implementation. 
+
+For more information, see this [issue thread](https://github.com/kubernetes-sigs/kind/issues/2576)
+
 ### Extra Port Mappings
 
 Extra port mappings can be used to port forward to the kind nodes. This is a 


doehyunbaek:
closes #2576

BenTheElder:
> The thing I'm confused is, is the note below still relevant? As I read your explaniation on the issue thread, setting hostPath in Docker desktop configuration would not really help. Is there something I am missing? If that explaination is misleading, maybe I should delete that note as well?

Yes it is still relevant, you cannot mount a path from the host that is not shared in the docker desktop configuration.
That's not related to mount propagation.

site/content/docs/user/configuration.md
```diff
@@ -289,6 +289,11 @@ included in the Preferences -> Resources -> File Sharing.
 
 For more information see the [Docker file sharing guide.](https://docs.docker.com/docker-for-mac/#file-sharing)
 
+**NOTE**: Mount propagation is not possible with Mac or Windows Docker Desktop,
+due to limitation of Docker Desktop implementation.
```
BenTheElder, line 293: I'm not sure how to tersely cover this off the top of my head but, this is not quite right.

You cannot use mount propagation *when the mount is to a path on the mac or windows host* *when using docker desktop* (and presumably other tools, but I can't say for certain, I don't other VM <> host boundaries can handle this either).

When using docker desktop you could still use this setting *if* the hostPath is to a location on the VM and not the mac/windows host. Which is generally unusual I think, but possible.

If #1966 lands, it should also be possible for that, because volumes are generally contained within the VM disk.

BenTheElder: So basically ... mount propagation doesn't work across the macOS / Windows | Docker Desktop Virtual Machine boundary. Mounts that cross that boundary can't be expected to support this, and that limitation comes from Docker.

doehyunbaek:
Thanks for your quick and kind reply. Here are some thoughts,

> When using docker desktop you could still use this setting if the hostPath is to a location on the VM and not the mac/windows host. Which is generally unusual I think, but possible.

I am unaware of this possibility. If you know any way to refer to VM path in docker cli with Docker desktop for mac(God this is wordy XD). Can you let me know? I have tinkered with both -v and --mount option and could not find success. From design perspective also it seems unreasonable for docker folks to let source be physical(mac) machine path and virtual(linux) machine path. Maybe this is the reason why docker docs resorts to just 'Mount propagation doesn't work with docker desktop'(https://github.com/docker/docker.github.io/pull/14355). And I think this statement is not only useful but also valid if there is no way to specificy vm path as bind source.

I think it is true that mount propagation is pretty advanced configuration as mentioned in [docker docs](https://docs.docker.com/storage/bind-mounts/#:~:text=Bind%20propagation%20is%20an%20advanced%20topic%20and%20many%20users%20never%20need%20to%20configure%20it.) and having this setting as HostToContainer [example](https://github.com/kubernetes-sigs/kind/blob/47d47043e041536ffa5d64730ff593e991130de3/site/static/examples/config-with-mounts.yaml#L18) trips many people up. So my suggestion is setting this option in example to None, as that would prevent many newcomers copy-pasting example and finding out it does not work and getting into rabbit hole world of linux mount propagation.

Also the fact that bind propagation would be mostly useless for non-advanced usecases(I tried to find compelling example for its use but couldn't find any... one I got [here](https://medium.com/kokster/kubernetes-mount-propagation-5306c36a4a2d#:~:text=available%20here.-,Use%20Cases,-Mount%20propagation%20mode) seems to be not so relevant for beginners.) supports this approach. And I presume people who need these advanced usecases would be familiar with these topics and could advance in their own with a little pointer to these threads.

So In summary, I suggest

1. Chaning propagation option in config-with-mounts.yaml from HostToContainer to None.

2-a. Provided there is no way to configure vm path as bind source, just say 'Bind propagation is not possible for docker desktop for mac or windows'

2-b. If there is some arcane way to configure vm path as bind source, be more specific and say 'Bind propagation does not work across VM boundaries. e.g. You can not bind propagate your mac host directory into container. use None option instead'.

Again, thanks for your past and current reply. Learned a lot investigating :)

BenTheElder:
> I am unaware of this possibility. If you know any way to refer to VM path in docker cli with Docker desktop for mac(God this is wordy XD), can you let me know?

KIND has some internally, for example `--volume /lib/modules:/lib/modules` or `--volume /dev/mapper:/dev/mapper` when appropriate. `/lib/modules` is the Linux VM's kernel modules and `/dev/mapper` the VM's device mapper, not something from the host (these don't exist on macOS or Windows).

> From design perspective also it seems unreasonable for docker folks to let source be both physical(mac) machine path and virtual(linux) machine path.

Not at all ... If the path isn't shared from mac and does exist on the VM, it makes sense to mount it from the VM, this just makes it behave the same as docker on linux without the docker desktop magic. The docker desktop filesystem magic comes into play for paths configured in the application.


> I think it is true that mount propagation is pretty advanced configuration as mentioned in [docker docs](https://docs.docker.com/storage/bind-mounts/#:~:text=Bind%20propagation%20is%20an%20advanced%20topic%20and%20many%20users%20never%20need%20to%20configure%20it.) and having this setting as HostToContainer [example](https://github.com/kubernetes-sigs/kind/blob/47d47043e041536ffa5d64730ff593e991130de3/site/static/examples/config-with-mounts.yaml#L18) trips many people up. So my suggestion is setting this option in example to None, as that would prevent many newcomers copy-pasting example and finding out it does not work and getting into rabbit hole world of linux mount propagation.


Sure, setting the default to `None` would be fine.

I think better yet, we should have an example of reasonable *actual* usage that doesn't bother with unnecessary fields, however we still need to document that they're available.

doehyunbaek:
Ah, my confusion was that I thought source path in -v option refers to path in physical machine, but apparently it is not. As can be shown from below example,
```
# fails because sharing is not configured and magic does not work.
$ docker run -it -v /Applications:/don ubuntu bash
docker: Error response from daemon: Mounts denied: 
The path /Applications is not shared from the host and is not known to Docker.
You can configure shared paths from Docker -> Preferences... -> Resources -> File Sharing.
See https://docs.docker.com/desktop/mac for more info.
ERRO[0000] error waiting for container: context canceled 
# succeeds because sharing is configured for /Users 
$ docker run -it -v /Users:/don ubuntu bash
/ # 
# succeeds because source path refers to /lib/modules in vm, not mac physical machine. /lib/modules does not exist in physical machine.
bash-5.1$ docker run -it -v /lib/modules:/don ubuntu bash
root@ad1d52570788:/# ls /don
5.10.104-linuxkit
```

It seems to make more sense this way as mount is happening at where the docker daemon is hosted not where cli lives. Thanks for correction.

I'm not sure how to progress on this issue though. As of actual usage, I am pretty fine using kind cluster with not much confiuration in my daily usage, so I wouldn't be able to find one in my configuration.  But I can look for any sensible usage of kind.x-k8s.io/v1alpha4 configuration throught github.

doehyunbaek:
I'm closing this PR as I'm not sure how to progress. Will reopen if I get some clue.

PR #2786, opened 2022-05-25
GIT-2663: optimized image load with re-tag support

### LINKED ISSUE #2663: kind load docker-image should add a tag instead of loading the image when possible
**What happened**:
Loading the same image with different tag loads the image again to kind:

```bash
$ kind load docker-image demo/demo1:v1
Image: "demo/demo1:v1" with ID "sha256:e9d87b36eeece1dac52c002a35d75205bdc4ea142e091f6b45861a78afbfb839" not yet present on node "kind-control-plane", loading...
$ kind load docker-image demo/demo1:latest
Image: "demo/demo1" with ID "sha256:e9d87b36eeece1dac52c002a35d75205bdc4ea142e091f6b45861a78afbfb839" not yet present on node "kind-control-plane", loading...
```

An image with the same hash (`sha256:e9d87b36eeece1dac52c002a35d75205bdc4ea142e091f6b45861a78afbfb839` in the previous example) should only need to add a "tag" instead of loading the image again.

Something like this should be expected:
Image: ID "sha256:e9d87b36eeece1dac52c002a35d75205bdc4ea142e091f6b45861a78afbfb839" found to be already present on all nodes, creating tag for "demo/demo1" .

```bash
$ kind load docker-image demo/demo1:v1
Image: "demo/demo1:v1" with ID "sha256:e9d87b36eeece1dac52c002a35d75205bdc4ea142e091f6b45861a78afbfb839" not yet present on node "kind-control-plane", loading...
$ kind load docker-image demo/demo1:latest
Image: "demo/demo1" with ID "sha256:e9d87b36eeece1dac52c002a35d75205bdc4ea142e091f6b45861a78afbfb839" not yet present on node "kind-control-plane", loading...
```

**Environment:**

- kind version: (use `kind version`): 0.12.0
- Kubernetes version: (use `kubectl version`): v1.23.4
- Docker version: (use `docker info`): 20.10.7
- OS (e.g. from `/etc/os-release`): Ubuntu 20.04.4 LTS

harshanarayana:
Currently, the way the `kind load docker-image` works is by detecting if there is an image with a matching image name and tab already exists in the cluster. However it is possible that the same image might be already present in the cluster with a different tag. 

With `ctr` 1.3.0+ supporting `ctr image tag` feature, we can avoid loading the images by just re-tagging the image hash with the new name.

Current changes fallback to using the default mode it currently has on following cases to retain compatibility.

1. If the `ctr` in the kind image is not supporting `ctr images tag` command
2. If the command to retrieve the RepoTags from Image ID runs into an error
3. If the Image ID itself is missing on the node
4. In case if the re-tag fails, it will fall back to loading the image all over again like the current workflow

Closes #2663 

### Missing Image
```bash
❯ bin/kind load docker-image --name helm-test docker.io/library/ubuntu:18.04
Image: "ubuntu:18.04" with ID "sha256:c6ad7e71ba7d4969784c76f57c4cc9083aa96bb969d802f2ea38f4aaed90ff93" not yet present on node "helm-test-control-plane", loading...
```
### Image Exists with Different Tag and Name
```bash
❯ bin/kind load docker-image --name helm-test docker.io/library/different-repo-ubuntu:18.04
Image with ID: sha256:c6ad7e71ba7d4969784c76f57c4cc9083aa96bb969d802f2ea38f4aaed90ff93 already present on the node helm-test-control-plane but is missing the tag different-repo-ubuntu:18.04. re-tagging...
```
### No image re-tag or load required
```bash
❯ bin/kind load docker-image --name helm-test docker.io/library/different-repo-ubuntu:18.04                                                                                                                                                                        
Image: "docker.io/library/different-repo-ubuntu:18.04" with ID "sha256:c6ad7e71ba7d4969784c76f57c4cc9083aa96bb969d802f2ea38f4aaed90ff93" found to be already present on all nodes.
```

### Sanitizing Registry Images
```bash
❯ kind load docker-image --name upstream-k8s kind/test-ubuntu:1.2.3
Image with ID: sha256:c6ad7e71ba7d4969784c76f57c4cc9083aa96bb969d802f2ea38f4aaed90ff93 already present on the node upstream-k8s-control-plane but is missing the tag kind/test-ubuntu:1.2.3. re-tagging...

root@upstream-k8s-control-plane:/# crictl images | grep ubun | grep kind
docker.io/kind/test-ubuntu                            1.2.3                                      c6ad7e71ba7d4       65.5MB
```

### DIFF
--- a/pkg/cluster/nodeutils/util.go
+++ b/pkg/cluster/nodeutils/util.go
@@ -126,3 +126,31 @@ func ImageID(n nodes.Node, image string) (string, error) {
 	}
 	return crictlOut.Status.ID, nil
 }
+
+// ImageTags is used to perform a reverse lookup of the ImageID to list set of available
+// RepoTags corresponding to the ImageID in question
+func ImageTags(n nodes.Node, imageID string) (map[string]bool, error) {
+	var out bytes.Buffer
+	tags := make(map[string]bool, 0)
+	if err := n.Command("crictl", "inspecti", imageID).SetStdout(&out).Run(); err != nil {
+		return tags, err
+	}
+	crictlOut := struct {
+		Status struct {
+			RepoTags []string `json:"repoTags"`
+		} `json:"status"`
+	}{}
+	if err := json.Unmarshal(out.Bytes(), &crictlOut); err != nil {
+		return tags, err
+	}
+	for _, tag := range crictlOut.Status.RepoTags {
+		tags[tag] = true
+	}
+	return tags, nil
+}
+
+// ReTagImage is used to tag an ImageID with a custom tag specified by imageName parameter
+func ReTagImage(n nodes.Node, imageID, imageName string) error {
+	var out bytes.Buffer
+	return n.Command("ctr", "--namespace=k8s.io", "images", "tag", "--force", imageID, imageName).SetStdout(&out).Run()
+}
--- a/pkg/cmd/kind/load/docker-image/docker-image.go
+++ b/pkg/cmd/kind/load/docker-image/docker-image.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/spf13/cobra"
 
@@ -35,6 +36,10 @@ import (
 
 	"sigs.k8s.io/kind/pkg/internal/cli"
 	"sigs.k8s.io/kind/pkg/internal/runtime"
+)
+
+type (
+	imageTagFetcher func(nodes.Node, string) (map[string]bool, error)
 )
 
 type flagpole struct {
@@ -128,16 +133,34 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
 	fns := []func() error{}
 	for i, imageName := range imageNames {
 		imageID := imageIDs[i]
+		processed := false
 		for _, node := range candidateNodes {
+			exists, reTagRequired := checkIfImageReTagRequired(node, imageID, imageName, nodeutils.ImageTags)
+			if exists && !reTagRequired {
+				continue
+			}
+
+			if reTagRequired {
+				// We will try to re-tag the image. If the re-tag fails, we will fall back to the default behavior of loading
+				// the images into the nodes again
+				logger.V(0).Infof("Image with ID: %s already present on the node %s but is missing the tag %s. re-tagging...", imageID, node.String(), imageName)
+				if err := nodeutils.ReTagImage(node, imageID, imageName); err != nil {
+					logger.Errorf("failed to re-tag image on the node %s due to an error %s. Will load it instead...", node.String(), err)
+					selectedNodes = append(selectedNodes, node)
+				} else {
+					processed = true
+				}
+				continue
+			}
 			id, err := nodeutils.ImageID(node, imageName)
 			if err != nil || id != imageID {
 				selectedNodes = append(selectedNodes, node)
 				logger.V(0).Infof("Image: %q with ID %q not yet present on node %q, loading...", imageName, imageID, node.String())
 			}
-		}
-		if len(selectedNodes) == 0 {
-			logger.V(0).Infof("Image: %q with ID %q found to be already present on all nodes.", imageName, imageID)
 			continue
+		}
+		if len(selectedNodes) == 0 && !processed {
+			logger.V(0).Infof("Image: %q with ID %q found to be already present on all nodes.", imageName, imageID)
 		}
 	}
 
@@ -214,4 +237,47 @@ func removeDuplicates(slice []string) []string {
 		}
 	}
 	return result
+}
+
+// checkIfImageExists makes sure we only perform the reverse lookup of the ImageID to tag map
+func checkIfImageReTagRequired(node nodes.Node, imageID, imageName string, tagFetcher imageTagFetcher) (exists, reTagRequired bool) {
+	tags, err := tagFetcher(node, imageID)
+	if len(tags) == 0 || err != nil {
+		exists = false
+		return
+	}
+	exists = true
+	imageName = sanitizeImage(imageName)
+	if ok := tags[imageName]; ok {
+		reTagRequired = false
+		return
+	}
+	reTagRequired = true
+	return
+}
+
+// sanitizeImage is a helper to return human readable image name
+// This is a modified version of the same function found under providers/podman/images.go
+func sanitizeImage(image string) (sanitizedName string) {
+	const (
+		defaultDomain    = "docker.io/"
+		officialRepoName = "library"
+	)
+	sanitizedName = image
+
+	if !strings.ContainsRune(image, '/') {
+		sanitizedName = officialRepoName + "/" + image
+	}
+
+	i := strings.IndexRune(sanitizedName, '/')
+	if i == -1 || (!strings.ContainsAny(sanitizedName[:i], ".:") && sanitizedName[:i] != "localhost") {
+		sanitizedName = defaultDomain + sanitizedName
+	}
+
+	i = strings.IndexRune(sanitizedName, ':')
+	if i == -1 {
+		sanitizedName += ":latest"
+	}
+
+	return
 }
--- a/pkg/cmd/kind/load/docker-image/docker-image_test.go
+++ b/pkg/cmd/kind/load/docker-image/docker-image_test.go
@@ -17,9 +17,12 @@ limitations under the License.
 package load
 
 import (
+	"errors"
 	"reflect"
 	"sort"
 	"testing"
+
+	"sigs.k8s.io/kind/pkg/cluster/nodes"
 )
 
 func Test_removeDuplicates(t *testing.T) {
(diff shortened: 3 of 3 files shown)


BenTheElder:
Thanks for the PR 😄 

> With ctr 1.3.0+ supporting ctr image tag feature, we can avoid loading the images by just re-tagging the image hash with the new name.

1.3.0 is older than I thought, we've had 1.3.0+ for a very long time, let's drop the check.


> In case if the re-tag fails, it will fall back to loading the image all over again like the current workflow

Trying to imagine a scenario where this makes sense 🤔
What were you thinking would happen in this case?

> Containerd internally tracks the docker images with docker.io/library which can be missing on the local image cache. In such cases, the detection of duplicate image just re-tags the image again. This is only when the dockerhub images are loaded without docker.io/library prefix.

This is going to be common, we should handle this translation. 
Kubernetes also does this internally, it is carried over from docker.

We need to take care to remember that while containerd does this, in podman on the host side the default registry is actually configurable, so we should only perform this normalization when interacting with containerd.

harshanarayana:
@BenTheElder  Thanks for the quick review and feedbacks. 

> let's drop the check.

All righty. Let me take care of that.

> What were you thinking would happen in this case?

Thinking about worst or the worst cases such as containerd getting restarted causing re-tag to fail or some other docker issue causing the command invocation on the container to fail. Just a precaution. Instead of returning an error, I am doing the next best thing of reloading again. Or do you think it is enough to return an error? Since we use the `--force` argument with `ctr` if the tag already exists, it will handle it and not return an error. So other than the worst case I mentioned above, I can't think of any other reason why `ctr images tag` would fail. 

> This is going to be common, we should handle this translation.

Sure. Let me take care of it. So, all we need to do is when the ID to repoTag match is being done, in the image if the registry part is missing, we default to `docker.io/library` and process it. Right ? Or do you see any other special handling to be done ?

harshanarayana:
> This is going to be common, we should handle this translation.

Refactored a bit of code that `podman` provider had into a common utils and using that in both places now. Relocated the tests as well to address the changes accordingly.

harshanarayana:
@BenTheElder PTAL when you can. Changes have been addressed

BenTheElder:
Sorry, I've been out and mostly unavailable.

pkg/cluster/nodeutils/util.go
```diff
@@ -126,3 +126,64 @@ func ImageID(n nodes.Node, image string) (string, error) {
...
+
+// ReTagImage is used to tag an ImageID with a custom tag specified by imageName parameter
+func ReTagImage(n nodes.Node, imageID, imageName string) error {
+	var out bytes.Buffer
+	return n.Command("ctr", "--namespace=k8s.io", "images", "tag", "--force", imageID, imageName).SetStdout(&out).Run()
+}
+
+// sanitizeImage is a helper to return human readable image name and
+// the podman pullable image name from the provided image
+func SanitizeImage(image string) (friendlyImageName, pullImageName string) {
```
BenTheElder, line 160: This one probably shouldn't be made public, and probably shouldn't be shared between podman and containerd, even if the method _currently_ does what we need for both.

What we're _currently_ doing for images passed to podman we should probably not do for all images and consider not doing in the future, and just make people actually specify docker.io when they want to use kind images from there.

For containerd, you could imagine this behavior could also become configurable in the future there.

TDLR: WET > DRY
Write Everything Twice. We'll know it's a good abstraction or not by the third time :-)

BenTheElder: we need a much simpler version of this for containerd loading, which doesn't care about the sha part etc., and will be easier to maintain someday.

podman and docker already have different versions of producing the "friendly" image for printing, which was also possibly a mis-feature (as it obscures what exactly we're running ...)

BenTheElder (commented):
one major nit, thanks for working on this
cc @aojea to help keep after reviews

pkg/cluster/nodeutils/util.go
harshanarayana: @BenTheElder All right. Let me put the refactoring back to the way it was and create a much simpler version of the function for sanitizing the image name for our current use-case. I will update the changes in a while

harshanarayana:
@BenTheElder  @aojea  PTAL at this PR when time permits. I've updated the changes as suggested by @BenTheElder

harshanarayana:
@BenTheElder  @aojea Just a gentle reminder if time permits to take a look at this one.

pkg/cmd/kind/load/docker-image/docker-image_test.go
```diff
@@ -60,3 +60,44 @@ func Test_removeDuplicates(t *testing.T) {
...
+			image:          "registry.k8s.io/kindest/node:latest",
+			sanitizedImage: "registry.k8s.io/kindest/node:latest",
+		},
+		{
+			image:          "k8s.gcr.io/pause:3.6",
+			sanitizedImage: "k8s.gcr.io/pause:3.6",
+		},
+		{
+			image:          "k8s.gcr.io/pause:3.6",
+			sanitizedImage: "k8s.gcr.io/pause:3.6",
```
aojea, line 88: these are the same, aren't they?

pkg/cmd/kind/load/docker-image/docker-image.go
```diff
@@ -215,3 +233,47 @@ func removeDuplicates(slice []string) []string {
...
+	if ok := tags[imageName]; ok {
+		exists = true
+		return
+	}
+	exists = true
+	reTagRequired = true
+	return
+}
+
+func sanitizeImage(image string) (sanitizedName string) {
```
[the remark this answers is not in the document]
aojea: oh, forget it , this is copied from
https://github.com/kubernetes-sigs/kind/blob/9ec6695660e203dc220581b66b62247b32e42018/pkg/cluster/internal/providers/podman/images.go#L84-L86

add a comment to the function then, to indicate the origin

pkg/cmd/kind/load/docker-image/docker-image_test.go
harshanarayana: Ack. Fixed.

pkg/cmd/kind/load/docker-image/docker-image.go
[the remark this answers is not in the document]
harshanarayana: Ack. Took care of it in the update.

pkg/cluster/nodeutils/util.go
```diff
@@ -126,3 +126,41 @@ func ImageID(n nodes.Node, image string) (string, error) {
...
+	}
+	for _, tag := range crictlOut.Status.RepoTags {
+		tags[tag] = true
+	}
+	return tags, nil
+}
+
+// ImageReTagSupported checks if the version of the ctr running in the cluster has support
+// for performing ctr images tag command
+func ImageReTagSupported(n nodes.Node) bool {
```
BenTheElder, line 154: I thought we were dropping this?

harshanarayana: Sorry, My bad. Yes we did. Forgot to cleanup. Let me take care of it.

pkg/cmd/kind/load/docker-image/docker-image.go
```diff
@@ -128,16 +133,33 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
...
+				// This could be because of one of two reasons. Either the previous command ran into an error or no image
+				// with matching Image ID was found. Let us try again and see if the image is indeed required to be loaded
+				id, err := nodeutils.ImageID(node, imageName)
+				if err != nil || id != imageID {
+					selectedNodes = append(selectedNodes, node)
+					logger.V(0).Infof("Image: %q with ID %q not yet present on node %q, loading...", imageName, imageID, node.String())
+				}
+				continue
+			}
+			if reTagRequired {
```
BenTheElder, line 149: go encourages only indenting the exceptional cases, I think it would be more idomatic and readable do to more like:
```go
if exists {
    continue
}
if reTagRequired {
   // do retagging instead of loading
    ...
    continue
}
// do loading
...
```

pkg/cmd/kind/load/docker-image/docker-image.go
```diff
@@ -37,6 +38,10 @@ import (
 	"sigs.k8s.io/kind/pkg/internal/runtime"
 )
 
+var (
+	imageTagFetcher = nodeutils.ImageTags
```
BenTheElder, line 42: instead of a global, we can pass the function, or split the function under test into two and pass the results of calling this.

harshanarayana: Let me refactor is to pass the function across and use that instead. Would make it cleaner.

pkg/cmd/kind/load/docker-image/docker-image.go
harshanarayana: Ack. Let me refactor this.

harshanarayana: I am going to refactor the code a bit and instead of returning two booleans, going to return an `ENUM` like entity that we can use to compare and make a decision. It will also keep the conditional checks cleaner.  Do you see any issue with this @BenTheElder 

```golang
func checkIfImageReTagRequired(node nodes.Node, imageID, imageName string, tagFetcher imageTagFetcher) processMode {
	tags, err := tagFetcher(node, imageID)
	if len(tags) == 0 || err != nil {
		return LoadImage
	}

	imageName = sanitizeImage(imageName)
	if ok := tags[imageName]; ok {
		return NoOp
	}
	return ReTagImage
}

-----
if imageProcessMode == NoOp {
    continue
}

if imageProcessMode == ReTagImage {
    // do stuff
    continue
}
// load Image
```

harshanarayana:
```bash
❯ ./kind load docker-image --name test-cluster ubuntu:18.04
Image: "ubuntu:18.04" with ID "sha256:c6ad7e71ba7d4969784c76f57c4cc9083aa96bb969d802f2ea38f4aaed90ff93" not yet present on node "test-cluster-control-plane", loading...

❯ ./kind load docker-image --name test-cluster kind/test-ubuntu:1.2.3
Image with ID: sha256:c6ad7e71ba7d4969784c76f57c4cc9083aa96bb969d802f2ea38f4aaed90ff93 already present on the node test-cluster-control-plane but is missing the tag kind/test-ubuntu:1.2.3. re-tagging...

❯ ./kind load docker-image --name test-cluster different-repo-ubuntu
ERROR: image: "different-repo-ubuntu" not present locally

❯ ./kind load docker-image --name test-cluster different-repo-ubuntu:18.04
Image with ID: sha256:c6ad7e71ba7d4969784c76f57c4cc9083aa96bb969d802f2ea38f4aaed90ff93 already present on the node test-cluster-control-plane but is missing the tag different-repo-ubuntu:18.04. re-tagging...

File: pkg/cmd/kind/load/docker-image/docker-image.go (hunk 5 above)
Comment (BenTheElder) [reply pkg/cmd/kind/load/docker-image/docker-image.go]: I wouldn't recommend it here because go enums are weak and we wind up with an unhandled case for additional values we didn't consider.

The booleans are reasonably idiomatic and are clearly exhaustively handled.

Comment (harshanarayana) [reply pkg/cmd/kind/load/docker-image/docker-image.go]: > I wouldn't recommend it here because go enums are weak

I think i kind of mentioned it wrong.  I meant to say a custom type. Not enum,. 

```golang
type (
	processMode     string
)

const (
	LoadImage  processMode = "load"
	ReTagImage processMode = "re-tag"
	NoOp       processMode = "none"
)
```

BenTheElder: I don't think this is an idiomatic approach in go, versus two return values, or inlining the method

BenTheElder: the problem is, even though there are only 3 states, I need to go read this other type and see what values exist to understand if they're all being handled.

With two booleans with local variable names, I can quickly comprehend if it's being handled.

In the future if a fourth mode is added to this type for some reason, subtle bugs will creep in because we're still using a very weak "enum" and it's not obvious if they're all being handled where they need to be, versus I know the booleans only have two values to handle.

BenTheElder:
Thank you for working on this, very sorry I've been so unavailable at the moment and not getting back to this reasonably.

I will be on vacation for a few weeks and then my work situation will change a bit. antonio can hopefully pick up remaining review / bikeshed :-)

pkg/cmd/kind/load/docker-image/docker-image.go
harshanarayana: @BenTheElder  All right. Let me move it back to the two bool return format and add checks accordingly.

harshanarayana:
cc @aojea  @BenTheElder  PTAL at the changes when time permits.

harshanarayana:
@aojea @BenTheElder  PTAL when time permits. The suggested changes have been addressed.

BenTheElder (approved):
/lgtm
/approve

PR #2787, opened 2022-05-27
Update yaml library for CVE

BenTheElder:
Fixes CVE-2022-28948

> An issue in the Unmarshal function in Go-Yaml v3 causes the program to crash when attempting to deserialize invalid input.

Not a big deal in our tools module, and users crashing themselves is a bit 🤷‍♂️, but we should be patching anyhow.


pkg/cluster/internal/providers/docker/network_test.go
```diff
@@ -154,9 +154,7 @@ func Test_sortNetworkInspectEntries(t *testing.T) {
 		tc := tc
 		t.Run(tc.Name, func(t *testing.T) {
 			toSort := make([]networkInspectEntry, len(tc.Networks))
-			for i := range tc.Networks {
-				toSort[i] = tc.Networks[i]
-			}
+			copy(toSort, tc.Networks)
```
BenTheElder, line 157: TODO: As discussed in commit message, we should consider disabling this linter instead, this is an inaccurate error and you might introduce a bug by switching.
We're fine in this case though.

PR #2788, opened 2022-05-31
Add known issue "Docker Desktop for macOS and Windows"

cavokz:


BenTheElder:
Thanks for the PR!

We should consider how to reconcile with https://kind.sigs.k8s.io/docs/user/loadbalancer/

cavokz:
> Thanks for the PR!

My pleasure.

> We should consider how to reconcile with https://kind.sigs.k8s.io/docs/user/loadbalancer/

In what sense?

BenTheElder:
There is a fairly detailed note there of the same nature, we should consider if some of that should just be cross referencing this new entry and if any of the text / links should be reused or rewritten to match in either location.

aojea:
/tide squash-merge
/lgtm
/approve

Thanks

PR #2789, opened 2022-06-01
Add repo LICENSE file to base image

alexmasi:
I would like to add the base repo LICENSE file to the base image.

Because the docker build env does not include the parent dir where the LICENSE file is I am using ADD to fetch from the URL.

Resulting image:

```
$ docker run -it --entrypoint sh 27ee7c05fa24
# ls
LICENSE  bin  boot  dev  etc  home  lib  lib32	lib64  libx32  media  mnt  opt	proc  root  run  sbin  srv  sys  tmp  usr  var
# cat LICENSE	
                                 Apache License
                           Version 2.0, January 2004
                        http://www.apache.org/licenses/

   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION

   1. Definitions.

      "License" shall mean the terms and conditions for use, reproduction,
      and distribution as defined by Sections 1 through 9 of this document.
...
```

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -79,6 +79,9 @@ COPY --chmod=0644 files/etc/sysctl.d/* /etc/sysctl.d/
 COPY --chmod=0644 files/etc/systemd/system/* /etc/systemd/system/
 COPY --chmod=0644 files/etc/systemd/system/kubelet.service.d/* /etc/systemd/system/kubelet.service.d/
 
+# copy the repo LICENSE file
+ADD https://raw.githubusercontent.com/kubernetes-sigs/kind/main/LICENSE /
+
 # Install dependencies, first from apt, then from release tarballs.
 # NOTE: we use one RUN to minimize layers.
 #


BenTheElder:
Thanks for the PR but I don't think this change is appropriate.

Kubernetes does not do this for any other images.

The repo's license does NOT apply to all files in the image, there are many things from other sources. This file would be misleading in my opinion.

Downloading from the repo via GitHub like this also isn't a great way to accomplish this.

alexmasi:
Thanks for taking a look Ben, my main goal here is to surface the appropriate licenses in the node container images so it can be re-hosted for use in KNE (github.com/google/kne) in order for us to reference tested versions in our deployment setups.

We have had issues in the past (not with kind but with other container deps) where we are referencing a container by a tag in a manifest or deployment config and that tag moves or the version is deleted leading to errors for the user.

By rehosting we can control the tags and ensure the containers work with KNE

But given that kind hasn't given us any issues outlined above (and adding license info is non-trivial) we probably won't have to rehost this at all.

PR #2801, opened 2022-06-14
fix blank line in rootless.md

3deep5me:
I think this makes more sense 🤗

### DIFF
--- a/site/content/docs/user/rootless.md
+++ b/site/content/docs/user/rootless.md
@@ -56,7 +56,8 @@ $ kind create cluster
 
 To create a kind cluster with Rootless Podman, just run:
 ```console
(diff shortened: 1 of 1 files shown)


site/content/docs/user/rootless.md
```diff
@@ -56,7 +56,8 @@ $ kind create cluster
 
 To create a kind cluster with Rootless Podman, just run:
 ```console
-$ KIND_EXPERIMENTAL_PROVIDER=podman kind create cluster
```
stmcginnis, line 59: I think we want to keep it as is.

aojea: agree with Sean

PR #2802, opened 2022-06-14
clarify the impact of Delegate=true when using rootless

### LINKED ISSUE #2800: Rootless docs do not explain the security aspects of Delegate=yes

https://kind.sigs.k8s.io/docs/user/rootless/ notes that on some systems, you may need to set `Delegate=true` in the systemd configuration. However, it completely fails to explain why this is necessary, or more precisely, why systemd is not configured this way by default, and what the potential downsides of changing the configuration are.

**Why is this needed**:

Well, it immediately reminded me of the longstanding recommendation to configure docker so that you could run the `docker` command as any user, which no one ever bothered to mention was equivalent to giving passwordless `sudo` access to every user... makes me wonder what other things I can do if I make this `Delegate=true` change...

danwinship:
The docs tell you to enable a mysterious systemd option when using rootless, but don't tell you why. I was assuming it was a security thing but it turns out it's actually just performance-related.

(The reindentation of the code blocks is because I couldn't figure out how to make the new paragraph be aligned under the previous bullet point without also making the code block be aligned under the bullet point.)

Fixes #2800 

/cc @aojea @AkihiroSuda

### DIFF
--- a/site/content/docs/user/rootless.md
+++ b/site/content/docs/user/rootless.md
@@ -22,18 +22,26 @@ running `sudo update-grub`.
 Also, depending on the host configuration, the following steps might be needed:
 
 - Create `/etc/systemd/system/user@.service.d/delegate.conf` with the following content, and then run `sudo systemctl daemon-reload`:
-```ini
-[Service]
-Delegate=yes
-```
+
+  ```ini
+  [Service]
+  Delegate=yes
+  ```
+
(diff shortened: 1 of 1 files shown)


aojea:
/approve
defer lgtm to @AkihiroSuda

AkihiroSuda (approved):

PR #2804, opened 2022-06-16
Add node build steps to Building Images section

stmcginnis:
In our quick start guide we have a Building Images section that talks
about being able to build custom images. It describes some of the
pieces, but it doesn't actually have the steps to take to actually
create the image.

There are notes in the repo README.md file that include these steps.
This adds a small amount of duplication, but this adds those steps to
the Building Images section to make it easier for folks to find out what
steps to take to build their images.

### DIFF
--- a/site/content/docs/user/quick-start.md
+++ b/site/content/docs/user/quick-start.md
@@ -260,6 +260,16 @@ Currently, kind supports one default way to build a `node-image`
 if you have the [Kubernetes][kubernetes] source in your host machine
 (`$GOPATH/src/k8s.io/kubernetes`), by using `docker`.
 
+To create a cluster from Kubernetes source:
+
+- ensure that Kubernetes is cloned in `$(go env GOPATH)/src/k8s.io/kubernetes`
+- build a node image and create a cluster with:
+
+```console
+kind build node-image
+kind create cluster --image kindest/node:latest
+```
+
 > **NOTE**: Building Kubernetes node-images requires everything building upstream
 > Kubernetes requires, we wrap the upstream build. This includes Docker with buildx.
 > See: https://git.k8s.io/community/contributors/devel/development.md#building-kubernetes-with-docker


site/content/docs/user/quick-start.md
```diff
@@ -260,6 +260,16 @@ Currently, kind supports one default way to build a `node-image`
 if you have the [Kubernetes][kubernetes] source in your host machine
 (`$GOPATH/src/k8s.io/kubernetes`), by using `docker`.
 
+To create a cluster from Kubernetes source:
+
+- ensure that Kubernetes is cloned in `$(go env GOPATH)/src/k8s.io/kubernetes`
```
BenTheElder, line 265: noting: this should also be covered in the **NOTE** below, and it may change in the future if Kubernetes stops depending on GOPATH (which is considered an outstanding bug)

stmcginnis: The duplication is a concern, but I think the README file is harder to discover than the official docs. Maybe the readme should point here instead?

When you say "this should be covered in the NOTE below", do you mean the `NOTE: Building Kubernetes node-images requires...` section? This should be inside that NOTE block? Or the link in the NOTE block should lead to the correct commands? If the former, happy to move this if that is preferred. If that latter, then I think it's still not very clear and warrants duplication here to make it easier for folks to find.

This was prompted after the question coming up a few times in Slack and then after pointing the person to this section in the docs, realizing that it has a lot of information to digest but doesn't actually make it clear what needs to be run to actually produce an image. Basically, I think it would be good to make it easier to discover how to do this so there is less confusion and less of a need for the question to keep being asked in Slack.

If the preference is to not make it so easy in order to force anyone doing this to actually read the docs more thoroughly, then I suppose that makes sense. Just let me know and I can close this out if so.

BenTheElder: sorry, this was overly terse, and I've been at an even this week and preparing to go OOO for a while now ...

I meant that the note block below currently leads to the docs for building kubernetes.
The correct requirements for building Kubernetes currently require GOPATH.

So the note below should more expansively cover the correct clone location by way of the upstream docs.

I would prefer not to cover kubernetes build requirements in depth in two places, if we have gaps in the upstream docs we should fill them there.

If the concern is that there's one or two tripping points people miss, I think it's OK to cover here, but I'd prefer to keep that to a minimum and not wind up duplicating the document because the upstream document isn't working for some reason.

The further point is: when kubernetes builds stop requiring GOPATH, so will kind node image builds. Duplicating build requirements gives more room to miss updating them since many folks updating Kubernetes docs probably won't look here.

BenTheElder: I don't think we want to _force_ reading docs more thoroughly, but ideally:
- we effectively call out the need to see the upstream doc for build requirements
- the upstream docs should be clear and easily read
- we should prefer to improve upstream docs where possible

The kubernetes website repo actually has a hard requirement on not duplicating third party content in their style / review guidelines, so for example they've purged all guides to use kind and instead just link to us.

But, I also understand that many kind users are not Kubernetes developers and are perhaps even very new to development and minimizing the reading required is desirable.

On the other hand, this is only _one_ of the requirements to get Kubernetes to build.
IIRC macOS developers on new machines need to install a few things with homebrew like gnu coreutils and an updated bash shell before the build will succeed (besides e.g. Docker).

BenTheElder (commented):
but, also I just want us to be thoughtful on those points, I'm sorry for the review delay and I'm not trying to be a pain here, I appreciate the contribution.

/approve

@aojea if you can take a look, I'm preemptively approving given the comment that this comes up previously a few times in slack

BenTheElder:
Sorry for even more text to read through, ordinarily I'd whittle this down but I'm pressed for time, one last thought ...

FWIW if we do cover this, I think we're going to need to consider expanding a lot more on GOPATH.

If you compare the install guide docs and how much nuance we've wound up covering ourselves there on using go to install we have precedent for covering content that really ought to not have to be duplicated everywhere, but we also wound up doing so more thoroughly than a single shell one-liner 🙃

stmcginnis:
> but, also I just want us to be thoughtful on those points, I'm sorry for the review delay and I'm not trying to be a pain here

Not at all, and I appreciate thinking this through thoroughly. It's a tricky balance to have just enough breadcrumbs here for someone to be able to quickly figure out what to do, versus duplicating too much information that doesn't belong.

I agree the proposed changes still probably don't strike that balance in an ideal way, but hoping it's a low-cost improvement for now until we can find a better way to guide those new users through what is really a much more detailed process.

But to be clear, I am absolutely fine dropping this if the feeling is we should use a different approach to these docs.

stmcginnis:
Cleaning up.

/close

PR #2825, opened 2022-07-05
Fix KubeletConfiguration template for k8s < 1.15

a-palchikov:
Fix the KubeletConfiguration template for < 1.15 w.r.t feature gates - otherwise the cluster construction fails with:
```
Creating cluster "dev.test" ...
 ✓ Ensuring node image (knode:v1.14) 🖼 
 ✓ Preparing nodes 📦  
 ✗ Writing configuration 📜 
ERROR: failed to create cluster: failed to generate kubeadm config content: error executing config template: template: kubeadm-config:105:19: executing "kubeadm-config" at <$.FeatureGates>: FeatureGates has arguments but cannot be invoked as function
make: *** [cluster] Error 1
```

### DIFF
--- a/pkg/cluster/internal/kubeadm/config.go
+++ b/pkg/cluster/internal/kubeadm/config.go
@@ -266,7 +266,7 @@ evictionHard:
   imagefs.available: "0%"
 {{if .FeatureGates}}featureGates:
 {{ range $key := .SortedFeatureGateKeys }}
-  "{{ $key }}": {{$.FeatureGates $key }}
+  "{{ $key }}": {{index $.FeatureGates $key }}
 {{end}}{{end}}
 {{if ne .KubeProxyMode "None"}}
 ---


BenTheElder:
A formal version support policy is something I drafted earlier this year but didn't get our yet.

Kubernetes as a project only supports back to 1.22: https://kubernetes.io/releases/

Kind has been more generous, but we can't realistically keep running arbitrarily large amounts of builds and CI etc.

Since the most recent release we've soft-switched to doubling Kubernetes here: we also support the latest development sources when using kind's latest sources and we support 3 more older versions (currently back to 1.18).

To an extent even this isn't very viable, Kubernetes will not accept patches to these unsupported versions and we cannot really host and will not use forks as a subproject, so we're stuck without any bug fixes when we try to support older versions.


Despite the simple patch here I'm somewhat inclined to say we should not be trying to support Kubernetes this old, as it also limits us from depending on kubeadm improvements in particular.

I'm doing a poor job of it but on vacation until late July. My co-maintainer Antonio is also out at the moment but should be back sooner and can also weigh in on this.

a-palchikov:
> I'm doing a poor job of it but on vacation until late July. My co-maintainer Antonio is also out at the moment but should be back sooner and can also weigh in on this.

No problem.

> Despite the simple patch here I'm somewhat inclined to say we should not be trying to support Kubernetes this old, as it also limits us from depending on kubeadm improvements in particular.
> 

I agree about the supporting older versions this old but the code is currently broken as-is. If it's broken the options are:
  1. it is patched to continue working (subject to change complexity, I guess)
  1. the code that is still in place in support for k8s < 1.15 should be removed entirely

Would you say the option 2 is the way to go with here?

aojea:
> Would you say the option 2 is the way to go with here?

this falls on the project bucket of "generousity"  Ben is talking, no need to remove the code so people can use it ;), but not bugfixing

> Kind has been more generous, but we can't realistically keep running arbitrarily large amounts of builds and CI etc.

BenTheElder (approved):
/lgtm
/approve
sorry for the long delay here

I still think we should drop these old Kubernetes versions at this point, but since we haven't yet and the patch is tiny, I'll merge this fix first, so it's possible to grab a recent commit with things working.

BenTheElder:
so yes, we should ultimately do 2) but I think there's value in having a moment where we did 1), maybe we will do 2) for v0.16 since we're about ready to release v0.15 as-is.

BenTheElder:
thanks!

PR #2831, opened 2022-07-19
Add field tags for JSON (un)marshalling

jkremser:
This makes the API types reusable across the k8s ecosystem. I've been working on cluster-api provider for kind and wanted to reuse as much configuration as possible in the CRD. if this pr is merged I could directly reference the kind upstream types, instead of introducing my own, converting them back and forth and making sure the contract is still the same.

Signed-off-by: Jirka Kremser <jiri.kremser@gmail.com>

### DIFF
--- a/pkg/apis/config/v1alpha4/types.go
+++ b/pkg/apis/config/v1alpha4/types.go
@@ -18,33 +18,33 @@ package v1alpha4
 
 // Cluster contains kind cluster configuration
 type Cluster struct {
-	TypeMeta `yaml:",inline"`
+	TypeMeta `yaml:",inline" json:",inline"`
 
 	// The cluster name.
 	// Optional, this will be overridden by --name / KIND_CLUSTER_NAME
-	Name string `yaml:"name,omitempty"`
+	Name string `yaml:"name,omitempty" json:"name,omitempty"`
 
 	// Nodes contains the list of nodes defined in the `kind` Cluster
 	// If unset this will default to a single control-plane node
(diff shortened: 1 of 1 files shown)


BenTheElder:
We have had custom un-marshaling with the YAML library in the past (can’t recall if we do currently) and it was much better to work with.

These types are component config but are not CRDs.

> I've been working on cluster-api provider for kind and wanted to reuse as much configuration as possible in the CRD

Are you aware of the upstream “docker” provider used by CAPI for development that is KIND based? This might be duplicated effort.

jkremser:
> Are you aware of the upstream “docker” provider used by CAPI for development that is KIND based? This might be duplicated effort.

Yes I know about it, I am doing it as a mental exercise to learn the CAPI :D

> These types are component config but are not CRDs.

Right, but it would be nice if they are reusable. I mean I can maintain my own copy, but hey "duplicated effort"

BenTheElder:
Sorry, more available now, been on vacation.

I recall now that we did away with special unmarshalling for the fields in favor of 1:1 disk format in memory already (we had stringly named enums => ints, now they're just strings), so we should be fine in that regard.

That was my only concern -- that this was something we could ensure continued to work as expected without issue, we should be fine as long as we stick to simple types like we have currently.

> Yes I know about it, I am doing it as a mental exercise to learn the CAPI :D

Cool :-)


/ok-to-test
/lgtm
/approve

jkremser:
thanks for the merge, btw. I am also back from the vacation :D

PR #2836, opened 2022-07-20
enable progress notification from etcd for all k8s e2e tests

p0lyn0mial:
the experimental-watch-progress-notify-interval flag specifies an interval
at which etcd sends data to the kube-api server.

It is used by the WatchBookmark feature which is GA since 1.17.
In addition to that, it will be used by a new WatchList feature which is Alpha since 1.25

Enables support for watch bookmark events.

xref: https://github.com/kubernetes/enhancements/issues/3157
xref: https://github.com/kubernetes/enhancements/pull/3142

### DIFF
--- a/hack/ci/e2e-k8s.sh
+++ b/hack/ci/e2e-k8s.sh
@@ -168,6 +168,10 @@ ${controllerManager_extra_args}
   scheduler:
     extraArgs:
 ${scheduler_extra_args}
+   etcd:
+     local:
+       extraArgs:
+         experimental-watch-progress-notify-interval: "5s"
   ---
   kind: InitConfiguration
   nodeRegistration:


neolit123 (commented):
kind supports a wide range of kubeadm / k8s versions. when was the experimental flag added in etcd and what is it's graduation timeline? given this is a kind pr do we want to enable it before it graduates for early test signal?

this would also mean that once the flag is no longer in etcd (experimental flag graduated to a new flag without the prefix) kind would need branching to aplly different flags for different etcd versions.

it would be easier for kind if kubeadm manages the flags. but we prefer to not add experimental features unless there are bugs at play - e.g. the member sync problem from a few months back.

on a relavent note, while the --experimental.. prefix pattern provides clear indication to users that something is experimental it just add usage complexity and instead documentation and cli stderr can show a warning to users.

aojea:
/hold

experiments should be done as experiments :smile: 

There are some examples in the test-infra folder, use a custom job with a customized script to test experimental options

https://github.com/kubernetes/test-infra/blob/master/experiment/kind-multizone-e2e.sh

This PR is a good example

https://github.com/kubernetes/test-infra/pull/25606/

BenTheElder:
Generally we prefer to change defaults in kubeadm and CI / kind overrides are for minor workarounds we can’t really avoid like dealing with DNS tests in an IPv6 CI environment without external IPv6 connectivity.

There are also specific CI jobs overriding and testing non-default things like alpha features but that doesn’t seem like the intent here.

It seems like this flag should be graduated to non-experimental and set in kubeadm?

BenTheElder:
@neolit123 naming aside (I agree FWIW) this does seem like something we’d want to enable broadly to improve efficiency and we know it is used in a ton of production clusters.

aojea:
> So I don't consider this an experimental feature. The problem is that there weren't any etcd release after 3.4 for ~2 years, and 3.5 that was released last year is still not fully production ready (not qualified enough).

Fair enough

/hold cancel

neolit123:
> @neolit123 naming aside (I agree FWIW) this does seem like something we’d want to enable broadly to improve efficiency and we know it is used in a ton of production clusters.

i have not seen metrics if this particular flag is used already in a lot of clusters. if the recommendation from the k8s / etcd integration side (api-machinery i guess) is to enable this flag in kubeadm we can do it.
we can handle the flag graduation on the kubeadm side and kind can consume it.

BenTheElder:
> FWIW - we enabled that setting in GKE (together with etcd 3.4) quite some time ago

That’s the lots of production clusters bit, GKE runs a lot of clusters and doesn’t enable e.g. kubernetes alpha features. So we have signal that it works.


I trust the SIG Scalability to tune etcd properly, I just generally prefer to minimize the CI overrides. We should consider doing this in the config generation in go instead of in the CI script until it lands in kubeadm.

wojtek-t:
> That’s the lots of production clusters bit, GKE runs a lot of clusters and doesn’t enable e.g. kubernetes alpha features. So we have signal that it works.

I can't give exact number, but we have a loooot of clusters having that flag. Also, all k8s CI that is using k8s-on-gce is using that flag. So I'm pretty convinced this is safe.

Now, I don't know kind so I will let you all to figure out how exactly this should be added.

p0lyn0mial:
Please let me know what are the next steps. Thanks!

aojea:
/ok-to-test
defer to @BenTheElder to get it running on the CI
@neolit123 for kubeadm maybe we should create a new issue to track it?

aojea:
it seems the patch is not well formatted

> ERROR: failed to create cluster: failed to generate kubeadm config content: failed to parse patches: failed to parse type meta for "kind: ClusterConfiguration\nmetadata:\n  name: config\napiServer:\n  extraArgs:\n    \"v\": \"4\"\ncontrollerManager:\n  extraArgs:\n    \"v\": \"4\"\nscheduler:\n  extraArgs:\n    \"v\": \"4\"\n etcd:\n   local:\n     extraArgs:\n       experimental-watch-progress-notify-interval: \"5s\"": error converting YAML to JSON: yaml: line 12: did not find expected key

neolit123:
> @neolit123 for kubeadm maybe we should create a new issue to track it?

yes. @p0lyn0mial please open an issue in k/kubeadm with a brief summary why we are adding the flag in kubeadm. you can then pr k/k/cmd/kubeadm.

we will use the issue to track the flag graduation.

p0lyn0mial:
> yes. @p0lyn0mial please open an issue in k/kubeadm with a brief summary why we are adding the flag in kubeadm. you can then pr k/k/cmd/kubeadm.

@neolit123 done, ptal: https://github.com/kubernetes/kubernetes/pull/111383 and https://github.com/kubernetes/kubernetes/issues/111382

PR #2869, opened 2022-08-08
bump to containerd 1.6.7, go 1.19

BenTheElder:
... and node image to 1.24.3


hack/tools/.golangci.yml
```diff
@@ -23,3 +23,9 @@ linters:
     - unparam
 
 linters-settings:
+  staticcheck:
+    checks:
+    - all
+    # TODO: require go1.16 for kind and sort out migrating without breakage
+    # until then, we still use io/ioutil for now
+    - '-SA1019'
```
BenTheElder, line 31: ioutil.ReadDir doesn't have a clean replacement

In the future we'll have to update go.mod to require go 1.16 and revive https://github.com/kubernetes-sigs/kind/pull/2443

Requiring go1.16 is not going to be ideal, because for example debian still appears to ship 1.15 in stable so we're going to break "go get" users.

We _are_ going to need to move off of io/ioutil at some point though.

aojea: deprecated is only a problem if we move to golang 2, no?

BenTheElder: I don't think these packages are actively maintained and the go standard library has moved off of them.

Amusingly `go`'s binary commands have not, because go itself needs to be bootstrapped by go 1.4 still (the first version of go written in go), but they're discussing moving bootstrap to 1.16.

BenTheElder: Also, due to them now actually being marked deprecated (in 1.19?) vscode started showing warnings, so I'm sure we'll have PRs coming for this again anyhow, we probably might as well figure out how to do it safely and efficiently without breaking `Copy` behavior etc.

BenTheElder:
flake https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/2869/pull-kind-e2e-kubernetes/1556765274663817216

aojea:
/retest
timeout

Issue #2868, opened 2022-08-08
[zfs] kind create cluster using podman provider fails to bootstrap the cluster - failed to mount rootfs component 


maciekmm:
**What happened**:

Creating a cluster with `kind create cluster` under rootful and rootless podman hangs on `⢎⡠ Starting control-plane 🕹️ ` and later fails with

> [kubelet-check] Initial timeout of 40s passed.
>     114 round_trippers.go:553] GET https://kind-control-plane:6443/healthz?timeout=10s  in 2 milliseconds

```bash
~ » sudo kind create cluster                                                                                                                                                                        
enabling experimental podman provider
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.24.0) 🖼 
 ✓ Preparing nodes 📦  
 ✓ Writing configuration 📜 
⢎⡠ Starting control-plane 🕹️ ^C
```

**What you expected to happen**:

I'd expect the `kind create cluster` to successfully create a cluster.

**How to reproduce it (as minimally and precisely as possible)**:

On a zfs backed Arch system installation running podman (rootful or rootless) with `fuse-overlayfs` storage driver execute `kind create cluster`.

**Anything else we need to know?**:

Running `kind` with `--retain` flag and peeking inside the container we see:

```
# journalctl -u kubelet.service
Aug 08 19:47:12 kind-control-plane kubelet[223]: E0808 19:47:12.457606     223 kubelet.go:2419] "Error getting node" err="node \"kind-control-plane\" not found"
Aug 08 19:47:12 kind-control-plane kubelet[223]: E0808 19:47:12.524655     223 dns.go:157] "Nameserver limits exceeded" err="Nameserver limits were exceeded, some nameservers have been omitted, the applied nameserver line is: fc00:f853:ccd:e793::1 10.89.0.1 84.208.20.110"
Aug 08 19:47:12 kind-control-plane kubelet[223]: E0808 19:47:12.558482     223 kubelet.go:2419] "Error getting node" err="node \"kind-control-plane\" not found"
[...]

PR #2874, opened 2022-08-10
rework snapshotter selection, detect fuseblk and use overlayfs-fuse

BenTheElder:
see: https://github.com/kubernetes-sigs/kind/issues/2868

### DIFF
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -101,22 +101,36 @@ overlayfs_preferrable() {
 
 configure_containerd() {
   local snapshotter=${KIND_EXPERIMENTAL_CONTAINERD_SNAPSHOTTER:-}
+
+  # handle userns (rootless)
   if [[ -n "$userns" ]]; then
-    # userns (rootless) configs
-
-    # Adjust oomScoreAdj
+    # enable restrict_oom_score_adj
     sed -i 's/restrict_oom_score_adj = false/restrict_oom_score_adj = true/' /etc/containerd/config.toml
-
     # Use fuse-overlayfs if overlayfs is not preferrable: https://github.com/kubernetes-sigs/kind/issues/2275
     if [[ -z "$snapshotter" ]] && ! overlayfs_preferrable; then
       snapshotter="fuse-overlayfs"
     fi
-  else
-    # we need to switch to the 'native' snapshotter on zfs
-    if [[ -z "$snapshotter" ]] && [[ "$(stat -f -c %T /kind)" == 'zfs' ]]; then
-      snapshotter="native"
+  fi
+
+  # if we have not already overridden the snapshotter, attempt to auto select
+  if [[ -z "$snapshotter" ]]; then
+    # we need to switch to 'native' or 'fuse-overlayfs' on zfs
+    container_filesystem="$(stat -f -c %T /kind)"
+    if [[ "$container_filesystem" == 'zfs' ]]; then
+      # if fuse is present, use fuse-overlayfs, else fallback to native
+      # we do not use the ZFS snapshotter because of skew issues vs the host
+      if [[ -e /dev/fuse ]]; then
+        snapshotter="fuse-overlayfs"
+      else
+        snapshotter="native"
+      fi
+    # fuse likely implies fuse-overlayfs, we should switch to fuse-overlayfs (or native)
+    elif [[ "$container_filesystem" == 'fuseblk' ]]; then
+      snapshotter="fuse-overlayfs"
     fi
   fi
+
+  # if we've overridden or auto-selected the snapshotter vs the default, update containerd
   if [[ -n "$snapshotter" ]]; then
     echo "INFO: changing snapshotter from \"overlayfs\" to \"$snapshotter\""
     sed -i "s/snapshotter = \"overlayfs\"/snapshotter = \"$snapshotter\"/" /etc/containerd/config.toml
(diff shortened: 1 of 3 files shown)


aojea:
@AkihiroSuda can review this better than me

aojea:
meant to use an emoji :smile:

images/base/files/usr/local/bin/entrypoint
```diff
@@ -101,22 +101,30 @@ overlayfs_preferrable() {
...
+
+  # if we have not already overridden the snapshotter, attempt to auto select
+  if [[ -z "$snapshotter" ]]; then
     # we need to switch to the 'native' snapshotter on zfs
-    if [[ -z "$snapshotter" ]] && [[ "$(stat -f -c %T /kind)" == 'zfs' ]]; then
+    container_filesystem="$(stat -f -c %T /kind)"
+    if [[ "$container_filesystem" == 'zfs' ]]; then
       snapshotter="native"
+    # fuse likely implies overlayfs-fuse, we should switch to overlayfs-fuse (or native)
+    elif [[ "$container_filesystem" == 'fuseblk' ]]; then
```
BenTheElder, line 122: this is the functional difference, the rest is rearranging a little.

we will select fuse-overlayfs if we detect we're in fuseblk and we haven't otherwise already selected a snapshotter due to userns or KIND_EXPERIMENTAL_CONTAINERD_SNAPSHOTTER

BenTheElder: I don't _think_ there's any supported snapshotter environment where we'd be in fuse within the container filesystem but not using fuse-overlayfs.

Technically the user could configure a mount over `/kind` but that's already an unsafe choice, we should perhaps but a README in there to let people know that this is an internal directory.

pkg/apis/config/defaults/image.go
```diff
@@ -18,4 +18,5 @@ limitations under the License.
 package defaults
 
 // Image is the default for the Config.Image field, aka the default node image.
-const Image = "kindest/node:v1.24.3@sha256:e1a5c449fb2556b5fbba29fb43f28055cacb12ce53d1f281e8be442d0dd5cd1d"
+// TODO: DO NOT MERGE with personal test image
+const Image = "docker.io/bentheelder/kind-node:v1.24.3@sha256:9fdef952a166325f89cae70b7e69767b50b7a7b0ace1f03df362228b82ee34e4"
```
maciekmm, line 22: The provided image does not seem to contain the changes listed in this PR.
See https://github.com/kubernetes-sigs/kind/issues/2868#issuecomment-1211183768

BenTheElder: sorry, fixed!

BenTheElder:
This works as expected in all our CI environments, locally, and in https://github.com/kubernetes-sigs/kind/issues/2868#issuecomment-1211253886

Other than maybe revamping the versioning scheme, I think this should be our last fix for v0.15.0 and we should roll these out v0.14.0 was march and we've soaked some notable fixes + the rootless fix ahead of K8s v1.25.

https://github.com/kubernetes-sigs/kind/compare/v0.14.0...main is looking pretty reasonable.

BenTheElder:
PV block test flake
/retest

images/base/files/usr/local/bin/entrypoint
```diff
@@ -101,22 +101,30 @@ overlayfs_preferrable() {
...
+  fi
+
+  # if we have not already overridden the snapshotter, attempt to auto select
+  if [[ -z "$snapshotter" ]]; then
     # we need to switch to the 'native' snapshotter on zfs
-    if [[ -z "$snapshotter" ]] && [[ "$(stat -f -c %T /kind)" == 'zfs' ]]; then
+    container_filesystem="$(stat -f -c %T /kind)"
+    if [[ "$container_filesystem" == 'zfs' ]]; then
       snapshotter="native"
+    # fuse likely implies overlayfs-fuse, we should switch to overlayfs-fuse (or native)
```
AkihiroSuda, line 121: s/overlayfs-fuse/fuse-overlayfs/

images/base/files/usr/local/bin/entrypoint
```diff
@@ -101,22 +101,30 @@ overlayfs_preferrable() {
...
-  else
+  fi
+
+  # if we have not already overridden the snapshotter, attempt to auto select
+  if [[ -z "$snapshotter" ]]; then
     # we need to switch to the 'native' snapshotter on zfs
-    if [[ -z "$snapshotter" ]] && [[ "$(stat -f -c %T /kind)" == 'zfs' ]]; then
+    container_filesystem="$(stat -f -c %T /kind)"
+    if [[ "$container_filesystem" == 'zfs' ]]; then
       snapshotter="native"
```
AkihiroSuda, line 120: Not new in this PR but I guess we can use fuse-overlayfs on zfs

BenTheElder: Hmm true, I think if we did we'd have to also check that fuse is actually present to not regress on https://github.com/kubernetes-sigs/kind/issues/2540

images/base/files/usr/local/bin/entrypoint
BenTheElder: fixed, thanks.

images/base/files/usr/local/bin/entrypoint
BenTheElder: done.

PR #1926, opened 2020-11-13
added: support for adding labels from the config file itself

yashvardhan-kukreja:
closes #1574 

**My approach:**
So, the labelling of the nodes is going to be one of the actions (`actionsToRun`) when cluster creation happens in the code.
Now, I have not coded a dedicated action for "just the labelling of nodes". 
I have rather created a file `postcreationconfiguration.go` which will consist of all the functions which are supposed to occur after the cluster gets created. (like labelling in our case)

And in the file `postcreationconfiguration.go`, I have written a `labelNodes` function which will label the nodes from the config file.

**Why did I do it?**
Because, say, in the future, there is another requirement like adding taints or tolerations (_or some other configuration, which usually in real life are done after cluster is created_) through the config file itself. 
Then, we can write functions for those things in the `postcreationconfigurations.go` itself instead of creating dedicated actions for them. That's because these things like adding taints or tolerations are usually done after the cluster is already created.

**What's happening behind the scenes?**
So, basically, at every iteration, if the labels are being applied to a control-plane node, then, that control-plane-node itself will self-apply those labels to itself.
But if the labels are being applied to a worker node, then, one of the control plane nodes will apply those labels to that worker node.

**Usage from user's side**:
```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
- role: control-plane
  labels:
    name: master
    index: 2
- role: worker
  labels:
    name: worker
    tier: app
- role: worker
```

**Output**: (Look over the end of every labels row)
```
NAME                   STATUS  ROLES   LABELS
master-control-plane   Ready   master  beta.kubernetes.io/arch=amd64,beta.kubernetes.io/os=linux,index=2,kubernetes.io/arch=amd64,kubernetes.io/hostname=master-control-plane,kubernetes.io/os=linux,name=master,node-role.kubernetes.io/master=
master-control-plane2  Ready   master  beta.kubernetes.io/arch=amd64,beta.kubernetes.io/os=linux,kubernetes.io/arch=amd64,kubernetes.io/hostname=master-control-plane2,kubernetes.io/os=linux,node-role.kubernetes.io/master=
master-worker          Ready   <none>  beta.kubernetes.io/arch=amd64,beta.kubernetes.io/os=linux,kubernetes.io/arch=amd64,kubernetes.io/hostname=master-worker,kubernetes.io/os=linux
[...]

Issue #1574, opened 2020-05-09
Support node labels within config files


nadenf:
**What would you like to be added**:

It would be nice if we had a simple:
```
- role: worker
  labels:
     type: zookeeper
     node-role.kubernetes.io/zookeeper=true: true
```

**Why is this needed**:

Using node-labels within the worker KubeAdm JoinConfiguration is:
a) Cumbersome for new users
b) Fails if you try to set a node-role which can't be specified in the JoinConfiguration

BenTheElder:
Definitely in favor of this, was sort of using https://github.com/kubernetes-sigs/kind/issues/1361 to track but this is clearer.

BenTheElder:
I have a mostly finished PR our that drops v1alpha3 amongst other improvements, to will be easier to add this after that.

yashvardhan-kukreja:
I would like to take this up, as there doesn't seem any PR resolving/referencing to this issue.

neolit123:
Note that the kubelet will not allow a node to self apply node-role labels if the NodeRestriction admission plugin is running.

yashvardhan-kukreja:
Yes, I have dealt with that.
the labels are being applied by the control plane node only.

> Note that the kubelet will not allow a node to self apply node-role labels if the NodeRestriction admission plugin is running.

PR #2877, opened 2022-08-13
Add extra labels section for per-node options

Benny-Git:
Ref. #1926 #1574
Add extra labels section for per-node options

This feature was introduced with https://github.com/kubernetes-sigs/kind/pull/1926
I needed something like this, and did not find it in the docs.
After searching some more I found the PR that introduced this, and decided to add it to the docs.

### DIFF
--- a/site/content/docs/user/configuration.md
+++ b/site/content/docs/user/configuration.md
@@ -367,6 +367,29 @@ spec:
 
 [Ingress Guide]: /docs/user/ingress
 
+### Extra Labels
+
+Extra labels might be useful for working with
+[nodeSelectors](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/).
+
+An example label for specifying a `tier` label:
+
+{{< codeFromInline lang="yaml">}}
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+nodes:
+- role: control-plane
+- role: worker
+  extraPortMappings:
+  - containerPort: 30950
+    hostPort: 80
+  labels:
+    tier: frontend
+- role: worker
+  labels:
+    tier: backend
+{{< /codeFromInline >}}
+
 ### Kubeadm Config Patches
 
 KIND uses [`kubeadm`](/docs/design/principles/#leverage-existing-tooling) 


BenTheElder:
Thank you!!

rendered: https://deploy-preview-2877--k8s-kind.netlify.app/docs/user/configuration/#extra-labels

given the huge delay I want to just go ahead and proactively merge this, but we should consider mentioning Kubernetes's restrictions on these in a follow-up: https://kubernetes.io/docs/reference/labels-annotations-taints/

> Kubernetes reserves all labels and annotations in the kubernetes.io namespace.

which, is actually missing something, it's k8s.io too. will need to go PR the main docs as well https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.16.md#deprecations-and-removals:~:text=removed%20the%20ability%20to%20set%20kubernetes.io%2D%20or%20k8s.io%2Dprefixed%20labels%20via%20%2D%2Dnode%2Dlabels%2C%20other%20than%20the%20specifically%20allowed%20labels/prefixes.%20(%2379305%2C%20%40paivagustavo)

Benny-Git:
> which, is actually missing something, it's k8s.io too. will need to go PR the main docs as well

I think the main doc's intention is to document the existing well-known labels and what they should be used for. It does not serve the purpose to inform which prefixes not to use?

The only thing hinting at that limitation is the sentence "Kubernetes reserves all labels and annotations in the kubernetes.io namespace." However I would not interpret the changelog's entry to mean `k8s.io` is genereally reserved - it looks like this only applies to Node labels?

BenTheElder:
> I think the main doc's intention is to document the existing well-known labels and what they should be used for. It does not serve the purpose to inform which prefixes not to use?

Perhaps, but _the first sentence_ of the page notes that kubernetes.io is reserved.

> However I would not interpret the changelog's entry to mean k8s.io is genereally reserved - it looks like this only applies to Node labels?

It's only enforced on node labels, it makes sense there because nodes are owned wholly by kubelet.

The namespaces are still broadly considered reserved for labels, apis, etc. AFAIK.

Unlike kubelet / Node, it makes sense to let third party controllers set labels like `app.kubernetes.io/created-by` on objects like pod, however it's *not* intended to invent your own k8s.io labels AFAIK.

There was also a cleanup to move people's CRDs out of k8s.io if they were not approved APIs reviewed by the upstream API review team. https://github.com/kubernetes/enhancements/pull/1111

----

In any case, I'm sure users are going to run into issues with failures trying to set k8s.io and kubernetes.io node labels which will not work since 1.16, we should make sure that's noted somewhere.

/override pull-kind-e2e-kubernetes-1-24

Issue #2882, opened 2022-08-17
current kind does not seem to support netavark backed rootless podman(4.x+)


wherka-ama:
**What happened**:
Failure during the cluster creation.

$ /usr/bin/kind create cluster --name test3
enabling experimental podman provider
Creating cluster "test3" ...
 ✓ Ensuring node image (kindest/node:v1.24.0) 🖼 
 ✓ Preparing nodes 📦  
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
sigs.k8s.io/kind/pkg/cluster/internal/providers/podman.getSubnets({0x77d9fb?, 0x7a6280?})
	sigs.k8s.io/kind/pkg/cluster/internal/providers/podman/provision.go:275 +0x199
sigs.k8s.io/kind/pkg/cluster/internal/providers/podman.getProxyEnv(0x77db16?, {0x77d9fb, 0x4})

**What you expected to happen**:
Cluster created successfully

**How to reproduce it (as minimally and precisely as possible)**:
The problem can be recreated on any rootless installation with podman backed with netavark. The key here is the output structure produced by `podman network inspect kind`. In older version of podman, where the CNI plugin was used the subnets where displayed  as `[n]plugins[n].ipam.ranges[n][m].subnet` while in a netavark it's more like `[n].subnets[n].subnet`

Examples:
*** cni:
```
[
    {
        "cniVersion": "0.4.0",
        "name": "kind",
        "plugins": [
            {
                "bridge": "cni-podman2",
                "hairpinMode": true,
                "ipMasq": true,
                "ipam": {
                    "ranges": [
                        [
                            {
                                "gateway": "fc00:8866:27d0:bd7e::1",
                                "subnet": "fc00:8866:27d0:bd7e::/64"
                            }
                        ],
                        [
                            {
                                "gateway": "10.89.1.1",
                                "subnet": "10.89.1.0/24"
                            }
                        ]
                    ],
                    "routes": [
                        {
                            "dst": "::/0"
                        },
                        {
                            "dst": "0.0.0.0/0"
                        }
                    ],
                    "type": "host-local"
                },
                "isGateway": true,
                "type": "bridge"
            },
            {
                "capabilities": {
                    "portMappings": true
                },
                "type": "portmap"
            },
            {
                "backend": "",
                "type": "firewall"
            },
            {
                "type": "tuning"
            },
            {
                "capabilities": {
                    "aliases": true
                },
                "domainName": "dns.podman",
[...]

PR #2883, opened 2022-08-17
Support for netavark backed podman(4.x+)  - fixing #2882

wherka-ama:
Adding support for netavark backed podman(4.x+) 

See: https://github.com/kubernetes-sigs/kind/issues/2882

### DIFF
--- a/pkg/cluster/internal/providers/podman/provision.go
+++ b/pkg/cluster/internal/providers/podman/provision.go
@@ -252,12 +252,12 @@ func getProxyEnv(cfg *config.Cluster, networkName string) (map[string]string, er
 			return nil, err
 		}
 		noProxyList := append(subnets, envs[common.NOProxy])
-		// Add pod, service and control plane(API server) dns names to no_proxy to allow in cluster
+		// Add pod and service dns names to no_proxy to allow in cluster
 		// Note: this is best effort based on the default CoreDNS spec
 		// https://github.com/kubernetes/dns/blob/master/docs/specification.md
 		// Any user created pod/service hostnames, namespaces, custom DNS services
 		// are expected to be no-proxied by the user explicitly.
-		noProxyList = append(noProxyList, ".svc", ".svc.cluster", ".svc.cluster.local", strings.Join([]string{cfg.Name, "control-plane"}, "-"))
+		noProxyList = append(noProxyList, ".svc", ".svc.cluster", ".svc.cluster.local")
 		noProxyJoined := strings.Join(noProxyList, ",")
 		envs[common.NOProxy] = noProxyJoined
 		envs[strings.ToLower(common.NOProxy)] = noProxyJoined
--- a/pkg/cluster/internal/providers/podman/provision.go
+++ b/pkg/cluster/internal/providers/podman/provision.go
@@ -265,7 +265,7 @@ func getProxyEnv(cfg *config.Cluster, networkName string) (map[string]string, er
 	return envs, nil
 }
 
-type PodmanNetworks []struct {
+type podmanNetworks []struct {
 	// v4+
 	Subnets []struct {
 		Subnet  string `json:"subnet"`
@@ -290,7 +290,7 @@ func getSubnets(networkName string) ([]string, error) {
 		return nil, errors.Wrap(err, "failed to get subnets")
 	}
 
-	networks := PodmanNetworks{}
+	networks := podmanNetworks{}
 	jsonErr := json.Unmarshal([]byte(out), &networks)
 	if jsonErr != nil {
 		return nil, errors.Wrap(jsonErr, "failed to get subnets")
--- a/pkg/cluster/internal/providers/podman/provision.go
+++ b/pkg/cluster/internal/providers/podman/provision.go
@@ -18,6 +18,7 @@ package podman
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


pkg/cluster/internal/providers/podman/provision.go
```diff
@@ -251,28 +252,67 @@ func getProxyEnv(cfg *config.Cluster, networkName string) (map[string]string, er
...
-		noProxyList = append(noProxyList, ".svc", ".svc.cluster", ".svc.cluster.local")
+		noProxyList = append(noProxyList, ".svc", ".svc.cluster", ".svc.cluster.local", strings.Join([]string{cfg.Name, "control-plane"}, "-"))
 		noProxyJoined := strings.Join(noProxyList, ",")
 		envs[common.NOProxy] = noProxyJoined
 		envs[strings.ToLower(common.NOProxy)] = noProxyJoined
 	}
 	return envs, nil
 }
 
+type PodmanNetworks []struct {
```
BenTheElder, line 268: This type shouldn’t be exported.

pkg/cluster/internal/providers/podman/provision.go
```diff
@@ -251,28 +252,67 @@ func getProxyEnv(cfg *config.Cluster, networkName string) (map[string]string, er
...
 		}
 		noProxyList := append(subnets, envs[common.NOProxy])
-		// Add pod and service dns names to no_proxy to allow in cluster
+		// Add pod, service and control plane(API server) dns names to no_proxy to allow in cluster
 		// Note: this is best effort based on the default CoreDNS spec
 		// https://github.com/kubernetes/dns/blob/master/docs/specification.md
 		// Any user created pod/service hostnames, namespaces, custom DNS services
 		// are expected to be no-proxied by the user explicitly.
-		noProxyList = append(noProxyList, ".svc", ".svc.cluster", ".svc.cluster.local")
+		noProxyList = append(noProxyList, ".svc", ".svc.cluster", ".svc.cluster.local", strings.Join([]string{cfg.Name, "control-plane"}, "-"))
```
BenTheElder, line 260: Changes to this method seem unrelated to netavark support and should also be kept consistent with the docker implementation

BenTheElder (commented):
Thanks for the PR!

pkg/cluster/internal/providers/podman/provision.go
BenTheElder: This should be a separate commit or PR, and it should probably not reimplement naming the control plane endpoint with different logic.

wherka-ama: @BenTheElder : agreed, this is a separate problem. Should I create a separate issue for that and take this part out from the current PR?

aojea:
/hold

holding waiting for podman developers to disclose more about the future plans of the network https://github.com/kubernetes-sigs/kind/issues/2882#issuecomment-1217734268

we already have several places that fork based on the  podman version, and we can't keep growing like that

pkg/cluster/internal/providers/podman/provision.go
wherka-ama: Corrected. Thanks!

pkg/cluster/internal/providers/podman/provision.go
wherka-ama: > This should be a separate commit or PR, and it should probably not reimplement naming the control plane endpoint with different logic.

Agreed. I've taken it out and will create a separate issue + another PR as an attempt to tackle the no_proxy inconsistencies. Thanks!

pkg/cluster/internal/providers/podman/provision.go
```diff
@@ -264,15 +265,54 @@ func getProxyEnv(cfg *config.Cluster, networkName string) (map[string]string, er
...
+				Gateway string `json:"gateway"`
+				Subnet  string `json:"subnet"`
+			} `json:"ranges"`
+		} `json:"ipam,omitempty"`
+	} `json:"plugins"`
+}
+
 func getSubnets(networkName string) ([]string, error) {
-	// TODO: unmarshall json and get rid of this complex query
-	format := `{{ range (index (index (index (index . "plugins") 0 ) "ipam" ) "ranges")}}{{ index ( index . 0 ) "subnet" }} {{end}}`
```
aojea, line 269: I can't remember now exactly, but I think we only care about the first plugin and subnet

wherka-ama: > I can't remember now exactly, but I think we only care about the first plugin and subnet

@aojea : would you like me to change that to replicate this behaviour here as well? I'm not sure what was the reasoning behind the first implementation, so would be happy to follow your guidance.

aojea: yeah. match the first implementation, it is safer

pkg/cluster/internal/providers/podman/provision.go
BenTheElder: thank you!

pkg/cluster/internal/providers/podman/provision.go
```diff
@@ -264,15 +265,54 @@ func getProxyEnv(cfg *config.Cluster, networkName string) (map[string]string, er
...
+			Ranges [][]struct {
+				Gateway string `json:"gateway"`
+				Subnet  string `json:"subnet"`
+			} `json:"ranges"`
+		} `json:"ipam,omitempty"`
+	} `json:"plugins"`
+}
+
 func getSubnets(networkName string) ([]string, error) {
-	// TODO: unmarshall json and get rid of this complex query
```
BenTheElder, line 268: New TODO: unit test parsing with some known real data to prevent regressions. (doesn't need to be this PR, we should have done that before)

BenTheElder (commented):
/lgtm
/approve
thanks so much!

pkg/cluster/internal/providers/podman/provision.go
wherka-ama: > New TODO: unit test parsing with some known real data to prevent regressions. (doesn't need to be this PR, we should have done that before)

@BenTheElder : funny you mentioned that cause I thought there is something missing in that area indeed and I've already started working on unit tests to cover at least what I've touched ;-)
We can either do it in the next PR, which I'm happy to deliver pretty soon(couple of days max, but hopefully tomorrow) or we can put this one on hold and bundle it with unit tests.  Please do let me know what's your preference and I'll go with it.

BenTheElder: another PR when you have time would be great!! no rush, I'm trying to land one or two more fixes ahead of k8s 1.25 / kind v0.15 but I think unit tests can wait / are more forward thinking 😅

pkg/cluster/internal/providers/podman/provision.go
wherka-ama: @aojea : I have a couple of points before I actually push any further commits to this PR

1. At one point in the issue (https://github.com/kubernetes-sigs/kind/issues/2882#issuecomment-1218019479) you pointed at the existing checks which I could use as my guideline. I'm fine with the version check(if we are to stick with explicit flows), but the part where there was a check to see if we are in the Rootless mode doesn't seem appropriate to me. The trigger of the error here is the presence of the environment variables linked with http(s) proxy. It makes no difference whether we run it as root or not. I've tested it and we have a consistent behaviour here. It is indeed generic in that regard.
2. I did a bit of a refactoring of the code to be very explicit with the versions/layouts used so that allows on very easy decommissioning of the legacy flow when the time comes. However, I must stress it became excessively verbose and I count on your judgement here to decide if that's the direction we should take.
3. In the refactored version I'm referring to the first plugin/subnet - mimicking the old query.  

@BenTheElder: I would welcome your opinion on that aspect as well.

[patch.txt](https://github.com/kubernetes-sigs/kind/files/9365140/patch.txt)

Issue #2884, opened 2022-08-17
no_proxy fed to the cluster is missing the control plane API dns name


wherka-ama:
**What happened**:
Kind failed to create the cluster in the environment with http(s) proxy
```
>kind create cluster
enabling experimental podman provider
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.24.3) 🖼 
 ✓ Preparing nodes 📦  
 ✓ Writing configuration 📜 
 ✗ Starting control-plane 🕹️ 
ERROR: failed to create cluster: failed to init node with kubeadm: command "podman exec --privileged kind-control-plane kubeadm init --skip-phases=preflight --config=/kind/kubeadm.conf --skip-token-print --v=6" failed with error: exit status 1
Command Output: I0817 10:16:02.075463     148 initconfiguration.go:255] loading configuration from "/kind/kubeadm.conf"
W0817 10:16:02.076834     148 initconfiguration.go:332] [config] WARNING: Ignored YAML document with GroupVersionKind kubeadm.k8s.io/v1beta3, Kind=JoinConfiguration
[init] Using Kubernetes version: v1.24.3
[certs] Using certificateDir folder "/etc/kubernetes/pki"
I0817 10:16:02.082340     148 certs.go:112] creating a new certificate authority for ca
[certs] Generating "ca" certificate and key
I0817 10:16:02.217935     148 certs.go:522] validating certificate period for ca certificate
[certs] Generating "apiserver" certificate and key
[certs] apiserver serving cert is signed for DNS names [kind-control-plane kubernetes kubernetes.default kubernetes.default.svc kubernetes.default.svc.cluster.local localhost] and IPs [10.96.0.1 10.89.0.14 127.0.0.1]
[certs] Generating "apiserver-kubelet-client" certificate and key
I0817 10:16:02.444010     148 certs.go:112] creating a new certificate authority for front-proxy-ca
[certs] Generating "front-proxy-ca" certificate and key
I0817 10:16:02.524854     148 certs.go:522] validating certificate period for front-proxy-ca certificate
[certs] Generating "front-proxy-client" certificate and key
I0817 10:16:02.713906     148 certs.go:112] creating a new certificate authority for etcd-ca
[certs] Generating "etcd/ca" certificate and key
I0817 10:16:02.897146     148 certs.go:522] validating certificate period for etcd/ca certificate
[certs] Generating "etcd/server" certificate and key
[...]

PR #2885, opened 2022-08-17
Extending the no_proxy env variable to automatically include all the nodes of the cluster

wherka-ama:
See: https://github.com/kubernetes-sigs/kind/issues/2884

### DIFF
--- a/pkg/cluster/internal/providers/docker/provision.go
+++ b/pkg/cluster/internal/providers/docker/provision.go
@@ -286,22 +286,12 @@ func getProxyEnv(cfg *config.Cluster, networkName string, nodeNames []string) (m
 
 		noProxyList := append(subnets, envs[common.NOProxy])
 		noProxyList = append(noProxyList, nodeNames...)
-		// Add pod,service and all the cluster nodes' dns names to no_proxy to allow in cluster
+		// Add pod and service dns names to no_proxy to allow in cluster
 		// Note: this is best effort based on the default CoreDNS spec
 		// https://github.com/kubernetes/dns/blob/master/docs/specification.md
 		// Any user created pod/service hostnames, namespaces, custom DNS services
 		// are expected to be no-proxied by the user explicitly.
-		var clusterNodeNames []string
-		nodeNamer := common.MakeNodeNamer(cfg.Name)
-		for _, node := range cfg.Nodes {
-			clusterNodeNames = append(clusterNodeNames, nodeNamer(string(node.Role)))
-		}
-		if len(cfg.Nodes) > 1 {
-			clusterNodeNames = append(clusterNodeNames, nodeNamer(string(constants.ExternalLoadBalancerNodeRoleValue)),
-				nodeNamer(string(constants.ExternalEtcdNodeRoleValue)))
-		}
 		noProxyList = append(noProxyList, ".svc", ".svc.cluster", ".svc.cluster.local")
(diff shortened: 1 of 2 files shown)


wherka-ama:
/cc @BenTheElder @aojea 
I'm pretty sure the test failure is unrelated to the change. It's worth rerunning it.

pkg/cluster/internal/providers/podman/provision.go
```diff
@@ -252,12 +252,23 @@ func getProxyEnv(cfg *config.Cluster, networkName string) (map[string]string, er
 			return nil, err
 		}
 		noProxyList := append(subnets, envs[common.NOProxy])
-		// Add pod and service dns names to no_proxy to allow in cluster
+		// Add pod,service and all the cluster nodes' dns names to no_proxy to allow in cluster
 		// Note: this is best effort based on the default CoreDNS spec
 		// https://github.com/kubernetes/dns/blob/master/docs/specification.md
 		// Any user created pod/service hostnames, namespaces, custom DNS services
 		// are expected to be no-proxied by the user explicitly.
+		var clusterNodeNames []string
```
BenTheElder, line 260: we should probably just restructure to pass these in or add them to the list at the call site?

pkg/cluster/internal/providers/docker/provision.go
```diff
@@ -286,12 +286,22 @@ func getProxyEnv(cfg *config.Cluster, networkName string, nodeNames []string) (m
 
 		noProxyList := append(subnets, envs[common.NOProxy])
 		noProxyList = append(noProxyList, nodeNames...)
```
BenTheElder, line 288: aren't we already adding these here?

pkg/cluster/internal/providers/podman/provision.go
aojea: I was thinking in adding them here

https://github.com/kubernetes-sigs/kind/blob/b6bc112522651d98c81823df56b7afa511459a3b/pkg/cluster/internal/providers/common/proxy.go#L41-L63

that will cascade for all the providers

BenTheElder: The docker implementation already has this AFAICT, we just need to modify podman to match and drop the docker changes, I think https://github.com/kubernetes-sigs/kind/pull/2885#discussion_r950321305

pkg/cluster/internal/providers/docker/provision.go
wherka-ama: Good point! Indeed it looks like we can revert for docker and try to pass the list the same way as it's done here indeed.. I'll do it right away. Thanks!

pkg/cluster/internal/providers/podman/provision.go
wherka-ama: If we were to match the podman flow with the docker one in that regard I guess the best option to start the restructuring from the `planCreation` and trickle it down. I'd be happy to do that.

@BenTheElder  @aojea : are you fine with that direction?

wherka-ama: I've just pushed a fresh commit with the refactoring. In short - it's an alignment with what we have for docker in that regards. I  must admit I like this version a lot more than the previous one. I hope you don't disagree ;-)

/cc @BenTheElder @aojea

PR #2891, opened 2022-08-19
image-archive: enable loading multiple image archives

### LINKED ISSUE #2881: kind load image-archive xx1.tar xx2.tar , cannot find xx2.tar's image in container of kind-node
**What happened**:
docker pull ghcr.io/kubean-io/kubean/spray-job:v0.1.0
docker pull ghcr.io/kubean-io/kubean/kubean-operator:v0.1.0
docker save ghcr.io/kubean-io/kubean/spray-job:v0.1.0       > kubean-job-v0.1.0.tar
docker save ghcr.io/kubean-io/kubean/kubean-operator:v0.1.0 > kubean-operator-v0.1.0.tar

cat /tmp/kind_config.yaml
kind: Cluster
nodes:
- role: control-plane
  extraPortMappings:
  - containerPort: 30080
    hostPort: 80

kind create cluster --name my-mgmt-cluster-installer --image kindest/node:v1.21.1 --config /tmp/kind_config.yaml

kind load image-archive offline/kubean/kubean-job-v0.1.0.tar offline/kubean/kubean-operator-v0.1.0.tar --name=my-mgmt-cluster-installer

####it's happened:####
docker exec -it my-mgmt-cluster-installer-control-plane bash
crictl images|grep kubean
企业微信截图_16607079917370

####I'll import it offline/kubean/kubean-operator-v0.1.0.tar####

kind load image-archive offline/kubean/kubean-operator-v0.1.0.tar --name=my-mgmt-cluster-installer

####it's happened:####
docker exec -it my-mgmt-cluster-installer-control-plane bash
crictl images|grep kubean
企业微信截图_16607080606723

**Environment:**

kind v0.14.0 go1.18.2 linux/amd64

Client Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.3", GitCommit:"aef86a93758dc3cb2c658dd9657ab4ad4afc21cb", GitTreeState:"clean", BuildDate:"2022-07-13T14:30:46Z", GoVersion:"go1.18.3", Compiler:"gc", Platform:"linux/amd64"}

Client: Docker Engine - Community
 Version:           20.10.17
 Engine:
  Version:          20.10.17

[root@shihang-operation ~/dce5-installer/offline]# cat /etc/os-release 
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"
CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="7"

harshanarayana:
Current command implementation of the `kind load image-archive` expects that only one tar can be loaded at a time. Though the command handler checks to see a minimum of one file is passed as argument to the command while invocation, it doesn't enforce the maximum number of files that can be passed to the command. 

However, the implementation picks the `args[0]` and ignores the rest without any error or warning. This can lead to a bit of confusion. It is useful to be able to load multiple image archives in one go anyway.

This PR enables the ability to load images from more than one archive at one go while honouring the original behavior.

Closes #2881 

```bash
❯ docker exec -it test-control-plane bash
root@test-control-plane:/# crictl images | grep alpine
root@test-control-plane:/#
exit

❯ docker save alpine:3.12 > archive1.tar

❯ docker save alpine:3.13 > archive2.tar

❯ bin/kind load image-archive archive1.tar archive2.tar --name test

❯ docker exec -it test-control-plane bash
root@test-control-plane:/# crictl images | grep alpine
docker.io/library/alpine                   3.12                                      24c8ece58a1aa       5.87MB
docker.io/library/alpine                   3.13                                      6b5c5e00213a4       5.91MB
root@test-control-plane:/#
exit

❯ bin/kind -v=3 load image-archive archive1.tar archive2.tar --name test
DEBUG: image-archive/image-archive.go:142] Loading Docker Image from archive archive1.tar to node test-control-plane
DEBUG: image-archive/image-archive.go:142] Loading Docker Image from archive archive2.tar to node test-control-plane
```

### DIFF
--- a/pkg/cmd/kind/load/image-archive/image-archive.go
+++ b/pkg/cmd/kind/load/image-archive/image-archive.go
@@ -49,6 +49,11 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 			}
 			return nil
 		},
+		PreRun: func(cmd *cobra.Command, args []string) {
+			if len(args) > 1 {
+				logger.Warn("It is suggested that you save multiple images into a common archive and load that instead of loading multiple archives for better performance")
+			}
+		},
 		Use:   "image-archive <IMAGE.tar>",
 		Short: "Loads docker image from archive into nodes",
 		Long:  "Loads docker image from archive into all or specified nodes by name",
@@ -78,19 +83,28 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
 		runtime.GetDefault(logger),
 	)
 
-	// Check if file exists
-	imageTarPath := args[0]
-	if _, err := os.Stat(imageTarPath); err != nil {
-		return err
+	for _, imageTarPath := range args {
+		if _, err := os.Stat(imageTarPath); err != nil {
+			return err
+		}
 	}
 
+	for _, imageTarPath := range args {
+		if err := loadArchiveToNodes(logger, provider, flags.Name, flags.Nodes, imageTarPath); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func loadArchiveToNodes(logger log.Logger, provider *cluster.Provider, clusterName string, nodeNames []string, imageArchivePath string) error {
 	// Check if the cluster nodes exist
-	nodeList, err := provider.ListInternalNodes(flags.Name)
+	nodeList, err := provider.ListInternalNodes(clusterName)
 	if err != nil {
 		return err
 	}
 	if len(nodeList) == 0 {
-		return fmt.Errorf("no nodes found for cluster %q", flags.Name)
+		return fmt.Errorf("no nodes found for cluster %q", clusterName)
 	}
 
 	// map cluster nodes by their name
@@ -104,9 +118,9 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
 	// pick only the user selected nodes and ensure they exist
 	// the default is all nodes unless flags.Nodes is set
 	selectedNodes := nodeList
-	if len(flags.Nodes) > 0 {
+	if len(nodeNames) > 0 {
 		selectedNodes = []nodes.Node{}
-		for _, name := range flags.Nodes {
+		for _, name := range nodeNames {
 			node, ok := nodesByName[name]
 			if !ok {
 				return fmt.Errorf("unknown node: %s", name)
@@ -120,18 +134,19 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
 	for _, selectedNode := range selectedNodes {
 		selectedNode := selectedNode // capture loop variable
 		fns = append(fns, func() error {
-			return loadImage(imageTarPath, selectedNode)
+			return loadImage(logger, imageArchivePath, selectedNode)
 		})
 	}
 	return errors.UntilErrorConcurrent(fns)
 }
 
 // loads an image tarball onto a node
-func loadImage(imageTarName string, node nodes.Node) error {
+func loadImage(logger log.Logger, imageTarName string, node nodes.Node) error {
 	f, err := os.Open(imageTarName)
 	if err != nil {
 		return errors.Wrap(err, "failed to open image")
 	}
 	defer f.Close()
+	logger.V(2).Infof("Loading Docker Image from archive %s to node %s", imageTarName, node.String())
 	return nodeutils.LoadImageArchive(node, f)
 }


pkg/cmd/kind/load/image-archive/image-archive.go
```diff
@@ -78,19 +78,25 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
 		runtime.GetDefault(logger),
 	)
 
-	// Check if file exists
-	imageTarPath := args[0]
-	if _, err := os.Stat(imageTarPath); err != nil {
-		return err
+	for _, imageTarPath := range args {
+		if _, err := os.Stat(imageTarPath); err != nil {
+			return err
```
aojea, line 83: I don't know if we should expand the command to multiple files, let's say we have
`kind load image file1 file2 file3`  and file2 fails.

The command exits with error , but file1 was already loaded.

I think that is safer to check that only one argument is passed

@BenTheElder WDYT?

harshanarayana: > The command exits with error , but file1 was already loaded.

This can be indicated clearly in the error message though. Or we can add a `--ignore-error` kind of an argument and turn the failed load into a warning and continue to load the rest of the archive. 

Having multiple archive load can be really useful though

BenTheElder: > Having multiple archive load can be really useful though

That's less clear actually, multiple invocations can be made at the same performance and you can instead create a single tarball with all your images at greater efficiency than any of these.

BenTheElder:
> However, the implementation picks the args[0] and ignores the rest without any error or warning. This can lead to a bit of confusion.

this I agree, it should have enforced Nargs = 1

It's less obvious that we should keep adding complexity to these vs fleshing out #2038, and we should actually nudge users to use a single tarball if possible to get better performance in most cases.

pkg/cmd/kind/load/image-archive/image-archive.go
harshanarayana: > a single tarball with all your images at greater efficiency than any of these

This is very much valid. But at the same time if you are loading up a bunch of archives collected from different places (Yes, we do this internally 😢 in our dev workflows),  it is not going to be easy to aggregate them into one common tar ball without doing a few things first. 

So shall I turn this PR to enforce the `Nargs=1` behavior instead then or will that be considered a backward incompatible changes of sort ?

harshanarayana: > multiple invocations can be made at the same performance 

True. But with support for multiple one, you can do `kind load image-archive *.tar --name test`. Would this not be a bit more user friendly ?

harshanarayana:
> this I agree, it should have enforced Nargs = 1

@BenTheElder @aojea Should I redo this PR to enforce this then ?

BenTheElder:
let me circle back to this later when I don't have COVID brain, I shouldn't be on here but I'm restless :+)

defer to @aojea in the meantime, I think it's OK to implement this but I almost think we should throw a warning or something to nudge people in the right direction, the performance using individual tarballs will be disappointing and for most users this is probably something they could avoid with something like `docker save foo bar baz`.

pkg/cmd/kind/load/image-archive/image-archive.go
aojea: I think that first we have to validate that all files exist, and then try to load all ... basically is split this in 2 loops

harshanarayana: @aojea Done. PTAL.

harshanarayana: > I think it's OK to implement this but I almost think we should throw a warning or something to nudge people in the right direction,

@aojea @BenTheElder  Took care of this one too by adding `PreRun` hook.

harshanarayana:
@BenTheElder  @aojea Just a ping to check if this is all good or anything else needs to be taken care of as part of this ?

BenTheElder (commented):
/lgtm
/approve

BenTheElder:
[refreshing github statuses]

BenTheElder:
Sorry, I thought I'd delegated this to antonio, I didn't get anymore pings and it fell off my radar with everything else going on.

I'm making a point to spend more time sweeping through old but not necessarily active open issues/PRs to try to avoid this.

In the future feel free to nudge another time if things sit, it's possible we were unavailable the last time and it got lost in the volume of notifications... sorry

aojea:
heh, 2022 , I totally forget about it, so sorry

PR #2894, opened 2022-08-24
doc: refactor copy to clipboard button

aude:
The background colour on the Copy to clipboard button is driving me nuts! :laughing: 

I constantly think I have marked the COPY buttons, but when I check, I haven't.

The background colour is so similar to the background colour for marked text, that I think it triggers my brain's habit and makes me believe they are marked:
image

I'm suggesting to change the background colour to something else than that important colour, and while I was at it I experimented with using emojis as well, the kind project seems fond of emojis :smile: 
image

What do you think?

### DIFF
--- a/site/assets/css/inline.css
+++ b/site/assets/css/inline.css
@@ -286,8 +286,8 @@ table.includecode thead th {
 }
 
 table.includecode thead th button {
-  background: #326be5;
-  border: .25em solid white;
+  background: transparent;
+  border: 0;
   color: white;
   font-weight: bold;
   margin: .25em;
--- a/site/assets/css/inline.css
+++ b/site/assets/css/inline.css
@@ -288,7 +288,6 @@ table.includecode thead th {
 table.includecode thead th button {
   background: transparent;
   border: 0;
-  color: white;
   font-weight: bold;
   margin: .25em;
   cursor: pointer;
--- a/site/assets/css/inline.css
+++ b/site/assets/css/inline.css
@@ -261,37 +261,33 @@ table.includecode {
   max-width: 100%;
   width: 100%;
   table-layout: fixed;
-  background: #f3f4f4;
-  border: .15em solid black;
+  border: 1px solid rgba(0,0,0,.125);
   border-spacing: 0;
 }
 
-table.includecode thead {
-  background: #333;
-  color: white;
-}
-
-table.includecode thead th a {
-  color: white;
-}
-
 table.includecode thead th {
-  border-bottom: .1em solid black;
+  border-bottom: 1px solid rgba(0,0,0,.125);
   font-weight: normal;
-  font-size: .9em;
+  font-size: 85%;
 }
 
 table.includecode thead th {
   text-align: right;
 }
 
+table.includecode thead th a {
+  font-family: monospace;
+}
+
 table.includecode thead th button {
+  vertical-align: middle;
   background: transparent;
   border: 0;
-  font-weight: bold;
-  margin: .25em;
   cursor: pointer;
-  padding: .1em .5em;
+}
+
+table.includecode tbody {
+  background: #f3f4f4;
 }
 
 table.includecode pre {
--- a/site/layouts/shortcodes/codeFromFile.html
+++ b/site/layouts/shortcodes/codeFromFile.html
@@ -9,8 +9,12 @@
 <table class="includecode" id="code-{{ $virtualPath }}">
   <thead>
     <tr>
-      <th><a href="/{{ $virtualPath }}">{{ $virtualPath }}</a> <button
-          onclick='copyText("code-{{ $virtualPath }}-hidden-copy-text");'>COPY</button></th>
+      <th>
+        <a href="/{{ $virtualPath }}">{{ $virtualPath }}</a>
+        <button onclick='copyText("code-{{ $virtualPath }}-hidden-copy-text");' title="Copy to clipboard">
+          COPY
+        </button>
+      </th>
     </tr>
   </thead>
   <tbody>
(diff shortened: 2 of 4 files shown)


BenTheElder:
Hi, thanks for the PR. Sorry it hasn't received any review yet.

> I constantly think I have marked the COPY buttons, but when I check, I haven't.

Hmm what do you mean by marked?

Tab-selecting adds a border (typical), or hovering changes the pointer.

Do you mean you highlight the text in the copy button? If so, why?
Highlighting the button text would only be useful to copy the text of the button itself ...?

> I'm suggesting to change the background colour to something else than that important colour, and while I was at it I experimented with using emojis as well, the kind project seems fond of emojis 😄

We do like emojis, but we don't use them as a sole representation of anything.

In this case the clipboard emoji can be less readable on the dark background, and is less accessible (not that the current site is nailing this yet 😞). On some systems without an emoji font available we will get a blank character or we will get a simple black design with no color which will be even harder to see.

In terminals we have whatever "standards" exist like TERM=dumb to avoid spamming visual flair, on the site we just don't use them for anything important and sparingly in general. 

> The background colour is so similar to the background colour for marked text, that I think it triggers my brain's habit and makes me believe they are marked:

The background marked color is browser / system dependent, avoiding it is ... challenging.
The button color we have currently is the same color as the link color, I think we can make it more obvious by removing the white border.

The kubernetes.io docs also use blue buttons with white text, though they don't look quite like these, the current design is more rounded and does change color slightly on hover, and for copying code they're using a special SVG icon.

Perhaps we should adopt this icon instead?

aude:
Hi, Ben, thanks for the thorough review!

I started replying to each question first, but then I realized that I think you have understood what I meant, you just are not overly enthusiastic about switching to an emoji. Did I understand that correctly?

Funny thing that you should mention the kubernetes.io docs; that is what I used for reference.

I think the idea of following kubernetes.io it good, I'll make a proposal and then we can see how it turns out.

BenTheElder:
I think that about sums it up 😅 

I think it should be pretty straightforward to swap in the icon they're using with an alt tag, though we might have to alter the layout a bit to get it off of a dark background as well 😅  ... maybe we should just use the codeblock background color for the header with the copy button.

the asthetics of the codeblocks could use a general refresh, I think we can drop the outline etc and replace the copy button with the more familiar icon and not rely on color .

BenTheElder:
Maybe something like this?
Screen Shot 2022-10-05 at 1 37 04 PM

(just trying something like this, edited in the browser directly)

aude:
That looks great! I'll try to implement it, I like everything except the gray whitespace area above the code, maybe the button can "float" over to the right instead of being a header?

aude:
How do you think this looks?
image

I pushed a new commit which makes it look like this.

stmcginnis:
> How do you think this looks?

+1 from me - I think that looks cleaner and visually clear.

BenTheElder:
that looks lovely, thank you!

BenTheElder:
https://deploy-preview-2894--k8s-kind.netlify.app/docs/user/local-registry/ live example

BenTheElder (commented):
what license does copycode.svg have? inlining it here does seem preferable ... unless we're violating the license 😅

aude:
Good point, I got it from https://github.com/kubernetes/website/blob/0088ea5da01fd82995122bcd2a4e26be8bbfe005/static/images/copycode.svg. That repo seems to be under the CC-BY-4.0 license: https://github.com/kubernetes/website/blob/0088ea5da01fd82995122bcd2a4e26be8bbfe005/LICENSE

aude:
Hi, Ben. I took a stab at complying with the [license](https://creativecommons.org/licenses/by/4.0/). It seems all that is needed is to add attribution, whatever that is.

I followed [Creative Commons' guide](https://creativecommons.org/use-remix/attribution/), and I think the attribution is added correctly now.

So I think it's possible to merge the PR without needing any changes now :slightly_smiling_face:

BenTheElder:
sorry it's been a hectic few weeks

github seems to make it hard to view SVG file raw source so using curl:

```console
$ curl -L  https://github.com/kubernetes-sigs/kind/raw/c018fda4bc70ddd4c38c2c7b284fd1ae5f8580ff/site/static/copycode.svg
<!--
copycode.svg by The Kubernetes Authors is licensed under CC BY 4.0

[copycode.svg]: https://github.com/kubernetes/website/blob/0088ea5da01fd82995122bcd2a4e26be8bbfe005/static/images/copycode.svg
[The Kubernetes Authors]: https://github.com/kubernetes
[CC BY 4.0]: https://github.com/kubernetes/website/blob/0088ea5da01fd82995122bcd2a4e26be8bbfe005/LICENSE
-->
<svg xmlns="http://www.w3.org/2000/svg" width="24px" height="24px" viewBox="0 0 24 24" fill="#184eb1">
<path d="M16 1H4c-1.1 0-2 .9-2 2v14h2V3h12V1zm3 4H8c-1.1 0-2 .9-2 2v14c0 1.1.9 2 2 2h11c1.1 0 2-.9 2-2V7c0-1.1-.9-2-2-2zm0 16H8V7h11v14z"/>
</svg
```

Oh!

If this is originally from https://github.com/kubernetes/website/blob/0088ea5da01fd82995122bcd2a4e26be8bbfe005/static/images/copycode.svg, then we're under the *same* license with the *same* authors ("The Kubernetes Authors") for `site/` and we can just copy in the file directly 😅

aude:
No worries. Okay, nice! I reverted the last commit then, so now `copycode.svg` is just a direct copy.

BenTheElder:
/lgtm
/approve
thanks so much! sorry for the delay

Issue #2673, opened 2022-03-14
kind + metallb stopped working last week


darint00:
**What happened**:
kind + metallb installation, per https://kind.sigs.k8s.io/docs/user/loadbalancer/  Stopped working last week.  No longer do we get an external IP address.

**What you expected to happen**:
To get an external IP address.  This was working for months on laptop b.  And for about a year on laptop a.  Doesn't work on either now.

**How to reproduce it (as minimally and precisely as possible)**:
install docker/kind and metallb.

**Anything else we need to know?**:

**Environment:**

- kind version: (use `kind version`): 0.11.1
- Kubernetes version: (use `kubectl version`): 1.21.1
- Docker version: (use `docker info`): 20.10.11
- OS (e.g. from `/etc/os-release`): Ubuntu 20.04

I suspect metallb made some changes that are incompatible with current kind?

aojea:
Can you add more details and logs about the failure?

darint00:
Please tell me what you would like to see.  I'm not versed in
troubleshooting k8s/kind/metallb

I am, on laptop a, installing 0.12.0 and will test out metallb with this.

aojea:
Kind export logs should output the important information to a folder, just zip that and attach it after you reproduce the issue

darint00:
Here is the kind export.

Two other pieces of information.

I noticed the k logs on the speaker pod/metallb showed these
errors/warnings:
grep -e error /tmp/out
{"caller":"level.go:63","error":null,"expected":3,"joined":2,"level":"error","msg":"partial
join","op":"memberDiscovery","ts":"2022-03-14T17:17:49.864534159Z"}
{"caller":"level.go:63","error":"1 error occurred:\n\t* Failed to join
172.18.0.6: dial tcp 172.18.0.6:7946: connect: connection
refused\n\n","expected":1,"joined":0,"level":"error","msg":"partial
join","op":"memberDiscovery","ts":"2022-03-14T17:17:49.864691392Z"}
***@***.***:/$ grep -e warning /tmp/out
W0314 17:17:49.550577       1 warnings.go:70] discovery.k8s.io/v1beta1
EndpointSlice is deprecated in v1.21+, unavailable in v1.25+; use
discovery.k8s.io/v1 EndpointSlice

And I tried with 0.12.0 Which is using k8s 1.23.4 and got the same
error/warings in the speaker pods of metallb.

aojea:
> Here is the kind export.

[...]

PR #2901, opened 2022-09-01
Suggest a more updated metallb version

fedepaol:
The latest metallb version is configurable via CRs instead of configmap.
Here we suggest using the latest version and we change the configuration
from configmap to CRs.


aojea:
we have users that are still using old kubernetes versions and I don't know if this can break them, last time we switched the docs here we had several reported issues https://github.com/kubernetes-sigs/kind/issues/2673

Should we wait a few releases of kubernetes more?

fedepaol:
> we have users that are still using old kubernetes versions and I don't know if this can break them, last time we switched the docs here we had several reported issues #2673
> 
> Should we wait a few releases of kubernetes more?

Actually, https://github.com/kubernetes-sigs/kind/issues/2673 was due to the fact that the doc was pointing at the `main` branch where we introduced the breaking change of switching from configmap to crds.

We validate metallb against the supported version of k8s, currently running CI against 1.21 and 1.24 (because 1.25 kind is not out yet :P ). It _should_ work with older versions, but I can't guarantee.

If we don't do this bump though, metallb manifests will fail because they deploy pod security policy which will be rejected in 1.25

aojea:
> If we don't do this bump though, metallb manifests will fail because they deploy pod security policy which will be rejected in 1.25

/ok-to-test

fedepaol:
Reference issue https://github.com/metallb/metallb/issues/1401

site/content/docs/user/loadbalancer.md
```diff
@@ -26,16 +26,10 @@ description: |-
 
 ## Installing metallb using default manifests
 
-### Create the metallb namespace
```
aojea, line 29: Should we add a small note, explaining this history that "since version x metallb moved to CRDs" ...

fedepaol: will do!

BenTheElder:
thank you!

PR #1548, opened 2020-05-02
install nfs common

BenTheElder:
this is necessary to enable NFS volumes, which should work on current kernels theoretically

PR #2905, opened 2022-09-02
install open-iscsi

fukuta-tatsuya-intec:
It is one of the most famous network storage implement.

#1548 NFS is already supported.

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -115,7 +115,7 @@ RUN echo "Installing Packages ..." \
       conntrack iptables iproute2 ethtool socat util-linux mount ebtables kmod \
       libseccomp2 pigz \
       bash ca-certificates curl rsync \
-      nfs-common fuse-overlayfs \
+      nfs-common fuse-overlayfs open-iscsi \
       jq \
     && find /lib/systemd/system/sysinit.target.wants/ -name "systemd-tmpfiles-setup.service" -delete \
     && rm -f /lib/systemd/system/multi-user.target.wants/* \
--- a/pkg/build/nodeimage/defaults.go
+++ b/pkg/build/nodeimage/defaults.go
@@ -20,4 +20,4 @@ package nodeimage
 const DefaultImage = "kindest/node:latest"
 
 // DefaultBaseImage is the default base image used
-const DefaultBaseImage = "docker.io/kindest/base:v20220811-20da9c7c"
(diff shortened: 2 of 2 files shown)


aojea:
we try to keep the images lean, nfs is the simplest shared storage with multiple of examples on how to use use it for volume provisioning ... 
is there any argument/use case to consider open-iscsi to be installed by default?

fukuta-tatsuya-intec:
Thanks for the comment. I respect the contributor's policy. (try to keep the images lean)
I sent this pr because of iscsi's fast file transfer rate. 
Our test data is 10GB, which is vast. Faster file transfer speed will increase our testing efficiency.

Tip: The open-iscsi installation has a disk impact of about 15 MB.
```bash
# apt install open-iscsi
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
The following additional packages will be installed:
  busybox-initramfs cpio finalrd initramfs-tools-bin initramfs-tools-core klibc-utils kmod libisns0 libklibc libkmod2 libopeniscsiusr netbase udev zstd
Suggested packages:
  libarchive1 bash-completion
The following NEW packages will be installed:
  busybox-initramfs cpio finalrd initramfs-tools-bin initramfs-tools-core klibc-utils kmod libisns0 libklibc libkmod2 libopeniscsiusr netbase open-iscsi udev
  zstd
0 upgraded, 15 newly installed, 0 to remove and 0 not upgraded.
Need to get 3278 kB of archives.
After this operation, 15.6 MB of additional disk space will be used.
```

Your Respectfully

aojea:
> Our test data is 10GB, which is vast. Faster file transfer speed will increase our testing efficiency.

It's hard for us add more dependencies without a general use case, however, you can always create and maintain your own images, just create Dockerfile with the kind image as based and install the openiscsi dependency

```
FROM kindest/node:1.25

RUN echo "Installing Packages ..." \
    && DEBIAN_FRONTEND=noninteractive clean-install \
    open-iscsi  
```

BenTheElder:
I'm not familiar with iscsi but enabling a [built-in](https://kubernetes.io/docs/concepts/storage/volumes/#iscsi) vendor-neutral remote storage type seems reasonable for < 16MB on-disk.

BenTheElder (commented):
/ok-to-test
/lgtm
/approve
rebased + updated base image

BenTheElder:
s390x build flake again
/retest

PR #2910, opened 2022-09-06
check for new podman overlap error message

BenTheElder:
seen in https://kubernetes.slack.com/archives/CEKK1KTN2/p1662384973519729

I think this is a 4.x / netavark thing

### DIFF
--- a/pkg/cluster/internal/providers/podman/network.go
+++ b/pkg/cluster/internal/providers/podman/network.go
@@ -109,9 +109,13 @@ func isUnknownIPv6FlagError(err error) bool {
 
 func isPoolOverlapError(err error) bool {
 	rerr := exec.RunErrorForError(err)
-	return rerr != nil &&
-		(strings.Contains(string(rerr.Output), "is being used by a network interface") ||
-			strings.Contains(string(rerr.Output), "is already being used by a cni configuration"))
+	if rerr == nil {
+		return false
+	}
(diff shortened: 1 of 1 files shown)


BenTheElder:
s390x flake 🙄 
/retest

pkg/cluster/internal/providers/podman/network.go
```diff
@@ -109,9 +109,13 @@ func isUnknownIPv6FlagError(err error) bool {
...
 func isPoolOverlapError(err error) bool {
 	rerr := exec.RunErrorForError(err)
-	return rerr != nil &&
-		(strings.Contains(string(rerr.Output), "is being used by a network interface") ||
-			strings.Contains(string(rerr.Output), "is already being used by a cni configuration"))
+	if rerr == nil {
+		return false
+	}
+	output := string(rerr.Output)
+	return strings.Contains(output, "is already used on the host or by another config") ||
```
rokkbert, line 116: Yes, I saw the matching message with podman 4.1.1:
"Command Output: Error: subnet fc00:f853:ccd:e793::/64 is already used on the host or by another config"

BenTheElder:
@aojea maybe also this one for v0.16?

aojea:
I blindly lgtm since it seems it has been confirmed the error message by another users

/lgtm

PR #2911, opened 2022-09-06
switch to safetext/yamltemplate for yaml templating

BenTheElder:
this is a small additional dependency, but it will prevents yaml injections


BenTheElder:
https://github.com/kubernetes-sigs/kind/runs/8213843373?check_suite_focus=true :D
/hold

BenTheElder:
filing https://github.com/kubernetes-sigs/kind/pull/2912 to reduce the config maintenance overhead before finishing this

BenTheElder:
Rebasing atop #2912, still needs a rework. I think a better strategy is to avoid yaml significant characters in the input, which is currently just the collapse featuregate / runtime config strings. We can use a (safe, no user injection) template method instead.

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -272,8 +282,8 @@ evictionHard:
   nodefs.inodesFree: "0%"
   imagefs.available: "0%"
 {{if .FeatureGates}}featureGates:
-{{ range $key := .SortedFeatureGateKeys }}
-  "{{ $key }}": {{ index $.FeatureGates $key }}
+{{ range $index, $gate := .SortedFeatureGates }}
+  "{{ (StructuralData $gate.Name) }}": {{ $gate.Value }}
```
BenTheElder, line 286: StructuralData is a special function to tell safetext that this is *supposed* to result in a dynamic input-controlled key and that's expected.

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -173,7 +183,7 @@ kubernetesVersion: {{.KubernetesVersion}}
 clusterName: "{{.ClusterName}}"
 {{ if .KubeadmFeatureGates}}featureGates:
 {{ range $key, $value := .KubeadmFeatureGates }}
-  "{{ $key }}": {{ $value }}
+  "{{ (StructuralData $key) }}": {{ $value }}
```
BenTheElder, line 186: StructuralData is a special function to tell safetext that this is *supposed* to result in a dynamic input-controlled key and that's expected.

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -94,8 +95,8 @@ type DerivedConfigData struct {
 	AdvertiseAddress string
 	// DockerStableTag is automatically derived from KubernetesVersion
 	DockerStableTag string
-	// SortedFeatureGateKeys allows us to iterate FeatureGates deterministically
-	SortedFeatureGateKeys []string
+	// SortedFeatureGates allows us to iterate FeatureGates deterministically
+	SortedFeatureGates []FeatureGate
```
BenTheElder, line 99: minor rework, because index is not supported. We didn't need it much other than cross-indexing featuregates by the sorted keys. So instead we just store sorted key + value.

BenTheElder: we could consider sorting both map forms, but currently we don't, no change there.

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -130,13 +136,17 @@ func (c *ConfigData) Derive() {
...
 	// create a sorted key=value,... string of FeatureGates
-	var featureGates []string
+	c.SortedFeatureGates = make([]FeatureGate, 0, len(c.FeatureGates))
+	featureGates := make([]string, 0, len(c.FeatureGates))
 	for _, k := range featureGateKeys {
 		v := c.FeatureGates[k]
 		featureGates = append(featureGates, fmt.Sprintf("%s=%t", k, v))
+		c.SortedFeatureGates = append(c.SortedFeatureGates, FeatureGate{
+			Name:  k,
+			Value: c.FeatureGates[k],
```
aojea, line 148:
```suggestion
			Value: v,
```

aojea:
/lgtm

a nit only for readability

PR #2912, opened 2022-09-07
Drop support for Kubernetes < 1.15

BenTheElder:
Simplifies configuration a bit, and we can consider taking advantage of newer kubeadm functionality soon (e.g. internally using phases), we announced this change in the release notes for kind v0.15.0 and Kubernetes 1.15 is >2 years old, so these releases are pretty old. Upstream Kubernetes only supports back to 1.23

We've also already been soft-limiting to 1.18 by not publishing new images older than Kubernetes 1.18 with KIND v0.13.0+


neolit123 (commented):
/lgtm

BenTheElder:
flake `Pods should run through the lifecycle of Pods and PodStatus` https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/2912/pull-kind-conformance-parallel-ga-only/1567388340699795456

aojea:
that one is a clasic

PR #2913, opened 2022-09-07
don't symlink /dev/console to /dev/kmsg

BenTheElder:
see https://github.com/kubernetes-sigs/kind/issues/662#issuecomment-1238911235

### DIFF
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -366,22 +366,6 @@ fix_product_uuid() {
   fi
 }
 
-fix_kmsg() {
-  # In environments where /dev/kmsg is not available, the kubelet (1.15+) won't
-  # start because it cannot open /dev/kmsg when starting the kmsgparser in the
-  # OOM parser.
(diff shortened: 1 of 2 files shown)


aojea:
is a kernel issue then?

BenTheElder:
> is a kernel issue then?

See the linked discussion for more but TLDR ...

They're not the same and systemd can reject this bogus symlink. lxc removed the option to do this, and the lxc option was the original justification for this being a reasonable solution.

We could instead fake out a file, but I think this is only in environments like LXC where the user can instead ensure the LXC environment has a placeholder.

BenTheElder:
we should do this after https://github.com/kubernetes-sigs/kind/pull/2905

BenTheElder:
not only is this a simplification, the previous change was incorrect.

if we see substantial issues with this, and those environments cannot be fixed, we can consider instead creating a fake file.

but I think this complexity belongs in these unusual, broken environments.

BenTheElder:
one more base image change to pickup before #2933 + v0.16.0 ... containerd 1.6.8

PR #2914, opened 2022-09-07
alphabetize maintainers and drop inactive

BenTheElder:


BenTheElder:
see: #2908

aojea:
:factory_worker:

PR #2918, opened 2022-09-08
increase node ready check timeout to 600 seconds

BodhiHu:
Hi,

When trying to run ```kind create cluster``` on Windows with WSL2 based Docker, it will timeout earlier before the node is ready.

This MR will increase the node ready timeout to 600 seconds, so that ```kind``` could wait enough time for the docker container to print the logs.

### DIFF
--- a/pkg/cluster/internal/providers/docker/provision.go
+++ b/pkg/cluster/internal/providers/docker/provision.go
@@ -401,7 +401,7 @@ func createContainerWithWaitUntilSystemdReachesMultiUserSystem(name string, args
 		return err
 	}
 
-	logCtx, logCancel := context.WithTimeout(context.Background(), 30*time.Second)
+	logCtx, logCancel := context.WithTimeout(context.Background(), 600*time.Second)
 	logCmd := exec.CommandContext(logCtx, "docker", "logs", "-f", name)
 	defer logCancel()
 	return common.WaitUntilLogRegexpMatches(logCtx, logCmd, common.NodeReachedCgroupsReadyRegexp())


neolit123 (commented):
if the timeout actually has to be increased i think 60s is more reasonable.

are you sure the machine has sufficient resources?

BenTheElder:
this timeout is only for the docker image to be running, 600s is an extremely long time and I doubt that Kubernetes will actually run well if starting an image is that slow. this does not include image pulling time.

BenTheElder:
Does it actually work well with this timeout? Have you tested this?

Conversely, increasing the timeout dramatically is going to make it take much longer to surface a failure to users.

We've had WSL2 users for years now without touching this, 10 minutes seems very excessive, I'm finding it hard to imagine the cluster actually functioning after taking this long to create the nodes, some resource must be severely limited?

BodhiHu:
Well... it should be just fine for machines that will not take that longer.
but in case it was ```unfortunately``` that slow (in this case users can go have a cup of tea and come back later) : D

In my case, after increasing the timeout, ```kind create cluster``` did run successfully and up on my computer.

aojea:
I remember one similar issue that was caused because the user has a slow network and the first boot always failed.

Can you check if once it boots the first time the next time to create the cluster it doesn't take so long?

BenTheElder:
We added a timeout recently to the IP address queries in the last release or so.

It should not take this long to start the node container on a host with sufficient resources to run Kubernetes, something is badly hanging and we should fix it.

BodhiHu:
Actually had tried several times, but it just timed out waiting for the logs... perhaps because of slow network.
As commented above, I'm closing this.

Issue #2856, opened 2022-08-02
Ambassador ingress setup documentation no longer valid


DmytryEmery:
**What would you like to be documented**:
Removal or updating of [Ambassdor ingress docs](https://kind.sigs.k8s.io/docs/user/ingress/#ambassador) 

**Why is this needed**:

Following the documentation for [ingress setup for Ambassdor](https://kind.sigs.k8s.io/docs/user/ingress/#ambassador) , it fails with current version of Kind.
```
error: resource mapping not found for name: "ambassadorinstallations.getambassador.io" namespace: "" from "https://github.com/datawire/ambassador-operator/releases/latest/download/ambassador-operator-crds.yaml": no matches for kind "CustomResourceDefinition" in version "apiextensions.k8s.io/v1beta1"
```

Digging around it appears: `apiextensions.k8s.io/v1beta1` is depreciated in new versions of Kubernetes.

As per: https://github.com/datawire/ambassador-operator/issues/73

I was able to get it to work by using an older version of the nodes 
e.g.
`kind create cluster --config .\kind-config.yaml --image kindest/node:v1.21.12​`

This could be mentioned in the docs, or alternately, move to `emissary-ingress` which seems to be the successor project?

BenTheElder:
> Removal or updating of [Ambassdor ingress docs](https://kind.sigs.k8s.io/docs/user/ingress/#ambassador)
> [...]
> This could be mentioned in the docs, or alternately, move to emissary-ingress which seems to be the successor project?

In this case I would lean towards remove, these are contributed and maintained by folks from or interested in the specific ingress typically.

While we've tried to permit all open options folks are willing to contribute, having more of these only has made the guide more confusing, and the kind maintainers do not intend to support N ingresses directly. Within the Kubernetes project we have ingress-nginx which works well enough for this purpose.

Given that nobody else has reported this previously since May when we last released kind and when v1.24.0 Kubernetes came out, it seems interest is minimal and we should probably drop this from the guide.

DmytryEmery:
Seems like a reasonable approach to me, if someone wants to use the others they can add those as needed.  👍

PR #2919, opened 2022-09-09
Remove Ambassador ingress docs

benmoss:
This came up in https://github.com/kubernetes-sigs/kind/issues/2856 but wasn't actually addressed, so here ya go 😄


aojea:
/lgtm
/approve

outdated docs are more problematic that no docs :smile:

BenTheElder:
Thank you!

Issue #2920, opened 2022-09-09
Add support for port ranges in extraPortMappings


bornaivankovic:
**What would you like to be added**:
Add support for port ranges in the extraPortMappings
e.g.
```
# cluster-config.yml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
  kubeadmConfigPatches:
  - |
    kind: ClusterConfiguration
    apiServer:
      extraArgs:
        service-node-port-range: 3000-4000
  extraPortMappings:
  - containerPortRange: 3000-4000
    hostPortRange: 3000-4000
    protocol: TCP
```
**Why is this needed**:
Allows more flexibility when running multiple services of type NodePort and it's not requiring to setup an ingress controller

BenTheElder:
Not completely opposed to this, but this would require a bit more fleshing out:
- for one, this will need to be a new kind API version
- podman and other runtimes need to be validated, port ranges are not something CRI supports and this makes this feature against our aims for portability https://kind.sigs.k8s.io/docs/design/principles/#target-cri-functionality
- what is the user experience for single ports? Are we going to break existing configs, or add awkwardly named new fields or ...


At the moment we're also primarily accepting code and docs fixes and making things more sustainable, @aojea and I are stretched a little thin as the two active maintainers who also maintain portions of Kubernetes, so it may be a bit.

In the meantime:
- on Linux you can just use the container IP and skip port mapping
- On all platforms you could run a sidecar on the kind network and ~socat ports to a node. KINd internally does something similar for multiple control plane nodes currently using haproxy.
- On all platforms you could generate a config with each port 🙈

BenTheElder:
https://github.com/kubernetes-sigs/kind/issues/1219#issuecomment-1242273059

bornaivankovic:
Hi,

Thanks for quick reply. I thought docker was CRI compliant so I was assuming `-p <range>:<range>` mapping was supported by CRI as well. 

The use case I was trying to cover was deploying kind in wsl and accessing deployed workloads from windows host without setting up an ingress. I also wanted to limit the needed configuration changes in my application. As docker doesn't allow to expose ports once container is running and I wanted to use the node port assigned by kubernetes to my service, I though I was being clever by defining the service-node-port-range to the same range exposed by docker to the wsl host. I even tested this with a PoC I made to support kind cluster configuration I posted above.

[...]

PR #2921, opened 2022-09-10
doc: alternative for port mapping or ingress

bornaivankovic:
addresses  [#2920](https://github.com/kubernetes-sigs/kind/issues/2920#issuecomment-1242507245)

### DIFF
--- a/site/content/docs/user/using-wsl2.md
+++ b/site/content/docs/user/using-wsl2.md
@@ -84,6 +84,8 @@ nodes:
 1. create service `kubectl create service nodeport nginx --tcp=80:80 --node-port=30000`
 1. access service `curl localhost:30000`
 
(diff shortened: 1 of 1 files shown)


site/content/docs/user/using-wsl2.md
```diff
@@ -84,6 +84,8 @@ nodes:
 1. create service `kubectl create service nodeport nginx --tcp=80:80 --node-port=30000`
 1. access service `curl localhost:30000`
 
+Alternatively, see [Helpful Tips for WSL2](/docs/user/using-wsl2/#helpful-tips-for-wsl2)
```
BenTheElder, line 87: nit: just use #helpful-tips-for-wsl2 (then we don't have to patch this when the path changes

BenTheElder (commented):
Sorry, I think the others have even more limited availability at the moment.

thank you for the PR. One minor request

BenTheElder:
thanks!

Issue #2924, opened 2022-09-13
Improve output of kind delete cluster


mk46:
**What would you like to be added**:
cluster delete cmd should acknowledge whether a cluster was `deleted successfully` or `cluster not found`
**Why is this needed**:
`kind delete cluster --name <--existing-cluster-->`  and 
`kind delete cluster --name <--not-existing-cluster--> ` are
giving the output as below
`Deleting cluster <clustername>  ...`
which silent the information about non-existing cluster or deleted successfully.

stmcginnis:
This is expected and by design. See first note under https://kind.sigs.k8s.io/docs/user/quick-start/#deleting-a-cluster

Basically you are telling kind that you don't want `foo` to exist, and it returns that it doesn't. Whether it existed beforehand or not does not matter.

If for some reason you need to know whether it exists first, you would need to get clusters to see if it is in the list. There shouldn't be a reason on delete to care.

BenTheElder:
This command is like `rm`.
We could provide more debug output around the specific actions taken but it's meant to be reentrant and safe to call from multiple tear down points in a script (like trapping EXIT).

BenTheElder:
/triage accepted
I think we should probably log if we take any actions ("deleting node xyz", "removing entry from kubeconfig"), or a message about not needing to take any action, to limit the confusion.

oxr463:
@BenTheElder I'd like to take a crack at adding this.

PR #2934, opened 2022-09-20
Improve log output for delete cluster

oxr463:
- [x] Deleting node
- [x] Deleting kubeconfig

Log output:

```
kind delete cluster -n kind
Deleting cluster "kind" ...
Removed context "kind" from kubeconfig
Deleted nodes: ["kind-control-plane"]
```

Closes: https://github.com/kubernetes-sigs/kind/issues/2924

### DIFF
--- a/pkg/cluster/internal/delete/delete.go
+++ b/pkg/cluster/internal/delete/delete.go
@@ -33,17 +33,25 @@ func Cluster(logger log.Logger, p providers.Provider, name, explicitKubeconfigPa
 		return errors.Wrap(err, "error listing nodes")
 	}
 
-	kerr := kubeconfig.Remove(name, explicitKubeconfigPath)
+	l, kerr := kubeconfig.Remove(name, explicitKubeconfigPath)
+	if l {
+		logger.V(0).Infof("Removed context %q from kubeconfig", name)
+	}
 	if kerr != nil {
 		logger.Errorf("failed to update kubeconfig: %v", kerr)
 	}
 
-	err = p.DeleteNodes(n)
-	if err != nil {
-		return err
+	if len(n) > 0 {
+		err = p.DeleteNodes(n)
+		if err != nil {
+			return err
+		}
+		logger.V(0).Infof("Deleted nodes: %q", n)
 	}
+
 	if kerr != nil {
-		return err
+		return kerr
 	}
+
 	return nil
 }
--- a/pkg/cluster/internal/delete/delete.go
+++ b/pkg/cluster/internal/delete/delete.go
@@ -33,10 +33,7 @@ func Cluster(logger log.Logger, p providers.Provider, name, explicitKubeconfigPa
 		return errors.Wrap(err, "error listing nodes")
 	}
 
-	l, kerr := kubeconfig.Remove(name, explicitKubeconfigPath)
-	if l {
-		logger.V(0).Infof("Removed context %q from kubeconfig", name)
-	}
+	kerr := kubeconfig.Remove(name, explicitKubeconfigPath)
 	if kerr != nil {
 		logger.Errorf("failed to update kubeconfig: %v", kerr)
 	}
--- a/pkg/cluster/internal/kubeconfig/internal/kubeconfig/remove.go
+++ b/pkg/cluster/internal/kubeconfig/internal/kubeconfig/remove.go
@@ -24,7 +24,7 @@ import (
 
 // RemoveKIND removes the kind cluster kindClusterName from the KUBECONFIG
 // files at configPaths
-func RemoveKIND(kindClusterName string, explicitPath string) error {
+func RemoveKIND(kindClusterName string, explicitPath string) (bool, error) {
 	// remove kind from each if present
 	for _, configPath := range paths(explicitPath, os.Getenv) {
 		if err := func(configPath string) error {
@@ -52,10 +52,10 @@ func RemoveKIND(kindClusterName string, explicitPath string) error {
 
 			return nil
 		}(configPath); err != nil {
-			return err
+			return false, err
 		}
 	}
-	return nil
+	return true, nil
 }
 
 // remove drops kindClusterName entries from the cfg
--- a/pkg/cluster/internal/kubeconfig/internal/kubeconfig/remove.go
+++ b/pkg/cluster/internal/kubeconfig/internal/kubeconfig/remove.go
@@ -24,7 +24,7 @@ import (
 
 // RemoveKIND removes the kind cluster kindClusterName from the KUBECONFIG
 // files at configPaths
-func RemoveKIND(kindClusterName string, explicitPath string) (bool, error) {
+func RemoveKIND(kindClusterName string, explicitPath string) error {
 	// remove kind from each if present
 	for _, configPath := range paths(explicitPath, os.Getenv) {
 		if err := func(configPath string) error {
@@ -52,10 +52,10 @@ func RemoveKIND(kindClusterName string, explicitPath string) (bool, error) {
 
 			return nil
 		}(configPath); err != nil {
-			return false, err
+			return err
 		}
 	}
-	return true, nil
+	return nil
 }
 
 // remove drops kindClusterName entries from the cfg
--- a/pkg/cluster/internal/kubeconfig/internal/kubeconfig/remove_test.go
+++ b/pkg/cluster/internal/kubeconfig/internal/kubeconfig/remove_test.go
@@ -183,7 +183,7 @@ users:
 	}
 
 	// ensure that we can write this merged config
-	if err := RemoveKIND("foo", existingConfigPath); err != nil {
+	if _, err := RemoveKIND("foo", existingConfigPath); err != nil {
 		t.Fatalf("Failed to remove kind from kubeconfig: %v", err)
 	}
 
@@ -251,7 +251,7 @@ users:
 	}
 
 	// ensure that we can write this merged config
-	if err := RemoveKIND("foo", existingConfigPath); err != nil {
+	if _, err := RemoveKIND("foo", existingConfigPath); err != nil {
 		t.Fatalf("Failed to remove kind from kubeconfig: %v", err)
 	}
 
--- a/pkg/cluster/internal/kubeconfig/internal/kubeconfig/remove_test.go
+++ b/pkg/cluster/internal/kubeconfig/internal/kubeconfig/remove_test.go
@@ -183,7 +183,7 @@ users:
 	}
 
 	// ensure that we can write this merged config
-	if _, err := RemoveKIND("foo", existingConfigPath); err != nil {
+	if err := RemoveKIND("foo", existingConfigPath); err != nil {
 		t.Fatalf("Failed to remove kind from kubeconfig: %v", err)
 	}
 
@@ -251,7 +251,7 @@ users:
 	}
 
 	// ensure that we can write this merged config
-	if _, err := RemoveKIND("foo", existingConfigPath); err != nil {
+	if err := RemoveKIND("foo", existingConfigPath); err != nil {
 		t.Fatalf("Failed to remove kind from kubeconfig: %v", err)
 	}
 
--- a/pkg/cluster/internal/kubeconfig/kubeconfig.go
+++ b/pkg/cluster/internal/kubeconfig/kubeconfig.go
@@ -44,7 +44,7 @@ func Export(p providers.Provider, name, explicitPath string, external bool) erro
 // either explicitPath being set or $KUBECONFIG or $HOME/.kube/config, following
 // the rules set by kubectl
 // clusterName must identify a kind cluster.
-func Remove(clusterName, explicitPath string) error {
+func Remove(clusterName, explicitPath string) (bool, error) {
 	return kubeconfig.RemoveKIND(clusterName, explicitPath)
 }
 
--- a/pkg/cluster/internal/kubeconfig/kubeconfig.go
+++ b/pkg/cluster/internal/kubeconfig/kubeconfig.go
@@ -44,7 +44,7 @@ func Export(p providers.Provider, name, explicitPath string, external bool) erro
 // either explicitPath being set or $KUBECONFIG or $HOME/.kube/config, following
 // the rules set by kubectl
 // clusterName must identify a kind cluster.
-func Remove(clusterName, explicitPath string) (bool, error) {
+func Remove(clusterName, explicitPath string) error {
 	return kubeconfig.RemoveKIND(clusterName, explicitPath)
 }
 


pkg/cmd/kind/delete/cluster/deletecluster.go
```diff
@@ -74,5 +74,6 @@ func deleteCluster(logger log.Logger, flags *flagpole) error {
 	if err := provider.Delete(flags.Name, flags.Kubeconfig); err != nil {
 		return errors.Wrapf(err, "failed to delete cluster %q", flags.Name)
 	}
+	logger.V(0).Infof("Deleted cluster: %q", flags.Name)
```
BenTheElder, line 77: Sorry, but this is not what we discussed adding in #2924.

We should log when we take specific actions like deleting node containers or if the kubeconfig is modified.
This long line doesn't add any additional information versus the log line above on line 73.

BenTheElder (changes requested):

pkg/cmd/kind/delete/cluster/deletecluster.go
oxr463: I've amended my PR. I don't think I can log the deletion of specific nodes unless I modify the `Provider.DeleteNodes()` which doesn't seem to have access to the logging interface.

Reference(s):
- https://github.com/kubernetes-sigs/kind/blob/main/pkg/cluster/internal/providers/provider.go#L41

oxr463:
The error handling doesn't seem to catch deleting non-existing clusters:

```
kind get clusters
No kind clusters found.

kind delete cluster -n non-existent
Deleting cluster "non-existent" ...
Removing context "non-existent" from kubeconfig ...
Deleting nodes for cluster "non-existent" ...
```

pkg/cluster/internal/delete/delete.go
```diff
@@ -33,16 +33,17 @@ func Cluster(logger log.Logger, p providers.Provider, name, explicitKubeconfigPa
 		return errors.Wrap(err, "error listing nodes")
 	}
 
+	logger.V(0).Infof("Removing context %q from kubeconfig ...", name)
```
BenTheElder, line 36: The call below may actually do nothing. We should only log more info when we actually did something.

pkg/cluster/internal/delete/delete.go
```diff
@@ -33,16 +33,17 @@ func Cluster(logger log.Logger, p providers.Provider, name, explicitKubeconfigPa
...
 	kerr := kubeconfig.Remove(name, explicitKubeconfigPath)
 	if kerr != nil {
 		logger.Errorf("failed to update kubeconfig: %v", kerr)
+		return err
```
BenTheElder, line 40: This is wrong

pkg/cluster/internal/delete/delete.go
```diff
@@ -33,16 +33,17 @@ func Cluster(logger log.Logger, p providers.Provider, name, explicitKubeconfigPa
...
 	}
 
+	logger.V(0).Infof("Deleting nodes for cluster %q ...", name)
 	err = p.DeleteNodes(n)
 	if err != nil {
-		return err
-	}
-	if kerr != nil {
+		logger.Errorf("failed to delete nodes: %v", kerr)
```
BenTheElder, line 46: This is redundant to returning the error.

pkg/cluster/internal/delete/delete.go
```diff
@@ -33,16 +33,17 @@ func Cluster(logger log.Logger, p providers.Provider, name, explicitKubeconfigPa
...
 	}
 
+	logger.V(0).Infof("Removing context %q from kubeconfig ...", name)
 	kerr := kubeconfig.Remove(name, explicitKubeconfigPath)
 	if kerr != nil {
 		logger.Errorf("failed to update kubeconfig: %v", kerr)
+		return err
 	}
 
+	logger.V(0).Infof("Deleting nodes for cluster %q ...", name)
```
BenTheElder, line 43: This doesn't tell the user if we actually found any containers and removed them.

BenTheElder:
> The error handling doesn't seem to catch deleting non-existing clusters:

From the past discussion in the issues ... that is working as intended.

there's no such thing as a cluster existing.

There is however node containers associated with a cluster name or kubeconfig entries.

Deleting cluster resources is reentrant, like `rm`, it makes it easier to script reliably. If the command fails in removing nodes that exist it will exit non-zero and log an error. Otherwise it will not and you can have multiple cleanup paths attempt to call it.

oxr463:
```
kind delete clusters -A
Removed context "bar" from kubeconfig
Removed context "foo" from kubeconfig
Deleted clusters: ["bar" "foo"]
```

pkg/cluster/internal/delete/delete.go
oxr463: Removed.

pkg/cluster/internal/delete/delete.go
oxr463: Fixed.

pkg/cluster/internal/delete/delete.go
oxr463: I've moved this into an else clause.

pkg/cluster/internal/delete/delete.go
BenTheElder: still redundant.

pkg/cluster/internal/delete/delete.go
```diff
@@ -36,13 +36,14 @@ func Cluster(logger log.Logger, p providers.Provider, name, explicitKubeconfigPa
 	kerr := kubeconfig.Remove(name, explicitKubeconfigPath)
 	if kerr != nil {
 		logger.Errorf("failed to update kubeconfig: %v", kerr)
+		return kerr
+	} else {
+		logger.V(0).Infof("Removed context %q from kubeconfig", name)
```
BenTheElder, line 41: this will tell the user we removed it from the kubeconfig even if the kubeconfig didn't contain the context. the context name is also not just the cluster name, and the context is not the only thing in the kubeconfig.

pkg/cluster/internal/delete/delete.go
```diff
@@ -36,13 +36,14 @@ func Cluster(logger log.Logger, p providers.Provider, name, explicitKubeconfigPa
...
 	}
 
 	err = p.DeleteNodes(n)
```
BenTheElder, line 44: we know here that we're deleting nodes n, we can log those if there are any.

pkg/cluster/internal/delete/delete.go
oxr463: Used context name instead of cluster name.

oxr463:
```
kind delete clusters -A
Removed context "kind-bar" from kubeconfig
Deleted nodes: ["bar-control-plane"]
Removed context "kind-foo" from kubeconfig
Deleted nodes: ["foo-control-plane"]
Deleted clusters: ["bar" "foo"]
```

pkg/cluster/internal/delete/delete.go
```diff
@@ -35,15 +35,16 @@ func Cluster(logger log.Logger, p providers.Provider, name, explicitKubeconfigPa
 
 	kerr := kubeconfig.Remove(name, explicitKubeconfigPath)
 	if kerr != nil {
-		logger.Errorf("failed to update kubeconfig: %v", kerr)
+		return kerr
+	} else {
+		logger.V(0).Infof("Removed context %q from kubeconfig", kubeconfig.ContextForCluster(name))
```
BenTheElder, line 40: this isn't strictly true, the context may not have been present in the cluster

this probably is most easily accomplished inside Remove

pkg/cluster/internal/delete/delete.go
```diff
@@ -35,15 +35,16 @@ func Cluster(logger log.Logger, p providers.Provider, name, explicitKubeconfigPa
...
 	}
 
 	err = p.DeleteNodes(n)
 	if err != nil {
 		return err
-	}
-	if kerr != nil {
-		return err
+	} else {
+		logger.V(0).Infof("Deleted nodes: %q", n)
```
BenTheElder, line 47: We shouldn't log this when len(nodes) = 0

oxr463: How can I test a cluster with no nodes?

pkg/cluster/internal/delete/delete.go
oxr463: Remove just returns RemoveKIND, which doesn't seem to have access to the logger.

oxr463: @BenTheElder what do you think about modifying RemoveKIND() to add the logger as a parameter?

pkg/cluster/internal/delete/delete.go
oxr463: @BenTheElder I don't see how this would ever trigger.

pkg/cluster/internal/delete/delete.go
oxr463: Ping @BenTheElder

pkg/cluster/internal/delete/delete.go
oxr463: Ping @BenTheElder

BenTheElder: this is a public API, so anyone can call it against this for a cluster name even if there are no nodes

pkg/cluster/internal/delete/delete.go
BenTheElder: alternatively we could return true/false if the cluster is removed or not.

I think it would be reasonable to add the logger as a parameter.

pkg/cluster/internal/delete/delete.go
BenTheElder: specifically by calling https://github.com/kubernetes-sigs/kind/blob/d53225750af998c6218b5c8c730b8b3ca0e2d472/pkg/cluster/provider.go#L186-L188

BenTheElder: we can check len(nodes) and avoid the deleteNodes call and the logging when there are no nodes

oxr463:
@BenTheElder ping 🙂

stmcginnis:
Looks like some tests need to be updated:

```txt
pkg/cluster/internal/kubeconfig/internal/kubeconfig/remove_test.go:186:30: not enough arguments in call to RemoveKIND
	have (string, string)
	want (log.Logger, string, string)
```

oxr463:
@stmcginnis any ideas why there are issues with the CSI driver and volumes? I don't think that's related to my changes.

stmcginnis:
> @stmcginnis any ideas why there are issues with the CSI driver and volumes? I don't think that's related to my changes.

I'm not sure, but I agree that doesn't look like it could be related to your changes. It might just be a transient failure.

/retest

oxr463:
@stmcginnis can you please add the label for tide?

pkg/cluster/internal/delete/delete.go
```diff
@@ -33,17 +33,19 @@ func Cluster(logger log.Logger, p providers.Provider, name, explicitKubeconfigPa
 		return errors.Wrap(err, "error listing nodes")
 	}
 
-	kerr := kubeconfig.Remove(name, explicitKubeconfigPath)
+	kerr := kubeconfig.Remove(logger, name, explicitKubeconfigPath)
 	if kerr != nil {
 		logger.Errorf("failed to update kubeconfig: %v", kerr)
+		return kerr
```
aojea, line 39: @BenTheElder we don't try to delete the nodes anymore?

stmcginnis: That's a good point. While it's not great if kubeconfig can't be updated, we probably shouldn't just error out completely at that point.

I could see either holding on to `kerr` and adding one more check after the `DeleteNodes` block is complete:

```go
if kerr != nil {
    return kerr
}

return nil
```

Or just ignoring it and going with just the logging of the error.

The current code uses the former approach. We should probably stick with that.

oxr463: Is there anything I need to change or is this good enough as is?

If you look through the history of this PR, I had made some changes to this but I ended up putting it back.

stmcginnis: I think I would prefer the change I noted above. I think that's what @aojea was alluding to as well.

oxr463: What about the `return nil` down on line 50?

oxr463: Maybe I don't understand. That's already in place.

stmcginnis: - [ ] Delete line 39
- [ ] Insert the following before the `return nil` at the end:

```go
if kerr != nil {
    return kerr
}
```

oxr463: Alright done. But what about the `return err` on line 44?

stmcginnis: True, otherwise if their is a failure removing the kubeconfig, then the node deletion fails, they never see that. Though then they would still need to fix the issue and clean things up anyway.

Maybe it would be better to move the kubeconfig deletion completely after the node deletion? That way if the kubeconfig deletion fails, at least the cluster itself has already been removed and the user would just need to clean up their kubeconfig.

I think either way is fine. Let's see if others have a stronger opinion on this.

Otherwise looks great - thanks for the updates!

oxr463: @aojea @BenTheElder any updates here?

pkg/cluster/internal/kubeconfig/internal/kubeconfig/remove.go
```diff
@@ -48,6 +49,7 @@ func RemoveKIND(kindClusterName string, explicitPath string) error {
 				if err := write(existing, configPath); err != nil {
 					return err
 				}
+				logger.V(0).Infof("Removed context %q from kubeconfig", kindClusterName)
```
BenTheElder, line 52: We should note this and log it later, we should not log while holding locks, the logging is comparatively much slower than the other operations.

pkg/cluster/internal/kubeconfig/internal/kubeconfig/remove_test.go
```diff
@@ -183,7 +184,7 @@ users:
 	}
 
 	// ensure that we can write this merged config
-	if err := RemoveKIND("foo", existingConfigPath); err != nil {
+	if err := RemoveKIND(cmd.NewLogger(), "foo", existingConfigPath); err != nil {
```
BenTheElder, line 187: We should be testing with a fake. There is a Noop logger in the log package.

BenTheElder (commented):
sorry, most of my energy in this project is still going to bug fixes and maintenance, elsewhere it's been quite a messy time with the kubernetes project almost running out of infra funding and other fun 🙃

pkg/cluster/internal/kubeconfig/internal/kubeconfig/remove.go
oxr463: How should I go about that?

BenTheElder: for what we're logging here currently we could toggle a boolean and then log based on that boolean after the logs are released

stmcginnis:
/ok-to-test
/lgtm

Looks like all comments have been addressed. This looks good to me. Thanks!

pkg/cluster/internal/kubeconfig/internal/kubeconfig/remove.go
```diff
@@ -52,10 +52,10 @@ func RemoveKIND(kindClusterName string, explicitPath string) error {
 
 			return nil
 		}(configPath); err != nil {
-			return err
+			return false, err
```
aojea, line 55: is there any situation that the bool is not false if err == nil ?

pkg/cluster/internal/delete/delete.go
```diff
@@ -33,17 +33,25 @@ func Cluster(logger log.Logger, p providers.Provider, name, explicitKubeconfigPa
 		return errors.Wrap(err, "error listing nodes")
 	}
 
-	kerr := kubeconfig.Remove(name, explicitKubeconfigPath)
+	l, kerr := kubeconfig.Remove(name, explicitKubeconfigPath)
+	if l {
+		logger.V(0).Infof("Removed context %q from kubeconfig", name)
+	}
```
aojea, line 39: do we really need this?

oxr463: This is what was asked for by @BenTheElder

oxr463: I’m fine to take it out if that’s what we want to do.

oxr463: Alright, I removed it.

pkg/cluster/internal/delete/delete.go
```diff
@@ -38,12 +38,17 @@ func Cluster(logger log.Logger, p providers.Provider, name, explicitKubeconfigPa
...
+		err = p.DeleteNodes(n)
+		if err != nil {
+			return err
+		}
+		logger.V(0).Infof("Deleted nodes: %q", n)
 	}
+
 	if kerr != nil {
-		return err
+		return kerr
```
aojea, line 50: heh, you also fixed this bug

BenTheElder:
This is released, please see the note at https://github.com/kubernetes-sigs/kind/releases/tag/v0.18.0#contributors and the "new features section" @oxr463

PR #2935, opened 2022-09-22
Bump base deps

BenTheElder:
containerd 1.6.8
cri-tools 1.25.0


BenTheElder:
Kubernetes e2e suite: [It] [sig-storage] Downward API volume should update annotations on modification [NodeConformance] [Conformance] 

PR #2938, opened 2022-09-23
misc: adding String() to KinD provider interface

yibozhuang:
Providers should be required to implement the fmt.Stringer String() API since the provider string representation is used by the --provider-id in the kubeadmin config.

See: https://github.com/kubernetes-sigs/kind/blob/main/pkg/cluster/internal/create/actions/config/config.go#L71

Otherwise, new providers will see the kubeadm config generated with the full struct representation
of the provider which can be an invalid yaml

Signed-off-by: Yibo Zhuang <yibzhuang@gmail.com>

### DIFF
--- a/pkg/cluster/internal/create/actions/config/config.go
+++ b/pkg/cluster/internal/create/actions/config/config.go
@@ -66,7 +66,7 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 	// create kubeadm init config
 	fns := []func() error{}
 
-	provider := fmt.Sprintf("%s", ctx.Provider)
+	provider := ctx.Provider.String()
 	configData := kubeadm.ConfigData{
 		NodeProvider:         provider,
 		ClusterName:          ctx.Config.Name,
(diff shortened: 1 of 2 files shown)


BenTheElder:
Please see the comments on the current usage.

we don't want more code depending on the string method.

BenTheElder:
Only the first part of the value is only used externally to identify if it is a kind cluster (`kind:///`), the rest is an internal detail and is not stable or standardized or meant to be used by external consumers, and isn't used internally by kind.

The value after the prefix is just something that seemed reasonable for future use, and needs to be per-node unique.

I don't think we should make string-ify-ing part of the general documented API, and String() often represents more than just the type, if we want to formalize this we should probably add .Name() or something.

And even then, this is the *only* use case for it in the code, we should maybe instead require the node provider to provide a node ID API.

BenTheElder:
P.S. thanks for the PR, this certainly is code smell at the moment :-)

BenTheElder:
If you're interested in pursuing this further, please file an issue where we can discuss the approach.

I think this PR is not how we want to handle it, thanks anyhow though!

PR #2940, opened 2022-09-26
doc: readme 0.16

raphaelauv:


BenTheElder:
we also need to do the docs, there's a data file for those

PR #2941, opened 2022-09-26
kindnetd remove wrong routes

chrischdi:
This adds logic to kindnetd to remove routes which do not match the expected gateway address for an (Pod)CIDR.

I'm currently working at
- https://github.com/kubernetes-sigs/cluster-api/issues/3728

Cluster API uses kindnetd as CNI for lots of e2e tests.

As of today, kindnetd only adds new routes and does never delete orphaned routes.

This could lead to the following situation in the workload Cluster created by Cluster API when kindnetd is used as CNI:
* a new Node joins the cluster which:
  * re-uses a PodCIDR of a deleted node
  * but has a different internal IP address compared to the deleted node
* All nodes which already had the route via the old node won't get the new route and pod to pod communication is broken

Broken pod to pod communication may include kube-apiserver to webhook communication, which is the case I detected during my tests.

### DIFF
--- a/images/kindnetd/cmd/kindnetd/routes.go
+++ b/images/kindnetd/cmd/kindnetd/routes.go
@@ -34,15 +34,31 @@ func syncRoute(nodeIP string, podCIDRs []string) error {
 			return err
 		}
 
-		// Check if the route exists to the other node's PodCIDR
+		// Declare the wanted route.
 		routeToDst := netlink.Route{Dst: dst, Gw: ip}
-		route, err := netlink.RouteListFiltered(nl.GetIPFamily(ip), &routeToDst, netlink.RT_FILTER_DST)
+		// List all routes which have the same dst set.
+		// RouteListFiltered ignores the gw for filtering because of the passed filterMask.
+		routes, err := netlink.RouteListFiltered(nl.GetIPFamily(ip), &routeToDst, netlink.RT_FILTER_DST)
 		if err != nil {
 			return err
 		}
 
+		// Check if the wanted route exists and delete wrong routes
+		found := false
+		for _, route := range routes {
+			if routeToDst.Gw.Equal(ip) {
+				found = true
+				continue
+			}
+			// Delete wrong route because of invalid gateway.
+			klog.Infof("Removing invalid route %v\n", route)
+			if err := netlink.RouteDel(&route); err != nil {
+				return err
+			}
+		}
+
 		// Add route if not present
-		if len(route) == 0 {
+		if !found {
 			klog.Infof("Adding route %v \n", routeToDst)
 			if err := netlink.RouteAdd(&routeToDst); err != nil {
 				return err
--- a/pkg/build/nodeimage/const_cni.go
+++ b/pkg/build/nodeimage/const_cni.go
@@ -20,7 +20,7 @@ package nodeimage
 The default CNI manifest and images are our own tiny kindnet
 */
 
-const kindnetdImage = "docker.io/kindest/kindnetd:v20220726-ed811e41"
+const kindnetdImage = "docker.io/kindest/kindnetd:v20220927-ce36d7c0"
 
 var defaultCNIImages = []string{kindnetdImage}
 


neolit123:
i was not aware capi uses kindnet for e2e.
node rotation is out of scope for kind, but for kubeadm e2e (also uses kindnet) we wanted to eventually test node addition / removal mainly around stacked etcd sustainability.

in any case, unclear if this is desired.

/cc @aojea

images/kindnetd/cmd/kindnetd/routes.go
```diff
@@ -24,30 +24,49 @@ import (
...
+	// List all routes to the other node's PodCIDR
+	routeToDst := netlink.Route{Dst: dst, Gw: gw}
+	routes, err := netlink.RouteListFiltered(nl.GetIPFamily(gw), &routeToDst, netlink.RT_FILTER_DST)
+	if err != nil {
+		return err
+	}
+
+	// Check if the route exists to the other node's PodCIDR and delete invalid routes
+	for _, route := range routes {
+		// Skip creation of a route if there is gateway matches the expected.
```
neolit123, line 53:
```suggestion
		// Skip creation of a route if there is gateway which matches the expected.
```
?

images/kindnetd/cmd/kindnetd/routes.go
```diff
@@ -24,30 +24,49 @@ import (
 	"k8s.io/klog/v2"
 )
 
-func syncRoute(nodeIP string, podCIDRs []string) error {
+func syncRoutes(nodeIP string, podCIDRs []string) error {
```
aojea, line 27: why don't we modify `syncRoute`only?

aojea: ah, ok, nevermind,

images/kindnetd/cmd/kindnetd/routes.go
```diff
@@ -24,30 +24,49 @@ import (
...
+func syncRoute(podCIDR string, gw net.IP) error {
+	// parse subnet
+	dst, err := netlink.ParseIPNet(podCIDR)
+	if err != nil {
+		return err
+	}
+
+	// List all routes to the other node's PodCIDR
+	routeToDst := netlink.Route{Dst: dst, Gw: gw}
+	routes, err := netlink.RouteListFiltered(nl.GetIPFamily(gw), &routeToDst, netlink.RT_FILTER_DST)
```
aojea, line 46: does this list the default route to? the 0.0.0.0 ?

aojea: puff, I'm asleep today, forget the comment, this is the same as before

aojea: hmm, but this will only list the routes that match `dst`and `gw`, you want to obtain only the routes that match destination and all the gateways

chrischdi: The `RouteListFiltered` is called with the mask `netlink.RT_FILTER_DST` which results in returning all routes which have the same `dst` as the given `routeToDst`. All other values of the Route are ignored during filtering.

aojea: I see, so it was wrong before

chrischdi: Example:

Consider the following active routing information:
```
$ ip -4 r
default via 172.18.0.1 dev eth0
172.18.0.0/16 dev eth0 proto kernel scope link src 172.18.0.6
192.168.0.0/24 via 172.18.0.7 dev eth0
192.168.1.0/24 via 172.18.0.5 dev eth0
192.168.3.0/24 via 172.18.0.7 dev eth0
192.168.4.0/24 via 172.18.0.9 dev eth0
192.168.5.0/24 via 172.18.0.5 dev eth0
192.168.6.0/24 via 172.18.0.4 dev eth0
192.168.8.0/24 via 172.18.0.9 dev eth0
```

* `syncRoutes("172.18.0.9", []string{"192.168.6.0/24"})`:
  * `netlink.RouteListFiltered` returns the following element: `192.168.6.0/24 via 172.18.0.4 dev eth0`
  * old implementation: would have skipped adding the new route which should look like `192.168.6.0/24 via 172.18.0.9 dev eth0`, if it even tried adding a new route, it would not have worked because there is already a route to `192.168.6.0/24` (error: `RTNETLINK answers: File exists`)
  * The new implementation removes `192.168.6.0/24 via 172.18.0.4 dev eth0` first and adds the new route instead

As visible in the example output above: there may still be orphaned routes, but they get recycled if the gateway IP changes.

images/kindnetd/cmd/kindnetd/routes.go
```diff
@@ -24,30 +24,49 @@ import (
...
+	if err != nil {
+		return err
+	}
+
+	// Check if the route exists to the other node's PodCIDR and delete invalid routes
+	for _, route := range routes {
+		// Skip creation of a route if there is gateway which matches the expected.
```
aojea, line 53: I think we should just continue, i.e., let's say the route is 10.0.0.0/8 and the new gw 192.168.1.1, if we have
```
10.0.0.0/8 via 192.168.1.1
10.0.0.0/8 via 192.168.1.5
```
despite we have the right route we want to delete the other route too

images/kindnetd/cmd/kindnetd/routes.go
```diff
@@ -24,30 +24,49 @@ import (
...
+		if route.Gw.Equal(gw) {
+			return nil
```
aojea, line 55: https://github.com/kubernetes-sigs/kind/pull/2941/files#r980086210

```suggestion
			continue
```

images/kindnetd/cmd/kindnetd/routes.go
```diff
@@ -24,30 +24,49 @@ import (
...
-			return err
+func syncRoute(podCIDR string, gw net.IP) error {
+	// parse subnet
+	dst, err := netlink.ParseIPNet(podCIDR)
+	if err != nil {
+		return err
+	}
+
+	// List all routes to the other node's PodCIDR
+	routeToDst := netlink.Route{Dst: dst, Gw: gw}
```
aojea, line 45: should we omit my previous mistake of using the gw?

images/kindnetd/cmd/kindnetd/routes.go
aojea: hmm, we have to add some boolean to indicate that we found it, to avoid to create it after the loop

images/kindnetd/cmd/kindnetd/routes.go
chrischdi: you're right 👍

images/kindnetd/cmd/kindnetd/routes.go
aojea: yes, yes, I'm a bit slow today, I see what you mean and I agree this reconciles it correctly, is just we never care about the case routes changed 🙃

images/kindnetd/cmd/kindnetd/routes.go
chrischdi: I think we can make the func easier.

We can expect to only have one route to our dst right?

So instead of using a for loop here, we could check for the length and have the following 4 cases:

* `len(routes) > 1`: --> error / unexpected
* `len(routes) == 1 && equal`: --> return nil
* `len(routes) == 1 && not equal`: --> remove and add route
* `len(routes) == 0`: add route

chrischdi: Another solution:
 
 ave two calls to `netlink.RouteListFiltered`
 * the first to check if a route exists which matches gateway and dst: if so we can return early
 * the second to get the list of routes which only match dst: to remove them all
 * as third step add the new route

images/kindnetd/cmd/kindnetd/routes.go
chrischdi: The `routeToDst` is still used for creating the route, but I'm ok with dropping the gw here and inline the definition below (where gw is included).

images/kindnetd/cmd/kindnetd/routes.go
aojea: the loop is not expensive and is easier to read, specially for people not used to networks. Also, we can solve the "routes > 1" , it is very cheap 

```
// myroute is the route we want to install
found := false
for _, r := routes {
   if r == myroute {
        found = true
   } else {
       delete(r)
   }
}
if !found {
  createRoute(myroute)
}
```

kindnet is a "toy" CNI, I prefer to keep it simple so "new" people to CNIs can

chrischdi:
Thank you very much for your review and guidance @aojea 👍 

I also inlined the func again because as you outlined correctly, we can't just do an early return.

images/kindnetd/cmd/kindnetd/routes.go
chrischdi: Updated the comment for `RouteListFiltered` and `routeToDst`

images/kindnetd/cmd/kindnetd/routes.go
chrischdi: Updated the parts to use continue and the boolean variable 👍

images/kindnetd/cmd/kindnetd/routes.go
```diff
@@ -34,15 +34,30 @@ func syncRoute(nodeIP string, podCIDRs []string) error {
...
+		routes, err := netlink.RouteListFiltered(nl.GetIPFamily(ip), &routeToDst, netlink.RT_FILTER_DST)
 		if err != nil {
 			return err
 		}
 
+		// Check if the wanted route exists and delete wrong routes
+		found := false
+		for _, route := range routes {
+			if routeToDst.Gw.Equal(ip) {
+				continue
```
aojea, line 50:
```suggestion
                                 found = true
				continue
```

chrischdi: 🙈

aojea:
/lgtm
/hold

because we have to build the image to run it in the CI, this code is not being exercised right now on the CI

@BenTheElder can you PTAL

BenTheElder (commented):
kind is generally -1 on adding code paths to support things we don't support (a tautology if you will!) but this change is small and seems reasonable and useful for multiple projects in-org.

+1 to antonio's review thus far :-)
/approve

BenTheElder:
/assign
rebasing, bumping kindneted

BenTheElder:
OK, all of the prow CI jobs (not github actions) should run with this now.

aojea:
/lgtm
/hold cancel
Thanks

PR #2946, opened 2022-09-29
kindnetd fix removal of wrong routes

chrischdi:
Hi, I'm sorry to disturb again on this topic, but I introduced a small bug in #2941 .

This PR is supposed to fix that.

The current implementation does not compare the node IP with the existing route, but instead with the wanted route which always results in true and never in removal of an orphaned ip route.

cc @BenTheElder @aojea


aojea:
damn, I should catch it during the review, sorry, it is indeed very subtle

/lgtm
/approve

aojea:
/hold

we have to rebuild the image

chrischdi:
Should be unrelated:

/test pull-kind-e2e-kubernetes-1-23

Issue #2875, opened 2022-08-10
containerd deprecated CRI hosts config


BenTheElder:
This is unfortunate, it will be much harder for users to configure things like local registry as you cannot configure it inline in containerd config anymore. Also, enabling the new non-deprecated configuration route will disable the old one AFAICT https://github.com/containerd/containerd/blob/main/docs/hosts.md#cri

You now need to create a directory with a host.toml file for each host, I'm also not sure if this means wildcard hosts are no longer supported.

see also: https://github.com/kubernetes/kubernetes/issues/110312

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/3233 Is out to give us a chance to pivot the default but it will be some time before user configs have this.

I think we're going to need a PSA in the next round of release notes

BenTheElder:
This is going to be a bit kludge without more functionality. I guess we can eventually flip on the config path and then tell users to extraMounts a directory with the host configs but that's not amazing and it's more error prone than the existing approach (commonly users don't have the right mount sharing)

BenTheElder:
https://github.com/kubernetes-sigs/kind/releases/tag/v0.20.0 has a PSA about this

BenTheElder:
This should've been resolved in v0.27 with the switch to containerd 2.x, along with many releases of warning users and having updated the sample scripts to handle this gracefully.

https://github.com/kubernetes-sigs/kind/releases#:~:text=Compare-,v0.27.0,-This%20release%20moves 

However, we need to follow up on that https://github.com/kubernetes-sigs/kind/issues/3918#issuecomment-2826311748

cc @stmcginnis @aojea 


We might in fact need https://github.com/kubernetes-sigs/kind/pull/3601

PR #2952, opened 2022-10-04
migrate containerd registry endpoint to hosts.toml

matthyx:
see https://github.com/kubernetes/kubernetes/issues/110312

### DIFF
--- a/images/base/files/etc/containerd/certs.d/hosts.toml
+++ b/images/base/files/etc/containerd/certs.d/hosts.toml
@@ -1,0 +1,3 @@
+# Enable registry.k8s.io as the primary mirror for k8s.gcr.io
+# See: https://github.com/kubernetes/k8s.io/issues/3411
+server = ["https://registry.k8s.io", "https://k8s.gcr.io"]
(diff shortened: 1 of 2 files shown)


BenTheElder:
Were it so easy https://kind.sigs.k8s.io/docs/user/local-registry/

https://github.com/kubernetes-sigs/kind/issues/2875

aojea:
🍿 https://github.com/kubernetes/kubernetes/issues/110312#issuecomment-1265470633

it seems this will take some time

BenTheElder:
I think we will probably have to design a KIND registry config format to shim this, so we can backend it with either containerd inline or the hosts dir.

When doing so we also need to consider the longstanding feature request for a built-in registry user experience.

matthyx:
do you have a KinDEP process ;-) ?

aojea:
> do you have a KinDEP process ;-) ?

there was some try IIRC, I think that Ben also opened an issue about that

BenTheElder:
https://kind.sigs.k8s.io/docs/contributing/getting-started/ points to discussing designs in an issue before moving to implementation. You can see some examples of issues outlining designs.

BenTheElder:
we just removed this config from the base image, we still need a good plan for how users can reasonably switch over with their custom registry configs. we have other much more pressing issues at the moment though.

BenTheElder:
#3122

matthyx:
Yeah, I let that rot a bit... thanks Ben.

BenTheElder:
thanks for the PR, this is going to be a mess to finish sorting out 😅

PR #2955, opened 2022-10-05
Fix load docker-image re-tagging with prefix

Vlatombe:
When loading a docker image that has already been loaded with a new tag, its name is not sanitized on import, leading to an inconsistent tag present in the kind container.

Reproducer with latest kind release (v0.16.0)

```
$ docker pull maven:3-jdk-11
$ docker tag maven:3-jdk-11 maven:test
```
Import the first tag
```
$ kind --name $cluster load docker-image maven:3-jdk-11
Image: "maven:3-jdk-11" with ID "sha256:a016b74d0cb2fdf1adcc87a85d6eb6fc4c299bf550def4d2b752b10f6f1cdbf1" not yet present on node "$cluster-control-plane", loading...
```
Then import the second tag
```
$ kind --name $cluster load docker-image maven:test
Image with ID: sha256:a016b74d0cb2fdf1adcc87a85d6eb6fc4c299bf550def4d2b752b10f6f1cdbf1 already present on the node $cluster-control-plane but is missing the tag maven:test. re-tagging...
```

Inspecting content of the internal registry shows that the new image hasn't been prefixed with `docker.io/library` as it was done for the first load.

```
$ docker exec $cluster-control-plane ctr -n k8s.io images list | grep maven
docker.io/library/maven:3-jdk-11                                                                                           application/vnd.docker.distribution.manifest.v2+json      sha256:89b095d4e90aaf193a782e16b53d0ba268ef75752a943b17aa346b632894c8fc 636.0 MiB linux/arm64                                                                  io.cri-containerd.image=managed
maven:test                                                                                                                 application/vnd.docker.distribution.manifest.v2+json      sha256:89b095d4e90aaf193a782e16b53d0ba268ef75752a943b17aa346b632894c8fc 636.0 MiB linux/arm64                                                                  io.cri-containerd.image=managed
```

### DIFF
--- a/pkg/cmd/kind/load/docker-image/docker-image.go
+++ b/pkg/cmd/kind/load/docker-image/docker-image.go
@@ -136,7 +136,7 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
 		imageID := imageIDs[i]
 		processed := false
 		for _, node := range candidateNodes {
-			exists, reTagRequired := checkIfImageReTagRequired(node, imageID, imageName, nodeutils.ImageTags)
+			exists, reTagRequired, imageName := checkIfImageReTagRequired(node, imageID, imageName, nodeutils.ImageTags)
 			if exists && !reTagRequired {
 				continue
 			}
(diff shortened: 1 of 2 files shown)


jglick:
Failed workaround:

```console
$ kind load docker-image docker.io/library/maven:test
Image: "docker.io/library/maven:test" with ID "sha256:…" found to be already present on all nodes.
```

jglick (approved):
Looks right (untested).

BenTheElder (commented):
/lgtm
/approve
thanks!

BenTheElder:
flake:
> Kubernetes e2e suite: [It] [sig-apps] StatefulSet Basic StatefulSet functionality [StatefulSetBasic] should provide basic identity 

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/2955/pull-kind-e2e-kubernetes/1578110956825743360
/retest

pkg/cmd/kind/load/docker-image/docker-image.go
```diff
@@ -241,15 +241,15 @@ func removeDuplicates(slice []string) []string {
...
+func checkIfImageReTagRequired(node nodes.Node, imageID, imageName string, tagFetcher imageTagFetcher) (exists, reTagRequired bool, sanitizedImage string) {
 	tags, err := tagFetcher(node, imageID)
 	if len(tags) == 0 || err != nil {
 		exists = false
 		return
 	}
 	exists = true
-	imageName = sanitizeImage(imageName)
-	if ok := tags[imageName]; ok {
+	sanitizedImage = sanitizeImage(imageName)
```
aojea, line 251: I was wondering if we should always sanitize the images names?

Issue #2770, opened 2022-05-14
Document installing linux docker (not docker desktop) on WSL 2


jankoprowski:
**What would you like to be documented**:
WSL 2 is not nesesery need Docker Desktop, which is not free.
You could document setup with https://docs.docker.com/engine/install/ubuntu/
After installing docker you just need add your user to `docker` group, restart wsl and run command `service docker start` 

**Why is this needed**:
Docker Desktop is not free - it would be nice to know you have an free alternative under WSL 2

BenTheElder:
PRs welcome, the WSL2 docs are maintained through community contributions, we don't develop on windows and don't have CI #1529 
/help
/good-first-issue

Aryan-Deshpande:
Hey, I would like to work on this issue. Just to clear up, documentation needs to be made to install docker/ setup docker on wsl2 right?.

BenTheElder:
Yes, the request is to document using kind without using docker desktop, on WSL2.

I think we should keep this relatively short / lightweight, "how to install docker" is not really in scope for this project, but we should document that this is an option with WSL2.

rakshitgondwal:
Hey @BenTheElder !!
I would like to work on this issue.

hrittikhere:
Do we need to expand this section? CC: @BenTheElder 
I can touch on that! 

 <img width="1346" alt="image" src="https://user-images.githubusercontent.com/67012359/193919146-de413cd7-1051-438d-98a2-f01781649404.png">

BenTheElder:
Yes, please do!

hrittikhere:
Hey @BenTheElder 
Found this article: https://dev.solita.fi/2021/12/21/docker-on-wsl2-without-docker-desktop.html and I think it will be a good reference point. Do you feel going with Debian & apt will be okay for the first PR? Please advice

BenTheElder:
I think we should prefer to document only that it is an option and link to external documentation.

We should not personally maintain "how to install docker on WSL2" ourselves, this is duplicated effort and Kubernetes SIG Docs prohibits this sort of thing for https://kubernetes.io

hrittikhere:
Thanks @BenTheElder for the information! 

Something like:

## Setting up Docker in WSL2 without Docker Desktop 

- 2/3 Links for external blogs


Does this look good with a few more lines of information?

PR #2966, opened 2022-10-12
docs: Add Docker in WSL2 without Docker Desktop

hrittikhere:
fixes: #2770 

CC: @BenTheElder can you please review?

### DIFF
--- a/site/content/docs/user/using-wsl2.md
+++ b/site/content/docs/user/using-wsl2.md
@@ -58,9 +58,15 @@ wsl --set-default-version 2
 1. Now, you can install your Linux distro of choice by searching the Windows Store. If you don't want to use the Windows Store, then follow the steps in the WSL docs for [manual install](https://docs.microsoft.com/en-us/windows/wsl/install-manual).
 1. Start up your distro with the shortcut added to the start menu
 
-## Setting up Docker in WSL2
+## Setting up Docker in WSL2 with Docker Desktop
 
 Install Docker with WSL2 backend here: https://docs.docker.com/docker-for-windows/wsl/
+
+
+## Setting up Docker in WSL2 without Docker Desktop
+
+An alternative approach that does not require the use of the heavy Docker Desktop. The installation procedure can be found at https://dev.to/bowmanjd/install-docker-on-windows-wsl-without-docker-desktop-34m9
+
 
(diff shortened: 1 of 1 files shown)


site/content/docs/user/using-wsl2.md
```diff
@@ -58,10 +58,16 @@ wsl --set-default-version 2
...
 
-## Setting up Docker in WSL2
+## Setting up Docker in WSL2 with Docker Desktop
 
 Install Docker with WSL2 backend here: https://docs.docker.com/docker-for-windows/wsl/
 
+
+## Setting up Docker in WSL2 without Docker Desktop
+
+An alternative approach that does not require the use of the heavy Docker Desktop. The installation procedure can be found at https://dev.to/bowmanjd/install-docker-on-windows-wsl-without-docker-desktop-34m9
```
BenTheElder, line 68: I'm not sure we should make judgements on the tool like "heavy", the kind project isn't favoring one over the other. (We *do* currently favor _Docker_ at all over podman because podman is less stable - as in it makes breaking major version changes etc - and as a result our support is as well)

hrittikhere: I actually was comparing Docker Desktop vs Engine when I meant heavy. Do you want me to rephrase or remove `too heavy ` for clarity?

BenTheElder: let's drop the heavy comment, it's not necessary. users can determine which works best for them, we're just noting kind will work with these.

BenTheElder:
```suggestion
Alternatively, docker can be installed in WSL2 without using Docker Desktop
See for example: https://dev.to/bowmanjd/install-docker-on-windows-wsl-without-docker-desktop-34m9
```

Though, this guide may be a confusing example as it also suggests installing podman, I'm not sure podman will work with kind in WSL2.

hrittikhere: Do you want me to add few more posts or find a newer better one? The posts work though if we neglect the podman section

BenTheElder (commented):
/lgtm
/approve
thank you!

hrittikhere:
Welcome Benjamin!

PR #2968, opened 2022-10-15
version pre release is always true

converge:
Signed-off-by: João Vanzuita <joaovanzuita@me.com>

### DIFF
--- a/pkg/cmd/kind/version/version.go
+++ b/pkg/cmd/kind/version/version.go
@@ -31,18 +31,16 @@ import (
 func Version() string {
 	v := versionCore
 	// add pre-release version info if we have it
-	if versionPreRelease != "" {
-		v += "-" + versionPreRelease
-		// If gitCommitCount was set, add to the pre-release version
-		if gitCommitCount != "" {
-			v += "." + gitCommitCount
-		}
-		// if commit was set, add the + <build>
-		// we only do this for pre-release versions
-		if gitCommit != "" {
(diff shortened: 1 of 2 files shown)


aojea:
Is it?

converge:
> Is it?

this is my very first PR to the project, pls correct me if I'm wrong. 

the condition checks if the constant value is != than "", the constant cannot be changed via LDFLAGS since only "var" are able to be updated during build time. 

reference: https://github.com/kubernetes-sigs/kind/blob/main/pkg/cmd/kind/version/version.go#L61

aojea:
yeah, it does, @BenTheElder knows how this works better and he may expand on this, but the constant is modified manually for the releases,. see https://github.com/kubernetes-sigs/kind/commit/d39dfa08bf2aba2a0e4f3fb51fa715147d299b84

converge:
> yeah, it does, @BenTheElder knows how this works better and he may expand on this, but the constant is modified manually for the releases,. see [d39dfa0](https://github.com/kubernetes-sigs/kind/commit/d39dfa08bf2aba2a0e4f3fb51fa715147d299b84)

I see.. there is also a script to update the version https://github.com/kubernetes-sigs/kind/blob/9ad13414051cf9f7b731f7213c456815b6ed3ade/hack/release/create.sh#L49

I'll close this PR.

BenTheElder:
It's done this way so that "go get" produces the right thing with release tags at least.

PR #2969, opened 2022-10-16
Add unit tests for version command

converge:
This PR adds unit tests for the version command.

Signed-off-by: João Vanzuita <joaovanzuita@me.com>

### DIFF
--- a/pkg/cmd/kind/version/version_test.go
+++ b/pkg/cmd/kind/version/version_test.go
@@ -61,7 +61,7 @@ func TestTruncate(t *testing.T) {
 	}
 }
 
-func TestVersion_Success(t *testing.T) {
+func TestVersion(t *testing.T) {
 	tests := []struct {
 		name           string
 		gitCommit      string
--- a/pkg/cmd/kind/version/version_test.go
+++ b/pkg/cmd/kind/version/version_test.go
@@ -61,65 +61,58 @@ func TestTruncate(t *testing.T) {
 	}
 }
 
-func TestVersion_SuccessWithGitCommitCountAndHash(t *testing.T) {
-
-	if versionPreRelease != "" {
-		gitCommitBackup := gitCommit
-		gitCommitCountBackup := gitCommitCount
-		defer func() {
-			gitCommit = gitCommitBackup
-			gitCommitCount = gitCommitCountBackup
-		}()
-		gitCommit = "mocked-hash"
-		gitCommitCount = "mocked-count"
-		want := versionCore + "-" + versionPreRelease + "." + gitCommitCount + "+" + gitCommit
-
-		got := Version()
-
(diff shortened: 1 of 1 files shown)


pkg/cmd/kind/version/version_test.go
```diff
@@ -60,3 +60,66 @@ func TestTruncate(t *testing.T) {
 		})
 	}
 }
+
+func TestVersion_SuccessWithGitCommitCountAndHash(t *testing.T) {
```
BenTheElder, line 64: If we're going to mutate these globals anyhow, we might as well use fixed inputs and assert the output and just do it in one Test function over each case

converge: > If we're going to mutate these globals anyhow, we might as well use fixed inputs and assert the output and just do it in one Test function over each case

I just pushed a commit, is that what you mean?

pkg/cmd/kind/version/version_test.go
```diff
@@ -60,3 +60,59 @@ func TestTruncate(t *testing.T) {
 		})
 	}
 }
+
+func TestVersion_Success(t *testing.T) {
```
BenTheElder, line 64: I would call this `TestVersion` since that's what it does and there is no error path, but I'm not going to block the PR on that

pkg/cmd/kind/version/version_test.go
```diff
@@ -60,3 +60,59 @@ func TestTruncate(t *testing.T) {
...
+		name           string
+		gitCommit      string
+		gitCommitCount string
+		want           string
+	}{
+		{
+			name:           "With git commit count and with commit hash",
+			gitCommit:      "mocked-hash",
+			gitCommitCount: "mocked-count",
+			want:           versionCore + "-" + versionPreRelease + "." + "mocked-count" + "+" + "mocked-hash",
```
BenTheElder, line 75: IMO we should fully control the inputs and have a specific, static expected output (IE a hardcoded string, not a computed result), otherwise we're just writing the logic twice. That can be a followup.

BenTheElder (commented):
/approve

pkg/cmd/kind/version/version_test.go
converge: updated

BenTheElder:
PPC64LE flake again
/retest

converge:
hey @BenTheElder , shall we retest it?

BenTheElder:
sorry, i'm at kubecon this week and i'm meeting with folks to try to keep the kubernetes project funded and online at the moment https://kubernetes.slack.com/archives/CCK68P2Q2/p1665109044532949

BenTheElder (commented):
/lgtm
/approve

Issue #2872, opened 2022-08-09
Running with rootless podman doesn't work as documented


cwrau:
**What happened**: I tried running kind with rootless podman and followed the [documentation](https://kind.sigs.k8s.io/docs/user/rootless/). But this didn't work

**What you expected to happen**: That it would work

**How to reproduce it (as minimally and precisely as possible)**: Install and configure podman for rootless, install kind. Use a terminal with systemd-scopes, like gnome-terminal. Use an OS that doesn't `Delegate` everything, like Arch Linux. (Seems to be done on Fedora, https://gitlab.gnome.org/GNOME/gnome-terminal/-/issues/7914#note_1523590)

**Anything else we need to know?**: 

**Environment:**

- kind version: (use `kind version`):
- Kubernetes version: (use `kubectl version`):
- Docker version: (use `docker info`):
- OS (e.g. from `/etc/os-release`):

BenTheElder:
> Docker version: (use docker info): command not found 😉

This is not helpful …
podman attempts to be docker compatible and contains docker sub commands like podman info. This command is rich with host environment debug info.


> I wonder if the kind authors just never got the message that units are now put into app.slice by default, and delegation need to be also enabled in intermediate slices. And if they are e.g. on Fedora, that is done by default ...

rootless podman support is contributed by @AkihiroSuda, nominally kind is developed for rootful docker.

Akihiro contributed some fedora based CI for rootless docker and podman, which does happen to be fedora based.

This is the first we’ve encountered a rootless podman user that wasn’t on fedora, as you can tell. Most of our users use docker, which has mature support. Podman support is experimental (the tool should be printing a warning when you use podman) and is fundamentally limited by some of the significant differences where it is not drop in compatible, we have entirely separate code paths for podman behavior.


> No idea what exactly your problem is. But if something wants to use cgroups, it really should run in its own systemd unit and enable delegation by setting the appropriate options there. As you know, that is possible to do by writing the .service file or using systemd-run --user.

The kind process *doesn’t touch cgroups*.
Our tool invokes docker or podman to spawn a container, inside the container we do touch cgroups.

BenTheElder:
Host cgroups configuration is not something we currently plan to touch from the kind process, for example podman we’re invoking may actually be talking to a remote instance anyhow and it’s difficult to detect reliably (see discussions linked to #2233)

cwrau:
> > Docker version: (use docker info): command not found 😉
> 
[...]

PR #2981, opened 2022-10-27
Document rootless podman quirks on some distributions

VannTen:
This should resolve the documentation issues in #2872

It apparently works without adding `--property=Delegate=yes` to the systemd-run.

### DIFF
--- a/site/content/docs/user/rootless.md
+++ b/site/content/docs/user/rootless.md
@@ -67,5 +67,10 @@ To create a kind cluster with Rootless Podman, just run:
 $ KIND_EXPERIMENTAL_PROVIDER=podman kind create cluster
 ```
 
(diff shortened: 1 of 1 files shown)


BenTheElder (commented):
/lgtm
/approve
/ok-to-test
thanks!

site/content/docs/user/rootless.md
```diff
@@ -67,5 +67,10 @@ To create a kind cluster with Rootless Podman, just run:
 $ KIND_EXPERIMENTAL_PROVIDER=podman kind create cluster
 ```
 
+On some distributions, you might need to use systemd-run to start kind into its own cgroup scope:
```
AkihiroSuda, line 70: What distribution?

VannTen: At least Archlinux. Might be related to cgroupsv2/unified hierarchy

Issue #2987, opened 2022-10-31
[zfs] kind create cluster crashes


jrwren:
<!-- Please use this template while reporting a bug and provide as much info as possible. Not doing so may result in your bug not being addressed in a timely manner. Thanks!-->


**What happened**:

I ran `kind create cluster` and have a Go stacktrace. I ran again with `--retain` for additional logs for this report.

```
 $ kind create cluster --retain                                         
Creating cluster "kind" ...                                                                              
 ✓ Ensuring node image (kindest/node:v1.25.3) 🖼                                                          
 ✓ Preparing nodes 📦                                                                                    
 ✓ Writing configuration 📜                                                                              
 ✗ Starting control-plane 🕹️                                                                              
ERROR: failed to create cluster: failed to init node with kubeadm: command "docker exec --privileged kind
-control-plane kubeadm init --skip-phases=preflight --config=/kind/kubeadm.conf --skip-token-print --v=6"
 failed with error: exit status 1                                                                        
Command Output: I1031 19:15:30.693671     150 initconfiguration.go:254] loading configuration from "/kind
/kubeadm.conf"                                                                                           
W1031 19:15:30.695745     150 initconfiguration.go:331] [config] WARNING: Ignored YAML document with Grou
pVersionKind kubeadm.k8s.io/v1beta3, Kind=JoinConfiguration                                              
[init] Using Kubernetes version: v1.25.3                                                                 
[certs] Using certificateDir folder "/etc/kubernetes/pki"                                                
I1031 19:15:30.702885     150 certs.go:112] creating a new certificate authority for ca                  
[certs] Generating "ca" certificate and key                                                              
I1031 19:15:30.813990     150 certs.go:522] validating certificate period for ca certificate             
[certs] Generating "apiserver" certificate and key                                                       
[certs] apiserver serving cert is signed for DNS names [kind-control-plane kubernetes kubernetes.default 
kubernetes.default.svc kubernetes.default.svc.cluster.local localhost] and IPs [10.96.0.1 172.19.0.2 127.
0.0.1]                                                                                                   
[certs] Generating "apiserver-kubelet-client" certificate and key                                        
I1031 19:15:31.060575     150 certs.go:112] creating a new certificate authority for front-proxy-ca      
[certs] Generating "front-proxy-ca" certificate and key                                                  
I1031 19:15:31.193915     150 certs.go:522] validating certificate period for front-proxy-ca certificate 
[...]

PR #2988, opened 2022-10-31
add zfsutils-linux to base Dockerfile

jrwren:
possibly a fix for #2163 so that the `snapshotter = "native"` config workaround is not needed.


BenTheElder:
One problem with this is that the tools will not match the kernel module version.

I do not recommend using the ZFS snapshotter, I don't think it's well tested and I don't think we can support it well containerized.

BenTheElder:
Thanks, and thank you for your help on https://github.com/kubernetes-sigs/kind/issues/2987, but we'll merge https://github.com/kubernetes-sigs/kind/pull/2989 to fix ZFS

PR #2990, opened 2022-11-02
upgrade go to 1.19.3

BenTheElder:


BenTheElder:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/2990/pull-kind-e2e-kubernetes/1587649922096697344

> [It] [sig-node] Pods Extended Pod Container Status should never report container start when an init container fails

/retest

aojea:
> https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/2990/pull-kind-e2e-kubernetes/1587649922096697344
> 
> > [It] [sig-node] Pods Extended Pod Container Status should never report container start when an init container fails
> 
> /retest

yeah, it is a classic lately

/lgtm

BenTheElder:
upstream regression caused CI on kubernetes master to flake a lot, the fix is coming but in the meantime we can go ahead with the 1.19.3 fix here.

PR #2995, opened 2022-11-07
update install instructions on site index

elmiko:
This change updates the instructions in the "Installation and usage" section of the index. It is being proposed because the `go get` method of installation is no longer current for recent version of Go.

### DIFF
--- a/site/content/_index.md
+++ b/site/content/_index.md
@@ -27,7 +27,7 @@ kind bootstraps each "node" with [kubeadm][kubeadm]. For more details see [the d
 
 For more detailed instructions see [the user guide][user guide].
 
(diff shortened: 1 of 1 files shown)


elmiko:
> Thanks, have we also missed this in the README.md?
> 

i added another commit to fix the readme as well

> We have an odd balance here, we have a much more detailed install guide in the quick start page but people often don't click links and we want to highlight the easy install.
> 
> It's really too bad go made this more confusing disappointed

agreed, even i missed the instructions when clicking around. i think it's nice to have an easy find `run this to get kind` style instruction, but having it repeated in several places makes it a little tough on the updates.

Issue #3006, opened 2022-11-18
Kind v0.17 Image name is blank when doing a load


sbkpilot1:
**What happened:**
Just upgraded to 0.17 and noticed that the image name is blank when doing a `kind load docker-image`, attached is a screen shot

**What you expected to happen:**
Image tag name should appear in the Image Loading message, this used to happen in earlier Kind versions

**How to reproduce it (as minimally and precisely as possible):**
Create a kind cluster
`kind create cluster --name test`

Load  a local docker image into the cluster
`kind load docker-image busybox:latest --name test`

Note Image is blank in the message should actually read `Image: "busybox:latest with ID...` instead it says `Image: ""`

**Anything else we need to know?:**
On WSL2 with Windows 11

**Environment:**
kind version: (use kind version): 0.17.0
Kubernetes version: (use kubectl version): v1.25.3
Docker version: (use docker info): 20.10.17
OS (e.g. from /etc/os-release): Debian 11 (bullseye) on WSL2

![kind_017_docker_load](https://user-images.githubusercontent.com/20803783/203151977-eb20fc01-e6c5-4374-ac8f-2ddb0ade7731.png)

BenTheElder:
Can you please fill out the rest of the bug template with the environment details (docker or podman info command output, os and kind version output, etc) annd include a minimal example to reproduce this?

sbkpilot1:
`❯ docker info
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Docker Buildx (Docker Inc., v0.8.2)
  compose: Docker Compose (Docker Inc., v2.6.1)
  extension: Manages Docker extensions (Docker Inc., v0.2.7)
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc., 0.6.0)
  scan: Docker Scan (Docker Inc., v0.17.0)

Server:
 Containers: 86
  Running: 2
  Paused: 0
  Stopped: 84
 Images: 37
 Server Version: 20.10.17
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
[...]

PR #3013, opened 2022-12-01
load docker-image: only use sanitized name for retagging

rmfitzpatrick:
Should resolve https://github.com/kubernetes-sigs/kind/issues/3006 by only using the sanitized name to retag as I think was originally intended by https://github.com/kubernetes-sigs/kind/pull/2955.

### DIFF
--- a/pkg/cmd/kind/load/docker-image/docker-image.go
+++ b/pkg/cmd/kind/load/docker-image/docker-image.go
@@ -136,7 +136,7 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
 		imageID := imageIDs[i]
 		processed := false
 		for _, node := range candidateNodes {
-			exists, reTagRequired, imageName := checkIfImageReTagRequired(node, imageID, imageName, nodeutils.ImageTags)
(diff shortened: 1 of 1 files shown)


pkg/cmd/kind/load/docker-image/docker-image.go
```diff
@@ -145,7 +145,7 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
 				// We will try to re-tag the image. If the re-tag fails, we will fall back to the default behavior of loading
 				// the images into the nodes again
 				logger.V(0).Infof("Image with ID: %s already present on the node %s but is missing the tag %s. re-tagging...", imageID, node.String(), imageName)
```
tao12345666333, line 147: we should use `sanitizedImageName`

rmfitzpatrick: I originally opted to respect what the user provided as an arg but this would be more accurate. updated.

BenTheElder (commented):
/lgtm
/approve

PR #3023, opened 2022-12-14
e2e-k8s.sh: support alpha features

pohly:
FEATURE_GATES and RUNTIME_CONFIG can be set to populate the corresponding kind config and thus modify the cluster.

### DIFF
--- a/hack/ci/e2e-k8s.sh
+++ b/hack/ci/e2e-k8s.sh
@@ -24,7 +24,14 @@ set -o errexit -o nounset -o xtrace
 # FOCUS: ginkgo focus regex
 # GA_ONLY: true  - limit to GA APIs/features as much as possible
 #          false - (default) APIs and features left at defaults
-# 
+# FEATURE_GATES:
+#          JSON or YAML encoding of a string/bool map: {"FeatureGateA": true, "FeatureGateB": false}
+#          Enables or disables feature gates in the entire cluster.
+#          Cannot be used when GA_ONLY=true.
+# RUNTIME_CONFIG:
+#          JSON or YAML encoding of a string/string (!) map: {"apia.example.com/v1alpha1": "true", "apib.example.com/v1beta1": "false"}
+#          Enables API groups in the apiserver via --runtime-config.
+#          Cannot be used when GA_ONLY=true.
 
 # cleanup logic for cleanup on exit
 CLEANED_UP=false
@@ -107,17 +114,25 @@ create_cluster() {
       \"logging-format\": \"${KUBELET_LOG_FORMAT}\""
   fi
 
-  # JSON map injected into featureGates config
-  feature_gates="{}"
-  # --runtime-config argument value passed to the API server
-  runtime_config="{}"
+  # JSON or YAML map injected into featureGates config
+  feature_gates="${FEATURE_GATES:-{\}}"
+  # --runtime-config argument value passed to the API server, again as a map
+  runtime_config="${RUNTIME_CONFIG:-{\}}"
 
   case "${GA_ONLY:-false}" in
   false)
-    feature_gates="{}"
-    runtime_config="{}"
+    :
     ;;
   true)
+    if [ "${feature_gates}" != "{}" ]; then
+      echo "GA_ONLY=true and FEATURE_GATES=${feature_gates} are mutually exclusive."
+      exit 1
+    fi
+    if [ "${runtime_config}" != "{}" ]; then
+      echo "GA_ONLY=true and RUNTIME_CONFIG=${runtime_config} are mutually exclusive."
+      exit 1
+    fi
+
     echo "Limiting to GA APIs and features for ${KUBE_VERSION}"
     feature_gates='{"AllAlpha":false,"AllBeta":false}'
     runtime_config='{"api/alpha":"false", "api/beta":"false"}'


pohly:
Let's test the modified script first in https://github.com/kubernetes/test-infra/pull/28248 before merging this PR here.

aojea:
/assign @BenTheElder 
too much bash for me 😄

BenTheElder:
Discussed more with aojea. I think we should support this. We won't make the go commands drop in replacements for this script and will migrate some core workflows to something else later while leaving this alone.

hack/ci/e2e-k8s.sh
```diff
@@ -24,7 +24,12 @@ set -o errexit -o nounset -o xtrace
 # FOCUS: ginkgo focus regex
 # GA_ONLY: true  - limit to GA APIs/features as much as possible
 #          false - (default) APIs and features left at defaults
-# 
+# FEATURE_GATES:
```
BenTheElder, line 27: I think maybe we should just require a valid yaml map string in the env and not try to do any bash parsing. Same for runtime_config

like `FEATURE_GATES='{foo: true}'`

pohly: Makes sense, will change it.

pohly: I keep forgetting that YAML supports JSON and thus everything in a single string with no line breaks :sweat_smile:

pohly: There's one small (and perfectly acceptable) loss of functionality when removing the parsing:

> For the sake of simplicity, not attempt is made to merge the settings from
> GA_ONLY with these new env variables. If GA_ONLY=true is set, the new env
> variables are ignored.

Revised commit pushed.

/hold

I still need to test it.

aojea:
+1

hack/ci/e2e-k8s.sh
pohly: I tested:
- no vars set
- GA_ONLY=true
- FEATURE_GATES/RUNTIME_CONFIG as in https://github.com/kubernetes/test-infra/pull/28248
- all variables

That all works as expected now.

/hold cancel

pohly: The script is also being used in https://testgrid.k8s.io/sig-instrumentation-tests#kind-json-logging-master through https://github.com/pohly/test-infra/blob/cc00ee53e0bee4a32e3171d86ddfc67438c81219/config/jobs/kubernetes/sig-instrumentation/sig-instrumentation-kind-periodics.yaml#L36-L43

That also works.

hack/ci/e2e-k8s.sh
```diff
@@ -106,15 +113,14 @@ create_cluster() {
...
   fi
 
-  # JSON map injected into featureGates config
-  feature_gates="{}"
-  # --runtime-config argument value passed to the API server
-  runtime_config="{}"
+  # JSON or YAML map injected into featureGates config
+  feature_gates="${FEATURE_GATES:-{\}}"
+  # --runtime-config argument value passed to the API server, again as a map
+  runtime_config="${RUNTIME_CONFIG:-{\}}"
```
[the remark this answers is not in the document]
pohly: A [prior version of this PR](https://github.com/kubernetes-sigs/kind/compare/9bde01149e6be33aad81dd1e57dd7d9fe7db413d..f5ece4804021e04dd3d4352b23dbec2ec296440a) allowed both to be set and used shell code to parse and merge all settings. This was deemed too complex and now the options are mutually exclusive: setting `GA_ONLY` overrides `FEATURE_GATES/RUNTIME_CONFIG`.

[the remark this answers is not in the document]
pohly: FWIW, I think that's sufficient: someone who has a need to set features individually can also add the `"api/alpha":"false", "api/beta":"false"` there. The special case for `"certificates.k8s.io/v1beta1":"true"` seems irrelevant now.

pohly:
@BenTheElder, @liggitt: have your questions regarding this PR been addressed? Please take another look.

hack/ci/e2e-k8s.sh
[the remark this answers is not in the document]
liggitt: in the GA_ONLY=true case below, can we detect and error if $feature_gates or $runtime_config are anything other than `{}`, rather than silently ignoring the envvar-set values

pohly:
I force-pushed without any changes to trigger a retest.

liggitt:
one comment to explicitly make GA_ONLY=true and FEATURE_GATES/RUNTIME_CONFIG use mutually exclusive, then lgtm

liggitt:
you'll need to rebase as well, the GA_ONLY=true path got cleaned up a couple months ago

hack/ci/e2e-k8s.sh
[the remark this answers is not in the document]
pohly: Added.

pohly:
... and rebased.

hack/ci/e2e-k8s.sh
```diff
@@ -24,7 +24,14 @@ set -o errexit -o nounset -o xtrace
 # FOCUS: ginkgo focus regex
 # GA_ONLY: true  - limit to GA APIs/features as much as possible
 #          false - (default) APIs and features left at defaults
-# 
+# FEATURE_GATES:
+#          JSON or YAML encoding of a string/bool map: {"FeatureGateA": true, "FeatureGateB": false}
+#          Enables or disables feature gates in the entire cluster.
+#          Ignored when GA_ONLY=true.
```
liggitt, line 30:
```suggestion
#          Cannot be used when GA_ONLY=true.
```

hack/ci/e2e-k8s.sh
```diff
@@ -24,7 +24,14 @@ set -o errexit -o nounset -o xtrace
...
+# RUNTIME_CONFIG:
+#          JSON or YAML encoding of a string/string (!) map: {"apia.example.com/v1alpha1": "true", "apib.example.com/v1beta1": "false"}
+#          Enables API groups in the apiserver via --runtime-config.
+#          Ignored when GA_ONLY=true.
```
liggitt, line 34:
```suggestion
#          Cannot be used when GA_ONLY=true.
```

hack/ci/e2e-k8s.sh
pohly: Changed.

hack/ci/e2e-k8s.sh
pohly: Changed.

hack/ci/e2e-k8s.sh
```diff
@@ -107,17 +114,25 @@ create_cluster() {
...
+  # JSON or YAML map injected into featureGates config
+  feature_gates="${FEATURE_GATES:-{\}}"
+  # --runtime-config argument value passed to the API server, again as a map
+  runtime_config="${RUNTIME_CONFIG:-{\}}"
 
   case "${GA_ONLY:-false}" in
   false)
-    feature_gates="{}"
-    runtime_config="{}"
+    :
```
aojea, line 124: puff, you make me learn things I wish I didn't know they existed 😄  https://stackoverflow.com/a/3224910

aojea:
/approve
/hold

Jordan for final approval (unhold)

aojea:
oh, he already did
/hold cancel

Thanks patrick

hack/ci/e2e-k8s.sh
serathius: Oh no, you incentivised me to click it, now I also wish to forget this.

Issue #2954, opened 2022-10-04
limit loadbalancer max sockets


killianmuldoon:
I'm currently dealing with this issue with haproxy: https://github.com/docker-library/haproxy/issues/194 It also impacts the kindest/haproxy docker images (which is where I came across it).

Setting a global `maxconn` in the haproxy.cfg file would fix the impact of the issue, though not the source issue which is somewhere between haproxy and docker AFAIK.

I was wondering if it would be acceptable to set that value in the haproxy config, and if so what value would be a good default? For fixing the above issue I think it can be arbitrarily high (as the issue seems to be the value being unset rather than what level it's set to).

killianmuldoon:
> CAPD is a bit of an odd duck here 😅

Fair :D It does copy-past the folder into a third_party directory (and the version used there is many versions out of date)

BenTheElder:
An odd duck we love :-)

Yeah, I think copying in is the safest approach, so CAPD can continue to update when ready.

Since both projects are kubernetes org owned and identical license, at least there's limited copy-paste license concerns 🙃

dlipovetsky:
> Setting a global maxconn in the haproxy.cfg file would fix the impact of the issue, though not the source issue which is somewhere between haproxy and docker AFAIK.

From what I understand, the root cause is that HAproxy allocates memory for each _possible_ concurrent connection, up to the maximum defined by the `maxconn` field in its configuration.

Kind's HAproxy configuration does not set `maxconn`, so HAProxy derives it from the file descriptor limit, which the container inherits:

> If this value is not set, it will automatically be calculated based on the current file descriptors limit reported by the "ulimit -n" command, possibly reduced to a lower value if a memory limit is enforced, based on the buffer size, memory allocated to compression, SSL cache size, and use or not of SSL and the associated maxsslconn (which can also be automatic). -- https://cbonte.github.io/haproxy-dconv/2.2/configuration.html#maxconn

Here's a demonstration using the latest image:

```console
❯ docker run --ulimit nofile=1000 --rm -it --memory=1gb --name haproxy-test kindest/haproxy:v20230227-d46f45b6 -d | grep maxconn
Note: setting global.maxconn to 453.

❯ docker run --ulimit nofile=10000 --rm -it --memory=1gb --name haproxy-test kindest/haproxy:v20230227-d46f45b6 -d | grep maxconn
Note: setting global.maxconn to 4953.

[...]

PR #3028, opened 2022-12-14
set nofile for loadbalancer

BenTheElder:
fixes https://github.com/kubernetes-sigs/kind/issues/2954

/hold

needs some more testing


dlipovetsky:
With #3115 merged, we explicitly limit HAProxy maximum concurrent connections. Therefore, I think this change is no longer needed.

Please see the explanation in https://github.com/kubernetes-sigs/kind/issues/2954#issuecomment-1453826595.

aojea:
I'm going to use the haproxy image for the cloud-provider-kind, I think that is better to set a sane default in haproxy config directly

BenTheElder:
thanks!

PR #3037, opened 2022-12-20
images/haproxy: overwrite instead of concatenating dpkg status

BenTheElder:
see: https://github.com/kubernetes/release/pull/2831


BenTheElder:
this change isn't affecting that job, we just have some flaky tests upstream

PR #3055, opened 2023-01-07
Adding support for 'kind provider' command

yibozhuang:
I'm very much open to UX feedback, as we can do something like `kind get provider`, but keeping it at `kind provider` for now as this seems to be somewhat separate from the other operations.

There are use cases where it would help to be able to list the supported providers as well as display the currently configured runtime provider.

This change adds a `kind provider` command which
will display all supported providers. The `--current` flag can be provided to show the currently configured provider which honors the KIND_EXPERIMENTAL_PROVIDER env if set.

Addresses: #3056 

Signed-off-by: Yibo Zhuang <yibzhuang@gmail.com>

### DIFF
--- a/pkg/cluster/provider.go
+++ b/pkg/cluster/provider.go
@@ -17,6 +17,7 @@ limitations under the License.
 package cluster
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


BenTheElder:
> I can also open an issue to discuss use cases / UX further, please let me know.

Please do this first, per our contributor guide.
https://kind.sigs.k8s.io/docs/contributing/getting-started/#4-reaching-out

BenTheElder:
Thanks but we're not moving forward with this at this time.
https://github.com/kubernetes-sigs/kind/issues/3056

Issue #3054, opened 2023-01-07
Networking somewhat broken with iptables nf_tables mode 1.8.8 on the host


BenTheElder:
Quick placeholder issue for https://github.com/kubernetes/minikube/issues/15573#issuecomment-1374286267

We're going to have to get creative to deal with this properly. cc @aojea

aojea:
@danwinship these is the second weird report about iptables, the first was this one https://github.com/kubernetes/kubernetes/issues/114537

Is there something new going on with the iptables binaries?

aojea:
Oh, I see, it seems to be related to https://bugzilla.netfilter.org/show_bug.cgi?id=1632

BenTheElder:
I actually think we should just do the minimum viable thing and unblock this. I think it's less fragile than parsing the rules and reconstructing the entire rule and hopefully iptables nftables is working on compatibility improvements now.

https://github.com/kubernetes-sigs/kind/pull/3121

BenTheElder:
mitigated in https://github.com/kubernetes-sigs/kind/pull/3121, should be released soon (maybe next week, lot going on atm http://k8s.io/image-registry-redirect)

PR #3059, opened 2023-01-08
Iptables mess

aojea:
As explained in https://bugzilla.netfilter.org/show_bug.cgi?id=1632
there is no compatibility guarantee for the userspace iptables
binaries, this causes that the output of iptables-save may be
different if the iptables host version doesn't match the version used
inside the kind node.

Since the iptables we need to process for doing the dns magic are well
known and can not change without breaking a lot of users, instead of
manipulating the existing rules, just obtain the parameters that we
need and remove and add the necessary rules.

Fixes: #3054

### DIFF
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -405,15 +405,49 @@ enable_network_magic(){
   fi
 
   # patch docker's iptables rules to switch out the DNS IP
-  iptables-save \
-    | sed \
-      `# switch docker DNS DNAT rules to our chosen IP` \
-      -e "s/-d ${docker_embedded_dns_ip}/-d ${docker_host_ip}/g" \
-      `# we need to also apply these rules to non-local traffic (from pods)` \
-      -e 's/-A OUTPUT \(.*\) -j DOCKER_OUTPUT/\0\n-A PREROUTING \1 -j DOCKER_OUTPUT/' \
-      `# switch docker DNS SNAT rules rules to our chosen IP` \
-      -e "s/--to-source :53/--to-source ${docker_host_ip}:53/g"\
-    | iptables-restore
+  # we can't use iptables-save | (sed) | iptables-restore because is not safe (#3054)
+  local iptables_original
+  iptables_original="$(iptables-save -t nat | grep DOCKER)"
+
+  # capture the forwarded TCP and UDP ports for the docker embedded DNS
+  local docker_embedded_dns_tcp_port
+  docker_embedded_dns_tcp_port="$( printf '%s' "${iptables_original}" | grep tcp | grep -E -o "\-\-to\-destination ${docker_embedded_dns_ip}:([0-9]+)" | cut -d\: -f2 | head -n1)"
+  local docker_embedded_dns_udp_port
+  docker_embedded_dns_udp_port="$( printf '%s' "${iptables_original}" | grep udp | grep -E -o "\-\-to\-destination ${docker_embedded_dns_ip}:([0-9]+)" | cut -d\: -f2 | head -n1)"
+
+  # Skip if this Node doesn't have embedded dns
+  if [[ -z ${docker_embedded_dns_tcp_port} ]] || [[-z ${docker_embedded_dns_udp_port} ]]; then
+    return
+  fi
+
+  # capture all the iptables rules we have to change, those that reference the docker embedded dns IP
+  local iptables_rules_change
+  iptables_rules_change="$(printf '%s' "${iptables_original}" | grep ${docker_embedded_dns_ip})"
+
+  # remove all rules related to the embedded dns server
+  local iptables_remove
+  iptables_remove="$( printf '%s' "${iptables_rules_change}" | sed "s/^-A/iptables -t nat -D/g" )"
+  eval "$iptables_remove"
+
+  # add new rules to forward the traffic directed to the docker_host_ip:53
+  # to the docker_embedded_dns_ip:[docker_embedded_dns_tcp_port,docker_embedded_dns_udp_port]
+  local iptables_add_rules
+  iptables_add_rules=(
+  # capture all traffic from the node and and the pods directed to the new nameserver (docker_host_ip)
+  "PREROUTING -d ${docker_host_ip} -j DOCKER_OUTPUT"
+  "OUTPUT -d ${docker_host_ip} -j DOCKER_OUTPUT"
+  "POSTROUTING -d ${docker_host_ip} -j DOCKER_POSTROUTING"
+  # forward the traffic directed to the docker_host_ip to the docker embedded dns
+  "DOCKER_OUTPUT -d ${docker_host_ip} -p tcp -m tcp --dport 53 -j DNAT --to-destination ${docker_embedded_dns_ip}:${docker_embedded_dns_tcp_port}"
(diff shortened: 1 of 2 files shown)


images/base/files/usr/local/bin/entrypoint
```diff
@@ -405,15 +405,49 @@ enable_network_magic(){
...
-      `# switch docker DNS DNAT rules to our chosen IP` \
-      -e "s/-d ${docker_embedded_dns_ip}/-d ${docker_host_ip}/g" \
-      `# we need to also apply these rules to non-local traffic (from pods)` \
-      -e 's/-A OUTPUT \(.*\) -j DOCKER_OUTPUT/\0\n-A PREROUTING \1 -j DOCKER_OUTPUT/' \
-      `# switch docker DNS SNAT rules rules to our chosen IP` \
-      -e "s/--to-source :53/--to-source ${docker_host_ip}:53/g"\
-    | iptables-restore
+  # we can't use iptables-save | (sed) | iptables-restore because is not safe (#3054)
+  local iptables_original
+  iptables_original="$(iptables-save -t nat | grep DOCKER)"
```
aojea, line 410: we only care about NAT rules, don't we?

images/base/files/usr/local/bin/entrypoint
```diff
@@ -405,15 +405,49 @@ enable_network_magic(){
...
+
+  # capture the forwarded TCP and UDP ports for the docker embedded DNS
+  local docker_embedded_dns_tcp_port
+  docker_embedded_dns_tcp_port="$( printf '%s' "${iptables_original}" | grep tcp | grep -E -o "\-\-to\-destination ${docker_embedded_dns_ip}:([0-9]+)" | cut -d\: -f2 | head -n1)"
```
aojea, line 414: it matches multiple entries, we just want one

aojea: the `head -n1` also swallow the exit code 1 if there is no match

aojea:
/retest
unrelated failures

BenTheElder:
We also need to start thinking about what happens when Kubernetes 1.X transitions to iptables 1.8.8, then the kind base image has ... 1.8.7 or 1.8.8 and mismatches other Kubernetes versions.

Right now, ignoring the host iptables, which are only used by dockerd, we have 1.8.7 or lower on all K8s versions and kind uses 1.8.7 in the base. But that will change and this fix won't be sufficient.

I think we might need to version match kind node and kube-proxy. We have a couple options for that but they're all sort of bad:

1) detect and switch between multiple versions in the kind node image

2) mount over iptables in the kube-proxy pod, thereby reducing the degree to which kind is usefully testing Kubernetes images but ensuring that kube-proxy environment always matches the node environment at least.

3) Extra problematic but mentioning it anyhow: attempt to locate and mount through the host iptables (... this has too many issues but sounds nice ...)

aojea:
The rules are only problematic when you rely on parsing, if you write the rules there is no problem

BenTheElder:
> The rules are only problematic when you rely on parsing, if you write the rules there is no problem

I mean consider when say kubernetes 1.27 ships iptables 1.8.8 in kube-proxy. Ignore the host for a moment.

KIND node / base image will have to pick a version to ship, and we may have the 1.8.7 vs 1.8.8 mismatch between node level rules and kube-proxy.

KIND could just upgrade to 1.8.8, but we support multiple kubernetes versions ...

kube-proxy 1.8.8+ upgrade is going to be a mess. KIND has enough control over the environment that we have options to mitigate it in KIND specifically, I don't think there will be a good general solution for kube-proxy. At some point iptables may implement the proposed backwards compatible rules-text storing, in which case there can be a breaking upgrade to that and then hopefully no future breakage.

aojea:
> between node level rules and kube-proxy.

 IIUIC only the actors that rely on manipulating the output of `iptables-save`are going to be impacted, kube-proxy doesn't have that problem,  @danwinship explain in details how kube-proxy process the output of iptables-save in the kube-router issue,

images/base/files/usr/local/bin/entrypoint
BenTheElder: I think so, also this is only relevant in the case of docker with embedded DNS.
We can skip this for podman.

I dislike how complex this is getting.

BenTheElder:
> IIUIC only the actors that rely on manipulating the output of iptables-saveare going to be impacted, kube-proxy doesn't have that problem, @danwinship explain in details how kube-proxy process the output of iptables-save in the kube-router issue,

Ah right. kube-proxy uses save but doesn't manipulate the output and only looks for chains.

OK.

I'm still not sure we should try to parse and recreate the entire rule. Docker could change the implementation entirely.

But we can match what we were doing before and re-insert --dport.

aojea:
> I'm still not sure we should try to parse and recreate the entire rule. Docker could change the implementation entirely.
> 
> But we can match what we were doing before and re-insert --dport.

I don't think we should parse, we know exactly what we want to do and the rule we need and we don't, parsing facilitates the code, but with this we

1. drop everything that NATs to embedded dns ip
2. insert the NAT rules we know we NEED

we may use our custom chains just in case.

If docker does some changes in these rules I think the drama is going to be bigger and kind is not going to be the first impacted

danwinship:
> I'm still not sure we should try to parse and recreate the entire rule. Docker could change the implementation entirely.

The current behavior definitely seems sketchy. Could you just create your own rule that ends up running before docker's rule (thus rendering it irrelevant)? You'd still have to be making assumptions about chain names and rule ordering and stuff, but at least you wouldn't depend on the specific contents of docker's rule.

BenTheElder:
> If docker does some changes in these rules I think the drama is going to be bigger and kind is not going to be the first impacted

Right now with the bug with version mismatch we only drop one trivial and very unlikely to change part of a known rule, that is that port 53 is the destination port we care about.

If docker mutates the rules in ways that should be fine to normal applications, we can wind up attempting to insert totally bogus rules, as opposed to slightly rewriting known rules _if they exist_.

I think it would be far less complicated to just ensure that the `--dport 53` is not missing. There's no reason the target port should change.

The existing mutation is pretty robust, version mismatch problems aside, because we're only substituting the IP address to hijack. We can do something similar to match the known problem rule and re-insert dport 53.

BenTheElder:
>  Could you just create your own rule that ends up running before docker's rule (thus rendering it irrelevant)? You'd still have to be making assumptions about chain names and rule ordering and stuff, but at least you wouldn't depend on the specific contents of docker's rule.

Docker's rule contains the _real_ ports of the DNS socket, I don't know why it doesn't just open a real listener on 127.0.0.1:53, but instead they have TCP and UDP sockets on two random ports on 127.0.0.11:$random, which 53 gets forwarded to selectively.

We take those injected rules and substitute the container's network bridge IP for 127.0.0.11, so we can route to this from the nested containers (with a non-local IP). 

All of the options are somewhat bad I think, KIND is ultimately a pile of dirty hacks :+)

aojea:
I think that we are looking from the wrong side 👀 , we don't depend on the rules at all, the only thing we should be worried about is the docker embedded dns IP and its ports. 
The nat rules can not change for what we want to do :) 
Dan suggestion is much cleaner than wiping and adding ours, but 🤷

BenTheElder:
FYI this PR was written on the upstream repo, which creates a branch here and then that branch gets protected by protection rules and shows up in clones. We should avoid that. This happens when you have write on a repo and click edit.

I'm going to delete the branch after manually opting out of the branch protection (which races with the protection automation).

aojea:
> FYI this PR was written on the upstream repo, which creates a branch here and then that branch gets protected by protection rules and shows up in clones. We should avoid that. This happens when you have write on a repo and click edit.

new laptop things ... sorry 🙇

Issue #3066, opened 2023-01-13
 enable containerd of docker desktop，can't build node image from k8s source


ktKongTong:
**What happened**:

after enable containerd，can't build node-image from k8s source code (release-1.26).

![image](https://user-images.githubusercontent.com/44502608/212328078-c17b57e7-4edc-4c55-a3e1-965c1360bf60.png)

![image](https://user-images.githubusercontent.com/44502608/212313870-7ae993fc-cef9-48ec-9ead-71592a997cef.png)

**What you expected to happen**:
Successfully build node image
**How to reproduce it (as _minimally_ and precisely as possible)**:
enable containerd to pull and store image, then use kind to build node image from k8s source.
**Anything else we need to know?**:
actually, after enable containerd,There is no `repositories` file in the tar file exported by `docker save`.

![image](https://user-images.githubusercontent.com/44502608/212320584-172b3857-3da3-49f3-ab3b-33c059ef8046.png)
![image](https://user-images.githubusercontent.com/44502608/212320685-d12fd34c-15af-4d9a-82d4-68c34d220542.png)

but it's seem that kind use it to get tags of image.

![image](https://user-images.githubusercontent.com/44502608/212321062-5f67c6b9-c465-433e-baf7-30d40c23993a.png)

**Environment:**
- kind version: (use `kind version`): v0.17.0
- Runtime info: (use `docker info` or `podman info`): 
```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.9.1)
  compose: Docker Compose (Docker Inc., v2.13.0)
  dev: Docker Dev Environments (Docker Inc., v0.0.5)
  extension: Manages Docker extensions (Docker Inc., v0.2.16)
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc., 0.6.0)
  scan: Docker Scan (Docker Inc., v0.22.0)
Server:
 Server Version: 20.10.21
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
[...]

PR #3067, opened 2023-01-13
fix getArchiveTags error when using containerd

ktKongTong:
fix getArchiveTags error when using containerd.

fix: https://github.com/kubernetes-sigs/kind/issues/3066

### DIFF
--- a/pkg/build/nodeimage/internal/container/docker/archive.go
+++ b/pkg/build/nodeimage/internal/container/docker/archive.go
@@ -53,7 +53,7 @@ func GetArchiveTags(path string) ([]string, error) {
 		if err != nil {
 			return nil, err
 		}
-		if hdr.Name == "repositories" {
+		if hdr.Name == "manifest.json" || hdr.Name == "repositories" {
 			break
 		}
 	}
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):

pkg/build/nodeimage/internal/container/docker/archive.go
```diff
@@ -215,3 +223,14 @@ func parseRepositories(data []byte) (archiveRepositories, error) {
 	}
 	return repoTags, nil
 }
+
+// return metadataEntries
```
AkihiroSuda, line 227: This line can be now removed. Also I suggest squashing commits.

ktKongTong: ok
```suggestion
```

pkg/build/nodeimage/internal/container/docker/archive.go
```diff
@@ -62,17 +62,25 @@ func GetArchiveTags(path string) ([]string, error) {
...
+			for tag := range tags {
+				res = append(res, fmt.Sprintf("%s:%s", repo, tag))
+			}
 		}
+	} else if hdr.Name == "manifest.json" {
+		manifest, err := parseDockerV1Manifest(b)
+		if err != nil {
+			return nil, err
+		}
+		res = append(res, manifest[0].RepoTags...)
```
aojea, line 83: I think that this failed before if was not able to parse it because at some points the marshalling will not work, and now it just returns an empty `[]string{},nil` ,
Shouldn't we keep failing if neither `manifest.json`or `repositories`is found?

maybe you can use a 
```go
switch hdr.Name:
 case manifest.json:
  // do something
 case repositories:
  // do something
 default:
   return nil, fmt.Errorf("unable to find tags, not compatible with specs ... 
```

aojea:
/lgtm
/approve

Thanks

BenTheElder:
Something is off with the merge robot, bringing up in #testing-ops in slack.

BenTheElder:
I will manually merge this if it continues to remain stuck, but SIG Testing might want a chance to debug the merge robot.

BenTheElder:
approving and running github actions to see if that's it. even though they had no status yet and aren't configured as required.

Issue #3063, opened 2023-01-12
Using "kind load docker-image img1 img2 ..." results in the image being loaded onto the node multiple times


artnaseef:
Running on a single-node Kind cluster.

Using "kind load docker-image img1 img2 ..." (where img1 and img2 are valid docker image names), the large, aggregate image created by "docker save" is uploaded to the single node multiple times - once per source image.

I tracked this down in the code and would offer a fix, but lack golang experience.  Google searching how to make a set in golang yielded no simple solution, so I decided to defer to more experienced hands here.

**What you expected to happen**:

The large image created by the "docker save" command should be loaded to the node only once, since it already contains ALL of the original docker images.

**How to reproduce it (as _minimally_ and precisely as possible)**:

"kind load docker-image img1 img2 img3"

**Anything else we need to know?**:

With almost no golang experience, managed to track this down to the following:

    Source File: pkg/cmd/kind/load/docker-image/docker-image.go

The following array is populated:

       selectedNodes := []nodes.Node{}
        fns := []func() error{}
        for i, imageName := range imageNames {
                imageID := imageIDs[i]
                for _, node := range candidateNodes {
                        id, err := nodeutils.ImageID(node, imageName)
                        if err != nil || id != imageID {
                                selectedNodes = append(selectedNodes, node)
                                logger.V(0).Infof("Image: %q with ID %q not yet present on node %q, loading...", imageName, imageID, node.String())
                        }
                }
                if len(selectedNodes) == 0 {
                        logger.V(0).Infof("Image: %q with ID %q found to be already present on all nodes.", imageName, imageID)
                        continue
                }
        }

Then iterated:

        // Load the images on the selected nodes
        for _, selectedNode := range selectedNodes {
                selectedNode := selectedNode // capture loop variable
                fns = append(fns, func() error {
                        return loadImage(imagesTarPath, selectedNode)
                })
        }

Notice there is no attempt to dedup the nodes, so the same selectedNode is processed once per image.

This can also be observed via the command "docker exec kind-control-plane ps axu | grep ctr" while the load is ongoing.

**Environment:**

- kind version: (use `kind version`):
kind v0.12.0 go1.13.8 linux/amd64

- Runtime info: (use `docker info` or `podman info`):
Docker Version 20.10.14
 
- OS (e.g. from `/etc/os-release`):
Ubuntu

- Kubernetes version: (use `kubectl version`):
1.23.4

- Any proxies or other special environment settings?:
no

[...]

PR #3070, opened 2023-01-15
fix: dedup nodes before loading images (#3063)

em-r:
this change avoids loading the same image multiple times onto a node when passing more than one image to `load docker-image`.
fixes: #3063


aojea:
/lgtm
/approve

Thanks

PR #3074, opened 2023-01-19
Enhance "kind delete cluster(s)" command

yuanchen8911:
### Why?

`kind delete cluster(s)` command doesn't report any error if the cluster(s) to be deleted doesn't exist. The output can be misleading.

### What?

Check if the cluster(s) to be deleted exist or not. If it doesn't exist, report an error `cluster xxx doesn't exist"

### Example 
Current version: 
```
$ kind delete clusters non-existing-cluster
Deleted clusters: ["non-existing-cluster"]
```

New version: 
```
$ kind delete clusters non-existing-cluster
failed to delete cluster "non-existing-cluster": cluster "non-existing-cluster" doesn't exist
Deleted clusters: []
```

### DIFF
--- a/pkg/cluster/internal/delete/delete.go
+++ b/pkg/cluster/internal/delete/delete.go
@@ -32,6 +32,9 @@ func Cluster(logger log.Logger, p providers.Provider, name, explicitKubeconfigPa
 	if err != nil {
 		return errors.Wrap(err, "error listing nodes")
 	}
+	if len(n) == 0 {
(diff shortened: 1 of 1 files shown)


aojea:
This is on purpose, I'm from mobile right now but there should be some issue or or explaining this decision

BenTheElder:
this is a breaking change we will not be accepting, with past discussions in the issue tracker e.g. https://github.com/kubernetes-sigs/kind/issues/2924

`kind delete cluster(s)` is like `rm -f`, for use re-entrantly in scripts.

PR #3079, opened 2023-01-27
fix: let kind run with podman and VirtualBox

wheelerlaw:
If a user has VirtualBox installed on the same machine they have `kind` installed, and they are using the podman experimental provider, they will likely be faced with the following error when they try to run `kind create cluster`:

```
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.25.3) 🖼
 ✗ Preparing nodes 📦  
ERROR: failed to create cluster: command "podman run --name kind-control-plane --hostname kind-control-plane --label io.x-k8s.kind.role=control-plane --privileged --tmpfs /tmp --tmpfs /run --volume e5b4dd4c571adb5257ddc3dbcc6bb6f0f220ccc935760ea7dc20539aac90d49d:/var:suid,exec,dev --volume /lib/modules:/lib/modules:ro -e KIND_EXPERIMENTAL_CONTAINERD_SNAPSHOTTER --detach --tty --net kind --label io.x-k8s.kind.cluster=kind -e container=podman --volume /dev/mapper:/dev/mapper --device /dev/fuse --publish=127.0.0.1:43527:6443/tcp -e KUBECONFIG=/etc/kubernetes/admin.conf docker.io/kindest/node@sha256:f52781bc0d7a19fb6c405c2af83abfeb311f130707a0e219175677e366cc45d1" failed with error: exit status 126
Command Output: Error: crun: error stat'ing file `/dev/vboxusb/005/006`: Permission denied: OCI permission denied
```

This is because the devices in the `/dev/vboxusb` belong to the `vboxusers` group. If the user executing the `kind` command is in that group, podman attempts to bind mount those hardware devices in the container when the `--privileged` flag is passed to podman. However, `crun` makes a `setgroups` syscall to reset the user groups so when it comes time for the bind mount to actually occur, it fails because the namespaced user in the container is no longer in that group and therefore does not have permission to `stat` the devices. 

This change passes the `--group-add keep-groups` to podman (which then passes the `run.oci.keep_original_groups=1` annotation to `crun`) so that the groups are preserved so the bind mount doesn't fail. 

(related to containers/podman#14284)


pkg/cluster/internal/providers/podman/provision.go
```diff
@@ -185,6 +185,9 @@ func runArgsForNode(node *config.Node, clusterIPFamily config.ClusterIPFamily, n
 		// including some ones podman would otherwise do by default.
 		// for now this is what we want. in the future we may revisit this.
 		"--privileged",
+		// For using podman when VirtualBox is also installed
+		// See https://github.com/containers/podman/issues/14284
```
aojea, line 189: this seems to be needed only for rootless, if you see in the `commonArgs`method we append optional arguments based on some environment details, we should have a check for rootless somewhere so we can use the same pattern for this option instead of setting it for everything

BenTheElder:
I'm a little concerned that we have diverging behavior here but I supposed that's always going to be the case with rootless.

I haven't seen similar issues with rootless runc + docker yet. Trying to think what other implications carrying through the groups might have.

wheelerlaw:
Just circling back to this now. I've been looking at the code for a little bit, and I am having a hard time understanding why the `provision.go` file exists at all. It doesn't look like it has a reference to an instance of the Provider (and so can't call the provider's cached Info() method), which would be nice to use to determine if it is running rootless or not. The functions in that file are making repeated calls to `podman info` instead of using the cached `Info()` method. I'd like to see if I can refactor it a little bit to reduce how many times it's calling out to `podman`.

BenTheElder:
> Just circling back to this now. I've been looking at the code for a little bit, and I am having a hard time understanding why the provision.go file exists at all. 

provision.go splits out the lengthy logic for organizing node creation from the rest of the provider logic.



> It doesn't look like it has a reference to an instance of the Provider (and so can't call the provider's cached Info() method), which would be nice to use to determine if it is running rootless or not. The functions in that file are making repeated calls to podman info instead of using the cached Info() method. I'd like to see if I can refactor it a little bit to reduce how many times it's calling out to podman.

That debt has happened over time but seems largely orthogonal, and has no relation to being in another file or not ..?

caching info for the duration of the process may also be a mis-feature. execing something like `podman info` is relatively cheap all things considered but persisting information without any eviction may cause issues.

BenTheElder:
> This is because the devices in the /dev/vboxusb belong to the vboxusers group. If the user executing the kind command is in that group, podman attempts to bind mount those hardware devices in the container when the --privileged flag is passed to podman.

Thinking about this again, is there any reason to expect that this is desired?
Can we configure podman to not include this mount instead?

https://github.com/containers/podman/issues/14284#issuecomment-1131645344 also comes to mind as a potential issue.

alhimik45:
Issue still exists, any solution?

BenTheElder:
@alhimik45 please see the discussion above. It's not clear what the best solution is, and the maintainers are not podman or virtualbox users: https://kind.sigs.k8s.io/docs/contributing/project-scope/

We will need more collaboration.

IIRC we have cleaned up some of the provider code tangent, but it's still not clear that we'd want to keep groups from the host, and if we did, to only do so in podman (and make the behavior diverge more from other backends).

I still think it would be best if we could make podman not attempt to add that mount, or to run under another user/group (Something you can do as the kind user without any code changes), but I'm not certain.

We generally don't actually _want_ to give these containers permissions on the host, we're unfortunately forced to do so to acquire some permissions we need to create kubelet in a container, but we don't really want to expand that into being more like root on the host and/or more variable across hosts and backends.

I would try changing your user/groups or look into configuring podman to not mount this.
https://docs.podman.io/en/stable/markdown/podman.1.html

PR #3083, opened 2023-01-31
Log cluster version in e2e-ci.sh

liggitt:


hack/ci/e2e-k8s.sh
```diff
@@ -192,6 +179,9 @@ EOF
     -v=3 \
     "--config=${ARTIFACTS}/kind-config.yaml"
 
+  # debug cluster version
+  kubectl version
```
BenTheElder, line 183: Also just for anyone else following along we do log the kubernetes version (like `v1.25.0`) in a file when we dump `kind export logs` to `$ARTIFACTS`, but that misses for example the go version. the version reported comes from the build system and is what kind uses internally to understand what version the node image is.

hack/ci/e2e-k8s.sh
```diff
@@ -118,22 +118,9 @@ create_cluster() {
     runtime_config="{}"
     ;;
   true)
-    case "${KUBE_VERSION}" in
```
BenTheElder, line 121: fair enough :+)

BenTheElder (commented):
/lgtm
/approve

PR #3091, opened 2023-02-09
test: update image registry from k8s.gcr.io to registry.k8s.io

mengjiao-liu:
Ref: https://github.com/kubernetes/k8s.io/issues/4738#issuecomment-1423021996


aojea:
/approve

yeah, these are unit tests, doesn't pull

BenTheElder:
Specifically this is test-data only, it's only parsed.

PR #3093, opened 2023-02-12
docs: add example to specify kubernetes source location

m-Bilal:
Kind docs currently ask users to "[ensure that Kubernetes is cloned in `$(go env GOPATH)/src/k8s.io/kubernetes`](https://github.com/kubernetes-sigs/kind#installation-and-usage)". This got me confused and led me to believe that we can't have the source anywhere else, since we can't specify its location (which is not the case). This PR aims to fix the documentation by adding an example which uses a different Kubernetes source location.

### DIFF
--- a/site/content/docs/user/quick-start.md
+++ b/site/content/docs/user/quick-start.md
@@ -260,6 +260,11 @@ Currently, kind supports one default way to build a `node-image`
 if you have the [Kubernetes][kubernetes] source in your host machine
 (`$GOPATH/src/k8s.io/kubernetes`), by using `docker`.
 
+You can also specify a different path to kubernetes source using 
+```
+kind build node-image /path/to/kubernetes/source
+```
+
 > **NOTE**: Building Kubernetes node-images requires everything building upstream
 > Kubernetes requires, we wrap the upstream build. This includes Docker with buildx.
 > See: https://git.k8s.io/community/contributors/devel/development.md#building-kubernetes-with-docker


README.md
```diff
@@ -97,10 +97,10 @@ kind delete cluster
 
 <!--TODO(bentheelder): improve this part of the guide-->
 To create a cluster from Kubernetes source:
-- ensure that Kubernetes is cloned in `$(go env GOPATH)/src/k8s.io/kubernetes`
```
stmcginnis, line 100: I think it would be good to keep some mention that the default path is `$(go env GOPATH)/src/k8s.io/kubernetes`, but can be pointed at a different path if provided on the command line.

BenTheElder:
> Kind docs currently ask users to "[ensure that Kubernetes is cloned in $(go env GOPATH)/src/k8s.io/kubernetes](https://github.com/kubernetes-sigs/kind#installation-and-usage)". 

AFAIK this is still best practice for developing Kubernetes, as some tools in github.com/kubernetes/kubernetes vendor + GOPATH mode.

m-Bilal:
@stmcginnis @BenTheElder I have reverted the previous change and added a new example that makes it clear. Thank you for your suggestions! Do let me know if anything else needs to be done. (sorry for the patchy availability, I only get time to do this on weekends :sweat_smile: )

stmcginnis:
/lgtm
/retest

No need to apologize! Thanks for trying to improve the docs.

BenTheElder (commented):
sorry this slipped off my radar :(

can we add this here instead https://kind.sigs.k8s.io/docs/user/quick-start/#building-images

the readme is meant to be a quick over view of how easily kind can be used, not a comprehensive usage guide. these pages are already quite long.

mimani68 (approved):

stmcginnis:
/lgtm

Good call - this location makes sense and fits well. Thanks for improving the docs!

BenTheElder (commented):
/lgtm
/approve

Issue #3097, opened 2023-02-14
Add DNS search configuration


howardjohn:
**What would you like to be added**:

Configuration to override the dns search in /etc/resolv.conf of nodes

**Why is this needed**:

Docker inherits resolv.conf of the host, which may have absurd number of search rules due to corporate policies, etc. These end up in all pods, making DNS latency quite high.

On my machine, a curl request takes 100ms because of this. Without search, its 10ms. This makes our tests nearly 10x slower.

https://kubernetes.slack.com/archives/CEKK1KTN2/p1676407003927219 has a bit more details and implementation ideas.

aojea:
for ref https://docs.docker.com/config/containers/container-networking/#dns-services

```
--dns-search
```

PR #3098, opened 2023-02-14
Add `dnsSearch` cluster option

howardjohn:
Fixes https://github.com/kubernetes-sigs/kind/issues/3097

The implementation here may seem kind of complex - docker has a `--dns-search`, after all. But there are a few reasons we cannot use it directly

* We cannot set *no* search through this mechanism, so users need to set at least one value there.
* When any DNS options are set, docker skips its own (complex) logic to determine the correct DNS config. So simply setting --dns-search actually breaks all DNS in many cases. For example, a 127.0.0.1 resolver  on the host would be used directly when the flag is set (broken), but without the flag docker rewrites this to 8.8.8.8.

We could re-implement dockers logic and deny empty search list, but this seems like a much simpler option.

### DIFF
--- a/pkg/apis/config/v1alpha4/types.go
+++ b/pkg/apis/config/v1alpha4/types.go
@@ -189,6 +189,8 @@ type Networking struct {
 	// KubeProxyMode defines if kube-proxy should operate in iptables or ipvs mode
 	// Defaults to 'iptables' mode
 	KubeProxyMode ProxyMode `yaml:"kubeProxyMode,omitempty" json:"kubeProxyMode,omitempty"`
+	// DNSSearch defines the DNS search domain to use for nodes. If not set, this will be inherited from the host.
+	DNSSearch *[]string `yaml:"dnsSearch,omitempty" json:"dnsSearch,omitempty"`
 }
 
 // ClusterIPFamily defines cluster network IP family
--- a/pkg/cluster/internal/providers/docker/provision.go
+++ b/pkg/cluster/internal/providers/docker/provision.go
@@ -201,18 +201,7 @@ func commonArgs(cluster string, cfg *config.Cluster, networkName string, nodeNam
 	}
 
 	if cfg.Networking.DNSSearch != nil {
-		if len(*cfg.Networking.DNSSearch) == 0 {
-			return nil, errors.New("docker provider requires a non-empty dnsSearch")
-		}
-		for _, s := range *cfg.Networking.DNSSearch {
-			args = append(args, "--dns-search", s)
-		}
-		args = append(args, "--dns", "8.8.8.8")
-		args = append(args, "--dns", "8.8.4.4")
-		if config.ClusterHasIPv6(cfg) {
-			args = append(args, "--dns", "2001:4860:4860::8888")
-			args = append(args, "--dns", "2001:4860:4860::8844")
-		}
+		args = append(args, "-e", "KIND_DNS_SEARCH="+strings.Join(*cfg.Networking.DNSSearch, " "))
 	}
 
 	return args, nil
--- a/pkg/cluster/internal/providers/docker/provision.go
+++ b/pkg/cluster/internal/providers/docker/provision.go
@@ -199,6 +199,22 @@ func commonArgs(cluster string, cfg *config.Cluster, networkName string, nodeNam
 	if mountFuse() {
 		args = append(args, "--device", "/dev/fuse")
 	}
+
+	if cfg.Networking.DNSSearch != nil {
+		if len(*cfg.Networking.DNSSearch) == 0 {
+			return nil, errors.New("docker provider requires a non-empty dnsSearch")
+		}
+		for _, s := range *cfg.Networking.DNSSearch {
+			args = append(args, "--dns-search", s)
+		}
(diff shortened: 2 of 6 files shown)


pkg/cluster/internal/providers/docker/provision.go
```diff
@@ -199,6 +199,22 @@ func commonArgs(cluster string, cfg *config.Cluster, networkName string, nodeNam
...
+		}
+		for _, s := range *cfg.Networking.DNSSearch {
+			args = append(args, "--dns-search", s)
+		}
+		args = append(args, "--dns", "8.8.8.8")
+		args = append(args, "--dns", "8.8.4.4")
+		if config.ClusterHasIPv6(cfg) {
+			args = append(args, "--dns", "2001:4860:4860::8888")
+			args = append(args, "--dns", "2001:4860:4860::8844")
+		}
```
howardjohn, line 215: I have 100% no clue why this works. Without it, things DNS gets blackholed. With it, things work perfectly and these don't actually end up in `/etc/resolv.conf` at all...

aojea: it seems an internal thing at first sight .. we should not set `--dns`, should we?

pkg/apis/config/v1alpha4/types.go
```diff
@@ -189,6 +189,8 @@ type Networking struct {
 	// KubeProxyMode defines if kube-proxy should operate in iptables or ipvs mode
 	// Defaults to 'iptables' mode
 	KubeProxyMode ProxyMode `yaml:"kubeProxyMode,omitempty" json:"kubeProxyMode,omitempty"`
+	// DNSSearch defines the DNS search domain to use for nodes. If not set, this will be inherited from the host.
+	DNSSearch *[]string `yaml:"dnsSearch,omitempty" json:"dnsSearch,omitempty"`
```
aojea, line 193: do we need a pointer ? is not the slice going to be nil if is empty?

pkg/cluster/internal/providers/docker/provision.go
howardjohn: I don't think we should set it, but somehow if we don't it doesn't work. I don't 100% understand why since even when we set it, it doesn't actually end up in.

I just found its something with `kind` network I guess, so will explore that more

```
$ docker run --dns-search=com --dns 8.8.8.8 howardjohn/nettools cat /etc/resolv.conf
search com
nameserver 8.8.8.8
options edns0 trust-ad
$ docker run --dns-search=com --dns 8.8.8.8 --dns 8.8.4.4 --network kind howardjohn/nettools cat /etc/resolv.conf
search com
nameserver 127.0.0.11
```

pkg/apis/config/v1alpha4/types.go
howardjohn: Intent was to allow `dnsSearch: []` to set it to empty, but I am open to suggestions.. I don't love it either.

pkg/cluster/internal/providers/docker/provision.go
howardjohn: I think there is some magic when you use non-default network that it configures the internal dns server. And when you DON'T set any dns configs it automatically replaces 127.0.0.1 from host with the google ones, but if you set and `--dns-*` flags that magic is gone... 

Low confidence

pkg/cluster/internal/providers/podman/provision.go
```diff
@@ -164,6 +164,15 @@ func commonArgs(cfg *config.Cluster, networkName string, nodeNames []string) ([]
 		args = append(args, "--device", "/dev/fuse")
 	}
 
+	if cfg.Networking.DNSSearch != nil {
+		if len(*cfg.Networking.DNSSearch) == 0 {
+			args = append(args, "--dns-search", "")
```
aojea, line 169: I see, so you want to use an empty value too, is this possible in docker?

pkg/apis/config/v1alpha4/types.go
aojea: got it https://github.com/kubernetes-sigs/kind/pull/3098/files#r1106429110

pkg/cluster/internal/providers/docker/provision.go
howardjohn: Updated approach and PR description with why

BenTheElder (commented):
Yeah, this looks about right, thank you!

@aojea, any objections?

I think we already have precedent for mutating /etc/resolv.conf in the image and that has gone fine, kubelet also does something kinda like this.
/lgtm

BenTheElder:
I'll be out next week, at least until thursday, possibly the whole week, but antonio can move forward on this or request changes. I want to see about rolling up a release again soon. 

We're kinda overdue for one but I've been re-prioritized to focus back on firefighting k8s infra costs for a bit while the impact is higher (since any changes we make to cut costs will have cumulative savings against this year's budget). I don't expect much more low hanging fruit to ship there and then I'll resume general k8s focus and catch up more here ...

I think antonio has been caught up in fixing up cloud provider build and test.

aojea:
no logs 🤔 

PR #3099, opened 2023-02-15
kindnet: add nodeSelector

howardjohn:
Currently, it is impossible to create a Node object and not have kindnet scheduled there. There is a blanket toleration for *any* taint, and no node selector.

This adds a nodeSelector for it, matching the one on kube-proxy. For most users this should have no impact, as kind nodes are linux.

however, for some cases where we are manually creating `Node`s like https://kwok.sigs.k8s.io/, it is useful to be able to prevent scheduling.


BenTheElder (commented):
/lgtm
/approve
thanks!

Issue #1529, opened 2020-04-30
Window CI


BenTheElder:
We should add CI for kind-on-windows.

AFAICT so far we cannot get github actions to create linux containers on windows, but we could probably bring up a GCE windows VM in a [boskos](https://github.com/kubernetes/test-infra/tree/master/boskos) managed project and run tests there?

Prow does not have windows nodes, but even if it did a windows container would not work, we need a windows VM or baremetal.

It would be nice to do this in a periodic job so we can check for breakage in a less manual way.

BenTheElder:
NOTE: it is probably non-trivial to navigate kubernetes's CI to make this work. While we're looking for help, don't expect this to necessarily be an easy starter task!

We're looking for a full test of creating a cluster. We already verify cross compliation.

wolf99:
Apologies if this is a naive suggestion; but could this help?

https://github.com/marketplace/actions/setup-wsl

BenTheElder:
I think the issue is folks are primarily on WSL2 not WSLv1 (e.g. the only docs contributed so far are for WSL2 https://kind.sigs.k8s.io/docs/user/using-wsl2/), which requires virtualization support. Per setup-wsl:
 
> WSLv2 is not explicitly supported, because as of this writing neither the Windows version running on GitHub hosted runners supports WSLv2, nor is Hyper-V enabled which is also necessary to enable WSLv2. 

This is the problem we ran into trying to setup CI ourselves previously. `@amwat` and I hacked on this previously but hit that limitation with no nested virtualization support in the github actions environment.

We learned that macOS runners (and only macOS) on github.com do have virtualization support, but didn't manage to get docker desktop running. Later someone contributed vagrant/linux rootless configurations running under macOS for testing rootless. We might be able to run windows in a VM under macOS 😂  but I'm not sure about say, licensing ...

"self hosted" runners using cloud VMs from SIG K8s Infra might? be an option, but perhaps higher effort and more to maintain.

PR #3101, opened 2023-02-16
ChromeOS needs KubeletInUserNamespace

dex4er:
I can run KIND cluster in ChromeOS sucesfully with this instruction. Actually it is not a specific to ChromeOS but it will be the same for running KIND in just any LXC container.

### DIFF
--- a/site/content/docs/user/known-issues.md
+++ b/site/content/docs/user/known-issues.md
@@ -35,7 +35,7 @@ description: |-
 * [Windows Containers](#windows-containers) (unsupported / infeasible)
 * [Non-AMD64 Architectures](#nonamd64-architectures) (images not pre-built yet)
 * [Unable to Pull Images](#unable-to-pull-images) (various)
-* [Chrome OS](#chrome-os) (unsupported)
+* [Chrome OS](#chrome-os) (needs KubeletInUserNamespace)
 * [AppArmor](#apparmor) (may break things, consider disabling)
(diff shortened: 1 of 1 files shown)


stmcginnis:
/lgtm

I have not tested and do not have a way to easily do so, but it looks like everything necessary is added and if the submitter has tested and it works for them, that's good enough for me.

One caveat I'm thinking here - should we have a statement in the docs that this is an experimental or "unsupported" use case? Great that this works and is an option for folks, but I think it strays a little outside the [primary scope](https://kind.sigs.k8s.io/docs/contributing/project-scope/) of the project.

In other words, it might be good to make it clear that the maintainers won't spend a lot of time troubleshooting if a user files an issue having a problem running on ChromeOS.

BenTheElder (commented):
/lgtm
/approve
Thank you!

BenTheElder:
That's a good point @stmcginnis but we're also not currently requiring that for say, https://kind.sigs.k8s.io/docs/user/using-wsl2/ even though those docs are entirely user-contributed and not something we test (#1529). Docker Desktop on macOS is in a similar state currently as well of supported-but-not-actively-tested and only best-effort back and forth debugging via github.

dex4er:
I see the difference between "unsupported" and "not working". I can thank you that you added the description of how to start KIND in LXC or ChromeOS even if I don't expect you can't troubleshoot such a setup.

PR #3108, opened 2023-02-22
Update during image build to get security patches

nr-dbuckwalter:
Pull in some security updates to the base image when building, as the Ubuntu base image is on a monthly-ish release cycle.  We're already updating listed packages with the `clean-install` script

### DIFF
--- a/images/base/files/usr/local/bin/clean-install
+++ b/images/base/files/usr/local/bin/clean-install
@@ -26,6 +26,7 @@ if [ $# = 0 ]; then
 fi
 
 apt-get update
+apt-get upgrade -y
 apt-get install -y --no-install-recommends "$@"
 apt-get clean -y
 rm -rf \


aojea:
we use to bump the base image, that should already bring the security updates ... why should we install those packages explicitily?

nr-dbuckwalter:
The [current ubuntu:22.04 image](https://git.launchpad.net/cloud-images/+oci/ubuntu-base/tree/oci/index.json?h=refs/tags/dist-jammy-amd64-20230126&id=d4e91f86e34e0de683c46f08178d93fcf83a515d) image does not include the fix for https://ubuntu.com/security/notices/USN-5844-1.  It [looks like the base Docker image for Ubuntu 22.04 is updated monthly-ish](https://github.com/docker-library/official-images/commits/master/library/ubuntu), so will get the patch mid-March plus whenever the next `kind` images are built/published.
The `clean-install` script effectively does an `apt-get upgrade` for listed packages, since it's not using `apt-get install --no-upgrades`, and my patch takes advantage of that.
A bigger change might be to just get all security updates when the `kind` images are built?

BenTheElder:
> A bigger/better change might be to just get all security updates when the kind images are built?

+1

aojea:
> > A bigger/better change might be to just get all security updates when the kind images are built?

is that feasible?

nr-dbuckwalter:
FWIW, the Ubuntu 22.04 base image was just updated and includes a version of openssl patched for the HIGH CVE I was looking to remove.  A rebuild/release without this PR will get the update

nr-dbuckwalter:
I updated this PR to simply install any available upgrades to packages while running `clean-install`.  This will grab any security or other update packages released after the official `ubuntu:22.04` base image is built.
This shrinks the window of time that security patches take to get into the `kind` base image.  If someone is cloning this Git repo and building a local copy of the `kind` images, any Ubuntu updates will be pulled in at the local build time.

BenTheElder (commented):
/lgtm
/approve
thanks!

BenTheElder:
we'll be rolling new images soon, kubernetes is pretty hectic today with code freeze but @aojea and I are discussing a release cut this week

Issue #3105, opened 2023-02-20
kind load docker-image does not work with podman provider


vyasgun:
**What happened**: Got the error that the image was not present locally (image is present) during `kind load docker-image` command when using podman provider.

**What you expected to happen**: The podman image to be loaded onto the kind cluster

**How to reproduce it (as _minimally_ and precisely as possible)**: 
```
KIND_EXPERIMENTAL_PROVIDER=podman kind create cluster
KIND_EXPERIMENTAL_PROVIDER=podman kind load docker-image <image-name>
```
Results in error:
```
using podman due to KIND_EXPERIMENTAL_PROVIDER
enabling experimental podman provider
ERROR: image: "<image-name>" not present locally
```
**Anything else we need to know?**:

**Environment:**

- kind version: (use `kind version`): kind v0.17.0 go1.19.3 darwin/amd64
- Runtime info: (use `docker info` or `podman info`):
```
host:
  arch: amd64
  buildahVersion: 1.28.0
  cgroupControllers:
  - cpu
  - io
  - memory
  - pids
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.1.5-1.fc37.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.1.5, commit: '
  cpuUtilization:
    idlePercent: 77.07
    systemPercent: 14.31
    userPercent: 8.62
  cpus: 1
  distribution:
    distribution: fedora
    variant: coreos
    version: "37"
  eventLogger: journald
  hostname: localhost.localdomain
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 1000000
    uidmap:
    - container_id: 0
      host_id: 501
      size: 1
    - container_id: 1
      host_id: 100000
      size: 1000000
  kernel: 6.1.9-200.fc37.x86_64
  linkmode: dynamic
  logDriver: journald
  memFree: 977330176
  memTotal: 2065821696
  networkBackend: netavark
  ociRuntime:
    name: crun
    package: crun-1.7.2-3.fc37.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 1.7.2
      commit: 0356bf4aff9a133d655dc13b1d9ac9424706cac4
      rundir: /run/user/501/crun
      spec: 1.0.0
[...]

Issue #2760, opened 2022-05-12
"kind load" tries to call docker though podman is used as runtime


Unveraenderbar:
**What happened**:
```kind load``` fails (with a misleading error message about the image not being available locally) if podman is used as runtime,
because it tries to do image management by calling a binary called "docker"

**What you expected to happen**:
If the cluster is provisioned by using podman as runtime management tool, kind should also call the podman bnary for image
management on load operations

**How to reproduce it (as minimally and precisely as possible)**:
* export ```KIND_EXPERIMENTAL_PROVIDER=podman``` as environment variable
* set up kind cluster
* pull an additional image into the local podman image store
* try to load said image into kind cluster

This provokes an error
```
$ podman images
REPOSITORY                                 TAG         IMAGE ID      CREATED       SIZE
docker.io/kindest/node                     v1.24.0     f607f70d3a1a  2 days ago    938 MB
docker.io/linuxserver/openssh-server       latest      6ecbeb69716b  4 days ago    35.5 MB
$ kind load docker-image --name $(kind get clusters) docker.io/linuxserver/openssh-server:latest
enabling experimental podman provider
enabling experimental podman provider
ERROR: image: "docker.io/linuxserver/openssh-server:latest" not present locally
$ ln -s /usr/bin/podman $HOME/.local/bin
$ kind load docker-image --name $(kind get clusters) docker.io/linuxserver/openssh-server:latest
enabling experimental podman provider
enabling experimental podman provider
Image: "docker.io/linuxserver/openssh-server:latest" with ID "6ecbeb69716b9cd757187df0ea154dd56d1bdc1093ff115559e70e25d094662a" not yet present on node "k8s-v1.24.0-control-plane", loading...
```

**Environment:**

- kind version: (use `kind version`): v0.13
- Kubernetes version: (use `kubectl version`): v1.24.0
- Docker version: (use `docker info`): none -- but podman v3.4.7 and v4.1.1 work fine
- OS (e.g. from `/etc/os-release`): Fedora 35 Silverblue

BenTheElder:
This is only true for "kind load docker-image" which yes that uses docker. See discussion in https://github.com/kubernetes-sigs/kind/issues/2729

Unveraenderbar:
I see.. thx for the quick answer! I'm fine with a symlink or image file pipe until  #2729 is realized.

BenTheElder:
[...]

Issue #2027, opened 2021-01-22
Unable to load image using Podman 


bostrt:
**What happened**: Unable to load image using Podman provider:

```
[root@localhost kind]# podman images
REPOSITORY                         TAG      IMAGE ID      CREATED       SIZE
localhost/ovn-daemonset-f          dev      abb63014304d  15 hours ago  634 MB
[root@localhost kind]# kind load docker-image ovn-daemonset-f:dev --name ovn 
enabling experimental podman provider
ERROR: image: "ovn-daemonset-f:dev" not present locally
[root@localhost kind]# kind load docker-image localhost/ovn-daemonset-f:dev --name ovn 
enabling experimental podman provider
ERROR: image: "localhost/ovn-daemonset-f:dev" not present locally
```

**What you expected to happen**: Be able to load image without any workaround (mentioned below).

**How to reproduce it (as minimally and precisely as possible)**:

1. Fresh install Fedora 33
2. Install podman (no docker)
3. clone and build kind
4. build an image using `podman build`
5. check the image exists `podman images`
6. try loading image via kind

**Anything else we need to know?**:
Yes, If I create a symlink like below, I'm able to load image:
```
[root@localhost kind]# ln -s /usr/bin/podman /usr/local/bin/docker

[root@localhost contrib]# kind load docker-image ovn-daemonset-f:dev --name ovn -v 10
enabling experimental podman provider
Image: "ovn-daemonset-f:dev" with ID "abb63014304dfae15ab5c7282b409f8a1b677bf14c6601a04e229ac489f71b29" not yet present on node "ovn-worker", loading...
...
```

**Environment:**

- kind version:
```
[root@localhost kind]# git rev-parse --short HEAD
db37e6cf
[root@localhost kind]# git status
On branch master
Your branch is up to date with 'origin/master'.

nothing to commit, working tree clean
[root@localhost kind]# kind --version
kind version 0.10.0-alpha+db37e6cfe88729
```

- Podman version:
```
[root@localhost tcpdump-docker]# podman version
Version:      2.2.1
API Version:  2.1.0
Go Version:   go1.15.5
Built:        Tue Dec  8 07:37:50 2020
OS/Arch:      linux/amd64
```

- OS:
```
[root@localhost kind]# cat /etc/redhat-release 
Fedora release 33 (Thirty Three)
```

BenTheElder:
[...]

PR #3109, opened 2023-02-23
load add image-load, support podman load

helen-frank:
/kind feature

Kind load add command image-load, extension supports podman load

### DIFF
--- a/pkg/cmd/kind/load/image-load/image-load.go
+++ b/pkg/cmd/kind/load/image-load/image-load.go
@@ -1,0 +1,305 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package load implements the `load` command
+package load
+
+import (
+	syserrors "errors"
+	"fmt"
+	"os"
+	osexec "os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"sigs.k8s.io/kind/pkg/cluster"
+	"sigs.k8s.io/kind/pkg/cluster/nodes"
+	"sigs.k8s.io/kind/pkg/cluster/nodeutils"
+	"sigs.k8s.io/kind/pkg/cmd"
+	"sigs.k8s.io/kind/pkg/errors"
+	"sigs.k8s.io/kind/pkg/exec"
+	"sigs.k8s.io/kind/pkg/fs"
+	"sigs.k8s.io/kind/pkg/log"
+
+	"sigs.k8s.io/kind/pkg/internal/cli"
+	"sigs.k8s.io/kind/pkg/internal/runtime"
+)
+
+type (
+	imageTagFetcher func(nodes.Node, string) (map[string]bool, error)
+)
+
+type flagpole struct {
+	Name  string
+	Nodes []string
+}
+
+// NewCommand returns a new cobra.Command for loading an image into a cluster
+func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
+	flags := &flagpole{}
+	cmd := &cobra.Command{
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) < 1 {
+				return errors.New("a list of image names is required")
+			}
+			return nil
+		},
+		Use:   "image-load <IMAGE> [IMAGE...]",
+		Short: "Loads images from host into nodes",
+		Long:  "Loads images from host into all or specified nodes by name",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			cli.OverrideDefaultName(cmd.Flags())
+			return runE(logger, flags, args)
+		},
+	}
+	cmd.Flags().StringVarP(
+		&flags.Name,
+		"name",
+		"n",
+		cluster.DefaultName,
+		"the cluster context name",
+	)
+	cmd.Flags().StringSliceVar(
+		&flags.Nodes,
+		"nodes",
+		nil,
+		"comma separated list of nodes to load images into",
+	)
+	return cmd
+}
+
+func runE(logger log.Logger, flags *flagpole, args []string) error {
+	provider := cluster.NewProvider(
+		cluster.ProviderWithLogger(logger),
+		runtime.GetDefault(logger),
+	)
+
+	runtimePath, err := GetDefaultRuntime()
+	if err != nil {
+		return err
+	}
+	// Check that the image exists locally and gets its ID, if not return error
+	imageNames := removeDuplicates(args)
+	var imageIDs []string
+	for _, imageName := range imageNames {
+		imageID, err := imageID(runtimePath, imageName)
+		if err != nil {
+			return fmt.Errorf("image: %q not present locally", imageName)
+		}
+		imageIDs = append(imageIDs, imageID)
+	}
+
+	// Check if the cluster nodes exist
+	nodeList, err := provider.ListInternalNodes(flags.Name)
+	if err != nil {
+		return err
+	}
+	if len(nodeList) == 0 {
+		return fmt.Errorf("no nodes found for cluster %q", flags.Name)
+	}
+
+	// map cluster nodes by their name
+	nodesByName := map[string]nodes.Node{}
+	for _, node := range nodeList {
+		// TODO(bentheelder): this depends on the fact that ListByCluster()
+		// will have name for nameOrId.
+		nodesByName[node.String()] = node
+	}
+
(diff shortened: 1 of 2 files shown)


aojea:
how is this different than current one that is not working in podman because of podman limitations to load images?

https://github.com/kubernetes-sigs/kind/issues/3105

helen-frank:
> how is this different than current one that is not working in podman because of podman limitations to load images?
> 
> #3105

My current implementation is feasible, I have tested it, and this problem is also found by wanting to use kwok on the kind-podman
```bash
podman pull docker.io/kindest/node:v1.25.0                                                                                                                                                                     ─╯
Trying to pull docker.io/kindest/node:v1.25.0...
Getting image source signatures
Copying blob 9602ae1e967a done
Copying blob 405bac4f49ab done
Copying config d3da246e12 done
Writing manifest to image destination
Storing signatures
d3da246e125a0644c137ffd5ecaea7658d4875581a82891b955b319aa6341d11
╭─   ~                                                                                                                                                            took  40s with root@10-29-16-89 at  10:32:24 ─╮
╰─❯ podman version                                                                                                                                                                                                 ─╯
Version:      3.4.4
API Version:  3.4.4
Go Version:   go1.17.3
Built:        Thu Jan  1 00:00:00 1970
OS/Arch:      linux/amd64
╭─   ~                                                                                                                                                                       with root@10-29-16-89 at  10:32:29 ─╮
╰─❯ podman ps                                                                                                                                                                                                      ─╯
CONTAINER ID  IMAGE                                                                                           COMMAND     CREATED         STATUS             PORTS                        NAMES
508321ce6a0f  docker.io/kindest/node@sha256:f52781bc0d7a19fb6c405c2af83abfeb311f130707a0e219175677e366cc45d1              7 days ago      Up 2 days ago                                   multi-node-worker2
49270a353240  docker.io/kindest/node@sha256:f52781bc0d7a19fb6c405c2af83abfeb311f130707a0e219175677e366cc45d1              7 days ago      Up 2 days ago                                   multi-node-worker
c9f635cd7edd  docker.io/kindest/node@sha256:f52781bc0d7a19fb6c405c2af83abfeb311f130707a0e219175677e366cc45d1              7 days ago      Up 2 days ago      10.29.16.89:41201->6443/tcp  multi-node-control-plane
a67d9f6f893b  docker.io/kindest/node:v1.26.0                                                                              11 minutes ago  Up 11 minutes ago  0.0.0.0:39493->6443/tcp      kwok-m1-control-plane
╭─   ~                                                                                                                                                                       with root@10-29-16-89 at  10:32:41 ─╮
╰─❯ kind load image-load --name kwok-m1 docker.io/kindest/node:v1.25.0                                                                                                                                             ─╯
Image: "docker.io/kindest/node:v1.25.0" with ID "d3da246e125a0644c137ffd5ecaea7658d4875581a82891b955b319aa6341d11" not yet present on node "kwok-m1-control-plane", loading...
╭─   ~                                                                                                                                                            took  50s with root@10-29-16-89 at  10:33:57 ─╮
╰─❯ podman exec a67d9f6f893b crictl images ls | grep "docker.io/kindest/node"                                                                                                                                      ─╯
docker.io/kindest/node                     v1.25.0              d3da246e125a0       875MB
```

aojea:
yeah, I was curious what is the difference with current docker-image command that makes this works and the other not 😄

helen-frank:
> yeah, I was curious what is the difference with current docker-image command that makes this works and the other not 😄

In fact, the difference between `image-load` and `docker-image` is whether the runtime is variable or not

The naming of `docker-image` prevents me from writing this logic to the boot

It's hard to predict how much of an impact `image-load` would have if it replaced `docker-image`

So I chose to start a new command, and it's best to use `image-load` instead of `docker-image` if you can

aojea:
> In fact, the difference between `image-load` and `docker-image` is whether the runtime is variable or not

do you mean that podman has changed something and now the `docker-image`command works with podman too?

helen-frank:
> > In fact, the difference between `image-load` and `docker-image` is whether the runtime is variable or not
> 
> do you mean that podman has changed something and now the `docker-image`command works with podman too?
> 
> if you do `kind load docker-image --name kwok-m1 docker.io/kindest/node:v1.25.0 ` does it work too? we have a bug open for a long time because it didn't worked with podman

`Kind load docker-image` cannot use podman because the code contains hard code.
Image-load is mirrored by the correct load.

pkg/cmd/kind/load/image-load/image-load.go
```diff
@@ -0,0 +1,305 @@
...
+func GetDefaultRuntime() (string, error) {
+	if _, err := osexec.LookPath("docker"); err != nil {
+		if !syserrors.Is(err, osexec.ErrNotFound) {
+			return "", err
+		}
+
+		if _, err := osexec.LookPath("podman"); err != nil {
+			return "", err
+		}
+		return "podman", nil
```
aojea, line 302: is this the difference?

aojea: because this was not working before https://github.com/kubernetes-sigs/kind/issues/2760

helen-frank: I tried to use docker-image in the form of soft links, and I succeeded.
```bash
❯ ln -s /usr/bin/podman /usr/bin/docker
❯ kind load docker-image --name kwok-m1 docker.io/kindest/node:v1.24.0
Image: "docker.io/kindest/node:v1.24.0" with ID "a7d31d743aca47d4cdb00d6f61ba58b2423a1ead57dffa234f8d4f402f0a049d" not yet present on node "kwok-m1-control-plane", loading...
❯ podman exec a67d9f6f893b crictl images ls | grep "docker.io/kindest/node"
docker.io/kindest/node                     v1.24.0              a7d31d743aca4       915MB
docker.io/kindest/node                     v1.25.0              d3da246e125a0       875MB
```

helen-frank:
```
❯ cat /etc/os-release
PRETTY_NAME="Ubuntu 22.04.2 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.2 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```

aojea: something seems to have changed in podman or my memory is confused 😄 

@BenTheElder it seems we can have the common load-image with no effort

helen-frank:
@aojea Can you give me an ok-to-test label?

aojea:
> @aojea Can you give me an ok-to-test label?

/ok-to-test

helen-frank:
Can we continue?😀 @aojea @danderson @BenTheElder 
First of all, it can be confirmed that image-load is already available on my podman environment.
What needs to be pushed forward in the future is which plan to adopt.
1. use `image-load`
2. use `image-load` and delete `docker-image`
3. Code ability to be combined into `docker-image` (not recommended, will lead to misunderstanding)

BenTheElder:
erm, Please see: https://github.com/kubernetes-sigs/kind/issues/2038, we should be discussing approaches to new features in the feature issues, per our contributing docs. [Please see: https://github.com/kubernetes-sigs/kind/issues/2038 ](https://kind.sigs.k8s.io/docs/contributing/getting-started/#4-reaching-out)

I'm sorry, but all hands on deck with Keeping the Kubernetes project funded right now (See SIG K8s Infra).
I will not be spending much time reviewing kind features until we have a plan to sustain the Kubernetes organization.

The next thing up for the kind project is a release with all the existing bug fixes, when @aojea gets time.

We need to get the API right before introducing a new command, people will begin depending on it immediately.

If podman is _actually_ working drop-in compatible here and we don't want to do all of #2038, we can design a third command (see the existing proposal) and give it a better name. `kind load image-load` is a confusing name.

mimani68 (approved):

stmcginnis (commented):
It looks like podman [does have the necessary commands](https://docs.podman.io/en/latest/markdown/podman-save.1.html) now. So from what I can tell, this should be good just swapping out `docker` for `podman` when needed.

If we have a general common command for this, should we add a note to `load docker-image` that it is deprecated and this new command should be used instead?

BenTheElder:
> If we have a general common command for this, should we add a note to load docker-image that it is deprecated and this new command should be used instead?

We probably shouldn't deprecate docker-image, it's more specific and widely used.

We really need a discussed approach (preferably in a tracking issue) that also considers nerdctl / other runtimes.

3deep5me:
Any news here? 
Would love to see kind load with podman 😊

aojea:
there is a workaround https://github.com/kubernetes-sigs/kind/issues/2027#issuecomment-765619736 in the meantime

BenTheElder:
> Any news here?
> Would love to see kind load with podman 😊

Please see the above comment and the tracking issue linked in it https://github.com/kubernetes-sigs/kind/pull/3109#issuecomment-1456932178

BenTheElder:
We require an agreed approach in feature discussion issue before adding features now, per our contributor guide.

For now I'm going to close this, pending discussion in https://github.com/kubernetes-sigs/kind/issues/2038 (I would _love_ to see help there ..)

PR #3113, opened 2023-02-27
Keep updating images

aojea:
The `local-ptah-provisioner`took a lot of time on s390x, no idea why but eventually it went through


BenTheElder (approved):
/lgtm
/approve

PR #3120, opened 2023-03-09
Update README.md

lixin963:
I think this phrase is a bit problematic


BenTheElder (changes requested):
I don't think this change makes sense.
We could drop "possibly" entirely but it's reasonably common to see this redundancy.

https://ell.stackexchange.com/a/200008.

lixin963:
OK,thanks!

PR #3121, opened 2023-03-10
add minimum viable iptables fix from https://github.com/kubernetes/mi…

BenTheElder:
…nikube/pull/15578

@aojea I think this is actually more maintainable and we're relatively unlikely to have more of these.

### DIFF
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -413,6 +413,7 @@ enable_network_magic(){
       -e 's/-A OUTPUT \(.*\) -j DOCKER_OUTPUT/\0\n-A PREROUTING \1 -j DOCKER_OUTPUT/' \
       `# switch docker DNS SNAT rules rules to our chosen IP` \
       -e "s/--to-source :53/--to-source ${docker_host_ip}:53/g"\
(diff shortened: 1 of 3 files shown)


images/base/files/usr/local/bin/entrypoint
```diff
@@ -413,6 +413,8 @@ enable_network_magic(){
       -e 's/-A OUTPUT \(.*\) -j DOCKER_OUTPUT/\0\n-A PREROUTING \1 -j DOCKER_OUTPUT/' \
       `# switch docker DNS SNAT rules rules to our chosen IP` \
       -e "s/--to-source :53/--to-source ${docker_host_ip}:53/g"\
+      `# ensure --dport on DNS rules, due to https://github.com/kubernetes-sigs/kind/issues/3054` \
```
aojea, line 416:
```suggestion
      `# nftables incompatibility between 1.8.8 and 1.8.7 omit the --dport flag on DNAT rules` \
      `# ensure --dport on DNS rules, due to https://github.com/kubernetes-sigs/kind/issues/3054` \
```

BenTheElder: ACK, agree

BenTheElder:
/hold
let's bump the base image and make sure this runs in CI

Issue #2799, opened 2022-06-08
How to overwrite resource config for kube-apiserver


krish7919:
I am new to KinD, and want to overwrite my `kube-apiserver` configs to assign different values for CPI/memory.. How do I do that..? Googling tells me I need to edit my `ClusterConfiguration`, but not sure as the only option supported is `apiServer.extraArgs` and I see no options for `apiserver.resources` or such?

CPU seems to be hardcoded in kubeadm to 250m: https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/phases/controlplane/manifests.go#L66

I have also created a patch file as:

```
spec:
  containers:
    - name: kube-apiserver
      resources:
        requests:
          cpu: 500m
        limits:
          cpu: 500
```

and written it to a file and mounted it in the control plane container. Then I do this in the kind config file:

```
- role: control-plane
    kubeadmConfigPatches:
      - |
        kind: JoinConfiguration
        patches:
          directory: /etc/kubernetes/manifests/patches
...
```

However the configs are never picked up.

The third approach I tried was contacting developers/community on the slack channel and there was no response.

**What would you like to be documented**:

Please provide an example of how to update the resource configuration in the static pod manifests generated by `kubeadm` when running `kind create cluster` command.

**Why is this needed**:

This will help overwrite the default value(s) hardcoded in code when running tests in Kind (or for other use cases).

Perhurb:
I have the same requirement. But I couldn't find any configuration for each node resource. Do you know it should be configuerd by kubeadm? Or, could you tell me if you have solute it?

aojea:
Those manifests are generated by kubeadm
cc: @neolit123

neolit123:
> However the configs are never picked up.

you probably have to also mount the patches on the kind nodes?
but yes kubeadm patches can be used to patch the static pod manifest resources:
https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/control-plane-flags/#patches

Perhurb:
How can I write the a specified patch file for my control-plane? Is there a complete example or some relative docs?

krish7919:
> we have unit tests / e2e tests that are green, so kubeadm patches work.

That's the reason why this is tagged as `documentation`; I tried a bunch of things and it didn't work, and an example of what you say would be helpful. Also, I would need to retrofit it to `kind` workflow with it configuration, but that's a different problem to solve.

neolit123:
[...]

PR #3123, opened 2023-03-11
Clarify use of mounts for merge patches

mattcary:
Fixes: #2799 

I've tested out the example patch for enabling the cache mutation detector. I haven't tested that the same strategy can be used on a worker as stated in my change, but it seems reasonable.

### DIFF
--- a/site/content/docs/user/configuration.md
+++ b/site/content/docs/user/configuration.md
@@ -455,5 +455,48 @@ nodes:
         node-labels: "my-label3=true"
 {{< /codeFromInline >}}
 
+If you need more control over patching, strategic merge and JSON6092 patches can
+be used as well. These are specified using files in a directory, for example
+`./patches/kube-controller-manager.yaml` could be the following.
+
+{{< codeFromInline lang="yaml" >}}
+apiVersion: v1
+kind: Pod
+metadata:
+  name: kube-controller-manager
+  namespace: kube-system
+spec:
+  containers:
+  - name: kube-controller-manager
+    env:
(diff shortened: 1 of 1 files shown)


site/content/docs/user/configuration.md
```diff
@@ -455,5 +455,47 @@ nodes:
...
+    containerPath: /patches
+
+kubeadmConfigPatches:
+  - |
+    kind: InitConfiguration
+    patches:
+      directory: /patches
+{{< codeFromInline >}}
+
+Note the `extraMounts` stanza. `kind` runs kubeadm from a container in the node,
```
BenTheElder, line 492: nit: I think this is slightly mis-worded / confusing. For kind the node is a container, and we run kubeadm in the node-container, but not in a container "within" the node (like say, kube-proxy).

BenTheElder:
thanks for writing this up. one nit. 

sorry for the delay, especially fun week:
https://github.com/kubernetes-sigs/kind/issues/3124
https://kubernetes.io/blog/2023/03/10/image-registry-redirect/
🙃

site/content/docs/user/configuration.md
mattcary: Good point thanks. Here'a another try.

mattcary:
> sorry for the delay, especially fun week: #3124 https://kubernetes.io/blog/2023/03/10/image-registry-redirect/ 🙃

No problem, I knew responses would be delayed. If I don't get these PRs started though I forget about them...

PR #3126, opened 2023-03-15
fix cloudbuild configs

BenTheElder:
follow up https://github.com/kubernetes-sigs/kind/pull/3125

entrypoint is a single string, need to split to args.

### DIFF
--- a/images/base/cloudbuild.yaml
+++ b/images/base/cloudbuild.yaml
@@ -3,5 +3,5 @@ options:
   substitution_option: ALLOW_LOOSE
 steps:
 - name: gcr.io/k8s-testimages/krte:latest-master
-  env:
-  entrypoint: make -C images/base push
+  entrypoint: make
+  args: ['-C', 'images/base', 'push']
--- a/images/haproxy/cloudbuild.yaml
+++ b/images/haproxy/cloudbuild.yaml
@@ -3,5 +3,5 @@ options:
   substitution_option: ALLOW_LOOSE
 steps:
 - name: gcr.io/k8s-testimages/krte:latest-master
-  env:
-  entrypoint: make -C images/haproxy push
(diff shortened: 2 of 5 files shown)


images/base/cloudbuild.yaml
```diff
@@ -3,5 +3,5 @@ options:
   substitution_option: ALLOW_LOOSE
 steps:
 - name: gcr.io/k8s-testimages/krte:latest-master
-  env:
-  entrypoint: make -C images/base push
+  entrypoint: make
+  args: ['-C' 'images/base' 'push']
```
aojea, line 7: don't you need a comma?

aojea: it seems you do https://cloud.google.com/build/docs/configuring-builds/create-basic-configuration

BenTheElder: yep

BenTheElder:
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-kind-push-base-image/1636158171183583232 🎉 

gcr.io/k8s-staging-kind/base:v20230316-a681949f@sha256:d6ccc95bab9972b302016790e9eb23ec25ccfb3a9d352ab9115802ec8c4c7fdd

BenTheElder:
two still failed:
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-kind-push-local-path-provisioner-image/1636158171330383872
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-kind-push-kindnetd-image/1636158171233914880

I think they're timing out because the slow cross-builds. we can bump the machine size or timeout limit, or we can get to removing those.

BenTheElder:
the rest passed.

aojea:
timed out in one hour?

PR #3131, opened 2023-03-21
Added CPU pinning to Docker container

hedibouattour:
This PR is related to optimizing resource usage, it is useful especially for large clusters with many nodes on the same machine.
The purpose is simply to allow pinning kind nodes to cpus using docker feature by implementing a new field in node config.

### DIFF
--- a/pkg/apis/config/v1alpha4/types.go
+++ b/pkg/apis/config/v1alpha4/types.go
@@ -118,6 +118,10 @@ type Node struct {
 	// binded to a host Port
 	ExtraPortMappings []PortMapping `yaml:"extraPortMappings,omitempty"`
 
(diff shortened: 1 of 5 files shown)


aojea:
> it is useful especially for large clusters with many nodes on the same machine.

can you open an issue and explain in more detail the problem and why this is the right solution to solve it?

/hold

dims:
xref: https://github.com/kubernetes-sigs/kind/issues/3138

stmcginnis:
This option is supported by [podman too](https://docs.podman.io/en/latest/markdown/podman-run.1.html). It would be good to add it to both providers.

BenTheElder:
Closing for now, pending sufficient discussion in https://github.com/kubernetes-sigs/kind/issues/3138

Issue #3117, opened 2023-03-08
The default CIDRs for serviceSubnet and podSubnet should be documented.


victor-sudakov:
**What would you like to be documented**:
I think that the default address pools for the serviceSubnet and podSubnet should be documented. Currently you can only learn them from studying the source code.

**Why is this needed**:
This is needed because some people use non-default address pools in the Docker daemon settings resulting in a possibility of Kind node addresses overlapping with the  serviceSubnet and podSubnet address pools which I was told kindnet does not detect and does not work with.

aojea:
This is a good issue for new contributors, the values are defined here

https://github.com/kubernetes-sigs/kind/blob/77675009ce7862102f2d4a6c6b61b8094f964ebb/pkg/apis/config/v1alpha4/default.go#L51-L72

/help

victor-sudakov:
BTW why do you have such an odd prefixlen for the IPv6  service subnet (longer than a /64)? `fd00:10:96::/112` looks kind of odd to me. A more appropriate ULA prefix length would be like /48 or /56.

aojea:
> `fd00:10:96::/112` looks kind of odd to me. A more appropriate ULA prefix length would be like /48 or /56. Is there a specific reason such a long prefixlen has been chosen?

hehe, long history https://github.com/kubernetes/enhancements/pull/1534

TL;DR this is a limitation in the implementation of the allocator that tracks teh assigned IPs to Services, it has to use a bitmap that is snapshotted to etcd, if the bitmap is too large it is not possible to store it in etcd or the performance is super bad ... hopefully I'll be able to solve this problem soon in kubernetes

Sajiyah-Salat:
I got the issue. we need to document podsubnet and servicesubnet in documentation which is ``10.96.0.0/16,  10.244.0.0/16``. Can you please specify the place where this would be useful to document.

aojea:
This looks like the better place

https://kind.sigs.k8s.io/docs/user/configuration/#networking

victor-sudakov:
> TL;DR this is a limitation in the implementation of the allocator that tracks teh assigned IPs to Services, it has to use a bitmap

OMG :-)

BenTheElder:
Besides the KEP pointer there's a rather involved PR ongoing to replace that Kubernetes allocator, Antonio's actually been authoring that and I'm intermittently attempting to help review :-)

https://github.com/kubernetes/kubernetes/pull/115075

[...]

PR #3137, opened 2023-03-24
Documented default pod and service subnets

RijulGulati:
Added default CIDR for pod subnet and service subnet.
This PR is an attempt to resolve issue #3117

### DIFF
--- a/site/content/docs/user/configuration.md
+++ b/site/content/docs/user/configuration.md
@@ -181,6 +181,8 @@ networking:
   podSubnet: "10.244.0.0/16"
 {{< /codeFromInline >}}
 
+By default, kind uses ```10.244.0.0/16``` pod subnet for IPv4 and ```fd00:10:244::/56``` pod subnet for IPv6.
+
 #### Service Subnet
 
 You can configure the Kubernetes service subnet used for service IPs by setting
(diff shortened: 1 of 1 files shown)


site/content/docs/user/configuration.md
```diff
@@ -181,6 +181,8 @@ networking:
   podSubnet: "10.244.0.0/16"
 {{< /codeFromInline >}}
 
+By default, kind uses ```10.244.0.0/16``` pod subnet for IPv4 and ```fd00:10:244::/56``` pod subnet for IPv6.
```
aojea, line 184:
```suggestion
By default, `kind` uses ```10.244.0.0/16``` pod subnet for IPv4 and ```fd00:10:244::/56``` pod subnet for IPv6.
```

@BenTheElder single ticks?

RijulGulati: I had thought of highlighting `kind` as well, but I saw kind being used without highlights in other places. So left it as it is.

aojea: Ben is the best person to decide, but I think ticks is better

aojea:
/lgtm
/approve

those were nits, this moves to a better state anyway

site/content/docs/user/configuration.md
BenTheElder: kind or KIND but not KinD
`kind` or `KIND` but not KinD

:-)

thanks for documenting this, been a hectic period with https://kubernetes.io/blog/2023/03/10/image-registry-redirect/, the docker hub announcements, the runc cves ...

BenTheElder: @aojea we could add a note to the contributing guide under docs development

PR #3139, opened 2023-03-26
Actions Updates - json/yaml validation

GrantBirki:
# Actions Updates - json/yaml validation

This pull request adds a new Actions workflow to validate all JSON / YAML files in the repository to ensure they are valid.

This workflow will run on pull requests to ensure that all files are valid from new contributors so that they do not introduce issues if they are missed in a review.

Related: https://github.com/kubernetes-sigs/external-dns/pull/3501

### DIFF
--- a/.github/workflows/json-yaml-validate.yml
+++ b/.github/workflows/json-yaml-validate.yml
@@ -10,7 +10,6 @@ on:
 
 permissions:
   contents: read
-  pull-requests: write # enable write permissions for pull request comments
 
 jobs:
   json-yaml-validate:
--- a/.github/workflows/json-yaml-validate.yml
+++ b/.github/workflows/json-yaml-validate.yml
@@ -21,5 +21,4 @@ jobs:
       - name: json-yaml-validate
         uses: GrantBirki/json-yaml-validate@v1.2.0
         with:
-          comment: "true" # enable comment mode
           yaml_exclude_regex: "(/static/examples/.*(usage|metallb-config).yaml|.*invalid-yaml)"
--- a/.github/workflows/json-yaml-validate.yml
+++ b/.github/workflows/json-yaml-validate.yml
@@ -22,3 +22,4 @@ jobs:
         uses: GrantBirki/json-yaml-validate@v1.2.0
         with:
           comment: "true" # enable comment mode
+          yaml_exclude_regex: "(/static/examples/.*(usage|metallb-config).yaml|.*invalid-yaml)"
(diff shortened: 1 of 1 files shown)


BenTheElder:
Hmm we have relatively few of these and a LOT of CI jobs already. Currently linting runs in pull-kind-verify and can be replicated locally (make verify).

We use actions to run CI with VMs more easily but generally we use Kubernetes' CI which has some properties like ensuring the latest commits are tested before merge and giving contributors a command to rerun tests.

.github/workflows/json-yaml-validate.yml
```diff
@@ -0,0 +1,25 @@
...
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  pull-requests: write # enable write permissions for pull request comments
```
BenTheElder, line 13: So we're granting write permission to an external repo at a tag ...?

pretty sure we already turned off write permissions for all actions at the repo level

GrantBirki: We can remove that `pull-requests: write` line and disable the Action's ability to "comment" back on PRs to lock things down a bit more. Can optionally pin the commit sha rather than using a tag as well.

I also did not notice the `make verify` command and if that is sufficient, then perhaps this PR is not needed. Just let me know and I can either leave it open or close it out 👍

.github/workflows/json-yaml-validate.yml
```diff
@@ -0,0 +1,25 @@
...
+jobs:
+  json-yaml-validate:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: json-yaml-validate
+        uses: GrantBirki/json-yaml-validate@v1.2.0
+        with:
+          comment: "true" # enable comment mode
```
GrantBirki, line 24:
```suggestion
```

Removing the ability to write comments to PRs

.github/workflows/json-yaml-validate.yml
```diff
@@ -0,0 +1,24 @@
...
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  pull-requests: write # enable write permissions for pull request comments
```
GrantBirki, line 13:
```suggestion
```

Removing PR write permissions

PR #3142, opened 2023-03-30
update containerd / runc / crictl

BenTheElder:
notably runc 1.1.5 with 3 vulns


BenTheElder:
for this release my plan is:
- let the postsubmit auto image builds we setup run
- in preparation for future releases / development, file a PR where initially we switch to the staging-GCR auto built new image(s)
- manually `crane cp` the image(s) from gcr to dockerhub, so nothing changes this release and we don't point lots of user traffic at k8s registry
- update the PR(s) to use the dockerhub copies

all of that will happen after this merges

FYI @aojea

BenTheElder:
Build passing. On to postsubmit.

e2es won't work here. Need to consider tweaking CI to skip them for docs and images only changes

BenTheElder:
`crane cp gcr.io/k8s-staging-kind/base:v20230330-89a4b81b kindest/base:v20230330-89a4b81b`

PR #3143, opened 2023-03-30
bump kindnetd dependencies

BenTheElder:
this should also trigger a build with latest go patch

ref https://github.com/kubernetes-sigs/kind/pull/3142#issuecomment-1489555245 for updated plan to actually pickup the new image

PR #3146, opened 2023-03-30
increase machine size for auto-builds

BenTheElder:
one of these took almost an hour to build a small image for two architectures

https://github.com/kubernetes-sigs/kind/pull/3143#issuecomment-1489678463

see e.g. https://github.com/kubernetes/ingress-nginx/blob/c84ae78bdfabb1f93bf0e87cdff34c75c744b8df/images/nginx/cloudbuild.yaml#L5

### DIFF
--- a/images/base/cloudbuild.yaml
+++ b/images/base/cloudbuild.yaml
@@ -1,6 +1,7 @@
 # See https://cloud.google.com/cloud-build/docs/build-config
 options:
   substitution_option: ALLOW_LOOSE
+  machineType: N1_HIGHCPU_8
 steps:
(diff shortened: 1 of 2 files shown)


BenTheElder:
no presubmit CI for these ...

BenTheElder:
much better for the base:
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-kind-push-base-image/1641300162355662848

better but still a little slow for kindnetd:
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-kind-push-kindnetd-image/1641300162452131840

PR #3151, opened 2023-03-30
bump main binary to 1.20.2

BenTheElder:
last patch before release. almost done writing relnotes


BenTheElder:
generally working fine, see some incoming e2e flakes on older releases that are the cluster, no kind go version bugs.

PR #3155, opened 2023-03-31
cleanup base image a bit

BenTheElder:
remove stray checksum file after usage
drop rsync (we don't use that anymore)
drop socat (containerd should be doing this with go now thanks to @aojea  and dockershim is long gone)
re-arrange installed packages and update doc comment
update containerd to tagged release


BenTheElder:
[not sure what test flaked but FWIW only the build and verify jobs are relevant to base image source changes now, we build after merge and then PR updating]

BenTheElder:
I think we want to get back on a tagged containerd release in particular, pushing an image

BenTheElder:
pull-kind-e2e* will be testing this now

BenTheElder:
all, green, rolling forward with stable containerd

PR #3156, opened 2023-03-31
Added platform flag to docker command in site Makefile

RijulGulati:
A minor fix. On M1 Mac, running `make build` gives the following warning -

`WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested`

Specifying platform explicitly resolves the same.

### DIFF
--- a/site/Makefile
+++ b/site/Makefile
@@ -2,7 +2,7 @@
 DOCKER       = docker
 HUGO_VERSION = 0.60.0
 DOCKER_IMAGE = jojomi/hugo:$(HUGO_VERSION)
-DOCKER_RUN   = $(DOCKER) run --rm --interactive --tty --volume $(realpath $(CURDIR)/..):/src -p 1313:1313 --workdir /src/site --entrypoint=hugo $(DOCKER_IMAGE)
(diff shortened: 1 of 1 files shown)


RijulGulati:
I wasn't sure if this is of value or not. So raised a PR anyway. Let me know what you think @aojea @BenTheElder.

site/Makefile
```diff
@@ -2,7 +2,7 @@
 DOCKER       = docker
 HUGO_VERSION = 0.60.0
 DOCKER_IMAGE = jojomi/hugo:$(HUGO_VERSION)
-DOCKER_RUN   = $(DOCKER) run --rm --interactive --tty --volume $(realpath $(CURDIR)/..):/src -p 1313:1313 --workdir /src/site --entrypoint=hugo $(DOCKER_IMAGE)
+DOCKER_RUN   = $(DOCKER) run --rm --interactive --tty --volume $(realpath $(CURDIR)/..):/src -p 1313:1313 --workdir /src/site --entrypoint=hugo --platform linux/amd64 $(DOCKER_IMAGE)
```
BenTheElder, line 5: TODO: switch to a multi-arch image (may also require updating the hugo version, which would be more involved to check nothing broke)

BenTheElder (commented):
thanks!
/lgtm
/approve

site/Makefile
RijulGulati: Should I create an issue for this? In order to track it better? I can take it up.

PR #3157, opened 2023-04-02
Delete ioutil as it is Deprecated since Go 1.16

yanggangtony:
Delete ioutil as it is Deprecated since Go 1.16
[ioutil info](https://cs.opensource.google/go/go/+/master:src/io/ioutil/ioutil.go).
```
// Deprecated: As of Go 1.16, the same functionality is now provided
// by package [io] or package [os], and those implementations
// should be preferred in new code.
// See the specific function documentation for details.
```

We can use os and io package for the same logic in the future.

### DIFF
--- a/pkg/fs/fs.go
+++ b/pkg/fs/fs.go
@@ -21,7 +21,6 @@ package fs
 
 import (
 	"io"
-	"io/ioutil"
 	"os"
 	"path"
 	"path/filepath"
@@ -29,10 +28,10 @@ import (
 	"strings"
 )
 
-// TempDir is like ioutil.TempDir, but more docker friendly
+// TempDir is like os.MkdirTemp, but more docker friendly
 func TempDir(dir, prefix string) (name string, err error) {
 	// create a tempdir as normal
-	name, err = ioutil.TempDir(dir, prefix)
+	name, err = os.MkdirTemp(dir, prefix)
 	if err != nil {
 		return "", err
 	}
@@ -138,14 +137,18 @@ func copyDir(src, dst string, info os.FileInfo) error {
 		return err
 	}
 	// copy every source dir entry
-	entries, err := ioutil.ReadDir(src)
+	entries, err := os.ReadDir(src)
 	if err != nil {
 		return err
 	}
 	for _, entry := range entries {
 		entrySrc := filepath.Join(src, entry.Name())
 		entryDst := filepath.Join(dst, entry.Name())
-		if err := copy(entrySrc, entryDst, entry); err != nil {
+		fileInfo, err := entry.Info()
+		if err != nil {
+			return err
+		}
+		if err := copy(entrySrc, entryDst, fileInfo); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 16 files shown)


stmcginnis:
The kind project still [supports Go versions back to 1.14](https://github.com/kubernetes-sigs/kind/blob/main/go.mod#L3). The replacements did not exist in the older releases. It should eventually be raised to only support newer go versions, but until that decision is made, this needs to stay with using `ioutils`.

BenTheElder:
Right, importantly KIND is embeddable as a library for [testing purposes](https://kind.sigs.k8s.io/docs/contributing/project-scope/), so we take great care in changing the requirements for library users (including adding more dependencies, bumping go version required, etc.)

It might be reasonable to start requiring go1.16, but we should actually do so by changing the [go.mod go statement](https://go.dev/doc/modules/gomod-ref#go) if we're going to. Go 1.16 came out February 2021 about a year after Go 1.14. While go only officially supports 1.19 and 1.20 currently, until very recently Kubernetes itself did not bump go versions on release branches so we could not drop version support as quickly as Go does upstream.

Does this improve anything immediately? Are there any bug fixes we're missing because we haven't moved?
Are these 100% compatible rewrites?

yanggangtony:
@stmcginnis @BenTheElder 
Thanks very much for the quick review..


Yes , you are right .
Now the go.mod dependencie is go1.14 . 
This pr will hold on until we support go's high version.
At that time , the go.mod and the new's dependencie will import .

pkg/fs/fs.go
```diff
@@ -138,14 +137,15 @@ func copyDir(src, dst string, info os.FileInfo) error {
...
-	entries, err := ioutil.ReadDir(src)
+	entries, err := os.ReadDir(src)
 	if err != nil {
 		return err
 	}
 	for _, entry := range entries {
 		entrySrc := filepath.Join(src, entry.Name())
 		entryDst := filepath.Join(dst, entry.Name())
-		if err := copy(entrySrc, entryDst, entry); err != nil {
+		fileInfo, _ := entry.Info()
```
BenTheElder, line 147: we shouldn't discard this error?

BenTheElder (commented):
I see at least one case that isn't clearly equivalent, otherwise I think this looks good and we can do this in the next kind release.

Let's update go.mod to 1.16

pkg/fs/fs.go
yanggangtony: ```entries, err := os.ReadDir(src)```
In the above code , If the err is nil , Maybe deduction the ```entry.Info()``` will return normal code without err .

But you are right , we should consider the error side..
I will update it later...

yanggangtony:
Another commit to bump go.mod to 1.16..

BenTheElder:
Hmm that's a concerning timeout https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/3157/pull-kind-e2e-kubernetes-1-27/1649154486473592832

I don't think it's related to this PR but ... that's also not something I'd been seeing. It shouldn't take that long to build.

BenTheElder (commented):
/lgtm
/approve
thanks!

PR #3161, opened 2023-04-07
Ensures that images are pulled from the insecure registry with http

essobedo:
Related to https://github.com/kubernetes-sigs/kind/issues/110

## Motivation

Inside k8s, the images are pulled from the insecure registry using HTTPS while they should be pulled with HTTP.

## Modifications:

* Modifies the registry mirrors' configuration to ensure that images pulled from `kind-registry` are pulled using HTTP 
* Defines the value of `hostFromClusterNetwork` of the `localRegistryHosting`
* Updates the documentation consequently

### DIFF
--- a/site/content/docs/user/local-registry.md
+++ b/site/content/docs/user/local-registry.md
@@ -25,7 +25,7 @@ The registry can be used like this.
 1. First we'll pull an image `docker pull gcr.io/google-samples/hello-app:1.0`
 2. Then we'll tag the image to use the local registry `docker tag gcr.io/google-samples/hello-app:1.0 localhost:5001/hello-app:1.0`
 3. Then we'll push it to the registry `docker push localhost:5001/hello-app:1.0`
-4. And now we can use the image `kubectl create deployment hello-server --image=localhost:5001/hello-app:1.0`
+4. And now we can use the image `kubectl create deployment hello-server --image=kind-registry:5000/hello-app:1.0`
 
 If you build your own image and tag it like `localhost:5001/image:foo` and then use
-it in kubernetes as `localhost:5001/image:foo`. And use it from inside of your cluster application as `kind-registry:5000`.
+it in kubernetes as `kind-registry:5000/image:foo`. And use it from inside of your cluster application as `kind-registry:5000`.
--- a/site/static/examples/kind-with-registry.sh
+++ b/site/static/examples/kind-with-registry.sh
@@ -16,7 +16,7 @@ kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 containerdConfigPatches:
 - |-
-  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."localhost:${reg_port}"]
+  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."${reg_name}:5000"]
     endpoint = ["http://${reg_name}:5000"]
 EOF
 
@@ -36,6 +36,7 @@ metadata:
 data:
   localRegistryHosting.v1: |
     host: "localhost:${reg_port}"
+    hostFromClusterNetwork: "${reg_name}:5000"
     help: "https://kind.sigs.k8s.io/docs/user/local-registry/"
 EOF
 


site/static/examples/kind-with-registry.sh
```diff
@@ -36,6 +36,7 @@ metadata:
 data:
   localRegistryHosting.v1: |
     host: "localhost:${reg_port}"
+    hostFromClusterNetwork: "${reg_name}:5000"
```
BenTheElder, line 39: neat!

BenTheElder:
Catching up after being on vacation for a week and some other kubernetes priorities ... Thanks for the PR!

Unfortunately we can't merge _anything_ without the CNCF's CLA 😅

essobedo:
@BenTheElder The CLA is signed now, it took me some time

site/content/docs/user/local-registry.md
```diff
@@ -25,7 +25,7 @@ The registry can be used like this.
 1. First we'll pull an image `docker pull gcr.io/google-samples/hello-app:1.0`
 2. Then we'll tag the image to use the local registry `docker tag gcr.io/google-samples/hello-app:1.0 localhost:5001/hello-app:1.0`
 3. Then we'll push it to the registry `docker push localhost:5001/hello-app:1.0`
-4. And now we can use the image `kubectl create deployment hello-server --image=localhost:5001/hello-app:1.0`
+4. And now we can use the image `kubectl create deployment hello-server --image=kind-registry:5000/hello-app:1.0`
```
BenTheElder, line 28: this change is unrelated and localhost:5001 should work.

the goal is to ensure that a consistent name is used for the images across hosts to avoid confusion for users.

site/static/examples/kind-with-registry.sh
```diff
@@ -16,7 +16,7 @@ kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 containerdConfigPatches:
 - |-
-  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."localhost:${reg_port}"]
+  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."${reg_name}:5000"]
```
BenTheElder, line 19: this change is wrong

BenTheElder: this configuration as-is allows us to remap pulls for images at `localhost:$registry_port` to the registry container, for a consistent reference between host and nodes

you can remap anything to anything, this is remapping is performed in the image resolution code in containerd.

site/content/docs/user/local-registry.md
```diff
@@ -25,7 +25,7 @@ The registry can be used like this.
...
 
 If you build your own image and tag it like `localhost:5001/image:foo` and then use
-it in kubernetes as `localhost:5001/image:foo`. And use it from inside of your cluster application as `kind-registry:5000`.
+it in kubernetes as `kind-registry:5000/image:foo`. And use it from inside of your cluster application as `kind-registry:5000`.
```
BenTheElder, line 31: same here: this doesn't need to change. the only time you need to use kind-registry instead is if you're in the docker network and not using the node container runtime

site/static/examples/kind-with-registry.sh
BenTheElder: so containerd sees `localhost:5001/foo` and instead it pulls from the endpoint we specify here.

BenTheElder:
> Inside k8s, the images are pulled from the insecure registry using HTTPS while they should be pulled with HTTP.

looking at the config again now, they are not, it is configured as http https://github.com/kubernetes-sigs/kind/blob/5cb68df9498344d5e3e33d1c9e073af67a0108ed/site/static/examples/kind-with-registry.sh#L20

I think there may be some confusion here either about how the containerd feature works or about pulling images from something other than the node runtimes.

essobedo:
I confirm that the current code works for the use case provided in the article https://kind.sigs.k8s.io/docs/user/local-registry/ but not in my use case where I need to be able to push an image from a pod with this approach we can access the local registry in the exact same manner inside the k8s cluster but I understand that it is probably too specific.

Issue #3163, opened 2023-04-09
Support command flag completions


iyear:
**What would you like to be added**:

Add flag completions for kind CLI.

**Why is this needed**:

Many commands in `kind` require the setting of command flags, but there are no completions, which causes some inconvenience in use:
For example, for `kind delete cluster -n`, I have to use `kind get clusters` to get the name first. After adding flag completion, I can just auto complete the name I want.

iyear:
And I'm also willing to work for it:)

iyear:
Hi, any news?

iyear:
They are basic completions. I want to be able to dynamically complete it when the flag is used. Such as `kind delete cluster -n [tab][tab]` and it displays all clusters name for completion

aojea:
ok, send a PR and we see how it looks like

PR #3166, opened 2023-04-13
feat: delete cluster flag completion

iyear:
#3163

Preview: 

[asciicast](https://asciinema.org/a/Af951OpEqU6MntMjEMqzYWBPA)

### DIFF
--- a/pkg/cmd/kind/delete/cluster/deletecluster.go
+++ b/pkg/cmd/kind/delete/cluster/deletecluster.go
@@ -49,13 +49,9 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 		},
 	}
 
-	const (
-		nameFlag = "name"
-	)
-
 	cmd.Flags().StringVarP(
 		&flags.Name,
-		nameFlag,
+		"name",
 		"n",
 		cluster.DefaultName,
 		"the cluster name",
@@ -67,7 +63,7 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 		"sets kubeconfig path instead of $KUBECONFIG or $HOME/.kube/config",
 	)
 
-	cobra.CheckErr(cmd.RegisterFlagCompletionFunc(nameFlag, func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	cobra.CheckErr(cmd.RegisterFlagCompletionFunc("name", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 		clusters, err := cluster.NewProvider(
 			cluster.ProviderWithLogger(logger),
 			runtime.GetDefault(logger),
--- a/pkg/cmd/kind/delete/cluster/deletecluster.go
+++ b/pkg/cmd/kind/delete/cluster/deletecluster.go
@@ -48,9 +48,14 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 			return deleteCluster(logger, flags)
 		},
 	}
+
+	const (
+		nameFlag = "name"
+	)
+
 	cmd.Flags().StringVarP(
 		&flags.Name,
-		"name",
+		nameFlag,
 		"n",
 		cluster.DefaultName,
 		"the cluster name",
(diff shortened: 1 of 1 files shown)


pkg/cmd/kind/delete/cluster/deletecluster.go
```diff
@@ -48,9 +48,14 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 			return deleteCluster(logger, flags)
 		},
 	}
+
+	var (
+		name = "name"
```
BenTheElder, line 53: Nit: This should be a `const` not a var and it should have a more descriptive name or else just explicitly paste `"name"` into each location. It's not really clearer to have `const name = "name" and it's strictly worse to have it as a var.

pkg/cmd/kind/delete/cluster/deletecluster.go
```diff
@@ -61,6 +66,19 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 		"",
 		"sets kubeconfig path instead of $KUBECONFIG or $HOME/.kube/config",
 	)
+
+	cobra.CheckErr(cmd.RegisterFlagCompletionFunc(name, func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		clusters, err := cluster.NewProvider(
+			cluster.ProviderWithLogger(logger),
```
BenTheElder, line 72: What happens when podman support is stable and we move from `KIND_EXPERIMENTAL_PROVIDER` to `KIND_PROVIDER` as a default for a `--provider` flag? Can flag completion check for other flag values?

This feature seems like it may unfortunately be at least as problematic to maintain as it is useful.

Using `KIND_CLUSTER_NAME` is another option to get around repeatedly typing these.

pkg/cmd/kind/delete/cluster/deletecluster.go
iyear: Thanks. Updated.

pkg/cmd/kind/delete/cluster/deletecluster.go
iyear: > Can flag completion check for other flag values?

Yes it can.

I think the best solution is to extract ·get clusters· as a common function. This way we can only maintain the code in the same place

BenTheElder: > Does this mean that the implementation of kind get clusters will change? The code in PR is copied from get clusters

Yes. It's defaulting based on availability or the env currently.

BenTheElder (commented):
sorry, i missed that my review comments were left in pending state

pkg/cmd/kind/delete/cluster/deletecluster.go
```diff
@@ -48,9 +48,14 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 			return deleteCluster(logger, flags)
 		},
 	}
+
+	const (
+		nameFlag = "name"
```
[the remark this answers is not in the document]
tao12345666333: +1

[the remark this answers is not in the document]
iyear: Agreed. Done

pkg/cmd/kind/delete/cluster/deletecluster.go
iyear: And how about `extract kind get clusters runE as a common function` ? IMO, the completion is like an alias of `get clusters`.

BenTheElder: Sure, I've just never used a tool where one flag completed based on another flag and didn't know if cobra can support this. If it can then let's go ahead :-)

stmcginnis:
This looks reasonable. Completion would be useful here.

/lgtm

BenTheElder:
cc @aojea, need another pair of eyes on this. trying this year to get through the backlog of things that fell through the cracks  😅

stmcginnis:
@aojea @BenTheElder reminder on this. I think it would be good to either get this through, or decide we aren't going to do it. I do think it's useful though, so it would be a nice usability improvement.

stmcginnis:
@aojea @BenTheElder We should really decide if we're going to move ahead with this or not. My personal opinion is it could improve the end user experience, for little risk, so probably worth adding.

PR #3172, opened 2023-04-17
ci: hack/release/get-contributors.sh: Fix incorrect token identification

wzshiming:
Currently, TOKEN is set and not work

### DIFF
--- a/hack/release/get-contributors.sh
+++ b/hack/release/get-contributors.sh
@@ -32,7 +32,7 @@ for contributor in "${contributors[@]}"; do
     # lookup the  commit info to get the login
     contributor_logins+=("$(curl \
         -sG \
-        ${GITHUB_OAUTH_TOKEN:+-H="Authorization: token ${GITHUB_OAUTH_TOKEN:?}"} \
+        ${GITHUB_OAUTH_TOKEN:+-H "Authorization: Bearer ${GITHUB_OAUTH_TOKEN:?}"} \
         --data-urlencode "q=${contributor}" \
         "https://api.github.com/repos/${ORG}/${REPO}/commits/${commit_for_contributor}" \
     | jq -r .author.login


aojea:
it seems Bearer covers more cases

> Note: In most cases, you can use Authorization: Bearer or Authorization: token to pass a token. However, if you are passing a JSON web token (JWT), you must use Authorization: Bearer.

https://docs.github.com/en/rest/overview/authenticating-to-the-rest-api?apiVersion=2022-11-28#about-authentication

/lgtm

hack/release/get-contributors.sh
```diff
@@ -32,7 +32,7 @@ for contributor in "${contributors[@]}"; do
     # lookup the  commit info to get the login
     contributor_logins+=("$(curl \
         -sG \
-        ${GITHUB_OAUTH_TOKEN:+-H="Authorization: token ${GITHUB_OAUTH_TOKEN:?}"} \
+        ${GITHUB_OAUTH_TOKEN:+-H "Authorization: Bearer ${GITHUB_OAUTH_TOKEN:?}"} \
```
BenTheElder, line 35: we should not switch to a space here. this is a single string expansion and it's confusing to try to make it a whitespace split pair of strings.

BenTheElder:
Are you using this script somewhere? :O

Currently Antonio and I use it here, last I checked it worked fine with PAT.

hack/release/get-contributors.sh
wzshiming: When GITHUB_OAUTH_TOKEN is set, it does not work, I try to debug it, I added `-vv` to `crul` and retry that the Header `authorization` was prefixed by `=`

``` console
> Host: api.github.com
> user-agent: curl/7.87.0
> accept: */*
> =authorization: token
```

I tried curl on my Mac, and it looks like `curl -H` is the problem

```
$ curl --version
curl 7.87.0 (x86_64-apple-darwin22.0) libcurl/7.87.0 (SecureTransport) LibreSSL/3.3.6 zlib/1.2.11 nghttp2/1.51.0
Release-Date: 2022-12-21

$ curl -vv -H="x: x" http://example.org
*   Trying 93.184.216.34:80...
* Connected to example.org (93.184.216.34) port 80 (#0)
> GET / HTTP/1.1
> Host: example.org
> User-Agent: curl/7.87.0
> Accept: */*
> =x: x
>
```

I was worried that curl might not be compatible with linux on the Mac, so I tried it again on CentOS, and it behaved the same way

```
$ curl --version
curl 7.29.0 (x86_64-redhat-linux-gnu) libcurl/7.29.0 NSS/3.53.1 zlib/1.2.7 libidn/1.28 libssh2/1.8.0
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps pop3 pop3s rtsp scp sftp smtp smtps telnet tftp
Features: AsynchDNS GSS-Negotiate IDN IPv6 Largefile NTLM NTLM_WB SSL libz unix-sockets

$ curl -vv -H="x: x" http://example.org
* About to connect() to example.org port 80 (#0)
*   Trying 93.184.216.34...
* Connected to example.org (93.184.216.34) port 80 (#0)
> GET / HTTP/1.1
> User-Agent: curl/7.29.0
> Host: example.org
> Accept: */*
> =x: x
>
```

wzshiming:
> Are you using this script somewhere? :O
> 
> Currently Antonio and I use it here, last I checked it worked fine with PAT.

Yes, I plan to use this script in other repo

hack/release/get-contributors.sh
BenTheElder: TIL, that's strange flag behavior, thanks

PR #3175, opened 2023-04-18
Raise error if host-ports are configured again with same listen address and protocol

### LINKED ISSUE #2975: validate that hostports are not configured twice with the same protocol
cc @aojea (thinking of that kubernetes issue)

daman1807:
fixes #2975

### DIFF
--- a/pkg/internal/apis/config/validate.go
+++ b/pkg/internal/apis/config/validate.go
@@ -23,6 +23,7 @@ import (
 	"strings"
 
 	"sigs.k8s.io/kind/pkg/errors"
+	"sigs.k8s.io/kind/pkg/internal/sets"
 )
 
 // similar to valid docker container names, but since we will prefix
@@ -118,15 +119,76 @@ func (n *Node) Validate() error {
 		if err := validatePort(mapping.HostPort); err != nil {
 			errs = append(errs, errors.Wrapf(err, "invalid hostPort"))
 		}
+
 		if err := validatePort(mapping.ContainerPort); err != nil {
 			errs = append(errs, errors.Wrapf(err, "invalid containerPort"))
 		}
+	}
+
+	if err := validatePortMappings(n.ExtraPortMappings); err != nil {
+		errs = append(errs, errors.Wrapf(err, "invalid portMapping"))
 	}
 
 	if len(errs) > 0 {
 		return errors.NewAggregate(errs)
 	}
 
+	return nil
+}
+
+func validatePortMappings(portMappings []PortMapping) error {
+	errMsg := "port mapping with same listen address, port and protocol already configured"
+
+	wildcardAddrIPv4 := net.ParseIP("0.0.0.0")
+	wildcardAddrIPv6 := net.ParseIP("::")
+
+	// bindMap has the following key-value structure
+	// PORT/PROTOCOL: [ IP ]
+	// { 80/TCP: [ 127.0.0.1, 192.168.2.3 ], 80/UDP: [ 0.0.0.0 ] }
+	bindMap := make(map[string]sets.String)
+
+	formatPortProtocol := func(port int32, protocol PortMappingProtocol) string {
+		return fmt.Sprintf("%d/%s", port, protocol)
+	}
+
+	for _, portMapping := range portMappings {
+		addr := net.ParseIP(portMapping.ListenAddress)
+		addrString := addr.String()
+
+		portProtocol := formatPortProtocol(portMapping.HostPort, portMapping.Protocol)
+		possibleErr := fmt.Errorf("%s: %s:%s", errMsg, addrString, portProtocol)
+
+		// in golang 0.0.0.0 and [::] are equivalent, convert [::] -> 0.0.0.0
+		// https://github.com/golang/go/issues/48723
+		if addr.Equal(wildcardAddrIPv6) {
+			addr = wildcardAddrIPv4
+			addrString = addr.String()
+		}
+
+		if _, ok := bindMap[portProtocol]; ok {
+
+			// wildcard address case:
+			// return error if there already exists any listen address for same port and protocol
+			if addr.Equal(wildcardAddrIPv4) {
+				if bindMap[portProtocol].Len() > 0 {
+					return possibleErr
+				}
+			}
+
+			// direct duplicate & wild card present check:
+			// return error if same combination of ip, port and protocol already exists in bindMap.
+			// return error if wildcard address is already present for same port & protocol
+			if bindMap[portProtocol].Has(addrString) || bindMap[portProtocol].Has(wildcardAddrIPv4.String()) {
+				return possibleErr
+			}
+		} else {
+			// initialize the set
+			bindMap[portProtocol] = sets.NewString()
+		}
+
+		// add the entry to bindMap
+		bindMap[portProtocol].Insert(addrString)
+	}
 	return nil
 }
 
--- a/pkg/internal/apis/config/validate_test.go
+++ b/pkg/internal/apis/config/validate_test.go
@@ -17,6 +17,8 @@ limitations under the License.
 package config
 
 import (
+	"fmt"
+	"sigs.k8s.io/kind/pkg/internal/assert"
 	"testing"
 
 	"sigs.k8s.io/kind/pkg/errors"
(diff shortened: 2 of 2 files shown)


pkg/cluster/internal/providers/common/util.go
```diff
@@ -0,0 +1,76 @@
...
+
+package common
+
+// AddressPortProtocolMap is used to keep track of declared listen address,
+// port and protocol in node extra port mappings.
+type AddressPortProtocolMap map[string]map[int32]map[string]struct{}
+
+// IsAddrPortProtocolReconfigured checks if a combination of listen address, port and protocol exists in the given map.
+// Also takes care of special cases like 0.0.0.0 wildcard address.
+func IsAddrPortProtocolReconfigured(addrPortProtocolMap AddressPortProtocolMap, addr string, port int32, protocol string) bool {
```
BenTheElder, line 25: Instead of exporting in common and doing in the providers, we should do this at config validation time?

daman1807: That make sense, will update it.

pkg/cluster/internal/providers/common/util.go
```diff
@@ -0,0 +1,76 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+// AddressPortProtocolMap is used to keep track of declared listen address,
+// port and protocol in node extra port mappings.
+type AddressPortProtocolMap map[string]map[int32]map[string]struct{}
```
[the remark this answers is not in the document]
daman1807: Do you mean like defining AddressPortProtocolMap structure with the nested map as a property and calling  pointer receivers on AddressPortProtocolMap ?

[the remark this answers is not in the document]
neolit123: never mind.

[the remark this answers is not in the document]
BenTheElder: I don't think this type is necessary at all. You can do a sets.String of protocol + net.JoinHostPort(hostAddr, hostPort). It will take less memory and less code.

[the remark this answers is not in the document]
BenTheElder: https://github.com/kubernetes-sigs/kind/tree/main/pkg/internal/sets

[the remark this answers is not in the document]
BenTheElder: we should probably round trip through parsing the IPs before joining them because there's multiple ways to represent the same IP address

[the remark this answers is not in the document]
daman1807: should I use [ParseIPSloppy](https://github.com/kubernetes/utils/blob/d93618cff8a2/net/parse.go#L27) or [ParseIP](https://pkg.go.dev/net#ParseIP) is good enough for this ?

pkg/internal/apis/config/validate.go
```diff
@@ -130,6 +140,68 @@ func (n *Node) Validate() error {
...
 }
 
+func formatPortMapping(addr net.IP, port int, protocol PortMappingProtocol) string {
+	return fmt.Sprintf("%s/%s",
+		net.JoinHostPort(addr.String(), strconv.Itoa(port)),
+		protocol,
+	)
+}
+
+func validatePortMapping(portMappingsSet sets.String, addrSet sets.String, mapping PortMapping) error {
```
BenTheElder, line 150: probably should be pointers to `sets.String`?

BenTheElder (commented):
thanks, I'm going to defer to antonio as the real networking expert :-)

pkg/internal/apis/config/validate.go
daman1807: `sets.String` is a map, and will be passed as a reference only 😇

BenTheElder: I guess we can depend on that since we have our own forked copy. If we were using a third-party set type I wouldn't depend on that detail.

pkg/internal/apis/config/validate_test.go
```diff
@@ -428,3 +431,69 @@ func TestPortValidate(t *testing.T) {
...
+			fmt.Sprintf("%s: 127.0.0.1:3000/TCP", errMsg)},
+
+		{newPortMapping("::", 5000, "TCP"), ""},
+
+		// error expected: subset of ::[::1] is already defined for same port and protocol
+		{newPortMapping("::1", 5000, "TCP"),
+			fmt.Sprintf("%s: [::1]:5000/TCP", errMsg)},
+	}
+
+	for _, tc := range cases {
```
aojea, line 484: use t.Run to avoid test pollution

pkg/internal/apis/config/validate_test.go
```diff
@@ -428,3 +431,69 @@ func TestPortValidate(t *testing.T) {
...
+	newPortMapping := func(addr string, port int, protocol string) PortMapping {
+		return PortMapping{
+			HostPort:      int32(port),
+			ListenAddress: addr,
+			Protocol:      PortMappingProtocol(protocol),
+		}
+	}
+	errMsg := "port mapping with same listen address, port and protocol already configured"
+	cases := []struct {
+		mapping   PortMapping
```
aojea, line 448: add a name field and use named variables , take previous test as an example, is easier to read and go through the test cases

pkg/internal/apis/config/validate_test.go
```diff
@@ -428,3 +431,141 @@ func TestPortValidate(t *testing.T) {
...
+			},
+			expectErr: "",
+		},
+		{
+			testName: "unique port mappings ipv4 & ipv6",
+			portMappings: []PortMapping{
+				newPortMapping("127.0.0.1", 80, "TCP"),
+				newPortMapping("0.0.0.0", 80, "UDP"),
+				newPortMapping("::1", 80, "TCP"),
+				newPortMapping("::", 80, "UDP"),
```
aojea, line 475: this should fail , in golang `::` and `0.0.0.0` are equivalent, https://github.com/golang/go/issues/48723

pkg/internal/apis/config/validate.go
```diff
@@ -113,14 +115,22 @@ func (n *Node) Validate() error {
...
 	for _, mapping := range n.ExtraPortMappings {
 		if err := validatePort(mapping.HostPort); err != nil {
 			errs = append(errs, errors.Wrapf(err, "invalid hostPort"))
 		}
+
 		if err := validatePort(mapping.ContainerPort); err != nil {
 			errs = append(errs, errors.Wrapf(err, "invalid containerPort"))
 		}
+
+		if err := validatePortMapping(portMappingsSet, addrSet, mapping); err != nil {
```
aojea, line 131: ok, I think you got the mechanics but this implementation inside this loop is confusing because we are really cross-validating between the different portmaps, and this loop is aggregating those inside that is not obvious

You should make  `validatePortMapping` to accept the array of `n.ExtraPortMappings ` directly, and create the sets and all inside it, moving the function outside of this loop

pkg/internal/apis/config/validate.go
```diff
@@ -130,6 +140,68 @@ func (n *Node) Validate() error {
...
+func formatPortMapping(addr net.IP, port int, protocol PortMappingProtocol) string {
+	return fmt.Sprintf("%s/%s",
+		net.JoinHostPort(addr.String(), strconv.Itoa(port)),
+		protocol,
+	)
+}
+
+func validatePortMapping(portMappingsSet sets.String, addrSet sets.String, mapping PortMapping) error {
+	wildcardAddrIPv4 := net.ParseIP("0.0.0.0")
+	wildcardAddrIPv6 := net.ParseIP("::")
```
aojea, line 152: for goalng is the same, so you can simplify into one unique variable `wildcardAdd` , left a comment below with a reference to a golang issue explaining it

pkg/internal/apis/config/validate.go
```diff
@@ -130,6 +140,68 @@ func (n *Node) Validate() error {
...
+
+	addr := net.ParseIP(mapping.ListenAddress)
+	port := int(mapping.HostPort)
+	protocol := mapping.Protocol
+	ipFamily := getIPFamily(addr)
```
aojea, line 157: having the address we don't really need the ip family, do we?

daman1807: that was based on the assumption that `::` and `0.0.0.0` are different /

pkg/internal/apis/config/validate.go
```diff
@@ -130,6 +140,68 @@ func (n *Node) Validate() error {
...
+
+	portMappingString := formatPortMapping(addr, port, protocol)
```
aojea, line 159: so, you have two dimensions "port-protocol" and then address, if you aggregate first by port protocol this seems it will be more simplers

key = port-protocol
and you create a map with key the port protocol and values the listen addresses
map[key][]string
then if there is a wildcard in the value and len(string) > 1 you have a collision
if there are 2 values with the same address in the string then you have a collision

I think you can also use a set instead of the string slice to use the existing methods to detect the collisions while processing these structures

pkg/internal/apis/config/validate.go
daman1807: keeping both the wild card addrs as `net.ParseIP("0.0.0.0").Equal(net.ParseIP("::"))` will return false in golang and the user might provide port mapping for both under the assumption that they won't overlap with each other and will bind to different IP protocols

aojea: yes, but for your validation porpuse see suggestion here https://github.com/kubernetes-sigs/kind/pull/3175#discussion_r1181035115

if you translate wildcard addresses to a constant and you build a map indexed by port/protocol, you must have only one wildecard in the value of the map, otherwise it means you'll have a collision problem

pkg/internal/apis/config/validate.go
```diff
@@ -118,18 +119,87 @@ func (n *Node) Validate() error {
...
+			// return error if there already exists any listen address for same port and protocol
+			if addr.Equal(wildcardAddrIPv4) {
+				if bindMap[portProtocol].Len() > 0 {
+					return possibleErr
+				}
+			}
+
+			// direct duplicate check:
+			// return error if same combination of ip, port and protocol already exists in bindMap.
+			if bindMap[portProtocol].Has(addrString) {
```
aojea, line 180: can not you simplify this to
```suggestion
			if bindMap[portProtocol].Has(addrString) || bindMap[portProtocol].Has(wildcardAddrIPv4.String()) {
```

aojea:
almost there, one last comment

aojea:
/lgtm
/approve

Thanks

PR #3178, opened 2023-04-18
use dnsSearch in CI

### LINKED ISSUE #1860: avoid excessive search lines in CI
There's not a good way to manage this in KIND, it's reasonable that we expect the host to have sane DNS and it's intentional that we use upstream DNS pointed at the host DNS resolution, but in CI we have additional search paths by way of running inside another cluster. We're already using FQDN for interacting with services in that cluster outside of KIND (namely the bazel build cache), so we shouldn't need search paths at all.

Basically we should mitigate this up front:
https://github.com/kubernetes/test-infra/issues/19080#issuecomment-685141853

/assign
/priority important-soon

xref: #303

BenTheElder:
fixes #1860 
thanks @howardjohn :-)

### DIFF
--- a/hack/ci/e2e-k8s.sh
+++ b/hack/ci/e2e-k8s.sh
@@ -151,6 +151,9 @@ apiVersion: kind.x-k8s.io/v1alpha4
 networking:
   ipFamily: ${IP_FAMILY:-ipv4}
   kubeProxyMode: ${KUBE_PROXY_MODE:-iptables}
+  # don't pass through host search paths
+  # TODO: possibly a reasonable default in the future for kind ...
+  dnsSearch: []
 nodes:
 - role: control-plane
 - role: worker


hack/ci/e2e-k8s.sh
```diff
@@ -151,6 +151,9 @@ apiVersion: kind.x-k8s.io/v1alpha4
 networking:
   ipFamily: ${IP_FAMILY:-ipv4}
   kubeProxyMode: ${KUBE_PROXY_MODE:-iptables}
+  # don't pass through host search paths
+  # TODO: possibly a reasonable default in the future for kind ...
```
BenTheElder, line 155: (in v1alpha5 that is)

aojea:
/lgtm

ref https://github.com/kubernetes-sigs/kind/pull/3098

PR #3192, opened 2023-05-02
feat(providers): add error when cluster does not exist

alegrey91:
When deleting clusters with `kind`, it could be nice to have a feedback from the application if the cluster was not found.
Sometimes happen that clusters remain up and running because no feedback is provided and at a first glance seems that the command did its job.
Here's the current behavior of kind when the cluster does not exist:
```
$ kind delete cluster --name cluster_does_not_exist
Deleting cluster "cluster_does_not_exist" ...
```
Here's how the feedback is given to the user with my implementation:
```
kind delete cluster --name cluster_does_not_exist
Deleting cluster "cluster_does_not_exist" ...
ERROR: failed to delete cluster "cluster_does_not_exist": error listing nodes: cluster does not exist
```

### DIFF
--- a/pkg/cluster/internal/providers/docker/provider.go
+++ b/pkg/cluster/internal/providers/docker/provider.go
@@ -108,6 +108,10 @@ func (p *provider) ListClusters() ([]string, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to list clusters")
 	}
+	// provide feedback to the user about cluster inexistence
+	if len(lines) == 0 {
+		return nil, errors.Errorf("cluster does not exist")
+	}
 	return sets.NewString(lines...).List(), nil
 }
 
@@ -124,6 +128,10 @@ func (p *provider) ListNodes(cluster string) ([]nodes.Node, error) {
 	lines, err := exec.OutputLines(cmd)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to list nodes")
+	}
+	// provide feedback to the user about cluster inexistence
+	if len(lines) == 0 {
+		return nil, errors.Errorf("cluster does not exist")
 	}
 	// convert names to node handles
 	ret := make([]nodes.Node, 0, len(lines))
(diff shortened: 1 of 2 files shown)


BenTheElder:
There's extensive past discussion of this in the issue tracker, in retrospect we might have made this different (perhaps a --force mode and encouraging that) but we can't make a substantial breaking change like this lightly and the benefits do not outweigh the breakage. We now log what actions were actually taken to help clarify the behavior.

alegrey91:
Hi @BenTheElder and thanks for your fast and exhaustive reply.
> If you'd like to do this when calling the library you can check List() isntead

Not sure I've got the point. Can you elaborate it more?
I'd really like to get a better feedback from the tool.
P.S. thanks for this amazing project.

BenTheElder:
After #2934 we log what actions are taken.

For library users you can make your own determination of "cluster exists" if for you that's only nodes.

It's NOT an error to try to delete cluster resources even if nodes don't exist, we don't only delete nodes. Changing the behavior of these function is a breaking change.

It is safe to do something like:

```bash

export KIND_CLUSTER_NAME=foo-test

trap 'kind delete cluster' exit
kind create cluster

run_my_tests

kind delete cluster
```

Those will log if any actions are taken and not fail if there's no action to take.

Also, even if all the node containers were deleted, deleting a cluster also cleans up the KUBECONFIG entry, and in the future could cover other resources.

stmcginnis:
See the first Note here: https://kind.sigs.k8s.io/docs/user/quick-start/#deleting-a-cluster

BenTheElder:
If we do find nodes and fail to delete them, that *is* an error and should produce one today. If the command fails to eliminate resources it will fail. If the resources have already been eliminated by some other process (e.g. a previous invocation of `kind delete cluster` or `docker rm`) then it will not.

This does mean that the command can succeed if you've failed to supply the correct cluster name, but consider also that users can currently delete one or more of the node containers using other tools and we still need to perform other kind-specific cleanup and remove any remaining nodes if they invoke `kind delete cluster`. I'm sure we have users depending on this in many interesting ways (e.g. testing behavior if they destroy a single node).

alegrey91:
Ok, I got the point of idempotency. Thanks for the deep explanation.
I'm going to close the pull request.

PR #3193, opened 2023-05-02
Update go and distroless-iptables

cpanato:
- bump go-version file to 1.20.4
- bump distroless-iptables to v0.2.3

/assign @aojea


go.mod
```diff
@@ -1,6 +1,6 @@
 module sigs.k8s.io/kind
 
-go 1.14
+go 1.20
```
cpanato, line 3: Any specific reason that was still in 1.14?

BenTheElder: Yes
/hold
See: https://github.com/kubernetes-sigs/kind/pull/3157

Note that go.mod go version implies the go feature level, NOT the go version used for building. 

That would be .go-version here in most cases.

.go-version
```diff
@@ -1 +1 @@
-1.20.2
+1.20.3
```
BenTheElder, line 1: This should be 1.20.4 today.

cpanato: they were faster :)

go.mod
cpanato: @BenTheElder i will drop this change for now to not block this PR, is that ok?

BenTheElder: Yes, please revert changes to the go.mod, we're probably going to bump to 1.16 soon but we're not using any go 1.20 features and this is risky for library consumers

BenTheElder:
thanks!

PR #3194, opened 2023-05-02
use git hash instead of git tag as best practice

cpanato:
- use git hash instead of git tag as best practice
is a best practice to use the hash and not a tag that can be mutable

/assign @aojea


BenTheElder (commented):
thanks I thought we'd gotten these already!
/lgtm
/approve

PR #3195, opened 2023-05-02
add dependabot config for Github actions

cpanato:
- add dependabot config for Github actions
to keep actions and the go dependencies up-to-date

/assign @aojea

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "area/dependency"
+      - "release-note-none"
+      - "ok-to-test"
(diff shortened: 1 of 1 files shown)


BenTheElder:
I'm actually kinda -1 on this.

It's easy to break things bumping to new versions without reviewing them carefully and not all dependency bumps are helpful, doing this also increases the minimum required version for consumers of kind as a library.

We don't do this in k/k either.

BenTheElder:
SGTM

cpanato:
@BenTheElder PTAL

BenTheElder (commented):
/lgtm
/approve

BenTheElder:
/hold
wait, will this clobber switching to hashes vs tags?

cpanato:
> /hold wait, will this clobber switching to hashes vs tags?

dependabot it will update the hashes and also the version comment :)

aojea:
Thanks Carlos for all these PRs

PR #3201, opened 2023-05-09
build all non-distro-packaged go binaries ourselves and respect .go-version

BenTheElder:
Pros:
- streamlines patching go version related CVEs
- ensures binaries with CGO are linked to the library version we'll be using
- allows controlling runc version independent of containerd version
- allows retiring https://github.com/kind-ci/containerd-nightlies, we can leave it as-is but we don't need to keep maintaining this vestigial hack and any contributor can PR changes to the containerd build as normal now
- allows us to control build-tags and compile out unused containerd snapshotters for some binary size savings
- ensures we can easily include LICENSE / notice info for binaries not coming from ubuntu packages

Cons:
- building the node image is more expensive, however we don't have to do this super often and we have automated builds now so we can throw cores at it and still have quick builds. The buildx multi-stage build is a DAG that can execute steps in parallel so with enough cores this runs pretty quickly.
- we had to put another copy of gimme in the image directory ... you can't use symlinks outside of the build context with docker build and we don't want the whole repo as a build context ... this is the path of least resistance for now. gimme doesn't really need updating for the forseeable future though so this is fine.
- dockerfile is somewhat more complex with more multi-stage steps, however, on the other hand these steps are typically actually simpler than the "fetch and check shasum" inline scripts, basically just clone and then `make`

TODO:
- Update image build postubmit jobs to trigger on `.go-version` and `images/Makefile.common.in` 
  - https://github.com/kubernetes/test-infra/pull/29474
- ~Rework `images/kindnetd`, `images/local-path-provisioner` to also respect `.go-version` w/ gimme~
  - DONE in 2nd and 3rd commits

When those are complete, go CVEs can be addressed in two PRs:
1. bump .go-version
2. bump references to images built as a result of merging 1)


images/base/Dockerfile
```diff
@@ -117,78 +82,119 @@ RUN echo "Installing Packages ..." \
...
+RUN git clone "${RUNC_CLONE_URL}" /runc \
+    && cd /runc \
+    && git checkout "${RUNC_VERSION}" \
+    && eval "$(gimme "${GO_VERSION}")" \
+    && make runc
+
+# stage for building crictl
+FROM go-build as build-crictl
+ARG GO_VERSION
+ARG CRICTL_CLONE_URL="https://github.com/kubernetes-sigs/cri-tools"
```
BenTheElder, line 134: this should be CRI_TOOLS

BenTheElder: eh, keeping these CRICTL_ maintains some level of compat with existing variables.

images/base/Dockerfile
```diff
@@ -117,78 +82,119 @@ RUN echo "Installing Packages ..." \
...
+ARG CRICTL_VERSION="v1.26.1"
+RUN git clone "${CRICTL_CLONE_URL}" /cri-tools \
+    && cd /cri-tools \
+    && git checkout "${CRICTL_VERSION}" \
+    && eval "$(gimme "${GO_VERSION}")" \
```
aojea, line 139: why eval gimme again? is not in the `build` image?

aojea:
cool

images/base/Dockerfile
BenTheElder: We call `gimme ${GO_VERSION}` in the `go-build` stage to fetch the toolchain (which then can be re-used by all the other stages that start `FROM go-build`). But the envs it sets are not persisted.

We call `eval $(gimme "${GO_VERSION}")` in the tooling specific stages before doing builds, in the same shell instance, to ensure that the go toolchain from gimme is used. gimme works by modifying `PATH` etc., when you eval it you're just setting those env to point to the go version it installed locally.

I forgot we could use `source ~/.gimme/envs/go$version.env` instead which might be a tiny bit faster.

https://github.com/travis-ci/gimme/blob/master/README.md

BenTheElder:
I've run `make -C images/base quick` locally and the image checks out, I haven't used it for a node image yet, but I don't have any reason to suspect issues there.

images/base/Dockerfile
```diff
@@ -117,78 +82,119 @@ RUN echo "Installing Packages ..." \
...
+    && go build -o ./bin/loopback -mod=vendor ./plugins/main/loopback \
+    && go build -o ./bin/ptp -mod=vendor ./plugins/main/ptp \
+    && go build -o ./bin/portmap -mod=vendor ./plugins/meta/portmap
+
+# stage for building containerd-fuse-overlayfs
+FROM go-build as build-fuse-overlayfs
+ARG GO_VERSION
+ARG CONTAINERD_FUSE_OVERLAYFS_VERSION="v1.0.5"
+ARG CONTAINERD_FUSE_OVERLAYFS_CLONE_URL="https://github.com/containerd/fuse-overlayfs-snapshotter"
+RUN git clone "${CONTAINERD_FUSE_OVERLAYFS_CLONE_URL}" /fuse-overlayfs-snapshotter \
```
aojea, line 162: ` --depth 1 ` ??

images/base/Dockerfile
```diff
@@ -117,78 +82,119 @@ RUN echo "Installing Packages ..." \
...
+
+# stage for building containerd
+FROM go-build as build-containerd
+ARG GO_VERSION
+ARG CONTAINERD_VERSION="v1.6.21"
+ARG CONTAINERD_CLONE_URL="https://github.com/containerd/containerd"
+# we don't build with optional snapshotters, we never select any of these
+# they're not ideal inside kind anyhow, and we save some disk space
+ARG BUILDTAGS=no_aufs,no_zfs,no_btrfs,no_devmapper
+RUN git clone "${CONTAINERD_CLONE_URL}" /containerd \
```
BenTheElder, line 114: We should do blobless clones for a minor speed up, most of these repos are small anyhow but it's low effort to add that flag

images/base/Dockerfile
BenTheElder: I added --filter=tree:0

BenTheElder:
I've updated this to also include LICENSE and notice files via https://github.com/google/go-license

Some other Kubernetes projects are doing this and we haven't been, this is easier to do now that we're cloning and building these packages.

debian packages that we're just installing the usual way already include comparable info in a standard way.

BenTheElder:
I've tested all of these builds locally in quick mode (single-architecture only w/o pushing).

images/base/Dockerfile
```diff
@@ -117,78 +83,133 @@ RUN echo "Installing Packages ..." \
...
+
+# shared stage to setup go version for building binaries
+FROM base as go-build
+COPY --chmod=0755 scripts/third_party/gimme/gimme /usr/local/bin/
+# tools needed at build-time only
+RUN clean-install git make libseccomp-dev gcc libc-dev pkg-config
+# set by makefile to .go-version
+ARG GO_VERSION
+RUN eval "$(gimme "${GO_VERSION}")" \
+    && GOBIN=/usr/local/bin go install github.com/google/go-licenses@latest
```
aojea, line 105: Latest?

BenTheElder: yes, the most recent tagged release. the API surface of this tool is small

images/base/Dockerfile
```diff
@@ -117,78 +83,133 @@ RUN echo "Installing Packages ..." \
...
+    && git checkout "${CONTAINERD_VERSION}" \
+    && eval "$(gimme "${GO_VERSION}")" \
+    && make bin/ctr bin/containerd bin/containerd-shim-runc-v2 \
+    && go-licenses save --save_path=/_LICENSES \
+        ./cmd/ctr ./cmd/containerd ./cmd/containerd-shim-runc-v2
+
+# stage for building runc
+FROM go-build as build-runc
+ARG GO_VERSION
+ARG RUNC_VERSION="v1.1.7"
```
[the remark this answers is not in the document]
BenTheElder: We did do that, but that version is just a recommendation / what they're using in CI in containerd*, we want to be able to control the pace of upgrading runc.

\* their own integration tests, maybe not k8s e2e ...

We may even have to split this up and use multiple versions depending on k8s version, see https://github.com/kubernetes/kubernetes/pull/117691#issuecomment-1529860122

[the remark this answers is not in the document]
BenTheElder: https://github.com/containerd/containerd/blob/c6d7e45c14896addf3b1e06db3f7999026b0191f/docs/RUNC.md#runc-version-requirements-for-containerd

It's recommended to use a version >= https://github.com/containerd/containerd/blob/main/script/setup/runc-version, but that may actually be broken with Kubernetes or we may want to be ahead.

runc versions will need more thought following the thread I linked above, but for now sticking to the current stable tag.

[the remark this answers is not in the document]
BenTheElder: should add: 1.1.7 is the version used by containerd 1.6.21 https://github.com/containerd/containerd/releases/tag/v1.6.21

semi-related: they're no longer including runc with containerd releases https://github.com/containerd/containerd/issues/6541

BenTheElder:
This has serious perf issues in the arm64 cross-build. Working on a fix

BenTheElder:
follow-up in https://github.com/kubernetes-sigs/kind/pull/3214

images/base/Dockerfile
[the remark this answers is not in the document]
BenTheElder: turns out we actually need to be able to stick to an older version at times. https://github.com/kubernetes-sigs/kind/pull/3220

older is fine, newer binary may not be. at least as long as we have to support cgroups v1

PR #3202, opened 2023-05-09
Update helpers.go

lixin963:
posixy -> posix


pkg/build/nodeimage/helpers.go
```diff
@@ -29,7 +29,7 @@ import (
 // ensuring the directory exists and writing contents to the file
 func createFile(containerCmder exec.Cmder, filePath, contents string) error {
 	// NOTE: the paths inside the container should use the path package
-	// and not filepath (!), we want posixy paths in the linux container, NOT
+	// and not filepath (!), we want posix paths in the linux container, NOT
```
BenTheElder, line 32: "Posixy" here meant "posix-like" as opposed to "strictly posix". Posix doesn't deeply specify this.

we should say "Linux style" or "Unix style"

PR #3214, opened 2023-05-10
docker images: cross build go binaries

BenTheElder:
_huge_ build speed-up for the arm64 build on amd64 hosts and vice versa

Notable change: Switched to debian base image from ubuntu. They have comparable enough packages but debian has *much* better cross platform package support.

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -20,7 +20,7 @@
 # start from ubuntu, this image is reasonably small as a starting point
 # for a kubernetes node image, it doesn't contain much we don't need
 # this stage will install basic files and packages
-ARG BASE_IMAGE=ubuntu:22.04
+ARG BASE_IMAGE=debian:bullseye-slim
 FROM $BASE_IMAGE as base
 
 # copy in static files
@@ -90,15 +90,18 @@ RUN echo "Enabling kubelet and containerd services ... " \
 RUN echo "Ensuring /etc/kubernetes/manifests" \
     && mkdir -p /etc/kubernetes/manifests
 
-RUN echo "Adjusting systemd-tmpfiles timer" \
-    && sed -i /usr/lib/systemd/system/systemd-tmpfiles-clean.timer -e 's#OnBootSec=.*#OnBootSec=1min#'
-
-
 # shared stage to setup go version for building binaries
-FROM base as go-build
+# NOTE we will be cross-compiling for performance reasons
+FROM --platform=$BUILDPLATFORM $BASE_IMAGE as go-build
+COPY --chmod=0755 files/usr/local/bin/* /usr/local/bin/
 COPY --chmod=0755 scripts/third_party/gimme/gimme /usr/local/bin/
+COPY --chmod=0755 scripts/target-cc /usr/local/bin/
 # tools needed at build-time only
-RUN clean-install git make libseccomp-dev gcc libc-dev pkg-config
+# first ensure we can install packages for both architectures
+RUN dpkg --add-architecture arm64 && dpkg --add-architecture amd64 \
+    && clean-install bash ca-certificates curl git make pkg-config \
+    crossbuild-essential-amd64 crossbuild-essential-arm64 \
+    libseccomp-dev:amd64 libseccomp-dev:arm64
 # set by makefile to .go-version
 ARG GO_VERSION
 RUN eval "$(gimme "${GO_VERSION}")" \
@@ -107,7 +110,7 @@ RUN eval "$(gimme "${GO_VERSION}")" \
 
 # stage for building containerd
 FROM go-build as build-containerd
-ARG GO_VERSION
+ARG TARGETARCH GO_VERSION
 ARG CONTAINERD_VERSION="v1.6.21"
 ARG CONTAINERD_CLONE_URL="https://github.com/containerd/containerd"
 # we don't build with optional snapshotters, we never select any of these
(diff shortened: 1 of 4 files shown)


stmcginnis:
/lgtm

Nice!

BenTheElder:
Found and fixed bug in kindnetd and local-path-provisioner changes.

I've built all of these locally now with `make -C images/$image` which builds both platforms ...

images/base/Dockerfile
```diff
@@ -117,44 +120,48 @@ RUN git clone --filter=tree:0 "${CONTAINERD_CLONE_URL}" /containerd \
...
+ARG TARGETARCH GO_VERSION
 ARG RUNC_VERSION="v1.1.7"
 ARG RUNC_CLONE_URL="https://github.com/opencontainers/runc"
 RUN git clone --filter=tree:0 "${RUNC_CLONE_URL}" /runc \
     && cd /runc \
     && git checkout "${RUNC_VERSION}" \
     && eval "$(gimme "${GO_VERSION}")" \
+    && export GOARCH=$TARGETARCH && export CC=$(target-cc) && export CGO_ENABLED=1 \
     && make runc \
     && go-licenses save --save_path=/_LICENSES .
```
BenTheElder, line 139: not sure if go-licenses is GOARCH aware, we might want to set GOARCH=$TARGETARCH for each of these

BenTheElder: done. it probably is indirectly via whatever it uses to list dependency packages.

images/base/Dockerfile
```diff
@@ -90,15 +90,18 @@ RUN echo "Enabling kubelet and containerd services ... " \
 RUN echo "Ensuring /etc/kubernetes/manifests" \
     && mkdir -p /etc/kubernetes/manifests
 
-RUN echo "Adjusting systemd-tmpfiles timer" \
-    && sed -i /usr/lib/systemd/system/systemd-tmpfiles-clean.timer -e 's#OnBootSec=.*#OnBootSec=1min#'
```
BenTheElder, line 94: the timer value is in a different location and already reasonable in the debian image

images/base/Dockerfile
```diff
@@ -20,7 +20,7 @@
 # start from ubuntu, this image is reasonably small as a starting point
 # for a kubernetes node image, it doesn't contain much we don't need
 # this stage will install basic files and packages
-ARG BASE_IMAGE=ubuntu:22.04
+ARG BASE_IMAGE=debian:bullseye-slim
```
BenTheElder, line 23: kubernetes's own debian-base image has broken held packages, I'm not sure it would actually make any more sense here. debian:.*-slim is already quite compact and seems well-maintained.

BenTheElder: I was pleasantly surprised to find that bullseye-slim even has the same iptables version as the existing image, and the build output looks reasonably sized :-)

If we find we need newer packages, we can switch to debian testing track or bullseye backports. It looks like everything in bullseye should be new enough though and this more closely aligns with k8s core images.

aojea: +1, this also will keep aligned kube-proxy and kindnetd iptables

aojea:
/lgtm


It took me a while to understand it, until I realized this is coming from https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

There is a lot of magic here with the special environment variables, it is good to keep a reference to the docker document at least in the PR comments

images/base/Dockerfile
```diff
@@ -117,64 +120,69 @@ RUN git clone --filter=tree:0 "${CONTAINERD_CLONE_URL}" /containerd \
...
-ARG GO_VERSION
+ARG TARGETARCH GO_VERSION
 ARG CNI_PLUGINS_VERSION="v1.2.0"
 ARG CNI_PLUGINS_CLONE_URL="https://github.com/containernetworking/plugins"
 RUN git clone --filter=tree:0 "${CNI_PLUGINS_CLONE_URL}" /cni-plugins \
     && cd /cni-plugins \
     && git checkout "${CNI_PLUGINS_VERSION}" \
     && eval "$(gimme "${GO_VERSION}")" \
     && mkdir ./bin \
+    && export GOARCH=$TARGETARCH && export CC=$(target-cc) && export CGO_ENABLED=1 \
```
BenTheElder, line 164: I'm not sure if we actually want CGO for these.

We do for containerd, runc. Maybe not for the others.
Thought most of them can already control that in their makefile etc. This is the only set we're just `go build` ing.

BenTheElder:
> There is a lot of magic here with the special environment variables, it is good to keep a reference to the docker document at least in the PR comments

Yes, FWIW in most places I think I left a comment about "buildx sets this" or "makefile sets this" for the ARGs, but that link is an excellent reference for this sort of multi-stage cross compilation and all the built-in envs.

images/base/Dockerfile
BenTheElder: I don't think it will cause problems, anyhow, and it looks like the usual builds do not disable CGO.

CGO is disabled by default when attempting to cross compile with go by setting GOARCH though

https://github.com/containernetworking/plugins/blob/38f18d26ecfef550b8bac02656cc11103fd7cff1/build_linux.sh#L20

BenTheElder:
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-kind-push-base-image/1656553571627307008

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-kind-push-kindnetd-image/1656553571673444352

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-kind-push-local-path-provisioner-image/1656553571719581696

images/base/Dockerfile
aojea: AFAIK this will impact only the libc binding  and mostly dns resolution, since we control the image it should be relatively safe, I honestly can't say what is better, but the cni plugins and crictl are small programs

BenTheElder:
base image built in only 5 minutes instead of hour+ timeout :-)

local-path-provisioner in 4 minutes

kindnetd in 3.5 minutes

images/base/Dockerfile
BenTheElder: crictl we're using the makefile anyhow so if they typically release/test with static that should remain the case.

for this one where we're `go build` ourselves it's worth double checking if that build is comparable to the released binaries. AFAICT it is.

BenTheElder: > since we control the image it should be relatively safe

yeah, that's a motivating factor in #3201 proceeding this, ran into some issues working on an image build and glibc symbol mismatch with external cgo binaries

images/base/Dockerfile
BenTheElder: The distro switch didn't quite work out smoothly, https://github.com/kubernetes-sigs/kind/pull/3218

PR #3216, opened 2023-05-11
Use dl.k8s.io to allow redirects

stmcginnis:
This updates references to allow redirects to other sources than the current GCS bucket.


BenTheElder:
Aplogies, https://github.com/kubernetes-sigs/kind/pull/3217#issuecomment-1544197971

PR #3213, opened 2023-05-10
bump haproxy, kindnetd, local-path, base image

BenTheElder:
bumps all of the images recently updated to their latest builds

PR #3218, opened 2023-05-11
add procps to packages

BenTheElder:
fix following #3214 

I'm going to do a bit more digging to see what else is different / missing, but this is sufficient to get a node image working locally at least.

https://github.com/kubernetes-sigs/kind/pull/3213#issuecomment-1544662636

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -72,7 +72,7 @@ RUN echo "Installing Packages ..." \
       conntrack iptables iproute2 ethtool util-linux mount ebtables kmod \
       libseccomp2 pigz fuse-overlayfs \
       nfs-common open-iscsi \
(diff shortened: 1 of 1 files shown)


BenTheElder:
left: image I just built
right: the default node image currently
https://www.diffchecker.com/aazDihg4/

rsync, socat and dependencies are expected differences because we dropped those before the distro switch.

python3 jumps out as a difference, but kind/kubernetes should not be depending on that anywhere, and if users are depending on something like that in the node image, that wasn't supported ...

mrbobbytables (commented):
/lgtm

BenTheElder:
build here https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-kind-push-base-image/1656768426636480512

PR #3219, opened 2023-05-11
[WIP] testing fixes ...

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/3218 clearly didn't resolve all of the issues with the new base image work, xref https://github.com/kubernetes-sigs/kind/pull/3213

Temp PR just to try some things in CI since I can't repro locally yet.

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -20,7 +20,7 @@
 # start from ubuntu, this image is reasonably small as a starting point
 # for a kubernetes node image, it doesn't contain much we don't need
 # this stage will install basic files and packages
-ARG BASE_IMAGE=debian:bookworm-slim
+ARG BASE_IMAGE=ubuntu:22.04
 FROM $BASE_IMAGE as base
 
 # copy in static files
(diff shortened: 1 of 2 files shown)


BenTheElder:
OK so it's not fixed by newer packages (I was thinking systemd version is slightly older in bullseye than 22.04) and still present when switching back to ubuntu hackily, so it must be in containerd or runc

We have had version changes for those AND we changed the build.

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/3219/pull-kind-conformance-parallel-ga-only/1656809136643379200

BenTheElder:
I'm just appending commits here with changes for the history of commits vs tests, at least for now.
Later I'll file another PR with actual proposed fixes.

I'm pretty sure at this point it's a variation on https://github.com/opencontainers/runc/issues/3849, we have runc ahead of the version in kubelet.

Thankfully the recent CVE fixes are in 1.1.5, so we can roll back as far as 1.1.5 for a bit while we decide what to do about Kubernetes releases < 1.24 (currently Kubernetes is bumping to 1.1.6, hopefully as far back as Kubernetes 1.24).

That also explains why I can't reproduce locally: this is not a problem with cgroupsv2, only v1.

However our main CI and many users are still on cgroups v1 ...

PR #3220, opened 2023-05-12
roll back runc to 1.1.5

BenTheElder:
see: 
https://github.com/kubernetes-sigs/kind/pull/3219#issuecomment-1544933062
https://kubernetes.slack.com/archives/CEKK1KTN2/p1683851267796889
https://github.com/opencontainers/runc/issues/3849

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -128,7 +128,7 @@ RUN git clone --filter=tree:0 "${CONTAINERD_CLONE_URL}" /containerd \
 # stage for building runc
 FROM go-build as build-runc
 ARG TARGETARCH GO_VERSION
-ARG RUNC_VERSION="v1.1.7"
+ARG RUNC_VERSION="v1.1.5"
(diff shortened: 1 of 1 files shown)


images/base/Dockerfile
```diff
@@ -128,7 +128,7 @@ RUN git clone --filter=tree:0 "${CONTAINERD_CLONE_URL}" /containerd \
 # stage for building runc
 FROM go-build as build-runc
 ARG TARGETARCH GO_VERSION
-ARG RUNC_VERSION="v1.1.7"
+ARG RUNC_VERSION="v1.1.5"
```
BenTheElder, line 131: per recent discussion: it's acceptable to have an older runc than the vendored copy in kubelet, but being ahead can cause problems on cgroups v1

so we're going to an older version, this is still new enough to have the recent CVE patches.

thankfully runc version is independent of containerd version in the base image now, so we can avoid rolling back containerd

PR #3228, opened 2023-05-15
use space seperate build tags for containerd build

BenTheElder:
fixes building 1.7+, which has it's own default tag(s) and uses spaces


stmcginnis:
/lgtm

TIL

BenTheElder:
yeah, go doesn't allow mixing these:
> #0 7.401 go: -tags space-separated list contains comma

usually commas have been used in the past, to allow embedding in GOFLAGS

PR #3230, opened 2023-05-17
mark v0.19.0

BenTheElder:
spoke to @aojea, marking a fixup release with the recent base image fixes and preloading fixes in particular


aojea:
+1

PR #3233, opened 2023-05-17
explicitly disable config_path in registry script, so we can eventual…

BenTheElder:
…ly enable it by default and pivot to using it.

I've tested this locally ant it works as expected with kind v0.19.0


BenTheElder:
xref: #2875

neolit123 (commented):
/lgtm

Issue #3223, opened 2023-05-12
upgrading to runc 1.1.6 / 1.1.7 breaks


BenTheElder:
See: #3220, https://kubernetes.slack.com/archives/CEKK1KTN2/p1683851267796889

#3221 and #3222 have test results.

The failure mode is like:
>  { failed [FAILED] failed to run command '/agnhost dns-suffix' on pod, stdout: , stderr: , err: Internal error occurred: error executing command in container: failed to exec in container: failed to start exec "69cc58960cac2f68567eec2a84011360392a649b12c91f955d6ddfeb6e34b80a": OCI runtime exec failed: exec failed: unable to start container process: error adding pid 78936 to cgroups: failed to write 78936: openat2 /sys/fs/cgroup/unified/kubelet.slice/kubelet-kubepods.slice/kubelet-kubepods-besteffort.slice/kubelet-kubepods-besteffort-pod60faed4d_72a9_453e_be90_24e96a46d7e5.slice/cri-containerd-d9b097ed59d27ead25448f023222ac459cc26d37682af62785f486008271b361.scope/cgroup.procs: no such file or directory: unknown: Internal error occurred: error executing command in container: failed to exec in container: failed to start exec "69cc58960cac2f68567eec2a84011360392a649b12c91f955d6ddfeb6e34b80a": OCI runtime exec failed: exec failed: unable to start container process: error adding pid 78936 to cgroups: failed to write 78936: openat2 /sys/fs/cgroup/unified/kubelet.slice/kubelet-kubepods.slice/kubelet-kubepods-besteffort.slice/kubelet-kubepods-besteffort-pod60faed4d_72a9_453e_be90_24e96a46d7e5.slice/cri-containerd-d9b097ed59d27ead25448f023222ac459cc26d37682af62785f486008271b361.scope/cgroup.procs: no such file or directory: unknown

This is limited to cgroups v1, this happens in our CI environment which is further awful by way of host node containerd => dockerd (in CI cluster pod) => kind (running against that nested dockerd)

BenTheElder:
xref: https://github.com/kubernetes/k8s.io/issues/5276 (we'll need to make sure we continue to cover v1 elsewhere for some time)

BenTheElder:
I ran an image built for k8s 1.27.1 with #3221 but on a GKE 1.26 / cgroupv2 node pool and no issues. While checking into https://github.com/kubernetes/k8s.io/issues/5276

[...]

Issue #3169, opened 2023-04-14
Any OOM Kill in the cluster leads to the entire cluster crashing irreparably


howardjohn:
**What happened**:

Whenever an OOM happens in any container in the cluster, the entire cluster crashes and cannot recover.

**What you expected to happen**:

OOM just kills the impacted container, which is restarted by k8s, etc.

**How to reproduce it (as _minimally_ and precisely as possible)**:

```
$  kind create cluster --name oom
$ cat <<EOF | kubectl apply -f -
apiVersion: apps/v1
kind: Deployment
metadata:
  name: echo
spec:
  selector:
    matchLabels:
      app: echo
  template:
    metadata:
      labels:
        app: echo
    spec:
      containers:
      - name: echo
        image: gcr.io/istio-testing/app:latest
        resources:
          limits:
            memory: 1Mi
EOF
```

Wait a minute or so and it will OOM and things will break.

**Anything else we need to know?**:

dmesg:
```
[  +0.000002] oom-kill:constraint=CONSTRAINT_MEMCG,nodemask=(null),cpuset=cri-containerd-548efed163da10fd7613e82876bc24c3a7f355fa740256659bfdb361519bb640.scope,mems_allowed=0,oom_memcg=/system.slice/docker-b2b75b11f95220f9267e53d48df2d654d585405c36b4c752302ab39da59d74e6.scope/kubelet.slice/kubelet-kubepods.slice/kubelet-kubepods-burstable.slice/kubelet-kubepods-burstable-pod68e0cbcc_0cf8_48fc_9a5c_bbb45350dd13.slice,task_memcg=/system.slice/docker-b2b75b11f95220f9267e53d48df2d654d585405c36b4c752302ab39da59d74e6.scope/kubelet.slice/kubelet-kubepods.slice/kubelet-kubepods-burstable.slice/kubelet-kubepods-burstable-pod68e0cbcc_0cf8_48fc_9a5c_bbb45350dd13.slice/cri-containerd-548efed163da10fd7613e82876bc24c3a7f355fa740256659bfdb361519bb640.scope,task=runc:[2:INIT],pid=3451192,uid=0
[  +0.000013] Memory cgroup out of memory: Killed process 3451192 (runc:[2:INIT]) total-vm:1088740kB, anon-rss:20kB, file-rss:7548kB, shmem-rss:0kB, UID:0 pgtables:140kB oom_score_adj:999
```

[...]

PR #3240, opened 2023-05-19
kubelet base environment fixes

BenTheElder:
- set KillMode=process (small patch while thinking about #3169, kOps already does this but not the varyingly broken k8s debs that we can't fix yet because they're still not versioned with k8s releases ...  https://github.com/kubernetes/kubernetes/issues/88832)
- fixup kubelet cgroup root even when using cgroupns (we're responsible for ensuring these when we set --cgroup-root)
- update kubelet flags
  - stop setting --fail-on-swap, we already set it in kubelet config and this flag is deprecated
  - stop setting cgroup-root, we already set it in kubelet config and this flag is deprecated
  - start setting runtime-cgroups, which cannot be set in kubelet config
- pickup containerd upstream service file improvements including OOMScoreAdjust (matches kubelet which self applies this), sd_notify, and waiting for local-fs.target

First patch towards #3223, I've tested most of this but we'll need to let this test broadly across CI after the base image is built

### DIFF
--- a/images/base/files/etc/default/kubelet
+++ b/images/base/files/etc/default/kubelet
@@ -1,1 +1,1 @@
-KUBELET_EXTRA_ARGS=--fail-swap-on=false
+KUBELET_EXTRA_ARGS=--runtime-cgroups=/system.slice/containerd.service
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -181,8 +181,8 @@ fix_mount() {
   mount --make-rshared /
 }
 
-# helper used by fix_cgroup
-mount_kubelet_cgroup_root() {
+# helper used by mount_kubelet_cgroup_root
+mount_kubelet_cgroup_root_subsystem() {
   local cgroup_root=$1
   local subsystem=$2
   if [ -z "${cgroup_root}" ]; then
@@ -199,6 +199,22 @@ mount_kubelet_cgroup_root() {
   # systemd might delete the cgroup unintentionally before the
   # kubelet starts.
   mount --bind "${subsystem}/${cgroup_root}" "${subsystem}/${cgroup_root}"
+}
+
+# helper used by fix_cgroup
+mount_kubelet_cgroup_root() {
+  local cgroup_subsystems=$1
+  echo "${cgroup_subsystems}" |
+  while IFS= read -r subsystem; do
+    mount_kubelet_cgroup_root_subsystem /kubelet "${subsystem}"
+    mount_kubelet_cgroup_root_subsystem /kubelet.slice "${subsystem}"
+  done
+  # workaround for hosts not running systemd
+  # we only do this for kubelet.slice because it's not relevant when not using
+  # the systemd cgroup driver
+  if [[ ! "${cgroup_subsystems}" = */sys/fs/cgroup/systemd* ]]; then
+    mount_kubelet_cgroup_root_subsystem /kubelet.slice /sys/fs/cgroup/systemd
+  fi
 }
 
 fix_cgroup() {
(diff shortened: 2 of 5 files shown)


BenTheElder:
we still have two deprecated flags:
> May 18 23:56:06 kind-control-plane kubelet[230]: Flag --pod-infra-container-image has been deprecated, will be removed in a future release. Image garbage collector will get sandbox image information from CRI.
> May 18 23:56:06 kind-control-plane kubelet[230]: Flag --provider-id has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.

provider-id: https://github.com/kubernetes/kubernetes/pull/116530

pod-infra-container-image is set by kubeadm and I'm not sure what the plan is yet

pacoxu:
> pod-infra-container-image is set by kubeadm and I'm not sure what the plan is yet

https://github.com/kubernetes/kubernetes/issues/106893#issuecomment-1465875895 and https://github.com/kubernetes/kubernetes/pull/114017#discussion_r1045031274. The flag `--pod-infra-container-image` removal should be postponed due to https://github.com/containerd/containerd/pull/7944.

BenTheElder:
Thanks!

I found one more:
> May 19 02:39:27 kind-control-plane kubelet[246]: Flag --container-runtime-endpoint has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.

I believe we're setting this via kubeadm config and kubeadm is setting this via kubelet flag.
We also set kubelet config and could set the option there in kind, but I think kubeadm will still set this flag.

With this PR's changes on 1.27.2:
> May 19 02:39:27 kind-control-plane kubelet[246]: Flag --container-runtime-endpoint has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.
> May 19 02:39:27 kind-control-plane kubelet[246]: Flag --pod-infra-container-image has been deprecated, will be removed in a future release. Image garbage collector will get sandbox image information from CRI.
> May 19 02:39:27 kind-control-plane kubelet[246]: Flag --provider-id has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.

images/base/files/etc/default/kubelet
```diff
@@ -1 +1 @@
-KUBELET_EXTRA_ARGS=--fail-swap-on=false
\ No newline at end of file
+KUBELET_EXTRA_ARGS=--runtime-cgroups=/system.slice/containerd.service
```
BenTheElder, line 1: This is the cgroup in which containerd operates (in kind, and in most cases since there's a standard in-tree service spec).

GKE nodes set this, noticed while debugging the runc issues, which reminded me that we aren't here.
So do the SIG node GCE containerd CI jobs.

https://github.com/kubernetes/test-infra/blob/6928f7825c26cd52dd2db97298db1882d6609584/config/jobs/kubernetes/sig-node/containerd.yaml#L20

https://github.com/kubernetes/kubernetes/blob/15a1f9a39db95a7f83ce5463a328566411d99d51/cluster/gce/util.sh#L823

images/base/files/usr/local/bin/entrypoint
```diff
@@ -201,6 +201,22 @@ mount_kubelet_cgroup_root() {
   mount --bind "${subsystem}/${cgroup_root}" "${subsystem}/${cgroup_root}"
 }
 
+# helper used by fix_cgroup
+mount_kubelet_cgroup_root() {
```
BenTheElder, line 205: This is just pulled out from the end of fix_cgroup for re-use in v1 both with/without cgroupns.

We're responsible for setting up the cgroup when we use --cgroup-root.

We did that previously for the "v1 no cgroupns and nested in another cluster" reasons, but we'll need to keep `--cgroup-root=kubelet` until we can phase it out at some point later following #3241.

justinsb:
Changes lgtm

/lgtm

PR #3241, opened 2023-05-19
enforce private cgroupns

BenTheElder:
Docker's had an option for this since 20.10.0 (https://docs.docker.com/engine/release-notes/20.10/#20100) but last time we looked at this that was too recent.

20.10.0 was released 2020-12-08 (2.5 years ago) and I think we can afford to require it now.

Podman added this in https://github.com/containers/podman/pull/3509 circa July 8th, 2019 which appears to be tagged all the way back to v1.5.0, so it should be available in all working podman versions anyhow.

Tentatively this should resolve issues with #3223, but it first requires #3240.
More generally this makes the node cgroups much easier to deal with.

### DIFF
--- a/pkg/cluster/internal/providers/podman/OWNERS
+++ b/pkg/cluster/internal/providers/podman/OWNERS
@@ -1,13 +1,13 @@
 # See the OWNERS docs at https://go.k8s.io/owners
 
 reviewers:
-  - amwat
   - aojea
   - BenTheElder
 approvers:
-  - amwat
   - aojea
   - BenTheElder
+emeritus_approvers:
+  - amwat
 
 labels:
   - area/provider/podman
--- a/pkg/apis/config/defaults/image.go
+++ b/pkg/apis/config/defaults/image.go
@@ -18,4 +18,4 @@ limitations under the License.
 package defaults
 
 // Image is the default for the Config.Image field, aka the default node image.
-const Image = "kindest/node:v1.27.1@sha256:b7d12ed662b873bd8510879c1846e87c7e676a79fefc93e17b2a52989d3ff42b"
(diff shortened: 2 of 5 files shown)


BenTheElder:
For the last discussion on cgroupns=private, see: https://github.com/kubernetes-sigs/kind/pull/2709#discussion_r84470932

It's been a year now and I'm more comfortable requiring users to have upgraded docker.

We can eventually drop some of the worst hacks here once we release `kind` with this change and phase out support for kind binaries without it.

pkg/cluster/internal/providers/podman/OWNERS
```diff
@@ -1,13 +1,13 @@
 # See the OWNERS docs at https://go.k8s.io/owners
 
 reviewers:
-  - amwat
   - aojea
   - BenTheElder
 approvers:
-  - amwat
```
BenTheElder, line 8: We already did this elsewhere but missed this file, which caused the bot to assign him to this PR

stmcginnis:
> it should be available in all working podman versions anyhow.

Agreed!

/lgtm

BenTheElder:
>  [It] [sig-network] HostPort validates that there is no conflict between pods with same hostPort but different hostIP and protocol [LinuxOnly] [Conformance]

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/3241/pull-kind-conformance-parallel-ipv6/1659678265620566016

/retest

BenTheElder:
> [It] [sig-node] Probing container should mark readiness on pods to false and disable liveness probes while pod is in progress of terminating

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/3241/pull-kind-e2e-kubernetes-1-25/1659678271631003648
/retest

PR #3243, opened 2023-05-19
fix lingering go get in quick start

### LINKED ISSUE #3242: [Docs] hyperLink mismatch
In quickstart->installation there is a hyperlink for the `go get option` which will not point to the go reference but the same quickstart page only and just the page will reload as it is a link mismatch

BenTheElder:
fixes #3242


stmcginnis:
/lgtm

Looks good!

PR #3245, opened 2023-05-22
switch registry script to use containerd hosts dir

BenTheElder:
This partially unblocks #2875 and moving forward with newer containerd releases ahead of containerd 2.0 (instead of sticking on 1.6 LTS longer).

We'll still need to highlight this in release notes in the next release to warn people to update their own local scripts to this approach.

It's also a little bit clunky, but it's backwards compatible with existing kind releases / doesn't depend on any new functionality, which should help us through the transition.

### DIFF
--- a/site/static/examples/kind-with-registry.sh
+++ b/site/static/examples/kind-with-registry.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 set -o errexit
 
-# create registry container unless it already exists
+# 1. Create registry container unless it already exists
 reg_name='kind-registry'
 reg_port='5001'
 if [ "$(docker inspect -f '{{.State.Running}}' "${reg_name}" 2>/dev/null || true)" != 'true' ]; then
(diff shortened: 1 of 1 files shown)


site/static/examples/kind-with-registry.sh
```diff
@@ -10,30 +10,46 @@ if [ "$(docker inspect -f '{{.State.Running}}' "${reg_name}" 2>/dev/null || true
...
+# This is necessary because localhost resolves to loopback addresses that are
+# network-namespace local.
+# In other words: localhost in the container is not localhost on the host.
+#
+# We want a consistent name that works from both ends, so we tell containerd to
+# alias localhost:${reg_port} to the registry container when pulling images
+REGISTRY_DIR="/etc/containerd/certs.d/localhost:${reg_port}"
+for node in $(kind get nodes); do
+  docker exec "${node}" mkdir -p "${REGISTRY_DIR}"
+  cat <<EOF | docker exec -i "${node}" cp /dev/stdin "${REGISTRY_DIR}/hosts.toml"
```
BenTheElder, line 41: there might be some clever way to do this with a single command, but this is how we write files internally to kind currently.

site/static/examples/kind-with-registry.sh
```diff
@@ -10,30 +10,46 @@ if [ "$(docker inspect -f '{{.State.Running}}' "${reg_name}" 2>/dev/null || true
...
+[host."http://${reg_name}:5000"]
```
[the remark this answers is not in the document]
BenTheElder: This config basically says "this endpoint http://kind-registry:5000 is the one mirror for localhost:${reg_port}" (so use http not https, and use kind-registry:5000 not localhost:5001).

It should be equivalent in behavior to what we did before.

site/static/examples/kind-with-registry.sh
aojea: `docker cp` ?

BenTheElder: Will cp create missing directory?

aojea: nevermind, I misunderstood the comment, we need to replace the var and create a folder

site/static/examples/kind-with-registry.sh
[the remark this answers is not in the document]
aojea: using localhost as an alias is 😬

site/static/examples/kind-with-registry.sh
BenTheElder: Yeah, we'd still have to mkdir and I don't think you can stain pipe a regular file in to docker cp, it takes a tar

site/static/examples/kind-with-registry.sh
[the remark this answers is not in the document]
BenTheElder: It seems less bad than possible alternatives. It's worked so far.

PR #3247, opened 2023-05-23
entrypoint: clean up logging and indentation

BenTheElder:


neolit123 (commented):
/lgtm

images/base/files/usr/local/bin/entrypoint
```diff
@@ -18,13 +18,24 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+# logging helpers
+log_info() {
+  echo "INFO: $1" >&2
```
aojea, line 23: we were not logging info to stderr, is this change on purpose?

BenTheElder: We mostly were (which is correct for diagnostic messages), but not always

PR #3253, opened 2023-05-23
make FindSource work even in CI without path_alias

BenTheElder:
… and without shelling out to go

go build context may shell out to go binary since go modules, but kind doesn't require a go binary on the host otherwise.

### DIFF
--- a/pkg/build/nodeimage/internal/kube/source.go
+++ b/pkg/build/nodeimage/internal/kube/source.go
@@ -19,6 +19,8 @@ package kube
 import (
 	"fmt"
 	"go/build"
+	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/alessio/shellescape"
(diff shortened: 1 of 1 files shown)


pkg/build/nodeimage/internal/kube/source.go
```diff
@@ -27,25 +29,43 @@ import (
...
+	if gopath == "" {
+		return "", errors.New("could not find Kubernetes source under current working directory and GOPATH is not set")
 	}
-	return "", fmt.Errorf("could not find %s module source under GOPATH=%s: %w", ImportPath, build.Default.GOPATH, err)
+	// try k8s.io/kubernetes first (old canonical GOPATH locaation)
+	if dir := filepath.Join(gopath, "src", "k8s.io", "kubernetes"); probablyKubeDir(dir) {
+		return dir, nil
+	}
+	// then try github.com/kubernetes/kubernetes (CI without path_alias set)
+	if dir := filepath.Join(gopath, "src", "github.com", "kubernetes", "kubernetes"); probablyKubeDir(dir) {
```
BenTheElder, line 52: Observed that some CI jobs are missing `path_alias` and would previously fail to locate sources.

We fixed those by passing the source path to `kind build node-image`, but ideally it should just work.

Prow will default to a GOPATH based on the repo URL if `path_alias` is not specified.

pkg/build/nodeimage/internal/kube/source.go
```diff
@@ -27,25 +29,43 @@ import (
...
-// maybeKubeDir returns true if the dir looks plausibly like a kubernetes
+// probablyKubeDir returns true if the dir looks plausibly like a kubernetes
 // source directory
-func maybeKubeDir(dir string) bool {
-	// TODO(bentheelder): consider adding other sanity checks
-	return dir != ""
+func probablyKubeDir(dir string) bool {
+	// TODO: should we do more checks?
+	// NOTE: go.mod with this line has existed since Kubernetes 1.15
+	const sentinelLine = "module k8s.io/kubernetes"
```
BenTheElder, line 63: over 4 years now https://github.com/kubernetes/kubernetes/blame/551868d863bea157af4d1ef9debbd0e16df17728/go.mod#L6

go is _really_ unlikely to change this part of the spec and we have no reason to rename the module.

I guess we could also check that this is the entire line but i'm not sure that's necessary and may break on line ending differences, and the user can always override by passing in the desired source path.

PR #3255, opened 2023-05-25
add workaround for misc controller on cgroups v1

BenTheElder:
While doing this I needed to work more with the system bins and systemd config, so I refactored kind vs non kind bins and systemd files and added a warning file to `/kind`.

The core of this is:
1. un-mount the misc controller
2. mask the misc controller from `/proc/cgroups` with a censored bind-mount that removes this controller before starting systemd, then undo the bind mount before we start containerd/kubelet

2) is a little gross but necessary because otherwise systemd will discover the controller is enabled and mount it.

We only do this on cgroupsv1.

We just do it unconditionally instead of depending on the Kubernetes version because we don't have any use for misc currently. Users that need misc should switch to cgroupsv2 (and there's never been a kind release where misc worked on v1 anyhow since runc only just gained support under v1), we can revisit when we choose to apply this later.

This should allow us to upgrade to runc 1.1.7 even when:
1. The host runtime is not handling misc
2. kubelet is not aware of misc (<1.23 or 1.24+ but not on the latest patches)


neolit123 (commented):
i must admit that i'm not well qualified to review this, but i see some ecosystem concerns around the v1 -> v2 migration for cgroups. also reading the abstraction idea of misc render me puzzled with design flaws in mind - "they could have done this 'better' / more verbose maybe".

it seems that not only kind but other tools on top of the migration process must handle cases like this. the entrypoint fake mount is highlights that systemd / runc are not helping either.

in any case, LGTM. but hopefully we get to v2 eventually linux + k8s wise. :))

perhaps @aojea has further comments.

BenTheElder:
> i must admit that i'm not well qualified to review this, but i see some ecosystem concerns around the v1 -> v2 migration for cgroups. also reading the abstraction idea of misc render me puzzled with design flaws in mind - "they could have done this 'better' / more verbose maybe".

Setting aside what I think about the controller, AFAICT currently it only exists for new resources that so far don't make sense for kind clusters like AMD ASIDs (https://lwn.net/Articles/856438/), and can be accessed under cgroupsv2 if truly needed.

runc only added support for this controller under v1 in 1.1.6, which is what caused this, so any existing kind users have not been using misc for anything unless they were on v2 already (and I _highly_ doubt it even then).

cgroups v2 + cgroupns actually removes most of the headaches from doing KIND-like nonsense :-)

> it seems that not only kind but other tools on top of the migration process must handle cases like this. 

Yes, I've been talking to minikube folks about this.

> the entrypoint fake mount highlights that systemd / runc are not helping either.

Well, it's a challenging problem for runc and kind is a pretty excessive edge-case that isn't necessarily supported.

systemd does nominally support running in a container but not in the way we're doing it (privileged ...) and the behavior there is pretty reasonable, it is responsible for ensuring cgroups are setup properly and inside a cgroupns this is ~fine, but then we turn around and do this in a screwy nested environment in k8s CI 🙃 

> in any case, LGTM. but hopefully we get to v2 eventually linux + k8s wise. :)) (and drop v1)

It's definitely coming, I think systemd set a target for EOY '23, and most of the distros have already switched.

cgroups v2 is already the default on docker desktop, debian stable, ubuntu, fedora ....

Similarly we can drop support for nodes without cgroupns sometime soon which will remove some of the worst hackery we're doing essentially trying to mimic cgroupns-like layout. Once #3241 has been released for a bit.

neolit123:
> It's definitely coming, I think systemd set a target for EOY '23, and most of the distros have already switched.

sounds good. hope they are no delays around that👍

BenTheElder:
We should have Prow CI results shortly, I've been testing in a GCE VM with misc + v1 configured.

We will not get CI results for podman / v2 / rootless / ... on github actions yet since I haven't updated the node image.

Given the false start on upgrading to runc 1.1.7 before, waiting to have confirmed Prow CI results first.

BenTheElder:
All e2e passed with no flakes, imagine that! :-)

images/base/files/kind/bin/undo-mount-hacks.sh
```diff
@@ -0,0 +1,28 @@
...
+set -o pipefail
+
+# don't run on cgroups v2
+if [[ -f "/sys/fs/cgroup/cgroup.controllers" ]]; then
+    exit 0
+fi
+
+# on cgroups v1 we want to undo bind mounting over this to hide misc
+# see entrypoint
+umount /proc/cgroups
```
aojea, line 28: don't this run after the entrypoint?

images/base/files/usr/local/bin/entrypoint
```diff
@@ -228,6 +228,27 @@ mount_kubelet_cgroup_root() {
...
+  local misc_controller='/sys/fs/cgroup/misc'
+  if [[ ! -d $misc_controller ]]; then
+    return 0
+  fi
+  log_info "removing misc controller"
+  umount $misc_controller || log_warn "failed umount $misc_controller"
+  rmdir $misc_controller || log_warn "failed rmdir $misc_controller"
+  # systemd will discover misc controller is available here and re-mount it
+  # we will pretend it isn't available with a bind mount
+  # this mount will be removed by undo-mount-hacks.service
```
aojea, line 247: I see https://github.com/kubernetes-sigs/kind/pull/3255/commits/89f3890a31353adc1ea55cad9e1372d2b0cdd5df#r1206047711

images/base/files/kind/bin/undo-mount-hacks.sh
BenTheElder: yes, the idea is we don't want to undo this until _after_ systemd read the file and setup cgroups. we don't terribly need to rush here and may not need to do it at all, but conceivably something like cadvisor could desire to read the real details and it's just confusing to have a fake here because the contents are dynamic, they include the number of active groups for the controller.

images/base/files/usr/local/bin/entrypoint
aojea: haha, so you mount it first to avoid systemd to mount it, and later you unmount it

images/base/files/kind/bin/undo-mount-hacks.sh
BenTheElder: https://man7.org/linux/man-pages/man7/cgroups.7.html

```
   /proc files
       /proc/cgroups (since Linux 2.6.24)
              This file contains information about the controllers that
              are compiled into the kernel.  An example of the contents
              of this file (reformatted for readability) is the
              following:

                  #subsys_name    hierarchy      num_cgroups    enabled
                  cpuset          4              1              1
                  cpu             8              1              1
                  cpuacct         8              1              1
                  blkio           6              1              1
                  memory          3              1              1
                  devices         10             84             1
                  freezer         7              1              1
                  net_cls         9              1              1
                  perf_event      5              1              1
                  net_prio        9              1              1
                  hugetlb         0              1              0
                  pids            2              1              1

              The fields in this file are, from left to right:

              1. The name of the controller.

              2. The unique ID of the cgroup hierarchy on which this
                 controller is mounted.  If multiple cgroups v1
                 controllers are bound to the same hierarchy, then each
                 will show the same hierarchy ID in this field.  The
                 value in this field will be 0 if:

                   a) the controller is not mounted on a cgroups v1
                      hierarchy;

                   b) the controller is bound to the cgroups v2 single
                      unified hierarchy; or

                   c) the controller is disabled (see below).

              3. The number of control groups in this hierarchy using
                 this controller.

              4. This field contains the value 1 if this controller is
                 enabled, or 0 if it has been disabled (via the
                 cgroup_disable kernel command-line boot parameter).
```

We're deleting the `misc` line and making a fixed copy of this what systemd sees when probing for available controllers.

aojea: systemd never retries, right?

images/base/files/usr/local/bin/entrypoint
BenTheElder: yes. we mount a modified copy of the proc entry so systemd won't discover misc, then we unmount the fake. 🙃 

we also have to unmount misc because the host runtime may have already set it up.

images/base/files/kind/bin/undo-mount-hacks.sh
BenTheElder: correct. from systemd POV everything succeeded and cgroup controller mounting is a one-time init thing.

BenTheElder: this is all because, if we controlled the host kernel etc. we'd either just have misc aware all the way down or disable the controller in the kernel. it doesn't make sense from systemd POV to have a flag to disable controllers because:

1) in a non containerized environment this could already be done in the kernel.
2) in a containerized environment systemd is not supposed to be in a privileged container and cgroups should be fully setup already.

and changing this isn't really worth it, because even if you could try to make this sound reasonable, it's all deprecated cgroups v1 anyhow, systemd and the rest of the ecosystem will drop that in the next year or so and v2 doesn't need any of this.

aojea:
/lgtm

the `fix_cgroup` function is really 😬

images/base/files/kind/bin/undo-mount-hacks.sh
```diff
@@ -0,0 +1,28 @@
...
+set -o nounset
+set -o pipefail
+
+# don't run on cgroups v2
+if [[ -f "/sys/fs/cgroup/cgroup.controllers" ]]; then
+    exit 0
+fi
+
+# on cgroups v1 we want to undo bind mounting over this to hide misc
+# see entrypoint
```
BenTheElder, line 27: this should probably `|| true` or otherwise check if we actually have a bind mount to remove, instead of only checking if we're on cgroups v1.

BenTheElder:
/hold
one little fix to do

BenTheElder:
addressed https://github.com/kubernetes-sigs/kind/pull/3255#discussion_r1206057319 and dropped the image bumps, we'll do the real ones in a follow-up once the base image auto-push job completes.

images/base/files/kind/bin/undo-mount-hacks.sh
BenTheElder: done. went with `|| true`, because doing this is harmless if it's not a mount and not worth the complexity to detect.

BenTheElder:
[new e2e failure is a flake and does NOT reflect these changes, since the image bumps were dropped]

BenTheElder:
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-kind-push-base-image/1661874202950504448 will build the image

Issue #3164, opened 2023-04-11
Support GPUs


tmc:
**What would you like to be added**: GPU support.

**Why is this needed**: Development of AI/ML models in kind.

https://docs.docker.com/config/containers/resource_constraints/

aojea:
it looks hard to generalize https://docs.docker.com/config/containers/resource_constraints/#access-an-nvidia-gpu , it seems you need some prerequisites and also run some nvidia code

tmc:
> it looks hard to generalize https://docs.docker.com/config/containers/resource_constraints/#access-an-nvidia-gpu , it seems you need some prerequisites and also run some nvidia code

Yes, put setting out a path for even making it possible (with some elbow grease) would be appreciated.

aojea:
how this is supposed to be used, a kind cluster with gpus ... asking from the ignorance

BenTheElder:
I think we're still lacking a portable standard for this and a use case that can't be solved with extraMount / the existing `--privileged`

samos123:
Could you consider allowing this to be added for now?

There is a lot of value to end-users that don't want to fork Kind just to get GPU support. I was going through [this blog post](https://jacobtomlinson.dev/posts/2022/running-kubeflow-inside-kind-with-gpu-support/) and feel like lacking a portable standard isn't a strong enough reason to have to fork. A fork is even less portable.

aojea:
https://github.com/jacobtomlinson/kind/pull/1/files

it does not seem as hard to maintain

BenTheElder:
Is passing through all GPUs expected to remain desirable?

we do need some limited abstraction that we can retarget to podman and maybe nerdctl in the future.

a book that simply passes all GPUs is one of those

As per our contributing guide some specific proposal must be discussed and we should be looking at podman as well, that PR doesn't appear to.

BenTheElder:
If we have to investigate podman etc ourselves this will remain on the backlog below pressing bug fixes like #3223.

If anyone is interested in helping move this along then please outline the approach to portability. Not just between runtimes but of the configurations across hosts.

The --gpu all mode in docker looks reasonable enough. It's better than hardcoding devices into config. But it seems somewhat unlikely to satisfy all users, and it doesn't look like podman has an equivilant at a glance so at the very least we need to address that with an error in the podman driver.

samos123:
[...]

PR #3257, opened 2023-05-30
Add GPU support

samos123:
WIP, still need to add podman support

This implements GPU support as discussed in #3164

### DIFF
--- a/pkg/apis/config/v1alpha4/types.go
+++ b/pkg/apis/config/v1alpha4/types.go
@@ -120,7 +120,7 @@ type Node struct {
 
 	// GPUs allows to access GPU devices from the kind node. Setting this to
 	// "all" will pass all the available GPUs to the kind node.
-	Gpus string `yaml:"gpus,omitempty" json:"gpus,omitempty"`
+	GPUs string `yaml:"gpus,omitempty" json:"gpus,omitempty"`
 
 	// KubeadmConfigPatches are applied to the generated kubeadm config as
 	// merge patches. The `kind` field must match the target object, and
--- a/pkg/apis/config/v1alpha4/types.go
+++ b/pkg/apis/config/v1alpha4/types.go
@@ -118,6 +118,10 @@ type Node struct {
 	// binded to a host Port
 	ExtraPortMappings []PortMapping `yaml:"extraPortMappings,omitempty" json:"extraPortMappings,omitempty"`
 
+	// GPUs allows to access GPU devices from the kind node. Setting this to
+	// "all" will pass all the available GPUs to the kind node.
+	Gpus string `yaml:"gpus,omitempty" json:"gpus,omitempty"`
+
 	// KubeadmConfigPatches are applied to the generated kubeadm config as
 	// merge patches. The `kind` field must match the target object, and
 	// if `apiVersion` is specified it will only be applied to matching objects.
--- a/pkg/cluster/internal/providers/docker/provision.go
+++ b/pkg/cluster/internal/providers/docker/provision.go
@@ -255,8 +255,8 @@ func runArgsForNode(node *config.Node, clusterIPFamily config.ClusterIPFamily, n
 		args = append(args, "-e", "KUBECONFIG=/etc/kubernetes/admin.conf")
 	}
 
-	if len(node.Gpus) > 0 {
-		args = append(args, fmt.Sprintf("--gpus=%v", node.Gpus))
+	if len(node.GPUs) > 0 {
+		args = append(args, fmt.Sprintf("--gpus=%v", node.GPUs))
 	}
 
 	// finally, specify the image to run
--- a/pkg/cluster/internal/providers/docker/provision.go
+++ b/pkg/cluster/internal/providers/docker/provision.go
@@ -255,6 +255,10 @@ func runArgsForNode(node *config.Node, clusterIPFamily config.ClusterIPFamily, n
 		args = append(args, "-e", "KUBECONFIG=/etc/kubernetes/admin.conf")
 	}
 
+	if len(node.Gpus) > 0 {
+		args = append(args, fmt.Sprintf("--gpus=%v", node.Gpus))
+	}
+
 	// finally, specify the image to run
 	return append(args, node.Image), nil
 }
--- a/pkg/internal/apis/config/types.go
+++ b/pkg/internal/apis/config/types.go
@@ -98,6 +98,10 @@ type Node struct {
 	// binded to a host Port
 	ExtraPortMappings []PortMapping
 
+	// GPUs allows to access GPU devices from the kind node. Setting this to
+	// "all" will pass all the available GPUs to the kind node.
+	Gpus string
+
 	// KubeadmConfigPatches are applied to the generated kubeadm config as
 	// strategic merge patches to `kustomize build` internally
 	// https://github.com/kubernetes/community/blob/a9cf5c8f3380bb52ebe57b1e2dbdec136d8dd484/contributors/devel/sig-api-machinery/strategic-merge-patch.md
--- a/pkg/internal/apis/config/types.go
+++ b/pkg/internal/apis/config/types.go
@@ -100,7 +100,7 @@ type Node struct {
 
 	// GPUs allows to access GPU devices from the kind node. Setting this to
 	// "all" will pass all the available GPUs to the kind node.
-	Gpus string
+	GPUs string
 
 	// KubeadmConfigPatches are applied to the generated kubeadm config as
 	// strategic merge patches to `kustomize build` internally
--- a/site/content/docs/user/configuration.md
+++ b/site/content/docs/user/configuration.md
@@ -287,7 +287,7 @@ nodes:
 
 ### GPU Support
 
-Kind nodes can utilize GPUs by setting the following:
+Kind nodes can utilize GPU devices from host by setting the following:
 {{< codeFromInline lang="yaml" >}}
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
--- a/site/content/docs/user/configuration.md
+++ b/site/content/docs/user/configuration.md
@@ -285,6 +285,20 @@ nodes:
 
 **Note**: Kubernetes versions are expressed as x.y.z, where x is the major version, y is the minor version, and z is the patch version, following [Semantic Versioning](https://semver.org/) terminology. For more information, see [Kubernetes Release Versioning.](https://github.com/kubernetes/sig-release/blob/master/release-engineering/versioning.md#kubernetes-release-versioning)
 
+### GPU Support
+
+Kind nodes can utilize GPUs by setting the following:
+{{< codeFromInline lang="yaml" >}}
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+nodes:
+- role: control-plane
+  gpus: all
+{{< /codeFromInline >}}
+
+As a pre-requisite you need to have installed the
+[NVIDIA Container Toolkit](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/install-guide.html).
+
 ### Extra Mounts
 
 Extra mounts can be used to pass through storage on the host to a kind node
--- a/pkg/cluster/internal/providers/podman/provision.go
+++ b/pkg/cluster/internal/providers/podman/provision.go
@@ -212,6 +212,10 @@ func runArgsForNode(node *config.Node, clusterIPFamily config.ClusterIPFamily, n
 		args...,
 	)
 
+	if len(node.GPUs) > 0 {
+		args = append(args, "--device", fmt.Sprintf("nvidia.com/gpu=%v", node.GPUs))
+	}
+
 	// convert mounts and port mappings to container run args
 	args = append(args, generateMountBindings(node.ExtraMounts...)...)
 	mappingArgs, err := generatePortMappings(clusterIPFamily, node.ExtraPortMappings...)
@@ -302,7 +306,6 @@ type podmanNetworks []struct {
 func getSubnets(networkName string) ([]string, error) {
 	cmd := exec.Command("podman", "network", "inspect", networkName)
 	out, err := exec.Output(cmd)
-
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get subnets")
 	}
--- a/pkg/internal/apis/config/convert_v1alpha4.go
+++ b/pkg/internal/apis/config/convert_v1alpha4.go
@@ -56,6 +56,7 @@ func convertv1alpha4Node(in *v1alpha4.Node, out *Node) {
 	out.ExtraMounts = make([]Mount, len(in.ExtraMounts))
 	out.ExtraPortMappings = make([]PortMapping, len(in.ExtraPortMappings))
 	out.KubeadmConfigPatchesJSON6902 = make([]PatchJSON6902, len(in.KubeadmConfigPatchesJSON6902))
+	out.Gpus = in.Gpus
 
 	for i := range in.ExtraMounts {
 		convertv1alpha4Mount(&in.ExtraMounts[i], &out.ExtraMounts[i])
--- a/pkg/internal/apis/config/convert_v1alpha4.go
+++ b/pkg/internal/apis/config/convert_v1alpha4.go
@@ -56,7 +56,7 @@ func convertv1alpha4Node(in *v1alpha4.Node, out *Node) {
 	out.ExtraMounts = make([]Mount, len(in.ExtraMounts))
 	out.ExtraPortMappings = make([]PortMapping, len(in.ExtraPortMappings))
 	out.KubeadmConfigPatchesJSON6902 = make([]PatchJSON6902, len(in.KubeadmConfigPatchesJSON6902))
-	out.Gpus = in.Gpus
+	out.GPUs = in.GPUs
 
 	for i := range in.ExtraMounts {
 		convertv1alpha4Mount(&in.ExtraMounts[i], &out.ExtraMounts[i])
--- a/pkg/internal/apis/config/validate.go
+++ b/pkg/internal/apis/config/validate.go
@@ -103,8 +103,7 @@ func (n *Node) Validate() error {
 
 	// validate node role should be one of the expected values
 	switch n.Role {
-	case ControlPlaneRole,
-		WorkerRole:
+	case ControlPlaneRole, WorkerRole:
 	default:
 		errs = append(errs, errors.Errorf("%q is not a valid node role", n.Role))
 	}
@@ -112,6 +111,10 @@ func (n *Node) Validate() error {
 	// image should be defined
 	if n.Image == "" {
 		errs = append(errs, errors.New("image is a required field"))
+	}
+
+	if err := validateGPUs(n.GPUs); err != nil {
+		errs = append(errs, errors.Wrapf(err, "invalid gpus"))
 	}
 
 	// validate extra port forwards
@@ -188,6 +191,13 @@ func validatePortMappings(portMappings []PortMapping) error {
 
 		// add the entry to bindMap
 		bindMap[portProtocol].Insert(addrString)
+	}
+	return nil
+}
+
+func validateGPUs(gpus string) error {
+	if len(gpus) > 0 && gpus != "all" {
+		return errors.Errorf("invalid gpus string: '%v'. Only 'all' is a valid value", gpus)
 	}
 	return nil
 }
--- a/pkg/internal/apis/config/validate_test.go
+++ b/pkg/internal/apis/config/validate_test.go
@@ -18,8 +18,9 @@ package config
 
 import (
 	"fmt"
-	"sigs.k8s.io/kind/pkg/internal/assert"
 	"testing"
+
+	"sigs.k8s.io/kind/pkg/internal/assert"
 
 	"sigs.k8s.io/kind/pkg/errors"
 )
@@ -251,7 +252,7 @@ func TestClusterValidate(t *testing.T) {
 	}
 
 	for _, tc := range cases {
-		tc := tc //capture loop variable
+		tc := tc // capture loop variable
 		t.Run(tc.Name, func(t *testing.T) {
 			t.Parallel()
 			err := tc.Cluster.Validate()
@@ -344,6 +345,15 @@ func TestNodeValidate(t *testing.T) {
 			ExpectErrors: 1,
 		},
 		{
+			TestName: "Invalid GPUs",
+			Node: func() Node {
+				cfg := newDefaultedNode(ControlPlaneRole)
+				cfg.GPUs = "not-all"
+				return cfg
+			}(),
+			ExpectErrors: 1,
+		},
+		{
 			TestName: "Invalid HostPort",
 			Node: func() Node {
 				cfg := newDefaultedNode(ControlPlaneRole)
@@ -360,7 +370,7 @@ func TestNodeValidate(t *testing.T) {
 	}
 
 	for _, tc := range cases {
-		tc := tc //capture loop variable
+		tc := tc // capture loop variable
 		t.Run(tc.TestName, func(t *testing.T) {
 			t.Parallel()
 			err := tc.Node.Validate()
@@ -414,7 +424,7 @@ func TestPortValidate(t *testing.T) {
 	}
 
 	for _, tc := range cases {
-		tc := tc //capture loop variable
+		tc := tc // capture loop variable
 		t.Run(tc.TestName, func(t *testing.T) {
 			t.Parallel()
 			err := validatePort(tc.Port)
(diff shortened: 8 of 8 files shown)


pkg/apis/config/v1alpha4/types.go
```diff
@@ -118,6 +118,10 @@ type Node struct {
 	// binded to a host Port
 	ExtraPortMappings []PortMapping `yaml:"extraPortMappings,omitempty" json:"extraPortMappings,omitempty"`
 
+	// GPUs allows to access GPU devices from the kind node. Setting this to
+	// "all" will pass all the available GPUs to the kind node.
+	Gpus string `yaml:"gpus,omitempty" json:"gpus,omitempty"`
```
neolit123, line 123: use GPUs?
>  All letters in the acronym should have the same case, using the appropriate case for the situation

https://github.com/zecke/Kubernetes/blob/master/docs/devel/api-conventions.md#naming-conventions

pkg/internal/apis/config/types.go
```diff
@@ -98,6 +98,10 @@ type Node struct {
 	// binded to a host Port
 	ExtraPortMappings []PortMapping
 
+	// GPUs allows to access GPU devices from the kind node. Setting this to
+	// "all" will pass all the available GPUs to the kind node.
+	Gpus string
```
neolit123, line 103: same

site/content/docs/user/configuration.md
```diff
@@ -285,6 +285,20 @@ nodes:
 
 **Note**: Kubernetes versions are expressed as x.y.z, where x is the major version, y is the minor version, and z is the patch version, following [Semantic Versioning](https://semver.org/) terminology. For more information, see [Kubernetes Release Versioning.](https://github.com/kubernetes/sig-release/blob/master/release-engineering/versioning.md#kubernetes-release-versioning)
 
+### GPU Support
+
+Kind nodes can utilize GPUs by setting the following:
```
neolit123, line 290:
```suggestion
Kind nodes can utilize GPU devices from the host, by setting the following:
```

aligning with the API godoc comment

pkg/cluster/internal/providers/docker/provision.go
```diff
@@ -255,6 +255,10 @@ func runArgsForNode(node *config.Node, clusterIPFamily config.ClusterIPFamily, n
 		args = append(args, "-e", "KUBECONFIG=/etc/kubernetes/admin.conf")
 	}
 
+	if len(node.Gpus) > 0 {
+		args = append(args, fmt.Sprintf("--gpus=%v", node.Gpus))
```
BenTheElder, line 259: let's not plumb this directly since the values are incompatible across backends

instead we can have a more structured format in the internal type (for now gpus true/false) and we need valdiatino on the external type.

pkg/apis/config/v1alpha4/types.go
BenTheElder: +1, also this field should be validated (with "all" being the only valid value currently), and we should put a note that in the future we'll look at supporting specifying specific devices.

BenTheElder: Commented more below and on the issue.

pkg/cluster/internal/providers/docker/provision.go
samos123: I added validation to ensure right now only "all" can be passed so it's now safe to to use node.Gpus directly in the docker provisioner. I prefer to keep the code simpler and passthrough instead of adding another layer of validation in the docker provisioning code. Happy to add additional validation in docker/povision.go if you have strong opinion on it.

For podman we can not plumb it through directly. So I will work out a way to convert node.Gpus into a podman compatible format. I don't have podman installed myself.

samos123:
One interesting thing I had to do was to run the following after kidn cluster came up:
```
helm install --wait --generate-name \
     -n gpu-operator --create-namespace \
     nvidia/gpu-operator \
     --set driver.enabled=false

docker exec -ti kind-control-plane bash
systemctl restart containerd
ln -s /sbin/ldconfig /sbin/ldconfig.real
```

So I'm thinking there might needs to be a change to the base image to include that symlink

site/content/docs/user/configuration.md
samos123: done

pkg/internal/apis/config/types.go
samos123: done

pkg/apis/config/v1alpha4/types.go
samos123: done

BenTheElder:
Thanks @klueska. As discussed in the linked issue I'd much rather adopt a standardized and portable approach.

https://kind.sigs.k8s.io/docs/design/principles/#target-cri-functionality

BenTheElder:
This should probably look closer to extraMounts.

samos123:
Why not adopt `--device` once docker 25 comes out and then do a version check? I rather have this in than encouraging people to fork kind or use other tools.  The end-users of kind are unaware whether `--device` or `--gpu` is used. Let's focus on end-user value vs implementation detail changing when people move to docker 25 in years. Everyone agrees to get this in before docker supports `--device` with gpus?

samos123:
That's exactly why I'm investigating the future so the Kind API stays the same even when eventually Docker 25 changes only the Docker implementation detail is changed to use `--device` instead of `--gpu`.  I don't want to waste time working on this if there is no desire to get this in before docker 25.

The Kind API should focus on what docker 25 and podman looks like, which I think is a fair and useful requirement.

So are we good as long as the Kind API stays the same that this will get implemented before Docker 25?

I do think it's important to have this PR account for what docker will be in docker 25 but I am opposed to only supporting GPUs in docker 25 and beyond.

BenTheElder:
I think we can make the API surface look like "devices" and some logic to detect a single "all nvidia" device value that shims to the GPU flag, but that will be more work for the implementer (AKA you currently) than requiring a docker upgrade.

I care more about the user facing API and how it evolves and has a maintainable backend going forward. 

If it currently has some workaround code that will be deleted later that's OK.
We'll have to consider what it looks like if the user tries to configure other devices as well.

klueska:
Does kind allow passing through envvars to the docker CLI? I ask because setting NVIDIA_VISIBLE_DEVICES=all is the same as passing —gpus=all of nvidia is set as the default runtime in docker‘s daemon.json config.

BenTheElder:
> Does kind allow passing through envvars to the docker CLI? I ask because setting NVIDIA_VISIBLE_DEVICES=all is the same as passing —gpus=all of nvidia is set as the default runtime in docker‘s daemon.json config.

We have some other experimental functionality based on passing through env but I forget if we explicitly pass them or pass in all env from the parent process.

BenTheElder:
I'm pretty sure that will work and could be a good stop gap until we flesh out devices support

klueska:
By parent process, do you mean whatever process runs „kind create cluster“. I was hoping there was a config option for envs to be set independently for each node.

BenTheElder:
> By parent process, do you mean whatever process runs „kind create cluster“.

Yes.

> I was hoping there was a config option for envs to be set independently for each node.

This hasn't currently made sense. Environment variables have been well for environment, like HTTP_PROXY, and overriding the snapshotter (due to the host filesystem). I don't think we'd want to newly introduce this for a stopgap, but it would be fine for single node test clusters in the meantime.

samos123:
I couldn't get it to work by simply passing the environment variable from the host. It looks like nvidia experts the env variable to be set in the container like this:
```
docker run --rm --runtime=nvidia \
    -e NVIDIA_VISIBLE_DEVICES=all nvidia/cuda nvidia-smi
```
which brings us back to needing to do a code change.

In that case, I propose for now to use `--gpus=all` and whenever docker 25 comes out then change the implementation to use --device. The API of Kind should be focused on podman and future --device like API of Docker 25. I will refactor the code to use a list of strings for gpus instead of it being a single string.

klueska:
@BenTheElder just to double back on this.

You mentioned previously that:
> We have some other experimental functionality based on passing through env but I forget if we explicitly pass them or pass in all env from the parent process.

Based on the most recent comment from @samos123 it seems as if only selected envs are passed through.

Would it be possible to whitelist the NVIDIA_VISIBLE_DEVICES envvar as one of the envs that get forwarded? Ideally we could set it per-node (so as to assign different nodes to different GPUs from the underlying machine), but in the absence of this, having the ability to set it globally for all nodes would be better than nothing.

klueska:
It occurred to me over the weekend that we can actually enable GPU support in `kind` __today__ by leveraging a feature we added to the nvidia-container-toolkit quite some time ago:

https://docs.google.com/document/d/1zy0key-EL6JH50MZgwg96RPYxxXXnVUdxLZwGiyqLd8/edit

It relies on passing the list of GPUs you want to inject as volume mounts rather than via an environment variable.

Steps to enable this:
1. Add `nvidia` as your default runtime in /etc/docker/daemon.json
2. Set `accept-nvidia-visible-devices-as-volume-mounts = true` in `/etc/nvidia-container-runtime/config.toml`
3. Restart docker (as necessary)
4. Add the following to any kind nodes you want to have access to all GPUs in the system:
```
  extraMounts:
    - hostPath: /dev/null
      containerPath: /var/run/nvidia-container-devices/all
```

I've tested this in my local environment and it works as expected. The only caveat is that it only works for passing `all` GPUs into a node (i.e. I can't pick and choose some GPUs for one node and another set of GPUs for another node).

In general the `nvidia-visible-devices-as-volume-mount` feature allows you to do fine-grained injection of a subset of GPUs into a container. However, `kind` runs all of its nodes in `--privileged` containers, meaning that there is no way to prevent it from seeing *all* GPUs (even if you tell it you only want it to see a subset of the GPUs). If / when `kind` ever supports running worker nodes without `--privileged` this restriction should go away.

grokspeed:
> Steps to enable this:
> 
> 1. Add `nvidia` as your default runtime in `/etc/docker/daemon.json`
[...]

Thanks @klueska I have linux server at home.  But I have been running K8s training for my team using kind on their W11 laptops.  Would like to dive into Kubeflow/Notebooks-with-GPUs scenario (I know, an overkill if not for training purpose, could have just used WSL2+CUDA).  Would you be able translate the above to Docker Desktop on Windows environment?

The daemon config is accessible via the Docker Desktop UI.  But I need some help on the nvidia-container-runtime/config.toml location on Windows.  Though, as I read the Nvidia page on container runtime and toolkit I came away thinking they may not be applicable to WSL2, as:

    docker run -it --rm --gpus all ubuntu nvidia-smi

seems to work with just installing CUDA support.   But Kubeflow seems different in that Nvidia runtime is still needed for the new containers I would be building... if I understand the container runtime concept correctly?

samos123:
I confirmed the steps provided by @klueska worked except hitting 1 minor issue. The only thing I had to do was run this:
```
docker exec -ti kind-control-plane ln -s /sbin/ldconfig /sbin/ldconfig.real
```
Should that be something that should be included in the base image? 

Here were the full steps I used to verify:
```
nvidia_config_file="/etc/nvidia-container-runtime/config.toml"
if [ -e "${nvidia_config_file}" ]; then
sudo sed -i '/accept-nvidia-visible-devices-as-volume-mounts/c\accept-nvidia-visible-devices-as-volume-mounts = true' ${nvidia_config_file}
fi

kind create cluster --name kind --config - <<EOF
kind: Cluster
nodes:
- role: control-plane
  image: kindest/node:v1.27.3@sha256:3966ac761ae0136263ffdb6cfd4db23ef8a83cba8a463690e98317add2c9ba72
  extraPortMappings:
  - containerPort: 30080
    hostPort: 30080
  # part of GPU workaround
  extraMounts:
    - hostPath: /dev/null
      containerPath: /var/run/nvidia-container-devices/all
EOF

docker exec -ti kind-control-plane ln -s /sbin/ldconfig /sbin/ldconfig.real || true

helm repo add nvidia https://helm.ngc.nvidia.com/nvidia || true
helm repo update
helm install --wait --generate-name \
     -n gpu-operator --create-namespace \
     nvidia/gpu-operator --set driver.enabled=false
```

klueska:
@samos123 You shouldn't need to create that symlink. Can you show me the contents of your `/etc/nvidia-container-runtime/config.toml` file?

samos123:
@klueska content of my file: https://gist.github.com/samos123/cc816b91a7a03651c71441e0949c3bb6

Note I don't seem to be the only one hitting this issue: https://github.com/NVIDIA/nvidia-docker/issues/614#issuecomment-423991632 That's the source of the workaround

xussof:
Following @samos123 steps I can't fully deploy nvidia-device-plugin-daemonset and nvidia-operator-validator when I have more than one node deployed by using this kind-config:

kind: Cluster
name: siv-dev-gpu
nodes:
  - role: control-plane
    image: kindest/node:v1.26.0
    extraMounts:
      - hostPath: /data/string-in-video/repos/git/github/string-in-video
        containerPath: /data/string-in-video/repos/git/github/string-in-video
      - hostPath: /data/string-in-video/files/dev
        containerPath: /data/string-in-video/files/dev
      - hostPath: /data/string-in-video/volumes/dev
        containerPath: /data/string-in-video/volumes/dev
      - hostPath: /dev/null
        containerPath: /var/run/nvidia-container-devices/all
    kubeadmConfigPatches:
      - |
        kind: InitConfiguration
        nodeRegistration:
          kubeletExtraArgs:
    extraPortMappings:
      - containerPort: 80
        hostPort: 80
        protocol: TCP
      - containerPort: 443
        hostPort: 443
        protocol: TCP

  - role: worker
    image: kindest/node:v1.26.0
    extraMounts:
      - hostPath: /data/string-in-video/repos/git/github/string-in-video
        containerPath: /data/string-in-video/repos/git/github/string-in-video
      - hostPath: /data/string-in-video/files/dev
        containerPath: /data/string-in-video/files/dev
      - hostPath: /data/string-in-video/volumes/dev
        containerPath: /data/string-in-video/volumes/dev

  - role: worker
    image: kindest/node:v1.26.0
    extraMounts:
      - hostPath: /data/string-in-video/repos/git/github/string-in-video
        containerPath: /data/string-in-video/repos/git/github/string-in-video
      - hostPath: /data/string-in-video/files/dev
        containerPath: /data/string-in-video/files/dev
      - hostPath: /data/string-in-video/volumes/dev
        containerPath: /data/string-in-video/volumes/dev
      - hostPath: /dev/null
        containerPath: /var/run/nvidia-container-devices/all

This is the error I get:

image

Failed to create pod sandbox: rpc error: code = Unknown desc = failed to get sandbox runtime: no runtime for "nvidia" is configured

But when I deploy it over a kind with a control-plane and a single worker it works.

samos123:
I think thats a limitation of the approach. Maybe @klueska as another cool workaround to solve that? :D

klueska:
Only one of your workers has:
```
      - hostPath: /dev/null
        containerPath: /var/run/nvidia-container-devices/all
```

xussof:
> Only one of your workers has:
> 
> ```
>       - hostPath: /dev/null
>         containerPath: /var/run/nvidia-container-devices/all
> ```

Yes because I only one one node to has gpu, the others I want to emulate that doesn't have GPU. Still, it fails if I add the hostpath to the another worker:

`kind: Cluster
name: siv-dev-gpu
nodes:
  - role: control-plane
    image: kindest/node:v1.26.0
    extraMounts:
      - hostPath: /data/string-in-video/repos/git/github/string-in-video
        containerPath: /data/string-in-video/repos/git/github/string-in-video
      - hostPath: /data/string-in-video/files/dev
        containerPath: /data/string-in-video/files/dev
      - hostPath: /data/string-in-video/volumes/dev
        containerPath: /data/string-in-video/volumes/dev
      - hostPath: /dev/null
        containerPath: /var/run/nvidia-container-devices/all
    kubeadmConfigPatches:
      - |
        kind: InitConfiguration
        nodeRegistration:
          kubeletExtraArgs:
    extraPortMappings:
      - containerPort: 80
        hostPort: 80
        protocol: TCP
      - containerPort: 443
        hostPort: 443
        protocol: TCP

  - role: worker
    image: kindest/node:v1.26.0
    extraMounts:
      - hostPath: /data/string-in-video/repos/git/github/string-in-video
        containerPath: /data/string-in-video/repos/git/github/string-in-video
      - hostPath: /data/string-in-video/files/dev
        containerPath: /data/string-in-video/files/dev
      - hostPath: /data/string-in-video/volumes/dev
        containerPath: /data/string-in-video/volumes/dev
      - hostPath: /dev/null
        containerPath: /var/run/nvidia-container-devices/all
        
  - role: worker
    image: kindest/node:v1.26.0
    extraMounts:
      - hostPath: /data/string-in-video/repos/git/github/string-in-video
        containerPath: /data/string-in-video/repos/git/github/string-in-video
      - hostPath: /data/string-in-video/files/dev
        containerPath: /data/string-in-video/files/dev
      - hostPath: /data/string-in-video/volumes/dev
        containerPath: /data/string-in-video/volumes/dev
      - hostPath: /dev/null
        containerPath: /var/run/nvidia-container-devices/all`

image


Error of nvidia-device-plugin-daemonset:



`NVIDIA_DRIVER_ROOT=/
CONTAINER_DRIVER_ROOT=/host
Starting nvidia-device-plugin
I0912 18:33:00.988910       1 main.go:154] Starting FS watcher.
E0912 18:33:00.988987       1 main.go:123] failed to create FS watcher: too many open files`


Error on nvidia-operator-validator:
`Failed to create pod sandbox: rpc error: code = Unknown desc = failed to get sandbox runtime: no runtime for "nvidia" is configured`

cceyda:
I had to downgrade the nvidia driver from 545->535 for it to work. Based on the compatibility stuff of the gpu-operator:
https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/release-notes.html
then followed the tutorial https://github.com/kubernetes-sigs/kind/pull/3257#issuecomment-1685220605 (Thank you)

jiangxiaobin96:
> It occurred to me over the weekend that we can actually enable GPU support in `kind` **today** by leveraging a feature we added to the `nvidia-container-toolkit` quite some time ago:
> 
> [Read list of GPU devices from volume mounts instead of NVIDIA_VISIBLE_DEVICES](https://docs.google.com/document/d/1uXVF-NWZQXgP1MLb87_kMkQvidpnkNWicdpO2l9g-fw/edit#)
> 
> It relies on passing the list of GPUs you want to inject as volume mounts rather than via an environment variable.
> 
> Steps to enable this:
> 
> 1. Add `nvidia` as your default runtime in `/etc/docker/daemon.json`
> 2. Restart `docker` (as necessary)
> 3. Set `accept-nvidia-visible-devices-as-volume-mounts = true` in `/etc/nvidia-container-runtime/config.toml`
> 4. Add the following to any kind nodes you want to have access to all GPUs in the system:
> 
> ```
>   extraMounts:
>     - hostPath: /dev/null
>       containerPath: /var/run/nvidia-container-devices/all
> ```
> 
> I've tested this in my local environment and it works as expected. The only caveat is that it only works for passing `all` GPUs into a node (i.e. I can't pick and choose some GPUs for one node and another set of GPUs for another node).
> 
> In general the `nvidia-visible-devices-as-volume-mount` feature allows you to do fine-grained injection of a subset of GPUs into a container. However, `kind` runs all of its nodes in `--privileged` containers, meaning that there is no way to prevent it from seeing _all_ GPUs (even if you tell it you only want it to see a subset of the GPUs). If / when `kind` ever supports running worker nodes without `--privileged` this restriction should go away.

Hello, I follow this comment and meet new error
```
Error: failed to create containerd task: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error running hook #0: error running hook: exit status 1, stdout: , stderr: nvidia-container-cli.real: mount error: stat failed: /proc/driver/nvidia/capabilities: no such file or directory: unknown
```
How to solve this problem?

xihajun:
```
Creating cluster "new-cluster" ...
 ✓ Ensuring node image (kindest/node:v1.27.3) 🖼
 ✓ Preparing nodes 📦 📦 📦
 ✓ Writing configuration 📜
 ✓ Starting control-plane 🕹️
 ✓ Installing CNI 🔌
 ✓ Installing StorageClass 💾
 ✗ Joining worker nodes 🚜
Deleted nodes: ["new-cluster-control-plane" "new-cluster-worker" "new-cluster-worker2"]
ERROR: failed to create cluster: failed to join node with kubeadm: command "docker exec --privileged new-cluster-worker kubeadm join --config /kind/kubeadm.conf --skip-phases=preflight --v=6" failed with error: exit status 1
Command Output: I1220 00:26:42.732281     158 join.go:412] [preflight] found NodeName empty; using OS hostname as NodeName
I1220 00:26:42.732373     158 joinconfiguration.go:76] loading configuration from "/kind/kubeadm.conf"
I1220 00:26:42.734232     158 controlplaneprepare.go:225] [download-certs] Skipping certs download
I1220 00:26:42.734258     158 join.go:529] [preflight] Discovering cluster-info
I1220 00:26:42.734283     158 token.go:80] [discovery] Created cluster-info discovery client, requesting info from "new-cluster-control-plane:6443"
I1220 00:26:42.754681     158 round_trippers.go:553] GET https://new-cluster-control-plane:6443/api/v1/namespaces/kube-public/configmaps/cluster-info?timeout=10s 200 OK in 19 milliseconds
```

jiangxiaobin96:
> Anyone can help with this issue?
> 
> ```
> Creating cluster "new-cluster" ...
>  ✓ Ensuring node image (kindest/node:v1.27.3) 🖼
>  ✓ Preparing nodes 📦 📦 📦
>  ✓ Writing configuration 📜
>  ✓ Starting control-plane 🕹️
>  ✓ Installing CNI 🔌
>  ✓ Installing StorageClass 💾
>  ✗ Joining worker nodes 🚜
> Deleted nodes: ["new-cluster-control-plane" "new-cluster-worker" "new-cluster-worker2"]
> ERROR: failed to create cluster: failed to join node with kubeadm: command "docker exec --privileged new-cluster-worker kubeadm join --config /kind/kubeadm.conf --skip-phases=preflight --v=6" failed with error: exit status 1
> Command Output: I1220 00:26:42.732281     158 join.go:412] [preflight] found NodeName empty; using OS hostname as NodeName
> I1220 00:26:42.732373     158 joinconfiguration.go:76] loading configuration from "/kind/kubeadm.conf"
> I1220 00:26:42.734232     158 controlplaneprepare.go:225] [download-certs] Skipping certs download
> I1220 00:26:42.734258     158 join.go:529] [preflight] Discovering cluster-info
> I1220 00:26:42.734283     158 token.go:80] [discovery] Created cluster-info discovery client, requesting info from "new-cluster-control-plane:6443"
> I1220 00:26:42.754681     158 round_trippers.go:553] GET https://new-cluster-control-plane:6443/api/v1/namespaces/kube-public/configmaps/cluster-info?timeout=10s 200 OK in 19 milliseconds
> ```
> 
> > Solved it, out of RAM
> 
> But now got
> 
> ```
> │ Failed to allocate device vector A (error code CUDA driver version is insufficient for CUDA runtime version)!                  │
> ```

cuda version and driver version do not match.

alexeadem:
> > Steps to enable this:
> > 
> > 1. Add `nvidia` as your default runtime in `/etc/docker/daemon.json`
> >    [...]
> 
> Thanks @klueska I have linux server at home. But I have been running K8s training for my team using kind on their W11 laptops. Would like to dive into Kubeflow/Notebooks-with-GPUs scenario (I know, an overkill if not for training purpose, could have just used WSL2+CUDA). Would you be able translate the above to Docker Desktop on Windows environment?
> 
> The daemon config is accessible via the Docker Desktop UI. But I need some help on the nvidia-container-runtime/config.toml location on Windows. Though, as I read the Nvidia page on container runtime and toolkit I came away thinking they may not be applicable to WSL2, as:
> 
> ```
> docker run -it --rm --gpus all ubuntu nvidia-smi
> ```
> 
> seems to work with just installing CUDA support. But Kubeflow seems different in that Nvidia runtime is still needed for the new containers I would be building... if I understand the container runtime concept correctly?
You may  want to look at this [comment](https://github.com/NVIDIA/k8s-device-plugin/issues/332#issuecomment-1917537232) running [qbo](https://ce.qbo.io/#/) with kind mages or directly into kind

utsumi-fj:
FYI, in addition to https://github.com/kubernetes-sigs/kind/pull/3257#issuecomment-1607287275, the workaround https://github.com/NVIDIA/nvidia-container-toolkit/issues/198#issuecomment-1884784524 is needed.

I installed GPU Operator in kind as follows:

1. https://github.com/kubernetes-sigs/kind/pull/3257#issuecomment-1607287275
2. https://github.com/NVIDIA/nvidia-container-toolkit/issues/198#issuecomment-1884784524 (`docker exec -it {{ kind node }} umount -R /proc/driver/nvidia`)
3. `helm install --wait --generate-name -n gpu-operator --create-namespace nvidia/gpu-operator --version=v24.9.1 --set driver.enabled=false`

Maybe, https://github.com/NVIDIA/gpu-operator/issues/441#issuecomment-1539649895 is needed too.

PR #3260, opened 2023-06-04
Typo Error

ShubhamKumar89:
Improve comment.


BenTheElder:
This is not an improvement, "AMD64" is another common name for x86_64:

https://en.wikipedia.org/wiki/X86-64

Importantly, it is the name the Go programming language uses.

PR #3263, opened 2023-06-05
add command to get podman subnet

maxerenberg:
This PR updates the LoadBalancer page to demonstrate how to obtain the IP address range used by the kind subnet when using Podman.
It is currently only known to work when using the netavark backend in rootful mode.


aojea:
/approve

Thanks

PR #3264, opened 2023-06-05
Remove `-sf 7` from haproxy image entrypoint

killianmuldoon:
Remove `-sf 7` from the haproxy entrypoint. This arg caused an issue in the Cluster API e2e tests causing the load balancer to exit prematurely. More info on the CAPI issue: https://github.com/kubernetes-sigs/cluster-api/issues/8641

### DIFF
--- a/images/haproxy/Dockerfile
+++ b/images/haproxy/Dockerfile
@@ -62,4 +62,4 @@ COPY haproxy.cfg /usr/local/etc/haproxy/haproxy.cfg
 
 # below roughly matches the standard haproxy image
 STOPSIGNAL SIGUSR1
-ENTRYPOINT ["haproxy", "-sf", "7", "-W", "-db", "-f", "/usr/local/etc/haproxy/haproxy.cfg"]
+ENTRYPOINT ["haproxy", "-W", "-db", "-f", "/usr/local/etc/haproxy/haproxy.cfg"]


images/haproxy/Dockerfile
```diff
@@ -62,4 +62,4 @@ COPY haproxy.cfg /usr/local/etc/haproxy/haproxy.cfg
 
 # below roughly matches the standard haproxy image
 STOPSIGNAL SIGUSR1
-ENTRYPOINT ["haproxy", "-sf", "7", "-W", "-db", "-f", "/usr/local/etc/haproxy/haproxy.cfg"]
+ENTRYPOINT ["haproxy", "-W", "-db", "-f", "/usr/local/etc/haproxy/haproxy.cfg"]
```
killianmuldoon, line 65: Alternative we could just make the change to remove the 7 and leave the `-sf` arg. The `-W` arg is enough to enable workers to spin up when the config is updated, however, and this matches the fix in CAPI.

killianmuldoon: Ref: https://github.com/kubernetes-sigs/cluster-api/pull/8685

neolit123: a couple of questions.

i don't remember how HA works in kind with haproxy.

if the haproxy config is prepared in advance post parsing the user node config and not iterated on each time a new CP node is added, wouldn't that mean that the -sf and -W flags are not needed - i.e. no dynamic reconfig?

else, if kind plans to support dynamic topology ALA 'kind add node', then it makes sense to keep these?

also, reading the docs, i do not understand the diff between -sf with and without a pid. what is it?

aojea: we don't really use -sf https://docs.haproxy.org/2.2/management.html , and having it harcoded to a number sounds even more suspicious that this was an oversight.

Kind just create the config in the container and send a HUP signal

https://github.com/kubernetes-sigs/kind/blob/5dfe2806e9affb7e24f1da26d0415b9f60c9f547/pkg/cluster/internal/create/actions/loadbalancer/loadbalancer.go#L94

but is not dynamically reconfigured, so this lgtm

aojea:
/approve

Thanks for following up and explain all the tests you have done and the conclusions

images/haproxy/Dockerfile
aojea: interestingly `-db` does not look like something recommended in production @BenTheElder 😄 


  -db : disable background mode and multi-process mode. The process remains in
    foreground. It is mainly used during development or during small tests, as
    Ctrl-C is enough to stop the process. Never use it in an init script.

PR #3184, opened 2023-04-24
podman: create ipv4 network if ipv6 is disabled

aojea:
Fixes: https://github.com/kubernetes-sigs/kind/issues/2892

Issue #3016, opened 2022-12-02
Kindnet fails during create cluster on RHEL8 when ipv6.disable=1


mmoore0011:
**What happened**:
Create cluster fails with:

```
ERROR: failed to create cluster: command "podman run --name kind-control-plane --hostname kind-control-plane --label io.x-k8s.kind.role=control-plane --privileged --tmpfs /tmp --tmpfs /run --volume df992b4a85f944c329890d2eccff715d0d79d7c16217b1f2e7cd81365f823f76:/var:suid,exec,dev --volume /lib/modules:/lib/modules:ro -e KIND_EXPERIMENTAL_CONTAINERD_SNAPSHOTTER --detach --tty --net kind --label io.x-k8s.kind.cluster=kind -e container=podman --volume /dev/mapper:/dev/mapper --publish=127.0.0.1:36055:6443/tcp -e KUBECONFIG=/etc/kubernetes/admin.conf docker.io/kindest/node:v1.22.15" failed with error: exit status 127

Command Output: time="2022-12-01T18:23:45-05:00" level=warning msg="Failed to load cached network config: network kind not found in CNI cache, falling back to loading network kind from disk"

Error: plugin type="bridge" failed (add): cni plugin bridge failed: open /proc/sys/net/ipv6/conf/eth0/accept_dad: no such file or directory
```

The server is booted with `ipv6.disable=1` so `/proc/sys/net/ipv6` doesn't exist

**What you expected to happen**:
Cluster starts

**How to reproduce it (as minimally and precisely as possible)**:
Not sure if this is specific to RHEL8 with podman but, boot a RHEL8 server with `ipv6.disable=1` and run `kind create cluster`

**Anything else we need to know?**:

**Environment:**

- kind version: (use `kind version`):
`kind version 0.17.0`

- Kubernetes version: (use `kubectl version`):
```
Client Version: v1.25.4
Kustomize Version: v4.5.7
```

- Docker version: (use `docker info`):
```
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
Client:       Podman Engine
Version:      4.0.2
API Version:  4.0.2
Go Version:   go1.17.7

Built:      Tue Apr 19 06:16:32 2022
OS/Arch:    linux/amd64
```

- OS (e.g. from `/etc/os-release`):
```
VERSION="8.6 (Ootpa)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="8.6"
[...]

PR #3270, opened 2023-06-14
Fix bug in podman provider with linux OS -

zvigrinberg:
When need to create network with ipv6, then in case ipv6 is disabled in linux kernel,  podman might not return error that ipv6 is not enabled in kernel ( for example - it happened on my computer, with RHEL 8.8), instead it's creating the network and failing later when creating the cluster in a container, so need in this case to check in case of running in Linux OS, and in case of ipv6 module is disabled in linux kernel , to create the network without ipv6 and return, otherwise, continue as before.  this solution makes the following manual workaround redundant:
[Comment in Issue 3016]( https://github.com/kubernetes-sigs/kind/issues/3016#issuecomment-1335556882) 
It's also extending the fix of the following PR:
podman: create ipv4 network if ipv6 is disablle 
[https://github.com/kubernetes-sigs/kind/pull/3184](https://github.com/kubernetes-sigs/kind/pull/3184)

### DIFF
--- a/pkg/cluster/internal/providers/podman/network.go
+++ b/pkg/cluster/internal/providers/podman/network.go
@@ -19,6 +19,7 @@ package podman
 import (
 	"crypto/sha1"
 	"encoding/binary"
+	"encoding/json"
 	"net"
 	"regexp"
 	"strings"
@@ -46,6 +47,13 @@ func ensureNetwork(name string) error {
 	// network already exists
 	if checkIfNetworkExists(name) {
 		return nil
+	}
+
+	// in case kind is running in linux and IPv6 is disabled in linux kernel, then create the podman network without Ipv6
+	// This is required here because i came across that even though ipv6 is disabled in kernel, podman doesn't stop you from creating a network with Ipv6 enabled,
+	// which later will result in crashing when it'll try to create the cluster.
+	if isIPv6DisabledInLinuxKernel() {
+		return createNetwork(name, "")
 	}
 
 	// generate unique subnet per network based on the name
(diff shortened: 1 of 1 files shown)


pkg/cluster/internal/providers/podman/network.go
```diff
@@ -48,6 +49,13 @@ func ensureNetwork(name string) error {
 		return nil
 	}
 
+	// in case kind is running in linux and IPv6 is disabled in linux kernel, then create the podman network without Ipv6
+	// This is required here because i came across that even though ipv6 is disabled in kernel, podman doesn't stop you from creating a network with Ipv6 enabled,
+	// which later will result in crashing when it'll try to create the cluster.
+	if isIPv6DisabledInLinuxKernel() {
+		return createNetwork(name, "")
```
aojea, line 56: we rely on the errors returned, see Lines 69-73
```
	if isUnknownIPv6FlagError(err) ||
		isIPv6DisabledError(err) {
```

if podman creates a network without ipv6 when we ask it to create it with ipv6 it looks a podman bug

zvigrinberg: @aojea 
podman is  creating the network with ipv6, the problem is that he does that even if the linux kernel ipv6 is disabled.

I Have ipv6 disabled in linux kernel,
```shell
[zgrinber@zgrinber kind]$ cat /sys/module/ipv6/parameters/disable
1

```

But trying creating a cluster with kind and podman failed, as you can see in the following printscreen:

image

and with my fix it succeeded:

image

image

zvigrinberg: @aojea
The Real problem in the above scenario, that it's not even  reaching code in lines 69-73 that you mentioned, as it's creating successfully the network:
```go
	subnet := generateULASubnetFromName(name, 0)
	err := createNetwork(name, subnet)
// this one causes to skip the treatment at Lines 69-73
	if err == nil {
		// Success!
		return nil
	}
```

zvigrinberg: @aojea I Believe that there are scenarios/use cases where podman returns errors and others where it doesn't return error...., and the second ones are captured by my treatment, whereas the first ones are captured by the treatment atlines 69-73

aojea: > as it's creating successfully the network:

why podman creates the network if ipv6 is disabled? that sounds like a bug

zvigrinberg: I Really Don't know, and it's not consistent , in some cases it returns error and in others it creates the network despite ipv6 is disabled....
My best smart guess is that sysctl net.ipv6.conf.all.disable_ipv6 property doesn't exists on my machine/system:
```shell
[zgrinber@zgrinber kind]$ sysctl net.ipv6.conf.all.disable_ipv6
sysctl: cannot stat /proc/sys/net/ipv6/conf/all/disable_ipv6: No such file or directory
```
instead of returning 0 or 1.
Maybe this cause podman to create the network possibly it reads that value

stmcginnis: I think the point @aojea is making is: this is a bug in podman and it should be fixed in podman. Kind should not have special handling to try to work around this issue.

PR #3281, opened 2023-06-22
site: simplify ingress example

### LINKED ISSUE #3181: Contour ingress isn't working
I went thru the Ingress documentation to set up Contour. I got to the bottom and ran the two `curl` commands and am not met with any echo "foo-app" or "bar-app" as it says should happen. Instead, I get:

```
> curl localhost:8100/bar/hostname
curl: (52) Empty reply from server
```

Note I have to use `:8100` because my kind cluster is being setup w/ a different template than the default shown in the walkthrough

<details>
<summary>my cluster template</summary>

```yaml
kind: Cluster
nodes:
- role: control-plane
  kubeadmConfigPatches:
  - |
    kind: InitConfiguration
    nodeRegistration:
      kubeletExtraArgs:
  extraPortMappings:
  - containerPort: 80
    hostPort: 8100
    protocol: TCP
  - containerPort: 443
    hostPort: 8443
    protocol: TCP
```

</details>

And so I modified the "Using ingress" template to replace all the 8080 with 80, then attempt to hit 8100.
If i try to just hit 80 I get
```
> curl localhost/bar/hostname
curl: (7) Failed to connect to localhost port 80 after 2242 ms: Couldn't connect to server
```

as I (believe) is correct given my cluster setup.

I am wondering if the images being used no longer echo back and, instead, just ACK as a way to signify e2e is working? If so, it seems this doc needs an update.

skriss:
Simplifies the ingress example to use a single
rule with basic prefix matching and no impl-
specific annotations.

Closes #3181.

### DIFF
--- a/site/content/docs/user/ingress.md
+++ b/site/content/docs/user/ingress.md
@@ -149,8 +149,8 @@ Refer [Using Ingress](#using-ingress) for a basic example usage.
 
 ## Using Ingress
 
-The following example creates simple http-echo services
-and an Ingress object to route to these services.
+The following example creates a simple agnhost service
+and an Ingress object to route to that service.
 
 ```yaml
 {{% readFile "static/examples/ingress/usage.yaml" %}}
@@ -165,8 +165,6 @@ kubectl apply -f {{< absURL "examples/ingress/usage.yaml" >}}
 Now verify that the ingress works
 
 {{< codeFromInline lang="bash" >}}
-# should output "foo-app"
-curl localhost/foo/hostname
-# should output "bar-app"
-curl localhost/bar/hostname
+# should output "agnhost"
+curl http://localhost/hostname
 {{< /codeFromInline >}}
--- a/site/static/examples/ingress/usage.yaml
+++ b/site/static/examples/ingress/usage.yaml
@@ -1,9 +1,9 @@
 kind: Pod
 apiVersion: v1
 metadata:
-  name: foo-app
+  name: agnhost
   labels:
-    app: foo
+    app: agnhost
 spec:
   containers:
   - command:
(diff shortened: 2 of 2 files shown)


aojea:
/hold
the key of the example is to demonstrate that you can route by url

aojea:
> /hold the key of the example is to demonstrate that you can route by url

/hold cancel

oh, I didn't read the issue https://github.com/kubernetes-sigs/kind/issues/3181#issuecomment-1602848669 

let's discuss it them, sorry for the noise

skriss:
another option could be to use the ingress conformance echoserver image for the backend, which wouldn't require any path rewriting and can include the pod name/other info in the JSON response; that would work fine with multiple rules using different path prefixes. Happy to implement whatever you all prefer as long as it's implementation-agnostic.

aojea:
I do not have a strong opinion, but this is just not giving much example on what an ingress does, ideally we should be able to send something to /foo and check that is different backend than sending to /bar

skriss:
OK, if you prefer to keep multiple rules in place then a couple other options that only make use of the core Ingress spec:
- use a different host for each rule instead of path matching; the example `curl` commands will need to set the `Host` header explicitly but can continue to use the `agnhost` echoserver
- switch to the ingress conformance echo server, which would be compatible with standard path matching; the JSON output returned from this echoserver is more verbose, though.

Let me know what you prefer, I'll happily update to either or any other suggestion you have.

site/content/docs/user/ingress.md
```diff
@@ -165,8 +165,6 @@ kubectl apply -f {{< absURL "examples/ingress/usage.yaml" >}}
 Now verify that the ingress works
 
 {{< codeFromInline lang="bash" >}}
-# should output "foo-app"
```
kundan2707, line 168: Existing one is correct.

skriss:
Since this PR hasn't gotten any traction I'll attempt an update to just the Contour section, though FWIW I still think it'd be better not to use nginx-specific annotations in the common example.

PR #3286, opened 2023-06-28
skip setting up default pod subnet when disabling default CNI

armstrongli:
default pod subnet is optional in kubernetes. the default subnet brought into kubelet/kube-proxy can bring extra NAT rules in kind node for service and bring unexpected troubles for custom cni.

here is the default NAT rule in kind node:

```
KUBE-MARK-MASQ  all  -- !10.244.0.0/16        anywhere             /* Kubernetes service cluster ip + port for masquerade purpose */ match-set KUBE-CLUSTER-IP dst,dst
```

### DIFF
--- a/pkg/apis/config/v1alpha4/default.go
+++ b/pkg/apis/config/v1alpha4/default.go
@@ -47,8 +47,8 @@ func SetDefaultsCluster(obj *Cluster) {
 			obj.Networking.APIServerAddress = "::1"
 		}
 	}
-	// default the pod CIDR
-	if obj.Networking.PodSubnet == "" {
+	// default the pod CIDR if not disabling default CNI
+	if obj.Networking.PodSubnet == "" && !obj.Networking.DisableDefaultCNI {
 		obj.Networking.PodSubnet = "10.244.0.0/16"
 		if obj.Networking.IPFamily == IPv6Family {
 			// node-mask cidr default is /64 so we need a larger subnet, we use /56 following best practices
--- a/pkg/internal/apis/config/default.go
+++ b/pkg/internal/apis/config/default.go
@@ -60,8 +60,8 @@ func SetDefaultsCluster(obj *Cluster) {
 		}
 	}
 
-	// default the pod CIDR
-	if obj.Networking.PodSubnet == "" {
+	// default the pod CIDR if not disabling default CNI
+	if obj.Networking.PodSubnet == "" && !obj.Networking.DisableDefaultCNI {
 		obj.Networking.PodSubnet = "10.244.0.0/16"
 		if obj.Networking.IPFamily == IPv6Family {
 			// node-mask cidr default is /64 so we need a larger subnet, we use /56 following best practices
--- a/pkg/internal/apis/config/validate.go
+++ b/pkg/internal/apis/config/validate.go
@@ -53,8 +53,10 @@ func (c *Cluster) Validate() error {
 	}
 
 	// podSubnet should be a valid CIDR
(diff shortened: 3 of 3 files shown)


aojea:
> default pod subnet is optional in kubernetes.

what do you mean by optiona, this is the subnet with all the pods 😄 

Thanks for your contributions, but this is not required, I suggest you to look for the `help-wanted` issues or open a feature request first to discuss the new feature
/close

armstrongli:
> what do you mean by optiona, this is the subnet with all the pods

if we use the default kindnetd and use the default subnet, there are no issues when we access to our service. but when we use external CNI to set up pod IPs and use other subnets, there are extra NAT for the traffic and it brings troubles to DSR or cilium in kind, especially we choose pod subnet to be 192.168.0.0/16 or other 10.x.x.x/xx subnets. 

we use our own CNI and boot up multiple kind clusters to simulate our datacenter topology: multiple availability zones, and multiple clusters. our CNI and external tools can make pods in kind clusters access each other. we encounter issues when we set up kind even disabling default CNI. extra NAT make DSR chaos, also IPIP turnal break, as well as bring problems into cilium rules. 

that's why we bring this change to allow blank pod subnet. and kubelet/kube-proxy won't set up such a in-cluster service NAT rule.

aojea:
you can disable kube-proxy or modify the kube-proxy configmap to not use the podCIDR

armstrongli:
> you can disable kube-proxy or modify the kube-proxy configmap to not use the podCIDR

good idea. i'll have a try, and send feedback here

armstrongli:
it doesn't work as expected. there are extra NAT rules added by kindnetd instead of kube-proxy. kindnetd consumes the PodSubnet as the parameter and set up extra NAT rules out of kube-proxy's control. that's the place brings troubles. 

```yaml
apiVersion: v1
kind: Pod
metadata:
  # ...
  name: kindnet-92pbp
  namespace: kube-system
  ownerReferences:
  - apiVersion: apps/v1
    blockOwnerDeletion: true
    controller: true
    kind: DaemonSet
    name: kindnet
  # ...
spec:
  // ...
  containers:
  - env:
    # ...
    - name: POD_SUBNET
      value: 10.244.0.0/16 # here is the place used by kindnet
      image: our.company.imagehub/foo/kindnetd:v20230618-dirty
    # ...
```

in the runtime, it has such a NAT rule to tell local traffic doesn't go through NAT:

```
# iptables -L -t nat
...
Chain POSTROUTING (policy ACCEPT)
...
KIND-MASQ-AGENT  all  --  anywhere             anywhere             ADDRTYPE match dst-type !LOCAL /* kind-masq-agent: ensure nat POSTROUTING directs all non-LOCAL destination traffic to our custom KIND-MASQ-AGENT chain */

Chain KIND-MASQ-AGENT (1 references)
...
RETURN     all  --  anywhere             10.244.0.0/16        /* kind-masq-agent: local traffic is not subject to MASQUERADE */
MASQUERADE  all  --  anywhere             anywhere             /* kind-masq-agent: outbound traffic is subject to MASQUERADE (must be last in chain) */
```

the chain is out of kube-proxy's control and any traffic on 192.168.0.0/16 or 10.x.x.x/xx go through NAT and the server getting requests are confused and loses DSR. the cilium rules block traffic, too, because the traiffic is not from source CIDR beause of the NAT.

additionally, the clusterCIDR also can only support 1 ipv4 cidr and 1 ipv6 cidr. it doesn't support multiple cidrs.

armstrongli:
there is also a race condition between kube-proxy and patch to customize the configmap for kube-proxy daemonset .

PR #3287, opened 2023-06-28
add no masquerade subnet support in kind netd

armstrongli:
original logic is to pick the no masquerade subnet from pod subnets. the change is to make kindnetd add extra subnets skip masquerading


aojea:
why do we need this in kind?

I'm maintaining the kindnet cni plugin in my own fork in case people need some special options that are reasonable https://github.com/aojea/kindnet but I do not see the need here in kind

Issue #3277, opened 2023-06-15
Rancher-Desktop [Alpine] can't create cluster with v0.20.0 [Previously Also Colima]


pmalek:
**What happened**:

After updating to v0.20.0 I cannot create a cluster anymore.

I'm using Mac with [colima](https://github.com/abiosoft/colima) 

```
Creating cluster "colima" ...
 ✓ Ensuring node image (kindest/node:v1.27.2) 🖼
 ✗ Preparing nodes 📦
Deleted nodes: ["colima-control-plane"]
ERROR: failed to create cluster: command "docker run --name colima-control-plane --hostname colima-control-plane --label io.x-k8s.kind.role=control-plane --privileged --security-opt seccomp=unconfined --security-opt apparmor=unconfined --tmpfs /tmp --tmpfs /run --volume /var --volume /lib/modules:/lib/modules:ro -e KIND_EXPERIMENTAL_CONTAINERD_SNAPSHOTTER --detach --tty --label io.x-k8s.kind.cluster=colima --net kind --restart=on-failure:1 --init=false --cgroupns=private --publish=127.0.0.1:52490:6443/TCP -e KUBECONFIG=/etc/kubernetes/admin.conf kindest/node:v1.27.2@sha256:3966ac761ae0136263ffdb6cfd4db23ef8a83cba8a463690e98317add2c9ba72" failed with error: exit status 125
Command Output: 3236752928bc442ebdaf6bd3b6b164643987d45b1a120ec3cd20ca14cc7f5dd7
docker: Error response from daemon: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "cgroup" to rootfs at "/sys/fs/cgroup": mount cgroup:/sys/fs/cgroup/openrc (via /proc/self/fd/7), flags: 0xe, data: openrc: invalid argument: unknown.
```

**What you expected to happen**:

No error and cluster creates successfully

**How to reproduce it (as _minimally_ and precisely as possible)**:

1. Try to create cluster with kind v0.20.0

**Environment:**

- kind version: (use `kind version`): v0.20.0
- Runtime info: (use `docker info` or `podman info`):

  ```
  Client: Docker Engine - Community
   Version:    24.0.2
   Context:    default
   Debug Mode: false
   Plugins:
    buildx: Docker Buildx (Docker Inc.)
      Version:  v0.10.5
      Path:     /usr/local/lib/docker/cli-plugins/docker-buildx
[...]

PR #3290, opened 2023-06-28
Add API for CDI --devices flag in Docker and Podman for mapping GPUs

lukeogg:
This PR adds support for passing GPU parameters to Nvidia Container Toolkit through the CDI specification. Although there is a way to map all GPUs to a single node with device mounts, more granularity is desired. This PR will support mapping various device combinations to different nodes as needed.

Would resolve [Issue 3164](https://github.com/kubernetes-sigs/kind/issues/3164)

- Adds API for devices with a list of strings in the format specified in the CDI specification.

```
  devices:
  - "nvidia.com/gpu=0"
  - "nvidia.com/gpu=1"
```

- Uses the CDI validation package to validate values passed in the devices API.
- Supports both docker and podman
- This relies on the [upstream PR for Docker](https://github.com/docker/cli/pull/4084) being merged and released - presumably in Docker v25.
- Checks the docker version if devices have been specified.
- Add documentation for mapping GPUs.

All GPUs mapped to a single control-plane:

```
{{< codeFromInline lang="yaml" >}}
kind: Cluster
nodes:
- role: control-plane
  devices:
  - "nvidia.com/gpu=all"
```

Specific GPUs mapped to specific worker nodes based on index:

```
kind: Cluster
nodes:
- role: control-plane
- role: worker
  devices:
  - "nvidia.com/gpu=0"
- role: worker
  devices:
  - "nvidia.com/gpu=1"
  ```

### CODE REVIEW

File: go.mod [hunk 1]
```diff
@@ -5,6 +5,7 @@ go 1.16
 require (
 	github.com/BurntSushi/toml v1.0.0
 	github.com/alessio/shellescape v1.4.1
+	github.com/container-orchestrated-devices/container-device-interface v0.5.5-0.20230516140309-1e6752771dc5 // indirect
```
BenTheElder, line 8: This unfortunately seems to pull in many dependencies, one of our core tenants is being extremely easy to embed and we also have to ensure we're in compliance with kubernetes / CNCF dependency licensing policies.

Maybe we should just leave validation to the underlying container runtimes?

lukeogg: Yes, I think underlying validation should be fine. I'll clean it up.

pkg/internal/apis/config/validate_test.go
```diff
@@ -18,9 +18,10 @@ package config
 
 import (
 	"fmt"
-	"sigs.k8s.io/kind/pkg/internal/assert"
 	"testing"
 
+	"sigs.k8s.io/kind/pkg/internal/assert"
+
```
stmcginnis, line 24: Very trivial nit, but if you do an update for any other reason, looks like the formatting should have grouped these two together and sorted them without the space between.

stmcginnis (commented):
/ok-to-test

Looks like this should be fine to podman, and I think the docker version handling makes sense.

Some nice markdown cleanup too. ;)

pkg/internal/apis/config/validate.go
```diff
@@ -192,6 +196,17 @@ func validatePortMappings(portMappings []PortMapping) error {
 	return nil
 }
 
+func validateDevices(devices []string) error {
+	for _, device := range devices {
+		device := strings.TrimSpace(device)
```
elezar, line 201: You should be able to use `"github.com/container-orchestrated-devices/container-device-interface/pkg/parser"` and call `parser.IsQualifiedName(device)` here. This is used in https://github.com/docker/cli/pull/4084 and significantly reduces the dependencies pulled in. 

It should also be fine to allow `podman` and / or `docker` to perform this validation though.

site/content/docs/user/configuration.md
```diff
@@ -281,10 +281,71 @@ nodes:
...
 
+### GPU Support
+
+There are two ways to map GPUs in to a KinD cluster. The first is using the `devices` API and the second is using the `extraMounts` API.
+
+#### Using the Devices API
+
+As a pre-requisite you install the [NVIDIA Container Toolkit](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/install-guide.html) installed on the host.
+
+Using `devices` for GPU support requires Docker v25 or later. See notes on CDI Container Support [here.](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/install-guide.html#container-device-interface-cdi-support)
```
elezar, line 296: It may be clearer to mention that the CDI specifications need to be generated explicitly.

Note, this is applicable to any CDI devices, so calling out support by other vendors here may also be good.

site/content/docs/user/configuration.md
```diff
@@ -281,10 +281,71 @@ nodes:
...
+  - "nvidia.com/gpu=1"
+{{< /codeFromInline >}}
+
+#### Using the Extra Mounts API
+
+GPUs can also be mapped using the `extraMounts` API. This method passes a list of GPUs to inject as volume mounts rather than the environment variable `NVIDIA_VISIBLE_DEVICES`.
+
+Steps to enable this:
+
+1. Add nvidia as your default runtime in /etc/docker/daemon.json
```
elezar, line 332: Assuming that you have a recent version of the NVIDIA Container Toolkit installed, running 
```
sudo nvidia-ctk runtime configure --runtime=docker --set-as-default
```
should set the default runtime.

site/content/docs/user/configuration.md
```diff
@@ -281,10 +281,71 @@ nodes:
...
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+nodes:
+- role: control-plane
+  extraMounts:
+  - hostPath: /dev/null
+    containerPath: /var/run/nvidia-container-devices/all
+{{< /codeFromInline >}}
+
+Note: this method only support adding `all` GPUs to a single node. If you want to add specific GPUs to specific nodes, you will need to use the `devices` API.
```
elezar, line 347: I don't know whether this is strictly true. Using `/var/run/nvidia-container-devices/0` or `/var/run/nvidia-container-devices/{{DEVICE_UUID}}` should allow individual devices to be made available.

cc @klueska

elezar (commented):
LGTM.

Some minor comments on the CDI device and documentation side of things.

site/content/docs/user/configuration.md
klueska: As mentioned in https://github.com/kubernetes-sigs/kind/pull/3257#issuecomment-1607287275, we can't achieve any finer granularity than `all` (regardless of whether we do volume mounts or use CDI) because the node's container is started with --privileged. Even if you tell it to selectively inject just a single GPU, the node will be able to see all of them because `--privileged` pulls in all device nodes under `/dev`.

pkg/internal/apis/config/validate.go
lukeogg: I'll give that a try.

pkg/cluster/internal/providers/docker/provision.go
```diff
@@ -255,6 +255,20 @@ func runArgsForNode(node *config.Node, clusterIPFamily config.ClusterIPFamily, n
...
+		if ver != "dev" || strings.Split(ver, ".")[0] < "25" {
+			return nil, errors.Errorf("using devices api in kind requires Docker >= v25, but found %q", ver)
+		}
+
+		// Append args for each device
+		for _, device := range node.CDIDevices {
+			args = append(args, "--device", strings.TrimSpace(device))
+		}
+	}
+
```
klueska, line 271: Will also allow passing "standard" devices via the `CDIDevices` setting? It looks like there is a new `validate()` call that will prevent this though. Is the validate call made sometime before this is done here?

pkg/internal/apis/config/validate.go
klueska: Please update to what Evan suggests to reduce the dependencies pulled in.

site/content/docs/user/configuration.md
```diff
@@ -281,10 +281,71 @@ nodes:
...
+Specific GPUs mapped to specific worker nodes based on index:
+
+{{< codeFromInline lang="yaml" >}}
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+nodes:
+- role: control-plane
+- role: worker
+  devices:
+  - "nvidia.com/gpu=0"
```
klueska, line 320: Note that this will *not* give you access to just GPU 0, but rather *all* GPUs on the system.

This is because --privileged is passed to the docker call that creates the node, thus giving it access to *all* devices under /dev (including all GPU devices).

Individual GPU injection will only be supported once if/when kind is able to start nodes without --privileged.

site/content/docs/user/configuration.md
```diff
@@ -281,10 +281,71 @@ nodes:
...
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+nodes:
+- role: control-plane
+  extraMounts:
+  - hostPath: /dev/null
+    containerPath: /var/run/nvidia-container-devices/all
+{{< /codeFromInline >}}
+
+Note: this method only support adding `all` GPUs to a single node. If you want to add specific GPUs to specific nodes, you will need to use the `devices` API.
```
klueska, line 347: This limitation of `all` is true even in the CDI case.

pkg/internal/apis/config/validate.go
lukeogg: @BenTheElder What do you think?

klueska (commented):
In general, this seems reasonable. The only real issue I see is the limitation around not being able to inject individual GPUs because of the --privileged flag. 

I'm assuming this will trip up many users thinking they should be able to selectively give different nodes access to different GPUs (based on their CDI device name), but then suddenly all nodes have access to all GPUs.

site/content/docs/user/configuration.md
lukeogg: Ok, this reduces the value of adding an API for CDI devices in my opinion.

site/content/docs/user/configuration.md
lukeogg: Ok, this reduces the value of adding an API for CDI devices in my opinion.

pkg/cluster/internal/providers/docker/provision.go
lukeogg: I am using `parser.ParseQualifiedName(device)` instead of `parser.ParseQualifiedName(device)`. The results are the same except you get the additional error message about why your device string is invalid. I think this is preferable. ParseQualifiedName() is called in IsQualifiedName().

This doesn't change the dependencies at all. There are tests on `github.com/container-orchestrated-devices/container-device-interface/pkg/parser` pulling in the additional packages. Maybe we should put the tests in a different package?

pkg/internal/apis/config/validate.go
lukeogg: @klueska I am using parser.ParseQualifiedName(device) instead of parser.ParseQualifiedName(device). The results are the same except you get the additional error message about why your device string is invalid. I think this is preferable. ParseQualifiedName() is called in IsQualifiedName().

This doesn't change the dependencies at all. There are tests on github.com/container-orchestrated-devices/container-device-interface/pkg/parser pulling in the additional packages. Maybe we should put the tests in a different package? @elezar wdyt?

elezar: I can update the tests in that package to use a different import for comparison or to be in the `parser_test` package if require. As a matter of interest, which package that is equivalent to `require` would be preferred?

site/content/docs/user/configuration.md
elezar: To be clear, this is the same for all other mechanisms too. The `--privileged` flag will mean that any container will have access to all devices and the only value add is that the driver libraries are also injected.

site/content/docs/user/configuration.md
lukeogg: Doing some testing today and this does allow access to all GPUs. I'll update the documentation. It is fairly convenient with Kind, however, as you don't need to inject drivers or install them on the node images.

pkg/internal/apis/config/validate.go
lukeogg: I don't personally have a preference - I haven't found any with fewer dependencies. I was thinking use the `parser_test` package to keep things very lightweight.

elezar: I have created https://github.com/container-orchestrated-devices/container-device-interface/pull/149 with an update to the test package. This should prevent `github.com/stretchr/testify/require` from getting pulled in.

site/content/docs/user/configuration.md
```diff
@@ -281,10 +281,71 @@ nodes:
...
+  - "nvidia.com/gpu=1"
+{{< /codeFromInline >}}
+
+#### Using the Extra Mounts API
+
+GPUs can also be mapped using the `extraMounts` API. This method passes a list of GPUs to inject as volume mounts rather than the environment variable `NVIDIA_VISIBLE_DEVICES`.
+
+Steps to enable this:
+
+1. Add nvidia as your default runtime in `/etc/docker/daemon.json` If you have the [NVIDIA Container Toolkit installed](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html) this can be done with: `sudo nvidia-ctk runtime configure --runtime=docker --set-as-default`
```
elezar, line 332:
```suggestion
1. Add `nvidia` as your default runtime in `/etc/docker/daemon.json` If you have the [NVIDIA Container Toolkit installed](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html) this can be done with: `sudo nvidia-ctk runtime configure --runtime=docker --set-as-default`
```

site/content/docs/user/configuration.md
elezar: > Ok, this reduces the value of adding an API for CDI devices in my opinion.

This isn't true. The use of the device mounts uses our "legacy" stack which has a number of shortcomings that using CDI addresses. It also means that users can access devices from other vendors such as Intel or their own device / resource definitions.

elezar (commented):
Thanks @lukeogg. Some minor comments / suggestions from my side.

site/content/docs/user/configuration.md
```diff
@@ -436,9 +497,9 @@ nodes:
           enable-admission-plugins: NodeRestriction,MutatingAdmissionWebhook,ValidatingAdmissionWebhook
 {{< /codeFromInline >}}
 
-On every additional node configured in the KIND cluster, 
+On every additional node configured in the KIND cluster,
 worker or control-plane (in HA mode),
-KIND runs `kubeadm join` which can be configured using the 
+KIND runs `kubeadm join` which can be configured using the
```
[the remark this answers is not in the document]
BenTheElder: The docs consistently use `KIND` or `kind` but not `KinD`. We're not using the acronym anymore as it may be podman or in the future something else.

site/content/docs/user/configuration.md
```diff
@@ -281,10 +281,71 @@ nodes:
...
+
+There are two ways to map GPUs in to a KinD cluster. The first is using the `devices` API and the second is using the `extraMounts` API.
+
+#### Using the Devices API
+
+As a pre-requisite you install the [NVIDIA Container Toolkit](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/install-guide.html) installed on the host.
+
+Using `devices` for GPU support requires Docker v25 or later. A [CDI specification](https://github.com/container-orchestrated-devices/container-device-interface) will need to be generated for your device. For Nvidia GPU devices see notes [here.](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/install-guide.html#container-device-interface-cdi-support)
+
+GPU devices can be mapped to Kind node copntainers with the devices API:
```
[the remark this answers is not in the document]
BenTheElder: `KIND` or `kind`. https://github.com/kubernetes-sigs/kind/pull/3290#discussion_r1304870057

site/content/docs/user/configuration.md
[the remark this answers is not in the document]
BenTheElder: We should put a note in https://kind.sigs.k8s.io/docs/contributing/development/#documentation

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.16
 require (
 	github.com/BurntSushi/toml v1.0.0
 	github.com/alessio/shellescape v1.4.1
+	github.com/container-orchestrated-devices/container-device-interface v0.6.0
```
BenTheElder, line 8: regarding dependency management, I'm working on https://github.com/kubernetes-sigs/kind/pull/3335 to switch the modules to 1.17+, which will enable https://go.dev/ref/mod#graph-pruning and make the transitive dependencies clear.

BenTheElder: just glancing at the go.sum changes, there's a concerningly large set of deps, but we can't tell with pre 1.17 go.mod files what is actually in the binary.

It remains important that `kind` the go libraries and CLI are trivially embeddable with minimal dependencies which also must meet kubernetes / CNCF's (allowed licenses) dependency standards. We have a number of users using kind directly in test tools etc and we don't want to bring in more dependencies.

https://github.com/cncf-tags/container-device-interface/blob/main/go.mod has a lot of deps for what I would expect to just be a spec/parser.

pkg/internal/apis/config/validate.go
BenTheElder: Sorry, a few things came up back in the real world and here with #3277 etc., the dependency discussion wound up forked in https://github.com/kubernetes-sigs/kind/pull/3290/files#r1304873501

I'm going to try to get https://github.com/kubernetes-sigs/kind/pull/3335 after which when this PR is rebased and `go mod tidy` it should be clearer what we're actually talking about for the new dependencies.

See linked comment for stance on deps.

BenTheElder: Currently for example we have a lot of CRI inspired types with in-tree dependency-free code instead, to avoid creating dep hell for our users. That may not be so reasonable here.

site/content/docs/user/configuration.md
BenTheElder: > the only value add is that the driver libraries are also injected.

I don't supposed we can just trigger driver library injection directly in a simple way and reduce the API surface to "yes I'd like the GPU drivers"?

lukeogg:
I will get back to this in the next week or so. Been out for a bit.

site/content/docs/user/configuration.md
klueska: Now that CDI support is available in docker, I wanted to revive this thread.

Can someone clarify why `--privileged` is needed in `kind` over something `--cap-add SYS_ADMIN`? As far as I understand things `--privileged` is basically just `(--cap-add SYS_ADMIN` + access to all devices)`. Are there specific devices that `kind` needs access to or is it really just about getting `CAP_SYS_ADMIN` privileges?

aojea: not my area of expertise, but there are some places you do a lot of "privileged things"

https://github.com/kubernetes-sigs/kind/blob/main/images/base/files/usr/local/bin/entrypoint

https://github.com/kubernetes-sigs/kind/blob/7c3e01fc1c25d833dd9938789a8f26b184472e8f/pkg/cluster/internal/providers/docker/node.go#L105-L108

I'm not sure but we may need NET_ADMIN too

BenTheElder: It's been a really long time since I've dug into the specifics, but I don't think that's quite it for `--privileged` which is special cased and not directly replicable with other flags IIRC. I definitely attempted to avoid this and IIRC even cap-add=ALL won't work, docker does a few other things like making /sys RW

lukeogg: That is how I understood it as well. It is more than just the capabilities that get added and there is no way to replicate with just other CLI options.

pkg/cluster/internal/providers/docker/provision.go
```diff
@@ -255,6 +255,20 @@ func runArgsForNode(node *config.Node, clusterIPFamily config.ClusterIPFamily, n
 		args = append(args, "-e", "KUBECONFIG=/etc/kubernetes/admin.conf")
 	}
 
+	// Append CDI device args (used for GPU support)
+	if len(node.CDIDevices) > 0 {
+		// Check for docker > 25
+		ver := Version()
+		if ver != "dev" || strings.Split(ver, ".")[0] < "25" {
+			return nil, errors.Errorf("using devices api in kind requires Docker >= v25, but found %q", ver)
+		}
```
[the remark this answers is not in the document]
aojea: I prefer this approach, rather than make kind to have awareness of the provider versions

site/content/docs/user/configuration.md
klueska: I chatted with @aojea at Kubecon the other day and he convinced me that it is impractical to remove `--privileged` when starting the docker containers that wrap a worker node. Some end users may be relying on `--privileged`, even if `kind` itself could be made to enumerate the settings necessary to give it "privileged-like" functionality without actually passing `--privileged`.

Therefore, the long-term solution for this will require running some sort of "post-processing" script that removes access to any GPUs that were not selected for injection. Since `kind` is not intended for production, I don't think we need to actually remove `cgroup` access to these GPUs, but rather just make them inaccessible to any workloads that want to make use of them.

For example, something like the following could be used to remove access to all GPUs except GPU 0 on a DGX A100 server
```
cp /proc/driver/nvidia/params root/gpu-params
sed -i 's/^ModifyDeviceFiles: 1$/ModifyDeviceFiles: 0/' root/gpu-params
mount --bind root/gpu-params /proc/driver/nvidia/params

export GPU_EXCLUDE_LIST=1,2,3,4,5,6,7
eval rm -rf /dev/nvidia{${GPU_EXCLUDE_LIST}}
```

Where the first three lines make it so that calls into `nvidia-smi` / `libnvidia-ml.so` do not attempt to recreate nvidia device nodes or reset their permissions if tampered with. And the last two lines remove the device nodes for all GPUs except GPU 0.

This would need to be extended in order to work for MIG devices, but it's a starting point to begin thinking about what we want to do here to support selective GPU injection into different kind nodes.

BenTheElder: Yeah, this is what I'd been thinking, it's going to be a bit gross and I'm not sure how to generalize it and give a good user experience. It is immediately possible to do this for testing by `docker exec` to your target nodes and running these steps though.

klueska: I spent today putting this together:
https://github.com/klueska/kind-with-gpus-examples

It should be a good starting point for using GPUs with kind independent of this PR

klueska: I've also pinged the docker maintainers that helped us push CDI through to see what their thoughts are here:
https://dockercommunity.slack.com/archives/C04MZQZJE94/p1712086101448089

klueska: Was pointed to this on that thread:
https://github.com/moby/moby/pull/39702#issuecomment-936707664

BenTheElder: If https://github.com/moby/moby/issues/47663 becomes available we'll have to version detect to use it to avoid breaking all of the users on older docker installs, and I'll be pretty wary of the branching codepath behaving differently vs --privileged.

We have only once required a newer docker version (when we started to require cgroupns=private be available) which caused a lot of consternation even though that change was aimed at mitigating the runc misc issues that caused kind to very flaky/broken ...

BenTheElder: It would've been nice when kind was being built, but now we have to consider that people undoubtedly have test environments that depend on being in a --privileged node :/

BenTheElder:
I think at this point we can assume a new PR instead?

Thank you anyhow 🙏

Issue #3293, opened 2023-07-03
Argument to controll passing proxy envs from host


shokerplz:
**What would you like to be added**:
An argument like "--no-env-pass" which would allow to not call getProxyEnv in pkg/cluster/internal//providers/docker/provision.go and pkg/cluster/internal//providers/podman/provision.go.

**Why is this needed**:
In my case I have a proxy listening on my local machine and my env vars are looking like HTTPS_PROXY="http://127.0.0.1:3128", so they are being passed to all kind containers and kind clusters doesn't work because of that.
I think it should be possible to disable passing envs.

neolit123:
not a big fan of the --no-foo-bar flag naming.
perhaps a more preferable way would be to have a positive naming such as --foo-bar which is defaulted to true.

in terms of the feature itself, it seems reasonable to me, but perhaps it should be added in the kind config and not as CLI flag.

shokerplz:
yeah, `--pass-env false` is probably better then my initial idea. 
But about config, from what I see, only kubernetes cluster related things (like amount of nodes, cni, etc) should be in config and passing proxy env option is not one of such things.

shokerplz:
yes we can live without this feature, but if kind is automatically grabs those envs from host's shell, why don't we add a feature that allows to disable such behaviour?

aojea:
right now they can do `HTTPS_PROXY="" kind create cluster` and you are suggesting they should do `kind create cluster --pass-env false` 

I can't see why adding a dependency to kind like this will benefit users, since they already have to modify their scripts to set the flag, when they can modify the script to override the env variables

BenTheElder:
The `env` tool or else `HTTPS_PROXY=""` seems more appropriate?

The proxy is an environment detail and it is completely standard behavior to respect these env and expect that they are set reasonably.

BenTheElder:
> I can't see why adding a dependency to kind like this will benefit users, since they already have to modify their scripts to set the flag, when they can modify the script to override the env variables

More importantly, third party scripts and tools should not be setting this flag, for most users respecting these envs is the right thing to do and e.g. test scripts in other projects using kind should not set this flag.

But you can set these env when going to use kind, and then theses other scripts will just work.

We could add a new non-standard env to make kind specifically ignore these, but then .... you could just set the existing standardized env?

PR #3295, opened 2023-07-04
Add argument to control passing proxy envs from host

AryanSharma9917:
Added argument to control passing proxy envs from host #3293

Here I added a new feature to kind that allows users to control the passing of proxy environment variables to containers. By introducing a --no-env-pass argument, users can disable the inclusion of proxy environment variables when creating kind clusters. This gives users more control over their development environment, especially when proxy settings may interfere with cluster functionality.

### DIFF
--- a/pkg/cluster/internal/providers/docker/provision.go
+++ b/pkg/cluster/internal/providers/docker/provision.go
@@ -35,7 +35,7 @@ import (
 )
 
 // planCreation creates a slice of funcs that will create the containers
-func planCreation(cfg *config.Cluster, networkName string) (createContainerFuncs []func() error, err error) {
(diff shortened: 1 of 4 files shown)


neolit123 (commented):
i will comment on the issue

stmcginnis:
/title Add argument to control passing proxy envs from host

Looks like you accidentally picked up a few extra commits on this.

stmcginnis:
/retitle Add argument to control passing proxy envs from host

Looks like you accidentally picked up a few extra commits on this.

aojea:
/hold 

env variables can be overriden , no need for flags, https://github.com/kubernetes-sigs/kind/issues/3293#issuecomment-1620345456

BenTheElder:
In the future we should wait for discussion of features before implementing them, per the contributor guide.

I don't think we're in agreement to accept this feature.

PR #3302, opened 2023-07-09
ExtraPortMapping Validation

### LINKED ISSUE #3301: A unique hostPort is not generated when using extraPortMappings without specifying hostPort.
**What happened**:

Creating a kind cluster without any `hostPort` defined in the `extraPortMappings` should use random but unique `hostPort`. However it seems like it's using the same and so fails before cluster creation during validation. This was/is working as expected with kind v0.18.0

There is a workaround if creating only 1 cluster, which is to manually add `hostPort` and select a unique port. However this does not scale when we need to run multiple kind cluster. (eg for demonstrating or testing a multi cluster to replacate a multi region behaviour etc).

The kind cluster to be created successfully

Given the below config file and creating a cluster will fail with the folloing error.
```
ERROR: failed to create cluster: invalid configuration for node 1: invalid portMapping: port mapping with same listen address, port and protocol already configured: <nil>:0/
```

command run: `kind create cluster --config config.yaml`

```yaml
# config.yaml

kind: Cluster
nodes:
  - role: control-plane
  - role: worker
    extraPortMappings:
      - containerPort: 8080
      - containerPort: 8443
```

This error only occurs after #3175 so all version from v0.19.0 on wards.
v0.18.0 continues to work as expected.

**Environment:**

- kind version: (use `kind version`): v0.20.0
- Runtime info: (use `docker info` or `podman info`): Docker engine 24.0.2
- OS (e.g. from `/etc/os-release`): OSX 13.4
- Kubernetes version: (use `kubectl version`): 1.24.15
- Any proxies or other special environment settings?: nil

aroradaman:
Using random host ports for validating extra port mapping defined without host ports  (default behaviour for provider docker and podman)

Fixes #3301

### DIFF
--- a/pkg/apis/config/v1alpha4/default.go
+++ b/pkg/apis/config/v1alpha4/default.go
@@ -31,13 +31,14 @@ func SetDefaultsCluster(obj *Cluster) {
 			},
 		}
 	}
+	if obj.Networking.IPFamily == "" {
+		obj.Networking.IPFamily = IPv4Family
+	}
+
 	// default the nodes
 	for i := range obj.Nodes {
 		a := &obj.Nodes[i]
-		SetDefaultsNode(a)
-	}
-	if obj.Networking.IPFamily == "" {
-		obj.Networking.IPFamily = IPv4Family
+		SetDefaultsNode(a, obj.Networking.IPFamily)
 	}
 	// default to listening on 127.0.0.1:randomPort on ipv4
 	// and [::1]:randomPort on ipv6
@@ -79,12 +80,26 @@ func SetDefaultsCluster(obj *Cluster) {
 }
 
 // SetDefaultsNode sets uninitialized fields to their default value.
-func SetDefaultsNode(obj *Node) {
+func SetDefaultsNode(obj *Node, ipFamily ClusterIPFamily) {
 	if obj.Image == "" {
 		obj.Image = defaults.Image
 	}
 
 	if obj.Role == "" {
 		obj.Role = ControlPlaneRole
+	}
+
+	for i := 0; i < len(obj.ExtraPortMappings); i++ {
+		if obj.ExtraPortMappings[i].ListenAddress == "" {
+			if ipFamily == IPv6Family {
+				obj.ExtraPortMappings[i].ListenAddress = "::"
+			} else {
+				obj.ExtraPortMappings[i].ListenAddress = "0.0.0.0"
+			}
+		}
+
+		if string(obj.ExtraPortMappings[i].Protocol) == "" {
+			obj.ExtraPortMappings[i].Protocol = PortMappingProtocolTCP
+		}
 	}
 }
--- a/pkg/internal/apis/config/validate.go
+++ b/pkg/internal/apis/config/validate.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"sigs.k8s.io/kind/pkg/errors"
@@ -137,7 +138,7 @@ func (n *Node) Validate() error {
 }
 
 func validatePortMappings(portMappings []PortMapping) error {
-	errMsg := "port mapping with same listen address, port and protocol already configured"
+	errMsg := "port mapping with same listen address, host port and protocol already configured"
 
 	wildcardAddrIPv4 := net.ParseIP("0.0.0.0")
 	wildcardAddrIPv6 := net.ParseIP("::")
@@ -152,11 +153,16 @@ func validatePortMappings(portMappings []PortMapping) error {
 	}
 
 	for _, portMapping := range portMappings {
+		// skipping validation if host port is not defined
+		if portMapping.HostPort == 0 {
+			continue
+		}
+
 		addr := net.ParseIP(portMapping.ListenAddress)
 		addrString := addr.String()
 
 		portProtocol := formatPortProtocol(portMapping.HostPort, portMapping.Protocol)
-		possibleErr := fmt.Errorf("%s: %s:%s", errMsg, addrString, portProtocol)
+		possibleErr := fmt.Errorf("%s: %s/%s", errMsg, net.JoinHostPort(addrString, strconv.Itoa(int(portMapping.HostPort))), portMapping.Protocol)
 
 		// in golang 0.0.0.0 and [::] are equivalent, convert [::] -> 0.0.0.0
 		// https://github.com/golang/go/issues/48723
--- a/pkg/internal/apis/config/validate_test.go
+++ b/pkg/internal/apis/config/validate_test.go
@@ -18,7 +18,6 @@ package config
 
 import (
 	"fmt"
-	"sigs.k8s.io/kind/pkg/internal/assert"
 	"testing"
 
 	"sigs.k8s.io/kind/pkg/errors"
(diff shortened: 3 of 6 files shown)


pkg/internal/apis/config/validate.go
```diff
@@ -151,7 +152,39 @@ func validatePortMappings(portMappings []PortMapping) error {
...
 	for _, portMapping := range portMappings {
+		// using default listen address
+		if portMapping.ListenAddress == "" {
+			portMapping.ListenAddress = wildcardAddrIPv4.String()
+		}
+
+		// using random hostport
+		// this is just for validation purpose, actual random host port will be assigned by pkg/cluster/internal/providers
+		if portMapping.HostPort == 0 {
+			portMapping.HostPort = getRandomHostPort()
```
aojea, line 180: I don't think we should fake the random port, it looks like if this value is empty we have to assume there will not be conflict and just skip its validation

aroradaman: makes more sense :)

pkg/internal/apis/config/validate_test.go
```diff
@@ -432,45 +432,66 @@ func TestPortValidate(t *testing.T) {
...
+		},
+		{
+			testName: "duplicate container ports",
+			portMappings: []PortMapping{
+				newPortMapping("", 0, 1000, ""),
+				newPortMapping("", 0, 2000, ""),
+				newPortMapping("", 0, 3000, ""),
+				newPortMapping("", 0, 3000, ""),
+			},
+			expectErr: "",
```
aojea, line 466: both testcases are the same, aren't they?

aroradaman: the second one has a duplicate containerPort, which is allowed by both kind and docker.

aojea: oh, heh, I started at it for one minute and didn't notice it XD

pkg/internal/apis/config/validate.go
```diff
@@ -152,11 +153,26 @@ func validatePortMappings(portMappings []PortMapping) error {
...
+
+		// using default listen address
+		if portMapping.ListenAddress == "" {
+			portMapping.ListenAddress = wildcardAddrIPv4.String()
+		}
+
+		// using default protool
+		if portMapping.Protocol == "" {
+			portMapping.Protocol = PortMappingProtocolTCP
+		}
```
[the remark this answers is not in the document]
aojea: point is we should default things in defaulting stage, not here in validation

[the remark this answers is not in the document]
aroradaman: should I remove defaults from `pkg/cluster/internal/providers/<provider>/provision.go` then and add it as part of **SetDefaultsCluster** in default stage .. defaulting with respect to provider ?

https://github.com/kubernetes-sigs/kind/blob/3610f606516ccaa88aa098465d8c13af70937050/pkg/cluster/internal/providers/docker/provision.go#L363-L377

https://github.com/kubernetes-sigs/kind/blob/3610f606516ccaa88aa098465d8c13af70937050/pkg/cluster/internal/providers/podman/provision.go#L376-L390

pkg/apis/config/v1alpha4/default.go
```diff
@@ -79,12 +79,26 @@ func SetDefaultsCluster(obj *Cluster) {
...
+				obj.ExtraPortMappings[i].ListenAddress = defaults.ExtraPortMappingListenAddressIPv6
+			} else {
+				obj.ExtraPortMappings[i].ListenAddress = defaults.ExtraPortMappingListenAddressIPv4
+			}
+		}
+
+		if string(obj.ExtraPortMappings[i].Protocol) == "" {
+			obj.ExtraPortMappings[i].Protocol = PortMappingProtocolTCP
+		}
+	}
```
aojea, line 103: @BenTheElder we just pre-default here, I think is ok, but it will better to have your thoughts

aojea:
/lgtm
/assign @BenTheElder 

for final approval

aojea:
/lgtm cancel

don't we need to modify `./pkg/internal/apis/config/default.go` too?

pkg/apis/config/v1alpha4/default.go
aroradaman: I will also remove the defaulting piece from docker/podman providers  if we finalize this approach.

aroradaman:
@aojea  Is `./pkg/internal/apis/config/default.go` generated by code-generator ?
If yes, can you point me to documentation for the same ?

pkg/apis/config/defaults/networking.go
```diff
@@ -0,0 +1,55 @@
...
+const ServiceSubnetIPv4 = "10.96.0.0/16"
+
+// ServiceSubnetIPv6 is the default IPv6 subnet for the Networking.PodSubnet field
+const ServiceSubnetIPv6 = "fd00:10:96::/112"
+
+// ServiceSubnetDualStack is the default DualStack subnet for the Networking.ServiceSubnet field
+const ServiceSubnetDualStack = "10.96.0.0/16,fd00:10:96::/112"
+
+// ExtraPortMappingListenAddressIPv4 is the default IPv4/DualStack listen address for ExtraPortMappings
+const ExtraPortMappingListenAddressIPv4 = "0.0.0.0"
```
BenTheElder, line 52: did we really need to make all of these public variables?

even if the values are documented, that doesn't mean we've committed to thoughtful names yet and that the specifics of how traffic works on a future provider will be the same.

BenTheElder (commented):
I've been out sick and I'm going to need some more time to think about how to better structure this, but I don't think exporting all of these values is the answer.

Thanks for working on this problem!

prahaladramji:
> I've been out sick and I'm going to need some more time to think about how to better structure this, but I don't think exporting all of these values is the answer.
> 
> Thanks for working on this problem!

Just following up to find out what's the current thoughts on this PR, I'm looking for a resolution to https://github.com/kubernetes-sigs/kind/issues/3301.

@BenTheElder Is the feedback more around the implementation details of this PR or is the above mentioned issue the desired behaviour of kind moving forward post v0.18.0?

BenTheElder:
I was hoping for a response to the existing comments before reviewing further.

pkg/apis/config/v1alpha4/default.go
BenTheElder: I guess this is OK, we can define the behavior based on docker. Not the first time kubernetes has done that.

pkg/apis/config/defaults/networking.go
BenTheElder: Still not thrilled about this aspect, we're exporting a lot of internal choices here that should be versioned with the API and we're doing so as a ton of unversioned globals ...

BenTheElder: This change does not seem necessary for the purpose of this PR?

aroradaman: moved back to versioned defaulting.

aroradaman:
@prahaladramji I don't think retest would help here, and also the error is not related to this PR. 

@BenTheElder do we still need `pull-kind-e2e-kubernetes-1-24` given that it has already reached EOL. I also see `pull-kind-e2e-kubernetes-1-28` to be missing.

aroradaman:
ping @BenTheElder

prahaladramji:
The 1 failing check seemes to be related to vagrant as part of the ci step and not related to this change.

A general look at that error seems like it could just be flakey vagrant step (potentially a retry may work). But if not, I'm happy to try and help look at fixing that step.

aroradaman:
/close in favor of https://github.com/kubernetes-sigs/kind/pull/3513
cc. @prahaladramji

PR #3315, opened 2023-07-22
Use `ImplementationSpecific` instead of `Prefix` for pathType example

sontek:
Since the example is using a regular express to match more than the prefix I think the recommendation would be to use
`ImplementationSpecific` instead.  You receive:

```
❯ k apply -f ingress.yml 
Warning: path /httpbin(/|$)(.*) cannot be used with pathType Prefix
```

When doing that.

### DIFF
--- a/site/static/examples/ingress/usage.yaml
+++ b/site/static/examples/ingress/usage.yaml
@@ -62,14 +62,14 @@ spec:
   rules:
   - http:
       paths:
-      - pathType: Prefix
+      - pathType: ImplementationSpecific
         path: /foo(/|$)(.*)
         backend:
           service:
             name: foo-service
             port:
               number: 8080
-      - pathType: Prefix
+      - pathType: ImplementationSpecific
         path: /bar(/|$)(.*)
         backend:
           service:


neolit123 (commented):
/lgtm

tao12345666333:
Thanks for your report.

This is because in the new version of the ingress-nginx project, some restrictions have been added to avoid security issues.

/lgtm

sontek:
@tao12345666333 @neolit123 Looks like we need more than the lgtm to get this merged

aojea:
unfortunately the ingress section grew organically and there are multiple implementations, is this compatible with all of them?

https://github.com/kubernetes-sigs/kind/pull/3281

BenTheElder:
@aojea I'm cutting it down to ingress-nginx and linking out to other guides, we should use whatever is appropriate for ingress-nginx and punt on the others for now https://github.com/kubernetes-sigs/kind/pull/3672

4llan:
The proposed changes in this PR fixes #3356 and has the correct implementation that fixes the warning of ingress-nginx with 0 colateral damage.

aojea:
> The proposed changes in this PR fixes #3356 and has the correct implementation that fixes the warning of ingress-nginx with 0 colateral damage.

it does not guarantee to work for other ingresses

4llan:
Quote from the docs:
> Deploy an Ingress controller, we document **Ingress NGINX here** but other ingresses may work including Contour and Kong, you should follow their docs if you choose to use them.

It's kind of misleading if the user guide says that it's documenting "Ingress NGINX", but fails to do so. To me, as a lay user learning this tool, the guide should be consistent.

Don't get me wrong, I'm just trying to collaborate (as a lay user). Thank you all for the tool and the documentation :)

4llan:
#3763 address this issue, has the same proposed solution, was submitted 1 year later and got merged. So, this PR should be closed.

stmcginnis:
Sorry this was missed when the new PRs came through. Glad the issue is resolved though. Thanks for the PR and the work on this.

/close

aojea:
we are going to change the section to have a better UX and focus on the standard API https://github.com/kubernetes-sigs/kind/pull/3767

PR #3326, opened 2023-08-09
tune the sequence to allow args take effect from kubeadm

armstrongli:
the default cgroup config is `--runtime-cgroups=/system.slice/containerd.service`. kind allows kubeadm patches to customize the kubelet args, but the default args goes after kubeadm and override the customized value from kubeadm.

### DIFF
--- a/images/base/files/etc/systemd/system/kubelet.service.d/10-kubeadm.conf
+++ b/images/base/files/etc/systemd/system/kubelet.service.d/10-kubeadm.conf
@@ -9,4 +9,4 @@ EnvironmentFile=-/var/lib/kubelet/kubeadm-flags.env
 # the .NodeRegistration.KubeletExtraArgs object in the configuration files instead. KUBELET_EXTRA_ARGS should be sourced from this file.
 EnvironmentFile=-/etc/default/kubelet
 ExecStart=
-ExecStart=/usr/bin/kubelet $KUBELET_KUBECONFIG_ARGS $KUBELET_CONFIG_ARGS $KUBELET_KUBEADM_ARGS $KUBELET_EXTRA_ARGS
+ExecStart=/usr/bin/kubelet $KUBELET_KUBECONFIG_ARGS $KUBELET_CONFIG_ARGS $KUBELET_EXTRA_ARGS $KUBELET_KUBEADM_ARGS


armstrongli:
sample command:

```
root@590325e0eaf2:/data# ./kubelet --runtime-cgroups=/system.slice/containerd.service --runtime-cgroups=/system.slice/containerd.service2 --runtime-cgroups=/system.slice/containerd.service3 --v=4
...
I0809 04:31:45.583283   12219 flags.go:33] FLAG: --runtime-cgroups="/system.slice/containerd.service3"
...
```

images/base/files/etc/systemd/system/kubelet.service.d/10-kubeadm.conf
```diff
@@ -9,4 +9,4 @@ EnvironmentFile=-/var/lib/kubelet/kubeadm-flags.env
 # the .NodeRegistration.KubeletExtraArgs object in the configuration files instead. KUBELET_EXTRA_ARGS should be sourced from this file.
 EnvironmentFile=-/etc/default/kubelet
 ExecStart=
-ExecStart=/usr/bin/kubelet $KUBELET_KUBECONFIG_ARGS $KUBELET_CONFIG_ARGS $KUBELET_KUBEADM_ARGS $KUBELET_EXTRA_ARGS
+ExecStart=/usr/bin/kubelet $KUBELET_KUBECONFIG_ARGS $KUBELET_CONFIG_ARGS $KUBELET_EXTRA_ARGS $KUBELET_KUBEADM_ARGS
```
BenTheElder, line 12: does this match kubeadm upstream? cc @neolit123

neolit123: the NodeRegistrationOptions.KubeletExtraArgs are supposed to be the last user override $KUBELET_EXTRA_ARGS

https://github.com/search?q=repo%3Akubernetes%2Frelease%20KUBELET_CONFIG_ARGS&type=code
`$KUBELET_KUBECONFIG_ARGS $KUBELET_CONFIG_ARGS $KUBELET_KUBEADM_ARGS $KUBELET_EXTRA_ARGS`

BenTheElder:
Neglected to comment back here: Seems like we should not merge this PR then re: https://github.com/kubernetes-sigs/kind/pull/3326#discussion_r1288933887

FWIW, I don't think we expect users to customize this particular flag either, we won't block it but it's unlikely to see priority support.

> the NodeRegistrationOptions.KubeletExtraArgs are supposed to be the last user override ($KUBELET_EXTRA_ARGS)

@armstrongli can you maybe file an issue to discuss your situation and share the config you're using?

PR #3327, opened 2023-08-09
Bump k8s.io dependencies to v0.26.4

### LINKED ISSUE #3324: Current client-go version (v0.26.1) contains possible panic in discovery
Please update kind to use client-go version to v0.26.4+

Why: Possible panic when using client-go v0.26.0 - v0.26.3 (looks like current client-go is affected: v0.26.1) for discovery.

kubernetes issue: https://github.com/kubernetes/kubernetes/issues/119840

liangyuanpeng:
part of #3324


aojea:
we still have to rebuild the images and update everything

BenTheElder:
I edited the PR body so this won't close #3324, since the fix will be in a follow-up PR that adopts the images built after this merges

/lgtm
/approve

liangyuanpeng:
I can working for update the image if it have released.

PR #3328, opened 2023-08-11
fix: ensure proper kubectl is used

rjsadow:
Ensure that the kubectl built is the one used throughout the script. 

ref: https://github.com/kubernetes/test-infra/issues/30345

/cc @BenTheElder @aojea


hack/ci/e2e-k8s.sh
```diff
@@ -73,6 +73,9 @@ build() {
   fi
   # make sure we have e2e requirements
   make all WHAT="cmd/kubectl test/e2e/e2e.test ${GINKGO_SRC_DIR}"
+
+  # Ensure the built kubecl is used instead of system
+  export PATH="${PWD}/bin/kubectl:$PATH"
```
BenTheElder, line 78: I don't think that works? For one thing Kubernetes doesn't build to `bin/`

rjsadow: You're right, I was looking at the wrong dir. I _think_ it'll need to be `export PATH="${PWD}/_output/bin/kubectl:$PATH" . This is based on https://github.com/kubernetes/kubernetes/blob/master/build/root/Makefile#L45-L46 and running e2e-k8s.sh locally.

rjsadow: Still isn't the right fix. Looking more. /hold

aojea: check in the kubernetes/hack folder, there are multiple scripts that looks for the `kubectl` binary there

rjsadow: :sweat:  `PATH="${PWD}/_output/bin/kubectl:$PATH"` vs `PATH="${PWD}/_output/bin:$PATH"`.... Too much time spent on realizing that. Update incoming.

aojea:
> /test pull-kind-conformance-parallel-dual-stack-ipv4-ipv6

not related, this is because https://github.com/kubernetes/kubernetes/issues/119927

This looks correct now

```
+ kubectl version
Client Version: v1.29.0-alpha.0.66+6e0cb243d57592
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.29.0-alpha.0.66+6e0cb243d57592
```

/assign @BenTheElder

hack/ci/e2e-k8s.sh
```diff
@@ -73,6 +73,9 @@ build() {
   fi
   # make sure we have e2e requirements
   make all WHAT="cmd/kubectl test/e2e/e2e.test ${GINKGO_SRC_DIR}"
+
+  # Ensure the built kubecl is used instead of system
```
aojea, line 77:
```suggestion
  # Ensure the built kubectl is used instead of system
```

hack/ci/e2e-k8s.sh
BenTheElder: I forget if we can rely on this always being the case but we can start here. When you build dockerized vs not the k8s build system emits to different subpaths under _output, but they might both be linked to this path? In the past in kubernetes upstream we've had to use a small utility to check multiple locations.

If this is working in CI we can start here though.

PR #3334, opened 2023-08-22
rootless.md: cgroup v2 is enabled by default on most distributions

AkihiroSuda:
cgroup v2 is now enabled by default on Debian, Ubuntu, Rocky, Alma, ArchLinux, etc. in addition to Fedora.

### DIFF
--- a/site/content/docs/user/rootless.md
+++ b/site/content/docs/user/rootless.md
@@ -14,10 +14,9 @@ Starting with kind 0.11.0, [Rootless Docker](https://docs.docker.com/go/rootless
 
 ## Host requirements
 The host needs to be running with cgroup v2.
-
-cgroup v2 is enabled by default on Fedora.
-On other distros, cgroup v2 can be typically enabled by adding `GRUB_CMDLINE_LINUX="systemd.unified_cgroup_hierarchy=1"` to `/etc/default/grub` and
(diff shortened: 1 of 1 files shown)


site/content/docs/user/rootless.md
```diff
@@ -14,10 +14,7 @@ Starting with kind 0.11.0, [Rootless Docker](https://docs.docker.com/go/rootless
 
 ## Host requirements
 The host needs to be running with cgroup v2.
-
-cgroup v2 is enabled by default on Fedora.
-On other distros, cgroup v2 can be typically enabled by adding `GRUB_CMDLINE_LINUX="systemd.unified_cgroup_hierarchy=1"` to `/etc/default/grub` and
```
aojea, line 19: don't you want to keep instructions for users? at least some comment about what they should do if the command does not contain Cgroup Version 2, is not uncommon people ask these questions (yes, I know people should google for it, but 🤷 )

AkihiroSuda: done

AkihiroSuda: Added back the instruction

aojea:
/lgtm
/approve

Thanks

BenTheElder:
Thanks!

PR #3335, opened 2023-08-24
Require go 1.17+

BenTheElder:
This enables better dependency management via https://go.dev/ref/mod#graph-pruning

We previously required 1.16+

We don't want to require especially recent versions because that will break users with `go install` using older packaged golang.

We use currently supported much more recent go for our own builds.

NOTE: there are no docs changes yet, because the docs have users installing v0.20 which is still compatible with go 1.16


BenTheElder:
had to update the tools while working on this, looks like we picked up some new linter issues, working on resolving.

images/kindnetd/cmd/kindnetd/main.go
```diff
@@ -252,7 +252,7 @@ func makeNodesReconciler(cniConfig *CNIConfigWriter, hostIP string, ipFamily IPF
 
 		// obtain the PodCIDR gateway
 		var nodeIPv4, nodeIPv6 string
-		for _, ip := range nodeIPs.List() {
+		for _, ip := range sets.List(nodeIPs) {
```
BenTheElder, line 255: [this is the non-deprecated way to use the sets package]

go.mod
```diff
@@ -1,6 +1,14 @@
...
-go 1.16
+// NOTE: This is the go language version, NOT the compiler version.
+//
+// This controls the *minimum* required go version and therefore available Go
+// language features.
+//
+// See ./.go-version for the go compiler version used when building binaries
+//
+// https://go.dev/doc/modules/gomod-ref#go
+go 1.17
```
aojea, line 11: well , this is changing in 1.21

> For example, a go.mod that says go 1.21.0 with no toolchain line is interpreted as if it had a toolchain go1.21.0 line.

https://go.dev/doc/toolchain

aojea:
just the question about the behavior change in golang with 1.21, the rest LGTM

go.mod
BenTheElder: We should rework everything around .go-version to switch to this toolchain line.

BenTheElder: Thanks, I forgot about this.

BenTheElder:
I'm going to finish this up today, ahead of generally bumping things for a v0.21 cut.

BenTheElder:
It's going to be a bit awkward to re-work this around plumbing GOTOOLCHAIN and I'm not sure what the best answer is, we've definitely got downstreams depending on patching .go-version. It might be best to just break that in favor of `toolchain` in the modules and put a release note. Need to think a bit.

BenTheElder:
I'm punting this to the next release because we really need to get that runc CVE out, but some veersion of the changes here probably need to be high on the list longterm as part of switching to go1.21+

BenTheElder:
Actually, I think the most reasonable answer is just to set both .go-version and toolchain to match (after confirming older go will ignore toolchain in the module file) and then deprecated .go-version and teach the gimme fallback to read it (though just getting 1.21 would be good enough since 1.21 will turn around and download toolchain).

BenTheElder:
rebased and started forcing GOTOOLCHAIN="go${GOVERSION}" everywhere for now.

go.mod
BenTheElder: so, I added a commit to deal with this, but the remaining downside (Also true in k/k currently), is that IDEs won't see all our bash hacks.

what we probably need to do is pivot to setting toolchain in the go mods and then also for the image builds of external binaries, but that's going to force us to decentralize setting the go version unfortunately.

BenTheElder: the current PR is enough to ensure our builds / tools work at least

BenTheElder:
> pull-kind-conformance-parallel-ga-only — Pod got deleted unexpectedly   
/retest

BenTheElder:
@aojea go 1.22 released so go 1.20 will be EOL and we'll have to sort this out. Will aim to look again later this week for better iteration, but the current version should be usable.

.go-version
```diff
@@ -1 +1 @@
-1.20.13
+1.21.6
```
aojea, line 1: this should be 1.22 no?

BenTheElder: this PR has been around awaiting review for a while :-)

BenTheElder: we're still on 1.20.13 @ HEAD

images/base/Dockerfile
```diff
@@ -141,6 +143,7 @@ RUN git clone --filter=tree:0 "${RUNC_CLONE_URL}" /runc \
     && cd /runc \
     && git checkout "${RUNC_VERSION}" \
     && eval "$(gimme "${GO_VERSION}")" \
+    && export GOTOOLCHAIN="go${GO_VERSION}" \
```
aojea, line 146: line159 seems is missing this export too

images/kindnetd/go.mod
```diff
@@ -1,6 +1,6 @@
 module sigs.k8s.io/kind/images/kindnetd
 
-go 1.13
+go 1.18
```
aojea, line 3: why don't you move to latest here then?

BenTheElder: it matched kubernetes at the time. note that this is not the go version used to compile, this is the language level features used.

images/base/Dockerfile
BenTheElder: yes. I will add this and bump up the go versions again

images/kindnetd/go.mod
BenTheElder: let's leave this for a later PR, this PR is already a lot and I don't want to deal with any breakage while trying to get the rest of the PR in.

images/base/Dockerfile
BenTheElder: fixed this one.

BenTheElder:
TODO: 
- rebump everything to latest
- explore using go.mod gotoolchain instead (need to handle things like the gimme fallback, and approach to keep these in sync or not)

leaving for follow-ups, I want to get in the core "we're unblocked to manage go versions with 1.21+ without switching language version to 1.21+" so we can collectively iterate from there.

BenTheElder:
There is a new linter failure from nerdctl code, fixing ...

BenTheElder:
/retest
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/3335/pull-kind-e2e-kubernetes-1-28/1774934849065324544

network glitch during cloning, not related.

BenTheElder:
hrm, that's not good.

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-kind-push-base-image/1774949333360185344

BenTheElder:
ah, I know what this is. patch coming in a bit (dealing with some internal work atm)

BenTheElder:
base is building again

PR #3342, opened 2023-08-31
Bump dependencies

aojea:


BenTheElder (commented):
/lgtm
/approve
with the understanding that we may have to roll this back after actually switching to images containing it and getting test coverage, runc upgrades need care.

aojea:
Dims update the GCE  jobs with these versions

BenTheElder:
> Dims update the GCE jobs with these versions

GCE has less issues than kind because it only has kubelet <> containerd/runc, we have another layer (or more).

PR #3294, opened 2023-07-04
Switch to debian bookworm as base image

saschagrunert:
Use the latest debian distribution for the haproxy and local-path-helper images.

Required by https://github.com/kubernetes/kubernetes/pull/118996

PR #3350, opened 2023-09-05
Fix haproxy image build

saschagrunert:
Inherited from:

- https://github.com/kubernetes/release/pull/3235
- https://github.com/kubernetes/release/pull/3237

As well as bumping the haproxy version.

Tested via:

```
docker buildx build --platform=linux/amd64,linux/arm64  --progress=auto -t gcr.io/k8s-staging-kind/haproxy:v20230905-7dc7aad7 --pull --build-arg GO_VERSION=1.20.4  .
```

Follow-up of https://github.com/kubernetes-sigs/kind/pull/3294

### DIFF
--- a/images/haproxy/Dockerfile
+++ b/images/haproxy/Dockerfile
@@ -28,7 +28,7 @@ RUN [ ! -f /usr/share/copyrights.tar.gz ] || tar -C / -xzvf /usr/share/copyright
 # - bash (ldd is a bash script and debian-base removes bash)
 # - procps (for `kill` which kind needs)
 RUN apt update && \
-    apt install -y --no-install-recommends haproxy=2.2.\* \
+    apt install -y --no-install-recommends haproxy=2.6.\* \
       procps bash
 
 # copy in script for staging distro provided binary to distroless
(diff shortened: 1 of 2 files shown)


images/haproxy/Dockerfile
```diff
@@ -28,7 +28,7 @@ RUN [ ! -f /usr/share/copyrights.tar.gz ] || tar -C / -xzvf /usr/share/copyright
 # - bash (ldd is a bash script and debian-base removes bash)
 # - procps (for `kill` which kind needs)
 RUN apt update && \
-    apt install -y --no-install-recommends haproxy=2.2.\* \
+    apt install -y --no-install-recommends haproxy=2.6.\* \
```
aojea, line 31: I can't remember now why we had to stick on 2.2, I think that there was some behavior change we depended on it

aojea: just switch back to the prevous debian version, we'll revisit later 
cc: @BenTheElder

images/haproxy/stage-binary-and-deps.sh
```diff
@@ -41,19 +41,30 @@ package_to_copyright() {
 # stage_file stages the filepath $1 to $2, following symlinks
 # and staging copyrights
 stage_file() {
-    cp -a --parents "${1}" "${2}"
+    # /lib is a symlink to /usr/lib in debian 12, means we just stick to
```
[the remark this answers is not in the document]
aojea: we don't need to modify this file, we need to fix `images/local-path-helper`

stmcginnis:
Closing as updated work is being done in #4005.

/close

PR #3352, opened 2023-09-05
Added iptables switch

cheina97:
# Description

This PR adds the possibility to switch from iptables legacy/nf_tables from a config file.

A field has been added inside config files, called **iptablesMode**
It can accept 3 values:
- auto: the default value if the field is not specified, it preserves the old behavior
- nf_tables: force the usage of iptables-nft
- legacy: force the usage of iptables-legacy

The new file looks like this:
```yaml
kind: Cluster
networking:
  serviceSubnet: "${SERVICE_CIDR}"
  podSubnet: "${POD_CIDR}"
  disableDefaultCNI: ${DISABLEDEFAULTCNI}
  iptablesMode: "nf_tables"
```

# Steps

- [x] Implement the feature in docker
- [ ] Implement the feature in podman
- [ ] Write unit tests
- [ ] Add documentation

# Why it could be necessary

In the last period, cloud providers and Linux distributions are leaving **iptables-legacy** to switch to **nf_tables** (**iptables_nft** or **nft**).
A way to switch from an **iptables implementation** to another will be appreciated by developers who work on Kubernetes networking and use KinD as the development platform.

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -228,6 +228,9 @@ COPY --from=build / /
 # first tell systemd that it is in docker (it will check for the container env)
 # https://systemd.io/CONTAINER_INTERFACE/
 ENV container docker
+# override the default iptables mode.
+# possible values are: legacy, nf_tables, auto
+ENV IPTABLES_MODE=auto 
 # systemd exits on SIGRTMIN+3, not SIGTERM (which re-executes it)
 # https://bugzilla.redhat.com/show_bug.cgi?id=1201657
 STOPSIGNAL SIGRTMIN+3
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -413,15 +413,26 @@ fix_product_uuid() {
 }
 
 select_iptables() {
-  # based on: https://github.com/kubernetes-sigs/iptables-wrappers/blob/97b01f43a8e8db07840fc4b95e833a37c0d36b12/iptables-wrapper-installer.sh
   local mode num_legacy_lines num_nft_lines
-  num_legacy_lines=$( (iptables-legacy-save || true; ip6tables-legacy-save || true) 2>/dev/null | grep -c '^-' || true)
-  num_nft_lines=$( (timeout 5 sh -c "iptables-nft-save; ip6tables-nft-save" || true) 2>/dev/null | grep -c '^-' || true)
-  if [ "${num_legacy_lines}" -ge "${num_nft_lines}" ]; then
-    mode=legacy
-  else
-    mode=nft
-  fi
+  case "${IPTABLES_MODE}" in
+    "legacy")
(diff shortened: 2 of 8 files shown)


aojea:
why you would want to do that?

legacy iptables has been deprecated long time ago ... we should let them go peacefully 😄

BenTheElder:
> A way to switch from an iptables implementation to another will be appreciated by developers who work on Kubernetes networking and use KinD as the development platform.

You should switch between them on the host?

I don't think this belongs in config, this is a detail of the host environment. We need to match what docker is using to inject rules into the node containers, which is dictated by the host. If you switch the host iptables between legacy and nftables future kind clusters should match.

cheina97:
> why you would want to do that?
> 
> legacy iptables has been deprecated long time ago ... we should let them go peacefully 😄

Hi @aojea. I agree with you but unfortunately, some cloud providers are still using it, and the majority of on-premises clusters (in my experience) are still running outdated node images.

aojea:
> Hi @aojea. I agree with you but unfortunately, some cloud providers are still using it, and the majority of on-premises clusters (in my experience) are still running outdated node images.

I do not understand, kind inherits and uses the iptables version that is used by the VM or node where it is running, why do you want to force it in the image?

cheina97:
I'm sorry, I didn't notice that it changes based on the host configuration and I don't understand how it is possible. I just tried and it works.
 
The entrypoint.sh file runs in a dedicated network namespace and not in the host network namespace, right? So how it can get the host information?

BenTheElder:
> The entrypoint.sh file runs in a dedicated network namespace and not in the host network namespace, right? So how it can get the host information? I'm quite curious.

because docker injects iptables rules into the namespace using the host binaries, and we inspect the state of the two iptables implementations to see which has rules

cheina97:
> > The entrypoint.sh file runs in a dedicated network namespace and not in the host network namespace, right? So how it can get the host information? I'm quite curious.
> 
> because docker injects iptables rules into the namespace using the host binaries, and we inspect the state of the two iptables implementations to see which has rules

It makes sense. Sorry for the inconvenience and thanks for your time.

Do you think It could be useful to make an addition to the documentation or is something useless?

BenTheElder:
> Do you think It could be useful to make an addition to the documentation or is something useless?

This is a pretty deep implementation detail and not specific to KIND, Kubernetes / kube-proxy also auto-selects iptables implementation details on the ~same basis.

There are pointers in the entrypoint script code comments, e.g. https://github.com/kubernetes-sigs/kind/pull/3352/files#diff-3c55751d83af635109cece495ee2ff38206764a8b95f4cb8f11fc08a5c0ea8dcR425

Eventually this won't be relevant as legacy iptables goes away.

I don't think we want to maintain separate docs for this, the iptables-wrappers readme is pretty detailed.

cheina97:
Ok thanks, I'm going to close the PR

PR #3376, opened 2023-09-30
Add installation instructions for `Scoop` and `Winget`

ste93cry:
Since `kind` is now installable via both [`Scoop`](https://scoop.sh/) and [`Winget`](https://learn.microsoft.com/it-it/windows/package-manager/winget/), the [`Quick Start`](https://kind.sigs.k8s.io/docs/user/quick-start/) guide should mention the installation instructions for those package managers.

### DIFF
--- a/site/content/docs/user/quick-start.md
+++ b/site/content/docs/user/quick-start.md
@@ -120,6 +120,17 @@ On Windows via Chocolatey (https://chocolatey.org/packages/kind)
 choco install kind
 {{< /codeFromInline >}}
 
+On Windows via Scoop (https://scoop.sh/#/apps?q=kind&id=faec311bb7c6b4a174169c8c02358c74a78a10c2)
+{{< codeFromInline lang="powershell" >}}
+scoop bucket add main
+scoop install main/kind
+{{< /codeFromInline >}}
+
+On Windows via Winget (https://github.com/microsoft/winget-pkgs/tree/master/manifests/k/Kubernetes/kind)
+{{< codeFromInline lang="powershell" >}}
+winget install Kubernetes.kind
+{{< /codeFromInline >}}
+
 ## Creating a Cluster
 
 Creating a Kubernetes cluster is as simple as `kind create cluster`.


BenTheElder:
blocked on https://github.com/kubernetes-sigs/kind/pull/3379
/hold

ste93cry:
I understand that you may not want to maintain the `Winget` manifest, but I still think it's worth documenting at least the ability to install the package as is possible with other package managers. Tools like `kubectl` and `minikube` are already available in `Winget`, and now that `kind` is too, it doesn't sound too bad to let people know about it. I don't know who maintains the other package manager manifests, but if they are community-driven I wonder why `Winget` should receive a different treatment just because it arrived late at the party 🤔

stmcginnis:
> I understand the reasons you don't want to maintain the manifest, but this has nothing to do with mentioning in the documentation that the tool can be installed with these package managers, right?

I see what you're saying, and I guess I agree. We don't maintain the brew or chocolatey packages, but we do have them listed in the documentation as possible ways to install kind. It almost feels like we should note a caveat there that the project has involvement in those packages and cannot vouch for their quality, but we've gone this long without it, so I don't see why these should be gated on doing something like that. If we even should.

So sorry for the long delay, but I do think this could be useful for someone to have in the docs.

/lgtm

BenTheElder:
> We don't maintain the brew or chocolatey packages, but we do have them listed in the documentation as possible ways to install kind.

 I'm in touch with the brew maintainers (heck somewhere there's a photo of me sitting with christoph at KubeCon while we developed the package on his laptop) and keep an eye on this one. 
 
 It's true that the chocolatey package has this issue.

We should split these into vetted and use at your own risk.

BenTheElder:
Let's add these, but follow-up by splitting first and third party and clarifying (more detailed proposal at https://github.com/kubernetes-sigs/kind/issues/3644)

Sorry for the delay. There's a LOT going on and there hasn't been much user demand for these versus the concerns and ongoing bug triage etc....

/lgtm
/approve

https://github.com/kubernetes-sigs/kind/issues/3644 to follow up on all third party install options.

BenTheElder:
oops, I meant to merge this already but missed the hold-lift, we'll have to rebase it after #3660 merged, I meant for that PR to rebase this work instead of the other way around

https://kind.sigs.k8s.io/docs/user/quick-start/#installing-with-a-package-manager

ste93cry:
Rebased 👍

BenTheElder (commented):
/lgtm
/approve
/hold cancel

thanks!

PR #3379, opened 2023-10-07
Add GitHub Workflow to publish to Winget

### LINKED ISSUE #2187: Add kind to Windows Package Manager repo
Add manifest to Windows Package Manager repo: https://github.com/microsoft/winget-pkgs

**Why is this needed**:
Would be nice to install kind from `winget`.

ste93cry:
Fixes #2187 by automating the publishing of the releases on [`Winget`](https://learn.microsoft.com/it-it/windows/package-manager/winget/). I decided to use `Komac` because the `YamlCreate` script requires cloning the entire [`winget-pkgs`](https://github.com/Trenly/winget-pkgs) repo to work, while `WinGetCreate` does not seems to work due to a failure in extracting the required metadata from the executable.

For the workflow to work, the `WINGET_TOKEN` secret should be added to the repository by the maintainers. The tool opens a PR in the upstream repo, so it has to be able to fork it.

### DIFF
--- a/.github/workflows/winget.yml
+++ b/.github/workflows/winget.yml
@@ -1,0 +1,26 @@
+name: Publish to Winget
+
+on:
+  release:
+    types: [released]
+
+permissions: {}
+
+jobs:
+  publish:
+    name: Publish
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Komac
+        run: |
+          curl -fSsL -o komac.jar https://github.com/russellbanks/Komac/releases/download/v1.11.0/Komac-1.11.0-all.jar
+          echo '1c26bb7bee6228ad095fe67d10ed254363099b3395f809bd6d95a3f72b2740fc komac.jar' | sha256sum --check --status -
+      
+      - name: Update and submit manifest
+        run: |
+          $JAVA_HOME_17_X64/bin/java -jar komac.jar update \
+            --id Kubernetes.kind \
+            --version $(echo '${{ github.event.release.name }}' | sed 's/^v//') \
+            --urls https://github.com/kubernetes-sigs/kind/releases/download/${{ github.event.release.name }}/kind-windows-amd64
+            --token ${{ secrets.WINGET_TOKEN }}
+            --submit


BenTheElder:
Thanks for contributing.

I don't know that we have the bandwidth to maintain winget, windows is community support currently and at minimum it seems we need to keep this komac tool up to date with any and all bugfixes/security patches and the issue has had relatively little activity/upvotes ...

cc @aojea

ste93cry:
Since the tool is automated, I don't expect it to need much maintenance: it does all the heavy work of cloning the `winget-pkgs` repository, updating its files, committing them and finally opening the pull request on the behalf of the owner of the API token. As `kind` itself is not frequently released, I think that it is reasonable to expect that it will have little to no impact on the life of the maintainers of this project.

> it seems we need to keep this komac tool up to date with any and all bugfixes/security patches

We could automate the downloading of the latest version of the tool rather than hardcoding the version in the workflow, and that would solve the point about the patches. The only problem I see is that `Komac` does not provide the `*.sha256sum` assets for each release like `kind` does, but maybe the developer would be open to adding them.

BenTheElder:
> We could automate the downloading of the latest version of the tool rather than hardcoding the version in the workflow, and that would solve the point about the patches. The only problem I see is that Komac does not provide the *.sha256sum assets for each release like kind does, but maybe the developer would be open to adding them.

It seems unlikely that there wouldn't be any breaking changes to running this tool?

ste93cry:
At the moment, the author is rewriting the tool in Rust, so it's likely that substantial changes will be made in the future. However, since he wants to give the new tool a different name, it seems unlikely that he will make any substantial changes to the current one. The author suggests using the [`vedantmgoyal2009/winget-releaser`](https://github.com/vedantmgoyal2009/winget-releaser) action to run the tool in the CI: the benefit would be that we could then use @dependabot to keep the action up-to-date, without worring about which tool is used behind. What do you think about this idea?

BenTheElder:
I really don't think we have the bandwidth to support another option, we are able to have homebrew without this and `go install` or downloadable binaries otherwise. cc @aojea

ste93cry:
Since `Winget` is installed by default since `Windows 11`, I find worth supporting it to install a tool the same as you would do on MacOS with `homebrew`. Moreover, downloading the whole `GoLang` installer just to be able to use `go install` seems a bit too much.

>I really don't think we have the bandwidth to support another option

I understand, but what would be the burden added to the maintenance of this package if the GitHub action does everything on its own for each release?

aojea:
everytime that we add a new tool we increase the surface of the projects and we receive new issues, 

I understand the sentiment of people trying to help adding documentation, but that is at a cost of the maintainers time, 

> I understand, but what would be the burden added to the maintenance of this package if the GitHub action does everything on its own for each release?

jobs break, url changes, projects get abandoned or rewritten

> At the moment, the author is rewriting the tool in Rust,

I truly appreciate your intent but winget will not be supported in the short term, we can always revisit in the future

/hold

Issue #3269, opened 2023-06-14
Copy button not working


Sajiyah-Salat:
**What you expected to happen**:
copy the code 

**How to reproduce it (as _minimally_ and precisely as possible)**:
On [this](https://kind.sigs.k8s.io/docs/user/quick-start/) page
Try to copy code with the use of copy symbol

stmcginnis:
This is working fine for me. Do you have more details such as OS and browser being used?

I've had copy/paste oddness in the past that has usually been cleared up by rebooting.

A lot of this is based on common documentation libraries, so even though it's the kind docs, the kind repo might not be the right place to address this (if there actually is a change needed).

AryanSharma9917:
Hi @Sajiyah-Salat , I would like to contribute to this issue. Could you please assign this issue to me?

AryanSharma9917:
\assign Aryansharma9917

BenTheElder:
There is a lot of variety on the web regarding this sort of thing, I'm not convinced that we should change this, lots of sites just insert it into your clipboard without any popups etc.

We're going to introduce new layout issues if we make the copy button itself dynamic width.

You can confirm that the content is in your clipboard by pasting it, and if not, click it again?

If the copy button doesn't work *at all* then that's a bug we should fix.

PR #3381, opened 2023-10-10
Updated Copy button

AryanSharma9917:
This PR fixes : #3269

### DIFF
--- a/site/layouts/shortcodes/codeFromFile.html
+++ b/site/layouts/shortcodes/codeFromFile.html
@@ -2,25 +2,41 @@
 {{ $lang := "" }}
 {{ $suffix := findRE "(\\.[^.]+)$" $file 1 }}
 {{ with  $suffix }}
-{{ $lang = (index . 0 | strings.TrimPrefix ".") }}
+  {{ $lang = (index . 0 | strings.TrimPrefix ".") }}
 {{ end }}
 {{ with .Get "lang" }}{{ $lang = . }}{{ end }}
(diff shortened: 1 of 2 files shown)


site/layouts/shortcodes/codeFromFile.html
```diff
@@ -2,25 +2,41 @@
...
     </tr>
   </tbody>
-  <textarea class="hidden-copy-text" id="code-{{ $virtualPath }}-hidden-copy-text">{{ ($file | readFile) }}</textarea>
+  <textarea class="hidden-copy-text" id="code-{{ $virtualPath }}-hidden-copy-text">{{ $file | readFile }}</textarea>
 </table>
+
+<script>
+  function copyText(elementId, successElementId) {
+    const textarea = document.getElementById(elementId);
+    textarea.select();
```
nekomeowww, line 32: `textarea` is possibly `undefined` if the HTML has been modified by either extensions or generated incorrectly. Should we at least determine it is safe to be accessed before using it?

```suggestion
    const textarea = document.getElementById(elementId);
    if (!textarea) {
      // perhaps some console.warn handling?
      return
    }
    textarea.select();
```

BenTheElder:
I'm not sure this is the right way to do this and in general we prefer to discuss approach before reviewing PRs to save everyone time (see our contributing guide)

https://github.com/kubernetes-sigs/kind/issues/3269#issuecomment-1942180032

I'm sorry I have not had the bandwidth for this myself, I've been primarily following up on bug fixes.

PR #3392, opened 2023-10-24
Be explicit about the source of the golang images.

adelton:


neolit123 (commented):
/lgtm

are these the only instances of this?

adelton:
These are the only two occurrences I hit when building with `podman` which is picky about having fully-qualified image names.

BenTheElder (commented):
Doing this can break people's proxies/image caching, which is why podman shouldn't hijack the default registry location :/

images/kindnetd/Dockerfile
```diff
@@ -14,7 +14,7 @@
 
 # first stage build kindnetd binary
 # NOTE: the actual go version will be overridden
-FROM --platform=$BUILDPLATFORM golang:latest
+FROM --platform=$BUILDPLATFORM docker.io/golang:latest
```
BenTheElder, line 17: Shouldn't this be:
```suggestion
FROM --platform=$BUILDPLATFORM docker.io/library/golang:latest
```

adelton:
> Doing this can break people's proxies/image caching, which is why podman shouldn't hijack the default registry location :/

I'm not sure it hijacks it -- it just prefers people to be explicit about what they mean, in the world where there are multiple registries.

images/kindnetd/Dockerfile
adelton: Technically perhaps yes. But at https://hub.docker.com/r/library/golang (which redirects to https://hub.docker.com/_/golang) the command shown is `docker pull golang`, not `docker pull library/golang`. I will change it to have the `/library` segment in the path, hoping that Docker won't change the setup of the registry.

BenTheElder: I mean the canonical location is `golang` but the fully qualified location is `docker.io/library/golang:latest`

BenTheElder:
> I'm not sure it hijacks it -- it just prefers people to be explicit about what they mean, in the world where there are multiple registries.

This defaulting has well defined meaning that has existed long before podman, in Kubernetes kubelet will convert `golang` => `docker.io/library/golang`.

Any image with a host in it goes to that particular registry, and any other image goes to the default ...

This sort of incompatibility is why we only support building with buildx specifically just as we require gnu make and not ninja.

BenTheElder (approved):

PR #3393, opened 2023-10-24
Avoid duplicating the `/kind/bin/*.sh` files in `/kind`.

adelton:


images/base/Dockerfile
```diff
@@ -25,11 +25,12 @@ FROM $BASE_IMAGE as base
...
 # copy in static files
 # all scripts are 0755 (rwx r-x r-x)
+# all non-scripts are 0644 (rw- r-- r--)
 COPY --chmod=0755 files/usr/local/bin/* /usr/local/bin/
+
+COPY --chmod=0644 files/kind/ /kind/
 COPY --chmod=0755 files/kind/bin/* /kind/bin/
 
-# all non-scripts are 0644 (rw- r-- r--)
-COPY --chmod=0644 files/kind/* /kind/
```
neolit123, line 32: can you explain what the bug is?
would it help if the COPY calls are more explicit about what is being copied - e.g. `COPY .... *.sh ....`

- the comment `# all non-scripts are 0644 (rw- r-- r--)` should be at the bottom where it was before
- `COPY --chmod=0644 ...`  calls should be under the comment, ideally.

adelton: The current content copies the `files/kind/*.sh` to `/kind/bin/ and then to `/kind/`:
```
$ podman run --rm -ti --entrypoint find docker.io/kindest/node:v1.27.3 /kind
/kind
/kind/README.txt
/kind/bin
/kind/bin/create-kubelet-cgroup-v2.sh
/kind/bin/mount-product-files.sh
/kind/bin/undo-mount-hacks.sh
/kind/create-kubelet-cgroup-v2.sh
/kind/mount-product-files.sh
/kind/undo-mount-hacks.sh
/kind/manifests
/kind/manifests/default-cni.yaml
/kind/manifests/default-storage.yaml
/kind/version
```

The problem with more specific `COPY` parameters is, you cannot say "everything but `*.sh`". And while the `COPY` to `/kind` could only list `README.txt` since that is the only non-`.sh` file there, it might make things less automatic should people add more content there in the future, or in their customizations.

I moved the comment about non-scripts to the top, specifically because I did not find an obvious way to have the order the way it was -- scripts first, non-scripts (and really just non-scripts) later.

neolit123: > The problem with more specific COPY parameters is, you cannot say "everything but *.sh". And while the COPY to /kind could only list README.txt since that is the only non-.sh file there, it might make things less automatic should people add more content there in the future, or in their customizations.

isn't /kind/version also a file?

> I moved the comment about non-scripts to the top, specifically because I did not find an obvious way to have the order the way it was -- scripts first, non-scripts (and really just non-scripts) later.

how about COPY-ing everything to preserve the structure and then calling `RUN chmod ...` separately where needed?
`COPY --chmod=....` is great, but i think it's causing the problem here, no?

adelton: > isn't /kind/version also a file?

Ah, right.

> how about COPY-ing everything to preserve the structure and then calling `RUN chmod ...` separately where needed? `COPY --chmod=....` is great, but i think it's causing the problem here, no?

That would work too.

neolit123: > Or we could chmod the files in the git repository -- that would make them executable in the container image as well.

my preference would be to chmod in the dockerfile.
but the others might give you a green light for the chmod of repo files.

BenTheElder: the files are already chmod in the repo, that doesn't work reliably because of umask 🙃 

this approach is fine, thanks!

BenTheElder (commented):
Thank you!

FYI: not sure what led you to discover this, but we don't support depending on the specifics of the node images and they're subject to change 😅 

/lgtm
/approve

PR #3395, opened 2023-10-25
add some rootless docs notes

kir4h:
Adding a couple of notes to the rootless setup on a couple of items worth mentioning:

* creation of the folder where the service unit file is created; this is debatable (let me know if I should remove it) as it might seem obvious, but including it to highlight that it's normal that such folder doesn't exist beforehand
* restart of Docker service if it was already running, otherwise changes are not reflected back


site/content/docs/user/rootless.md
```diff
@@ -33,6 +33,15 @@ Also, depending on the host configuration, the following steps might be needed:
   Beware that changing this configuration may affect system
   performance.)
 
+  Please note that:
+
+  - `/etc/systemd/system/user@.service.d/` folder needs to be created if not already present on your host
```
AkihiroSuda, line 38: s/folder/directory/

kir4h: updated

AkihiroSuda: Could you squash the commits?

BenTheElder (commented):
/lgtm
/approve

thanks!

Issue #2317, opened 2021-06-19
containerd / nerdctl provider support


xiaofan-linux:
kind provider already supports podman and docker. What supports containerd?

BenTheElder:
Nothing. See some discussion here: https://github.com/kubernetes-sigs/kind/issues/1369

What's your use case?

Also to be clear these are used to create the "nodes". KIND runs the Kubernetes pods on containerd _inside the nodes_ always since around v0.2.0 off the top of my head.

xiaofan-linux:
> Nothing. See some discussion here: #1369
> 
> What's your use case?
> 
> Also to be clear these are used to create the "nodes". KIND runs the Kubernetes pods on containerd _inside the nodes_ always since around v0.2.0 off the top of my head.

If containerd is installed on a virtual machine, can kind call containerd to create a kubernetes cluster?  Not calling docker or podman

xiaofan-linux:
> Nothing. See some discussion here: #1369
> 
> What's your use case?
> 
> Also to be clear these are used to create the "nodes". KIND runs the Kubernetes pods on containerd _inside the nodes_ always since around v0.2.0 off the top of my head.

If containerd is installed on a virtual machine, can kind call containerd to create a kubernetes cluster?  Not calling docker or podman

BenTheElder:
Currently: No. Why wouldn't you be able to install docker in the VM?

xiaofan-linux:
> Currently: No. Why wouldn't you be able to install docker or podman in the VM?

docker and podman can be installed. Would you like to ask if you consider adding containerd support later?

BenTheElder:
I haven't heard a compelling reason to maintain this yet. We've found the podman support very time consuming to maintain and the differences between the tools limiting. I think containerd makes sense for running production workloads but for development and test I would still easily recommend Docker instead (or podman if you have reasons against the docker/containerd ecosystem).

It seems unlikely that developer machines and CI environments will switch to not support docker and that's where we target kind. Podman is available for the redhat ecosystem / as a local docker alternative.

I don't yet see any argument other than "just because" and it's already gotten painful trying to get full CI coverage and maintenance for the two existing options, so I don't think is yet more compelling than the expected cost.

AkihiroSuda:
Supporting containerd via nerdctl (https://github.com/containerd/nerdctl) would be quite straight forward.

afbjorklund:
[...]

PR #3396, opened 2023-10-26
[WIP]Add ipv6 error message check to support the nerdctl

yankay:
**Do not merge**

HI dears.

I'm one of the maintainers of nerdctl, and I'm working on the nerdctl to support the kind recently.

* https://github.com/containerd/nerdctl/issues/349
* https://github.com/kubernetes-sigs/kind/issues/2317

The cluster has been successfully created by kind with nerdctl:
```
root@kay201:~/oss/kind# ln -s /usr/local/bin/nerdctl /usr/local/bin/docker ## Using the nerdctl instead of docker
root@kay201:~/oss/kind# export KUBE_VERSION="1.28.0"
root@kay201:~/oss/kind#  ./bin/kind create cluster --config /root/kind.yml --image release-ci.daocloud.io/docker.m.daocloud.io/kindest/node:v${KUBE_VERSION}

Creating cluster "kind" ...
 ✓ Ensuring node image (release-ci.daocloud.io/docker.m.daocloud.io/kindest/node:v1.28.0) 🖼
 ✓ Preparing nodes 📦 📦 📦 📦
 ✓ Writing configuration 📜
 ✓ Starting control-plane 🕹️
 ✓ Installing CNI 🔌
 ✓ Installing StorageClass 💾
 ✓ Joining worker nodes 🚜

root@kay201:~/oss/kind# ./kubectl get nodes
NAME                 STATUS   ROLES           AGE   VERSION
kind-control-plane   Ready    control-plane   76s   v1.28.0
kind-worker          Ready    <none>          51s   v1.28.0
kind-worker2         Ready    <none>          53s   v1.28.0
kind-worker3         Ready    <none>          52s   v1.28.0

```

The issues of nerdctl has been almost fixed, only 2 PR is pending:
- https://github.com/containerd/nerdctl/pull/2594
- https://github.com/containerd/nerdctl/pull/2592

And there needs a change of kind.

Because the nerdctl do not support the ipv6 network now, so it needs kind to skip the ipv6 config.
If the kind can read the error message of kind, and skip it , the cluster can be create successully.

Thanks :-)

### DIFF
--- a/pkg/cluster/internal/providers/docker/network.go
+++ b/pkg/cluster/internal/providers/docker/network.go
@@ -77,7 +77,7 @@ func ensureNetwork(name string) error {
 	// https://github.com/kubernetes-sigs/kind/issues/1544
 	// Otherwise if it's not a pool overlap error, fail
 	// If it is, make more attempts below
-	if isIPv6UnavailableError(err) {
+	if isIPv6UnavailableError(err) || isIPv6UnsupportError(err) {
 		// only one attempt, IPAM is automatic in ipv4 only
 		return createNetworkNoDuplicates(name, "", mtu)
 	}
@@ -262,6 +262,12 @@ func checkIfNetworkExists(name string) (bool, error) {
 func isIPv6UnavailableError(err error) bool {
 	rerr := exec.RunErrorForError(err)
 	return rerr != nil && strings.HasPrefix(string(rerr.Output), "Error response from daemon: Cannot read IPv6 setup for bridge")
+}
+
+// The nerdctl do not support the ipv6 network, https://github.com/containerd/nerdctl/issues/1547
+func isIPv6UnsupportError(err error) bool {
+	rerr := exec.RunErrorForError(err)
+	return rerr != nil && strings.Contains(string(rerr.Output), "unknown flag: --ipv6")
 }
 
 func isPoolOverlapError(err error) bool {


aojea:
the problem I see is that once nerdctl implement the --ipv6 flag we'll have to remove the error, right?

what is blocking `nerdctl` to implement the ipv6 flag? is there any link or place where I can look more?

BenTheElder:
> the problem I see is that once nerdctl implement the --ipv6 flag we'll have to remove the error, right

I think it depends on how long it takes to ship IPv6, right now released nerdctl won't work even with this patch AIUI but there might be intermediate versions worth supporting and this path could be safely left in place.

I do wonder if we should fork to a nerdctl provider though, which would also let us avoid requiring the user to put a fake docker symlink in their path. We've had a bad experience with this with podman in the past versus detecting that it's actually podman and having dedicated provider / code paths.

aojea:
> I do wonder if we should fork to a nerdctl provider though, which would also let us avoid requiring the user to put a fake docker symlink in their path. We've had a bad experience with this with podman in the past versus detecting that it's actually podman and having dedicated provider / code paths.

@AkihiroSuda has some list somewhere with the missing things, his opinion will be interesting

AkihiroSuda:
> what is blocking `nerdctl` to implement the ipv6 flag? is there any link or place where I can look more?

The lack of the CI for IPv6 is considered to be a blocker for this PR:
- https://github.com/containerd/nerdctl/pull/1558

I think this is just a soft blocker, as Docker doesn't have IPv6 CI either (IIUC).

AkihiroSuda:
> I do wonder if we should fork to a nerdctl provider though, which would also let us avoid requiring the user to put a fake docker symlink in their path. 

SGTM, but I think the implementation should try to avoid the amount of the code clones (unlike `podman` driver).

aojea:
> The lack of the CI for IPv6 is considered to be a blocker for this PR:

we run ipv6 in github actions, what is exactly nlockling ... I'll try to take a look later

BenTheElder:
> SGTM, but I think the implementation should try to avoid the amount of the code clones (unlike podman driver).

FWIW It's not too much code (~1200 loc for the entire "provider") and when we review changes to the existing drivers we ask that they be copied between them when PRs aren't obviously specific to that driver.

We can start with a copy paste and just swap `docker` => `nerdctl`, if enough code stays common and/or they get larger we can refactor out more common code now that we have N=3 examples to derive APIs from.

```
$ echo $(pwd) && cloc .
/usr/local/google/home/bentheelder/go/src/sigs.k8s.io/kind/pkg/cluster/internal/providers/docker
      10 text files.
       9 unique files.                              
       1 file ignored.

github.com/AlDanial/cloc v 1.96  T=0.01 s (648.8 files/s, 124063.7 lines/s)
-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
Go                               9            170            354           1197
-------------------------------------------------------------------------------
SUM:                             9            170            354           1197
-------------------------------------------------------------------------------
```

I don't think IPv6 support should be a blocker, kind already works on hosts without IPv6 and will need to continue to for the forseeable future. IPv6 is also somewhat broken in docker, kind.

I think we'll also want to get some minimal CI up, at least testing one configuration of nerdctl

yankay:
Thanks @aojea @BenTheElder @AkihiroSuda for the PR review

The ipv6 has been supported by the `nerdctl` by https://github.com/containerd/nerdctl/pull/1558
The kind cluster can be created successfully  without the change :-)

Thank you all very much. :-)

-------

And what's the next step to do to make kind support nerdctl ? for example:

- [docs] add experimetal support for nerdctl
- e2e test (minimal CI up)
- find bug and fix, eg: 
  -   ./bin/kind get clusters error , https://github.com/containerd/nerdctl/issues/2598

Issue #2694, opened 2022-03-25
local registry script with rootless podman Error: "slirp4netns" is not supported: invalid network mode


mondo192:
**What happened**: 
slirp4netns networking error 

**What you expected to happen**: 
no error

**How to reproduce it (as minimally and precisely as possible)**:
1. Install kind with go
2. KIND_EXPERIMENTAL_PROVIDER=podman kind create cluster
3. Copy the script from https://kind.sigs.k8s.io/docs/user/local-registry/ and replace docker commands with podman
4. bash kind-with-registry.sh

**Anything else we need to know?**: 
Using rootless podman 

**Environment:**

- kind version: (use `kind version`): kind v0.12.0 go1.18 linux/amd64
- Kubernetes version: (use `kubectl version`):
```
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.5", GitCommit:"c285e781331a3785a7f436042c65c5641ce8a9e9", GitTreeState:"clean", BuildDate:"2022-03-16T15:58:47Z", GoVersion:"go1.17.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.4", GitCommit:"e6c093d87ea4cbb530a7b2ae91e54c0842d8308a", GitTreeState:"clean", BuildDate:"2022-03-06T21:32:53Z", GoVersion:"go1.17.7", Compiler:"gc", Platform:"linux/amd64"}

```
- Docker version: (use `docker info`):
```
host:
  arch: amd64
  buildahVersion: 1.23.1
  cgroupControllers:
  - memory
  - pids
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.1.0-2.fc35.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.1.0, commit: '
  cpus: 12
  distribution:
    distribution: fedora
    variant: workstation
    version: "35"
  eventLogger: journald
  hostname: fedora
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.16.16-200.fc35.x86_64
  linkmode: dynamic
  logDriver: journald
  memFree: 5251850240
[...]

PR #3407, opened 2023-10-31
Connect container run to kind network as bridge

matzew:
I am coming to this b/c I ran into the **Error: "slirp4netns" is not supported: invalid network mode** problem.

Than I saw the comment [here](https://github.com/kubernetes-sigs/kind/issues/2694#issuecomment-1080946864)

Which made me think to update the example script and do the `run` with a `--net`. this seems to also make the `docker network connect` not needed (at least for `podman` this is fine).

**UPDATE:**
After discussing on the PR I updated the script and we `run` the regisrty by default as `bridge` network mode, which is supported by both: `docker` and `podman`

### DIFF
--- a/site/static/examples/kind-with-registry.sh
+++ b/site/static/examples/kind-with-registry.sh
@@ -6,7 +6,7 @@ reg_name='kind-registry'
 reg_port='5001'
 if [ "$(docker inspect -f '{{.State.Running}}' "${reg_name}" 2>/dev/null || true)" != 'true' ]; then
   docker run \
-    -d --restart=always -p "127.0.0.1:${reg_port}:5000" --net=kind --name "${reg_name}" \
+    -d --restart=always -p "127.0.0.1:${reg_port}:5000" --network bridge --name "${reg_name}" \
     registry:2
 fi
 
@@ -43,7 +43,13 @@ for node in $(kind get nodes); do
 EOF
 done
 
-# 4. Document the local registry
+# 4. Connect the registry to the cluster network if not already connected
+# This allows kind to bootstrap the network but ensures they're on the same network
+if [ "$(docker inspect -f='{{json .NetworkSettings.Networks.kind}}' "${reg_name}")" = 'null' ]; then
+  docker network connect "kind" "${reg_name}"
+fi
+
+# 5. Document the local registry
 # https://github.com/kubernetes/enhancements/tree/master/keps/sig-cluster-lifecycle/generic/1755-communicating-a-local-registry
 cat <<EOF | kubectl apply -f -
 apiVersion: v1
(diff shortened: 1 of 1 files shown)


neolit123 (commented):
/lgtm

unclear why the error is happening but the flag change / step 4 removal seems OK to me.

site/static/examples/kind-with-registry.sh
```diff
@@ -6,7 +6,7 @@ reg_name='kind-registry'
 reg_port='5001'
 if [ "$(docker inspect -f '{{.State.Running}}' "${reg_name}" 2>/dev/null || true)" != 'true' ]; then
   docker run \
-    -d --restart=always -p "127.0.0.1:${reg_port}:5000" --name "${reg_name}" \
+    -d --restart=always -p "127.0.0.1:${reg_port}:5000" --net=kind --name "${reg_name}" \
```
BenTheElder, line 9: this doesn't work reliably because the network doesn't exist yet if you've never created any clusters, which is why we use "network connect"

BenTheElder (changes requested):

site/static/examples/kind-with-registry.sh
```diff
@@ -43,13 +43,7 @@ for node in $(kind get nodes); do
 EOF
 done
 
-# 4. Connect the registry to the cluster network if not already connected
-# This allows kind to bootstrap the network but ensures they're on the same network
```
BenTheElder, line 47: (this comment explains why the script works this way)

we can expand this script to support podman but the current approach will break lots of people running this in clean CI environments

BenTheElder:
We usually use `DOCKER=${DOCKER:-docker}` env for overriding docker => podman where the same command will actually work in kubernetes scripts.

Ideally we'd auto detect and match but that's a chicken and egg.

We could check `KIND_EXPERIMENTAL_NODE_PROVIDER` though that's only a partial answer.

site/static/examples/kind-with-registry.sh
matzew: We could, perhaps run this container in `bridge` mode too? Like the `kind-control-plane` container?

I think the podman team is not allowing to "mix" network modes, see https://github.com/containers/podman/issues/13683#issuecomment-1081446705

site/static/examples/kind-with-registry.sh
matzew: with `--network bridge` as stated above, this section is not being removed

site/static/examples/kind-with-registry.sh
BenTheElder: How would we do that? Creating another throwaway network for the registry?

BenTheElder: Specifying `--network bridge` is probably OK, I think docker always has this network but it's not clear if podman does or we'd have to create it.

matzew: I think the `bridge` network will be there. It's well documented 
https://github.com/containers/podman/blob/main/docs/tutorials/basic_networking.md#bridge

matzew: I updated the PR to this

BenTheElder (commented):
/lgtm
/approve

thanks!

PR #3418, opened 2023-11-11
[WIP] Fix --wait flag to accept numeric arguments

radeksm:
**Problem:**
When user attempts to  use **--wait** option cluster create command fails like below:
```
[radek@fed38s08 ~]$ kind create cluster --name c01 --wait 10
ERROR: invalid argument "10" for "--wait" flag: time: missing unit in duration "10"
```

The problems seems to be that Cobra which parses the arguments does not understand _time.Duration_ type.

### DIFF
--- a/pkg/cmd/kind/create/cluster/createcluster.go
+++ b/pkg/cmd/kind/create/cluster/createcluster.go
@@ -37,7 +37,7 @@ type flagpole struct {
 	Config     string
 	ImageName  string
 	Retain     bool
-	Wait       time.Duration
+	Wait       int
 	Kubeconfig string
 }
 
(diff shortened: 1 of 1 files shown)


pkg/cmd/kind/create/cluster/createcluster.go
```diff
@@ -79,11 +79,12 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
...
 		"retain nodes for debugging when cluster creation fails",
 	)
-	cmd.Flags().DurationVar(
+
+	cmd.Flags().IntVar(
 		&flags.Wait,
 		"wait",
-		time.Duration(0),
-		"wait for control plane node to be ready (default 0s)",
+		0,
```
neolit123, line 86: `10` is not a valid input to `time.ParseDuration()` which is used by pflag and cobra.
```
d, err := time.ParseDuration("10")
fmt.Println(err, d) // output: `time: missing unit in duration "10" 0s`
```

also, this is a breaking change on the kind CLI, as users who used to pass `10s` for `--wait` (a correct `time.Duration` input), will no longer be able to.

radeksm: Ok, I see, fully agree.  So  this is not a code bug but more a lack of clear documentation, how  **--wait** option format should look like.

radeksm:
Not a bug.

PR #3419, opened 2023-11-13
Be more clear on the expected format of --wait option

radeksm:
This is just a very minor follow up on this pull request [3418](https://github.com/kubernetes-sigs/kind/pull/3418).

`kind create cluster --help` does not explain what is expected format of _**--wait**_ option. This is documented well in site/content/docs/user/quick-start.md but this simple change to command line help is going to save some time and confusion.


neolit123 (commented):
/lgtm

BenTheElder:
FWIW, this problem exists in kubectl as well, e.g. `kubectl wait --help` `--timeout` docs, these are go https://pkg.go.dev/time#Duration if we want to be more specific.

This change seems fine, but I prefer to align with `kubectl` where possible on the CLI flags and behaviors for consistency, maybe we need (or have already and can just reference?) a kubernetes.io docs page to cover these duration strings, they're used in other places in the project.

aojea:
1.24 seems to be permanently failing https://prow.k8s.io/pr-history/?org=kubernetes-sigs&repo=kind&pr=3419, keep retesting is not going to solve the problem ;)

radeksm:
> 1.24 seems to be permanently failing https://prow.k8s.io/pr-history/?org=kubernetes-sigs&repo=kind&pr=3419, keep retesting is not going to solve the problem ;)

Yeahhh but it seems to be not related to my change and I thought it's some CI problem with lack of resources. It was the same with pull-kind-e2e-kubernetes-1-25 which eventually passed. 
Sorry but I am not familiar with the CI.

pkg/cmd/kind/create/cluster/createcluster.go
```diff
@@ -83,7 +83,7 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 		&flags.Wait,
 		"wait",
 		time.Duration(0),
-		"wait for control plane node to be ready (default 0s)",
+		"the length of time to wait for control plane node to be ready (default 0s). Non-zero values should contain a corresponding time unit (e.g. 1s, 2m, 3h). A value of zero means don't timeout requests.",
```
BenTheElder, line 86:
```suggestion
		"the length of time to wait for control plane node to be ready (default 0s). Non-zero values should contain a corresponding time unit (e.g. 1s, 2m, 3h). A value of zero means don't wait.",
```
(I see this is 1:1 from kubectl, we are doing something slightly different here so one small tweak)

BenTheElder (commented):
... we (=kind approvers) really should've gotten back to this PR a very long time ago, apologies, I have mostly been chasing bug fixes and debugging when I have time, I'm aiming to make sure we keep up with the PR backlog going forward.

I have one minor correction request, if you're not available after some time given the duration we can apply it in a follow-up so as to retain the attribution for this commit.

BenTheElder:
... So the github "commit suggestion" button broke the CLA :/

BenTheElder:
> The email address for the commit (https://github.com/kubernetes-sigs/kind/commit/56db1281bc0666368896a5aabf24584c5eedb6dd) is not linked to the GitHub account, preventing the EasyCLA check.

... ah, this is because of the existing commit, I think due to @radeksm changing account info since this PR was started and then neglected on our end (sorry!)

pkg/cmd/kind/create/cluster/createcluster.go
```diff
@@ -83,7 +83,7 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 		&flags.Wait,
 		"wait",
 		time.Duration(0),
-		"wait for control plane node to be ready (default 0s)",
+		"the length of time to wait for control plane node to be ready (default 0s). Non-zero values should contain a corresponding time unit (e.g. 1s, 2m, 3h). A value of zero means don't wait.",
```
stmcginnis, line 86: Now that this is two full sentences, it would be nice if it started with capitalization.

stmcginnis (commented):
Sorry this has sat for so long. @radeksm would you be able to update to fix the CLA issues now? If so, we should be able to get this through much faster than it's been.

pkg/cmd/kind/create/cluster/createcluster.go
BenTheElder: @stmcginnis maybe at this point send an updated PR, I feel bad for how this went but we also can't merge CLA failures and you're suggesting changes anyhow (and the PR is single line ...)

stmcginnis:
Redone with #3992

PR #3438, opened 2023-11-29
Kindnetd: ignore nodes with empty PodCIDR.

mneverov:
Nodes don't get podCIDR assigned when controller manager is run with [allocate-node-cidrs=false](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/#options). In this case node routing set up should be skipped.
This patch adds check if a node's `node.Spec.PodCIDR` is empty (because even if the field is optional the type is string, so the default value is ""). `node.Spec.PodCIDRs` is not checked because the default value for the `[]string` will be nil.

### DIFF
--- a/images/kindnetd/cmd/kindnetd/main.go
+++ b/images/kindnetd/cmd/kindnetd/main.go
@@ -236,11 +236,14 @@ func makeNodesReconciler(cniConfig *CNIConfigWriter, hostIP string, ipFamily IPF
 		}
 
 		// This is another node. Add routes to the POD subnets in the other nodes
-		// don't do anything unless there is a PodCIDR
+		// don't do anything unless there is a non-empty PodCIDR
 		var podCIDRs []string
 		if ipFamily == DualStackFamily {
 			podCIDRs = node.Spec.PodCIDRs
 		} else {
+			if node.Spec.PodCIDR == "" {
+				klog.Infof("IPFamily is not dual and node %s has no CIDR, ignoring", node.Name)
+			}
 			podCIDRs = []string{node.Spec.PodCIDR}
 		}
 		if len(podCIDRs) == 0 {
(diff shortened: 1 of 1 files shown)


images/kindnetd/cmd/kindnetd/main.go
```diff
@@ -236,11 +236,14 @@ func makeNodesReconciler(cniConfig *CNIConfigWriter, hostIP string, ipFamily IPF
...
-		// don't do anything unless there is a PodCIDR
+		// don't do anything unless there is a non-empty PodCIDR
 		var podCIDRs []string
 		if ipFamily == DualStackFamily {
 			podCIDRs = node.Spec.PodCIDRs
 		} else {
+			if node.Spec.PodCIDR == "" {
+				klog.Infof("IPFamily is not dual and node %s has no CIDR, ignoring", node.Name)
+			}
 			podCIDRs = []string{node.Spec.PodCIDR}
```
aojea, line 247:
```suggestion
		} else if node.Spec.PodCIDR != "" {
			podCIDRs = []string{node.Spec.PodCIDR}
```

aojea:
/ok-to-test

TODO for myself build the new image and append to the PR
/cc @BenTheElder 

The context is that Max is working in a new sig-network project to enhance the nodeipam controller, so there will be a time that nodes will not get the node.Spec.PodCIDR , that is fine, that field is informational and CNI plugins may use it or not.

Fixing this bug in kindnet will allow this new project to use kind for CI

aojea:
@BenTheElder can you remind me how do we promote the images with the  new cloudbuild model?

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-kind-push-kindnetd-image/1732408886066941952

Issue #3441, opened 2023-11-30
leverage containerd image pinning


BenTheElder:
So we can potentially re-enable image GC. #735 

@ruiwen-zhao added this to containerd 1.6+ (not sure specific version yet) and we should leverage it for the pre-loaded images in kind.

roman-kiselenko:
Hey! I want to work on this.

BenTheElder:
hey!

So I started looking into this and we'll need to label the images with `--label io.cri-containerd.pinned=pinned --label io.cri-containerd.image=managed`. 

I know this is available for `ctr images pull`, we need to verify if it is for import and if not we'll need to PR that upstream first.

Also, we need to be on a sufficiently new conatinerd release first. I didn't track down exactly which version yet, that would be helpful.

We then need to decide which images this applies to.
I think to start we should do all images in `kindest/node` but NOT later added `kind load ...` images.

Turning on GC afterwards remains an open question, there's some bad trade-offs there:
1. What about user side-loaded images? Maybe we should? This could be surprising behavior.
2. image GC knobs are based on % of disk, but the host disk is not being used exclusively for this cluster and on a desktop maybe 90% full disk doesn't mean we should GC images ..... turning this on is definitely a breaking change and it's not clear what the threshold is.

Still, we can at least go ahead and clearly label kind's core bundled images as "don't GC this".

BenTheElder:
> Yep, the labeling is only available for pull and there is no --label option for import. 😞

We should work with containerd to add this feature.

roman-kiselenko:
> > Yep, the labeling is only available for pull and there is no --label option for import. 😞
> 
> We should work with containerd to add this feature.

I'll give it a try and propose changes to the containerd. 🙏

roman-kiselenko:
I've opened a PR with related changes.

roman-kiselenko:
PR with `--label` flag [merged](https://github.com/containerd/containerd/pull/9476). 

Flag will be available in the next major release.

PR #3444, opened 2023-12-04
Node-image: tag the core bundled images that we want to preserve with a pinned label

roman-kiselenko:
Ref #3441 

Tag the core bundled images with `pinned` label.

```shell
crictl inspecti registry.k8s.io/coredns/coredns:v1.11.1
{
  "status": {
    "id": "sha256:2437cf762177702dec2dfe99a09c37427a15af6d9a57c456b65352667c223d93",
    "repoTags": [
      "registry.k8s.io/coredns/coredns:v1.11.1"
    ],
    "repoDigests": [],
    "size": "16482581",
    "uid": null,
    "username": "nonroot",
    "spec": null,
    "pinned": true
  },
```

Additional context

- kubernetes/kubernetes#103299
- https://github.com/containerd/containerd/pull/7944

### DIFF
--- a/pkg/build/nodeimage/imageimporter.go
+++ b/pkg/build/nodeimage/imageimporter.go
@@ -55,7 +55,7 @@ func (c *containerdImporter) Pull(image, platform string) error {
 func (c *containerdImporter) LoadCommand() exec.Cmd {
 	return c.containerCmder.Command(
 		// TODO: ideally we do not need this in the future. we have fixed at least one image
-		"ctr", "--namespace=k8s.io", "images", "import", "--all-platforms", "--no-unpack", "--digests", "-",
+		"ctr", "--namespace=k8s.io", "images", "import", "--label=io.cri-containerd.pinned=pinned", "--all-platforms", "--no-unpack", "--digests", "-",
 	)
 }
 


neolit123 (commented):
is the pin feature even working properly at this point? this is a recent bug fix PR in containerd:
https://github.com/containerd/containerd/issues/9328

i don't think we need to pin any image of active installed components - i.e. if we run a static pod, the image is "active", same goes for the deployments and DS-es that kubeadm / kind deploy.

BenTheElder:
> is the pin feature even working properly at this point? this is a recent bug fix PR in containerd:

I know it's rolling out to another distro, even if not AFAICT it's not doing anything actively harmful in the bugs.

We aren't ready to re-enable GC by default yet without further conversation anyhow, but we can go ahead and start labeling these.

> i don't think we need to pin any image of active installed components - i.e. if we run a static pod, the image is "active", same goes for the deployments and DS-es that kubeadm / kind deploy.

It's a layering violation to try to know which ones are currently run as static pods are not, it makes sense to just label all of the images we pre-loaded because they may not be pullable in the target environment and we're expressing intent that they should not be GCed.

pkg/build/nodeimage/imageimporter.go
```diff
@@ -48,17 +50,25 @@ func (c *containerdImporter) End() error {
 
 func (c *containerdImporter) Pull(image, platform string) error {
 	return c.containerCmder.Command(
-		"ctr", "--namespace=k8s.io", "content", "fetch", "--platform="+platform, image,
+		"ctr", "--namespace="+commonNS, "content", "fetch", "--platform="+platform, image,
```
BenTheElder, line 53: `+commonNS` is harder to read / obscures the value and is unrelated to this PR, please revert this part

roman-kiselenko:
Should we update the [containerd version](https://github.com/kubernetes-sigs/kind/blob/main/images/base/Dockerfile#L121) to 1.7.10 in this PR?

BenTheElder (commented):
/lgtm
/approve

thanks!

PR #3449, opened 2023-12-09
Remove "Non-AMD64 Architectures" section from Known Issues

### LINKED ISSUE #3447: Remove "Non-AMD64 Architectures" section from Known Issues
I would suggest removing the "Non-AMD64 Architectures" section from Known Issues. It is not accurate given that you now do have ARM pre-built images.

**Why is this needed**:
Accuracy

mausearce:
Fixes #3447

### DIFF
--- a/site/content/docs/user/known-issues.md
+++ b/site/content/docs/user/known-issues.md
@@ -33,7 +33,7 @@ description: |-
 * [Pod Errors Due to "too many open files"](#pod-errors-due-to-too-many-open-files) (likely [inotify] limits which are not namespaced)
 * [Docker Permission Denied](#docker-permission-denied) (ensure you have permission to use docker)
 * [Windows Containers](#windows-containers) (unsupported / infeasible)
-* [Non AMD64 and ARM64 Architectures](#non-amd64-and-arm64-architectures) (images not pre-built yet)
+* [Unsupported Architectures](#unsupported-architectures) (images not pre-built yet)
 * [Unable to Pull Images](#unable-to-pull-images) (various)
 * [Chrome OS](#chrome-os) (needs KubeletInUserNamespace)
 * [AppArmor](#apparmor) (may break things, consider disabling)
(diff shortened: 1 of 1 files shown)


aojea:
/ok-to-test

you have to remove the reference to the issue from the commit

aojea:
/approve

Thanks

aojea:
/hold

I think we should be more specific, we only publish ARM, but s390x and others more esoteric are not going to be published

site/content/docs/user/known-issues.md
```diff
@@ -33,7 +33,6 @@ description: |-
 * [Pod Errors Due to "too many open files"](#pod-errors-due-to-too-many-open-files) (likely [inotify] limits which are not namespaced)
 * [Docker Permission Denied](#docker-permission-denied) (ensure you have permission to use docker)
 * [Windows Containers](#windows-containers) (unsupported / infeasible)
-* [Non-AMD64 Architectures](#nonamd64-architectures) (images not pre-built yet)
```
aojea, line 36: Non-AMD64 and ARM64

aojea:
thinking more about this we should keep the disclaimer for those esoteric architectures, so they are aware they need to build their own images, just make more clear that is non-amd64 and non-arm64

site/content/docs/user/known-issues.md
stmcginnis: I wonder if "Unsupported Architectures" might be better than explicitly listing out the supported ones. That could keep it flexible in the unlikely event more than AMD64 and ARM64 are ever supported.

aojea: I like that idea

aojea:
/hold cancel
/lgtm
/approve

Thanks

site/content/docs/user/known-issues.md
mausearce: I just pushed a new commit with this suggestion. Let's hope the 3rd time is the charm :-)

BenTheElder:
Thanks!

Issue #3244, opened 2023-05-22
kind cluster on WSL2 does not restart when Docker restarts


inliquid:
**What happened**:

Hey! I have created a cluster on Ubuntu@WSL2, but found that it's not restarted when Docker restarts. 

In order to start the cluster I have to run manually:
```
$ docker start kind-control-plane
```

I have also tried changing `RestartPolicy` to `unless-stopped` and `always` - it didn't work. There are a couple of containers with `unless-stopped` policy, and they all restart as expected, problem is only with `kind-control-plane`.

When I run `docker start kind-control-plane` everything seems fine, so the problem is only with restart, I don't see any failure.

**What you expected to happen**:

Kind cluster being started with Docker restart.

**How to reproduce it (as _minimally_ and precisely as possible)**:

If I'm correct:
1. Install Ubuntu on latest WSL2@Windows 11
2. Install Docker Desktop and configure it to integrate with Ubuntu@WSL2
![изображение](https://github.com/kubernetes-sigs/kind/assets/6797881/9ba8146c-68c0-4649-85cb-a7c2d5a25328)
3. Create kind cluster
4. Restart Docker Desktop
5. Check if kind cluster restarted

**Anything else we need to know?**:

**Environment:**

- kind version: (use `kind version`):
```
$ kind version
kind v0.19.0 go1.20.4 linux/amd64
```
- Runtime info: (use `docker info` or `podman info`):
```
$ docker info
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.10.4
    Path:     /usr/local/lib/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.17.3
    Path:     /usr/local/lib/docker/cli-plugins/docker-compose
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.0
    Path:     /usr/local/lib/docker/cli-plugins/docker-dev
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.19
    Path:     /usr/local/lib/docker/cli-plugins/docker-extension
  init: Creates Docker-related starter files for your project (Docker Inc.)
    Version:  v0.1.0-beta.4
    Path:     /usr/local/lib/docker/cli-plugins/docker-init
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc.)
    Version:  0.6.0
[...]

Issue #2917, opened 2022-09-08
cluster reboots doesn't preserve node ip


lazedo:
**What happened**:
node ip wasn't preserved

**What you expected to happen**:
node ip preserved

**How to reproduce it (as minimally and precisely as possible)**:
start cluster, attach a network to a node, restart container

**Anything else we need to know?**:
entrypoint script assumes `head` for detecting ip which doesn't match the omitted (by `head`) existing
the node ip in 172.18.x.x the network attached is 192.168.x.x. the ip from attached network may be listed first
```bash
  # fixup IPs in manifests ...
  curr_ipv4="$( (head -n1 <(timeout 5 getent ahostsv4 "$(hostname)") | cut -d' ' -f1) || true)"
```
**Environment:**
ubuntu-2204
- kind version: (use `kind version`): 0.15
- Kubernetes version: (use `kubectl version`):
```
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.1", GitCommit:"86ec240af8cbd1b60bcc4c03c20da9b98005b92e", GitTreeState:"clean", BuildDate:"2021-12-16T11:41:01Z", GoVersion:"go1.17.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.3", GitCommit:"aef86a93758dc3cb2c658dd9657ab4ad4afc21cb", GitTreeState:"clean", BuildDate:"2022-08-08T21:15:39Z", GoVersion:"go1.18.3", Compiler:"gc", Platform:"linux/amd64"}
```
- Docker version: (use `docker info`):
```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Docker Buildx (Docker Inc., v0.8.1-docker)
  scan: Docker Scan (Docker Inc., v0.17.0)

Server:
 Containers: 7
  Running: 7
  Paused: 0
  Stopped: 0
 Images: 350
 Server Version: 20.10.14
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
[...]

Issue #2824, opened 2022-07-05
[MacOS-intel] Ports are not reachable after a docker restart


giejay:
**What happened**:
When Im restarting Docker (or the Macbook itself), the ports, which I have configured in the extraPortMappings section, are no longer available to the host. 

**What you expected to happen**:
Ports should remain exposed after a Docker restart.

**How to reproduce it (as minimally and precisely as possible)**:
I have a Cassandra image, including a service which exposes 9042 on NodePort 30000:
```
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: cassandra
  labels:
    app: cassandra
spec:
  serviceName: cassandra
  replicas: 1
  selector:
    matchLabels:
      app: cassandra
  template:
    metadata:
      labels:
        app: cassandra
    spec:
      terminationGracePeriodSeconds: 1800
      containers:
        - name: cassandra
          image: cassandra:3.0.21
          imagePullPolicy: Always
          ports:
            - containerPort: 7000
              name: intra-node
            - containerPort: 7001
              name: tls-intra-node
            - containerPort: 7199
              name: jmx
            - containerPort: 9042
              name: cql
          resources:
            limits:
              cpu: "2000m"
              memory: 6Gi
            requests:
              cpu: "1000m"
              memory: 2Gi
          securityContext:
            capabilities:
              add:
                - IPC_LOCK
          lifecycle:
            preStop:
              exec:
                command:
                  - /bin/sh
                  - -c
                  - nodetool drain
          env:
            - name: MAX_HEAP_SIZE
              value: 512M
            - name: HEAP_NEWSIZE
              value: 100M
            - name: CASSANDRA_SEEDS
              value: "cassandra-0.cassandra.default.svc.cluster.local"
            - name: CASSANDRA_CLUSTER_NAME
              value: "K8Demo"
            - name: CASSANDRA_DC
              value: "DC1-K8Demo"
            - name: CASSANDRA_RACK
              value: "Rack1-K8Demo"
            - name: POD_IP
              valueFrom:
                fieldRef:
                  fieldPath: status.podIP
---
apiVersion: v1
kind: Service
metadata:
  labels:
    app: cassandra
  name: cassandra
spec:
  type: NodePort
  ports:
    - nodePort: 30000
      port: 9042
      protocol: TCP
      targetPort: 9042
  selector:
    app: cassandra
```

That NodePort is forwarded to the host on 9042 again: 

```
[...]

Issue #2383, opened 2021-07-23
DNS not working after reboot


hobyte:
**What happened**:
I created a new kind cluster, then rebooted my computer. After reboot, the dns cannot resolve adresses

**What you expected to happen**:

dns can resolve adresses

**How to reproduce it (as minimally and precisely as possible)**:

- create a new kind cluster
- test dns: it's working
- reboot your machine (dont't stop docker before reboot)
- test dns again: 
```
#APISERVER=https://kubernetes.default.svc
#SERVICEACCOUNT=/var/run/secrets/kubernetes.io/serviceaccount
#NAMESPACE=$(cat ${SERVICEACCOUNT}/namespace)
#TOKEN=$(cat ${SERVICEACCOUNT}/token)
#CACERT=${SERVICEACCOUNT}/ca.crt
#curl --cacert ${CACERT} --header "Authorization: Bearer ${TOKEN}" -X GET ${APISERVER}/api
curl: (6) Could not resolve host: kubernetes.default.svc
```
Taken from https://kubernetes.io/docs/tasks/run-application/access-api-from-pod/#without-using-a-proxy

**Anything else we need to know?**:
- dns pods are running
- dns logs:
```
.:53
[INFO] plugin/reload: Running configuration MD5 = db32ca3650231d74073ff4cf814959a7
CoreDNS-1.8.0
linux/amd64, go1.15.3, 054c9ae
```
dns lookup:
```
#nslookup kubernetes.default
Server:         10.96.0.10
Address:        10.96.0.10#53

Name:   kubernetes.default.svc.cluster.local
Address: 10.96.0.1
#nslookup kubernetes.default.svc
;; connection timed out; no servers could be reached
```
rslov.conf:
```
#cat /etc/resolv.conf 
search default.svc.cluster.local svc.cluster.local cluster.local fritz.box
nameserver 10.96.
```

**Environment:**

- kind version: (use `kind version`): kind v0.11.1 go1.16.4 linux/amd64
- Kubernetes version: (use `kubectl version`): Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.1", GitCommit:"5e58841cce77d4bc13713ad2b91fa0d961e69192", GitTreeState:"clean", BuildDate:"2021-05-12T14:18:45Z", GoVersion:"go1.16.4", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #3458, opened 2023-12-20
command add start cluster

helen-frank:
# Feature

Adds the command start cluster to command, which is used to start the cluster again.

# User story

I am using kind to create multiple clusters on my mobile devices for development. When my mobile devices reboot, I have to manually start each container.
I wrote some docker related scripts to avoid having to start each container manually.
Now I'm thinking that I can use kind to start the clusters I want individually, instead of using docker to write a regular match to the cluster name to start the container for the corresponding cluster.

# Usage example

```bash
# kind start cluster -n k1
Start cluster "k1" ...
Started nodes: ["k1-worker2" "k1-control-plane" "k1-worker"]
```

### DIFF
--- a/pkg/cluster/internal/providers/docker/provider.go
+++ b/pkg/cluster/internal/providers/docker/provider.go
@@ -154,6 +154,25 @@ func (p *provider) DeleteNodes(n []nodes.Node) error {
 	return nil
 }
 
+// StartNodes is part of the providers.Provider interface
+func (p *provider) StartNodes(n []nodes.Node) error {
+	if len(n) == 0 {
+		return nil
+	}
+	const command = "docker"
+	args := make([]string, 0, len(n)+1) // allocate once
+	args = append(args,
+		"start",
+	)
+	for _, node := range n {
+		args = append(args, node.String())
+	}
+	if err := exec.Command(command, args...).Run(); err != nil {
+		return errors.Wrap(err, "failed to start nodes")
+	}
+	return nil
+}
+
 // GetAPIServerEndpoint is part of the providers.Provider interface
 func (p *provider) GetAPIServerEndpoint(cluster string) (string, error) {
 	// locate the node that hosts this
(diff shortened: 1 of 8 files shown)


neolit123 (commented):
hi, new features should ideally be discussed in an issue first before a PR is sent.

with respect to "starting a cluster" it would also apply to cluster creation where it's automatic. if that's true would it make sense to also have a flag for "kind create cluster", that does not start the containers, letting the user to do it async?

stmcginnis:
I believe there are still some issues with restarting clusters (#1689, #2383, #2824, #2917, #3244). We would need to make sure those are addressed before adding an official command to support that operation, otherwise there will be an expectation that restarting is a supported functionality that will work.

It really is best to just delete and recreate your cluster, since kind clusters are meant to be disposable.

helen-frank:
> hi, new features should ideally be discussed in an issue first before a PR is sent.
> 
> with respect to "starting a cluster" it would also apply to cluster creation where it's automatic. if that's true would it make sense to also have a flag for "kind create cluster", that does not start the containers, letting the user to do it async?

ok, I'll take care of it from now on

I agree with you about adding a flag to 'kind create cluster'.

helen-frank:
> I believe there are still some issues with restarting clusters (#1689, #2383, #2824, #2917, #3244). We would need to make sure those are addressed before adding an official command to support that operation, otherwise there will be an expectation that restarting is a supported functionality that will work.
> 
> It really is best to just delete and recreate your cluster, since kind clusters are meant to be disposable.

Thank you for summarizing your questions

My idea is to provide a simple way to start the cluster first, similar to batch boot nodes, to reduce the difficulty of manual operation. For more network or physical machine related issues, I don't agree that they should be solved in start, start should solve the problem of quickly starting the cluster nodes.

Regarding your point about clusters being disposable, I've probably seen that it's more common to reuse clusters except when running bulk tests. This is because you're deploying applications in the cluster to provide services for development, and you don't want to redeploy them.

BenTheElder:
Per our contributing guide (linked in CONTRIBUTING.md, the welcome message, our readme, and our docs site), features like this should be discussed in an issue first and the approach agreed upon.

You can workaround this for now with something like `kind get nodes | xargs docker start` if this implementation is sufficient for your useage.

As-is, support for restarts is limited to a single podman cluster or docker clusters, not guaranteed, and disposable usage is highly encouraged / optimized for.

Issue #3459, opened 2023-12-22
[kubeadm init] sed: couldn't flush stdout: Device or resource busy


sword-jin:
setup env

```
KIND_EXPERIMENTAL_PROVIDER=podman kind create cluster --name test-podman --image=kindest/node:v1.29.0
```

I am trying to install kind cluster by podman, encounter this problem, I found the kubelet service could not run. Checked the log by `

```
░░ The job identifier is 59922 and the job result is done.
Dec 22 09:07:25 test-podman-control-plane systemd[1]: Starting kubelet: The Kubernetes Node Agent...
░░ Subject: A start job for unit kubelet.service has begun execution
░░ Defined-By: systemd
░░ Support: https://www.debian.org/support
░░
░░ A start job for unit kubelet.service has begun execution.
░░
░░ The job identifier is 59922.
Dec 22 09:07:25 test-podman-control-plane sh[6097]: sed: couldn't flush stdout: Device or resource busy
Dec 22 09:07:25 test-podman-control-plane systemd[1]: kubelet.service: Control process exited, code=exited, status=4/NOPERMISSION
░░ Subject: Unit process exited
░░ Defined-By: systemd
░░ Support: https://www.debian.org/support
░░
░░ An ExecStartPre= process belonging to unit kubelet.service has exited.
░░
░░ The process' exit code is 'exited' and its exit status is 4.
Dec 22 09:07:25 test-podman-control-plane systemd[1]: kubelet.service: Failed with result 'exit-code'.
░░ Subject: Unit failed
░░ Defined-By: systemd
░░ Support: https://www.debian.org/support
░░
░░ The unit kubelet.service has entered the 'failed' state with result 'exit-code'.
Dec 22 09:07:25 test-podman-control-plane systemd[1]: Failed to start kubelet: The Kubernetes Node Agent.
░░ Subject: A start job for unit kubelet.service has failed
░░ Defined-By: systemd
░░ Support: https://www.debian.org/support
░░
░░ A start job for unit kubelet.service has finished with a failure.
░░
░░ The job identifier is 59922 and the job result is failed.
```

Then check the script, and run it line by line, 

```
ensure_subtree_control /
sed -e 's/ / +/g' -e 's/^/+/' <"/sys/fs/cgroup///cgroup.controllers" >"/sys/fs/cgroup///cgroup.subtree_control"
```

__And my cgroup.controllers is a empty file__

```
[...]

PR #3462, opened 2023-12-26
Support podman runc runtime

sword-jin:
Actually, I am fixing https://github.com/kubernetes-sigs/kind/issues/3459, I found the cgroups has some problem.

So I force to use runc as the podman's container, then it works.

I think we can provide a env to use runc.

as you know, `--runtime` is a global flag of podman, so we can use a `newPodmanCmd` for all the podman command execution.

### DIFF
--- a/pkg/cluster/internal/providers/podman/command.go
+++ b/pkg/cluster/internal/providers/podman/command.go
@@ -1,0 +1,30 @@
+package podman
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"sigs.k8s.io/kind/pkg/exec"
+)
+
+// newPodmanCmd returns a new exec.Cmd for podman.
+func newPodmanCmd(args ...string) exec.Cmd {
+	args = appendPodmanRuntimeArg(args...)
+	return exec.Command("podman", args...)
+}
+
+// newPodmanCmdWithContext returns a new exec.Cmd for podman with the given context.
+func newPodmanCmdWithContext(ctx context.Context, args ...string) exec.Cmd {
+	args = appendPodmanRuntimeArg(args...)
+	return exec.CommandContext(ctx, "podman", args...)
+}
+
+// appendPodmanRuntimeArg use the KIND_PODMAN_RUNTIME environment variable if set.
+func appendPodmanRuntimeArg(args ...string) []string {
+	runtime := os.Getenv("KIND_PODMAN_RUNTIME")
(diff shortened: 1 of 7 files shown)


pkg/cluster/internal/providers/podman/command.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"sigs.k8s.io/kind/pkg/exec"
+)
+
+// newPodmanCmd returns a new exec.Cmd for podman.
+// It will use the KIND_PODMAN_RUNTIME environment variable if set.
```
neolit123, line 12: this comment line should be above appendPodmanRuntimeArg

neolit123 (commented):
i don't know if there is history about supporting or not supporting different runtimes for podman in kind, but i think the refactor to have a newPodmanCmd and context variant is fine.

the new exposed flag and env var, seem good too.

LGTM, minus one nit.

pkg/cluster/internal/providers/podman/command.go
sword-jin: Yes, thank you

neolit123 (commented):
please keep the commits squashed to one.

LGTM, on my side would let others review.
thanks for the PR.

sword-jin:
@neolit123 is it possible to run e2e locally, I don’t find the manual on testing guide

neolit123:
> @neolit123 is it possible to run e2e locally, I don’t find the manual on testing guide

i don't have tips for that.

the errors in this run are unrelated flakes:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/3462/pull-kind-e2e-kubernetes-1-24/1739572487202541568

you can type `/retest`.

/retest

sword-jin:
I guessed it’s flaky test before so I ran rerun but still failed, is this case so flaky?

aojea:
/hold

AFAIK this can be configured in the host https://serverfault.com/questions/989509/how-can-i-change-the-oci-runtime-in-podman , what should we add this knob to kind?

sword-jin:
> /hold
> 
> AFAIK this can be configured in the host https://serverfault.com/questions/989509/how-can-i-change-the-oci-runtime-in-podman , why should we add this knob to kind?

Oh, really, thank you. I should close this PR.

PR #3469, opened 2024-01-04
Setup OSSF scorecard and Codeql workflows

mmorel-35:
- Setup OSSF Scorecard and CodeQL workflow to improve security vulnerabilities detection
- Pin Docker images versions
- Setup dependabot for the Dockerfiles

Signed-off-by: Matthieu MOREL <matthieu.morel35@gmail.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,6 +1,38 @@
 ---
 version: 2
 updates:
+  - package-ecosystem: "docker"
+    directory: "/images/base"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "area/dependency"
+      - "release-note-none"
+      - "ok-to-test"
+  - package-ecosystem: "docker"
+    directory: "/images/haproxy"
+    schedule:
+      interval: "weekly"
+    labels:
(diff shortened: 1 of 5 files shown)


BenTheElder:
TLDR: Thank you, but no thank you. These are not the approaches we'd prefer to take.

Also, we normally discuss changes like this first before developing them: https://kind.sigs.k8s.io/docs/contributing/getting-started/

> Setup OSSF Scorecard [...]

This is a highly flawed and biased metric and we're not going to endorse it, nor do other Kubernetes projects.

> [..] and CodeQL workflow to improve security vulnerabilities detection

We use golangci-lint linters for vuln scanning.

Vuln scanners won't work otherwise for our built-from source binaries, and the rest we keep at latest on each build.


> Pin Docker images versions

The go images intentionally use the latest go, which we then use to fetch a pinned go version. Pinning these images is counter-productive. It does ~nothing for reproducibility but it does mean we'll have to start bumping them regularly.

> Setup dependabot for the Dockerfiles

Unnecessary if we don't pin.

PR #3475, opened 2024-01-10
Switch to new containerd version - v1.7.11

dims:
While we are at it, ability to set the desired containerd version in the make file itself, rather than hardcoding in the Dockerfile

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -116,9 +116,10 @@ RUN eval "$(gimme "${GO_VERSION}")" \
 
 
 # stage for building containerd
+# set by makefile to the desired containerd version
+ARG CONTAINERD_VERSION
 FROM go-build as build-containerd
 ARG TARGETARCH GO_VERSION
-ARG CONTAINERD_VERSION="v1.7.5"
 ARG CONTAINERD_CLONE_URL="https://github.com/containerd/containerd"
 # we don't build with optional snapshotters, we never select any of these
 # they're not ideal inside kind anyhow, and we save some disk space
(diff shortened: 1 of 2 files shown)


images/base/Dockerfile
```diff
@@ -116,9 +116,10 @@ RUN eval "$(gimme "${GO_VERSION}")" \
 
 
 # stage for building containerd
+# set by makefile to the desired containerd version
+ARG CONTAINERD_VERSION
 FROM go-build as build-containerd
 ARG TARGETARCH GO_VERSION
-ARG CONTAINERD_VERSION="v1.7.5"
```
BenTheElder, line 121: this is inconsistent with every other version in the file ..?

dims: i was mirroring `ARG GO_VERSION` in line 113 above.

aojea: why don't keep the default version here?
Also, what is the benefit, this is more a personal commodity no? I don't imaging now we are going to release multiple images with different containerd versions

dims: @aojea this is to test newer versions of containerd in CI :)

BenTheElder: The go version had to be moved out for consistency with the rest of the repo, but the containerd version only exists here.

You can already do `EXTRA_BUILD_OPT` to pass additional args to the dockerfile, and we can already PR updates to the CONTAINERD_VERSION.

dims: Ack, was trying to make it more obvious/explicit with this PR. will just try to use `EXTRA_BUILD_OPT`

Issue #3436, opened 2023-11-28
Support Kubernetes UserNamespacesSupport alpha feature gate


dgl:
**What happened**:

I'm working on parts of the Kubernetes user namespace support (currently an alpha feature).  I'd like to use kind for testing it.

I enabled the `UserNamespacesSupport` feature gate. Pods that set `hostUsers: false` fail with:

```
Warning  FailedCreatePodSandBox  6s    kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = failed to start sandbox "75fd33edcf39433911025ac0e045581bd19688190cd1e5f7166d279056dc592c": failed to create containerd task: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "sysfs" to rootfs at "/sys": mount src=sysfs, dst=/sys, dstFD=/proc/self/fd/10, flags=0xf: operation not permitted: unknown
```

After fixing that (below), I also saw:

```
Warning  Failed                  5s    kubelet            Error: failed to create containerd task: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error running createContainer hook #0: fork/exec /kind/bin/mount-product-files.sh: permission denied: unknown
```

**What you expected to happen**:

Sweet user namespace based isolation.

**How to reproduce it (as _minimally_ and precisely as possible)**:

Update runc to main in the base image, but also set `runc_nodmz` (because of the bug I reported in https://github.com/opencontainers/runc/pull/4129):

```diff
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -135,13 +135,13 @@ RUN git clone --filter=tree:0 "${CONTAINERD_CLONE_URL}" /containerd \
 # stage for building runc
 FROM go-build as build-runc
 ARG TARGETARCH GO_VERSION
-ARG RUNC_VERSION="v1.1.9"
+ARG RUNC_VERSION="main"
 ARG RUNC_CLONE_URL="https://github.com/opencontainers/runc"
 RUN git clone --filter=tree:0 "${RUNC_CLONE_URL}" /runc \
     && cd /runc \
     && git checkout "${RUNC_VERSION}" \
     && eval "$(gimme "${GO_VERSION}")" \
-    && export GOARCH=$TARGETARCH && export CC=$(target-cc) && export CGO_ENABLED=1 \
[...]

PR #3477, opened 2024-01-11
Avoid setting /kind/bin to 0644

dgl:
This allows runc hooks to work when not running as root (such as with user namespaces, see issue #3436).

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -24,12 +24,14 @@ ARG BASE_IMAGE=debian:bookworm-slim
 FROM $BASE_IMAGE as base
 
 # copy in static files
-# all scripts are 0755 (rwx r-x r-x)
+# all scripts and directories are 0755 (rwx r-x r-x)
 # all non-scripts are 0644 (rw- r-- r--)
 COPY --chmod=0755 files/usr/local/bin/* /usr/local/bin/
 
 COPY --chmod=0644 files/kind/ /kind/
-COPY --chmod=0755 files/kind/bin/* /kind/bin/
+# COPY only applies to files, not the directory itself, so the permissions are
+# fixed in RUN below with a chmod.
+COPY --chmod=0755 files/kind/bin/ /kind/bin/
 
 COPY --chmod=0644 files/LICENSES/* /LICENSES/*
 COPY --chmod=0644 files/etc/* /etc/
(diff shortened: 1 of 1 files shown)


images/base/Dockerfile
```diff
@@ -24,12 +24,14 @@ ARG BASE_IMAGE=debian:bookworm-slim
...
-# all scripts are 0755 (rwx r-x r-x)
+# all scripts and directories are 0755 (rwx r-x r-x)
 # all non-scripts are 0644 (rw- r-- r--)
 COPY --chmod=0755 files/usr/local/bin/* /usr/local/bin/
 
-COPY --chmod=0644 files/kind/ /kind/
-COPY --chmod=0755 files/kind/bin/* /kind/bin/
+# Don't specify a chmod for /kind itself, else /kind/bin ends up with 0644
+# which isn't the correct permission for a directory.
+COPY files/kind/ /kind/
```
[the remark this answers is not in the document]
BenTheElder: there's no kind binary inside the image.

images/base/Dockerfile
```diff
@@ -24,12 +24,14 @@ ARG BASE_IMAGE=debian:bookworm-slim
...
 
 # copy in static files
-# all scripts are 0755 (rwx r-x r-x)
+# all scripts and directories are 0755 (rwx r-x r-x)
 # all non-scripts are 0644 (rw- r-- r--)
 COPY --chmod=0755 files/usr/local/bin/* /usr/local/bin/
 
-COPY --chmod=0644 files/kind/ /kind/
-COPY --chmod=0755 files/kind/bin/* /kind/bin/
+# Don't specify a chmod for /kind itself, else /kind/bin ends up with 0644
```
BenTheElder, line 31: this isn't suitable, because then the build depends on umask.

we'll need another workaround, e.g. `RUN chmod` unfortunately

BenTheElder:
We definitely want some version of this fix, but as-is this regresses versus why we started explicitly setting chmod

images/base/Dockerfile
dgl: Sorry about the delay, using a chmod now.

images/base/Dockerfile
```diff
@@ -42,8 +44,6 @@ COPY --chmod=0644 files/etc/systemd/system/kubelet.service.d/* /etc/systemd/syst
 # Install dependencies, first from apt, then from release tarballs.
 # NOTE: we use one RUN to minimize layers.
```
BenTheElder, line 45: This comment is out of date (we squash with multi-stage) and I'd probably move the chmod to it's own RUN, but those cleanups aren't important and I want to get this into the next release and we have some other bug fixes to prompt a quick follow-up release currently

PR #3480, opened 2024-01-14
Pass proxy environment variables to the building container when building node images

Zumium:
When used in a restricted network environment, KIND may not be able to download necessary resources directly from the Internet during the building process, which makes the building fail. Adding` --http-proxy` option to the "build node-image" command helps to solve the problem by setting up the building container with `http_proxy` and `https_proxy` environment variables and making the downloading to be able to use the provided proxy.

### DIFF
--- a/pkg/build/nodeimage/buildcontext.go
+++ b/pkg/build/nodeimage/buildcontext.go
@@ -35,6 +35,15 @@ import (
 	"sigs.k8s.io/kind/pkg/internal/version"
 )
 
+const (
+	// httpProxy is the HTTP_PROXY environment variable key
+	httpProxy = "HTTP_PROXY"
+	// httpsProxy is the HTTPS_PROXY environment variable key
+	httpsProxy = "HTTPS_PROXY"
+	// noProxy is the NO_PROXY environment variable key
+	noProxy = "NO_PROXY"
+)
+
 // buildContext is used to build the kind node image, and contains
 // build configuration
 type buildContext struct {
@@ -135,6 +144,9 @@ func (c *buildContext) buildImage(bits kube.Bits) error {
 		"docker", "commit",
 		// we need to put this back after changing it when running the image
 		"--change", `ENTRYPOINT [ "/usr/local/bin/entrypoint", "/sbin/init" ]`,
+		// remove proxy settings since they're for the building process
+		// and should not be carried with the built image
+		"--change", `ENV HTTP_PROXY="" HTTPS_PROXY="" NO_PROXY=""`,
 		containerID, c.image,
 	).Run(); err != nil {
 		c.logger.Errorf("Image build Failed! Failed to save image: %v", err)
@@ -314,16 +326,28 @@ func (c *buildContext) createBuildContainer() (id string, err error) {
 	// and a little random bits in case we have multiple builds simultaneously
 	random := rand.New(rand.NewSource(time.Now().UnixNano())).Int31()
 	id = fmt.Sprintf("kind-build-%d-%d", time.Now().UTC().Unix(), random)
+	runArgs := []string{
+		"-d", // make the client exit while the container continues to run
+		// the container should hang forever, so we can exec in it
+		"--entrypoint=sleep",
+		"--name=" + id,
+		"--platform=" + dockerBuildOsAndArch(c.arch),
+		"--security-opt", "seccomp=unconfined", // ignore seccomp
+	}
+	// pass proxy settings from environment variables to the building container
+	// to make them work during the building process
+	for _, name := range []string{httpProxy, httpsProxy, noProxy} {
+		val := os.Getenv(name)
+		if val == "" {
+			val = os.Getenv(strings.ToLower(name))
+		}
+		if val != "" {
+			runArgs = append(runArgs, "--env", name+"="+val)
+		}
+	}
 	err = docker.Run(
 		c.baseImage,
-		[]string{
-			"-d", // make the client exit while the container continues to run
-			// the container should hang forever, so we can exec in it
-			"--entrypoint=sleep",
-			"--name=" + id,
-			"--platform=" + dockerBuildOsAndArch(c.arch),
-			"--security-opt", "seccomp=unconfined", // ignore seccomp
-		},
+		runArgs,
 		[]string{
 			"infinity", // sleep infinitely to keep the container around
 		},


aojea:
we inherit proxy variables from the environment, is this not the case with building ?

https://kind.sigs.k8s.io/docs/user/quick-start/#configure-kind-to-use-a-proxy

I don't expect to add flags for this

pkg/build/nodeimage/buildcontext.go
```diff
@@ -135,6 +136,9 @@ func (c *buildContext) buildImage(bits kube.Bits) error {
 		"docker", "commit",
 		// we need to put this back after changing it when running the image
 		"--change", `ENTRYPOINT [ "/usr/local/bin/entrypoint", "/sbin/init" ]`,
+		// remove http proxy setting
```
aojea, line 139: why remove?

pkg/build/nodeimage/buildcontext.go
```diff
@@ -314,16 +318,23 @@ func (c *buildContext) createBuildContainer() (id string, err error) {
...
+	runArgs := []string{
+		"-d", // make the client exit while the container continues to run
+		// the container should hang forever, so we can exec in it
+		"--entrypoint=sleep",
+		"--name=" + id,
+		"--platform=" + dockerBuildOsAndArch(c.arch),
+		"--security-opt", "seccomp=unconfined", // ignore seccomp
+		"--network=host",
+	}
+	if c.httpProxy != "" {
```
aojea, line 330: this can be obtained from https://github.com/kubernetes-sigs/kind/blob/40c81f187425254daf2bf84360a6257a278252df/pkg/cluster/internal/providers/common/proxy.go#L35-L37

pkg/cmd/kind/build/nodeimage/nodeimage.go
```diff
@@ -86,6 +87,12 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 		"",
 		"architecture to build for, defaults to the host architecture",
 	)
+	cmd.Flags().StringVar(
```
aojea, line 90: no flags

pkg/build/nodeimage/buildcontext.go
Zumium: It's because during my testing, I found that the control panel wasn't able to start up with my built node image, and the built image was carrying the proxy settings. When I removed them, it worked.

Zumium:
@aojea Hi, aojea.  The "http_proxy" and "https_proxy" env vars are not working with building yet because they are not passed to the building container, which is what happened when I tested it. If no flags is the preferred way, I'll adjust my code to read proxy settings from env vars instead of a flag.

pkg/build/nodeimage/buildcontext.go
Zumium: It seems that this `GetProxyEnvs` function is for using in provisioning clusters, because it requires a cluster config as its parameter, which we don't have when building node images. I'll just read from env vars directly.

pkg/build/nodeimage/buildcontext.go
```diff
@@ -314,16 +326,29 @@ func (c *buildContext) createBuildContainer() (id string, err error) {
...
 	random := rand.New(rand.NewSource(time.Now().UnixNano())).Int31()
 	id = fmt.Sprintf("kind-build-%d-%d", time.Now().UTC().Unix(), random)
+	runArgs := []string{
+		"-d", // make the client exit while the container continues to run
+		// the container should hang forever, so we can exec in it
+		"--entrypoint=sleep",
+		"--name=" + id,
+		"--platform=" + dockerBuildOsAndArch(c.arch),
+		"--security-opt", "seccomp=unconfined", // ignore seccomp
+		"--network=host",
```
aojea, line 336: this was not here before, was it?

pkg/build/nodeimage/buildcontext.go
```diff
@@ -135,6 +144,9 @@ func (c *buildContext) buildImage(bits kube.Bits) error {
 		"docker", "commit",
 		// we need to put this back after changing it when running the image
 		"--change", `ENTRYPOINT [ "/usr/local/bin/entrypoint", "/sbin/init" ]`,
+		// remove proxy settings since they're for the building process
+		// and should not be carried with the built image
+		"--change", `ENV HTTP_PROXY="" HTTPS_PROXY="" NO_PROXY=""`,
```
aojea, line 149: I'm not familiar with this, what is this change required?

pkg/build/nodeimage/buildcontext.go
Zumium: The line `--network=host` is added by me, the others were there before. The purpose of `--network=host` is to be able to use local host proxy like "http://127.0.0.1:1080".

pkg/build/nodeimage/buildcontext.go
Zumium: Sorry I may not fully understand your question.  The `--change` option accepts a Dockerfile instruction and applies it to the outputted image. Detail is here: https://docs.docker.com/engine/reference/commandline/commit/#options

pkg/build/nodeimage/buildcontext.go
aojea: you can not submit a PR with multiple changes without informing the reviewers, this is a change in behavior, it requires additional discussion, and in this case it seems that is only to fit your personal workflow

aojea: please do only one change at a time

pkg/build/nodeimage/buildcontext.go
aojea: but why do we need to use it?

pkg/build/nodeimage/buildcontext.go
aojea: this does not gives much confidence on the solution proposed :/

pkg/build/nodeimage/buildcontext.go
Zumium: Right now it's only based on my local testing result (control panel won't start successfully if they're not removed), and I haven't figured out what's going on underneath. I'll offer a detailed explanation in a few days cause I can only do this after work, please bear with me.

aojea: there is no rush, take your time, but if we add any option we need to be sure we don't break any existing workflow

pkg/build/nodeimage/buildcontext.go
liangyuanpeng: > The purpose of --network=host is to be able to use local host proxy like "http://127.0.0.1:1080/".

@Zumium 
This is not necessary for the current PR, you can still use the proxy even without the host network, doesn't it?

pkg/build/nodeimage/buildcontext.go
BenTheElder: this is necessary so we don't wind up saving the env to the saved image

pkg/build/nodeimage/buildcontext.go
BenTheElder: > The line --network=host is added by me, the others were there before. The purpose of --network=host is to be able to use local host proxy like "http://127.0.0.1:1080/".

We should drop this, because you can't run node containers this way anyhow and will need a different solution for proxy?

BenTheElder:
sorry for the delay, we should still make one of the requested changes re: host network

pkg/build/nodeimage/buildcontext.go
Zumium: @liangyuanpeng Well, I'm afraid it's necessary especially for a proxy running on localhost. Imagine your proxy server is running locally on host, for example http://127.0.0.1:1080, which I believe is a common case. If the address is directly set to the container and not using host network, "127.0.0.1" will refer to the container's localhost, and the programs inside the container will fail to connect to proxy because they're not connecting to the right place.

Zumium: @BenTheElder I believe that the core of the problem is the meaning of "localhost" changes inside & outside of the container, and proxies usually bind on localhost for safety reason, preventing a process inside container from connecting to it directly. If host network is not accepted, we may have to add some kind of relay-er, and it will be a more complex solution.

aojea: you can use the IP of your host, proxies use to listen in all host ip addresses, is easier to modify your proxy than force this pod to run with hostnetwork

Zumium: It's also ok to just keep KIND simple and leave the worry to users. I'll modify my code later.

aojea: I don't disagree, but this was this way for years and everytime we change something that sounds simple, it always have consequences

aojea:
can you squash all commits?

Zumium:
> can you squash all commits?

Done.

Zumium:
Thank everyone for helping! What a coincidence that today happens to be 除夕(chu xi) festival, which marks the beginning of a new Chinese year. Again, thank you everyone and happy Chinese new year!

liangyuanpeng:
happy new year :)

PR #3483, opened 2024-01-18
Add ipFamily validation

tnqn:
Setting ipFamily to invalid value didn't return any error and kind just created an ipv4 cluster silently. It could be a bit confusing when it is set to some plausible but invalid values like IPv4, IPv6, Dual, ds.

### DIFF
--- a/pkg/internal/apis/config/validate.go
+++ b/pkg/internal/apis/config/validate.go
@@ -52,6 +52,11 @@ func (c *Cluster) Validate() error {
 		}
 	}
 
+	// ipFamily should be ipv4, ipv6, or dual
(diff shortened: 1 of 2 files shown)


BenTheElder:
/hold
[going to merge this after we tag the next patch release currently in flight, just because that release is already a regression fix]

tnqn:
@aojea @BenTheElder it seems specific role is needed to rerun the failed github actions, could you help rerun them?

aojea:
> @aojea @BenTheElder it seems specific role is needed to rerun the failed github actions, could you help rerun them?

rebase and repush, I think that it triggers the github actions again

BenTheElder:
we can manually re-run, but I think actions only picks up config changes in the current branch

I'm OK merging this without the actions passing since they're not really testing for this and the change isn't node-runtime specific, we can always manually bypass.

PR #3382, opened 2023-10-12
GitHub Actions: use "larger" runners for nested virt

AkihiroSuda:
Fix #3364

PR #3484, opened 2024-01-19
GitHub Actions: use Linux runners for nested virt

AkihiroSuda:
"Larger" runners are no longer needed for Linux with nested virt.

Replaces:
- #3382

### DIFF
--- a/.github/workflows/cgroup2.yaml
+++ b/.github/workflows/cgroup2.yaml
@@ -14,8 +14,7 @@ permissions:
 jobs:
   docker:
     name: Cgroup v2
-    # nested virtualization is only available on macOS hosts
-    runs-on: macos-12
+    runs-on: ubuntu-22.04
     timeout-minutes: 30
     strategy:
       fail-fast: false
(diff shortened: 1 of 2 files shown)


.github/workflows/cgroup2.yaml
```diff
@@ -14,15 +14,15 @@ permissions:
 jobs:
   docker:
     name: Cgroup v2
-    # nested virtualization is only available on macOS hosts
-    runs-on: macos-12
+    runs-on: ubuntu-22.04
```
aojea, line 17: interesting

AkihiroSuda:
https://github.com/kubernetes-sigs/kind/actions/runs/7579837506/job/20644695181?pr=3484

```
==> default: Running provisioner: install-packages (shell)...
    default: Running: inline script
    default: ++ modprobe tap ip_tables iptable_nat ip6_tables ip6table_nat
    default: ++ sh
    default: ++ curl -fsSL https://get.docker.com
    default: # Executing docker install script, commit: e5543d473431b782227f8908005543bb4389b8de
    default: + sh -c 'dnf install -y -q dnf-plugins-core'
    default: main: line 593:  2636 Killed                  $sh_c "$pkg_manager install -y -q $pre_reqs"
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
```

🤔

aojea:
how can we get more verbosity on the error?  some package failing to install ?

AkihiroSuda:
Switched back from libvirt to VirtualBox.
Now CI is green.

hack/ci/Vagrantfile
```diff
@@ -8,9 +8,7 @@ Vagrant.configure("2") do |config|
   # Mirrors can be found at here: https://admin.fedoraproject.org/mirrormanager/mirrors/Fedora/37/x86_64
   config.vm.box = "dummy"
   config.vm.box_url = "https://iad.mirror.rackspace.com/fedora/releases/37/Cloud/x86_64/images/Fedora-Cloud-Base-Vagrant-37-1.7.x86_64.vagrant-virtualbox.box"
-  # macOS github actions runners have 3 cores and 14 GB of ram
   # assume some ram is needed for the host environment but very little CPU
-  # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners
   memory = 10240
   cpus = 3
```
aojea, line 13: linux workers seems to have only 2 cpus and 7GB of ram 🤔 https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

AkihiroSuda: https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/

> We are pleased to announce that we completed that upgrade today, and we now provide machines that are double their previous specification, with 4-vCPUs, 16 GiB of memory, and 150GiB storage.

aojea: love it

PR #3485, opened 2024-01-21
CI: rename cgroup2.yaml to vm.yaml

AkihiroSuda:
"cgroup2.yaml" is becoming misnomer, as PR #3409 is going to enable native ubuntu-22.04 runners with cgroup2.

### DIFF
--- a/.github/workflows/vm.yaml
+++ b/.github/workflows/vm.yaml
@@ -1,4 +1,4 @@
-name: Cgroup v2
+name: VM
 
 on:
   workflow_dispatch:
@@ -12,8 +12,9 @@ permissions:
   contents: read
 
 jobs:
-  docker:
-    name: Cgroup v2
+  vm:
+    # Fedora is different from Ubuntu in LSM (SELinux), filesystem (btrfs), kernel version, etc.
+    name: "VM (Fedora)"
     runs-on: ubuntu-22.04
     timeout-minutes: 30
     strategy:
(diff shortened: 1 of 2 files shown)


aojea:
we don't want to do platform validation, only functionality, I thought we were only using Fedora because of the lack of cgroupsv2, if we are going to have them natively why we need a Fedora runner? can we run the rootless tests on the new runners?

AkihiroSuda:
> we don't want to do platform validation, only functionality, I thought we were only using Fedora because of the lack of cgroupsv2, if we are going to have them natively why we need a Fedora runner? can we run the rootless tests on the new runners?

Fedora is also used for covering SELinux, btrfs, newer kernel, etc.
It is possible to run the rootless tests on the native ubuntu-22.04 runners, I suggest to keep Fedora as well.

aojea:
> Fedora is also used for covering SELinux, btrfs, newer kernel, etc.
> While It is possible to run the rootless tests on the native ubuntu-22.04 runners, I suggest to keep Fedora as well.

fair enough but I'd prefer to be more neutral to distros, can we come up with another name that represents these features more generically? `cutting-edge features`, `advanced features ` , ...

AkihiroSuda:
> > Fedora is also used for covering SELinux, btrfs, newer kernel, etc.
> > While It is possible to run the rootless tests on the native ubuntu-22.04 runners, I suggest to keep Fedora as well.
> 
> fair enough but I'd prefer to be more neutral to distros, can we come up with another name that represents these features more generically? `cutting-edge features`, `advanced features ` , ...

What about just calling it "vm"?

aojea:
ok

aojea:
/lgtm
/approve
Thanks

PR #3495, opened 2024-01-30
update Contour ingress docs

### LINKED ISSUE #3181: Contour ingress isn't working

skriss:
Updates the Contour ingress docs to link
to the Contour website for a usage example
since the kind example is nginx-specific.

Closes #3181.

### DIFF
--- a/site/content/docs/user/ingress.md
+++ b/site/content/docs/user/ingress.md
@@ -76,8 +76,10 @@ Apply it by running:
 kubectl patch daemonsets -n projectcontour envoy -p '{{< minify file="static/examples/ingress/contour/patch.json" >}}'
 {{< /codeFromInline >}}
 
-Now the Contour is all setup to be used.
-Refer to [Using Ingress](#using-ingress) for a basic example usage.
+Now Contour is ready to be used.
+Refer to Contour's [Getting Started](https://projectcontour.io/getting-started/#test-it-out) documentation for a basic usage example.
+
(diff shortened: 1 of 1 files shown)


kundan2707 (commented):
Thanks for this important update 
/lgtm

BenTheElder (commented):
/lgtm
/approve

PR #3497, opened 2024-02-01
Update runc and containerd.

ameukam:
Runc Release notes: https://github.com/opencontainers/runc/releases/tag/v1.1.12
Containerd release notes: https://github.com/containerd/containerd/releases/tag/v1.7.13

PR #3499, opened 2024-02-02
update images

BenTheElder:
picks up #3497 & #3498, k8s 1.29.1


BenTheElder:
We probably need to give the e2e's building k8s >4 CPU these days, the node-image builds are taking a long time.
We do in k/k AFAIK.

BenTheElder:
Also at some point we stopped logging messages from the e2e tests until the test is over ...

Issue #3506, opened 2024-02-05
Interaction between kind/kubeadm and new metallb versions breaks single-node metallb usage


howardjohn:
metallb issue: https://github.com/metallb/metallb/issues/2274

As of 0.14, metallb respects the `node.kubernetes.io/exclude-from-external-load-balancers` and will not make announcements for these nodes.

As of k8s 1.21:

> Kubeadm: apply the "node.kubernetes.io/exclude-from-external-load-balancers" label on control plane nodes during "init", "join" and "upgrade" to preserve backwards compatibility with the lagacy LB mode where nodes labeled as "master" where excluded. To opt-out you can remove the label from a node. See #97543 and the linked KEP for more details. ([#98269](https://github.com/kubernetes/kubernetes/pull/98269), [@neolit123](https://github.com/neolit123)) [SIG Cluster Lifecycle]

As a result, this documentation https://kind.sigs.k8s.io/docs/user/loadbalancer/ does not work on single-node clusters with recent metallb versions.

Possible paths forward:
* Change metallb
* Document to remove the label from the node
* Default to removing the label from the node

BenTheElder:
Ouch, I didn't expect this to still be unhandled.

We should not change the labeling behavior in kind, most software will have already switched since 1.21 is pretty old now and changing it again will just confuse everyone.

Most likely we should fix metallb, if they're unwilling to accept this then we can fallback to just documenting the workaround.

BenTheElder:
[prior comment deleted, I skimmed this => confused with another kubeadm label change around the same time frame]

Auto-removing this for single-node clusters seems reasonable given we auto-remove the comparable general scheduling taint, in the meantime we can document the workaround.

BenTheElder:
v0.22.0 is out with this fix, thanks @howardjohn !

PR #3507, opened 2024-02-05
Remove `exclude-from-external-load-balancers` from single node clusters

howardjohn:
Part of https://github.com/kubernetes-sigs/kind/issues/3506

### DIFF
--- a/pkg/cluster/internal/create/actions/kubeadminit/init.go
+++ b/pkg/cluster/internal/create/actions/kubeadminit/init.go
@@ -142,6 +142,17 @@ func (a *action) Execute(ctx *actions.ActionContext) error {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


neolit123 (commented):
/lgtm
we should probably mention this in the kubeadm "create a cluster" page too.

BenTheElder (commented):
just confirming: this was added in 1.21? so this is a no-op for _really_ old unsupported clusters and not worth worrying about version checks.

thanks! LGTM

BenTheElder:
> we should probably mention this in the kubeadm "create a cluster" page too.

Agreed, do you prefer to track that sort of thing at k/website?

neolit123:
> > we should probably mention this in the kubeadm "create a cluster" page too.
> 
> Agreed, do you prefer to track that sort of thing at k/website?

i can log a new k/website ticket.

BenTheElder:
[also confirming for next time, thanks!]

Issue #3510, opened 2024-02-08
v0.21.0: kind container makes changes to binfmt configutation


f84anton:
**What happened**:
I'm using orbstack on macbook m1, upgraded to kind 0.21.0 and created a new cluster with k8s 1.23, image `kindest/node:v1.23.17@sha256:fbb92ac580fce498473762419df27fa8664dbaa1c5a361b5957e123b4035bdcf`
After that I've got problem with running amd64 images in docker or inside kind:
```
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
exec /service: exec format error
```

config file:
```
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
featureGates:
  TTLAfterFinished: true
containerdConfigPatches:
- |-
  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."localhost:5000"]
    endpoint = ["http://kind-registry:5000"]
  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."kind-registry:5000"]
    endpoint = ["http://kind-registry:5000"]
kubeadmConfigPatches:
- |
  apiVersion: kubelet.config.k8s.io/v1beta1
  kind: KubeletConfiguration
  evictionHard:
    nodefs.available: "1%"
    imagefs.available: "1%"
    nodefs.inodesFree: "1%"
    memory.available: "50Mi"
  imageMinimumGCAge: "1h"
  imageGCHighThresholdPercent: 90
  imageGCLowThresholdPercent: 85
  allowedUnsafeSysctls:
  - "net.*"
nodes:
- role: control-plane
  extraPortMappings:
  - containerPort: 30080
    hostPort: 80
  - containerPort: 30443
    hostPort: 443
```

Here is how I created a cluster:
```
kind create cluster \
 --name kind \
 --image  "kindest/node:v1.23.17@sha256:fbb92ac580fce498473762419df27fa8664dbaa1c5a361b5957e123b4035bdcf"\
 --config /path/to/config \
 --wait 5m`
 ```

Here is binfmt info after kind container was created:
```
%  docker run --privileged --rm tonistiigi/binfmt
{
  "supported": [
    "linux/arm64"
  ],
  "emulators": [
    "python3.11"
  ]
}
```

After `kind cluster delete` and orbstack restart I see that the list goes back to normal:
```
%  docker run --privileged --rm tonistiigi/binfmt
{
  "supported": [
[...]

PR #3511, opened 2024-02-08
[WIP] disable systemed-binfmt.service

BenTheElder:
related to #3510

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -86,10 +86,12 @@ RUN echo "Installing Packages ..." \
     && echo "ReadKMsg=no" >> /etc/systemd/journald.conf \
     && ln -s "$(which systemd)" /sbin/init
 
-RUN echo "Enabling services ... " \
+# NOTE: systemd-binfmt.service will register things into binfmt_misc which is kernel-global
+RUN echo "Enabling / Disabling services ... " \
     && systemctl enable kubelet.service \
     && systemctl enable containerd.service \
-    && systemctl enable undo-mount-hacks.service
+    && systemctl enable undo-mount-hacks.service \
+    && systemctl mask systemd-binfmt.service
 
 RUN echo "Ensuring /etc/kubernetes/manifests" \
     && mkdir -p /etc/kubernetes/manifests


images/base/Dockerfile
```diff
@@ -86,10 +86,12 @@ RUN echo "Installing Packages ..." \
...
     && ln -s "$(which systemd)" /sbin/init
 
-RUN echo "Enabling services ... " \
+# NOTE: systemd-binfmt.service will register things into binfmt_misc which is kernel-global
+RUN echo "Enabling / Disabling services ... " \
     && systemctl enable kubelet.service \
     && systemctl enable containerd.service \
-    && systemctl enable undo-mount-hacks.service
+    && systemctl enable undo-mount-hacks.service \
+    && systemctl mask systemd-binfmt.service
```
aojea, line 94: it seems something happened with this in bookworm, a quick search on google show several hits https://github.com/microsoft/WSL/issues/8843 

AFAIK this reads from a folder, can we have any side effect of doing this hammer approach or should we try to be more precise and understand what is exactly breaking this?

aojea: oh, I see, it screws the list entirely https://github.com/kubernetes-sigs/kind/issues/3510

aojea: I imagine then that the best way is to say the docker container must use the one existing in the host

/lgtm

BenTheElder: so often binfmt_misc is intentionally set by containers (see e.g. our multi-arch build setup), but kind nodes shouldn't be doing it, and python scripts should keep using shebangs and not depend on this anyhow ...

dgl: Per https://systemd.io/CONTAINER_INTERFACE/ (execution environment number 2) `/proc/sys` being a read-only mount is the recommended way to tell systemd not to change things. However as that mentions also making `/proc/sys/net` writable may be desirable (and likely would be needed for many k8s workloads).

It likely needs more thought than will fit in a review comment, so this targeted workaround makes sense for now.

BenTheElder: Yeah, the base image comments link to this page -- We have other similar workarounds for similar reasons, we're really stretching "systemd in a container" when we go `--privileged` but without it we can't get the "kubernetes 'in' a container" part. Previously in practice the key part was `/sys` mounted ro (mentioned further down as a trigger for udev) which we do.

PR #3513, opened 2024-02-09
Support multiple random ExtraPortMappings

### LINKED ISSUE #3512: Cannot use multiple extraPortMappings with default hostPort
**What happened**:

When creating a kind cluster using the config:

```
  extraPortMappings:
  - containerPort: 80
    listenAddress: 127.0.0.1
    protocol: TCP
  - containerPort: 443
    listenAddress: 127.0.0.1
    protocol: TCP
```

kind errors with:

```
ERROR: failed to create cluster: invalid configuration for node 0: invalid portMapping: port mapping with same listen address, port and protocol already configured: 127.0.0.1:0/TCP
```

Both hostPorts to be picked at random by kind, instead of an error about port 0 already being chosen.

```bash
kind: Cluster
nodes:
- role: control-plane
  extraPortMappings:
  - containerPort: 80
    listenAddress: 127.0.0.1
    protocol: TCP
  - containerPort: 443
    listenAddress: 127.0.0.1
    protocol: TCP
EOF
```

* The same happens when using hostPort -1
* It works fine when only specifying one port mapping, so the issue seems to be with the validation which should allow duplicate hostPorts of 0 and -1. For example, this works:

```bash
kind: Cluster
nodes:
- role: control-plane
  extraPortMappings:
  - containerPort: 80
    listenAddress: 127.0.0.1
    protocol: TCP
EOF

docker port kind-control-plane       
80/tcp -> 127.0.0.1:40705
6443/tcp -> 127.0.0.1:35909
```

**Environment:**

```
kind v0.21.0 go1.21.6 linux/amd64
```

- Runtime info: (use `docker info` or `podman info`):

```
Client:
 Version:    24.0.7
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  0.12.1
    Path:     /usr/lib/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  2.24.3
    Path:     /usr/lib/docker/cli-plugins/docker-compose
```

```
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
```

```
Client Version: v1.29.1
```

wouterh-dev:
When the default value, 0, is used for the HostPort in ExtraPortMappings then Kind will determine a random HostPort to use for this mapping. The validation only allowed a single instance of such a mapping, but now allows multiple.

Also delay the closing of the randomly determined port until all random ports have been determined to ensure the same port cannot be returned multiple times by the operating system.

Fixes #3512

### DIFF
--- a/pkg/cluster/internal/providers/common/getport_test.go
+++ b/pkg/cluster/internal/providers/common/getport_test.go
@@ -41,7 +41,10 @@ func TestPortOrGetFreePort(t *testing.T) {
 		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
-			got, err := PortOrGetFreePort(tt.port, "localhost")
+			got, releaseHostPortFn, err := PortOrGetFreePort(tt.port, "localhost")
+			if releaseHostPortFn != nil {
+				releaseHostPortFn()
+			}
 			if (err != nil) != tt.wantErr {
 				t.Errorf("PortOrGetFreePort() error = %v, wantErr %v", err, tt.wantErr)
 				return
@@ -84,7 +87,10 @@ func TestGetFreePort(t *testing.T) {
 		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
-			got, err := GetFreePort(tt.listenAddr)
+			got, releaseHostPortFn, err := GetFreePort(tt.listenAddr)
+			if releaseHostPortFn != nil {
+				releaseHostPortFn()
+			}
 			if (err != nil) != tt.wantErr {
 				t.Errorf("GetFreePort() error = %v, wantErr %v", err, tt.wantErr)
 				return
--- a/pkg/internal/apis/config/validate.go
+++ b/pkg/internal/apis/config/validate.go
@@ -155,6 +155,11 @@ func validatePortMappings(portMappings []PortMapping) error {
 		addr := net.ParseIP(portMapping.ListenAddress)
 		addrString := addr.String()
 
+		if (portMapping.HostPort == -1 || portMapping.HostPort == 0) && (portMapping.Protocol == PortMappingProtocolTCP || portMapping.Protocol == "") {
+			// Port -1 and 0 cause a random port to be selected, thus duplicates are allowed
+			continue
+		}
+
 		portProtocol := formatPortProtocol(portMapping.HostPort, portMapping.Protocol)
 		possibleErr := fmt.Errorf("%s: %s:%s", errMsg, addrString, portProtocol)
 
--- a/pkg/internal/apis/config/validate.go
+++ b/pkg/internal/apis/config/validate.go
@@ -152,13 +152,13 @@ func validatePortMappings(portMappings []PortMapping) error {
 	}
 
 	for _, portMapping := range portMappings {
-		addr := net.ParseIP(portMapping.ListenAddress)
-		addrString := addr.String()
-
-		if (portMapping.HostPort == -1 || portMapping.HostPort == 0) && (portMapping.Protocol == PortMappingProtocolTCP || portMapping.Protocol == "") {
+		if portMapping.HostPort == -1 || portMapping.HostPort == 0 {
 			// Port -1 and 0 cause a random port to be selected, thus duplicates are allowed
 			continue
 		}
+
+		addr := net.ParseIP(portMapping.ListenAddress)
+		addrString := addr.String()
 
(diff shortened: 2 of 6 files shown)


pkg/cluster/internal/providers/common/getport_test.go
```diff
@@ -41,7 +41,10 @@ func TestPortOrGetFreePort(t *testing.T) {
 		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
-			got, err := PortOrGetFreePort(tt.port, "localhost")
+			got, releaseHostPortFn, err := PortOrGetFreePort(tt.port, "localhost")
+			if releaseHostPortFn != nil {
+				releaseHostPortFn()
```
aojea, line 46: why do you want to defer the close  here?

pkg/internal/apis/config/validate.go
```diff
@@ -155,6 +155,11 @@ func validatePortMappings(portMappings []PortMapping) error {
 		addr := net.ParseIP(portMapping.ListenAddress)
 		addrString := addr.String()
 
+		if (portMapping.HostPort == -1 || portMapping.HostPort == 0) && (portMapping.Protocol == PortMappingProtocolTCP || portMapping.Protocol == "") {
+			// Port -1 and 0 cause a random port to be selected, thus duplicates are allowed
+			continue
+		}
```
aojea, line 161: this does not look great to me, seems very hacky

pkg/cluster/internal/providers/common/getport_test.go
aojea: > Also delay the closing of the randomly determined port until all random ports have been determined to ensure the same port cannot be returned multiple times by the operating system.

I see, there is a TIME_WAIT period that by default is 1 min or 2 min, I can't remember now exactly which one

pkg/internal/apis/config/validate.go
wouterh-dev: Do you have a suggestion on how to improve this condition? The goal is to match TCP port 0 and -1 which already were considered special cases by Kind. 

I noticed that the unit tests set the portMapping Protocol to an empty string, and when parsing my config file it's "TCP", which is what this condition checks for.

pkg/cluster/internal/providers/common/getport_test.go
wouterh-dev: I defer the close so there is no chance the operating system might assign the same port twice when asking for a random ("0" port).

Relying on TIME_WAIT seemed risky since I'm not sure how the kernel behaves for a socket that never had a connection, but I haven't actually seen the same port be assigned twice.

If you prefer I'll revert the deferred closing.

pkg/internal/apis/config/validate.go
aojea: the problem is the feature itself is very hacky https://github.com/kubernetes-sigs/kind/issues/3512#issuecomment-1936124620 , you are delegating something to kind "assign random ports to container ports" for later require an external tool to return those values

wouterh-dev: This feature already existed before this PR. I did not introduce the feature. I am merely allowing it to be used multiple times.

aojea: yeah, we talked on the referenced issue, I'm going to resolve this for the sake of starting the conversation from scratch

pkg/cluster/internal/providers/common/getport_test.go
aojea: hmm, my bad, time_wait does not apply here 
image

aojea: ok, this looks correct, since we generate ports in a loop, this holds all these random ports during the loop

pkg/internal/apis/config/validate.go
aojea, line 161: why the protocol check?
I think this should just go at the beginning of the loop and remove the check for protocol, no?

wouterh-dev: I added the protocol check because the allocation logic in Kind can only allocate TCP port, but docker does support multiple random UDP ports, so you're right and the lack of support in Kind for allocating random UDP ports should probably be considered a different issue (if at all).

wouterh-dev: I pushed a commit to move the check to the begin of the loop & drop the protocol condition

aojea:
/lgtm
/approve

Thanks for the constructive discussion

PR #3516, opened 2024-02-09
Support multiple -1 ExtraPortMappings

wouterh-dev:
When the value -1 is used then Kind will translate this to 0 and pass it to the backend, which causes the backend to assign a random port. Since these ports are random, duplicates can be allowed.

This is an alternative to https://github.com/kubernetes-sigs/kind/pull/3513 as discussed in https://github.com/kubernetes-sigs/kind/issues/3512


aojea:
/close

Let's get https://github.com/kubernetes-sigs/kind/pull/3513 in

PR #3518, opened 2024-02-13
[WIP] Make /proc/sys read-only with carve-outs for some sysctls

dgl:
As mentioned on #3511 this could be a more complete way to ensure systemd or other components don't change sysctls unexpectedly. This also makes sysfs mountable per #3436 (but that is just the mount of sysfs on `/kind/private/sys`, so can easily be split, aside from any naming preferences).

WIP as I'm not sure it's the best option, but possibly better than fragile breakage due to unexpected sysctl changes.

The downside is it needs an allow list of sysctls which is probably going to need additions for other use cases, but it does mean kind can be explicit about what is supported.

The workaround to add a sysctl as writable would be:

```
docker exec a-node mount --rbind /kind/private/proc/sys/some-sysctl /proc/sys/some-sysctl
```

(This currently won't support running in some userns configurations yet, but it should be a case of just ignoring the error from mount if it errors (it can work, it depends on the exact userns environment). In a user namespace the host's sysctls can't be modified anyway. I can test userns cases if this option is worth taking further.)

[1]: https://systemd.io/CONTAINER_INTERFACE/

### DIFF
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -172,13 +172,24 @@ fix_mount() {
     sync
   fi
 
+  # Mount sysfs and proc as read-write, on a known, but kind specific location.
+  # This allows bind mounting, below and is also required to run some workloads
+  # which need to mount proc and sysfs themselves (this avoids the proc and
+  # sysfs mounts being "masked", as far as the kernel is concerned).
+  # XXX, better ref for fs_fully_visible than kernel code?
+  # https://github.com/torvalds/linux/commit/1b852bceb0d1
+  log_info 'mounting /kind/private filesystems'
+  mount -t sysfs -o rw sysfs /kind/private/sys
+  mount -t proc -o rw proc /kind/private/proc
+
   log_info 'remounting /sys read-only'
   # systemd-in-a-container should have read only /sys
   # https://systemd.io/CONTAINER_INTERFACE/
   # however, we need other things from `docker run --privileged` ...
   # and this flag also happens to make /sys rw, amongst other things
   #
-  # This step is ignored when running inside UserNS, because it fails with EACCES.
+  # This step is ignored when running inside UserNS, because it can fail with
+  # EACCES.
   if ! mount -o remount,ro /sys; then
     if [[ -n "$userns" ]]; then
       log_info 'UserNS: ignoring mount fail'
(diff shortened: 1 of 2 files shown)


images/base/files/usr/local/bin/entrypoint
```diff
@@ -187,6 +198,23 @@ fix_mount() {
...
+  for mount_point in \
+      kernel/shmall kernel/shmmax kernel/shmmni kernel/shm_rmid_forced kernel/msgmax kernel/msgmnb kernel/msgmni \
+      fs/mqueue \
+      net \
+      vm/overcommit_memory vm/panic_on_oom kernel/panic kernel/panic_on_oops \
+      kernel/keys/root_maxkeys kernel/keys/root_maxbytes; do
+    if [[ -f /kind/private/proc/sys/"${mount_point}" ]]; then
+      mount --bind -o rw /kind/private/proc/sys/"${mount_point}" /proc/sys/"${mount_point}"
+    fi
+  done
```
AkihiroSuda, line 216: I'm not sure about the robustness of this.
I think this should be opt-in.

images/base/files/usr/local/bin/entrypoint
```diff
@@ -172,13 +172,24 @@ fix_mount() {
...
+
   log_info 'remounting /sys read-only'
   # systemd-in-a-container should have read only /sys
   # https://systemd.io/CONTAINER_INTERFACE/
   # however, we need other things from `docker run --privileged` ...
   # and this flag also happens to make /sys rw, amongst other things
   #
-  # This step is ignored when running inside UserNS, because it fails with EACCES.
+  # This step is ignored when running inside UserNS, because it can fail with
+  # EACCES.
```
AkihiroSuda, line 192: Unnecessary change

AkihiroSuda:
>  ensure systemd or other components don't change sysctls unexpectedly

Rootless mode ( https://kind.sigs.k8s.io/docs/user/rootless/ ) almost solves this issue.

BenTheElder:
> As mentioned on https://github.com/kubernetes-sigs/kind/pull/3511 this could be a more complete way to ensure systemd or other components don't change sysctls unexpectedly. This also makes sysfs mountable per https://github.com/kubernetes-sigs/kind/issues/3436 (but that is just the mount of sysfs on /kind/private/sys, so can easily be split, aside from any naming preferences).

I'm really hesitant to ship a change like this because it's hard to say how we'll break users that have come to rely on this over the years and disabling something like udev/binfmt misc on the other hand is cheap and reasonable, at the risk of missing some future systemd behavior.

stmcginnis:
Is this still a WIP? As Ben mentioned, it does seem like a risky change to make, so if we're not going to do it, it might be good to close.

BenTheElder:
I appreciate the contribution and intent, but I think this is probably riskier than future systemd changes, we at least know to watch for these in the future. Closing for now

PR #3520, opened 2024-02-13
images: allow overriding IMAGE_NAME

BenTheElder:
then you can do like `make -C images/base IMAGE_NAME=kind-base REGISTRY=$(whoami)` instead of having to clobber `IMAGE` which contains the computed tag


neolit123 (commented):
/lgtm

PR #3530, opened 2024-02-26
enable CDI on containerd

aojea:
I think this is what we just need for supporting CDI devices in kind

 I still need to test it

### DIFF
--- a/images/base/files/etc/containerd/config.toml
+++ b/images/base/files/etc/containerd/config.toml
@@ -40,3 +40,6 @@ version = 2
   tolerate_missing_hugepages_controller = true
   # restrict_oom_score_adj needs to be true when running inside UserNS (rootless)
   restrict_oom_score_adj = false
+  # enable CDI https://github.com/cncf-tags/container-device-interface
+  enable_cdi = true
+  cdi_spec_dirs = ["/etc/cdi", "/var/run/cdi"]


klueska:
Yes, this is all that is required. Though the specification of the dirs is unnecessary, as the ones you provide are the defaults.

See the following for how we have been enabling this in one of our repos via `containerdConfigPatches`:
https://github.com/NVIDIA/k8s-dra-driver/blob/main/demo/clusters/kind/scripts/kind-cluster-config.yaml#L20-L25

klueska (approved):

images/base/files/etc/containerd/config.toml
```diff
@@ -40,3 +40,6 @@ version = 2
   tolerate_missing_hugepages_controller = true
   # restrict_oom_score_adj needs to be true when running inside UserNS (rootless)
   restrict_oom_score_adj = false
+  # enable CDI https://github.com/cncf-tags/container-device-interface
+  enable_cdi = true
```
BenTheElder, line 44: why is this default false? 👀

aojea: @estesp @mikebrow can you help us here ? ^^

aojea: I see, is a 2.0 thing https://github.com/containerd/containerd/commit/c8e8a093ce3dd9231173042e9828a679972f5d6b

estesp: correct

aojea:
it will be part of 2.0 containerd

PR #3549, opened 2024-03-13
go.mod: include indirect dependencies

AkihiroSuda:
Bump up the go.mod format so as to help analyzing all the dependencies

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module sigs.k8s.io/kind
 
-go 1.16
+go 1.21
 
 require (
 	github.com/BurntSushi/toml v1.0.0
@@ -14,4 +14,11 @@ require (
 	github.com/spf13/pflag v1.0.5
 	gopkg.in/yaml.v3 v3.0.1
 	sigs.k8s.io/yaml v1.3.0
+)
+
+require (
+	github.com/inconshreveable/mousetrap v1.0.0 // indirect
+	github.com/kr/text v0.2.0 // indirect
+	golang.org/x/sys v0.0.0-20210630005230-0f9fa26af87c // indirect
+	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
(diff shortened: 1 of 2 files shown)


stmcginnis:
Makes sense if we are bumping up the minimum go version. But that has intentionally been kept back. I would defer to @BenTheElder on whether that is something we want right now or not.

AkihiroSuda:
@BenTheElder Could you take a look?

go.mod
```diff
@@ -1,6 +1,6 @@
 module sigs.k8s.io/kind
 
-go 1.16
+go 1.21
```
BenTheElder, line 3: we've intentionally held this back, people `go get` / `go install` kind and use systems with outdated go and we have no particular need for latest go features in the core tool

BenTheElder: we've been hacking on bumping it to 1.17 for related reasons but not as far as 1.21 currently https://github.com/kubernetes-sigs/kind/pull/3335

BenTheElder: I've been waiting for @aojea to take another review pass at https://github.com/kubernetes-sigs/kind/pull/3335, but if we wanted to land a more targeted change, just bumping to 1.17 here will accomplish the same thing with less impact.

AkihiroSuda: They mustn’t use Go 1.20, as it has reached EOL and no longer receives security fixes.

BenTheElder: That's their choice, the kind binary is not particularly sensitive to these as a local non-networked CLI. We document that 1.16+ is required to install currently and have no use case for newer go features for the core project.

Being able to `go install` on e.g., debian stable, without issues, is something we do not want to break lightly.

BenTheElder: And again, to get the module pruning, you only need to increase this to 1.17

BenTheElder: We don't break users without a good reason, lots of new k8s users are installing the kind CLI via their system go and it works fine.

similarly we still have limited support for kubernetes versions that kubernetes no longer supports (e.g. so lagging users can test before/after updating), it would be easier to force only current versions but it has value to users not to and the burden is not too high

AkihiroSuda: Sure, closing this for niw

PR #3557, opened 2024-03-21
Determine which nerdctl implementation

estesp:
When called via the library path, the nerdctl provider is instantiated without a binary name. We still need to do a lookup to determine if finch or nerdctl is the installed binary to provide the local runtime command line

### DIFF
--- a/pkg/cluster/internal/providers/nerdctl/provider.go
+++ b/pkg/cluster/internal/providers/nerdctl/provider.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
+	osexec "os/exec"
 	"path/filepath"
 	"strings"
 
@@ -40,9 +41,21 @@ import (
 
 // NewProvider returns a new provider based on executing `nerdctl ...`
 func NewProvider(logger log.Logger, binaryName string) providers.Provider {
-	// if unset, default to nerdctl
+	// if binaryName is unset, do a lookup; we may be here via a
+	// library call to provider.DetectNodeProvider(), which returns
+	// true from nerdctl.IsAvailable() by checking for both finch
+	// and nerdctl. If we don't redo the lookup here, then a finch
+	// install that triggered IsAvailable() to be true would fail
+	// to be used if we default to nerdctl when unset.
 	if binaryName == "" {
+		// default to "nerdctl"; but look for "finch" if
+		// nerctl binary lookup fails
 		binaryName = "nerdctl"
+		if _, err := osexec.LookPath("nerdctl"); err != nil {
+			if _, err := osexec.LookPath("finch"); err == nil {
+				binaryName = "finch"
+			}
+		}
 	}
 	return &provider{
 		logger:     logger,


estesp:
Would love for this to be in the `v0.23.0` milestone if at all possible so that new users trying our nerdctl/finch integration don't end up with this hiccup.

pkg/cluster/internal/providers/nerdctl/provider.go
```diff
@@ -40,9 +41,25 @@ import (
...
-	// if unset, default to nerdctl
+	// if binaryName is unset, do a lookup; we may be here via a
+	// library call to provider.DetectNodeProvider(), which returns
+	// true from nerdctl.IsAvailable() by checking for both finch
+	// and nerdctl. If we don't redo the lookup here, then a finch
+	// install that triggered IsAvailable() to be true would fail
+	// to be used if we default to nerdctl when unset.
 	if binaryName == "" {
-		binaryName = "nerdctl"
+		if _, err := osexec.LookPath("nerdctl"); err != nil {
```
aojea, line 51: what do you think about this approach to remove some `if` indentantion
```diff
diff --git a/pkg/cluster/internal/providers/nerdctl/provider.go b/pkg/cluster/internal/providers/nerdctl/provider.go
index 1ed094b5..fa46a621 100644
--- a/pkg/cluster/internal/providers/nerdctl/provider.go
+++ b/pkg/cluster/internal/providers/nerdctl/provider.go
@@ -21,6 +21,7 @@ import (
        "encoding/json"
        "fmt"
        "net"
+       osexec "os/exec"
        "path/filepath"
        "strings"
 
@@ -40,9 +41,22 @@ import (
 
 // NewProvider returns a new provider based on executing `nerdctl ...`
 func NewProvider(logger log.Logger, binaryName string) providers.Provider {
-       // if unset, default to nerdctl
+       // if binaryName is unset, do a lookup; we may be here via a
+       // library call to provider.DetectNodeProvider(), which returns
+       // true from nerdctl.IsAvailable() by checking for both finch
+       // and nerdctl. If we don't redo the lookup here, then a finch
+       // install that triggered IsAvailable() to be true would fail
+       // to be used if we default to nerdctl when unset.
        if binaryName == "" {
+               // if neither finch nor nerdctl are available; we shouldn't
+               // get here as IsAvailable() returned true if we are being
+               // instantiated; default to "nerdctl" anyway
                binaryName = "nerdctl"
+               if _, err := osexec.LookPath("nerdctl"); err != nil {
+                       if _, err := osexec.LookPath("finch"); err == nil {
+                               binaryName = "finch"
+                       }
+               }
        }
        return &provider{
                logger:     logger,
```

estesp: yeah, that looks better; I'll update!

aojea:
/lgtm
/approve

Thanks

aojea:
/test pull-kind-e2e-kubernetes-1-27

unrelated test failure

PR #3563, opened 2024-04-01
add nftables to the base image

aojea:
Iptables started deprecation on some distros and nftables is the replacement, also the new kube-proxy will have a nftables backend and we are working on using them for [network policies too](https://github.com/aojea/kube-netpol/) , so we should start having them in the base image

cc: @danwinship


neolit123 (commented):
/lgtm

aojea:
/hold

I still need to build the image :)

PR #3548, opened 2024-03-13
bump containerd, cri-tools, CNI plugins, and containerd-fuse-overlayfs

AkihiroSuda:
containerd:
- https://github.com/containerd/containerd/releases/tag/v1.7.14

cri-tools:
- https://github.com/kubernetes-sigs/cri-tools/releases/tag/v1.29.0

CNI plugins:
- https://github.com/containernetworking/plugins/releases/tag/v1.4.0
- https://github.com/containernetworking/plugins/releases/tag/v1.4.1

containerd-fuse-overlayfs
- https://github.com/containerd/fuse-overlayfs-snapshotter/releases/tag/v1.0.7
- https://github.com/containerd/fuse-overlayfs-snapshotter/releases/tag/v1.0.8

PR #3564, opened 2024-04-01
bump base image

aojea:
It gets nftables plus all the security fixes of rebuilding the image


BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/3548#pullrequestreview-1971659146

Issue #3533, opened 2024-02-28
flaky test: Create multi node cluster error with nerdctl


yankay:
**What happened**:

Create multi node cluster error with nerdctl.  The error logs:
```
2024-02-28T04:45:05.0673700Z  • Preparing nodes 📦 📦 📦   ...
2024-02-28T04:45:11.6001644Z  ✗ Preparing nodes 📦 📦 📦 
2024-02-28T04:45:11.6006839Z ERROR: failed to create cluster: command "nerdctl run --name kind-worker --hostname kind-worker --label io.x-k8s.kind.role=worker --privileged --security-opt seccomp=unconfined --security-opt apparmor=unconfined --tmpfs /tmp --tmpfs /run --volume /var --volume /lib/modules:/lib/modules:ro -e KIND_EXPERIMENTAL_CONTAINERD_SNAPSHOTTER --detach --tty --label io.x-k8s.kind.cluster=kind --net kind --restart=on-failure:1 --init=false kindest/node:v1.29.2@sha256:51a1434a5397193442f0be2a297b488b6c919ce8a3931be0ce822606ea5ca245" failed with error: exit status 1
2024-02-28T04:45:11.6010402Z 
2024-02-28T04:45:11.6015593Z Command Output: time="2024-02-28T04:45:11Z" level=fatal msg="failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error running hook #0: error running hook: exit status 1, stdout: , stderr: time=\"2024-02-28T04:45:07Z\" level=fatal msg=\"failed to call cni.Setup: plugin type=\\\"firewall\\\" failed (add): running [/usr/sbin/ip6tables -t filter -N CNI-ISOLATION-STAGE-2 --wait]: exit status 4: ip6tables v1.8.7 (nf_tables):  CHAIN_USER_ADD failed (File exists): chain CNI-ISOLATION-STAGE-2\\n\"\nFailed to write to log, write /var/lib/nerdctl/1935db59/containers/default/d1ffddcc574c429168ee06332a604e9cd10349bc51813eb644a04ffc8d54ea5b/oci-hook.createRuntime.log: file already closed: unknown"
2024-02-28T04:45:11.6021322Z 
2024-02-28T04:45:11.6021601Z Stack Trace: 
2024-02-28T04:45:11.6022409Z sigs.k8s.io/kind/pkg/errors.WithStack
[...]

PR #3567, opened 2024-04-03
Fix nerdctl create multi-node cluster fail by do not create container concurrently

yankay:
Fix: #3533

The issue is caused by creating a concurrent container with the network.
Because there is no lock for the container creation, there is some conflict with the CNI creation.

And there are no test cases of nerdctl for creating containers concurrently.

### DIFF
--- a/pkg/cluster/internal/providers/nerdctl/provider.go
+++ b/pkg/cluster/internal/providers/nerdctl/provider.go
@@ -107,6 +107,8 @@ func (p *provider) Provision(status *cli.Status, cfg *config.Cluster) (err error
 	}
 
 	// actually create nodes
+	// TODO: remove once nerdctl handles concurrency better
+	// xref: https://github.com/containerd/nerdctl/issues/2908
 	for _, f := range createContainerFuncs {
 		if err := f(); err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


aojea:
> And there are no test cases of nerdctl for creating containers concurrently.

This seems an important and missing feature, can we track this in the nerdctl project add a TODO in the code to reference it?

pkg/cluster/internal/providers/nerdctl/provider.go
```diff
@@ -107,7 +107,12 @@ func (p *provider) Provision(status *cli.Status, cfg *config.Cluster) (err error
 	}
 
 	// actually create nodes
-	return errors.UntilErrorConcurrent(createContainerFuncs)
+	for _, f := range createContainerFuncs {
```
aojea, line 110:
```suggestion
        // TODO: remove once nerdctl handles concurrency better
        // xref: https://github.com/containerd/nerdctl/issues/2908
	for _, f := range createContainerFuncs {
```

aojea:
ping @yankay can you please add the comment so we can merge?

yankay:
> ping @yankay can you please add the comment so we can merge?

Thanks @aojea

aojea:
/lgtm
/approve

Thanks

PR #3573, opened 2024-04-10
fix: remove NoArgs positional arguments validation

### LINKED ISSUE #3466: If the CLI shows help because of a malformed command, but `--help` was not passed, return nonzero
Thanks for your work on kind, it's very helpful.

---

I mistyped a command: `kind clusters get`.  That's not how the CLI works.  It should have been `kind get clusters`.

Helpfully, kind showed me the help.  Unhelpfully, kind returned 0, as if everything had gone according to plan.  Later, I tried to recall this command via fzf, expecting to chose from a list of past successes.

Instead I was shown both failures and successes:

- `kind get clusters`
- `kind clusters get`

This happened because kind returned 0 for both, so my history search couldn't tell the successes from the failures.

Git behaves appropriately:

```
$ git status > /dev/null && echo good || echo bad
good
$ git schmatus 2> /dev/null && echo good || echo bad
bad
```

Kind does not:

```
$ kind get clusters > /dev/null && echo good || echo bad
good
$ kind clusters get > /dev/null && echo good || echo bad
good
```

**Environment:**

```
kind v0.20.0 go1.21.3 darwin/amd64
```

yashvardhan-kukreja:
fixes #3466 

```sh
~/Projects/GoProjects/kind issue-3466/no-args-fix                                                                                                                                                   05:26:35
❯ ./bin/kind clusters get > /dev/null && echo good || echo bad
ERROR: unknown command "clusters" for "kind"
bad

~/Projects/GoProjects/kind issue-3466/no-args-fix                                                                                                                                                   05:27:00
❯ ./bin/kind get clusters > /dev/null && echo good || echo bad
good

~/Projects/GoProjects/kind issue-3466/no-args-fix                                                                                                                                                   05:27:06
❯ ./bin/kind get clusters dfkni > /dev/null && echo good || echo bad
ERROR: unknown command "dfkni" for "kind get clusters"
bad

~/Projects/GoProjects/kind issue-3466/no-args-fix                                                                                                                                                   05:27:14
❯ ./bin/kind create cluster sds d sd> /dev/null && echo good || echo bad
ERROR: unknown command "sds" for "kind create cluster"
bad


```

### DIFF
--- a/pkg/cmd/kind/build/build.go
+++ b/pkg/cmd/kind/build/build.go
@@ -30,7 +30,6 @@ import (
 // NewCommand returns a new cobra.Command for building
 func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 	cmd := &cobra.Command{
-		Args: cobra.NoArgs,
 		// TODO(bentheelder): more detailed usage
 		Use:   "build",
 		Short: "Build one of [node-image]",
--- a/pkg/cmd/kind/completion/completion.go
+++ b/pkg/cmd/kind/completion/completion.go
@@ -32,7 +32,6 @@ import (
 // NewCommand returns a new cobra.Command for cluster creation
 func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 	cmd := &cobra.Command{
-		Args:  cobra.NoArgs,
 		Use:   "completion",
 		Short: "Output shell completion code for the specified shell (bash, zsh or fish)",
 		Long:  longDescription,
--- a/pkg/cmd/kind/create/create.go
+++ b/pkg/cmd/kind/create/create.go
@@ -30,7 +30,6 @@ import (
 // NewCommand returns a new cobra.Command for cluster creation
 func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 	cmd := &cobra.Command{
-		Args:  cobra.NoArgs,
 		Use:   "create",
 		Short: "Creates one of [cluster]",
 		Long:  "Creates one of local Kubernetes cluster (cluster)",
--- a/pkg/cmd/kind/delete/delete.go
+++ b/pkg/cmd/kind/delete/delete.go
@@ -31,7 +31,6 @@ import (
 // NewCommand returns a new cobra.Command for cluster deletion
 func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 	cmd := &cobra.Command{
-		Args: cobra.NoArgs,
 		// TODO(bentheelder): more detailed usage
 		Use:   "delete",
 		Short: "Deletes one of [cluster]",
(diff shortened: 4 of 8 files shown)


aojea:
what happen now if I pass arguments that are not going to be used?

yashvardhan-kukreja:
> what happen now if I pass arguments that are not going to be used?

Can you give a more specific example which might be going through your mind?

Meanwhile I did try providing unusable args/flag

```sh
❯ ./bin/kind get clusters -v3 --a3 -b3 > /dev/null && echo good || echo bad 
ERROR: unknown flag: --a3
bad
```

```sh
❯ kind get clusters -v3 --a3 -b3 > /dev/null && echo good || echo bad
ERROR: unknown flag: --a3
bad
```

aojea:
I'm not familiar enough with cobra, I was wondering what change on behavior is causing the removal of this option

yashvardhan-kukreja:
Sure, let me elaborate on what I found when diggin originally into this issue.

Cobra calls `Find` method here when executing the commands - [Ref](https://github.com/spf13/cobra/blob/main/command.go#L1092)

[Here](https://github.com/spf13/cobra/blob/main/command.go#L746-L748) you can see the `Find` method is ignoring the validation which we can benefit from i.e. the [legacyArgs validation](https://github.com/spf13/cobra/blob/main/args.go#L35-L37) which enforces the fact that any parent-less subcommand (i.e. wrong subcommand) is immediately returned as an error.

A more graceful way would have been to already have a `PositionalArgument` function written (like cobra.NoArgs) exactly matching our requirements but that doesn't seem to exist at the moment. Moreover, we could have chosen to write our own such function but it would've been an overkill in my opinion.


Finally, it would have been much nicer if cobra returned an error [here](https://github.com/spf13/cobra/blob/main/command.go#L1123) instead of just returning a `nil` error alongside logging out the `help` when a wrong set of command are provided.

(It almost looks like a bug/mistake to me that they added a `return cmd, nil`. I would've expected to return an error there and leave to the invoker to compare it against flag.ErrHelp to handle it on their own).

yashvardhan-kukreja:
> I was wondering what change on behavior is causing the removal of this option

The only change I noticed was the change in how things were logged when wrong subcommands were provided.

Previously, `kind` had to reach [here](https://github.com/spf13/cobra/blob/main/command.go#L1117) to eventually face an error log the information accordingly.

But after this PR, any wrong subcommands would fail much before i.e. [here](https://github.com/spf13/cobra/blob/main/command.go#L1092) only

yashvardhan-kukreja:
@aojea mind taking a look at this once you get free?

stmcginnis:
In this case, being less explicit does seem to be the better option so we get the appropriate non-zero return code.

/lgtm

BenTheElder (commented):
/lgtm
/approve
thank you!

BenTheElder:
/ok-to-test
Sorry, we've gotten on a terrible turnaround time for many non-critical bug fixes, I'm working on getting us caught back up going forward.

BenTheElder:
These failures are not related to your PR, the CI is having some issues currently (see #testing-ops in https://slack.k8s.io)
/retest

PR #3574, opened 2024-04-10
Detect /etc/containerd/certs.d automatically for mirror registry config.

Romain-Geissler-1A:
This will allow a slightly smoother transition for people who wish to already move to the new non deprecated way to configure registries in containerd and who build their own node image (or bind mount the config).

Eventually since the old way won't work, this should be the default configuration, written directly in the config.toml main file.

### DIFF
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -145,6 +145,12 @@ configure_containerd() {
       systemctl enable containerd-fuse-overlayfs
     fi
   fi
+
+  if [[ -e "/etc/containerd/certs.d" ]]; then
+    cat >> /etc/containerd/config.toml <<EOF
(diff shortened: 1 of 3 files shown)


images/base/files/usr/local/bin/entrypoint
```diff
@@ -145,6 +145,12 @@ configure_containerd() {
       systemctl enable containerd-fuse-overlayfs
     fi
   fi
+
+  if [[ -e "/etc/containerd/certs.d" ]]; then
```
BenTheElder, line 149: This is still technically a breaking change because the user could've been using this directory with the old config approach.

I'm not sure this is better than just announcing that we've changed the config and we've already encouraged users customizing the config to switch and updated our registry sample script to switch.

BenTheElder:
This will be default-enabled for future releases, thanks for the PRs.

PR #3576, opened 2024-04-10
feat(pkg/cluster/internal/providers): log container init and before

### LINKED ISSUE #3575: Log container init when max verbosity is set

Hello, I propose to log all the container logs from before init when setting the maximum verbosity supported.

**Why is this needed**:

This is extremely helpful to debug at runtime with logs from init and before, without the need to read logs afterward with `export logs`.

A lot of points can go wrong e.g. when setting up cgroups and in that case logs from Systemd are essential.
Combining them with logs from kind itself first and from Kubernetes then in a single output can be vital to better understand at runtime the initialization sequence of the containerized nodes, and possibly debug and resolve issues.

As of now only the log of the multi-user systemd target reached event is filtered, leading to ambiguous outputs like [this one](https://github.com/kubernetes-sigs/kind/issues/3423#issue-1992335390).

maxgio92:
**What it introduces**

This PR proposes to log all the container logs from before init, when setting the maximum verbosity supported.

**Why is needed**

This is extremely helpful to debug at runtime with logs from init and before, without the need to read individual logs afterward with export logs.

A lot of points can go wrong e.g. when setting up cgroups and in that case logs from Systemd are essential.
Combining them with logs from kind itself first and from Kubernetes then in a single output can be vital to better understand at runtime the initialization sequence of the containerized nodes, and possibly debug and resolve issues.

Closes #3575

### DIFF
--- a/pkg/cluster/internal/providers/common/cgroups.go
+++ b/pkg/cluster/internal/providers/common/cgroups.go
@@ -5,7 +5,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+	http://www.apache.org/licenses/LICENSE-2.0
 
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


BenTheElder:
pending agreed solution in https://github.com/kubernetes-sigs/kind/issues/3575 (which is our documented preferred approach to features per the contributor guide), thank you

Issue #3434, opened 2023-11-27
Allow  kubeProxyMode: "nftables" in the networking configuration


uablrek:
**What would you like to be added**:

The [nftables proxier](https://github.com/kubernetes/kubernetes/pull/121046) is added in "alpha" with feature-gate `NFTablesProxyMode` in K8s v1.29. When it becomes more available, it should be possible to specify:

```
networking:
  kubeProxyMode: "nftables"
```
in the kind configuration.

**Why is this needed**:

For improved UE

### Background

To try the nftables proxier before this feature, please use:
```
featureGates:
  "NFTablesProxyMode": true
nodes:
- role: control-plane
  kubeadmConfigPatches:
  - |
    kind: KubeProxyConfiguration
    mode: "nftables"
    nftables:
      masqueradeAll: false
      masqueradeBit: 14
      minSyncPeriod: 1s
      syncPeriod: 30s
```
and, use an image built on "master" or K8s v1.29 or later.

Please see also the [thread in slack](https://kubernetes.slack.com/archives/CEKK1KTN2/p1701084832395189).

aojea:
@uablrek since we are already in 1.30 , can you add the new proxy mode?

 https://github.com/kubernetes-sigs/kind/blob/40c81f187425254daf2bf84360a6257a278252df/pkg/apis/config/v1alpha4/types.go#L212-L215

aojea:
@uablrek oh wait, I read now the slack thread, I forgot the option is feature gated, despite we allow the option we also have to enable the feature gate to be able to use it ... not a nice if :/

danwinship:
You could add the enum value now and just note that it also requires setting a feature gate for now.

aojea:
> les jobs kubernetes/test-infra#31539

the experience is that people don't read and they will be opening bugs, is just because of that, for our needs we can just go with this https://github.com/kubernetes/test-infra/pull/31539

PR #3577, opened 2024-04-11
support kube-proxy nftables mode

aojea:
/hold

wait until 1.31 graduates it to beta

Fixes: #3434

### DIFF
--- a/site/content/docs/user/configuration.md
+++ b/site/content/docs/user/configuration.md
@@ -217,14 +217,14 @@ networking:
 
 #### kube-proxy mode
 
-You can configure the kube-proxy mode that will be used, between iptables and ipvs. By
-default iptables is used
+You can configure the kube-proxy mode that will be used, between iptables, ipvs and nftables.
+By default iptables is used
 
 {{< codeFromInline lang="yaml" >}}
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 networking:
-  kubeProxyMode: "ipvs"
+  kubeProxyMode: "nftables"
 {{< /codeFromInline >}}
 
 To disable kube-proxy, set the mode to `"none"`.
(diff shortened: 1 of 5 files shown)


aojea:
@BenTheElder the problem with this option is to keep compatibility with old version, as it only will work with new versions ... I can't remember if we have some precedent like this

BenTheElder:
> @BenTheElder the problem with this option is to keep compatibility with old version, as it only will work with new versions ... I can't remember if we have some precedent like this

This is true with feature gates, not exactly the same as those are totally user supplied strings.

site/content/docs/user/configuration.md
```diff
@@ -217,14 +217,14 @@ networking:
 
 #### kube-proxy mode
 
-You can configure the kube-proxy mode that will be used, between iptables and ipvs. By
-default iptables is used
+You can configure the kube-proxy mode that will be used, between iptables, ipvs and nftables.
```
BenTheElder, line 220:
```suggestion
You can configure the kube-proxy mode that will be used, between iptables, nftables (Kubernetes v1.30+), and ipvs.
```

BenTheElder (commented):
Actually, we will need to internally version-check at cluster config time anyhow and enable NFTablesProxyMode feature gate? Or do we expect the user to do this?

It's new that we have our own config option for an alpha feature.

danwinship:
> It's new that we have our own config option for an alpha feature.

no, you missed:

> wait until 1.31 graduates it to beta

BenTheElder:
Sorry, yes, SGTM.

I think we should just document that it's 1.31+, we allow clusters with mixed versions and only advanced users will be setting this field anyhow.

aojea:
/hold cancel

K/K PR is already merged and this will allow to set up CI without having to use additional scripting https://github.com/kubernetes/kubernetes/pull/124383

PR #3582, opened 2024-04-16
e2e-k8s.sh: support --ginkgo.label-filter

pohly:
The label filter query is more expressive (logical operations) and readable (no regexp unless absolutely required). Such a query can be combined with focus + skip, but in practice a single label filter can replace both of those and is easier to understand.

Kubernetes has supported ginkgo v2 and thus --label-filter since v1.25.0. This makes it safe to pass that command line flag unconditionally when invoking the E2E suite. However, actual labels were only added in Kubernetes 1.29.

Note: I have not tested this yet. If we are confident that it'll work, then we can merge and I'll start using it in a small job (most likely DRA) once it is available.

### DIFF
--- a/hack/ci/e2e-k8s.sh
+++ b/hack/ci/e2e-k8s.sh
@@ -22,6 +22,12 @@ set -o errexit -o nounset -o xtrace
 # Settings:
 # SKIP: ginkgo skip regex
 # FOCUS: ginkgo focus regex
+# LABEL_FILTER: ginkgo label query for selecting tests (see "Spec Labels" in https://onsi.github.io/ginkgo/#filtering-specs)
+#
+# The default is to focus on conformance tests. Serial tests get skipped when
+# parallel testing is enabled. Using LABEL_FILTER instead of combining SKIP and
+# FOCUS is recommended (more expressive, easier to read than regexp).
+#
 # GA_ONLY: true  - limit to GA APIs/features as much as possible
 #          false - (default) APIs and features left at defaults
 # FEATURE_GATES:
@@ -237,9 +243,13 @@ run_tests() {
     printf '%s' "${fixed_coredns}" | kubectl apply -f -
   fi
 
-  # ginkgo regexes
+  # ginkgo regexes and label filter
   SKIP="${SKIP:-}"
-  FOCUS="${FOCUS:-"\\[Conformance\\]"}"
+  FOCUS="${FOCUS:-}"
+  LABEL_FILTER="${LABEL_FILTER:-}"
+  if [ -z "${FOCUS}" ] && [ -z "${LABEL_FILTER}" ]; then
+    FOCUS="\\[Conformance\\]"
+  fi
   # if we set PARALLEL=true, skip serial tests set --ginkgo-parallel
   if [ "${PARALLEL:-false}" = "true" ]; then
     export GINKGO_PARALLEL=y
@@ -262,7 +272,7 @@ run_tests() {
   # interrupt
   ./hack/ginkgo-e2e.sh \
     '--provider=skeleton' "--num-nodes=${NUM_NODES}" \
-    "--ginkgo.focus=${FOCUS}" "--ginkgo.skip=${SKIP}" \
+    "--ginkgo.focus=${FOCUS}" "--ginkgo.skip=${SKIP}" "--ginkgo.label-filter=${LABEL_FILTER}" \
     "--report-dir=${ARTIFACTS}" '--disable-log-dump=true' &
   GINKGO_PID=$!
   wait "$GINKGO_PID"


pohly:
Is my understanding correct that this change will be usable right after merging, i.e. without a formal kind release?

aojea:
> Is my understanding correct that this change will be usable right after merging, i.e. without a formal kind release?

yes

@pohly @onsi this output looks weird, `Will run 2144 of 7407 specs`  but it `Ran 875 of 7407 Specs `

```
/home/prow/go/src/k8s.io/kubernetes/_output/bin/ginkgo --nodes=25 --poll-progress-after=60m --poll-progress-interval=5m --source-root=/home/prow/go/src/k8s.io/kubernetes/cluster/.. --timeout=24h --flake-attempts=1 --no-color /home/prow/go/src/k8s.io/kubernetes/_output/bin/e2e.test -- --kubeconfig=/root/.kube/kind-test-config --host=https://127.0.0.1:41971/ --provider=skeleton --gce-project= --gce-zone= --gce-region= --gce-multizone=false --gke-cluster= --kube-master= --cluster-tag= --cloud-config-file= --repo-root=/home/prow/go/src/k8s.io/kubernetes/cluster/.. --node-instance-group= --prefix=e2e --network=e2e --node-tag= --master-tag= --docker-config-file= --dns-domain=cluster.local --prepull-images=false --report-complete-ginkgo --report-complete-junit --provider=skeleton --num-nodes=2 --ginkgo.focus=. '--ginkgo.skip=\[Serial\]|\[Slow\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|PodSecurityPolicy|LoadBalancer|load.balancer|Simple.pod.should.support.exec.through.an.HTTP.proxy|subPath.should.support.existing' --ginkgo.label-filter= --report-dir=/logs/artifacts --disable-log-dump=true
Running Suite: Kubernetes e2e suite - /home/prow/go/src/k8s.io/kubernetes/_output/bin
=====================================================================================
Random Seed: 1713281512 - will randomize all specs
Will run 2144 of 7407 specs
Running in parallel across 25 processes
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS
... snipped ...
Ran 875 of 7407 Specs in 924.114 seconds
SUCCESS! -- 875 Passed | 0 Failed | 0 Pending | 6532 Skipped
```

hack/ci/e2e-k8s.sh
```diff
@@ -262,7 +272,7 @@ run_tests() {
   # interrupt
   ./hack/ginkgo-e2e.sh \
     '--provider=skeleton' "--num-nodes=${NUM_NODES}" \
-    "--ginkgo.focus=${FOCUS}" "--ginkgo.skip=${SKIP}" \
+    "--ginkgo.focus=${FOCUS}" "--ginkgo.skip=${SKIP}" "--ginkgo.label-filter=${LABEL_FILTER}" \
```
aojea, line 275: should we avoid to add the flag ` "--ginkgo.label-filter=${LABEL_FILTER}" \` if LABEL_FILTER is empty?

onsi:
@aojea i don’t know if y’all also call `Skip()` in some specs at runtime.  That would explain the discrepancy.

pohly:
We do a lot of runtime skipping. I think that explains it.

hack/ci/e2e-k8s.sh
pohly: If it's empty, then it doesn't make a difference. I didn't want to make the shell code more complicated by adding further if checks.

aojea: fair

hack/ci/e2e-k8s.sh
```diff
@@ -237,9 +243,13 @@ run_tests() {
     printf '%s' "${fixed_coredns}" | kubectl apply -f -
   fi
 
-  # ginkgo regexes
+  # ginkgo regexes and label filter
   SKIP="${SKIP:-}"
-  FOCUS="${FOCUS:-"\\[Conformance\\]"}"
+  FOCUS="${FOCUS:-}"
+  LABEL_FILTER="${LABEL_FILTER:-}"
+  if [ -z "${SKIP}" ] && [ -z "${FOCUS}" ] && [ -z "${LABEL_FILTER}" ]; then
```
aojea, line 250: SKIP was not influencing the defaulting of Focus before, to be backwards compatible should not this be ?
```suggestion
  if  [ -z "${FOCUS}" ] && [ -z "${LABEL_FILTER}" ]; then
```

aojea:
one last comment and we are good to go  https://github.com/kubernetes-sigs/kind/pull/3582/files#r1635882841

hack/ci/e2e-k8s.sh
pohly: Yes, I think you are right. Changed.

As you were asking about SKIP, I also looked at the surrounding lines a bit more and started to wonder whether we should still use FOCUS and SKIP at all as defaults. We could get the same effect with the new label filter.

I decided against it (again?) because it's better to be conservative and to keep using what we used before. It also minimizes the changes to the script.

But I could also be convinced to be more audacious and clean this up so that FOCUS and SKIP are only set when explicitly requested. That would expedite our transition to using labels. :shrug:

aojea: > That would expedite our transition to using labels.

better do baby steps, so we can identify clearly what change do what

aojea:
we should also think in phasing this:
1.31 start the transition to labels and deprecation focus and skip , warn on test-infra to use filters

1.32 phase out and forbid focus and skip in test-infra with a. linter

PR #3584, opened 2024-04-21
use cloud-provider-kind as recommended loadbalancer

aojea:
We are very thankful to metallb but with cloud-provider-kind we are able to run load balancers in multiple platforms and also to provider a more "cloud provider" experience, so we'll use it as the recommended solution for Services of type Load balancer

/kind documentation
/assign @BenTheElder


site/content/docs/user/loadbalancer.md
```diff
@@ -6,68 +6,29 @@ menu:
...
 
-Wait until the MetalLB pods (controller and speakers) are ready:
+Cloud Provider KIND can be installed using golang
 
 {{< codeFromInline lang="bash" >}}
-kubectl wait --namespace metallb-system \
-                --for=condition=ready pod \
-                --selector=app=metallb \
-                --timeout=90s
+go install sigs.k8s.io/cloud-provider-kind@latest
```
stmcginnis, line 26: Are we publishing anywhere that we can use kubectl instead of `go install`? I suppose the target audience for `kind` will have `go` available, but it would be nice to not have to have the whole go toolchain installed to use Cloud Provider Kind.

The use of `latest` gives me a little concern too, just because we try to discourage the use of latest elsewhere when installing things and prefer to be more explicit.

site/content/docs/user/loadbalancer.md
```diff
@@ -6,68 +6,29 @@ menu:
...
 {{< /codeFromInline >}}
 
-### Setup address pool used by loadbalancers
+or downlading one of the [released binaries](https://github.com/kubernetes-sigs/cloud-provider-kind/releases).
```
stmcginnis, line 29: Sorry, I guess I should read through first before commenting. This addresses most of my concern.

site/content/docs/user/loadbalancer.md
```diff
@@ -6,68 +6,29 @@ menu:
...
-```yaml
-{{% readFile "static/examples/loadbalancer/metallb-config.yaml" %}}
-```
-
-Apply the contents
-
-{{< codeFromInline lang="bash" >}}
-kubectl apply -f https://kind.sigs.k8s.io/examples/loadbalancer/metallb-config.yaml
-{{< /codeFromInline >}}
+Cloud Provider KIND runs as a standalone binary in your host and connects to your KIND cluster and provision new Load Balancer containers for your Services. It requires privileges to open ports on the system and to connect to the container runtime.
```
stmcginnis, line 31: Small typo:

```suggestion
Cloud Provider KIND runs as a standalone binary in your host and connects to your KIND cluster and provisions new Load Balancer containers for your Services. It requires privileges to open ports on the system and to connect to the container runtime.
```

site/content/docs/user/loadbalancer.md
aojea: there is a typo :) 
```suggestion
or downloading one of the [released binaries](https://github.com/kubernetes-sigs/cloud-provider-kind/releases).
```

site/content/docs/user/loadbalancer.md
aojea: latest in `go install` means the latest released version https://pkg.go.dev/cmd/go#hdr-Compile_and_install_packages_and_dependencies , is not like in containers that is unspecified

BenTheElder: yeah this is confusing versus docker, go consistently means "highest release tag" when using the latest placeholder, which should be fine here. We didn't pin to a particular metallb version in the docs.

users should absolutely consider pinning a specific version but i'm not sure we want to do that here, we want to maintain compatibility with kind and that should be fairly easy since it only minimally interacts (famous last words) and mostly talks to docker/podman and the api server

aojea: We are testing kubernetes HEAD with cloud-provider-kind HEAD so this combination MUST work or we'll have a bigger problem:)

PR #3585, opened 2024-04-21
fix docs

aojea:
follow up on previous PR since I got wrong the title and the comments


stmcginnis:
Shoot, I missed that too with the diff display. Glad you caught it so quickly.

/lgtm

PR #3595, opened 2024-04-29
Remove redundant default type value from flag doc

timofurrer:
This change set removes the redundant mention of the `--type` flag default value of the `kind build node-image` command. This basically changes the help message from:

```
--type string         build type, default is docker (default "docker")
```

to:

```
--type string         build type (default "docker")
```


aojea:
/approve

Thanks

PR #3596, opened 2024-04-29
Add github workflow for approve to run github action.

liangyuanpeng:
I have see that this PR have `approve` & `lgtm` & `ok-to-test` and pending for github action approval. https://github.com/kubernetes-sigs/kind/pull/3595

> 4 workflows awaiting approval
This workflow requires approval from a maintainer. [Learn more about approving workflows.](https://docs.github.com/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks)
9 pending, 2 neutral, and 6 successful checks

So i open this PR to add a github workflow for approve to run github action if PR have label of `ok-to-test`, Speed up the PR process.

Basically, this is copied from https://github.com/etcd-io/etcd/blob/main/.github/workflows/gh-workflow-approve.yaml

### DIFF
--- a/.github/workflows/gh-workflow-approve.yaml
+++ b/.github/workflows/gh-workflow-approve.yaml
@@ -1,0 +1,40 @@
+---
+name: Approve GitHub Workflows
+
+on:
+  pull_request_target:
+    types:
+      - labeled
+      - synchronize
+
+jobs:
+  approve:
+    name: Approve ok-to-test
+    if: contains(github.event.pull_request.labels.*.name, 'ok-to-test')
+    runs-on: ubuntu-latest
+    permissions:
+      actions: write
+    steps:
+      - name: Update PR
+        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+        continue-on-error: true
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          debug: ${{ secrets.ACTIONS_RUNNER_DEBUG == 'true' }}
+          script: |
+            const result = await github.rest.actions.listWorkflowRunsForRepo({
+              owner: context.repo.owner,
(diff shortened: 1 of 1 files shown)


liangyuanpeng:
feel free to close it if we don't want it.

stmcginnis:
I don't believe we want to automatically approve workflow runs. They are held until a member is able to review and decide to allow it for a reason. You can read more about the process here: https://www.kubernetes.dev/docs/guide/pull-requests/#the-testing-and-merge-workflow

aojea:
I tend to agree with Sean, most of the time to merge is on reviews

.github/workflows/gh-workflow-approve.yaml
```diff
@@ -0,0 +1,40 @@
...
+      - labeled
+      - synchronize
+
+jobs:
+  approve:
+    name: Approve ok-to-test
+    if: contains(github.event.pull_request.labels.*.name, 'ok-to-test')
+    runs-on: ubuntu-latest
+    permissions:
+      actions: write
```
BenTheElder, line 16: this makes me very nervous. github PR workflows have a history of privilege escalation issues

BenTheElder:
> I don't believe we want to automatically approve workflow runs. They are held until a member is able to review and decide to allow it for a reason. 

This PR is running them if `ok-to-test` label exists?

liangyuanpeng:
> I tend to agree with Sean, most of the time to merge is on reviews

So this just let github action to run when PR have `ok-to-test`,  and PR merging depends on lgtm and approve, which is not the problem this PR hopes to solve.

.github/workflows/gh-workflow-approve.yaml
liangyuanpeng: OK, let me close it, 

The best approach is probably still to implement it on the prow side.

PR #3599, opened 2024-05-01
fix kind version unit test failures for released versions

hp685:
Version() string assumes that preRelease version is set for commit hash and count to be taken into account. Update the unit tests to pass in a pre release version per test case

### DIFF
--- a/pkg/cmd/kind/version/version.go
+++ b/pkg/cmd/kind/version/version.go
@@ -58,7 +58,7 @@ const versionCore = "0.23.0"
 
 // versionPreRelease is the base pre-release portion of the kind CLI version per
 // Semantic Versioning 2.0.0
-const versionPreRelease = "alpha"
+var versionPreRelease = "alpha"
 
 // gitCommitCount count the commits since the last release.
 // It is injected at build time.
--- a/pkg/cmd/kind/version/version_test.go
+++ b/pkg/cmd/kind/version/version_test.go
@@ -63,34 +63,53 @@ func TestTruncate(t *testing.T) {
 
 func TestVersion(t *testing.T) {
 	tests := []struct {
-		name           string
-		gitCommit      string
-		gitCommitCount string
-		want           string
+		name              string
+		versionPreRelease string
+		gitCommit         string
+		gitCommitCount    string
+		want              string
 	}{
 		{
-			name:           "With git commit count and with commit hash",
-			gitCommit:      "mocked-hash",
-			gitCommitCount: "mocked-count",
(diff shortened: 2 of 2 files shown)


BenTheElder:
> fix kind version unit test failures on non-main branches

We don't have or support any branch other than main ...?

Can you explain in more detail what you're attempting to fix?

pkg/cmd/kind/version/version.go
```diff
@@ -58,7 +58,7 @@ const versionCore = "0.23.0"
 
 // versionPreRelease is the base pre-release portion of the kind CLI version per
 // Semantic Versioning 2.0.0
-const versionPreRelease = "alpha"
+var versionPreRelease = "alpha"
```
BenTheElder, line 61: .Not a fan of mutating global vars to implement unit tests

hp685:
> > fix kind version unit test failures on non-main branches
> 
> We don't have or support any branch other than main ...?
> 
> Can you explain in more detail what you're attempting to fix?

I can update the description. Here's what I meant:

the version unit tests fail if you we the test for any released artifacts since the versionPreRelease="" after its released. 
Any downstreams that rely on building and testing a released version would encounter this failure

pkg/cmd/kind/version/version.go
hp685: Alternatively, we can skip this test if versionPreRelease == "". 
Open to other suggestions

BenTheElder: re-reading this, I see we've already done that for other fields.

What we should really do is make Version() a shim to version() passing the globals and test `version` for invariants, but we can leave that to a follow-up.

BenTheElder: (since we already had this global mutating pattern prior to this PR)

BenTheElder:
Ah -- That's an interesting gap. Our release process creates a new pre-release commit immediately following the release tag commit so we're never testing that commit alone.

BenTheElder (commented):
/lgtm
/approve

pkg/cmd/kind/version/version.go
hp685: SG, thanks for the approval. I can look into your suggestion of making Version() a shim in a follow up.

hp685:
Looking at the logs for the e2e jobs the timeout failure doesn't seem relevant to this change

BenTheElder:
correct

pkg/cmd/kind/version/version.go
BenTheElder: filed a follow-up in https://github.com/kubernetes-sigs/kind/pull/3856 just for reference

PR #3600, opened 2024-05-02
Bump containerd to v1.7.16

abhay-krishna:
This PR bumps the containerd version in the Kind base image to the latest upstream release (version `v1.7.16`).

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -122,7 +122,7 @@ RUN eval "$(gimme "${GO_VERSION}")" \
 # stage for building containerd
 FROM go-build as build-containerd
 ARG TARGETARCH GO_VERSION
-ARG CONTAINERD_VERSION="v1.7.15"
+ARG CONTAINERD_VERSION="v1.7.16"
 ARG CONTAINERD_CLONE_URL="https://github.com/containerd/containerd"
 # we don't build with optional snapshotters, we never select any of these
 # they're not ideal inside kind anyhow, and we save some disk space


images/base/Dockerfile
```diff
@@ -122,7 +122,7 @@ RUN eval "$(gimme "${GO_VERSION}")" \
 # stage for building containerd
 FROM go-build as build-containerd
 ARG TARGETARCH GO_VERSION
-ARG CONTAINERD_VERSION="v1.7.15"
+ARG CONTAINERD_VERSION="v1.7.16"
```
BenTheElder, line 125: any notable changes to be aware of?

we also usually see runc changed at the same time

abhay-krishna: I didn't see any breaking changes, just bug fixes and a CVE fix.

runc has a new RC tag [`v1.2.0-rc.1`](https://github.com/opencontainers/runc/releases/tag/v1.2.0-rc.1) but this is a minor version upgrade from what Kind uses currently, which is `v1.1.12`, so I was wary about upgrading to that one.

BenTheElder:
[I forgot to submit my comment here, lost in all the tabs, sorry!]

Let's hold this until after the next release, there don't seem to be any super pressing bug fixes, we've had time to soak the current version in CI, and we're ~overdue for a release.

I think we'll get a release in this week and then shortly after double back and get this in.

Thanks for the PR and your patience!

abhay-krishna:
> [I forgot to submit my comment here, lost in all the tabs, sorry!]
> 
> Let's hold this until after the next release, there don't seem to be any super pressing bug fixes, we've had time to soak the current version in CI, and we're ~overdue for a release.
> 
> I think we'll get a release in this week and then shortly after double back and get this in.
> 
> Thanks for the PR and your patience!

Thanks for the response! Sounds good to me, I'll keep an eye on this PR and rebase if necessary.

PR #3607, opened 2024-05-13
upgrade to go1.21.10

BenTheElder:
note: still not switching to 1.22 because of [runc issue 4233](https://github.com/opencontainers/runc/issues/4233)


BenTheElder:
[merging security bump ahead of v0.23, this is not affecting k/k e2es]

Issue #3609, opened 2024-05-13
Mass CI Failures for NFS Tests


BenTheElder:
This has been plaguing us recently and slowing down testing etc, did some digging finally and ...

Pretty sure this is: https://github.com/kubernetes/kubernetes/pull/123423#issuecomment-2108949914

/sig storage
/sig testing
/priority important-soon
/triage accepted
/assign

BenTheElder:
This will be fixed by test cherry-picks, see this thread https://github.com/kubernetes/kubernetes/pull/123423#issuecomment-2109144815

BenTheElder:
These are:
https://github.com/kubernetes/kubernetes/pull/124858
https://github.com/kubernetes/kubernetes/pull/124856
https://github.com/kubernetes/kubernetes/pull/124857

BenTheElder:
Nudged #release-management to PTAL at merging the cherry picks.

BenTheElder:
this should be fixed :-)

PR #3614, opened 2024-05-14
Add a new directory/remote builder - new modes to specify k8s binaries to build node-images

dims:
Related to https://github.com/kubernetes-sigs/kind/issues/381

Enabling the following command lines to build node image

1. Specify a tar gzipped file from a local disk or from an url
```
kind build node-image $HOME/Downloads/kubernetes-server-linux-amd64.tar.gz
kind build node-image --type url https://dl.k8s.io/v1.30.0/kubernetes-server-linux-arm64.tar.gz
```
This is aimed at someone who may have a different build system and do not use the k8s make file targets, but would love to build a node image for kind with their own patches and build systems

2.
```
kind build node-image v1.30.0
```
This is aimed at the majority of the current requests on slack and other avenues who just want to play with a version we haven't created a node image for (but has an upstream k8s tag and release artifacts)

Please take a look!

Here are the list of files and directory structure needed by the new builder in the tar gzipped file (including the `kubernetes` directory. which should be at the root of the passed in root directory)
```
kubernetes
└── server
    └── bin
        ├── kube-apiserver.docker_tag
        ├── kube-apiserver.tar
        ├── kube-controller-manager.tar
        ├── kube-proxy.tar
        ├── kube-scheduler.tar
        ├── kubeadm
        ├── kubectl
        └── kubelet
```

With this PR, here are some of the possibilities:
```
kind build node-image https://dl.k8s.io/v1.30.0/kubernetes-server-linux-arm64.tar.gz
kind build node-image $HOME/Downloads/kubernetes-server-linux-amd64.tar.gz
kind build node-image v1.30.0
kind build node-image $HOME/go/src/k8s.io/kubernetes/

kind build node-image --type url https://dl.k8s.io/v1.30.0/kubernetes-server-linux-arm64.tar.gz
kind build node-image --type file $HOME/Downloads/kubernetes-server-linux-amd64.tar.gz
kind build node-image --type release v1.30.0
kind build node-image --type source $HOME/go/src/k8s.io/kubernetes/

kind build node-image
```

### DIFF
--- a/pkg/build/nodeimage/build.go
+++ b/pkg/build/nodeimage/build.go
@@ -17,10 +17,14 @@ limitations under the License.
 package nodeimage
 
 import (
+	"fmt"
+	"net/url"
+	"os"
 	"runtime"
 
 	"sigs.k8s.io/kind/pkg/build/nodeimage/internal/kube"
 	"sigs.k8s.io/kind/pkg/errors"
+	"sigs.k8s.io/kind/pkg/internal/version"
 	"sigs.k8s.io/kind/pkg/log"
 )
 
@@ -46,24 +50,98 @@ func Build(options ...Option) error {
 		ctx.logger.Warnf("unsupported architecture %q", ctx.arch)
 	}
 
-	// locate sources if no kubernetes source was specified
-	if ctx.kubeRoot == "" {
-		kubeRoot, err := kube.FindSource()
-		if err != nil {
-			return errors.Wrap(err, "error finding kuberoot")
+	if ctx.buildType == "" {
+		ctx.buildType = detectBuildType(ctx.kubeParam)
+		if ctx.buildType != "" {
+			ctx.logger.V(0).Infof("Detected build type: %q", ctx.buildType)
 		}
-		ctx.kubeRoot = kubeRoot
 	}
 
-	// initialize bits
-	builder, err := kube.NewDockerBuilder(ctx.logger, ctx.kubeRoot, ctx.arch)
-	if err != nil {
-		return err
+	if ctx.buildType == "url" {
+		ctx.logger.V(0).Infof("Building using URL: %q", ctx.kubeParam)
+		builder, err := kube.NewURLBuilder(ctx.logger, ctx.kubeParam)
+		if err != nil {
+			return err
+		}
+		ctx.builder = builder
 	}
-	ctx.builder = builder
+
+	if ctx.buildType == "file" {
+		ctx.logger.V(0).Infof("Building using local file: %q", ctx.kubeParam)
+		if info, err := os.Stat(ctx.kubeParam); err == nil && info.Mode().IsRegular() {
+			builder, err := kube.NewTarballBuilder(ctx.logger, ctx.kubeParam)
+			if err != nil {
+				return err
+			}
+			ctx.builder = builder
+		}
+	}
+
+	if ctx.buildType == "release" {
+		ctx.logger.V(0).Infof("Building using release %q artifacts", ctx.kubeParam)
+		kubever, err := version.ParseSemantic(ctx.kubeParam)
+		if err == nil {
+			builder, err := kube.NewReleaseBuilder(ctx.logger, "v"+kubever.String(), ctx.arch)
+			if err != nil {
+				return err
+			}
+			ctx.builder = builder
+		} else {
+			if _, err := os.Stat(ctx.kubeParam); err != nil {
+				ctx.logger.V(0).Infof("%s is not a valid kubernetes version", ctx.kubeParam)
+				return fmt.Errorf("%s is not a valid kubernetes version", ctx.kubeParam)
+			}
+		}
+	}
+
+	if ctx.builder == nil {
+		// locate sources if no kubernetes source was specified
+		if ctx.kubeParam == "" {
+			kubeRoot, err := kube.FindSource()
+			if err != nil {
+				return errors.Wrap(err, "error finding kuberoot")
+			}
+			ctx.kubeParam = kubeRoot
+		}
+		ctx.logger.V(0).Infof("Building using source: %q", ctx.kubeParam)
+
+		// initialize bits
+		builder, err := kube.NewDockerBuilder(ctx.logger, ctx.kubeParam, ctx.arch)
+		if err != nil {
+			return err
+		}
+		ctx.builder = builder
+	}
 
 	// do the actual build
 	return ctx.Build()
+}
+
+// detectBuildType detect the type of build required based on the param passed in the following order
+// url: if the param is a valid http or https url
+// file: if the param refers to an existing regular file
+// source: if the param refers to an existing directory
+// release: if the param is a semantic version expression (does this require the v preprended?
+func detectBuildType(param string) string {
+	u, err := url.ParseRequestURI(param)
+	if err == nil {
+		if u.Scheme == "http" || u.Scheme == "https" {
+			return "url"
+		}
+	}
+	if info, err := os.Stat(param); err == nil {
+		if info.Mode().IsRegular() {
+			return "file"
+		}
+		if info.Mode().IsDir() {
+			return "source"
+		}
+	}
+	_, err = version.ParseSemantic(param)
+	if err == nil {
+		return "release"
+	}
+	return ""
 }
 
 func supportedArch(arch string) bool {
--- a/pkg/build/nodeimage/internal/kube/builder_remote.go
+++ b/pkg/build/nodeimage/internal/kube/builder_remote.go
@@ -1,0 +1,156 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kube
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"sigs.k8s.io/kind/pkg/log"
+)
+
+type remoteBuilder struct {
+	version string
+	logger  log.Logger
+	url     string
+}
+
+var _ Builder = &remoteBuilder{}
+
+// NewURLBuilder used to specify a complete url to a gzipped tarball
+func NewURLBuilder(logger log.Logger, url string) (Builder, error) {
+	return &remoteBuilder{
+		version: "",
+		logger:  logger,
+		url:     url,
+	}, nil
+}
+
+// NewReleaseBuilder used to specify a release semver and constructs a url to release artifacts
+func NewReleaseBuilder(logger log.Logger, version, arch string) (Builder, error) {
+	url := "https://dl.k8s.io/" + version + "/kubernetes-server-linux-" + arch + ".tar.gz"
+	return &remoteBuilder{
+		version: version,
+		logger:  logger,
+		url:     url,
+	}, nil
+}
+
+// Build implements Bits.Build
+func (b *remoteBuilder) Build() (Bits, error) {
+
+	tmpDir, err := os.MkdirTemp(os.TempDir(), "k8s-tar-extract-")
+	if err != nil {
+		return nil, fmt.Errorf("error creating temporary directory for tar extraction: %w", err)
+	}
+
+	tgzFile := filepath.Join(tmpDir, "kubernetes-"+b.version+"-server-linux-amd64.tar.gz")
+	err = b.downloadURL(b.url, tgzFile)
+	if err != nil {
+		return nil, fmt.Errorf("error downloading file: %w", err)
+	}
+
+	err = extractTarball(tgzFile, tmpDir, b.logger)
+	if err != nil {
+		return nil, fmt.Errorf("error extracting tgz file: %w", err)
+	}
+
+	binDir := filepath.Join(tmpDir, "kubernetes/server/bin")
+	contents, err := os.ReadFile(filepath.Join(binDir, "kube-apiserver.docker_tag"))
+	if err != nil {
+		return nil, err
+	}
+	sourceVersionRaw := strings.TrimSpace(string(contents))
+	return &bits{
+		binaryPaths: []string{
+			filepath.Join(binDir, "kubeadm"),
+			filepath.Join(binDir, "kubelet"),
+			filepath.Join(binDir, "kubectl"),
+		},
+		imagePaths: []string{
+			filepath.Join(binDir, "kube-apiserver.tar"),
+			filepath.Join(binDir, "kube-controller-manager.tar"),
+			filepath.Join(binDir, "kube-scheduler.tar"),
+			filepath.Join(binDir, "kube-proxy.tar"),
+		},
+		version: sourceVersionRaw,
+	}, nil
+}
+
+func (b *remoteBuilder) downloadURL(url string, destPath string) error {
+	output, err := os.Create(destPath)
+	if err != nil {
+		return fmt.Errorf("error creating file for download %q: %v", destPath, err)
+	}
+	defer output.Close()
+
+	b.logger.V(0).Infof("Downloading %q", url)
+
+	// Create a client with custom timeouts
+	// to avoid idle downloads to hang the program
+	httpClient := &http.Client{
+		Transport: &http.Transport{
+			Proxy: http.ProxyFromEnvironment,
+			DialContext: (&net.Dialer{
+				Timeout:   30 * time.Second,
+				KeepAlive: 30 * time.Second,
+			}).DialContext,
+			TLSHandshakeTimeout:   10 * time.Second,
+			ResponseHeaderTimeout: 30 * time.Second,
+			IdleConnTimeout:       30 * time.Second,
+		},
+	}
+
+	// this will stop slow downloads after 10 minutes
+	// and interrupt reading of the Response.Body
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
+	defer cancel()
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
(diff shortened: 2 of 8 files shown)


dims:
built an image using:
```
kind build node-image --type release v1.30.0
```

and looks like that does build the node image properly
```
$ kind create cluster --image kindest/node:latest
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:latest) 🖼
 ✓ Preparing nodes 📦
 ✓ Writing configuration 📜
 ✓ Starting control-plane 🕹️
 ✓ Installing CNI 🔌
 ✓ Installing StorageClass 💾

Have a nice day! 👋

$ kubectl cluster-info --context kind-kind
Kubernetes control plane is running at https://127.0.0.1:35929
CoreDNS is running at https://127.0.0.1:35929/api/v1/namespaces/kube-system/services/kube-dns:dns/proxy

To further debug and diagnose cluster problems, use 'kubectl cluster-info dump'.
```

pkg/build/nodeimage/internal/kube/builder_remote.go
```diff
@@ -0,0 +1,201 @@
...
+func NewRemoteBuilder(logger log.Logger, version string) (Builder, error) {
+	return &remoteBuilder{
+		version: version,
+		logger:  logger,
+	}, nil
+}
+
+// Build implements Bits.Build
+func (b *remoteBuilder) Build() (Bits, error) {
+	url := "https://dl.k8s.io/" + b.version + "/kubernetes-server-linux-amd64.tar.gz"
```
aojea, line 51: if we make this parmetizable as default `https://dl.k8s.io/"` but that people can define their own url or `file:://home/aojea/mybuild` we don't need the directory, people with custom builds just have to create the tarball as we expect (with current approach it has to do it anyway, as it has to have the same directory layout

pkg/build/nodeimage/build.go
```diff
@@ -46,21 +47,43 @@ func Build(options ...Option) error {
 		ctx.logger.Warnf("unsupported architecture %q", ctx.arch)
 	}
 
-	// locate sources if no kubernetes source was specified
-	if ctx.kubeRoot == "" {
-		kubeRoot, err := kube.FindSource()
+	if ctx.buildType == "directory" {
```
aojea, line 50: I think we can omit this if `release` accepts a parameter that allows to override the default url

`https://dl.k8s.io` can be overriden via `file://local`

pkg/build/nodeimage/internal/kube/builder_remote.go
dims: i am holding off on the parameterizable url and dedicating this to JUST our released artifacts as I did add another option to use tgz from file on disk.

pkg/build/nodeimage/build.go
dims: i removed buildType totally from the PR. just 2 things now ... either a semver or a path-to-a-tgz-file.

pkg/build/nodeimage/build.go
```diff
@@ -17,7 +17,13 @@ limitations under the License.
 package nodeimage
 
 import (
+	"fmt"
+	"os"
+	"regexp"
 	"runtime"
+	"strconv"
+
+	"github.com/blang/semver/v4"
```
aojea, line 26: @BenTheElder forked it from apimachinery to avoid dependencies `pkg/internal/version/version.go`

go.mod
```diff
@@ -13,6 +13,7 @@ go 1.17
 require (
 	github.com/BurntSushi/toml v1.0.0
 	github.com/alessio/shellescape v1.4.1
+	github.com/blang/semver/v4 v4.0.0
```
aojea, line 16: I think we can avoid it https://github.com/kubernetes-sigs/kind/pull/3614/files#r1601717270

pkg/cmd/kind/build/nodeimage/nodeimage.go
```diff
@@ -97,7 +94,7 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
 	if err := nodeimage.Build(
 		nodeimage.WithImage(flags.Image),
 		nodeimage.WithBaseImage(flags.BaseImage),
-		nodeimage.WithKuberoot(kubeRoot),
+		nodeimage.WithKubeParam(kubeRoot),
```
aojea, line 97: @BenTheElder it seems with this approach we infer from the input if we should: build from source, fetch from url or extract from tarball

Is it too magical? should we recover type ?
At least we should document it , the `Use: ` ....

better wait for @BenTheElder he is good with these cli UX things

dims: definitely! no rush.

go.mod
dims: Done!

pkg/cmd/kind/build/nodeimage/nodeimage.go
BenTheElder: > it seems with this approach we infer from the input if we should: build from source, fetch from url or extract from tarball

I think we should infer it and allow the user to override the inference.

We can say:
- empty input == source build and autodetect
- starts with a drive letter, `/`, or `.` => source build from the path
- build from a release for a parseable kube version OR a few magical well-known like the "latest" tags

- `--type` explicitly overrides this argument parsing

how about that? Then the common path is just like `kind build node-image .`, `kind build node-image $HOME/kubernetes`, `kind build node-image v1.28.2` and only the confusing case of "I have a directory named v1.28.2 and want to do ./v1.28.2 but pass it as `v1.28.2` requires setting `--type` or users that want to be certain and avoid any chance.

BenTheElder: We could have similar detection for "is a release tar" if we want to support that.

pkg/build/nodeimage/internal/kube/builder_remote.go
```diff
@@ -0,0 +1,203 @@
+/*
```
BenTheElder, line 1: Let's call this one "release" ?

"remote" sort of implies there's a build happening on a remote machine, "release" implies we're using an upstream release. I think. Maybe there's an even better name.

pkg/build/nodeimage/internal/kube/builder_remote.go
```diff
@@ -0,0 +1,203 @@
...
+	return &remoteBuilder{
+		version: version,
+		arch:    arch,
+		logger:  logger,
+	}, nil
+}
+
+// Build implements Bits.Build
+func (b *remoteBuilder) Build() (Bits, error) {
+	url := "https://dl.k8s.io/" + b.version + "/kubernetes-server-linux-" + b.arch + ".tar.gz"
```
BenTheElder, line 53: Do we forsee supporting downstream release hosts? Maybe we punt that to an env override or a totally different `--type` later. That could get messy to support (e.g. we have binaries versus images, and the releases hosting layout is sort of a weak API that is more reasonable when we're just consuming upstream)

pkg/build/nodeimage/internal/kube/builder_remote.go
BenTheElder: Or maybe "download" if that's too ambiguous versus "tarball"

dims:
@BenTheElder @aojea here's what i came up with. Please verify if we can go with this.

|           --type      |         CLI parameter       |                   Action             |
| --------------------- | --------------------------- | ------------------------------------ |
|  NONE	                |          NONE               |   Autodetect k8s directory to build  |
|  NONE                 |  resolves as a Directory    |   Use directory as source to build   | 
|  NONE                 |  resolves as a tarball file |   Treat as binary tar ball           |
|  NONE                 |  parses as a semver         |   Download from dl.k8s.io            |
|  source               |       Directory             |   Use directory as source to build   | 
|  file                 |      tarball file           |   Treat as binary tar ball           |
|  release              |        semver               |   Download from dl.k8s.io            |
|  url                  |     URL to tarball          |   Download binary tarball            |

pkg/build/nodeimage/internal/kube/builder_remote.go
dims: covered this use case, we just let them specify direct url(s) - https://github.com/kubernetes-sigs/kind/pull/3614#issuecomment-2121474989

pkg/build/nodeimage/internal/kube/builder_remote.go
dims: see consolidated table below - https://github.com/kubernetes-sigs/kind/pull/3614#issuecomment-2121474989

pkg/build/nodeimage/build.go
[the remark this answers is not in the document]
dims: agree - see https://github.com/kubernetes-sigs/kind/pull/3614#issuecomment-2121474989

pkg/cmd/kind/build/nodeimage/nodeimage.go
dims: consolidated table https://github.com/kubernetes-sigs/kind/pull/3614#issuecomment-2121474989

pkg/build/nodeimage/build.go
```diff
@@ -46,21 +49,49 @@ func Build(options ...Option) error {
...
 		}
-		ctx.kubeRoot = kubeRoot
 	}
 
-	// initialize bits
-	builder, err := kube.NewDockerBuilder(ctx.logger, ctx.kubeRoot, ctx.arch)
-	if err != nil {
-		return err
+	if ctx.builder == nil && ctx.kubeParam != "" {
+		if info, err := os.Stat(ctx.kubeParam); err == nil && info.Mode().IsRegular() {
```
stmcginnis, line 69: I think I'd feel safer swapping this with the SemVer check above. Maybe it's rare, but if I'm a user testing out things with different versions and I make a local source directory named `v1.30.0`, it might be surprising why my local source isn't being used.

Or maybe a note somewhere declaring the expected order of precedence. Which reminds me - should we update the docs with this change to note the additional options?

dims: @stmcginnis good points, let me revise the PR with the table i had above as well.

dims:
@stmcginnis @aojea @BenTheElder here are the latest command lines i tested

```
kind build node-image https://dl.k8s.io/v1.30.0/kubernetes-server-linux-arm64.tar.gz
kind build node-image $HOME/junk/kubernetes-server-linux-amd64.tar.gz
kind build node-image v1.30.0
kind build node-image $HOME/go/src/k8s.io/kubernetes/

kind build node-image --type url https://dl.k8s.io/v1.30.0/kubernetes-server-linux-arm64.tar.gz
kind build node-image --type file ~/junk/kubernetes-server-linux-amd64.tar.gz
kind build node-image --type release v1.30.0
kind build node-image --type source $HOME/go/src/k8s.io/kubernetes/

kind build node-image
```

stmcginnis:
This looks reasonable to me - thanks @dims!

dims:
Updated docs and CLI help. thanks @stmcginnis 

@aojea @BenTheElder PTAL

pkg/build/nodeimage/build.go
```diff
@@ -46,26 +50,95 @@ func Build(options ...Option) error {
...
+		}
+		ctx.builder = builder
 	}
-	ctx.builder = builder
 
 	// do the actual build
 	return ctx.Build()
 }
 
+func detectBuildType(param string) string {
```
aojea, line 120:
```suggestion
// detectBuildType detect the type of build required based on the param passed in the following order
// url: if the param is a valid http or https url
// file: if the param refers to an existing regular file
// source: if the param refers to an existing directory
// release: if the param is a semantic version expression (does this require the v preprended?
func detectBuildType(param string) string {
```

pkg/build/nodeimage/build.go
```diff
@@ -46,26 +50,95 @@ func Build(options ...Option) error {
...
 	}
 
-	// initialize bits
-	builder, err := kube.NewDockerBuilder(ctx.logger, ctx.kubeRoot, ctx.arch)
-	if err != nil {
-		return err
+	if ctx.buildType == "release" {
+		ctx.logger.V(0).Infof("Building using release %q artifacts", ctx.kubeParam)
+		kubever, err := version.ParseSemantic(ctx.kubeParam)
+		if err == nil {
```
aojea, line 83: @BenTheElder does it makes sense to fallback to file if the type is release but the parameters is not correct?
if we want to do it can we add a comment explaining this logic

pkg/build/nodeimage/internal/kube/builder_remote.go
```diff
@@ -0,0 +1,211 @@
...
+			TLSHandshakeTimeout:   10 * time.Second,
+			ResponseHeaderTimeout: 30 * time.Second,
+			IdleConnTimeout:       30 * time.Second,
+		},
+	}
+
+	// this will stop slow downloads after 3 minutes
+	// and interrupt reading of the Response.Body
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
+	defer cancel()
```
aojea, line 128: :/ 

image


Let's make this 10 minutes, assuming there can be people using 512KB downloads ... my fathers had onle 1MBps only one year ago XD

aojea: you have the idleconntimeout that will kick off in case the connection goes idle

pkg/build/nodeimage/internal/kube/builder_remote.go
```diff
@@ -0,0 +1,211 @@
...
+			return fmt.Errorf("extracting image data: %w", err)
+		}
+		f.Close()
+
+		numFiles++
+	}
+
+	logger.V(2).Infof("Successfully extracted %d files from image tarball %s", numFiles, tarPath)
+	return err
+}
```
aojea, line 211: this is used by pkg/build/nodeimage/internal/kube/builder_tarball.go , I wonder if we should put it in an independent fail tar.go or something

pkg/build/nodeimage/internal/kube/builder_remote.go
```diff
@@ -0,0 +1,211 @@
...
+	if response.StatusCode >= 400 {
+		return fmt.Errorf("error response from %q: HTTP %v", url, response.StatusCode)
+	}
+
+	start := time.Now()
+	defer func() {
+		b.logger.V(2).Infof("Copying %q to %q took %q", url, destPath, time.Since(start))
+	}()
+
+	_, err = io.Copy(output, response.Body)
```
aojea, line 150: TODO it will be nice to have a progressbar here https://github.com/joncrlsn/go-examples/blob/master/http-download-with-progress.go

pkg/build/nodeimage/build.go
```diff
@@ -46,26 +50,95 @@ func Build(options ...Option) error {
...
-	// locate sources if no kubernetes source was specified
-	if ctx.kubeRoot == "" {
-		kubeRoot, err := kube.FindSource()
+	if ctx.buildType == "" {
+		ctx.buildType = detectBuildType(ctx.kubeParam)
+		if ctx.buildType != "" {
+			ctx.logger.V(0).Infof("Detected build type: %q", ctx.buildType)
+		}
+	}
+
```
aojea, line 59: TODO consider to encapsulate this into `func getBuilder(buildType string) (builder, error)`

pkg/build/nodeimage/internal/kube/builder_remote.go
BenTheElder: (and users can control-c / sig-term ... ) +1 to extend the timeout

BenTheElder:
> If these are good, then i'll update the docs in this PR itself.

We'll want to add a note that modes other than source directory are only kind v0.24+

pkg/build/nodeimage/internal/kube/builder_remote.go
dims: updated to 10 mins

pkg/build/nodeimage/internal/kube/builder_remote.go
dims: Done!

pkg/build/nodeimage/internal/kube/builder_remote.go
dims: added TODO

pkg/build/nodeimage/build.go
dims: will leave this for a follow up PR

pkg/build/nodeimage/build.go
dims: i'd rather fail here. we can revisit if we feel strongly in a follow up

dims:
> We'll want to add a note that modes other than source directory are only kind v0.24+

Done!

dims:
@BenTheElder is `VM / VM (Fedora) (docker, rootless) (pull_request)` failure a flake? `exec format error` seems like something wrong in the GH itself?

aojea:
/lgtm
Thank you very much Dims, this is really great

dims:
> Thank you very much Dims, this is really great

thanks @aojea !

/assign @BenTheElder

pkg/build/nodeimage/internal/kube/builder_remote.go
```diff
@@ -0,0 +1,156 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
stmcginnis, line 2: nit: Should probably use the current year for new files?

stmcginnis (commented):
I can see this being very useful!

/lgtm

pkg/build/nodeimage/internal/kube/builder_remote.go
dims: Fixed @stmcginnis thanks! please re-lgtm if it looks good.

stmcginnis:
Thanks @dims!

/lgtm

BenTheElder:
> @BenTheElder is VM / VM (Fedora) (docker, rootless) (pull_request) failure a flake? exec format error seems like something wrong in the GH action itself?

yeah, that sounds like a failed download or bad binfmt_misc, we have some flakes in these jobs with the environment and some flakes in prow with the e2e tests (like #3609, still waiting for those backports to merge)

BenTheElder (commented):
/lgtm
/approve

thanks dims! 🎉

tanvp112:
Hi, this is a kind v0.24.0 feature, right? Any thought when this will be release. Thanks!

BenTheElder:
It is. When we're ready. There are other essentials to sort out and all of us only work on kind part time so no particular ETA.

you can install kind @ HEAD with the makefile, instructions in the docs, but it won't be fully supported yet and we may make some changes before releasing.

PR #3624, opened 2024-05-21
set skipPhases in Init and JoinConfiguration

### LINKED ISSUE #3606: Unable to skipPhases: addons/coredns
**What happened**:

When trying to install `kind` cluster with coredns disabled using `skipPhases`, the `skipPhases` are not honored. From the verbose log I attached you can see that `skipPhases` were added to InitConfiguration using `kubeadmConfigPatches` I used in my config (see how to reproduce).

Attached Log: [kind-test.log](https://github.com/kubernetes-sigs/kind/files/15292297/kind-test.log)

The result is this:

```bash
$ kubectl get pods -A
NAMESPACE            NAME                                                   READY   STATUS    RESTARTS   AGE
kube-system          coredns-76f75df574-g4xnp                               0/1     Pending   0          92s
kube-system          coredns-76f75df574-kd54l                               0/1     Pending   0          92s
kube-system          etcd-my-dev-cluster-control-plane                      1/1     Running   0          108s
kube-system          kube-apiserver-my-dev-cluster-control-plane            1/1     Running   0          106s
kube-system          kube-controller-manager-my-dev-cluster-control-plane   1/1     Running   0          108s
kube-system          kube-proxy-2gwdt                                       1/1     Running   0          92s
kube-system          kube-scheduler-my-dev-cluster-control-plane            1/1     Running   0          106s
local-path-storage   local-path-provisioner-7577fdbbfb-ct4ff                0/1     Pending   0          92s
```

CoreDNS will not be installed.

Create cluster with following configuration:

```yaml
---
kind: Cluster
networking:
  disableDefaultCNI: true
nodes:
- role: control-plane
  kubeadmConfigPatches:
  - |
    kind: InitConfiguration
    skipPhases:
    - addon/coredns
```

For example `kind create cluster --name my-dev-cluster --config=kind-test.yaml -v9`.

Probably it is overriden here - https://github.com/kubernetes-sigs/kind/blob/v0.22.0/pkg/cluster/internal/create/actions/kubeadminit/init.go#L63-L80

**Environment:**

- kind version: (use `kind version`): kind v0.22.0 go1.20.13 linux/amd64
- Runtime info: (use `docker info`, `podman info` or `nerdctl info`): Server Version: 20.10.22
- OS (e.g. from `/etc/os-release`): Fedora 35
```bash
- $ kubectl version
Client Version: v1.28.2
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.29.2
```

nojnhuh:
Fixes #3606.

This PR moves the `--skip-phases` flag from the `kubeadm init` and `kubeadm join` commands to the config file for recent versions of Kubernetes to allow users to patch them from a Cluster config. Older (< 1.23) versions of Kubernetes will still specify the `--skip-phases` flag for those commands for compatibility.

### DIFF
--- a/pkg/cluster/internal/create/actions/kubeadmjoin/join.go
+++ b/pkg/cluster/internal/create/actions/kubeadmjoin/join.go
@@ -24,6 +24,7 @@ import (
 	"sigs.k8s.io/kind/pkg/cluster/nodes"
 	"sigs.k8s.io/kind/pkg/errors"
 	"sigs.k8s.io/kind/pkg/exec"
+	"sigs.k8s.io/kind/pkg/internal/version"
 	"sigs.k8s.io/kind/pkg/log"
 
 	"sigs.k8s.io/kind/pkg/cluster/nodeutils"
@@ -117,18 +118,31 @@ func joinWorkers(
 
 // runKubeadmJoin executes kubeadm join command
 func runKubeadmJoin(logger log.Logger, node nodes.Node) error {
-	// run kubeadm join
-	// TODO(bentheelder): this should be using the config file
-	cmd := node.Command(
-		"kubeadm", "join",
+	kubeVersionStr, err := nodeutils.KubeVersion(node)
+	if err != nil {
+		return errors.Wrap(err, "failed to get kubernetes version from node")
+	}
+	kubeVersion, err := version.ParseGeneric(kubeVersionStr)
+	if err != nil {
+		return errors.Wrapf(err, "failed to parse kubernetes version %q", kubeVersionStr)
+	}
+
+	args := []string{
+		"join",
 		// the join command uses the config file generated in a well known location
 		"--config", "/kind/kubeadm.conf",
-		// skip preflight checks, as these have undesirable side effects
-		// and don't tell us much. requires kubeadm 1.13+
-		"--skip-phases=preflight",
 		// increase verbosity for debugging
 		"--v=6",
-	)
+	}
+	// Newer versions set this in the config file.
+	if kubeVersion.LessThan(version.MustParseSemantic("v1.23.0")) {
+		// skip preflight checks, as these have undesirable side effects
+		// and don't tell us much. requires kubeadm 1.13+
+		args = append(args, "--skip-phases=preflight")
+	}
+
+	// run kubeadm join
+	cmd := node.Command("kubeadm", args...)
 	lines, err := exec.CombinedOutputLines(cmd)
 	logger.V(3).Info(strings.Join(lines, "\n"))
 	if err != nil {
--- a/pkg/cluster/internal/kubeadm/config.go
+++ b/pkg/cluster/internal/kubeadm/config.go
@@ -107,6 +107,10 @@ type DerivedConfigData struct {
 	IPv6 bool
 	// kubelet cgroup driver, based on kubernetes version
 	CgroupDriver string
+	// JoinSkipPhases are the skipPhases values for the JoinConfiguration.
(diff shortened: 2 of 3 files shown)


pkg/cluster/internal/create/actions/kubeadmjoin/join.go
```diff
@@ -117,18 +118,31 @@ func joinWorkers(
 
 // runKubeadmJoin executes kubeadm join command
 func runKubeadmJoin(logger log.Logger, node nodes.Node) error {
-	// run kubeadm join
-	// TODO(bentheelder): this should be using the config file
```
nojnhuh, line 121: @BenTheElder It seemed like this was the last missing piece for this TODO. Please let me know if there's still a gap here where this comment is still valuable.

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -107,6 +107,10 @@ type DerivedConfigData struct {
 	IPv6 bool
 	// kubelet cgroup driver, based on kubernetes version
 	CgroupDriver string
+	// JoinSkipPhases are the skipPhases values for the JoinConfiguration.
+	JoinSkipPhases []string
+	// InitSkipPhases are the skipPhases values for the InitConfiguration.
+	InitSkipPhases []string
```
neolit123, line 113: was this discussed on the ticket?
kind doesn't have to wrap these kubeadm options, and users can pass a custom slice via kind kubeadm config patches, which would override the default to skip the preflight.

nojnhuh: Not sure we got this deep into detail in the issue beyond setting `skipPhases` in the config (https://github.com/kubernetes-sigs/kind/issues/3606#issuecomment-2123157692), but I can see where we could get rid of these fields by hardcoding `preflight` and conditionally adding `addon/kube-proxy` right in the template. Is that what you mean? Or are you suggesting not to specify any `skipPhases` by default? Overall I was trying to preserve the existing default behavior since I don't fully understand the implications of running the preflight checks here.

neolit123: NVM, i mistakenly thought that this exposes the skip phase slices in the public facing kind config, which is not the case.

BenTheElder: yeah, naming is hard and all that, this is internal / plumbing options to the config templates

pkg/cluster/internal/create/actions/kubeadmjoin/join.go
BenTheElder: yep! thanks for fixing my ancient TODO 😅

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -166,6 +170,14 @@ func (c *ConfigData) Derive() {
 		runtimeConfig = append(runtimeConfig, fmt.Sprintf("%s=%s", k, v))
 	}
 	c.RuntimeConfigString = strings.Join(runtimeConfig, ",")
+
+	// skip preflight checks, as these have undesirable side effects
```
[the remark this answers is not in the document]
BenTheElder: (also the 1.13+ comment is now redundant, we stopped supporting Kubernetes < 1.15 in https://github.com/kubernetes-sigs/kind/releases/tag/v0.16.0)

BenTheElder (commented):
/lgtm
/approve
thank you!

we can rewrite those old comments in a follow-up

PR #3627, opened 2024-05-24
update pause image to 3.10

bzsuni:
start using registry.k8s.io/pause:3.10

### DIFF
--- a/images/base/files/etc/containerd/config.toml
+++ b/images/base/files/etc/containerd/config.toml
@@ -34,7 +34,7 @@ version = 2
 
 [plugins."io.containerd.grpc.v1.cri"]
   # use fixed sandbox image
-  sandbox_image = "registry.k8s.io/pause:3.7"
+  sandbox_image = "registry.k8s.io/pause:3.10"
   # allow hugepages controller to be missing
   # see https://github.com/containerd/cri/pull/1501
   tolerate_missing_hugepages_controller = true


bzsuni:
ref https://github.com/kubernetes/kubernetes/issues/125092

neolit123 (commented):
/lgtm

pkg/cluster/nodeutils/util_test.go
```diff
@@ -77,7 +77,7 @@ func TestParseSnapshotter(t *testing.T) {
 		max_container_log_line_size = 16384
 		netns_mounts_under_state_dir = false
 		restrict_oom_score_adj = false
-		sandbox_image = "registry.k8s.io/pause:3.7"
+		sandbox_image = "registry.k8s.io/pause:3.10"
```
BenTheElder, line 80: this test shouldn't need changing, it's just test data. let's not change it and then it won't match a search and replace for the tag next time.

BenTheElder:
FWIW we've not been bumping this because all of the changes since ... 3.3? 3.6? have been windows-only and kind nodes do not run on windows.

I am still +1 to update though, pending dropping the unrelated test-data change (it's harmless, but we don't want to keep churning this every time and it's a bad sign if the tests fail when we change the real file)

pkg/cluster/nodeutils/util_test.go
bzsuni: Yes, that's right, I've revert it, thank you

stmcginnis:
/lgtm

Took a little closer look since it sounds like updating isn't strictly necessary. The last time the Linux pause source was changed was ~4 years ago, and newer updates to the container image [have all been for Windows](https://github.com/kubernetes/kubernetes/blob/ec93d3b71a6634659b59bf435959f4befafe3796/build/pause/CHANGELOG.md). That said, it looks like something is overlooked if we don't keep this at least somewhat updated. There's a 10K difference in size, so not much harm there.

BenTheElder (commented):
/approve
thanks

PR #3628, opened 2024-05-24
Update Go dependencies

stmcginnis:
This updates the go dependencies to the latest available that will work with our chosen minimum supported runtime.

Everything works as expected with local testing. Since the latest release just went out, and presumably the next one will be awhile, it would be a good time to bumps these deps to get as much runtime using them to watch for any odd issues that aren't covered by the main happy path.


BenTheElder (commented):
/lgtm
/approve

I kind of regret removing vendor, properly reviewing these is a pain. I think these look good though, thank you.

Issue #3311, opened 2023-07-19
v0.20.0 cannot create clusters on RHEL 7


ncouse:
We are using RHEL7 VMs and have been successfully using these with KinD for quite some time (thank you).

When trying to upgrade to 0.20.0, the cluster fails to install.

Docker is using cgroups v1, and kernel is a rather antique version `3.10.0-1160.81.1.el7.x86_64`.

The root problem seems to be the use of `--cgroupns=private`, which will *not* work in this environment. I presume the issue is with kernel support for the feature - I believe 4.6 is required.

While RHEL 7 is quite old, it is still in support, even with the old Kernel 3.

**What happened**:

KinD 0.20.0 fails to install on a RHEL 7 VM (kernel 3.10.0).

**What you expected to happen**:

Cluster to be created.

**How to reproduce it (as _minimally_ and precisely as possible)**:

```text
$ kind create cluster
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.27.3) 🖼
 ✗ Preparing nodes 📦
Deleted nodes: ["kind-control-plane"]
ERROR: failed to create cluster: command "docker run --name kind-control-plane --hostname kind-control-plane --label io.x-k8s.kind.role=control-plane --privileged --security-opt seccomp=unconfined --security-opt apparmor=unconfined --tmpfs /tmp --tmpfs /run --volume /var --volume /lib/modules:/lib/modules:ro -e KIND_EXPERIMENTAL_CONTAINERD_SNAPSHOTTER --detach --tty --label io.x-k8s.kind.cluster=kind --net kind --restart=on-failure:1 --init=false --cgroupns=private --volume /dev/mapper:/dev/mapper --publish=127.0.0.1:33587:6443/TCP -e KUBECONFIG=/etc/kubernetes/admin.conf kindest/node:v1.27.3@sha256:3966ac761ae0136263ffdb6cfd4db23ef8a83cba8a463690e98317add2c9ba72" failed with error: exit status 125
Command Output: WARNING: Your kernel does not support cgroup namespaces.  Cgroup namespace setting discarded.
83f54548a6e5f603f7eac309719806364f9d7c226c77849a07cd363773f40d4b
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: cgroup namespaces aren't enabled in the kernel: unknown.
```

**Anything else we need to know?**:

[...]

PR #3631, opened 2024-05-27
Add cgroupns requirement to known issues

stmcginnis:
This adds an entry to the known issues to call out the requirement for kernel support for cgroup namespaces. This is a common gotcha for those running older Linux distros. While this won't eliminate all bug reports for those attempting to use kind in these environments, it will at least add a clue to help explain when they encounter errors.

Closes: #3311


BenTheElder (commented):
/lgtm
/approve
thanks!

PR #3632, opened 2024-05-29
update comments regarding skipping preflight kubeadm phase

nojnhuh:
Follow-up from https://github.com/kubernetes-sigs/kind/pull/3624#discussion_r1608929179 to clarify why the `preflight` phase is skipped for `kubeadm init` and `kubeadm join`.


BenTheElder (commented):
/lgtm
/approve
thank you!

Issue #3359, opened 2023-09-11
Ephemeral storage not available when using (docker/podman) rootless providers


josecastillolema:
**What happened**:

When using kind with rootless docker/podman pods with ephemeral storage requests never get scheduled, i.e.:
```
apiVersion: v1
kind: Pod
metadata:
  name: test
spec:
  containers:
  - command:
    - sleep
    - infinity
    image: centos
    name: compute
    resources:
      requests:
        ephemeral-storage: 50M
```

**What you expected to happen**:
The pod gets properly scheduled, as it does with rootfull docker/podman

**How to reproduce it (as _minimally_ and precisely as possible)**:
1. Create a kind cluster with rootless docker/podman
2. Try to create the pod

**Anything else we need to know?**:
With rootfull docker/podman:
```
kubectl describe node ..
Capacity:
  cpu:                8
  ephemeral-storage:  71658616Ki
  hugepages-1Gi:      0
  hugepages-2Mi:      0
  memory:             32629028Ki
  pods:               110
Allocatable:
  cpu:                8
  ephemeral-storage:  71658616Ki
  hugepages-1Gi:      0
  hugepages-2Mi:      0
  memory:             32629028Ki
  pods:               110
```

With rootless docker/podman (no ephemeral storage):
```
Capacity:
  cpu:            8
  hugepages-1Gi:  0
  hugepages-2Mi:  0
  memory:         32629028Ki
  pods:           110
Allocatable:
  cpu:            8
  hugepages-1Gi:  0
  hugepages-2Mi:  0
  memory:         32629028Ki
  pods:           110
```

**Environment:**

- kind version: (use `kind version`): v0.20.0 go1.20.7 linux/amd64
- Runtime info: (use `docker info` or `podman info`): 
   - Docker: 24.0.1
   - Podman: 4.6.1
- OS (e.g. from `/etc/os-release`): Fedora Linux

giuseppe:
it seems to be done on purpose with rootless, I've tried the following patch and it solves the problem for me:

```diff
diff --git a/pkg/cluster/internal/kubeadm/config.go b/pkg/cluster/internal/kubeadm/config.go
index 6aa17581..0a17f64e 100644
--- a/pkg/cluster/internal/kubeadm/config.go
+++ b/pkg/cluster/internal/kubeadm/config.go
@@ -79,10 +79,6 @@ type ConfigData struct {
 	// RootlessProvider is true if kind is running with rootless mode
[...]

PR #3360, opened 2023-09-12
cluster: drop DisableLocalStorageCapacityIsolation

giuseppe:
it was used to workaround a kubelet crash issue with rootless providers.

The Kubelet seems to work fine now with localStorageCapacityIsolation enabled in a user namespace so drop the special handling.  After this change, ephemeral storage can be used in a rootless cluster.

Closes: https://github.com/kubernetes-sigs/kind/issues/3359

PR #3651, opened 2024-06-10
[rebased #3360] cluster: drop DisableLocalStorageCapacityIsolation

BenTheElder:
Rebased https://github.com/kubernetes-sigs/kind/pull/3360, original description below:

it was used to workaround a kubelet crash issue with rootless providers.

The Kubelet seems to work fine now with localStorageCapacityIsolation enabled in a user namespace so drop the special handling.  After this change, ephemeral storage can be used in a rootless cluster.

Closes: https://github.com/kubernetes-sigs/kind/issues/3359

### DIFF
--- a/pkg/cluster/internal/kubeadm/config.go
+++ b/pkg/cluster/internal/kubeadm/config.go
@@ -79,10 +79,6 @@ type ConfigData struct {
 	// RootlessProvider is true if kind is running with rootless mode
 	RootlessProvider bool
 
-	// DisableLocalStorageCapacityIsolation is typically set true based on RootlessProvider
-	// based on the Kubernetes version, if true kubelet localStorageCapacityIsolation is set false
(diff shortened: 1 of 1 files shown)


aojea:
all fedora jobs failed?

BenTheElder:
There was a flake SSH-ing to the VM in each of them, doesn't seem related.

BenTheElder:
it never even ran kind in this flake https://github.com/kubernetes-sigs/kind/actions/runs/9457245522/job/26088691929#step:7:22

BenTheElder:
I'd like all the CI to pass here ideally before merge.

I don't think the issue is with these changes though so we should probably debug elsewhere.

https://github.com/kubernetes-sigs/kind/pull/648 will run these without changes

BenTheElder:
Rebased again after #3653

BenTheElder:
All tests passing now.

PR #3653, opened 2024-06-11
fedora CI image: switch to an active mirror

BenTheElder:
https://mirrormanager.fedoraproject.org/mirrors/Fedora/38/x86_64


stmcginnis:
I think #3654 would be better, but if we don't want to make too many changes at once, this looks good.

/lgtm

BenTheElder:
https://github.com/kubernetes-sigs/kind/actions/runs/9471527503/job/26094936846?pr=3653 failed with no output whatsoever ... I think we can treat that as an actions bug and just merge this.

PR #3654, opened 2024-06-11
switch to currently supported fedora

BenTheElder:
~stop hardcoding exact download~

xref https://github.com/kubernetes-sigs/kind/pull/3653

/hold

TODO: figure out why fedora 40 isn't working

### DIFF
--- a/hack/ci/Vagrantfile
+++ b/hack/ci/Vagrantfile
@@ -3,11 +3,9 @@
 
 # Vagrant box for testing kind with non-Ubuntu
 Vagrant.configure("2") do |config|
-  # `config.vm.box = "fedora/37-cloud-base"` seems flaky,
-  # so we specify the URL explicitly.
-  # Mirrors can be found at here: https://admin.fedoraproject.org/mirrormanager/mirrors/Fedora/37/x86_64
-  config.vm.box = "dummy"
-  config.vm.box_url = "https://mirror.math.princeton.edu/pub/fedora-archive/fedora/linux/releases/37/Cloud/x86_64/images/Fedora-Cloud-Base-Vagrant-37-1.7.x86_64.vagrant-virtualbox.box"
+  config.vm.box = "fedora/39-cloud-base"
+  # https://mirrormanager.fedoraproject.org/mirrors/Fedora/39/x86_64
+  config.vm.box_url = "https://gsl-syd.mm.fcix.net/fedora/linux/releases/39/Cloud/x86_64/images/Fedora-Cloud-Base-Vagrant-39-1.5.x86_64.vagrant-virtualbox.box"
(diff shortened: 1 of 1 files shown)


stmcginnis:
#3653 is good, but I kind of think it might be good to just make the bigger jump while we're at it.

/lgtm

BenTheElder:
/override pull-kind-e2e-kubernetes-1-29

pod scheduling timeout in prow is not great but also not relevant to this PR

BenTheElder:
So far it looks like booting things is going fine. I notice the actions are off though, we have "docker/docker" ,"nerdctl/nerdctl", "podman/podman" and ... "vm/vm"

that's not related to this PR, but there's clearly a bug in the actions config

BenTheElder:
ohhhh "VM" is cgroup v2 ...

BenTheElder:
at some point we will need cgroup v2 => v1 inversion ...

BenTheElder:
This change isn't working yet.

BenTheElder:
Looks like Fedora 39 works. It's still supported but 40 would be more current.

It also looks like the downloads are slow and we should switch back to downloading from a mirror.

stmcginnis:
Just capturing failure message here for convenience and in case we lose the job logs.

```txt
Pod "pfpod2" is invalid: [spec.containers[0].resources.requests: Invalid value: "600m": must be less than or equal to cpu limit of 500m, spec.containers[0].resources.requests: Invalid value: "600Gi": must be less than or equal to ephemeral-storage limit of 500Gi, spec.containers[0].resources.requests: Invalid value: "600Mi": must be less than or equal to memory limit of 500Mi]
```

BenTheElder:
So that's a bug in the upstream tests and we should file an issue there, it's defintiely not related to this PR which only affects some of the github actions.

BenTheElder:
/override pull-kind-e2e-kubernetes-1-26 pull-kind-e2e-kubernetes-1-27

(prow e2e not relevant for this change)

BenTheElder:
actions are passing, I lost context and I can't see how to pull up older runs easily on github for comparison, so  I'm not sure if the download is _faster_ but it seems sufficiently fast.

stmcginnis:
> actions are passing, I lost context and I can't see how to pull up older runs easily on github for comparison, so I'm not sure if the download is _faster_ but it seems sufficiently fast.

Looking under the Actions tab, I took a look at this and past runs.

The current `VM/CGroupv2` runs took anywhere between 5:37 and 5.51 for the "Boot Fedora" step.

The same jobs with the "e2e-k8s.sh: support --ginkgo.label-filter" PR took between 4:35 and 5:43.

So very slightly more time with the newer Fedora, but that image is probably very slightly larger than the older one anyway. That seems reasonable to me.

/lgtm

PR #3655, opened 2024-06-11
Clean up GitHub Action definitions

stmcginnis:
This defines a local action for the common actions performed by most of our workflows. This reduces duplication between them to make sure they are consistent, and also make the workflow definitions more concise to just the operations that are unique to the particular testing activities.

### DIFF
--- a/.github/workflows/vm.yaml
+++ b/.github/workflows/vm.yaml
@@ -14,7 +14,7 @@ permissions:
 jobs:
   vm:
     # Fedora is different from Ubuntu in LSM (SELinux), filesystem (btrfs), kernel version, etc.
-    name: "VM (Fedora)"
+    name: "CGroupv2 (Fedora)"
     runs-on: ubuntu-22.04
     timeout-minutes: 30
     strategy:
--- a/.github/actions/setup-env/action.yaml
+++ b/.github/actions/setup-env/action.yaml
@@ -1,0 +1,34 @@
+name: "Setup environment"
+description: "Performs common setup operations."
+runs:
+  using: "composite"
+  steps:
+    - name: Get go version
+      id: golangversion
+      run: |
+        echo "go_version=$(cat .go-version)" >> "$GITHUB_OUTPUT"
+      shell: bash
+
+    - name: Set up Go
+      id: go
+      uses: actions/setup-go@cdcb36043654635271a94b9a6d1392de5bb323a7 # v5.0.1
+      with:
+        go-version: ${{ steps.golangversion.outputs.go_version }}
+        check-latest: true
+
+    - name: Install kind
+      run: sudo make install INSTALL_DIR=/usr/local/bin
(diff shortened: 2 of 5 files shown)


.github/workflows/vm.yaml
```diff
@@ -1,4 +1,4 @@
-name: VM
+name: CGroupv2
```
aojea, line 1: can you avoid this in this PR?

stmcginnis: Yeah, I'll switch it back. Ben had commented on some confusion based on that naming at the same time I had the file open, so I quick threw that in. But it's a different concern from the rest of the change.

BenTheElder: I think we should do this, it's non-obvious.

In the future we'll rewrite these and be testing with cgroup v1 in a vagrant job while the rest of CI is on v2.

stmcginnis: How about this... the overall workflow is named `VM` and the single job under the workflow is also named `VM`. This is why we end up with the displayed name of `VM/VM`. I can keep the workflow named `VM` to denote these are jobs that run with a VM, then rename the job itself to be `CGroupv2`. That would result in the running jobs (multiple because it's a matrix job for each combination) being named `VM/CGroupv2`.

Then if we choose to add other jobs that need to be run within a VM, we could have something like:

- `VM/CGroupv2`
- `VM/CGroupv1`
- `VM/Windows`
- `VM/foo`

etc.

If we have some consensus if that makes sense, I can quick update again.

stmcginnis: Especially since it looks like I'll have to do an update anyway because I missed something. :D

aojea:
> Error: GitHub.DistributedTask.ObjectTemplating.TemplateValidationException: The template is not valid. /home/runner/work/kind/kind/./.github/actions/setup-env/action.yaml (Line: 6, Col: 7): Required property is missing: shell

stmcginnis:
> > Error: GitHub.DistributedTask.ObjectTemplating.TemplateValidationException: The template is not valid. /home/runner/work/kind/kind/./.github/actions/setup-env/action.yaml (Line: 6, Col: 7): Required property is missing: shell

Yep, I have the missing `shell` line staged locally. Just wanted to wait until a [response on the naming](https://github.com/kubernetes-sigs/kind/pull/3655#discussion_r1635496986) before I pushed it up so I don't waste resources running tests if another update is needed.

.github/workflows/vm.yaml
aojea: I remember now, I just didn't want to call this Fedora https://github.com/kubernetes-sigs/kind/pull/3485 so it move to something generic ... basically because this can be a magnet for distro validation

+1 on the VM/faeture format

aojea:
/lgtm
/approve

Thanks

Issue #3470, opened 2024-01-05
extra_mounts: Support docker --mount parameter in addition to current --volume


an-dr-eas-k:
**What would you like to be added**:
currently the configuration for the extra_mounts config structure is translated in the docker argument ```--volume``` with syntax ```--volume=host_path:container_path```. I followed the guide here https://kind.sigs.k8s.io/docs/user/local-registry/ and wanted to terraform the process. I am now stuck at the point to dynamically mount my local config dir to the container directory with name ```/etc/containerd/certs.d/localhost:5001/```. The colon in the directory name cannot be mapped with the ```--volume``` argument. 

I suggest an additional extra_mounts structure that is translated to the --mount parameter of docker. 

**Why is this needed**:
Doing so would allow colons in directory names.

BenTheElder:
xref: https://github.com/kubernetes-sigs/kind/issues/1966

This is probably happening in some future API revision, subject to bandwidth.

FWIW I recommend mounting the whole [..]`certs.d/`

an-dr-eas-k:
thanks, that helps!

BenTheElder:
Notes to anyone looking at this: 
- We have provisional support for both podman and nerdctl
- We very much avoid breaking behavior changes where possible, it's likely that these flags have subtle behavior differences, so we'll have to consider that. On the other hand we probably don't want to make `--volume` any more a part of the API than it already is, as this is pretty leaky to docker/podman/... and wasn't the intent (originally we intended to provide the ability to bind mount test configurations)
- There is more discussion in https://github.com/kubernetes-sigs/kind/issues/1966 and further issues linked from there

PR #3667, opened 2024-06-21
chore(provisioning): enable docker bind mounts #3470

scarlson:
… colon


BenTheElder (commented):
Hi, this is a behavior change in the API, please follow our contributing guide and discuss the approach in an issue first before sending us an implementation to review.

Also, if/when we're reviewing implementation changes, unrelated markdown changes should be in a seperate commit or better yet seperate PR.

BenTheElder:
Thanks for looking into this though :-)

I left some notes on the referenced issue, this will be a little more complex now as we also have to target podman / nerdctl and try to avoid breaking behavior changes.

I think it might be worth considering alternatives such as introducing a new API version and switching exclusively to mount only when using v1alpha5+, unless we know we can do this in a fully backwards compatible way across all 3 runtimes.

PR #3670, opened 2024-06-26
add imagePullPolicy IfNotPresent for default cni

Sigthror:
This should make `kind create cluster` independent from Internet

### DIFF
--- a/pkg/build/nodeimage/const_cni.go
+++ b/pkg/build/nodeimage/const_cni.go
@@ -99,6 +99,7 @@ spec:
       containers:
       - name: kindnet-cni
         image: ` + kindnetdImage + `
+        imagePullPolicy: IfNotPresent
         env:
         - name: HOST_IP
           valueFrom:


pkg/build/nodeimage/const_cni.go
```diff
@@ -99,6 +99,7 @@ spec:
       containers:
       - name: kindnet-cni
         image: ` + kindnetdImage + `
+        imagePullPolicy: IfNotPresent
```
aojea, line 102: this is already the default when the image is not set to `latest`, that is this case, so this is making it explicit but is not actually changing the behavior 

https://github.com/kubernetes/kubernetes/blob/66126bbdc6c28dea6a3994d801974a9bbf892453/pkg/apis/core/v1/defaults.go#L73-L83

Sigthror: I'm a little bit confused. I've tested the behavior with `kindest/node:v1.26.3` and had ImagePullBackOff, but with newest version 1.30.0 all works perfectly

aojea: hmm, @BenTheElder is it possible that this was when we did shas instead of tags?, anyway , being explicit seems good to me

BenTheElder:
Not opposed to setting this, but it will make zero behavioral difference and is not the solution to your issues.

pkg/build/nodeimage/const_cni.go
BenTheElder: Arguably this should be imagePullPolicy: Never if we're being explicit @aojea?

I think we should just leave it, this isn't fixing any bug.

Sigthror: I think it's probably better to make imagePullPolicy explicit.
I've done some investigation and pretty sure that cni images was present on nodes for v1.26.3 images.

<details>
<summary>ctr output</summary>
root@kind-control-plane:/# ctr -n k8s.io i check
REF                                                                                                                             TYPE                                                      DIGEST                                                                  STATUS           SIZE                UNPACKED
docker.io/kindest/kindnetd:v20230330-48f316cd@sha256:c19d6362a6a928139820761475a38c24c0cf84d507b9ddf414a078cf627497af           application/vnd.docker.distribution.manifest.list.v2+json sha256:c19d6362a6a928139820761475a38c24c0cf84d507b9ddf414a078cf627497af complete (3/3)   24.1 MiB/24.1 MiB   true
</details>

<details>
<summary>describe pod</summary>
❯ kubectl -n kube-system describe pod kindnet-dn6wx
Name:             kindnet-dn6wx
Namespace:        kube-system
Priority:         0
Service Account:  kindnet
Node:             kind-control-plane/172.18.0.2
Start Time:       Wed, 26 Jun 2024 20:36:35 +0300
Labels:           app=kindnet
                  controller-revision-hash=cf7b65fb6
                  k8s-app=kindnet
                  pod-template-generation=1
                  tier=node
Annotations:      <none>
Status:           Running
IP:               172.18.0.2
IPs:
  IP:           172.18.0.2
Controlled By:  DaemonSet/kindnet
Containers:
  kindnet-cni:
    Container ID:   containerd://a3dfebaa8de1d15dc19d430900f211cd131fd6e3d4c1ceb1875a8cd4568b8386
    Image:          docker.io/kindest/kindnetd:v20230330-48f316cd@sha256:c19d6362a6a928139820761475a38c24c0cf84d507b9ddf414a078cf627497af
    Image ID:       docker.io/kindest/kindnetd:v20230330-48f316cd@sha256:c19d6362a6a928139820761475a38c24c0cf84d507b9ddf414a078cf627497af
    Port:           <none>
    Host Port:      <none>
    State:          Running
      Started:      Wed, 26 Jun 2024 20:37:15 +0300
    Ready:          True
    Restart Count:  0
    Limits:
      cpu:     100m
      memory:  50Mi
    Requests:
      cpu:     100m
      memory:  50Mi
    Environment:
      HOST_IP:                  (v1:status.hostIP)
      POD_IP:                   (v1:status.podIP)
      POD_SUBNET:              10.244.0.0/16
      CONTROL_PLANE_ENDPOINT:  kind-control-plane:6443
    Mounts:
      /etc/cni/net.d from cni-cfg (rw)
      /lib/modules from lib-modules (ro)
      /run/xtables.lock from xtables-lock (rw)
      /var/run/secrets/kubernetes.io/serviceaccount from kube-api-access-9cw28 (ro)
</details>

Also i've tested the behavior with imagePullPolicy: IfNotPresent by replacing `kind/manifests/default-cni.yaml` and it works even for v1.26.3.

I'm totally agree that right now this isn't fixing anything, but we can avoid such strange behavior in the future by making imagePullPolicy explicit

PR #3673, opened 2024-06-28
Remove DOCKER_CLI_EXPERIMENTAL usage

stmcginnis:
The DOCKER_CLI_EXPERIMENTAL environment variable was used to enable experimental features in the docker CLI. This capability was [deprecated in Docker v19.03 and completely removed in v23.0](https://docs.docker.com/engine/deprecated/#configuration-options-for-experimental-cli-features).

This removes the setting of DOCKER_CLI_EXPERIMENTAL from our scripts as it no longer has any effect and is not needed.

### DIFF
--- a/hack/build/init-buildx.sh
+++ b/hack/build/init-buildx.sh
@@ -15,8 +15,6 @@
 
 set -o errexit -o nounset -o pipefail
 
-export DOCKER_CLI_EXPERIMENTAL=enabled
-
 # We can skip setup if the current builder already has multi-arch
 # AND if it isn't the docker driver, which doesn't work
 current_builder="$(docker buildx inspect)"
--- a/hack/release/build/push-node.sh
+++ b/hack/release/build/push-node.sh
@@ -71,7 +71,6 @@ for arch in "${__arches__[@]}"; do
 done
 
 # combine to manifest list tagged with kubernetes version
-export DOCKER_CLI_EXPERIMENTAL=enabled
 # images must be pushed to be referenced by docker manifest
 # we push only after all builds have succeeded
 for image in "${images[@]}"; do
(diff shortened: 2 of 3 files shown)


BenTheElder:
> This removes the setting of DOCKER_CLI_EXPERIMENTAL from our scripts as it no longer has any effect and is not needed.

unless on an old docker? 19.03 is pretty old though ...

stmcginnis:
> unless on an old docker? 19.03 is pretty old though ...

Yeah, I don't know if we publish an official supported version range anywhere, but I'm fairly sure things would no longer work with that old of a version. If we want to add this statement somewhere I am happy to add it to this PR, but I think we are safe to assume if someone actually is running this old of a docker install they are going to run in to other issues.

BenTheElder:
We're soft-requiring 20.10 (december 2020 and later) at runtime now (https://github.com/kubernetes-sigs/kind/releases/tag/v0.20.0) with cgroupns=private (... though we really may have to revisit that ... none of our options have been great there).

I don't think we want to make a more specific support statement because we're not really equipped to test all of the versions and even with the 20.10 thing, you may get earlier versions working fine if you have cgroups v2 enabled.

/lgtm
/approve

BenTheElder:
We very rarely intentionally adopt anything requiring newer versions of kernel / docker / ..., so far we've made that exception maybe two times, once for the problems with cgroups v1 and cgroup awareness skew => cgroupns, I'm struggling to recall the other one but it was quite early in the project and less impactful as we had less users, I think it was only build-time and not runtime but I'm not finding it.

We don't however, actually test with a matrix of versions and I don't think we want to take that on versus checking requirements when adopting new docker/podman/nerdctl features and reviewing incoming bugs. So far we haven't introduced any unexpected compat issues that I can recall ... but we also don't guarantee anything, similar to Kubernetes which doesn't make any comment about containerd/cri-o versions.

PR #3675, opened 2024-07-01
feature: support kubeadm.k8s.io/v1beta4

ls-2018:
- https://github.com/kubernetes/kubernetes/pull/125029
- https://github.com/kubernetes/kubeadm/issues/2890

Need to wait for the release


pkg/cluster/internal/kubeadm/config.go
```diff
@@ -480,6 +480,161 @@ conntrack:
 {{end}}{{end}}
 `
 
+// ConfigTemplateBetaV4 is the kubeadm config template for API version v1beta4
+const ConfigTemplateBetaV4 = `# config generated by kind
+apiVersion: kubeadm.k8s.io/v1beta4
```
BenTheElder, line 485: any changes we need to be aware of ...?

ls-2018: Here are some of the changes, such as :
- When I redeploy the cluster using kind, I need to pass some env to apiserver, which is not supported by the current beta3 version, but is supported by beta4. Of course, the template I provide in pr may not cover all scenarios, which will need to be added later

#### v1beta3
``` 
// ControlPlaneComponent holds settings common to control plane component of the cluster
type ControlPlaneComponent struct {
	// ExtraArgs is an extra set of flags to pass to the control plane component.
	// A key in this map is the flag name as it appears on the
	// command line except without leading dash(es).
	// TODO: This is temporary and ideally we would like to switch all components to
	// use ComponentConfig + ConfigMaps.
	// +optional
	ExtraArgs map[string]string `json:"extraArgs,omitempty"`

	// ExtraVolumes is an extra set of host volumes, mounted to the control plane component.
	// +optional
	ExtraVolumes []HostPathMount `json:"extraVolumes,omitempty"`
}
```
#### v1beta4
```  v1beta4
// ControlPlaneComponent holds settings common to control plane component of the cluster
type ControlPlaneComponent struct {
	// ExtraArgs is an extra set of flags to pass to the control plane component.
	// A key in this map is the flag name as it appears on the
	// command line except without leading dash(es).
	// TODO: This is temporary and ideally we would like to switch all components to
	// use ComponentConfig + ConfigMaps.
	// +optional
	ExtraArgs map[string]string `json:"extraArgs,omitempty"`

	// ExtraVolumes is an extra set of host volumes, mounted to the control plane component.
	// +optional
	ExtraVolumes []HostPathMount `json:"extraVolumes,omitempty"`

	// ExtraEnvs is an extra set of environment variables to pass to the control plane component.
	// Environment variables passed using ExtraEnvs will override any existing environment variables, or *_proxy environment variables that kubeadm adds by default.
	// +optional
	ExtraEnvs []corev1.EnvVar `json:"extraEnvs,omitempty"`
}
```

BenTheElder: > When I redeploy the cluster using kind, I need to pass some env to apiserver, which is not supported by the current beta3 version, but is supported by beta4. Of course, the template I provide in pr may not cover all scenarios, which will need to be added later

There's no question that we will adopt v1beta4, my question is are there any breaking changes or other changes we should be aware of when adopting it.

ls-2018: I didn't find anything else

neolit123: the above v1beta4 struct is not correct in terms of extra args for v1beta4. in the new versions they are a slice of key-value pairs:
https://groups.google.com/g/kubernetes-sig-cluster-lifecycle/c/9HgtTf0W1hk

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -480,6 +480,161 @@ conntrack:
...
+  bindPort: {{.APIBindPort}}
+nodeRegistration:
+  criSocket: "unix:///run/containerd/containerd.sock"
+  kubeletExtraArgs:
+    node-ip: "{{ .NodeAddress }}"
+    provider-id: "kind://{{.NodeProvider}}/{{.ClusterName}}/{{.NodeName}}"
+    node-labels: "{{ .NodeLabels }}"
+{{ if .InitSkipPhases -}}
+skipPhases:
+  {{ range $phase := .InitSkipPhases -}}
```
dominicqi, line 549: https://github.com/kubernetes-sigs/kind/pull/3693   Will this issue occur again?

pkg/cluster/internal/kubeadm/config.go
BenTheElder: Changing extraArgs like this makes sense but a lot of kind configs will need to start targeting specific kubeadm versions when we roll this out because they've been patching extraArgs portably across versions as a subset of the config patched

@ls-2018 that includes https://github.com/kubernetes-sigs/kind/blob/00d659bd28be6ab79be8bd7db60a2d69742673d0/hack/ci/e2e-k8s.sh#L97

which will be somewhat complicated.

pkg/cluster/internal/kubeadm/config.go
BenTheElder: yeah, we need to fix this here too

BenTheElder:
I think we'll need to take a little bit more care when switching here, maybe we'll target kind switching to this for 1.32+ instead after e.g. our own scripts are updated to handle this ...

(where possible we prefer to gate breaking changes against kubernetes versions users couldn't have been using yet)

BenTheElder:
> v1beta3 is now marked as deprecated but will continue to be supported until version 1.34 or later.

So we need to soon, but we can afford to think a bit about how to communicate the extraArgs change and rework various CI scripts (not just the one in this repo, we're likely to break a lot of other CI in the project when we turn this on, other subprojects have their own scripts)

pkg/cluster/internal/kubeadm/config.go
ls-2018: @dominicqi thank you for your notice.

ls-2018:
@BenTheElder @neolit123 

I think that using v1beta4 should not affect the tools I previously wrote. They do not have any new features. If later tools need to test the new features of v1beta4, they can be tested through patches. I added a case, I'm not sure if everyone can accept this.

neolit123:
> > v1beta3 is now marked as deprecated but will continue to be supported until version 1.34 or later.
> 
> So we need to soon, but we can afford to think a bit about how to communicate the extraArgs change and rework various CI scripts (not just the one in this repo, we're likely to break a lot of other CI in the project when we turn this on, other subprojects have their own scripts)

the extra args change will certainly be disruptive. based on feedback we might keep v1beta3 for longer than 1 year.

ls-2018:
Do you have any good ideas about changing args from map to slice, if not patch

ls-2018:
> > > v1beta3 is now marked as deprecated but will continue to be supported until version 1.34 or later.
> > 
> > 
> > So we need to soon, but we can afford to think a bit about how to communicate the extraArgs change and rework various CI scripts (not just the one in this repo, we're likely to break a lot of other CI in the project when we turn this on, other subprojects have their own scripts)
> 
> the extra args change will certainly be disruptive. based on feedback we might keep v1beta3 for longer than 1 year.

Do you have any good ideas about changing args from map to slice.
@neolit123 , Is there any guidance available?

neolit123:
> Do you have any good ideas about changing args from map to slice.
> @neolit123 , Is there any guidance available?

if you have your v1beta3 config on disk as JSON/YAML the guidance is to use the `kubeadm config migrate` command to upgrade it to v1beta4.

BenTheElder:
> I think that using v1beta4 should not affect the tools I previously wrote. They do not have any new features. If later tools need to test the new features of v1beta4, they can be tested through patches.

Er, see https://github.com/kubernetes-sigs/kind/pull/3675/files#r1714482152

When we start generating for v1beta4 by default, any script attempting to patch extra args has to start patching the v1beta4 format.

Also, previously those scripts could get away with not specifying the kubeadm config API version and letting the trivial patch target all versions.

Now they will need to do version targeting patches.

That includes the e2e script in this repo to start.

/ok-to-test

BenTheElder:
> the extra args change will certainly be disruptive. based on feedback we might keep v1beta3 for longer than 1 year.

It's also a powerful change, and non-version-specific patches were just a shortcut, however kind will be responsible for when v1beta4 transition becomes required for its users.

I think the currently pending release to go along with 1.31 has enough changes to absorb as-is and 1.31 just released so we'll probably have to move this to target 1.32

reneleonhardt:
@ls-2018 would it be better to rebase this onto 1.32.0-alpha.0 in order to fix the remaining tests?

aojea:
prow already rebases, the failures are not related to the versions

```
pkg/cluster/internal/create/actions/config/config_test.go:89:2: expected ';', found g
WARN invalid TestEvent: FAIL	sigs.k8s.io/kind/pkg/cluster/internal/create/actions/config [setup failed]
bad output from test2json: FAIL	sigs.k8s.io/kind/pkg/cluster/internal/create/actions/config [setup failed]
```

reneleonhardt:
@ls-2018 with this information, could you fix the code or tests?

Issue #2826, opened 2022-07-06
support install https proxy certificate


shiying2012:
**What would you like to be added**:

The office we work in requires setting the https_proxy proxy parameter and providing a self-signed CA certificate to access the Internet. For example, for ubuntu, we would put the CA certificate in the /usr/local/share/ca-certificates/ directory and run update-ca-certificates. 

We expect to implement this by modifying the configuration file and mounting the certificate on the host into the container instead of building the node image ourselves.

**Why is this needed**:

The office requires access to the Internet via an HTTPS proxy, and KinD's node image requires an additional certificate to be installed. Very inconvenient.

BenTheElder:
> We expect to implement this by modifying the configuration file and mounting the certificate on the host into the container instead of building the node image ourselves.

https://kind.sigs.k8s.io/docs/user/configuration/#extra-mounts

Then `docker exec` to run update-ca-ceerificates

BenTheElder:
> I find it not very satisfying having to exec in the container to run udpate-ca-certificates... 

understandable

> It seems to me that it would be possible to run the update-ca-certificates from the entrypoint script to apply whatever cert would have been mounted in the container, no?

We'd have to know where it would be mounted. In which case, I'm not sure this is the right model either.

> At the moment I chose to inherit the kindest/node image like this:

This is a pretty good work-around IMHO, my only addition would be to do something like `FROM kindest/node:v1.25.2@sha256: 9be91e9e9cdf116809841fc77ebdb8845443c4c72fe5218f3ae9eb57fdb4bace` (pin to the image from the release you are using, this one is from v0.16.0)

If you're aware of any similar tools (VM tools, minikube, microk8s etc.) that have support for this I'd love to see what established UX looks like.

If this were more standardized I'd love to look at automatically importing the cert root.

pataluc:
> We'd have to know where it would be mounted. In which case, I'm not sure this is the right model either.

As far as I understand, `update-ca-certificates` (on Ubuntu but probably all Debian-like distrib) reads this directory: `/usr/local/share/ca-certificates` (cf. https://manpages.ubuntu.com/manpages/xenial/man8/update-ca-certificates.8.html)

```
Furthermore  all  certificates  with  a  .crt  extension  found below /usr/local/share/ca-
       certificates are also included as implicitly trusted.
```

;)

siredmar:
@BenTheElder i can give you an example for the UX perspective how talos does it. 

[...]

PR #3682, opened 2024-07-11
fix: ensure update-ca-certificates during startup

anthr76:
Related issue https://github.com/kubernetes-sigs/kind/issues/2826

In some environments it's common to configure Podman/Docker to unconditionally mount in certificates with all containers via [mounts.conf](https://www.mankier.com/5/containers-mounts.conf)

This ensures during bootup `update-ca-certificates` is ran

### DIFF
--- a/images/base/files/usr/local/bin/entrypoint
+++ b/images/base/files/usr/local/bin/entrypoint
@@ -567,6 +567,7 @@ fix_product_name
 fix_product_uuid
 select_iptables
 enable_network_magic
+update-ca-certificates
 
 # we want the command (expected to be systemd) to be PID1, so exec to it
 log_info 'starting init'


images/base/files/usr/local/bin/entrypoint
```diff
@@ -567,6 +567,7 @@ fix_product_name
 fix_product_uuid
 select_iptables
 enable_network_magic
+update-ca-certificates
```
BenTheElder, line 570: how much time does this take when it isn't mounted?

we're generally pretty sensitive to startup time and we might be able to do something more clever if this is slow

it's also unnecessary when the container is restarted, right? (unless certs were updated in between ...)

BenTheElder:
Right now those mounting certs can run this manually (we could even provide a daemonset or script for this), but most users who are not doing this have zero impact.

I don't think this is strictly a fix and I'm not sure this is the right approach to making additional certs easier.

BenTheElder:
Let's discuss this in a feature issue first per our contributing guide, I'm not sure this is clearly the right approach.

Thank you.

Issue #3451, opened 2023-12-13
Flaky ingress behavior using ingress-nginx and rootless podman


thomasjm:
**What happened**:

I'm trying to run the basic Ingress example with Ingress Nginx from [here](https://kind.sigs.k8s.io/docs/user/ingress/). I'm using rootless podman.

Once I create the [example services and ingress](https://kind.sigs.k8s.io/docs/user/ingress/#using-ingress), I try to do `curl localhost:12345/foo/hostname`. Note that there's a random port `12345`, due to the fact that I pass `0` for the `hostPort` in the Kind cluster config (see below).

About 10% of the time it works and I get the desired response `foo-app`. The rest of the time, the `curl` command hangs indefinitely.

When I look in the nginx controller logs, I see a lot of messages like the following:

```
2023/12/13 13:51:36 [alert] 353#353: pthread_create() failed (11: Resource temporarily unavailable)
2023/12/13 13:51:36 [alert] 39#39: fork() failed while spawning "cache loader process" (11: Resource temporarily unavailable)
2023/12/13 13:51:36 [alert] 39#39: sendmsg() failed (9: Bad file descriptor)
2023/12/13 13:51:37 [alert] 39#39: worker process 52 exited with fatal code 2 and cannot be respawned
```

This looks to me like `nginx` is spawning a bunch of worker threads, and most of them are failing to create properly. Maybe there's some problem with rootless podman?

**What you expected to happen**:

HTTP requests to the ingress should work reliably.

**How to reproduce it (as _minimally_ and precisely as possible)**:

Just following the Ingress instructions. My exact Kind config is as follows:

<details>
  <summary>Kind config file</summary>

```
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane

  extraMounts:
  - hostPath: /nix/store/mrcy594mjgm5zcckr1f4i901isxiwj0s-binary-cache
    containerPath: /binary-cache
    readOnly: false
    propagation: HostToContainer

  kubeadmConfigPatches:
  - |
    kind: InitConfiguration
    nodeRegistration:
      kubeletExtraArgs:
        node-labels: "ingress-ready=true"
        authorization-mode: "AlwaysAllow"
        streaming-connection-idle-timeout: "0"
  extraPortMappings:
  - containerPort: 80
    hostPort: 0
```

</summary>
</details>

**Environment:**

- kind version: (use `kind version`): `0.20.0`
[...]

PR #3687, opened 2024-07-18
docs(user/rootless.md) Mention possible pids_limit issues with rootless podman

netguino:
When running rootless podman, there can be issues with processes not able to create new pids. This is caused by podman's default limit being too low for scenarios like running nginx and spawning workers.

Following up on this discussion: https://github.com/kubernetes-sigs/kind/issues/3451   

This simply adds a notice to the rootless section and suggests a way to disable said limit if desired.

### DIFF
--- a/site/content/docs/user/rootless.md
+++ b/site/content/docs/user/rootless.md
@@ -52,6 +52,14 @@ Also, depending on the host configuration, the following steps might be needed:
   iptable_nat
   ```
 
+- If using podman, be aware that by default there is a [limit](https://docs.podman.io/en/v4.3/markdown/options/pids-limit.html#pids-limit-limit) to the number of pids that can be created. This can cause problems like nginx workers inside a container not spawning correctly.
+    - If you want to disable this limit, edit your `containers.conf` file (generally located in `/etc/containers/containers.conf`). Note that this could cause things like pid exhaustion to happen on the host machine. Alternatively, change `0` to your desired new limit:
+
+    ```ini
+    [containers]
+    pids_limit = 0
+    ```
+
 ## Restrictions
 
 The restrictions of Rootless Docker apply to kind clusters as well.


site/content/docs/user/rootless.md
```diff
@@ -52,6 +52,12 @@ Also, depending on the host configuration, the following steps might be needed:
   iptable_nat
   ```
 
+- If using podman, be aware that by default there is a [limit](https://docs.podman.io/en/v4.3/markdown/options/pids-limit.html#pids-limit-limit) to the number of pids that can be created. This can cause problems like nginx workers inside a container not spawning correctly.
+    - If you want to disable this limit, add the following to your `containers.conf` file (generally located in `/etc/containers/containers.conf`):
+    ```ini
+    [containers]
+    pids_limit = 0
+    ```
```
BenTheElder, line 60: nit: should hae a blank line before the next heating

site/content/docs/user/rootless.md
```diff
@@ -52,6 +52,12 @@ Also, depending on the host configuration, the following steps might be needed:
   iptable_nat
   ```
 
+- If using podman, be aware that by default there is a [limit](https://docs.podman.io/en/v4.3/markdown/options/pids-limit.html#pids-limit-limit) to the number of pids that can be created. This can cause problems like nginx workers inside a container not spawning correctly.
```
BenTheElder, line 55: we should probably mention potential downsides as well? (to disabling the limit), users could also set a higher limit instead of no limit?

BenTheElder (commented):
thanks!

site/content/docs/user/rootless.md
netguino: What about something like...

If you want to raise this limit, edit your `containers.conf` file ( generally located in `/etc/containers/` ). Note that setting this to `0` will disable the limit, potentially allowing things like pid exhaustion to happen on the host machine. 
```ini
[containers]
pids_limit=10000
```

I'm not 100% sure about doing it this way because most people will likely copy the arbitrary number I chose for the block. 

Alternatively, I can just go with the current way, but rewrite it like this:

If you want to disable this limit, edit your `containers.conf` file ( generally located in `/etc/containers/` ). Note that setting could cause things like pid exhaustion to happen on the host machine. If you prefer, change `0` for your desired maximum number for the new limit. 

```ini
[containers]
pids_limit=0
```

stmcginnis: It seems reasonable to me to show an example using 0, along with that warning about pid exhaustion being a possibility.

But I could also see the inverse. Havintg10000 as the example at least does set some sort of limit. Then there could just be a note stating it is possible to have no limit with 0.

Either approach seems fine to me. Your latest update looks good.

stmcginnis (commented):
/lgtm

site/content/docs/user/rootless.md
BenTheElder: I agree.

I think we just need to at least mention the risk of exhaustion and the possibility to set a specific higher limit instead, with the alternate downside that it may not be high enough and/or may still enable exhaustion.

I think your sample with 0 is sufficient otherwise. Thank yu!

BenTheElder (commented):
/lgtm
/approve

thanks!!

PR #3688, opened 2024-07-18
Add exclued node flag for load image command

Monokaix:
When using kind load command I want to exclude some nodes，for example，I want to load to all nodes except master node，because it has taint and is just a control plane and I don't want load image to this node，so add a new flag just exclude the specific node would be helpful.

### DIFF
--- a/pkg/cmd/kind/load/docker-image/docker-image.go
+++ b/pkg/cmd/kind/load/docker-image/docker-image.go
@@ -43,8 +43,9 @@ type (
 )
 
 type flagpole struct {
-	Name  string
-	Nodes []string
+	Name         string
+	Nodes        []string
+	ExcludeNodes []string
 }
 
 // NewCommand returns a new cobra.Command for loading an image into a cluster
@@ -77,6 +78,12 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 		"nodes",
 		nil,
 		"comma separated list of nodes to load images into",
+	)
+	cmd.Flags().StringSliceVar(
+		&flags.ExcludeNodes,
+		"exclude-nodes",
+		nil,
+		"comma separated list of exclude nodes not to load images into",
 	)
 	return cmd
 }
@@ -126,6 +133,16 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
 				return fmt.Errorf("unknown node: %q", name)
 			}
 			candidateNodes = append(candidateNodes, node)
+		}
+	}
+
+	if len(flags.ExcludeNodes) > 0 {
+		for _, name := range flags.ExcludeNodes {
+			_, ok := nodesByName[name]
+			if !ok {
+				return fmt.Errorf("unknown node: %q", name)
+			}
+			candidateNodes = nodeutils.RemoveNode(name, candidateNodes)
 		}
 	}
 
--- a/pkg/cmd/kind/load/image-archive/image-archive.go
+++ b/pkg/cmd/kind/load/image-archive/image-archive.go
@@ -35,8 +35,9 @@ import (
 )
 
 type flagpole struct {
-	Name  string
-	Nodes []string
+	Name         string
+	Nodes        []string
+	ExcludeNodes []string
 }
 
 // NewCommand returns a new cobra.Command for loading an image into a cluster
@@ -75,6 +76,12 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 		nil,
 		"comma separated list of nodes to load images into",
 	)
+	cmd.Flags().StringSliceVar(
+		&flags.ExcludeNodes,
+		"exclude-nodes",
+		nil,
+		"comma separated list of exclude nodes not to load images into",
+	)
 	return cmd
 }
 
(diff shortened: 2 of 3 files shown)


pkg/cmd/kind/load/docker-image/docker-image.go
```diff
@@ -129,6 +136,16 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
 		}
 	}
 
+	if len(flags.ExcludeNodes) > 0 {
+		for _, name := range flags.ExcludeNodes {
+			_, ok := nodesByName[name]
+			if !ok {
```
neolit123, line 142:
```suggestion
			if _, ok := nodesByName[name]; ok {
```

pkg/cmd/kind/load/docker-image/docker-image.go
```diff
@@ -129,6 +136,16 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
...
+				return fmt.Errorf("unknown node: %q", name)
+			}
+			candidateNodes = nodeutils.RemoveNode(name, candidateNodes)
```
neolit123, line 145: the end result `candidateNodes` is processed with too many loops.
instead try using maps or a smarter way to process the excluded nodes.

for example, only add a node to candiateNodes if it's not in the excluded nodes.

pkg/cmd/kind/load/image-archive/image-archive.go
```diff
@@ -130,6 +137,16 @@ func loadArchiveToNodes(logger log.Logger, provider *cluster.Provider, clusterNa
 		}
 	}
 
+	if len(excludeNodeNames) > 0 {
+		for _, name := range excludeNodeNames {
+			_, ok := nodesByName[name]
+			if !ok {
+				return fmt.Errorf("unknown node: %q", name)
+			}
+			selectedNodes = nodeutils.RemoveNode(name, selectedNodes)
```
neolit123, line 146: same problem here.

pkg/cmd/kind/load/docker-image/docker-image.go
Monokaix: Thanks for your reply! Will update later.

aojea:
you can already specify the ones you want to load the image, why to use the negative too?

BenTheElder:
Yeah ... not a fan of adding exclusive / confusing flag pairs, you can currently specify the set of nodes to load to with an existing flag

Monokaix:
> you can already specify the ones you want to load the image, why to use the negative too?

Because when there are many kind nodes and I just want not loading to a few nodes, without this flag I have to specify most nodes, if we can support exclude semantics, it would be very convenient: )

BenTheElder:
We generally ask that features like this be discussed as a design first, so these discussions are resolved up front.

The problem with adding every convenient little feature is bloat and poor test coverage / support ability.

How many nodes is "many nodes"? We generally don't expect kind clusters to have many, in nearly all cases they should have one. For some specific use cases they might have a few.

Monokaix:
> We generally ask that features like this be discussed as a design first, so these discussions are resolved up front.
> 
> The problem with adding every convenient little feature is bloat and poor test coverage / support ability.
> 
> https://kind.sigs.k8s.io/docs/contributing/getting-started/
> 
> How many nodes is "many nodes"? We generally don't expect kind clusters to have many, in nearly all cases they should have one. For some specific use cases they might have a few.

I can truly understand that, there are three cases I have to create a kind cluster with not only one node.

1. In some benchmark case, we need hundreds or thousands of  nodes to do performance check.

1. A cloud native scheduler project's e2e test needs at least four work nodes and one control plane.

1. I deployed kubeflow with kind, because kubeflow has many components which occupied lots of resources, I have to create six nodes to meet the resource requirements.

As kind becomes more and more popular, users use kind to deploy more than one node should be considered as a normal case：）

stmcginnis:
> deploy more than one node should be considered as a normal case

That is quite normal, and is used regularly for testing. Usually that is a small number of nodes though, and anything scaling beyond that starts to be something better done with other tools than kind.

It sounds like you've been able to get an impressive amount of scale for something not intended to scale that far. It would be interested to see how you've gotten around some of the limitations like inotify, etc. But I think type of setup would still be a very rare thing, so adding a (potentially) confusing option that is the inverse of an existing option may not be worth the maintenance and confusion it would cause.

It may be better to add a step to query the cluster for the nodes you care about, then use that as input for the `--nodes` argument.

aojea:
exclude node
```sh
$ kind get nodes | grep -v kind-control-plane
kind-worker
kind-worker2
```

convert to comma separated list to pass to kind command
```sh
$ kind get nodes | grep -v kind-control-plane | paste -sd, -
kind-worker,kind-worker2
```

voila

```sh
$ kind load docker-image busybox:stable --nodes $(kind get nodes | grep -v kind-control-plane | paste -sd, -)
Image with ID: sha256:65ad0d468eb1c558bf7f4e64e790f586e9eda649ee9f130cd0e835b292bbc5ac already present on the node kind-worker but is missing the tag docker.io/library/busybox:stable. re-tagging...
Image with ID: sha256:65ad0d468eb1c558bf7f4e64e790f586e9eda649ee9f130cd0e835b292bbc5ac already present on the node kind-worker2 but is missing the tag docker.io/library/busybox:stable. re-tagging...
```

BenTheElder:
I think at least for now we're not going through with this, if you'd like to discuss options further, please file a feature discussion issue per our contributing guide.

Apologies for the delay, and thank you for the contributions.

Issue #3685, opened 2024-07-15
WSL environment needs additional cgroupv2 configuration


tppalani:
**What happened**:

I'm have created kind cluster using kindest node with base image using config.yaml i can control plane and node are in read state but when i see application pod i can see some error related i don't see this error in older release version  kindest/node:v1.27.

**What you expected to happen**:

**How to reproduce it (as _minimally_ and precisely as possible)**:

**Anything else we need to know?**:

logs
```
combined from similar events): Liveness probe errored: rpc error: code = Unknown desc = failed to exec in container: failed to start exec "d755ebc8e8c9df8392a0819a9a9873f8b3dc92bc8adc2e0cb78a373effb85c2d": OCI runtime exec failed: exec failed: unable to start container process: error adding pid 655569 to cgroups: failed to write 655569: openat2 /sys/fs/cgroup/unified/kubelet.slice/kubelet-kubepods.slice/kubelet-kubepods-burstable.slice/kubelet-kubepods-burstable-pod9aca007f_efe0_4ef4_98f7_17751468c3e5.slice/cri-containerd-1872cb535b10d6ae6b00f2e0891
```

**Environment:**

- kind version: (use `kind version`): kind v0.18.0 go1.20.2 windows/amd64
- Runtime info: (use `docker info`, `podman info` or `nerdctl info`): 
```
host:
  arch: amd64
  buildahVersion: 1.36.0
  cgroupControllers:
  - cpuset
  - cpu
  - cpuacct
  - blkio
  - memory
  - devices
  - freezer
  - net_cls
  - perf_event
  - net_prio
  - hugetlb
  - pids
  - rdma
  - misc
  cgroupManager: cgroupfs
  cgroupVersion: v1
  conmon:
    package: conmon-2.1.10-1.fc40.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.1.10, commit: '
  cpuUtilization:
    idlePercent: 83.06
    systemPercent: 7.27
    userPercent: 9.67
  cpus: 12
  databaseBackend: sqlite
  distribution:
    distribution: fedora
    variant: container
    version: "40"
  eventLogger: journald
  freeLocks: 2042
  hostname: LDD4C6G3
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 5.15.153.1-microsoft-standard-WSL2
  linkmode: dynamic
  logDriver: journald
  memFree: 407486464
  memTotal: 16566644736
  networkBackend: netavark
[...]

PR #3689, opened 2024-07-18
Add known issue for WSL2 and cgroupv2 support

stmcginnis:
This adds a note to the Known Issues document with details related to proper setup requirements for cgroupv2 support and links to the known issues page for extra visibility from the Using WSL2 page.

Includes a separate commit that cleans up some of the stale information from the known issues page and minor formatting issues found while performing updates.

Closes: #3685


neolit123 (commented):
/lgtm
/kind documentation

site/content/docs/user/known-issues.md
```diff
@@ -389,37 +375,46 @@ docker: Error response from daemon: open /dev/dma_heap: permission denied.
...
+```
+
+Using KIND in these environments will require upgrading your OS to a more recent version that supports cgroup namespaces.
+Another option is to run a virtual machine using a newer kernel.
+
+## Failure to Create Cluster on WSL2
 
-Docker Desktop 4.3.0+ uses Cgroups v2 (see [release notes](https://docs.docker.com/release-notes/)) and can only work with Kubernetes versions >= 1.19 (see [failure to create cluster with Cgroups v2](https://kind.sigs.k8s.io/docs/user/known-issues/#failure-to-create-cluster-with-cgroups-v2)).
+Some Linux kernel options for WSL2 do not have cgroup configured in a way that
+Kind and other Linux-focused tools may expect. This may result in a failure
```
BenTheElder, line 400: nit: `kind`  or `KIND`

site/content/docs/user/known-issues.md
```diff
@@ -389,37 +375,46 @@ docker: Error response from daemon: open /dev/dma_heap: permission denied.
...
+unable to start container process: error adding pid 655569 to cgroups
+```
 
-You may also be able to make a configuration change to your Docker Desktop
-settings. Starting with Docker Desktop 4.4.2 a deprecated option has been added
-to support those requiring cgroupv1. As noted in the [4.4.2 release
-notes](https://docs.docker.com/desktop/mac/release-notes/#docker-desktop-442),
-the setting `deprecatedCgroupv1` can be set to `true` in `settings.json`. After
-restarting the Docker Engine, the VM used by Docker Desktop will use cgroupv1.
+The Kind development team is not able to provide support with Windows and WSL, so
```
BenTheElder, line 407: nit again, we don't use `Kind` in the rest of the docs.

site/content/docs/user/known-issues.md
```diff
@@ -389,37 +375,46 @@ docker: Error response from daemon: open /dev/dma_heap: permission denied.
 
 Although the policy has been fixed in Fedora 34, the fix has not been backported to Fedora 33 as of June 28, 2021. Putting SELinux in permissive mode (`setenforce 0`) is one known workaround. This disables SELinux until the next boot. For more details, see [kind#2296].
 
-## Failure to Create Cluster with Docker Desktop as Container Runtime
+## Docker Desktop for macOS and Windows
+
+Docker containers cannot be executed natively on macOS and Windows, therefore Docker Desktop runs them in a Linux VM. As a consequence, the container networks are not exposed to the host and you cannot reach the kind nodes via IP. You can work around this limitation by configuring [extra port mappings](https://kind.sigs.k8s.io/docs/user/configuration/#extra-port-mappings) though.
```
BenTheElder, line 380: This isn't the only solution and we should elaborate more if we're going to mention it here, they have a number of options including the recent work in cloud provider kind, proxies, running tests inside the cluster network, etc.

Maybe at least "You can work around this limitation by, for example, configuring [extra port mappings] amongst other options"

site/content/docs/user/known-issues.md
```diff
@@ -389,37 +375,46 @@ docker: Error response from daemon: open /dev/dma_heap: permission denied.
...
+
+## Older Linux Distributions
+
+KIND uses a cgroup setting of `cgroupns=private`. The cgroup namespace functionality was added in 2016, so some of the
```
BenTheElder, line 384: (which is on by default in cgroups v2 on all major container runtimes)

BenTheElder (commented):
thanks for working on this, did a first pass

cc @aojea

site/content/docs/user/known-issues.md
stmcginnis: This is showing up as new in the diff, but this is an existing section: https://kind.sigs.k8s.io/docs/user/known-issues/#docker-desktop-for-macos-and-windows

I'll tweak the wording a bit to try to improve it though.

site/content/docs/user/known-issues.md
```diff
@@ -209,19 +202,12 @@ docker image prune
 
 You can verify the issue by exporting the logs (`kind export logs`) and looking
 at the kubelet logs, which may have something like the following:
-```
+
+```txt
 Dec 07 00:37:53 kind-1-control-plane kubelet[688]: I1207 00:37:53.229561     688 eviction_manager.go:340] eviction manager: must evict pod(s) to reclaim ephemeral-storage
 Dec 07 00:37:53 kind-1-control-plane kubelet[688]: E1207 00:37:53.229638     688 eviction_manager.go:351] eviction manager: eviction thresholds have been met, but no pods are active to evict
 ```
-## Failure to Create Cluster with Cgroups v2
```
BenTheElder, line 216: I guess we can refer back to this in git / ask people to please upgrade. It's surprising how often we still get issues from really old releases of both kind and Kubernetes

BenTheElder (commented):
/lgtm
/approve
thanks and sorry for the delay

PR #3690, opened 2024-07-20
Reword comments and fix typo in buildcontext.go

joycecodes:
Non-functional change.

### DIFF
--- a/pkg/build/nodeimage/buildcontext.go
+++ b/pkg/build/nodeimage/buildcontext.go
@@ -187,8 +187,8 @@ func (c *buildContext) prePullImagesAndWriteManifests(bits kube.Bits, parsedVers
 		return repository
 	}
 
-	// correct set of built tags using the same logic we will use to rewrite
-	// the tags as we load the archives
+	// Determine accurate built tags using the logic that will be applied
+	// when rewriting tags during archive loading
 	fixedImages := sets.NewString()
 	fixedImagesMap := make(map[string]string, builtImages.Len()) // key: original images, value: fixed images
 	for _, image := range builtImages.List() {
@@ -316,7 +316,7 @@ func (c *buildContext) prePullImagesAndWriteManifests(bits kube.Bits, parsedVers
 		}
 	}
 
-	// run all image re-tragging concurrently until one fails or all succeed
+	// run all image re-tagging concurrently until one fails or all succeed
 	if err := errors.UntilErrorConcurrent(tagFns); err != nil {
 		c.logger.Errorf("Image build Failed! Failed to re-tag images %v", err)
 		return nil, err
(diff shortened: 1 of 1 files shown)


pkg/build/nodeimage/buildcontext.go
```diff
@@ -125,7 +125,7 @@ func (c *buildContext) buildImage(bits kube.Bits) error {
 		return err
 	}
 
-	// pre-pull images that were not part of the build and write CNI / storage
+	// pre-pull images that were not part of the build and write CNI / Storage
```
stmcginnis, line 128: Storage isn't a proper name here, I don't think it needs to be capitalized.

pkg/build/nodeimage/buildcontext.go
```diff
@@ -249,7 +249,7 @@ func (c *buildContext) prePullImagesAndWriteManifests(bits kube.Bits, parsedVers
 		c.logger.Errorf("Image build Failed! Failed write default Storage Manifest: %v", err)
 		return nil, err
 	}
-	// all builds should install the default storage driver images currently
+	// all builds should install the default Storage driver images currently
```
stmcginnis, line 252: Same here.

stmcginnis (commented):
Some slight improvements here, but nothing I would deem too important. But we will need the EasyCLA issues sorted before we can accept this. Please address the issues pointed out by the comment from EasyCLA and let us know if anything doesn't make sense.

kundan2707:
@joycecodes  please sign cncf-cla

pkg/build/nodeimage/buildcontext.go
BenTheElder: CNI is only capitalized because it is an acronym AND a proper name. In this case storage is neither.

joycecodes:
> It looks like maybe your second commit is OK with easycla but not your first one. It would be good to squash commits anyway.
> 
> You can squash you commits and update it to be the correct author using something like the following:
> 
> ```shell
> git rebase -i HEAD~2
> # change the second commit to squash, fix up combined commit message
> git commit --amend --author="First Last <email@example.com>" --no-edit
> git push -f origin comments
> ```
> 
> We need to make sure changes are covered by the CLA before we can consider accepting them.

Thank you for your patience!

pkg/build/nodeimage/buildcontext.go
```diff
@@ -327,19 +327,18 @@ func (c *buildContext) prePullImagesAndWriteManifests(bits kube.Bits, parsedVers
...
 	runArgs := []string{
-		"-d", // make the client exit while the container continues to run
-		// the container should hang forever, so we can exec in it
-		"--entrypoint=sleep",
+		"-d",                 // make the client exit while the container continues to run
+		"--entrypoint=sleep", // the container should hang forever, so we can exec in it
 		"--name=" + id,
 		"--platform=" + dockerBuildOsAndArch(c.arch),
-		"--security-opt", "seccomp=unconfined", // ignore seccomp
+		"--security-opt", "seccomp=unconfined", // disable secure computing mode
```
BenTheElder, line 341: This change is wrong, seccomp is a specific technology, which we're ignoring by setting this container to unconfined. The old comment is not very detailed, but the new comment is inaccurate. If the reader doesn't know what seccomp is, they should not depend on an inline explanation though, it's rather complicated.

BenTheElder (commented):
one comment, the rest looks good!

pkg/build/nodeimage/buildcontext.go
BenTheElder: I think we should just drop this comment, it's not helpful.

joycecodes: Makes sense! Thanks for the catch :)

BenTheElder (commented):
/lgtm
/approve

PR #3692, opened 2024-07-23
Fix Ingress migration guide link error

rogerogers:
Fix Ingress migration guide link error


aojea:
upps

/approve

Thanks

PR #3693, opened 2024-07-25
Fix Kubeadm template format

dominicqi:
When using `kubeProxyMode: none` The current kubeadm template will produce  

> ERROR: failed to create cluster: failed to generate kubeadm config content: error executing config template: Invalid YAML Template

I found out the kubeadm template will drop the line breaker if  `InitSkipPhases` is more than one, playground  https://go.dev/play/p/-wJhmxW6w43 .


neolit123 (commented):
thanks for the go playground reproducer.
`  {{- range $phase := .InitSkipPhases }}` is correct
/lgtm

BenTheElder (commented):
/lgtm
/approve
nice catch!

PR #3703, opened 2024-08-07
Extract selection of nodes without image to a private function

SimonBaeumer:
**Description**

This PR imrpoves the readability of the code a bit by moving the selection of nodes without an image present into a private function.

### DIFF
--- a/pkg/cmd/kind/load/docker-image/docker-image.go
+++ b/pkg/cmd/kind/load/docker-image/docker-image.go
@@ -130,8 +130,40 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
 	}
 
 	// pick only the nodes that don't have the image
-	selectedNodes := map[string]nodes.Node{}
+	selectedNodes := getNodesWithoutImage(logger, imageNames, imageIDs, candidateNodes)
+
+	// return early if no node needs the image
+	if len(selectedNodes) == 0 {
+		return nil
+	}
+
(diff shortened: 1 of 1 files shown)


pkg/cmd/kind/load/docker-image/docker-image.go
```diff
@@ -130,8 +130,40 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
...
+	for _, selectedNode := range selectedNodes {
+		selectedNode := selectedNode // capture loop variable
+		fns = append(fns, func() error {
+			return loadImage(imagesTarPath, selectedNode)
+		})
+	}
+	return errors.UntilErrorConcurrent(fns)
+}
+
+// getNodesWithoutImage
```
BenTheElder, line 164: This needs to be more descriptive, for example just glancing at the name you probably wouldn't expect that it also mutates the nodes by re-tagging images.

BenTheElder (commented):
Thanks for the PR

I'm not sure this is actually more readable, this is core logic from the command and the new abstraction feels awkward / misleading

SimonBaeumer:
> Thanks for the PR
> 
> I'm not sure this is actually more readable, this is core logic from the command and the new abstraction feels awkward / misleading

Thanks for reviewing @BenTheElder. Yeah, while reading the code I did the refactoring, if in doubt I am in favor closing this PR.

PR #3707, opened 2024-08-13
switch to go 1.22.6

BenTheElder:
ahead of v0.24, we'll also want to follow up with base image bumps after this merges


BenTheElder:
> { failed [FAILED] kubelet should only have a single PID in the system (pidof returned "138075 258")

that's a neat flake. I'm going to go ahead and merge this though, there's NO way this PR has impacted inside the container yet.

BenTheElder:
builds are in https://prow.k8s.io/?repo=kubernetes-sigs%2Fkind&type=postsubmit

then I'll "promote" with crane and open a PR (we do NOT want to reference staging directly and registry.k8s.io has other concerns we'll follow up on later ... no reason to rock the boat right this moment)

Issue #3068, opened 2023-01-13
Allow adding container labels for kind nodes


hellt:
Hi @aojea @BenTheElder 
I would like to see if you consider the enhancement request to be able to provide container metadata via container labels a feasible request.

In containerlab (containerlab.dev) we would like to add support for spinning up kind clusters as part of a more extensive containerlab lab. In containerlab, the containers that are "orchestrated" by it, tagged accordingly so that additional tools around containerlab can filter containers and provide some automation around it.

When looking into kind API we didn't find a way to add metadata to kind containers (these are worker and master nodes). Is is possible to enhance API and allow providing tags to kind containers?

aojea:
If by ` container metadata via container labels` you refer to the kindnode containers, they already have information necessary to identify them

```
            "Labels": {
                "io.x-k8s.kind.cluster": "kind",
                "io.x-k8s.kind.role": "control-plane"
            },
```

does it work for you?

hellt:
@aojea, the enhancement request is about being able to assign custom labels to the ones you do by default.

aojea:
I can see how that can be difficult to support with docker/podman duality
Do you have time next week and we talk over this, this way I can use the opportunity to learn more about the project

hellt:
@aojea yes, we can sync up in twitter DMs to choose some time for a sync

BenTheElder:
@aojea what came of this?

Previously we've rejected this feature. We're trying to keep the options to minimal essential ones as the project is tricky enough to maintain already and there's existing labels for identifying nodes from any particular kind cluster, as well as exported APIs and commands for listing nodes which can then be interacted with via docker/podman.

logikone:
I have a use case where we're creating multiple kind clusters in a CI container for testing. The CI container has the hosts docker.sock mounted in the container. Sometimes when something fails and the job exits you can end up having some dangling kind clusters remaining on the CI host. The job containers that are created by the runner have a bunch of labels on them identifying which repo, job, etc the container belongs to. I'd like to be able to put some of those same labels on the containers that kind creates so I can better identify dangling containers. 

I'd be happy to contribute in my free time if this is a feature that you'd like to see added. Thanks!

BenTheElder:
can you use a computed cluster name instead?

logikone:
Ya, that's what I'm doing currently 👍

icefed:
[...]

PR #3709, opened 2024-08-14
add ContainerLabels for node (#3068)

icefed:
example usage:
```
- role: worker
  containerLabels:
    dev.orbstack.domains: "*.kind.local"
```

hope this is helpful for https://github.com/kubernetes-sigs/kind/issues/3068.


stmcginnis:
> the issue is not closed, and i can see it's useful for some users.

That is just ignoring the negative feedback from the maintainers on the issue though. The concerns raised are still valid. The issue should probably be closed at this point, but I think it was left open to allow for more discussion if anyone had ideas that would address those concerns. I don't believe anything has changed to make it worth adding this.

There are other easy ways outside of kind to handle something like this. With the small number of users where this would be needed, it is better to use those approaches than to add more code to kind to cover it.

BenTheElder:
Pleas see https://kind.sigs.k8s.io/docs/contributing/getting-started/#4-reaching-out

We have not agreed to this feature. In general we try to keep the surface of integration to container runtimes minimal and only add essential features, not just useful features.

icefed:
thanks for your reply.

stmcginnis:
Going to close since this isn't a direction we want to take.

/close

Issue #3714, opened 2024-08-16
new builder modes don't work with pre-release tags


BenTheElder:
**What happened**:
```
$ kind build node-image https://dl.k8s.io/ci/$(curl -L https://dl.k8s.io/ci/latest.txt)/kubernetes-server-linux-amd64.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   138  100   138    0     0   2716      0 --:--:-- --:--:-- --:--:--  2760
100    34  100    34    0     0    322      0 --:--:-- --:--:-- --:--:--   322
Detected build type: "url"
Building using URL: "https://dl.k8s.io/ci/v1.32.0-alpha.0.156+983875b2f5bf64/kubernetes-server-linux-amd64.tar.gz"
Starting to build Kubernetes
Downloading "https://dl.k8s.io/ci/v1.32.0-alpha.0.156+983875b2f5bf64/kubernetes-server-linux-amd64.tar.gz"
Finished building Kubernetes
Building node image ...
Building in container: kind-build-1723833436-1564394937
ERROR: error building node image: invalid Kubernetes version: could not parse pre-release/metadata (-alpha.0.156_983875b2f5bf64) in version "v1.32.0-alpha.0.156_983875b2f5bf64
```

**What you expected to happen**:

should build succesfully

**How to reproduce it (as _minimally_ and precisely as possible)**:

see bove

**Anything else we need to know?**:

**Environment:**

- kind version: (use `kind version`): `kind v0.24.0 go1.22.6 linux/amd64`
- Runtime info: (use `docker info`, `podman info` or `nerdctl info`):
- OS (e.g. from `/etc/os-release`):
- Kubernetes version: (use `kubectl version`):
- Any proxies or other special environment settings?:

/assign

PR #3715, opened 2024-08-16
fix getting version from tarball

BenTheElder:
the docker tags are mangled, and the tarball has the `version` file matching what we'd get from a local source build

fixes https://github.com/kubernetes-sigs/kind/issues/3714

cc @dims @aojea @stmcginnis

### DIFF
--- a/pkg/build/nodeimage/internal/kube/builder_remote.go
+++ b/pkg/build/nodeimage/internal/kube/builder_remote.go
@@ -77,7 +77,7 @@ func (b *remoteBuilder) Build() (Bits, error) {
 	}
 
 	binDir := filepath.Join(tmpDir, "kubernetes/server/bin")
-	contents, err := os.ReadFile(filepath.Join(binDir, "kube-apiserver.docker_tag"))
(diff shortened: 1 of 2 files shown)


BenTheElder:
For release builds, there's no additional metadata, so the docker tag and version file are the same.

The version file represents the binary version, what you'd get from `kubectl version` etc.
The docker tag file represents the mangled form that isn't quite semver because of docker tag character-set limitations.

aojea:
I can see both are the same
```
$ more kubernetes/server/bin/kube-apiserver.docker_tag
v1.32.0-alpha.0.156_983875b2f5bf64
aojea@aojea:~/src/tmp$ more kubernetes/version
v1.32.0-alpha.0.156+983875b2f5bf64

```

the version seems more correct though

/lgtm

stmcginnis (approved):

BenTheElder:
The + is valid semver while _ is not, which is why we only use the _ for the image tags in Kubernetes, it doesn't parse.

aojea:
> The + is valid semver while _ is not, which is why we only use the _ for the image tags in Kubernetes, it doesn't parse.

Missed that details

PR #3716, opened 2024-08-16
cleanup long-deprecated flags

BenTheElder:
/hold

We need to cleanup references around the project first, a number of subprojects are still using these:

https://cs.k8s.io/?q=--kube-root&i=nope&files=&excludeFiles=&repos=

Most of them appear to be copies of a CSI script (which uses both flags)

### DIFF
--- a/pkg/cmd/kind/build/nodeimage/nodeimage.go
+++ b/pkg/cmd/kind/build/nodeimage/nodeimage.go
@@ -30,7 +30,6 @@ type flagpole struct {
 	BuildType string
 	Image     string
 	BaseImage string
-	KubeRoot  string
 	Arch      string
 }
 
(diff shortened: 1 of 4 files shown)


BenTheElder:
A lot of kubernetes projects use kind @ HEAD to enable testing kubernetes @ HEAD, so we should probably track down and fix those references first. https://github.com/kubernetes-sigs/kind/issues/3717

BenTheElder:
still blocked on https://github.com/kubernetes-sigs/kind/issues/3717#issuecomment-2332557751

BenTheElder:
@stmcginnis filed cleanup PRs to downstreams back in August, we should stop waiting here.

/hold cancel

PR #3718, opened 2024-08-17
Add CRI-O Kind tutorial to Kind user docs

kannon92:
Add information about how to use Kind and CRI-O. The source was added recently and it can help CRI-O users use Kind.

### DIFF
--- a/site/content/docs/user/resources.md
+++ b/site/content/docs/user/resources.md
@@ -46,6 +46,10 @@ https://golangexample.com/connect-directly-to-docker-for-mac-containers-via-ip-a
 
 https://docs.gitlab.com/charts/development/kind/
 
+### Using CRI-O with KIND
+
+https://github.com/cri-o/cri-o/blob/release-1.31/tutorials/crio-in-kind.md#cri-o-in-kind
+
 ## KubeCon Talks
 
 The authors have given the following talks relating to KIND:


aojea:
/lgtm
/approve

Thanks

neolit123 (commented):
/lgtm

site/content/docs/user/resources.md
```diff
@@ -46,6 +46,10 @@ https://golangexample.com/connect-directly-to-docker-for-mac-containers-via-ip-a
 
 https://docs.gitlab.com/charts/development/kind/
 
+### Using CRI-O with KIND
+
+https://github.com/cri-o/cri-o/blob/release-1.31/tutorials/crio-in-kind.md#cri-o-in-kind
```
BenTheElder, line 51: Can we add a note that this isn't strictly supported, but is possible?

we don't support depending on any details of the node image other than it contains kubernetes at some version and we reserve the right to switch distros and other implementation details, so this is likely to break and I don't want people thinking it was supposed to be fully supported

(this is also how we treat being your own CNI, we leave room for it but we note that it's not supported and don't take bugs filed for alternative CNIs)

BenTheElder: To be clear: people are going to do this anyhow and I'm happy to link to a reference, I just don't want to set a false expectation that it's supported, we have too many support issues as-is.

BenTheElder: We haven't previously linked to e.g. "using cilium with kind" in the docs, but that one is also pretty common and cilium maintains docs for, and in that case you don't even have to worry about OS packages since CNI install is done with a daemonset using standard paths.

BenTheElder: https://github.com/kubernetes-sigs/kind/pull/3720

Issue #2916, opened 2022-09-08
Running with rootless podman seems not to respect property Delegate=yes


plluksie:
**What happened**: I'm not able to create cluster.

**What you expected to happen**: I'm able to create cluster.

**How to reproduce it (as minimally and precisely as possible)**:

```
❯ sudo cat /etc/systemd/system/user@.service.d/delegate.conf
[Service]
Delegate=yes

❯ KIND_EXPERIMENTAL_PROVIDER=podman kind -v 5 create cluster
using podman due to KIND_EXPERIMENTAL_PROVIDER
enabling experimental podman provider
ERROR: failed to create cluster: running kind with rootless provider requires setting systemd property "Delegate=yes", see https://kind.sigs.k8s.io/docs/user/rootless/
Stack Trace: 
sigs.k8s.io/kind/pkg/errors.New
	sigs.k8s.io/kind/pkg/errors/errors.go:28
sigs.k8s.io/kind/pkg/cluster/internal/create.validateProvider
	sigs.k8s.io/kind/pkg/cluster/internal/create/create.go:253
sigs.k8s.io/kind/pkg/cluster/internal/create.Cluster
	sigs.k8s.io/kind/pkg/cluster/internal/create/create.go:70
sigs.k8s.io/kind/pkg/cluster.(*Provider).Create
	sigs.k8s.io/kind/pkg/cluster/provider.go:182
sigs.k8s.io/kind/pkg/cmd/kind/create/cluster.runE
	sigs.k8s.io/kind/pkg/cmd/kind/create/cluster/createcluster.go:80
sigs.k8s.io/kind/pkg/cmd/kind/create/cluster.NewCommand.func1
	sigs.k8s.io/kind/pkg/cmd/kind/create/cluster/createcluster.go:55
github.com/spf13/cobra.(*Command).execute
	github.com/spf13/cobra@v1.4.0/command.go:856
github.com/spf13/cobra.(*Command).ExecuteC
	github.com/spf13/cobra@v1.4.0/command.go:974
github.com/spf13/cobra.(*Command).Execute
	github.com/spf13/cobra@v1.4.0/command.go:902
sigs.k8s.io/kind/cmd/kind/app.Run
	sigs.k8s.io/kind/cmd/kind/app/main.go:53
sigs.k8s.io/kind/cmd/kind/app.Main
	sigs.k8s.io/kind/cmd/kind/app/main.go:35
main.main
	sigs.k8s.io/kind/main.go:25
[...]

PR #3721, opened 2024-08-20
Add install note to rootless docs

khanhtc1202:
Follow the discussed at https://github.com/kubernetes-sigs/kind/issues/2916#issuecomment-2057035416
Kudos @jocutajar


BenTheElder (commented):
/lgtm
/approve

thank you!

I think as a follow-up we can look into if this should be the default suggestion

PR #3724, opened 2024-08-29
clarify when port-mapping is needed

network-charles:
This PR clarifies when port-mapping is needed.

### Why is this important?

When developers are studying for the CKS exam, they may want to study using a Kind cluster running on a public cloud VM. Stating this information helps them configure use cases that involve exposing a nodePort; ingress, exposing the API server, accessing a pod via a browser, etc.

Referenced page: https://kind.sigs.k8s.io/docs/user/configuration/#extra-port-mappings

### DIFF
--- a/site/content/docs/user/configuration.md
+++ b/site/content/docs/user/configuration.md
@@ -308,6 +308,9 @@ With the installation of the Docker Desktop Application, whether it is on macOs,
 
 You may also want to see the [Ingress Guide].
 
+> **NOTE**: If you're running Kind on a remote host and need to send traffic to Kind node
+> IPs from a different host than where kind is running, you need to configure port-mapping.
+
 {{< codeFromFile file="static/examples/config-with-port-mapping.yaml" lang="yaml" >}}
 
 An example http pod mapping host ports to a container port.


site/content/docs/user/configuration.md
```diff
@@ -304,7 +304,7 @@ Extra port mappings can be used to port forward to the kind nodes. This is a
 cross-platform option to get traffic into your kind cluster. 
 
 If you are running Docker without the Docker Desktop Application on Linux, you can simply send traffic to the node IPs from the host without extra port mappings. 
-With the installation of the Docker Desktop Application, whether it is on macOs, Windows or Linux, you'll want to use these.
+Without the installation of the Docker Desktop Application, whether it is on macOs, Windows or Linux, you'll want to use these.
```
stmcginnis, line 307: I believe the original text was accurate, though this whole section could probably be reworded to make it more clear. Basically it is saying:

- If you are using Docker without Docker Desktop, you don't need extra port mappings.
- If you are using Docker Desktop you do need extra port mappings.

I believe this also now applies to more than just Docker Desktop: colima, Rancher Desktop, etc. Basically anything that enables running Docker on the OS by putting it inside a VM.

network-charles: Ok, thanks for the clarification. 

I don't if you're familiar with the [AWS Cloud9](https://aws.amazon.com/cloud9/) service. I am running one. I had to configure port mapping to reach my pod exposed via a NodePort service. Does it mean that Docker Desktop is installed on my Linux machine which has no GUI?

stmcginnis: Hmm, good question @network-charles. I'm not actually sure what they are using in Cloud9. I would expect either Docker directly, or maybe Finch. It's possible they are doing some sort of special network handling that is causing problems.

network-charles: @stmcginnis, actually I meant that I had to use that configuration to reach my pod from my non-remote machine. Sorry, I wasn't clear enough.

curl traffic `non-remote machine > remote machine > kind node > k8s service > k8s pod`

I just deployed a regular virtual machine on AWS without using Cloud9 and I experienced the same issue. I had to configure port mapping to reach the pod from my non-remote machine.

network-charles: The initial documentation could still be correct. Maybe we need to add a new sentence about accessing the Kind node if you are running Kind on a remote machine and need to send traffic to the Kind node's IP from a non-remote machine.

site/content/docs/user/configuration.md
```diff
@@ -308,6 +308,9 @@ With the installation of the Docker Desktop Application, whether it is on macOs,
 
 You may also want to see the [Ingress Guide].
 
+> **NOTE**: If you're running Kind on a remote host and need to send traffic to Kind node
+> IPs from a non-remote host, you need to configure port-mapping.
```
BenTheElder, line 312: Er, port forwarding is generally expected to be used in this guide, even in the local case (especially considering usage like docker desktop)

network-charles: Hi Ben, could you clarify? I don't understand what you mean.

BenTheElder: ... I'm sorry, I read this as the ingress guide.

----

We have a comment above that outlines when you can not use port mappings to reach services listening on the node already, I'm not sure what case this is trying to cover?

What is "non-remote host", wouldn't that just be the host machine?

You need a port mapping to reach nodePort / hostPort services UNLESS you're on the same host as the kind node and actually local to the nodes / not using a VM solution like docker desktop.

(You can also avoid this by sending traffic from another container in all cases)

network-charles: > What is "non-remote host", wouldn't that just be the host machine?

I am running Kind on a cloud VM, the non-remote host is the machine I am using to access the cloud VM.

stmcginnis: I think I agree with Ben here that "non-remote host" is a little confusing and would be better stated as "local host machine" or "your local machine". Or maybe "from a different host than where kind is running"?

network-charles: I love this suggestion "from a different host than where kind is running". 

I'll update it, thanks for the feedback, @stmcginnis.

stmcginnis:
Thanks @network-charles - I think this question has come up enough that having a mention like this might help avoid some confusion.

/lgtm

network-charles:
It’s a pleasure.

aojea:
/ok-to-test
/lgtm
/approve

It seems you reached consensus in https://github.com/kubernetes-sigs/kind/pull/3724#discussion_r1743925323

Issue #3729, opened 2024-09-02
kind-with-registry.sh breaks with podman


wbrefvem:
**What should be cleaned up or changed**:

The script `site/static/examples/kind-with-registry.sh` uses `kind get nodes` and then iterates over the raw output in order to run `docker exec` to update each node's `hosts.toml`. The output when using podman looks something like this:

```
$ kind get nodes
using podman due to KIND_EXPERIMENTAL_PROVIDER
enabling experimental podman provider
kind-control-plane
kind-worker2
kind-worker3
kind-worker
```
This breaks because the script assumes no text in the output other than node names.

On the other hand,

```
$ kubectl get nodes -o jsonpath='{.items[*].metadata.name}'
kind-control-plane kind-worker kind-worker2 kind-worker3
```

achieves the desired result without breaking podman users.

**Why is this needed**:

It's common for podman users to alias docker to podman. This small change allows them to use the script without any fiddling.

wbrefvem:
This is a one-line fix that I'm happy to contribute.

aojea:
> This is a one-line fix that I'm happy to contribute.

please do

aojea:
I apologize, I didn't read your proposal was to use kubectl, I don't think that is valid, let's keep discussing over the PR

BenTheElder:
Hmm, the other text output should be to stderr, is it not, or are we capturing both?

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/3731#discussion_r1742454017

Can you share more details about the script failure, the kind and podman versions, etc?

I'm confused as to why this wouldn't already be working, we should only be capturing stdout and we should be logging to stderr.

aojea:
it seems there is no problem with the script https://github.com/kubernetes-sigs/kind/pull/3731#discussion_r1742814329

should we close?

aojea:
I think with more verbosity you'll get a more meaningful error

BenTheElder:
let's insert `kind export logs` right before the `kind get nodes` call and upload the log files if you can please, something is wrong if `exec` is broken following `kind create cluster` which internally performed many exec after waiting for the containers to be ready.

It sounds like the kubectl call delays things enough that they're healthy again, but that shouldn't be happening and doesn't make much sense, maybe they're restarting?

wbrefvem:
[...]

PR #3731, opened 2024-09-03
fix: make kind-with-registry.sh work with podman

wbrefvem:
Addresses #3729

### DIFF
--- a/site/static/examples/kind-with-registry.sh
+++ b/site/static/examples/kind-with-registry.sh
@@ -36,7 +36,7 @@ EOF
 # We want a consistent name that works from both ends, so we tell containerd to
 # alias localhost:${reg_port} to the registry container when pulling images
 REGISTRY_DIR="/etc/containerd/certs.d/localhost:${reg_port}"
-for node in $(kind get nodes); do
+for node in $(kubectl get nodes -o jsonpath='{.items[*].metadata.name}'); do
   docker exec "${node}" mkdir -p "${REGISTRY_DIR}"
   cat <<EOF | docker exec -i "${node}" cp /dev/stdin "${REGISTRY_DIR}/hosts.toml"
 [host."http://${reg_name}:5000"]


site/static/examples/kind-with-registry.sh
```diff
@@ -36,7 +36,7 @@ EOF
 # We want a consistent name that works from both ends, so we tell containerd to
 # alias localhost:${reg_port} to the registry container when pulling images
 REGISTRY_DIR="/etc/containerd/certs.d/localhost:${reg_port}"
-for node in $(kind get nodes); do
+for node in $(kubectl get nodes -o jsonpath='{.items[*].metadata.name}'); do
```
neolit123, line 39: isn't the right fix to make `kind get nodes` work with podman?

aojea: podman works with kind get nodes, it logs the experimental disclaimer ... podman is not stable enough and still lacks some important capabilities https://github.com/kubernetes-sigs/kind/issues/1778

stmcginnis: There's some small discussion on the linked issue that highlights why the current `kind get nodes` output is a challenge with podman.

aojea: oh, my bad again, it does rely on `kubectl` at the end of the script ... nevermind, let me reassing this to @BenTheElder , he knows better this script

/assign @BenTheElder

BenTheElder: > podman works with kind get nodes, it logs the experimental disclaimer ... podman is not stable enough and still lacks some important capabilities https://github.com/kubernetes-sigs/kind/issues/1778

the log output should not be capture here, are we logging to stdout??

aojea: it uses the logger, I was trying to remember what we log to stderr and what to stdout, I remember some issues with that before

BenTheElder: It should be stderr, which should not be captured:

https://github.com/kubernetes-sigs/kind/blob/0483056594cb49bec92dc5c3f396e7390a39765f/pkg/internal/runtime/runtime.go#L16

https://github.com/kubernetes-sigs/kind/blob/0483056594cb49bec92dc5c3f396e7390a39765f/pkg/cmd/logger.go#L31-L36

BenTheElder: https://github.com/kubernetes-sigs/kind/blob/0483056594cb49bec92dc5c3f396e7390a39765f/pkg/cluster/internal/providers/podman/provider.go#L45

BenTheElder:
```
$ (out="$(KIND_EXPERIMENTAL_PROVIDER=podman kind get nodes || true)"; echo "out: $out")
ERROR: failed to list nodes: command "podman ps -a --filter label=io.x-k8s.kind.cluster=kind --format '{{.Names}}'" failed with error: exec: "podman": executable file not found in $PATH
Command Output: 
out: 
```

Are we sure the script is broken??

wbrefvem: > Are we sure the script is broken?

For podman running on macOS, yes, but I may have misunderstood the reason. Here's what I'm seeing when the script starts to `docker exec` the nodes:

```
Error: can only create exec sessions on running containers: container state improper
```

@BenTheElder you're right that the two lines about podman are not going to stdout. It must be a race condition, and since my solution involves an API call, it's slow enough to lose the race.

BenTheElder: Interesting, we really should be safe to exec once `kind create cluster` succeeds ... uhhh

Did the container crash or something?

Let's follow up more on the issue.

BenTheElder:
This seems to be accidentally mitigating some deeper issue, closing the PR for now but watching the issue for more details so we can determine the root cause and fix

PR #3736, opened 2024-09-06
deploy local registry with ssl

kriegmaster56:
Due to kubernetes not allowing plain http image registry, 

i modify the script to generate a self signed certificate and deploy the registry with it. The script also configure kubernetes to use the registry with ssl enabled.

the registry is still available from the host using the same address localhost:5001 to push images.


BenTheElder:
> Due to kubernetes not allowing plain http image registry,

Huh? Kubernetes does not talk to a registry, only the container runtime does.

And we control the container runtime.

BenTheElder:
I don't think we want to make the demo script more complicated for a non-essential feature, this is only used locally for development anyhow.

You can of course make more complicated versions of it for personal use, but I have not seen this failing to work in any use case yet.

Issue #3713, opened 2024-08-16
DNS resolution unreliable in v0.24.0 when using network policies


jglick:
I tried using the new built-in `NetworkPolicy` support in v0.24.0 (#842 / #3612) instead of Calico. It did not work well, apparently due to DNS problems.

**What happened**:

Requests to either pods inside the cluster by service name, or external hosts, were often delayed or failed due to DNS failures.

**What you expected to happen**:

Requests should succeed immediately, up to the responsiveness of the service.

**How to reproduce it (as _minimally_ and precisely as possible)**:

Will try to put together a minimal test case. In the meantime:

In my original setup a cluster was created with

```yaml
networking:
  disableDefaultCNI: true
  podSubnet: 192.168.0.0/16
```

and I then installed Calico using

```bash
kubectl apply -f https://raw.githubusercontent.com/projectcalico/calico/v3.27.3/manifests/calico.yaml
kubectl -n kube-system set env daemonset/calico-node FELIX_IGNORELOOSERPF=true
```

and defined some network policies. This worked fine, including after creating a fresh cluster using v0.24.0; for example from a pod I could access another pod in the same namespace using a service name

```bash
curl -iv http://othersvc/
```

provided of course that `othersvc` permitted that incoming connection; the command prints content essentially instantly. Contacting external (Internet) services was also as reliable as the site itself.

After deleting the Calico setup and rerunning the exact same scenario, sometimes the `curl` command works instantly as before; sometimes it fails and claims the `othersvc` host could not be found; other times it works, but only after a delay of several seconds. Contacting hosts on the Internet is also unreliable, sometimes giving a hostname resolution error. For example

```bash
curl -Iv http://www.google.com/
```

sometimes works and sometimes does not

```
curl: (6) Could not resolve host: www.google.com
```

whereas

```bash
curl -Iv http://74.125.21.99/
```

works reliably.

**Environment:**

<details>
<summary>Kind v024.0, Docker 27.1.2, Ubuntu Noble</summary>

```console
$ kind version
kind v0.24.0 go1.22.6 linux/amd64
$ docker info
Client: Docker Engine - Community
 Version:    27.1.2
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.16.2
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.29.1
[...]

PR #3752, opened 2024-10-07
kindnetd: kube-network-policies bump to 0.6.0 to fix dns race

aojea:
Fixes: https://github.com/kubernetes-sigs/kind/issues/3713


stmcginnis:
/lgtm

🎉

aojea:
/retest

forgot to push the kindnet image :)

> docker.io/kindest/kindnetd:v20241007-36f62932: digest: sha256:a454aa48d8e10631411378503103b251e3f52856d8be2535efb73a92fa2c0387 size: 685

pkg/build/nodeimage/const_cni.go
```diff
@@ -20,7 +20,7 @@ package nodeimage
 The default CNI manifest and images are our own tiny kindnet
 */
 
-const kindnetdImage = "docker.io/kindest/kindnetd:v20240813-c6f155d6"
+const kindnetdImage = "docker.io/kindest/kindnetd:v20241007-36f62932"
```
BenTheElder, line 23: we should probably also push a new node image based on this change, you can push to the kind staging registry first so we can test it in CI

BenTheElder: (the github actions jobs don't build images, they run the default node image, also users won't have an easy time trying the fix without a node image)

BenTheElder: also nit but: this should be a seperate commit from the source changes. otherwise the git hash is for a commit that doesn't exist.

aojea: is not a separate commit already?

aojea: going to sleep https://console.cloud.google.com/gcr/images/k8s-staging-kind/global/node@sha256:3b24a9aaeaafb74c822848ba551d79b21b0b92270d931efb498bca7af01a6dc0/details?inv=1&invt=AbeKYg&tab=pull , I can´t remember how to create the multi arch node image, will take a look tomorrow

aojea: find it `KUBEROOT=~/src/kubernetes REGISTRY=gcr.io/k8s-staging-kind hack/release/build/push-node.sh`

aojea: pushed and tested

stmcginnis:
/lgtm

Will be nice to get the follow up in #3753!

aojea:
overriding the CI since the failure are going to be fixed by https://github.com/kubernetes-sigs/kind/pull/3753!

PR #3756, opened 2024-10-08
feat: Ability to set SANs to the API server certificate

### LINKED ISSUE #3755: Ability to set extra SANs to the API server certificate
Ability to set extra SANs for the API server certificate

**Why is this needed**:
In some cases, if `apiServerAddress` is modified (which is highly not recommended for sure, and should be done if implications are clear) it may be needed to se custom SANs. Example use-case:
* Temporary AWS instance with the ephemeral pub IP
* Kind cluster set up in that AWS instance
* `apiServerAddress` is set to the internal EC2 IP
* External CI system communicates with the temporary Kind cluster
* Error `Unable to connect to the server: tls: failed to verify certificate: x509: certificate is valid for 10.96.0.1, 172.18.0.2, 0.0.0.0, not 18.118.189.168`

Setting `apiServerAddress` to the external IP will fail because it is not the EC2 IP and `failed to listen on the ${EXTERNAL_IP}` will be risen, so adding extra SANs capabilities would help in such cases.

maksym-iv:
In some cases, if `apiServerAddress` is modified (which is highly not recommended for sure, and should be done if implications are clear) it may be needed to se custom SANs. Example use-case:
* Temporary AWS instance with the ephemeral pub IP
* Kind cluster set up in that AWS instance
* `apiServerAddress` is set to the internal EC2 IP
* External CI system communicates with the temporary Kind cluster
* Error `Unable to connect to the server: tls: failed to verify certificate: x509: certificate is valid for 10.96.0.1, 172.18.0.2, 0.0.0.0, not 18.118.189.168`

Fixes #3755

### DIFF
--- a/pkg/apis/config/v1alpha4/types.go
+++ b/pkg/apis/config/v1alpha4/types.go
@@ -177,6 +177,10 @@ type Networking struct {
 	//
 	// Defaults to 127.0.0.1
 	APIServerAddress string `yaml:"apiServerAddress,omitempty" json:"apiServerAddress,omitempty"`
+	// APIServerExtraSANs is the list of additional SANs to add to the Kubernetes API server certificate
+	//
+	// Defaults to []
+	APIServerExtraSANs []string `yaml:"apiServerExtraSANs,omitempty" json:"apiServerExtraSANs,omitempty"`
 	// PodSubnet is the CIDR used for pod IPs
 	// kind will select a default if unspecified
 	PodSubnet string `yaml:"podSubnet,omitempty" json:"podSubnet,omitempty"`
--- a/pkg/cluster/internal/create/actions/config/config.go
+++ b/pkg/cluster/internal/create/actions/config/config.go
@@ -73,6 +73,7 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 		ControlPlaneEndpoint: controlPlaneEndpoint,
 		APIBindPort:          common.APIServerInternalPort,
 		APIServerAddress:     ctx.Config.Networking.APIServerAddress,
+		APIServerExtraSANs:   ctx.Config.Networking.APIServerExtraSANs,
 		Token:                kubeadm.Token,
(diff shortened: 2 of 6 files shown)


BenTheElder:
Hi, our process is to discuss the approach to features first and agree to the feature and an approach before reviewing code that we will have to maintain.

This is covered in our contributor guide: https://kind.sigs.k8s.io/docs/contributing/getting-started/

maksym-iv:
> Hi, our process is to discuss the approach to features first and agree to the feature and an approach before reviewing code that we will have to maintain.
> 
> This is covered in our contributor guide: https://kind.sigs.k8s.io/docs/contributing/getting-started/

Sure, my bad. I've created the issue before PR and opened a PR just to demonstrate the potential approach, can close the PR.
And yep, my bad, I somehow missed the 
>Please file an enhancement / [feature request issue](https://github.com/kubernetes-sigs/kind/issues/new?labels=kind%2Ffeature&template=enhancement.md) to discuss features before filing a PR

aojea:
So, you can already do this with kubeadm patches https://kind.sigs.k8s.io/docs/user/configuration/#kubeadm-config-patches

```
kind: Cluster
kubeadmConfigPatchesJSON6902:
- group: kubeadm.k8s.io
  version: v1beta2
  kind: ClusterConfiguration
  patch: |
    - op: add
      path: /apiServer/certSANs/-
      value: 172.16.0.2
networking:
  apiServerAddress: "0.0.0.0"
  apiServerPort: 6443
```

this is just exposing this through the kind API, but for the end user is just 1 line of yaml vs 8 lines of yaml, and will conflict with existing solution of using patches

maksym-iv:
> this is just exposing this through the kind API, but for the end user is just 1 line of yaml vs 8 lines of yaml, and will conflict with existing solution of using patches

Sure, makes sense, closing than. Thanks for clarification

BenTheElder:
Thank you!

PR #3749, opened 2024-10-04
docker: detect ip6tables setup failure and fallback to ipv4 only

BenTheElder:
/assign @aojea 
/hold

Issue #3748, opened 2024-10-04
KinD no longer working in GitHub Codespaces


avillela:
**What happened**:

Up until sometime this summer (~July 2024) I used to be able to run KinD in GitHub codespaces. Recently, however, I am no longer able to do so.

**What you expected to happen**:

After installing KinD in my Codespace, I expect it to spin up a Kubernetes cluster; however, it fails to do so.

**How to reproduce it (as _minimally_ and precisely as possible)**:

After creating my codepsace, I run the following:

```shell
# Install KinD
[ $(uname -m) = x86_64 ] && curl -Lo ./kind https://kind.sigs.k8s.io/dl/v0.20.0/kind-linux-amd64
chmod +x ./kind
sudo mv ./kind /usr/local/bin/kind
kind --version

# Create KinD cluster
kind create cluster --name otel-python-lab
```

After running `kind create cluster`, I get the following error:

```text
ERROR: failed to create cluster: failed to ensure docker network: command "docker network create -d=bridge -o com.docker.network.bridge.enable_ip_masquerade=true -o com.docker.network.driver.mtu=1500 --ipv6 --subnet fc00:f853:ccd:e793::/64 kind" failed with error: exit status 1
Command Output: Error response from daemon: Failed to Setup IP tables: Unable to enable NAT rule:  (iptables failed: ip6tables --wait -t nat -I POSTROUTING -s fc00:f853:ccd:e793::/64 ! -o br-6ded57439d75 -j MASQUERADE: ip6tables v1.8.7 (legacy): can't initialize ip6tables table `nat': Table does not exist (do you need to insmod?)
Perhaps ip6tables or your kernel needs to be upgraded.
 (exit status 3))
```

I used to not get this error before and was able to create a cluster successfully.

**Anything else we need to know?**:

**Environment:**

- kind version: (use `kind version`): [v0.24.0](https://github.com/kubernetes-sigs/kind/releases/tag/v0.24.0)
- Runtime info: (use `docker info`, `podman info` or `nerdctl info`): 

```text
Client:
 Version:    27.0.3-1
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.17.1
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.29.7
    Path:     /usr/libexec/docker/cli-plugins/docker-compose
Server:
[...]

PR #3759, opened 2024-10-15
fix missing space in ip6tables error matching

BenTheElder:
follow up to #3749 based on user testing in https://github.com/kubernetes-sigs/kind/issues/3748#issuecomment-2414946811

... the error message from docker has two spaces.

It might make sense to instead check a regex, but I want to put out a trivial fix before thinking about how we want to handle that (we should probably establish a better pattern consistently)


stmcginnis:
/lgtm

Darn double spacers.

PR #3763, opened 2024-10-24
Fix pathType setting in usage.yaml

richburroughs:
The pathType Prefix no longer works, it generates errors. This is the workaround suggested in ingress-nginx issue #10200.

https://github.com/kubernetes/ingress-nginx/issues/10200

I tested it and it worked for me.

### DIFF
--- a/site/static/examples/ingress/usage.yaml
+++ b/site/static/examples/ingress/usage.yaml
@@ -62,14 +62,14 @@ spec:
   rules:
   - http:
       paths:
-      - pathType: Prefix
+      - pathType: ImplementationSpecific
         path: /foo(/|$)(.*)
         backend:
           service:
             name: foo-service
             port:
               number: 8080
-      - pathType: Prefix
+      - pathType: ImplementationSpecific
         path: /bar(/|$)(.*)
         backend:
           service:


stmcginnis:
/lgtm

I think there was some concern in the past that not all IngressControllers would use this same type, but since this example already is using the nginx annotation for the rewrite, I think it makes sense here.

aojea:
can not we change the regex?

aojea:
> ince this example already is using the nginx annotation for the rewrite, I think it makes sense here

I see, is it possible to make the example agnostic of the implementation instead of keep going down the ingress-nginx path?

aojea:
/hold

can we modify the example instead?

it seems based on ingress-nginx maintainers feedback https://github.com/kubernetes-sigs/kind/pull/3764#issuecomment-2437299133 it is doable

richburroughs:
> > ince this example already is using the nginx annotation for the rewrite, I think it makes sense here
> 
> I see, is it possible to make the example agnostic of the implementation instead of keep going down the ingress-nginx path?

I think it's meant to be an example that people can actually run, so I don't expect that can be agnostic.

Also, keep in mind that it's currently broken. I tried to run it and it just doesn't work, which is why I opened the PR.

BenTheElder:
Do we really even want a regex here? Maybe we should craft a simpler example.

BenTheElder:
Thanks for the PR, I'm personally fine with this as a fix, rather than leaving it broken, but I also agree we should pursue simplifying the example to drop the regex and annotations.

aojea:
/hold cancel

don't have time so it is better to have something working than nothing

/approve

Thanls

aojea:
follow up in https://github.com/kubernetes-sigs/kind/pull/3767

PR #3764, opened 2024-10-25
Update pathType to ImplementationSpecific

jjshanks:
While following the guide at https://kind.sigs.k8s.io/docs/user/ingress/#using-ingress the following errors came up when applying the ingress

```
Warning: path /foo(/|$)(.*) cannot be used with pathType Prefix
Warning: path /bar(/|$)(.*) cannot be used with pathType Prefix
Error from server (BadRequest): error when creating "usage.yaml": admission webhook "validate.nginx.ingress.kubernetes.io" denied the request: ingress contains invalid paths: path /foo(/|$)(.*) cannot be used with pathType Prefix
path /bar(/|$)(.*) cannot be used with pathType Prefix
```
I was able to get it to work by changing pathType to ImplementationSpecific which I got from the example at https://kubernetes.github.io/ingress-nginx/examples/rewrite/#examples

### DIFF
--- a/site/static/examples/ingress/usage.yaml
+++ b/site/static/examples/ingress/usage.yaml
@@ -57,19 +57,20 @@ kind: Ingress
 metadata:
   name: example-ingress
   annotations:
+    nginx.ingress.kubernetes.io/use-regex: "true"
     nginx.ingress.kubernetes.io/rewrite-target: /$2
 spec:
   rules:
   - http:
       paths:
-      - pathType: Prefix
+      - pathType: ImplementationSpecific
         path: /foo(/|$)(.*)
         backend:
           service:
             name: foo-service
             port:
               number: 8080
-      - pathType: Prefix
+      - pathType: ImplementationSpecific
         path: /bar(/|$)(.*)
         backend:
           service:


aojea:
we want to encourage standard APIs that favor the ecosystem instead of relying on specific implementations that can cause fragmentation

@rikatz @gacko do you know what existing regex in the example is not valid and suggest how can we change it to be valid?

rikatz:
@aojea the usage of regex on path on Ingress resources is supported just when path is implSpecific. 

Kubernetes project missed adding a validation on the API when it was promoted to v1 and we decided to keep it as is to not be a breaking change :) 

That said, instead of doing implSpecific here, I think just removing the regex from the path should be enough as the desired outcome here seems already to be a path prefix behavior

rikatz:
Ah and remove the annotations :) it doesn't seem you really need a regex and rewrite here

rikatz:
Actually I was thinking we may have some room for improvement on ingress nginx as well. 

The referred workflow here is pretty common, where the path on proxy is like /bla but the path on backend is /. 

@strongjz we may want to support this workflow and reduce usage of regexes. Wdyt?

strongjz:
I thought we also did as part of fixing a CVE as well?

strongjz:
https://kubernetes.github.io/ingress-nginx/faq/#validation-of-path

aojea:
I think the question here is if we can have an example in kind that does not have to use anything that is not specific, it does not matter if we need to change the pod manifests or anything else

jjshanks:
Closing this out in favor of #3763 which was submitted just before mine.

aojea:
we are going to change the section anyway https://github.com/kubernetes-sigs/kind/pull/3767

PR #3767, opened 2024-10-26
improve the ingress controller guide

aojea:
Use Cloud Provider Kind to provisiong loadbalancer and provide a realistic user experience, avoiding to the portmapping and selectors.

Remove implementation specific details so the guide is more agnostic.

Depends on https://github.com/kubernetes/ingress-nginx/issues/12231


aojea:
This is going to merge https://github.com/kubernetes/ingress-nginx/pull/12232/

We need to merge this to get both repos in sync

strongjz:
> This is going to merge [kubernetes/ingress-nginx#12232](https://github.com/kubernetes/ingress-nginx/pull/12232)
> 
> We need to merge this to get both repos in sync

You may need to point to the main till we have a release that will update the deploy manifests for a released version.

site/content/docs/user/ingress.md
```diff
@@ -10,16 +10,8 @@ description: |-
...
-creating a cluster to forward ports from the host
-to an ingress controller running on a node.
-
-We can also setup a custom node label by using `node-labels`
-in the kubeadm `InitConfiguration`, to be used
-by the ingress controller `nodeSelector`.
-
-
 1. [Create a cluster](#create-cluster)
+2. Use [Cloud Provider KIND](./loadbalancer.md) to provision a Loadbalancer for the Ingress Controller.
```
BenTheElder, line 14: I think maybe I would phrase this as installing and running cloud provider kind to enable the loadbalancer controller which ingress-nginx will use through the loadbalancer API, unless we're going to do something where we actually have the user create a loadbalancer object, which we don't appear to be doing.


Ditto for how we describe it below.

BenTheElder (commented):
There is an overall trade-off in requiring installing an additional tool and moving away from the known-portable port-forward example.

We should consider if we need an alternate port-forwarding example now. It's still a useful option without another host installed binary. Currently the configuration guide links back to the Ingress guide for this purpose, which will need fixing (preferably in this PR as that's going to become a bad link otherwise).

BenTheElder:
> In addition, replace the hashicorp images in the Loadbalancer example manifest and use the agnhost ones

[which are maintained by SIG testing upstream]

+1

aojea:
@BenTheElder I reworked the wide to describe both  options: LoadBalancer and extraPortMapping, so people can choose one or the other

site/content/docs/user/ingress.md
```diff
@@ -10,16 +10,12 @@ description: |-
...
 
-We can leverage KIND's `extraPortMapping` config option when
-creating a cluster to forward ports from the host
-to an ingress controller running on a node.
+Ingress exposes HTTP and HTTPS routes from outside the cluster to services within the cluster.
 
-We can also setup a custom node label by using `node-labels`
-in the kubeadm `InitConfiguration`, to be used
-by the ingress controller `nodeSelector`.
+1. [Create a cluster](#create-cluster): There are two primary methods to direct external traffic to Services inside the cluster"
```
stmcginnis, line 15:
```suggestion
1. [Create a cluster](#create-cluster): There are two primary methods to direct external traffic to Services inside the cluster:
```

stmcginnis (commented):
Looks good to me. One minor typo.

PR #3770, opened 2024-10-28
Use version tags for 3rd party GH Actions

stmcginnis:
This switches our third party GitHub Action references to use the major version tag rather than the SHA. This means the actions will automatically use the latest tag within that versions, and Dependabot updates will only be necessary when there is a major version bump where we will want to evaluate the included changes before updating.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,5 +13,4 @@ updates:
     groups:
       actions:
         update-types:
-          - "minor"
-          - "patch"
+          - "major"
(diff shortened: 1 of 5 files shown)


stmcginnis:
I thought that might be the case. :+1:

BenTheElder:
@aojea might have other thoughts, but I think especially "we can trivially revert any issues" is worth it, even if we wind up with more dependabot PRs to review :-)

aojea:
No strong opinion here, I always prefer pinned dependencies, I don't really had much issues with GitHub actions dependencies so I can't really say ... I expect them to be very stable

BenTheElder:
>  I don't really had much issues with GitHub actions dependencies so I can't really say ... I expect them to be very stable

There's also the concern about effectively curl | bash into your CI infra (versus e.g. at least hash pinning) behind a mutable pointer. We currently don't allow github actions access to much but ...

PR #3771, opened 2024-10-31
aggregate masq sync errors

sword-jin:
resolve the TODO, also we should break the loop if context is closed.

### DIFF
--- a/images/kindnetd/cmd/kindnetd/masq.go
+++ b/images/kindnetd/cmd/kindnetd/masq.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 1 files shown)


BenTheElder (commented):
/lgtm
/approve
/hold
(holding for @aojea input)

Also @aojea as a TODO should we consider something else in the future, like N *consecutive* failures instead of 3 lifetime errors?

BenTheElder:
thank you!

stmcginnis:
> should we consider something else in the future, like N consecutive failures instead of 3 lifetime errors?

I had the same thought as I started reviewing, but isn't that covered by the else clause:

```go
errs = errs[:0]
```

BenTheElder:
Oh, you're right.
We structured this a little unusually.

Issue #3772, opened 2024-11-02
Increase the log


aojea:
The default values for l[og rotation on the kubelet ](https://kubernetes.io/docs/concepts/cluster-administration/logging/#log-rotation)
> containerLogMaxSize (default 10Mi) and containerLogMaxFiles (default 5),

are not enough to keep all the logs of a job 

https://gcsweb.k8s.io/gcs/kubernetes-ci-logs/pr-logs/pull/128474/pull-kubernetes-e2e-kind/1852600889194319872/artifacts/kind-control-plane/pods/kube-system_kube-apiserver-kind-control-plane_88820ae9448fa5cfa46a5444a7bf898e/kube-apiserver/

This makes hard to debug or troubleshoot CI problems, specially rare flakes that happen very infrequently, since it is impossible to understand what happened without logs.

We need to use some values that allow to store all the logs that happen during the entire job duration

I think that increasing the  values  to:

```
container-log-max-files: "10"
container-log-max-size: "5Mi"
```

The place to change them is in 
https://github.com/kubernetes-sigs/kind/blob/4f444f8afa344eb2de27e68a2a24b823301fc8d9/hack/ci/e2e-k8s.sh#L101-L124

https://knowledge.broadcom.com/external/article/327480/customizing-the-kubelet-parameters-for-l.html

PR #3774, opened 2024-11-02
update container-log-max-files-max-size

elieser1101:
As requested in https://github.com/kubernetes-sigs/kind/issues/3772

### DIFF
--- a/hack/ci/e2e-k8s.sh
+++ b/hack/ci/e2e-k8s.sh
@@ -115,7 +115,9 @@ create_cluster() {
       apiServer_extra_args="${apiServer_extra_args}
       \"logging-format\": \"${CLUSTER_LOG_FORMAT}\""
   fi
-  kubelet_extra_args="      \"v\": \"${KIND_CLUSTER_LOG_LEVEL}\""
+  kubelet_extra_args="      \"v\": \"${KIND_CLUSTER_LOG_LEVEL}\"
+      \"container-log-max-files\": \"10\"
+      \"container-log-max-size\": \"5Mi\""
   KUBELET_LOG_FORMAT=${KUBELET_LOG_FORMAT:-$CLUSTER_LOG_FORMAT}
   if [ -n "$KUBELET_LOG_FORMAT" ]; then
       check_structured_log_support "KUBECTL_LOG_FORMAT"
--- a/hack/ci/e2e-k8s.sh
+++ b/hack/ci/e2e-k8s.sh
@@ -117,7 +117,7 @@ create_cluster() {
   fi
   kubelet_extra_args="      \"v\": \"${KIND_CLUSTER_LOG_LEVEL}\"
       \"container-log-max-files\": \"10\"
-      \"container-log-max-size\": \"5Mi\""
+      \"container-log-max-size\": \"100Mi\""
   KUBELET_LOG_FORMAT=${KUBELET_LOG_FORMAT:-$CLUSTER_LOG_FORMAT}
   if [ -n "$KUBELET_LOG_FORMAT" ]; then
       check_structured_log_support "KUBECTL_LOG_FORMAT"


hack/ci/e2e-k8s.sh
```diff
@@ -115,7 +115,9 @@ create_cluster() {
       apiServer_extra_args="${apiServer_extra_args}
       \"logging-format\": \"${CLUSTER_LOG_FORMAT}\""
   fi
-  kubelet_extra_args="      \"v\": \"${KIND_CLUSTER_LOG_LEVEL}\""
+  kubelet_extra_args="      \"v\": \"${KIND_CLUSTER_LOG_LEVEL}\"
+      \"container-log-max-files\": \"10\"
+      \"container-log-max-size\": \"5Mi\""
```
aojea, line 120: still not enough 🤔 

https://gcsweb.k8s.io/gcs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kind/3774/pull-kind-e2e-kubernetes-1-30/1852821515305750528/artifacts/kind-control-plane/pods/kube-system_kube-apiserver-kind-control-plane_9c7fae38e19217590379c7ebcfd2e100/kube-apiserver/

The max files seems to be respected, but the max size does not, can you set this to 100Mi?
```suggestion
      \"container-log-max-size\": \"100Mi\""
```

aojea:
@BenTheElder now it keeps the whole run in two files

https://gcsweb.k8s.io/gcs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kind/3774/pull-kind-e2e-kubernetes/1853039395674787840/artifacts/kind-control-plane/pods/kube-system_kube-apiserver-kind-control-plane_f15c9aff8d8905a46ac69983e98c39c7/kube-apiserver/

This is important to be able to debug problems on CI runs, specially rare flakes that does not happen frequently, is the second time this week it happens to me I can not debug because there are no logs when the problem happened

/lgtm
/approve

/hold 

Wait for Ben

/assign @BenTheElder

hack/ci/e2e-k8s.sh
```diff
@@ -115,7 +115,9 @@ create_cluster() {
       apiServer_extra_args="${apiServer_extra_args}
       \"logging-format\": \"${CLUSTER_LOG_FORMAT}\""
   fi
-  kubelet_extra_args="      \"v\": \"${KIND_CLUSTER_LOG_LEVEL}\""
+  kubelet_extra_args="      \"v\": \"${KIND_CLUSTER_LOG_LEVEL}\"
+      \"container-log-max-files\": \"10\"
+      \"container-log-max-size\": \"100Mi\""
```
BenTheElder, line 120: I'm concerned whenever we have to do CI-only configuration @aojea, because it means we're no longer actually testing the defaults, do we think this is because of the verbosity on the components?

aojea: compared with an e2e gce log the sizes are similar https://gcsweb.k8s.io/gcs/kubernetes-ci-logs/pr-logs/pull/128474/pull-kubernetes-e2e-gce/1852746024750157824/artifacts/e2e-44626956c7-674b9-master/

the difference is that those components logs to the filesystem, with the gorunner thingy

BenTheElder: > the difference is that those components logs to the filesystem, with the gorunner thingy, in kind everything logs to the pods and is rotated by the kubelet IIRC

This seems like a gap to revisit later, ideally we want to make sure component logs have robust retention, but maybe not random pods.

BenTheElder:
Also, we could just change the defaults in kind, kubeadm, or kubelet. Again, I think we should avoid tuning the cluster "for CI", but in the fullness of capturing logs for debugging _now_

/hold cancel

PR #3776, opened 2024-11-05
bump go to 1.22.8

BenTheElder:
spoke to @aojea: not doing a full bump of everything this time, we don't want to break runc/containerd/... right now, we'd like to get a bugfix release out ahead of kubecon

however, the go patches should be safe and we're behind


BenTheElder:
we will need to bump the images, however we have post-submit builds, planning tentatively to lean on those and do a follow-up

Issue #3777, opened 2024-11-06
When using podman, the configuration obtained using `get kubeconfig` cannot be used


kebe7jun:
**What happened**:

```
root@kebe-home:~# kind get kubeconfig --name my-cluster-installer > /tmp/x
using podman due to KIND_EXPERIMENTAL_PROVIDER
enabling experimental podman provider
root@kebe-home:~# kubectl --kubeconfig /tmp/x get no
E1106 01:55:36.092466 2212989 memcache.go:265] couldn't get current server API group list: Get "https://:16443/api?timeout=32s": tls: either ServerName or InsecureSkipVerify must be specified in the tls.Config
E1106 01:55:36.093429 2212989 memcache.go:265] couldn't get current server API group list: Get "https://:16443/api?timeout=32s": tls: either ServerName or InsecureSkipVerify must be specified in the tls.Config
E1106 01:55:36.095009 2212989 memcache.go:265] couldn't get current server API group list: Get "https://:16443/api?timeout=32s": tls: either ServerName or InsecureSkipVerify must be specified in the tls.Config
E1106 01:55:36.095988 2212989 memcache.go:265] couldn't get current server API group list: Get "https://:16443/api?timeout=32s": tls: either ServerName or InsecureSkipVerify must be specified in the tls.Config
E1106 01:55:36.097525 2212989 memcache.go:265] couldn't get current server API group list: Get "https://:16443/api?timeout=32s": tls: either ServerName or InsecureSkipVerify must be specified in the tls.Config
Unable to connect to the server: tls: either ServerName or InsecureSkipVerify must be specified in the tls.Config
root@kebe-home:~# cat /tmp/x
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: LS0tLS1CRU...
    server: https://:16443
  name: kind-my-cluster-installer
contexts:
- context:
    cluster: kind-my-cluster-installer
    user: kind-my-cluster-installer
  name: kind-my-cluster-installer
current-context: kind-my-cluster-installer
...
```

**What you expected to happen**:

kubectl command does not report an error.

**How to reproduce it (as _minimally_ and precisely as possible)**:

**Anything else we need to know?**:

**Environment:**

- kind version: (use `kind version`): kind v0.19.0 go1.20.4 linux/amd64
- Runtime info: (use `docker info`, `podman info` or `nerdctl info`):
```
host:
  arch: amd64
  buildahVersion: 1.33.7
  cgroupControllers:
[...]

PR #3778, opened 2024-11-06
Setting the default hostIP for `get kubeconfig`

kebe7jun:
Fix https://github.com/kubernetes-sigs/kind/issues/3777


aojea:
see if this https://github.com/kubernetes-sigs/kind/issues/3777 is happening with latest version of kind, there were multiple fixes since then , specially for podman that broke compatibility with portmaps some time ago

pkg/cluster/internal/providers/podman/provider.go
```diff
@@ -171,6 +171,13 @@ func (p *provider) DeleteNodes(n []nodes.Node) error {
 	return deleteVolumes(nodeVolumes)
 }
 
+func getHostIPOrDefault(hostIP string) string {
```
aojea, line 174:
```suggestion
// getHostIPOrDefault defaults HostIP to localhost if is not set 
// xref: https://github.com/kubernetes-sigs/kind/issues/3777
func getHostIPOrDefault(hostIP string) string {
```

aojea:
/ok-to-test

One suggestion so we understand why we are doing this so we don't have to remember :)

pkg/cluster/internal/providers/podman/provider.go
stmcginnis: +1

kebe7jun:
@aojea Would you please help me figure out why this test(`pull-kind-verify `) fails? I executed `make verify` locally and it was successful. Sorry for not knowing much about kind testing.

pkg/cluster/internal/providers/podman/provider.go
```diff
@@ -171,6 +171,15 @@ func (p *provider) DeleteNodes(n []nodes.Node) error {
 	return deleteVolumes(nodeVolumes)
 }
 
+// getHostIPOrDefault defaults HostIP to localhost if is not set
```
stmcginnis, line 174: The test failure appears to be coming from here. I had to look locally to see you have a space at the end of this line that gets removed by the `make update` target. So you should be able to just run that and check in the updated changes.

kebe7jun: Thank you very much!

aojea:
@kebe7jun please squash all the commits

aojea:
/lgtm
/approve

let's go with 127.0.0.1 ,

PR #3779, opened 2024-11-06
Fix site link to LoadBalancer page

stmcginnis:
The current rendered site does not correctly link local in-repo documents. This results in the LoadBalancer links on the Ingress User Guide pages going to a 404.

Unfortunately this behavior is different between "normal" markdown rendering in local IDEs and while viewing source files in GitHub. But this change makes the rendered public site correctly link to the page.

---

Current example: https://kind.sigs.k8s.io/docs/user/ingress/#option-1-loadbalancer


aojea:
/lgtm
/approve

Thanks

Issue #3781, opened 2024-11-07
Want option for `kind delete cluster` to not update kubeconfig


MikeSpreitzer:
**What would you like to be added**:
I would like an option on the `kind delete cluster` command to skip any updates to the kubeconfig.

**Why is this needed**:
I want to delete multiple clusters concurrently. When I try it (see https://github.com/kubestellar/kubestellar/issues/2558#issuecomment-2461253900), the clusters get deleted but all but one of the `kind` commands fail because the other is holding the kubeconfig lock file.

I can `kubectl config delete-context` myself. Or, in the case just referenced, not even care because a subsequent `kind create cluster` is going to write the context.

stmcginnis:
Having a retry loop seems reasonable.

There may be a workaround you can use for this. But `kind create cluster` and `kind delete cluster` have a `--kubeconfig` argument. This is useful to keep kind cluster settings separate from the default kubeconfig, or to help organize kubeconfigs into individual files that are explicitly used for connecting to different clusters.

You could use that for the creation to keep these clusters separate. But even if you don't use it with `kind create cluster` I believe you should be able to do something like `kind delete cluster --name test123 --kubeconfig $(mktemp)`.

aojea:
wow, this look wrong, I expect delete to actually delete the context

@BenTheElder you are the more familiar with the kubectl config code, can you please take a look?

BenTheElder:
The current implementation of interacting with KUBECONFIG closely mirrors client-go, We need to be careful about that part. But we can add bounded retry around the entire KUBECONFIG edit call when we get a lock failure.

that said, if you want to guarantee the call succeeds, your immediate options are:
- set KUBECONFIG to a no op location when calling this
- retry calling delete cluster

PR #3782, opened 2024-11-07
Retry obtaining KUBECONFIG file lock

stmcginnis:
When there are multiple concurrent create or delete operations, there can be different processing trying to lock the KUBECONFIG file at the same time. This leads to small windows where getting the lock files.

We don't fail the operation if something like "kind delete cluster" isn't able to remove its settings from KUBECONFIG. This is ideal and what most users would expect, but it also means there may be some leftover crud left behind that they are not aware of.

This adds some basic retries to obtaining the file lock. It potentially adds a very small delay (in a much longer process), but makes it more likely that the process is able to get the lock and complete the KUBECONFIG update.

Related: #3781

### DIFF
--- a/pkg/cluster/internal/kubeconfig/internal/kubeconfig/lock.go
+++ b/pkg/cluster/internal/kubeconfig/internal/kubeconfig/lock.go
@@ -19,9 +19,12 @@ package kubeconfig
 import (
 	"os"
 	"path/filepath"
+	"time"
 )
 
-// these are from
+const lockFileRetryAttemps = 5
+
+// these are based on
 // https://github.com/kubernetes/client-go/blob/611184f7c43ae2d520727f01d49620c7ed33412d/tools/clientcmd/loader.go#L439-L440
 
 func lockFile(filename string) error {
@@ -32,12 +35,21 @@ func lockFile(filename string) error {
 			return err
 		}
 	}
-	f, err := os.OpenFile(lockName(filename), os.O_CREATE|os.O_EXCL, 0)
-	if err != nil {
-		return err
+
+	// Retry obtaining the file lock a few times to accommodate concurrent operations.
+	var lastErr error
+	for i := 0; i < lockFileRetryAttemps; i++ {
+		f, err := os.OpenFile(lockName(filename), os.O_CREATE|os.O_EXCL, 0)
+		if err == nil {
+			f.Close()
+			return nil
+		}
+
+		lastErr = err
+		time.Sleep(100 * time.Millisecond)
 	}
-	f.Close()
-	return nil
+
+	return lastErr
 }
 
 func unlockFile(filename string) error {


pkg/cluster/internal/kubeconfig/internal/kubeconfig/lock.go
```diff
@@ -32,12 +35,21 @@ func lockFile(filename string) error {
 			return err
 		}
 	}
-	f, err := os.OpenFile(lockName(filename), os.O_CREATE|os.O_EXCL, 0)
-	if err != nil {
-		return err
+
+	// Retry obtaining the file lock a few times to accommodate concurrent operations.
```
BenTheElder, line 39: I'm not sure this is safe, if there is a lock we're racing with another process anyhow? (which may logically conflict) and we're deviating in behavior from client-go

I think we probably need to retry the entire kubeconfig edit call instead?

stmcginnis: I believe this _should_ be the safe route. Once we obtain the lock, there shouldn't be any concerns about other client-go (or other libraries) that also use file locks to update the KUBECONFIG. All of the update operations happen between that window of obtaining the file lock and removing it.

When we update the contents, we do both [read and write within this locked section](https://github.com/kubernetes-sigs/kind/blob/30bc6ab4203b9157f6055df8d0106521a680c9fa/pkg/cluster/internal/kubeconfig/internal/kubeconfig/merge.go#L33-L52).

Same with removing an entry. The [file lock is obtained first, and only then are updates made](https://github.com/kubernetes-sigs/kind/blob/30bc6ab4203b9157f6055df8d0106521a680c9fa/pkg/cluster/internal/kubeconfig/internal/kubeconfig/remove.go#L32-L46).

The only difference from client-go is it will immediately fail if another process already has that lock. We will be a little more robust and just retry, where typical operations should be complete within milliseconds.

So I believe the only risk if there is another process that is _not_ using the same file locking mechanism to control access to KUBECONFIG. As long is it operates under the same principle of getting the lock before reading and updating, then the only difference is we might be able to wait until the other process is done before completely failing the operation.

BenTheElder: I think I still feel safer about simply retrying the top level kubeconfig mutation method, this code is really hairy, it's a bit broken by design but that's to be compatible with client-go, the whole ecosystem is kind of stuck because there's no obvious way to improve the behavior in say, client-go without breaking the multi-process interaction.

Ideally this would've used a "real" file lock or something, or just kept a single file with "current-context" and moved the kubeconfigs to individual files in a directory, but I think we're a long way from that world still :(

BenTheElder: > The only difference from client-go is it will immediately fail if another process already has that lock. We will be a little more robust and just retry, where typical operations should be complete within milliseconds.

I believe the reason for this is to avoid deadlocking?

BenTheElder: er, I think the missing bit here is that we lock *n* files potentially, sequentially, and that can deadlock between processes with different lists if you don't bail out instead of re-attempting the next lock.

I think it's easier to have confidence about compatibility by backing out and retrying the entire set

the entire approach is ~wrong, but it sort of mostly works and I don't think we want to deviate much

stmcginnis: Makes sense. It's been awhile since I've looked at this. I'll try to get back and update it soon.

stmcginnis:
Took another look through the code, and I still think locking at this level is the most correct. I only see one other place where there are multiple files, and [that's only for determining the right path for the kubeconfig](https://github.com/kubernetes-sigs/kind/blob/30bc6ab4203b9157f6055df8d0106521a680c9fa/pkg/cluster/internal/kubeconfig/internal/kubeconfig/remove.go#L29).

This is a very rare issue though, so rather than introducing something that could cause any risk, leaving the current failure corner case is probably the safest.

Issue #3743, opened 2024-09-12
Move HA Config Snippet to Configuration Section


josephaw1022:
**What would you like to be documented**:

All I simply am requesting is that this is moved (or copy) this section here 

![image](https://github.com/user-attachments/assets/785a37d2-1529-4ebe-a44a-166b4c8a7803)

to the configuration page

![image](https://github.com/user-attachments/assets/4bbf3cf4-a6ea-4d2d-90c7-f1c7596f3f01)

**Why is this needed**:
- Make it easier for anyone to have a page that has details about specific configurations, but also has useful examples that the user can see on the same page

BenTheElder:
HA is a feature, not a config option. Adding nodes is a config option, which is already covered in the config docs (and the quick start).

BenTheElder:
The reason HA is mentioned in this page is to cover the airgapping requirements, the config example is to clarify what we mean by "HA".

What's your use case for "HA"? This feature is extremely niche and poorly supported currently.

josephaw1022:
The only reason I brought this issue up was is specifically because I noticed it in the offline section and thought that it would be an interesting thing to note under the mutli node cluster intro section. 


![image](https://github.com/user-attachments/assets/58d8f4d3-9ac6-4948-8570-bc74b4353f56)

As just a way to say "you can configure an ha cluster as well"

That's all that I was getting at

PR #3784, opened 2024-11-08
Add note to configuration about HA control plane

stmcginnis:
Adds a small note to the Nodes configuration section to point out that multiple control plane nodes may be added to set up a "high availability" control plane cluster.

Closes: #3743


BenTheElder (commented):
/lgtm
/approve

TODO: craft a note more strongly recommending single-node clusters (w/ reasons) unless you have specific need for multi-node behaviors

PR #3787, opened 2024-11-08
update ingress docs

aojea:
The existing manifest was not updated, we need to wait for a new rrelease but until then, it is better we host it ourselves so the guide works correctly.

In addition, remove a comment that no longer applies since current configuration of Ingress uses only standard options.

Tested locally


stmcginnis:
Great!

/lgtm

PR #3794, opened 2024-11-19
chores: add a warning message when delete a not-exists cluster

sword-jin:
a typo error of clean-script cause we have more and more clusters in our test machine, and we don't observe any message.

So I think this error message is useful to help others to encounter this similar problem

### DIFF
--- a/pkg/cluster/internal/delete/delete.go
+++ b/pkg/cluster/internal/delete/delete.go
@@ -44,6 +44,8 @@ func Cluster(logger log.Logger, p providers.Provider, name, explicitKubeconfigPa
 			return err
 		}
 		logger.V(0).Infof("Deleted nodes: %q", n)
+	} else {
+		logger.V(0).Infof("Whooops, no nodes found for cluster %q", name)
 	}
 
 	if kerr != nil {
--- a/pkg/cluster/internal/loadbalancer/config.go
+++ b/pkg/cluster/internal/loadbalancer/config.go
@@ -18,6 +18,7 @@ package loadbalancer
 
 import (
 	"bytes"
+	"fmt"
 	"text/template"
 
 	"sigs.k8s.io/kind/pkg/errors"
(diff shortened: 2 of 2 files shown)


pkg/cluster/internal/loadbalancer/config.go
```diff
@@ -68,16 +69,22 @@ backend kube-apiservers
...
-// Config returns a kubeadm config generated from config data, in particular
-// the kubernetes version
-func Config(data *ConfigData) (config string, err error) {
+var configTpl *template.Template
+
+func init() {
 	t, err := template.New("loadbalancer-config").Parse(DefaultConfigTemplate)
 	if err != nil {
-		return "", errors.Wrap(err, "failed to parse config template")
+		panic(fmt.Sprintf("BUG: invalid default loadbalancer config template, check the template: %v", err))
```
aojea, line 77: :/

pkg/cluster/internal/delete/delete.go
```diff
@@ -44,6 +44,8 @@ func Cluster(logger log.Logger, p providers.Provider, name, explicitKubeconfigPa
 			return err
 		}
 		logger.V(0).Infof("Deleted nodes: %q", n)
+	} else {
+		logger.V(0).Infof("Whooops, no nodes found for cluster %q", name)
```
stmcginnis, line 48: The idempotent nature of cluster deletion has been brought up several times in the past. We do not want to do this. If you need to know if a cluster exists or not before deleting, then you should list the clusters first.

The line above was added as a compromise. So another option is to check the output an make sure it contains the string "Deleted nodes" if you want to verify something was actually deleted.

But functionality-wise, if you tell kind "I don't want a cluster named X", at the end you will not have a cluster named X. Whether there was one to begin with or not.

See also the [note in the docs](https://kind.sigs.k8s.io/docs/user/quick-start/#deleting-a-cluster) explicitly stating this expected behavior.

/hold

BenTheElder:
It's important to understand that "clusters" are not a single object and are not merely the nodes, they are also any additional assets like kubeconfig entries. The lack of nodes does not mean we didn't need to delete kubeconfig still.

You can detect if your cleanup script has the right name by if any nodes were logged as deleted, which we already support.

Thank you, but this has been discussed extensively in the issues and I don't think we'll be going forward with it. Also per our contributor guide we'd discuss a feature like this before reviewing PRs ordinarily.

pkg/cluster/internal/delete/delete.go
BenTheElder: Right, the idempotent nature is important to reliable CI because it makes it safe and even encouraged to have both the normal cleanup path and a signal / exit handler for cleanup without worrying about coordinating them. You can just cleanup on exit always and it's fine and reliable.

logging warnings like this will discourage implementing reliable cleanup and the output is misleading anyhow (a node or nodes is not the sum of a kind cluster)

sword-jin:
> It's important to understand that "clusters" are not a single object and are not merely the nodes, they are also any additional assets like kubeconfig entries. The lack of nodes does not mean we didn't need to delete kubeconfig still.
> 
> You can detect if your cleanup script has the right name by if any nodes were logged as deleted, which we already support.
> 
> Thank you, but this has been discussed extensively in the issues and I don't think we'll be going forward with it. Also per our contributor guide we'd discuss a feature like this before reviewing PRs ordinarily.

oh, thanks for the context, make sense, I'll close this PR

BenTheElder:
Thanks!

PR #3796, opened 2024-11-21
Added more kind config examples

3deep5me:
Hello, 
do you maybe want to add some more common examples? It would be great if you could have a look on my PR.
It could save some time for others. (Same for me if I have lost my configurations again 🤣)

### DIFF
--- a/site/static/examples/config-multi-node.yaml
+++ b/site/static/examples/config-multi-node.yaml
@@ -1,0 +1,7 @@
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+nodes:
+- role: control-plane
+- role: worker
+- role: worker
+- role: worker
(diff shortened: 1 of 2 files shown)


BenTheElder:
have you seen https://kind.sigs.k8s.io/docs/user/configuration/#nodes ?

3deep5me:
Hi @BenTheElder,
thanks for the hint - no haven't seen it. 😅
What do you think about the metrics example?

BenTheElder:
> What do you think about the metrics example?

Sorry we fell behind on replying here, I'm a bit worried about keeping it up to date, I would prefer to have a dedicated guide for installing metrics server, possibly hosted upstream with that project.

We've found that when we host guides ourselves for external projects they tend to get out of date because when changes are made to those projects, they don't get reflected in the guides.

Kubernetes's core docs site has a similar policy of strongly preferring to link out to external guides for external projects.

At the moment we don't support metrics server, metrics in kind may be inaccurate and metrics are NOT part of Kubernetes conformance currently.

I appreciate the PR though, thank you.

Issue #3797, opened 2024-11-22
I cannot use kind build node-image, complains about invalid directory and it true its not there


jamiemacdonald04:
**What happened**:

I tried to build and it failed.  See command below, no such file or directory

**What you expected to happen**:
A different version of k8s node to be created

**How to reproduce it (as _minimally_ and precisely as possible)**:
 kind build node-image --type url https://dl.k8s.io/v1.30.0/kubernetes-server-linux-arm64.tar.gz

**Anything else we need to know?**:
This is the output from the ocmmand 
```
Building using URL: "https://dl.k8s.io/v1.30.0/kubernetes-server-linux-arm64.tar.gz"
Starting to build Kubernetes
Downloading "https://dl.k8s.io/v1.30.0/kubernetes-server-linux-arm64.tar.gz"
Failed to build Kubernetes: open /var/folders/zr/tc5l708x4_s_dm1f4qtmpv4m0000gn/T/k8s-tar-extract-140925075/kubernetes/version: no such file or directory
```

**Environment:**

- kind version: (kind v0.25.0 go1.23.3 darwin/arm64
- Runtime info: Version:    25.0.3
- OS (e.g. from `/etc/os-release`): mac
- Kubernetes version: (use `kubectl version`): Client Version: v1.30.2
- Any proxies or other special environment settings?: No

stmcginnis:
It looks like the format of the published tarball has changed:

```txt
$ tree /tmp/k8s-tar-extract-2532324674/kubernetes/ -L 3
/tmp/k8s-tar-extract-2532324674/kubernetes/
├── kubernetes-src.tar.gz
├── LICENSES
│   ├── LICENSE
│   ├── OWNERS
│   ├── third_party
│   │   ├── forked
│   │   ├── gimme
│   │   └── multiarch
│   └── vendor
│       ├── bitbucket.org
│       ├── cloud.google.com
│       ├── github.com
│       ├── go.etcd.io
│       ├── golang.org
│       ├── google.golang.org
│       ├── go.opencensus.io
│       ├── go.opentelemetry.io
│       ├── gopkg.in
│       ├── go.starlark.net
│       ├── go.uber.org
│       ├── k8s.io
│       └── sigs.k8s.io
└── server
    └── bin
        ├── apiextensions-apiserver
        ├── kubeadm
        ├── kube-aggregator
        ├── kube-apiserver
        ├── kube-apiserver.docker_tag
        ├── kube-apiserver.tar
[...]

PR #3800, opened 2024-11-25
node build: Extract version from remote URL

stmcginnis:
At some point changes were made to the structure of the community published tarballs where the expected "version" file is no longer present. This causes the build to fail due to not being able to capture the version of the source being built.

Most of the time the requested version is part of the URL for the remote source anyway. This creates a fallback where the version will be extracted from the URL if it cannot be read explicitly from a "version" file in the tarball content.

This does not cover the local tarball file path, or if the user points to a non-community published URL. But in the "normal" usage of pointing to community published tarballs, this will prevent the node build from failing in the cases where there is not a version file present.

Closes: #3797

### DIFF
--- a/pkg/build/nodeimage/internal/kube/builder_remote.go
+++ b/pkg/build/nodeimage/internal/kube/builder_remote.go
@@ -24,11 +24,14 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"time"
 
 	"sigs.k8s.io/kind/pkg/log"
 )
+
+var versionRE = regexp.MustCompile(`^v\d+\.\d+\.\d+`)
 
 type remoteBuilder struct {
 	version string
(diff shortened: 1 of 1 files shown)


BenTheElder:
> At some point changes were made to the structure of the community published tarballs where the expected "version" file is no longer present. This causes the build to fail due to not being able to capture the version of the source being built.

This is likely a bug in the upstream release process and we should open an issue to discuss this with release engineering.

BenTheElder (commented):
This is a breaking change for anyone that was just hosting a tarball at an arbitrary path (versus upstream)

I would really prefer that our release artifacts actually contain the version information in a self-contained way, as a project, and I'm all but certain they did at some point ..

stmcginnis:
> This is a breaking change for anyone that was just hosting a tarball at an arbitrary path (versus upstream)

I don't think this is a breaking change. It will still try to read the version from an included "version" file. If that file is not present, and the URL does not contain a version number in the path, there will be no difference in behavior. I.e., it will still fail, or the version file is present and it will work exactly like it did before.

Agree we need to root cause the artifact issue, but think this could be useful either way.

stmcginnis:
This could be useful as a final fallback if all else fails, but I wasn't aware of the docker_tag available in the source. So looks like https://github.com/kubernetes-sigs/kind/pull/3815 is a good way to go.

PR #3801, opened 2024-11-25
bump containerd / runc to v1.7.23 / v1.1.14

zhangzujian:
Bump containerd to v1.7.22 to fix [pod deletion failure](https://github.com/kubeovn/kube-ovn/actions/runs/12002458122/job/33454663485) introduced in v1.7.14.

I cannot figure out which version/commit fixes the problem. After upgrading containerd to v1.7.22 no failure occurs anymore in [our test](https://github.com/kubeovn/kube-ovn/actions/runs/12006969275?pr=4760).

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -122,7 +122,7 @@ RUN eval "$(gimme "${GO_VERSION}")" \
 # stage for building containerd
 FROM go-build AS build-containerd
 ARG TARGETARCH GO_VERSION
-ARG CONTAINERD_VERSION="v1.7.18"
+ARG CONTAINERD_VERSION="v1.7.23"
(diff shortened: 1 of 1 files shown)


aojea:
the latest in the 1.7.x series is 1.7.23 https://github.com/containerd/containerd/releases/tag/v1.7.23 so I think is better to use that

/cc @BenTheElder

BenTheElder (commented):
anything of notes in the changelogs before we commit to this upgrade?

ditto runc

BenTheElder:
I've been digging through the release notes and working on an upgrade to 1.7.24 in https://github.com/kubernetes-sigs/kind/pull/3816

Issue #3795, opened 2024-11-20
Unable to load images to `kind` cluster


iamvinov-atlassian:
**What happened**:
I am attempting to load Docker images into `kind` cluster: `kind load docker-image busybox -n nebulae` but getting the following error:
`
❯ kind load docker-image busybox -n nebulae
Image: "busybox" with ID "sha256:5b0f33c83a97f5f7d12698df6732098b0cdb860d377f6307b68efe2c6821296f" not yet present on node "nebulae-control-plane", loading...
ERROR: failed to load image: command "docker exec --privileged -i nebulae-control-plane ctr --namespace=k8s.io images import --all-platforms --digests --snapshotter=overlayfs -" failed with error: exit status 1
Command Output: ctr: content digest sha256:83e82a8dd385e27d95f2118c1332d414684aa665552f7f837f86da33674308c4: not found
`

**What you expected to happen**:
I expected the image to load successfully. I already have this image pulled locally using `docker pull busybox`. Upon further investigation, it seems to me that `kind` or (`containerd`) expects the image for all platforms to be present on the host for the `load` command to succeed. 

**How to reproduce it (as _minimally_ and precisely as possible)**:
```
docker pull busybox
kind create cluster --name nebulae
kind load -v 10 docker-image -n nebulae busybox
```

**Anything else we need to know?**:
From looking at other answers on the internet, it seems generally this error occurs when the image arch doesn't match the host arch. But this is not the case. I did perform `docker images --tree` and made sure the images match my host (M3 MacBook Pro) OS.
```
busybox:latest                                                                         5b0f33c83a97       12.6MB            4MB
├─ linux/arm64/v8                                                                      6ca1ac3927a1       6.02MB         1.85MB
├─ linux/amd64                                                                         a3e1b257b47c       6.56MB         2.16MB
├─ linux/arm/v5                                                                        3076001161ce           0B             0B
├─ linux/arm/v6                                                                        a9fc789b4096           0B             0B
├─ linux/arm/v7                                                                        fb632082f5cb           0B             0B
├─ linux/386                                                                           c0d2f0e7a91f           0B             0B
├─ linux/mips64le                                                                      0e1d386b0b5d           0B             0B
├─ linux/ppc64le                                                                       fc082c5fdd21           0B             0B
├─ linux/riscv64                                                                       d55b3027f77f           0B             0B
└─ linux/s390x                                                                         4bc8b19fe938           0B             0B
```

**Environment:**

[...]

PR #3805, opened 2024-12-03
Use --local=false on amd64, to workaround containerd image loading

dgl:
This is a potential but not pretty workaround for #3795.

### DIFF
--- a/pkg/build/nodeimage/imageimporter.go
+++ b/pkg/build/nodeimage/imageimporter.go
@@ -18,6 +18,7 @@ package nodeimage
 
 import (
 	"io"
+	"runtime"
 
 	"sigs.k8s.io/kind/pkg/exec"
 )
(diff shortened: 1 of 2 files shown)


pkg/build/nodeimage/imageimporter.go
```diff
@@ -53,9 +54,16 @@ func (c *containerdImporter) Pull(image, platform string) error {
 }
 
 func (c *containerdImporter) LoadCommand() exec.Cmd {
-	return c.containerCmder.Command(
+	var opt string
+	// TODO: Hack to workaround #3795.
+	if runtime.GOARCH == "amd64" {
+		opt = "--local=false"
```
BenTheElder, line 60: can you elaborate a bit on what this is intended to do and why only amd64?

it will be a bit before I can dig into this

dgl: I dug a bit more into the actual issue on the issue thread, which will hopefully answer how to actually fix this in containerd / docker (which I do believe it is actually a bug in).

The reason for the arm64 / amd64 split is confusing and what makes this a hack:

* This is happening with packages where there is an amd64 / 386 mix, hence why targeting amd64 for `--local=false`.
* On the version of containerd kind is currently using `--all-platforms` with `--local=false` doesn't work.

So the bug has been seen on amd64, people using arm64 do want to mix platforms. It's also possible for arm64 to mix v8 or v9, but I think that's far less common and unlikely to break anyone.

Clearly this should be fixed in containerd (or docker, if it should actually include more blobs in its images), but given a kind release is tied to a containerd release I think it would be safe to include this hack until kind is on a containerd with a fix.

aojea: I don't like much the idea of we carry patching compensation for known bugs

BenTheElder: As a general rule we actually permit using them across releases and notify when a release has a compatibility change that requires images built after a certain release or something like that. It hasn't happened recently as well.

BenTheElder:
see also: https://github.com/kubernetes-sigs/kind/pull/3828#discussion_r1905958503 (containerd 2.0 and setting --local=true ...)

BenTheElder:
So if I followed correctly, after we finish adopting containerd 2.x, we will be using local=false on all platforms, with some upstream bugfixes. I think we should probably prefer that?

PR #3816, opened 2024-12-11
upgrade base image and tooling dependencies

BenTheElder:
I've been reading through the release notes for these one by one ...

- go 1.23 should be fine for kind, and appears to be supported (or even required) by the upgraded dependencies
- Containerd 1.7.x should be safe (reviewed release notes)
- runc makes me nervous but should be safe ... (kind does weird stuff ...)
- crictl should be safe
- containerd fuse-overlayfs doesn't mention any breaking changes
- cni-plugins should be safe

I think we should also do local-path-provisioner but I haven't finished reviewing that and I want to let some of the CI go ahead and test this much. NOTE: that doesn't yet include the github actions which won't build a node image.


BenTheElder:
looks like we'll have to upgrade golangci-lint

BenTheElder:
ok, go updates are going to be a slightly bigger can of worms ...

BenTheElder:
will revisit sometime tomorrow

BenTheElder:
https://github.com/opencontainers/runc/releases/tag/v1.2.3 dropped since I started this PR :-)

aojea:
runc looks important, several fixes for regressions

BenTheElder:
The good news is all the prow e2e CI passed flawlessly.

The bad news is https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kind/3816/pull-kind-verify/1866663406874923008

`go mod tidy` is starting to mutate hack/tools/go.mod with:
```patch
$ git diff
diff --git a/hack/tools/go.mod b/hack/tools/go.mod
index ee468e10..ce7cba29 100644
--- a/hack/tools/go.mod
+++ b/hack/tools/go.mod
@@ -1,6 +1,8 @@
 module sigs.k8s.io/kind/hack/tools
 
-go 1.17
+go 1.22.1
+
+toolchain go1.23.4
 
 require (
        github.com/golangci/golangci-lint v1.62.2
```

stmcginnis:
> go mod tidy is starting to mutate hack/tools/go.mod

Do we need to be as concerned with `hack/tools`? As long as the top level go.mod isn't being modified, it seems mostly safe to allow internal tooling to be different.

BenTheElder:
> Do we need to be as concerned with hack/tools? As long as the top level go.mod isn't being modified, it seems mostly safe to allow internal tooling to be different.

I think this is OK, but I haven't root-caused yet, and it means we'll have to start duplicating the toolchain version

BenTheElder:
I'm looking at `make generate` issues with deepcopy-gen changes.

BenTheElder:
Just updating the tools module language version to 1.23 works as expected, and I think that makes the most sense.

At some point we'll want to bump the main module up, but we don't have a strong use case for any of the new-fangled features for the CLI binary and we will break people doing `go get` with distro provided go packages until toolchain is more widely available in those versions. At that point it will mostly be about the go version for tools embedding the packages.

BenTheElder:
fixed `make generate` for gengo changes, re-bumping runc now ...

images/base/Dockerfile
```diff
@@ -140,7 +140,7 @@ RUN git clone --filter=tree:0 "${CONTAINERD_CLONE_URL}" /containerd \
 # stage for building runc
 FROM go-build AS build-runc
 ARG TARGETARCH GO_VERSION
-ARG RUNC_VERSION="v1.1.13"
+ARG RUNC_VERSION="v1.2.2"
```
aojea, line 143:
```suggestion
ARG RUNC_VERSION="v1.2.3"
```

?

BenTheElder: I intend to. this is still WIP, been following up on this nightmare: with unknown automation accessing k8s-ci-robot (which is owner/admin of all the github orgs) 🙃 https://kubernetes.slack.com/archives/C09QZ4DQB/p1733947637418569

BenTheElder:
aside: I don't generally recommend "big update all the dependencies" PRs, but these are high-touch and the last incoming PR was unresponsive about reviewing release notes. At this point a number of them are inter-related (like go1.23 and some of the base image bumps), and we don't risk disrupting 1.32 release anymore, and I have some time to review the downstream changes and test things out now ...

still WIP, will need new base image and other follow-ups

BenTheElder:
[I'm intending that we roll this + the `kind build node-image` fixes into a v0.26.0 that defaults to v1.32 kubernetes O(soon) but instead of blocking on that pushed a v1.32 image w/ v0.25.0 for now]

BenTheElder:
very happy to see actions (with assorted host runtimes etc) working with updated runc in particular via WIP commit node image (https://github.com/kubernetes-sigs/kind/pull/3816/commits/309db31deb938df235a5b6aeacc493ef5a838a86)

BenTheElder:
looks like golangci-lint timed out
/retest

BenTheElder:
golangci-lint timeout is not a flake. seems likely to be related to https://github.com/kubernetes-sigs/kind/pull/3816/commits/8f97ed0d01063ce758cedb2271d78fcee52d7fac given it was passing last night and the dockerfile changes should be unrelated

BenTheElder:
looks like it's while linting kindnetd

BenTheElder:
we did increase the size of that binary recently, maybe just bump the linter timeout. through 2m feels somewhat generous given the amount of in-repo source code.

BenTheElder:
the whole verify job only took 4m so I guess that's fine https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kind/3816/pull-kind-verify/1866986720868700160

BenTheElder:
OK, now everything is working.

I think we should also upgrade kindnetd and local-path-provisioner but this is already pretty huge.

I'm going to drop the temp image bump commits and we'll get new images built normally in postsubmit / follow-ups, I'm fairly satisfied that we've exercised runc+containerd in CI in particular.

BenTheElder:
I think let's split the rest into further PRs

BenTheElder:
/retest
flake on 1.31 https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kind/3816/pull-kind-e2e-kubernetes-1-31/1866999826344841216

NOTE: I dropped the image bump PRs, so that's not caused by this PR, just a flaky test

PR #3818, opened 2024-12-12
more upgrades for v0.26.0

BenTheElder:
1) pickup new base image (but not new node image yet, less disruptive when we're in-flight upgrading the rest, and we already tested a staging node image)

2) bump local-path-provisioner to build v0.0.30, I looked over this and will do an upgrade as the next PR, after the image builds. there are some nice improvements and it should be compatible based on release notes and the manifest etc.

PR #3820, opened 2024-12-12
fix local-path-provisioner build

BenTheElder:
see: https://github.com/rancher/local-path-provisioner/blob/master/scripts/build

https://github.com/kubernetes-sigs/kind/pull/3818#issuecomment-2540083143

### DIFF
--- a/images/local-path-provisioner/Dockerfile
+++ b/images/local-path-provisioner/Dockerfile
@@ -18,13 +18,14 @@ COPY --chmod=0755 scripts/third_party/gimme/gimme /usr/local/bin/
 RUN git clone --filter=tree:0 https://github.com/rancher/local-path-provisioner
 ARG VERSION
 # set by makefile to .go-version
+# TODO: scripts/build builds for multiple platforms, so we waste a little time here
 ARG TARGETARCH GO_VERSION
 RUN eval "$(gimme "${GO_VERSION}")" \
     && export GOTOOLCHAIN="go${GO_VERSION}" \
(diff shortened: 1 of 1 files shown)


images/local-path-provisioner/Dockerfile
```diff
@@ -18,13 +18,14 @@ COPY --chmod=0755 scripts/third_party/gimme/gimme /usr/local/bin/
...
 # set by makefile to .go-version
+# TODO: scripts/build builds for multiple platforms, so we waste a little time here
 ARG TARGETARCH GO_VERSION
 RUN eval "$(gimme "${GO_VERSION}")" \
     && export GOTOOLCHAIN="go${GO_VERSION}" \
     && cd local-path-provisioner \
     && git fetch && git checkout "${VERSION}" \
     && GOARCH=$TARGETARCH scripts/build \
-    && mv bin/local-path-provisioner /usr/local/bin/local-path-provisioner \
+    && mv bin/local-path-provisioner-$TARGETARCH /usr/local/bin/local-path-provisioner \
```
BenTheElder, line 28: basically they started building multiple arches and putting -$GOARCH at the end

there isn't a build target for one arch, we don't build this that often so ... just continue using the upstream build logic and adapt copying the binary to match the arch-specific file names

BenTheElder:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/post-kind-push-local-path-provisioner-image/1867334732086251520 🟢

PR #3824, opened 2024-12-13
update default node image to 1.32 with upgraded containerd etc

BenTheElder:
tested locally with rootful docker


BenTheElder:
we had a DNS flake in one of the actions that seems to be on the VM side for cgroup v2, not the image etc

https://github.com/kubernetes-sigs/kind/actions/runs/12321896969/job/34394268047?pr=3824

rerunning it

PR #3827, opened 2024-12-16
mark v0.26.0

BenTheElder:
I'm going to automate this a little bit more after the release is done ...

Already drafted notes from https://github.com/kubernetes-sigs/kind/compare/v0.25.0...main and built images (so much faster with #3825 🎉 thanks @dims)


aojea:
👏

PR #3828, opened 2024-12-20
update containerd to v2.0.2

### LINKED ISSUE #3768: Containerd 2.0 support
Containerd is close to release a new major version 2.0 and this will may bring some changes 

Per example, I had to test something with containerd 2.0 and the `kind load docker-images` command fails with

```
kind load docker-image ghcr.io/aojea/kindnetd:v1.6.0
Image: "ghcr.io/aojea/kindnetd:v1.6.0" with ID "sha256:03579ba24d18613b5f6af6fe639a877d59e899ce88214598ef228e98484b0354" not yet present on node "kind-worker", loading...
Image: "ghcr.io/aojea/kindnetd:v1.6.0" with ID "sha256:03579ba24d18613b5f6af6fe639a877d59e899ce88214598ef228e98484b0354" not yet present on node "kind-worker2", loading...
Image: "ghcr.io/aojea/kindnetd:v1.6.0" with ID "sha256:03579ba24d18613b5f6af6fe639a877d59e899ce88214598ef228e98484b0354" not yet present on node "kind-control-plane", loading...
ERROR: failed to detect containerd snapshotter
```

```[tasklist]
### Tasks
- [ ] kind load docker-images for containerd 2.0
- [ ] add CI job with containerd 2.0
```

AkihiroSuda:
Fix #3768

### DIFF
--- a/pkg/cluster/nodeutils/util.go
+++ b/pkg/cluster/nodeutils/util.go
@@ -19,6 +19,7 @@ package nodeutils
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"io"
 	"path"
 	"strings"
@@ -102,9 +103,24 @@ func parseSnapshotter(config string) (string, error) {
 	if err != nil {
 		return "", errors.Wrap(err, "failed to detect containerd snapshotter")
 	}
-	snapshotter, ok := parsed.GetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "containerd", "snapshotter"}).(string)
+	configVersion, ok := parsed.Get("version").(int64)
 	if !ok {
-		return "", errors.New("failed to detect containerd snapshotter")
+		return "", errors.New("failed to detect containerd config version")
+	}
+	var snapshotter string
+	switch configVersion {
+	case 2: // Introduced in containerd v1.3. Still supported in containerd v2.
+		snapshotter, ok = parsed.GetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "containerd", "snapshotter"}).(string)
+		if !ok {
+			return "", errors.New("failed to detect containerd snapshotter (config version 2)")
+		}
+	case 3: // Introduced in containerd v2.0.
+		snapshotter, ok = parsed.GetPath([]string{"plugins", "io.containerd.cri.v1.images", "snapshotter"}).(string)
+		if !ok {
+			return "", errors.New("failed to detect containerd snapshotter (config version 3)")
+		}
+	default:
+		return "", fmt.Errorf("unknown containerd config version: %d (supported versions: 2 and 3)", configVersion)
 	}
 	return snapshotter, nil
 }
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -122,7 +122,7 @@ RUN eval "$(gimme "${GO_VERSION}")" \
 # stage for building containerd
 FROM go-build AS build-containerd
 ARG TARGETARCH GO_VERSION
-ARG CONTAINERD_VERSION="v1.7.24"
+ARG CONTAINERD_VERSION="v2.0.2"
 ARG CONTAINERD_CLONE_URL="https://github.com/containerd/containerd"
 # we don't build with optional snapshotters, we never select any of these
 # they're not ideal inside kind anyhow, and we save some disk space


stmcginnis:
I plan on doing some more local testing, but this looks good to me and CI is happy.

/lgtm

pkg/cluster/nodeutils/util.go
```diff
@@ -82,7 +83,9 @@ func LoadImageArchive(n nodes.Node, image io.Reader) error {
 	if err != nil {
 		return err
 	}
-	cmd := n.Command("ctr", "--namespace=k8s.io", "images", "import", "--all-platforms", "--digests", "--snapshotter="+snapshotter, "-").SetStdin(image)
+	// --local=true: Disables the Transfer API mode that is enabled by default since containerd v2.0.
+	// Workaround for `ctr: rpc error: code = InvalidArgument desc = unable to initialize unpacker: no unpack platforms defined: invalid argument`
```
BenTheElder, line 87: We probably need a long term solution for this?

pkg/cluster/nodeutils/util.go
```diff
@@ -102,9 +105,24 @@ func parseSnapshotter(config string) (string, error) {
...
 	}
-	snapshotter, ok := parsed.GetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "containerd", "snapshotter"}).(string)
+	configVersion, ok := parsed.Get("version").(int64)
 	if !ok {
-		return "", errors.New("failed to detect containerd snapshotter")
+		return "", errors.New("failed to detect containerd config version")
+	}
+	var snapshotter string
+	switch configVersion {
+	case 2: // Introduced in containerd v1.3. Still supported in containerd v2.
```
BenTheElder, line 114: we probably need to allow user supplied config patches to similarly target 2 vs 3 separately, though that doesn't need to be done in this PR

aojea:
/lgtm

I'm positively surprised the diff is so small ... IIRC we already had issues with `ctr` and storage formats in the past

pkg/cluster/nodeutils/util.go
aojea: do we depend on the version of the config?

https://github.com/kubernetes-sigs/kind/blob/90d943900d5fe34fdbcea46dbe69abecb8015331/pkg/cluster/internal/create/actions/config/config.go#L119-L146

BenTheElder: > do we depend on the version of the config?

user's patches will not work as intended, they will wind up merging in keys that are not valid for v3 because all the keys have new namespacing again.

BenTheElder: (and if they simply switch to v3 keys, it will not work with older kind images)

AkihiroSuda:
>  since CRI only has v1 and not v1alpha2 in containerd v2, we break kubernetes < 1.26


CRI v1 seems introduced in Kubernetes 1.20: https://github.com/kubernetes/kubernetes/commit/9fcede9d5b37fccfe3269bd93559c551e47de5ff

aojea:
> and since CRI only has v1 and not v1alpha2 in containerd v2, we break kubernetes < 1.26 (which is probably OK, but worth noting, we have not actively broken an old k8s version for some time now).

v1 was added in 1.20  https://github.com/kubernetes/kubernetes/pull/96387 
it was made default in 1.23 https://github.com/kubernetes/kubernetes/pull/106501
In 1.26 v1alpha2 was dropped https://github.com/kubernetes/kubernetes/pull/110618,  but we should support until 1.20 IIUIC

pkg/cluster/nodeutils/util.go
aojea: ah, so you refer to the "existing" user patches, I think is a matter of how you see it then ... for me those are containerd specific patches, and yes, users need to adapt them to containerd 2.0 unless there is some containerd feature that allows to migrate between formats

BenTheElder: >  for me those are containerd specific patches, and yes, users need to adapt them to containerd 2.0 unless there is some containerd feature that allows to migrate between formats

But then their config won't work for older kind releases/images.

The way we do this with kubeadm config patches is you can have multiple patches that are targeting different kubeadm API versions. So far that's been irrelevant because for nearly all of our history we've only had containerd 2.0 config.

BenTheElder: But actually, that is for a future PR, because in this PR we're still using a v2.0 config (`images/base/files/etc/containerd/config.toml`), this part of the diff seems to be technically unecessary for now.

BenTheElder:
> CRI v1 seems introduced in Kubernetes 1.20: https://github.com/kubernetes/kubernetes/commit/9fcede9d5b37fccfe3269bd93559c551e47de5ff

ACK, definitely no problem with that.

We will still need to make sure to note this for the release, and we should consider if this is sufficient excuse to drop other < 1.20 bits from the project. We've been very lenient about dropping support outright so far.

And re: other parts: https://github.com/kubernetes-sigs/kind/pull/3828#discussion_r1906139310

pkg/cluster/nodeutils/util.go
BenTheElder: Also: https://github.com/kubernetes-sigs/kind/pull/3805

pkg/cluster/nodeutils/util.go
aojea: I see, I got it know, so we keep using containerd 2.0 version for the config and let containerd deal with the translation internall to version 3 https://github.com/containerd/containerd/blob/main/docs/cri/config.md#config-versions

BenTheElder: Yeah, before we switch our config to v3 we should give users a way to deal with it smoothly, but we're not actually switching to v3 in this PR, even though this PR adds partial support for v3.

pkg/cluster/nodeutils/util.go
AkihiroSuda: Opened an issue:
- https://github.com/containerd/containerd/issues/11228

AkihiroSuda: PR
- https://github.com/containerd/containerd/pull/11229

pkg/build/nodeimage/imageimporter.go
```diff
@@ -55,7 +55,8 @@ func (c *containerdImporter) Pull(image, platform string) error {
 func (c *containerdImporter) LoadCommand() exec.Cmd {
 	return c.containerCmder.Command(
 		// TODO: ideally we do not need this in the future. we have fixed at least one image
-		"ctr", "--namespace=k8s.io", "images", "import", "--label=io.cri-containerd.pinned=pinned", "--all-platforms", "--no-unpack", "--digests", "-",
+		// --local=true: see the comment in pkg/cluster/nodeutils/util.go
```
BenTheElder, line 58: we set --no-unpack, isn't the issue in the other location with unpacking?

pkg/cluster/nodeutils/util.go
BenTheElder: thank you!

AkihiroSuda: Updated to v2.0.2

aojea:
/lgtm
/assign @BenTheElder 

Thanks

BenTheElder:
/lgtm
/approve
/hold cancel

Note: we have automated base image builds *after merge*, so most of the CI here is using the current base image still. After this merges we will send another PR to pickup the image, it's possible we'll find issues and revert. This is fine, just pointing it out in case.

Sometimes we preemptively test by temporarily adding a commit using a locally built & pushed image, but only to test. We don't need to do that here, but we will get more signal when we adopt a built image.

BenTheElder:
Thanks for working on this and the upstream bug fixes, as always!

BenTheElder:
filed https://github.com/kubernetes-sigs/kind/pull/3848 after the build job completed so we can keep this moving along

BenTheElder:
I did find an issue https://github.com/kubernetes-sigs/kind/pull/3848#issuecomment-2599215072

We can mitigate, but the underlying cause may be worth investigating.

Previously we did not wait for containerd to be ready because it was fully booted in ~0.3s even when cross-architecture, so we never had any issue, now that takes approximately 1s so image pulls fail because the socket is not available yet.

Mitigating that would be pretty trivial (we can either retry image pulls which we should consider anyhow, watch for containerd to be ready before doing any pulls, or both), but that seems like a pretty big regression that will impact startup time at runtime as well (both initially, and anytime we need to restart). Not sure if that's something the containerd project would track or just accept.

cc @samuelkarp

BenTheElder:
(These times are v1.7.24 vs v2.0.2, arm64 builds on an amd64 GCE machine, more details in https://github.com/kubernetes-sigs/kind/pull/3848#issuecomment-2599215072)

BenTheElder:
Update: this is probably not worth discussing upstream, because 2.0.2 is still < 0.07s with amd64 + amd64 host. It is however consistently longer than 1.7.4. Something with arm64 qemu must be even more pathological.

BenTheElder:
Moving further discussion to https://github.com/kubernetes-sigs/kind/pull/3848

PR #3838, opened 2025-01-05
Add description to delete cluster subcommands

stmcginnis:
This adds a more detailed description to the `kind delete cluster` and `kind delete clusters` commands explaining expected delete behavior.

Hopefully this reduces some of the regular confusion from users submitting issues due to expecting an error deleting a cluster that does not exist.

Updated command output:

```txt
$ kind delete cluster --help
Deletes a Kind cluster from the system.

This is an idempotent operation, meaning it may be called multiple times without
failing. If the cluster exists it will be deleted, and if the cluster is already
gone it will just return success. Errors will only occur if the cluster exists
and was not able to be deleted.

Usage:
  kind delete cluster [flags]

Flags:
  -h, --help                help for cluster
      --kubeconfig string   sets kubeconfig path instead of $KUBECONFIG or $HOME/.kube/config
  -n, --name string         the cluster name (default "kind")

Global Flags:
  -q, --quiet             silence all stderr output
  -v, --verbosity int32   info log verbosity, higher value produces more output
```

### DIFF
--- a/pkg/cmd/kind/delete/cluster/deletecluster.go
+++ b/pkg/cmd/kind/delete/cluster/deletecluster.go
@@ -38,11 +38,17 @@ type flagpole struct {
 func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 	flags := &flagpole{}
 	cmd := &cobra.Command{
-		Args: cobra.NoArgs,
-		// TODO(bentheelder): more detailed usage
+		Args:  cobra.NoArgs,
 		Use:   "cluster",
 		Short: "Deletes a cluster",
-		Long:  "Deletes a resource",
+		Long: `Deletes a Kind cluster from the system.
+
+This is an idempotent operation, meaning it may be called multiple times without
+failing (like "rm -f"). If the cluster resources exist they will be deleted, and
+if the cluster is already gone it will just return success.
+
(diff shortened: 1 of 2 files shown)


pkg/cmd/kind/delete/cluster/deletecluster.go
```diff
@@ -39,10 +39,15 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
...
-		// TODO(bentheelder): more detailed usage
 		Use:   "cluster",
 		Short: "Deletes a cluster",
-		Long:  "Deletes a resource",
+		Long:  `Deletes a Kind cluster from the system.
+
+This is an idempotent operation, meaning it may be called multiple times without
+failing. If the cluster exists it will be deleted, and if the cluster is already
+gone it will just return success. Errors will only occur if the cluster exists
+and was not able to be deleted.
```
BenTheElder, line 49:
```suggestion
This is an idempotent operation, meaning it may be called multiple times without
failing (like `rm -f`). If cluster resources exist they will be deleted, and if the cluster is already
gone it will just return success. Errors will only occur if cluster resource exist
and are not able to be deleted.
```

BenTheElder (commented):
thanks! some suggestions + gofmt is required

BenTheElder (commented):
/lgtm
/approve

PR #3843, opened 2025-01-13
Refactor: improve shell scripts

namejsjeongkr:
- These changes aim to make the scripts more robust, easier to maintain, and more efficient while ensuring they follow the project's established coding patterns. (Please refer to the below key changes.)

- e2e-k8s.sh
Enhanced error handling by adding set -o pipefail.
Improved code structure by extracting repeated tasks into functions.
Added informative log messages for better execution tracking.
Standardized variable usage to ${VARIABLE} format for consistency.
Retained temporary directory creation and cleanup logic for better resource management.

- push-latest-cli.sh
Updated error handling to use set -o errexit -o nounset -o pipefail.
Introduced an upload_file function to reduce code duplication.
Implemented use of gsutil -m cp for parallel uploads to improve performance.
Added log messages at key steps for clearer execution tracking.
Standardized variable declarations using double quotes.

### DIFF
--- a/hack/ci/e2e-k8s.sh
+++ b/hack/ci/e2e-k8s.sh
@@ -13,164 +13,83 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# hack script for running a kind e2e
-# must be run with a kubernetes checkout in $PWD (IE from the checkout)
-# Usage: SKIP="ginkgo skip regex" FOCUS="ginkgo focus regex" kind-e2e.sh
-
 set -o errexit -o nounset -o xtrace
 
-# Settings:
-# SKIP: ginkgo skip regex
-# FOCUS: ginkgo focus regex
-# LABEL_FILTER: ginkgo label query for selecting tests (see "Spec Labels" in https://onsi.github.io/ginkgo/#filtering-specs)
-#
-# The default is to focus on conformance tests. Serial tests get skipped when
-# parallel testing is enabled. Using LABEL_FILTER instead of combining SKIP and
-# FOCUS is recommended (more expressive, easier to read than regexp).
-#
-# GA_ONLY: true  - limit to GA APIs/features as much as possible
-#          false - (default) APIs and features left at defaults
-# FEATURE_GATES:
-#          JSON or YAML encoding of a string/bool map: {"FeatureGateA": true, "FeatureGateB": false}
-#          Enables or disables feature gates in the entire cluster.
-#          Cannot be used when GA_ONLY=true.
-# RUNTIME_CONFIG:
-#          JSON or YAML encoding of a string/string (!) map: {"apia.example.com/v1alpha1": "true", "apib.example.com/v1beta1": "false"}
-#          Enables API groups in the apiserver via --runtime-config.
-#          Cannot be used when GA_ONLY=true.
+SKIP="${SKIP:-}"
+FOCUS="${FOCUS:-}"
+LABEL_FILTER="${LABEL_FILTER:-}"
+GA_ONLY="${GA_ONLY:-false}"
+FEATURE_GATES="${FEATURE_GATES:-{\}}"
+RUNTIME_CONFIG="${RUNTIME_CONFIG:-{\}}"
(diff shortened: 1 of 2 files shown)


hack/ci/e2e-k8s.sh
```diff
@@ -13,303 +13,184 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# hack script for running a kind e2e
-# must be run with a kubernetes checkout in $PWD (IE from the checkout)
-# Usage: SKIP="ginkgo skip regex" FOCUS="ginkgo focus regex" kind-e2e.sh
-
 set -o errexit -o nounset -o xtrace
 
-# Settings:
```
stmcginnis, line 22: I don't think all of the comments should be removed.

hack/ci/e2e-k8s.sh
```diff
@@ -13,303 +13,184 @@
...
-    kind "export" logs "${ARTIFACTS}" || true
-    kind delete cluster || true
-  fi
-  rm -f _output/bin/e2e.test || true
-  # remove our tempdir, this needs to be last, or it will prevent kind delete
-  if [ -n "${TMP_DIR:-}" ]; then
-    rm -rf "${TMP_DIR:?}"
-  fi
-  CLEANED_UP=true
+    if [ "${CLEANED_UP:-false}" = "true" ]; then
```
stmcginnis, line 36: There are a lot of unnecessary whitespace changes included that make this difficult to review.

stmcginnis (commented):
What is the motivation for these changes? These scripts are used quite extensively, for some time now, without any known issues I'm aware of.

hack/ci/push-latest-cli/push-latest-cli.sh
```diff
@@ -14,18 +14,25 @@
...
 
 set -o errexit -o nounset -o pipefail
-set -x;
+set -x
+
+# Function: Upload file to GCS bucket
+upload_file() {
+    local file="$1"
+    local version="$2"
+    gsutil -m cp -P "${file}" "gs://${BUCKET}/${version}/$(basename "${file}")"
```
BenTheElder, line 23: `-m` doesn't make sense for small, single file interactions?

And otherwise this was just `gsutil cp -P <call site specific arguments>` which is clearer inline, similar to `cp -R ...`

This seems over-abstracted

BenTheElder (commented):
Thanks, but this is hard to review with the large amount of unnecessary changes.

> e2e-k8s.sh
> Enhanced error handling by adding set -o pipefail.
Didn't happen? Also this script is written in POSIX shell (not bash, unlike most of thme) for wide-usage, which doesn't have pipefail. (note the shebang)

> Improved code structure by extracting repeated tasks into functions.

... this is debatable, a lot of straightforward inline context was removed.

> Introduced an upload_file function to reduce code duplication.

Invoking a command with a single flag and two call-site-specific arguments is hardly code duplication?

> Implemented use of gsutil -m cp for parallel uploads to improve performance.

I'm doubtful that this is faster for these small files being uploaded one at a time, do we have evidence? 
Anyhow we're not having a performance issue here, this script runs in automation and runtime is dominated by cloning and compiling. https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/logs/post-kind-push-binaries

> push-latest-cli.sh
> Updated error handling to use set -o errexit -o nounset -o pipefail.
Huh? we already were?

> Standardized variable declarations using double quotes.

Double quotes should only be used when expansion is being performed. Single quotes prevents accidental expansion (or unquoted when forming an array).

hack/ci/push-latest-cli/push-latest-cli.sh
```diff
@@ -34,36 +41,42 @@ SHORT_COMMIT="${COMMIT:0:8}"
 BUCKET="${BUCKET:-k8s-staging-kind}"
 # under each of these
 VERSIONS=(
-  latest
-  "${SHORT_COMMIT}"
+    "latest"
```
BenTheElder, line 44: these are arguments in an array, and we want the literal `latest` as a distinct word, quoting here is superfluous, as a literal it doesn't need double quotes

PR #3848, opened 2025-01-17
update to latest base image with containerd 2.0.2, ensure containerd is ready before importing images

BenTheElder:
contains the base image built from https://github.com/kubernetes-sigs/kind/pull/3828 in https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/post-kind-push-base-image/1880330906284068864

TODO: node image (prow e2es will use this, github actions will use the default node image only, though I expect we are more likely to catch issues in the full kubernetes e2e tests anyhow for this particular change)

### DIFF
--- a/pkg/build/nodeimage/buildcontext.go
+++ b/pkg/build/nodeimage/buildcontext.go
@@ -267,18 +267,22 @@ func (c *buildContext) prePullImagesAndWriteManifests(bits kube.Bits, parsedVers
 	}()
 
 	fns := []func() error{}
+	osArch := dockerBuildOsAndArch(c.arch)
 	for _, image := range requiredImages {
 		image := image // https://golang.org/doc/faq#closures_and_goroutines
-		fns = append(fns, func() error {
-			if !builtImages.Has(image) {
-				if err = importer.Pull(image, dockerBuildOsAndArch(c.arch)); err != nil {
+		if !builtImages.Has(image) {
+			fns = append(fns, func() error {
+				if err = importer.Pull(image, osArch); err != nil {
 					c.logger.Warnf("Failed to pull %s with error: %v", image, err)
 					runE := exec.RunErrorForError(err)
 					c.logger.Warn(string(runE.Output))
+					c.logger.Warnf("Retrying %s pull after 1s ...", image)
+					time.Sleep(time.Second)
+					return importer.Pull(image, osArch)
 				}
-			}
-			return nil
-		})
+				return nil
+			})
+		}
 	}
 	if err := errors.AggregateConcurrent(fns); err != nil {
 		return nil, err
--- a/pkg/build/nodeimage/buildcontext.go
+++ b/pkg/build/nodeimage/buildcontext.go
@@ -284,6 +284,11 @@ func (c *buildContext) prePullImagesAndWriteManifests(bits kube.Bits, parsedVers
 			})
 		}
 	}
+	// Wait for containerd socket to be ready, which may take 1s when running under emulation
+	if err := importer.WaitForReady(); err != nil {
+		c.logger.Errorf("Image build failed, containerd did not become ready %v", err)
+		return nil, err
+	}
 	if err := errors.AggregateConcurrent(fns); err != nil {
 		return nil, err
 	}
--- a/pkg/apis/config/defaults/image.go
+++ b/pkg/apis/config/defaults/image.go
@@ -18,4 +18,4 @@ limitations under the License.
 package defaults
 
 // Image is the default for the Config.Image field, aka the default node image.
-const Image = "kindest/node:v1.32.0@sha256:c48c62eac5da28cdadcf560d1d8616cfa6783b58f0d94cf63ad1bf49600cb027"
+const Image = "kindest/node:v1.32.1@sha256:6afef2b7f69d627ea7bf27ee6696b6868d18e03bf98167c420df486da4662db6"
--- a/pkg/build/nodeimage/defaults.go
+++ b/pkg/build/nodeimage/defaults.go
@@ -22,4 +22,4 @@ const DefaultImage = "kindest/node:latest"
 // DefaultBaseImage is the default base image used
 // TODO: come up with a reasonable solution to digest pinning
 // https://github.com/moby/moby/issues/43188
-const DefaultBaseImage = "docker.io/kindest/base:v20241212-9f82dd49"
+const DefaultBaseImage = "docker.io/kindest/base:v20250117-f528b021"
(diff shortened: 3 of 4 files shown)


BenTheElder:
/retest
[pod scheduling timeout, CI never ran]

BenTheElder:
All of CI passed the first try (ignoring failure to schedule the CI workload itself, unrelated, timing mismatch with autoscaling vs job scheduler), however when building a node image locally:

```
Failed to pull docker.io/kindest/kindnetd:v20241212-9f82dd49 with error: command "docker exec --privileged kind-build-1737147227-770353019 ctr --namespace=k8s.io content fetch --platform=linux/arm64 docker.io/kindest/kindnetd:v20241212-9f82dd49" failed with error: exit status 1
time="2025-01-17T20:53:53Z" level=warning msg="Failed to check deprecations" error="connection error: desc = \"transport: Error while dialing: dial unix:///run/containerd/containerd.sock: timeout\""
ctr: connection error: desc = "transport: Error while dialing: dial unix:///run/containerd/containerd.sock: timeout"
Failed to pull docker.io/kindest/local-path-provisioner:v20241212-8ac705d0 with error: command "docker exec --privileged kind-build-1737147227-770353019 ctr --namespace=k8s.io content fetch --platform=linux/arm64 docker.io/kindest/local-path-provisioner:v20241212-8ac705d0" failed with error: exit status 1
time="2025-01-17T20:53:53Z" level=warning msg="Failed to check deprecations" error="connection error: desc = \"transport: Error while dialing: dial unix:///run/containerd/containerd.sock: timeout\""
ctr: connection error: desc = "transport: Error while dialing: dial unix:///run/containerd/containerd.sock: timeout"
Failed to pull docker.io/kindest/local-path-helper:v20241212-8ac705d0 with error: command "docker exec --privileged kind-build-1737147227-770353019 ctr --namespace=k8s.io content fetch --platform=linux/arm64 docker.io/kindest/local-path-helper:v20241212-8ac705d0" failed with error: exit status 1
time="2025-01-17T20:53:53Z" level=warning msg="Failed to check deprecations" error="connection error: desc = \"error reading server preface: read unix @->/run/containerd/containerd.sock: use of closed network connection\""
ctr: connection error: desc = "error reading server preface: read unix @->/run/containerd/containerd.sock: use of closed network connection"
```

Debugging, my current suspicion is that we need to wait for containerd to be ready, it takes longer to start?

Around 1s on a fairly large cloud VM:
> INFO[2025-01-17T20:58:38.686355016Z] containerd successfully booted in 0.956085s`

If i start containerd in the previous base image like this to simulate the build process:

```
docker run -d --entrypoint=sleep --name="test-old-base" --platform=linux/arm64 --security-opt=seccomp=unconfined docker.io/kindest/base:v20241212-9f82dd49 infinity
docker exec -it test-old-base containerd
```

>  INFO[2025-01-17T21:03:28.470803043Z] containerd successfully booted in 0.357321s

These times are pretty representative of repeated attempts, containerd 2.0.2 takes about 3x to start versus v1.7.24

(note, these are the arm64 cross-build on an amd64 host)

BenTheElder:
To replicate:
1) start both versions
```console
docker run -d --entrypoint=sleep --name="test-old-base" --platform=linux/arm64 --security-opt=seccomp=unconfined docker.io/kindest/base:v20241212-9f82dd49 infinity
docker run -d --entrypoint=sleep --name="test-new-base" --platform=linux/arm64 --security-opt=seccomp=unconfined docker.io/kindest/base:v20250117-f528b021 infinity
```
2) try starting containerd in each of these:
```console
docker exec -it test-old-base containerd
```

```console
docker exec -it test-new-base containerd
```
(then watch for the log line like "containerd successfully booted in 0.972910s")

BenTheElder:
Update: this is probably not worth discussing upstream, because 2.0.2 is still < 0.07s with amd64 + amd64 host. It is however consistently longer than 1.7.4. Something with arm64 qemu must be even more pathological.

I think let's add image pull retries + waiting for it to start. We only need to do this for pulling, not imports, which is a nice idea anyhow to handle transient network issues.

aojea:
> Update: this is probably not worth discussing upstream, because 2.0.2 is still < 0.07s with amd64 + amd64 host. It is however consistently longer than 1.7.4. Something with arm64 qemu must be even more pathological.
> 
> I think let's add image pull retries + waiting for it to start. We only need to do this for pulling, not imports (we do all pulling first), which is a nice idea anyhow to handle transient network issues.

cc: @samuelkarp @AkihiroSuda 

to put this in their radar

BenTheElder:
I don't think this is a valid upstream issue, on further digging, it's only present under emulation (arm64 on amd64 host), or at least I don't currently have access to verify on an arm64 host. On amd64 2.x is technically slower than 1.7 but with a much smaller difference (both in absolute and relative terms).

aojea:
/lgtm

unhold once you feel we are ready

BenTheElder:
> unhold once you feel we are ready

iterating on retry/wait for ready as a fix, we need the images to build multi-arch.

the patch I just pushed is sufficient to get a working build but it's just a quick hack.

`ctr --connect-timeout` doesn't seem to work, if I run `ctr --connect-timeout=30s --timeout=30s info` without containerd running it fails just as quickly as without those flags (appears to be 1s, though the usage suggests it should be 0s)

BenTheElder:
OK, it both waits for containerd to be ready AND implements retry on image pull now.

Next I'm pushing a test node image we can try in the github actions.

BenTheElder:
Everything seems to work locally, and all the github actions are passing with the node image pushed to staging, so I'll promote the node image and I think we can proceed, barring review comments.

pkg/build/nodeimage/buildcontext.go
```diff
@@ -267,18 +267,27 @@ func (c *buildContext) prePullImagesAndWriteManifests(bits kube.Bits, parsedVers
...
+					return importer.Pull(image, osArch)
 				}
-			}
-			return nil
-		})
+				return nil
+			})
+		}
+	}
+	// Wait for containerd socket to be ready, which may take 1s when running under emulation
```
BenTheElder, line 287: we could also consider starting the image importer earlier on, but I would prefer to get the functional fix merged first over further micro-optimization. it won't make any meaningful difference in the 99% case of users building for their host architecture anyhow, and even when it does it will be negligible, basically one `sleep` in WaitForReady vs 0.

BenTheElder:
One more issue cropped up: https://github.com/kubernetes/kubernetes/issues/129818

This one is for an alpha feature though.

PR #3850, opened 2025-01-19
test: upgrade nerdctl to v2

tao12345666333:
> This release of nerdctl is expected to be used with containerd v1.6, v1.7, or v2.0.
> https://github.com/containerd/nerdctl/releases/tag/v2.0.2

The nerdctl v2.x is also expected to be compatible with containerd v2. 

This pull request includes updates to version numbers for dependencies in the `nerdctl` GitHub workflow and the default base image in the `nodeimage` package.

Version updates:

* [`.github/workflows/nerdctl.yaml`](diffhunk://#diff-4e2320dfb704eab40a465b6a57332f8c4b000bc4dd27248ac3a7f64b8b0c766eL29-R29): Updated `NERDCTL_VERSION` from "1.7.4" to "2.0.2".
~* [`pkg/build/nodeimage/defaults.go`](diffhunk://#diff-4283c324e37748a6ec55fe240d2d06561a180ca2b170d684b30edd2d9564e743L25-R25): Updated `DefaultBaseImage` from "docker.io/kindest/base:v20241212-9f82dd49" to "docker.io/kindest/base:v20250117-f528b021".~

### DIFF
--- a/.github/workflows/nerdctl.yaml
+++ b/.github/workflows/nerdctl.yaml
@@ -26,7 +26,7 @@ jobs:
     env:
       JOB_NAME: "nerdctl-${{ matrix.deployment }}-${{ matrix.ipFamily }}"
       IP_FAMILY: ${{ matrix.ipFamily }}
-      NERDCTL_VERSION: "1.7.4"
+      NERDCTL_VERSION: "2.0.2"
       KIND_EXPERIMENTAL_PROVIDER: "nerdctl"
     steps:
       - name: Check out code into the Go module directory
@@ -91,7 +91,7 @@ jobs:
       - name: Export logs
         if: always()
         run: |
-          sudo cat /etc/cni/net.d/* 
+          sudo find /etc/cni/net.d/ -type f -exec sh -c 'echo "{}" && cat "{}"' \;
           sudo mkdir -p /tmp/kind/logs
           sudo /usr/local/bin/kind export logs /tmp/kind/logs
           sudo chown -R $USER:$USER /tmp/kind/logs


tao12345666333:
The CI failure was due to using cat to read the configuration in this directory within the workflow. However, a default directory was added in v2.

I will fix it.

```
sudo tree  /etc/cni/net.d/
/etc/cni/net.d/
├── default
│   └── nerdctl-kind.conflist
└── nerdctl-bridge.conflist

2 directories, 2 files
```

.github/workflows/nerdctl.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
     env:
       JOB_NAME: "nerdctl-${{ matrix.deployment }}-${{ matrix.ipFamily }}"
       IP_FAMILY: ${{ matrix.ipFamily }}
-      NERDCTL_VERSION: "1.7.4"
+      NERDCTL_VERSION: "2.0.2"
```
BenTheElder, line 29: do you think we need to be testing both? were there major breaking changes in 2.x vs 1.x?

tao12345666333: As for the functions used in kind, there are no major changes except the support of containerd v2.

I'm fine if you want to test both of them.

BenTheElder: sounds like we don't need to for now, thanks

pkg/build/nodeimage/defaults.go
```diff
@@ -22,4 +22,4 @@ const DefaultImage = "kindest/node:latest"
 // DefaultBaseImage is the default base image used
 // TODO: come up with a reasonable solution to digest pinning
 // https://github.com/moby/moby/issues/43188
-const DefaultBaseImage = "docker.io/kindest/base:v20241212-9f82dd49"
+const DefaultBaseImage = "docker.io/kindest/base:v20250117-f528b021"
```
BenTheElder, line 25: can we leave this out of this commit? this will merge in another commit, see discussion in https://github.com/kubernetes-sigs/kind/pull/3848

BenTheElder: (also, I would expect the PR to work with the old base image / node images anyhow?? this is nerdctl on the host, right?, github actions doesn't build node images currently)

tao12345666333: Yes, sure.

.github/workflows/nerdctl.yaml
```diff
@@ -91,7 +91,7 @@ jobs:
       - name: Export logs
         if: always()
         run: |
-          sudo cat /etc/cni/net.d/* 
+          sudo find /etc/cni/net.d/ -type f -exec sh -c 'echo "{}" && cat "{}"' \;
```
AkihiroSuda, line 94: ?

tao12345666333: I have explained here https://github.com/kubernetes-sigs/kind/pull/3850#issuecomment-2600945612 that the previous method would lead to errors.

failed job: https://github.com/kubernetes-sigs/kind/actions/runs/12855249247/job/35840567917

BenTheElder:
/lgtm
thank you!

PR #3858, opened 2025-02-05
replace dated intial design docs with a clear warning to not depend on implementation details

BenTheElder:
These were meant for education about how kind is accomplished, but they've mostly been referernced by users attempting to extend and otherwise depend on internals

/cc @aojea @stmcginnis


aojea:
/lgtm

:)

Issue #3853, opened 2025-01-24
Failure to load images with kind v0.26.0 and kindes/node:v1.32.1


sorindumitru:
**What happened**:

Our CI tries to figure out the latest versions of kind and kindest/node to run integration tests against various k8s versions. After the image with v1.32.1 (roughly, I know it worked on Tuesday the 21st, but not on Thursday the 23rd) we started seeing CI failures for the combination of kind 0.26.0 and kindes/node v1.32.1. The error occurs when trying to load an image into the cluster:
```
[2025-01-24T08:51:40Z] executing 00-setup...
Ensuring kind version v0.27.0-alpha is available...
Want kubectl version: v1.32.1
Have kubectl version: v1.32.1
[2025-01-24T08:51:40Z] starting cluster...
Creating cluster "k8stest" ...
 ✓ Ensuring node image (kindest/node:v1.32.1) 🖼
 ✓ Preparing nodes 📦
 ✓ Writing configuration 📜
 ✓ Starting control-plane 🕹️
 ✓ Installing CNI 🔌
 ✓ Installing StorageClass 💾
Set kubectl context to "kind-k8stest"
You can now use your cluster with:

kubectl cluster-info --context kind-k8stest

Have a nice day! 👋
[2025-01-24T08:51:55Z] loading container images...
Image: "spire-server:latest-local" with ID "sha256:83de609b1ad3b445d26a9355b2926140f1bb0010649aa9858a270bd2acb14870" not yet present on node "k8stest-control-plane", loading...
ERROR: failed to detect containerd snapshotter
Stack Trace:
sigs.k8s.io/kind/pkg/errors.New
        /home/sorin/github/kubernetes-sig/kind/pkg/errors/errors.go:28
sigs.k8s.io/kind/pkg/cluster/nodeutils.parseSnapshotter
        /home/sorin/github/kubernetes-sig/kind/pkg/cluster/nodeutils/util.go:107
sigs.k8s.io/kind/pkg/cluster/nodeutils.getSnapshotter
        /home/sorin/github/kubernetes-sig/kind/pkg/cluster/nodeutils/util.go:97
sigs.k8s.io/kind/pkg/cluster/nodeutils.LoadImageArchive
        /home/sorin/github/kubernetes-sig/kind/pkg/cluster/nodeutils/util.go:81
sigs.k8s.io/kind/pkg/cmd/kind/load/docker-image.loadImage
[...]

PR #3859, opened 2025-02-10
Statically link CNI plugins

dims:
Picking a release of cni plugins and inspecting the binaries shows that they are statically linked. So we should do the same.

Following up to a multus-cni testing issue: 
- https://github.com/k8snetworkplumbingwg/multus-cni/pull/1213#issuecomment-2626094985
- https://github.com/k8snetworkplumbingwg/multus-cni/pull/1213#issuecomment-2635853338

```
$ find . -type f | xargs file | grep executable | cut -f 1-4 -d ','
./vrf:                                   ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked
./host-device:                           ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked
./vlan:                                  ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked
./portmap:                               ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked
./dhcp:                                  ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked
./sbr:                                   ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked
./firewall:                              ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked
./tuning:                                ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked
./host-local:                            ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked
./tap:                                   ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked
./bridge:                                ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked
./macvlan:                               ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked
./bandwidth:                             ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked
./ptp:                                   ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked
./ipvlan:                                ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked
./static:                                ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked
./loopback:                              ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked
./dummy:                                 ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked
```

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -176,7 +176,7 @@ RUN git clone --filter=tree:0 "${CNI_PLUGINS_CLONE_URL}" /cni-plugins \
     && eval "$(gimme "${GO_VERSION}")" \
     && export GOTOOLCHAIN="go${GO_VERSION}" \
     && mkdir ./bin \
-    && export GOARCH=$TARGETARCH && export CC=$(target-cc) && export CGO_ENABLED=1 \
+    && export GOARCH=$TARGETARCH && export CC=$(target-cc) && export CGO_ENABLED=0 \
     && go build -o ./bin/host-local -mod=vendor ./plugins/ipam/host-local \
     && go build -o ./bin/loopback -mod=vendor ./plugins/main/loopback \
     && go build -o ./bin/ptp -mod=vendor ./plugins/main/ptp \


aojea:
It seems we already discussed this in the original PR https://github.com/kubernetes-sigs/kind/pull/3214#discussion_r1190712828 , can't remember why we did one and not the other since there should not be any requirement on the plugins to use cgo
/assign @BenTheElder

BenTheElder:
> It seems we already discussed this in the original PR https://github.com/kubernetes-sigs/kind/pull/3214#discussion_r1190712828 , can't remember why we did one and not the other since there should not be any requirement on the plugins to use cgo

IIRC we were matching the upstream builds at the time?

Not using CGO makes DNS weird, for things "on the host" we generally build with cgo, unless they're not using the network.

I think none of the CNI plugins should need to use DNS though?

BenTheElder:
> Following up to a multus-cni testing issue:

So, to be clear though: https://kind.sigs.k8s.io/docs/design/node-image/

> DO NOT DEPEND ON THE INTERNALS OF THE NODE IMAGES.

statically linking these seems ~ok given upstream is and we're not aware of a specific need to dynamically link _these_ binaries

(also we should consider deleting the ones we're not using, but lets do that separately)

... but I generally expect CNI installs to ship their own binaries and not depend on ours, the same as installing on metal etc.

Commented in https://github.com/k8snetworkplumbingwg/multus-cni/pull/1213#issuecomment-2652347814

That shouldn't block matching the upstream binaries though, we don't have a strong reason to deviate in that regard.

BenTheElder:
NOTE: We're also not currently releasing at the moment due to being part way through sorting out containerd 2.0

https://github.com/kubernetes/kubernetes/issues/129818 / https://github.com/containerd/containerd/pull/11323 /  https://github.com/kubernetes-sigs/kind/issues/3795 /  https://github.com/containerd/containerd/issues/11344 / https://github.com/kubernetes-sigs/kind/issues/3853

BenTheElder (commented):
/lgtm
/approve

BenTheElder:
this will be available in pre-built node images starting after https://github.com/kubernetes-sigs/kind/pull/3861

though it sounds like there's some subtler bug with the multus install that this won't actually fix, for that particular use case

PR #3863, opened 2025-02-14
bump go to 1.23.6

BenTheElder:


BenTheElder:
preparing for a release, I don't think we're very meaningfully impacted by the 1.23.5 CVEs and certainly not 1.23.6

the go version used to build kind shouldn't impact the `pull-kind-e2e*` jobs, however it will _later_ when we pickup new images (kindnetd etc) built with this.

PR #3864, opened 2025-02-14
bump golang images

BenTheElder:
picks up go 1.23.6 for builds


BenTheElder:
[all actions passed, moving image from staging to "prod" and updating commit]

PR #3865, opened 2025-02-14
bump actions runners to ubuntu 22.04, as 20.04 is deprecated

BenTheElder:
https://github.com/actions/runner-images/issues/11101

the others were already on 22.04

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -14,7 +14,7 @@ permissions:
 jobs:
   docker:
     name: Docker
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     timeout-minutes: 30
     strategy:
       fail-fast: false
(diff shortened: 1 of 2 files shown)


BenTheElder:
permissions error w/ podman https://github.com/kubernetes-sigs/kind/actions/runs/13339133886/job/37260520002?pr=3865

BenTheElder:
looks like that was just a flake

BenTheElder:
nudge @aojea @stmcginnis 🙏

stmcginnis:
I'd say just move to 24.04, but maybe that's a little too "bleeding edge" for us. Or at least could cause some bruising.

/lgtm

BenTheElder:
> I'd say just move to 24.04, but maybe that's a little too "bleeding edge" for us. Or at least could cause some bruising.

I wound up sending a PR for this with podman, where we have some reliability issues with the older version in 22.04, but for the others I think we sort of want to intentionally check that we're not depending on anything newer. Last time we did (cgroupns) it was probably worth it to solve some bugs but it caused a lot of headache and I appreciate the smoke test on an older distro.

PR #3869, opened 2025-02-16
Attempt to tidy up documentation

adrianmoisey:
This tries to fix up the hanging sentance

### DIFF
--- a/site/content/docs/user/configuration.md
+++ b/site/content/docs/user/configuration.md
@@ -102,7 +102,7 @@ Multiple details of the cluster's networking can be customized under the
 
 #### IP Family
 
-KIND has support for IPv4, IPv6 and dual-stack clusters, it is possible to switch away from the default (IPv4).
+KIND has support for IPv4, IPv6 and dual-stack clusters, with the default being `ipv4`. You can change this by setting `ipFamily` under `networking` to `ipv6` or `dual`, see below for more requirements.
 
 ##### IPv6 clusters
 You can run IPv6 single-stack clusters using `kind`, if the host that runs the docker containers support IPv6.
--- a/site/content/docs/user/configuration.md
+++ b/site/content/docs/user/configuration.md
@@ -102,7 +102,7 @@ Multiple details of the cluster's networking can be customized under the
 
 #### IP Family
 
-KIND has support for IPv4, IPv6 and dual-stack clusters, you can switch from the default of IPv4 by setting:
+KIND has support for IPv4, IPv6 and dual-stack clusters, it is possible to switch away from the default.
 
 ##### IPv6 clusters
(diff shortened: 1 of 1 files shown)


site/content/docs/user/configuration.md
```diff
@@ -102,7 +102,7 @@ Multiple details of the cluster's networking can be customized under the
 
 #### IP Family
 
-KIND has support for IPv4, IPv6 and dual-stack clusters, you can switch from the default of IPv4 by setting:
```
adrianmoisey, line 105: This "you can switch from the default of IPv4 by setting:" felt strange. When I read it it felt like a config option was missing after the colon, but I see everything is described in the text below.
This attempts to fix that problem, but It feels un-elegant to me.

aojea: we need english native people, @BenTheElder , we still need to mention that the default is IPv4 in the new sentence

stmcginnis:
This reads better to me. It's possible the original text has evolved where it made sense to have the `:`, but now with it leading to different sections it doesn't seem right.

/lgtm

site/content/docs/user/configuration.md
```diff
@@ -102,7 +102,7 @@ Multiple details of the cluster's networking can be customized under the
...
+KIND has support for IPv4, IPv6 and dual-stack clusters, it is possible to switch away from the default.
```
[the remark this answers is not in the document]
stmcginnis: /remove-lgtm

+1, making the default very clear would be great.

stmcginnis:
/lgtm

Thanks!

site/content/docs/user/configuration.md
```diff
@@ -102,7 +102,7 @@ Multiple details of the cluster's networking can be customized under the
 
 #### IP Family
 
-KIND has support for IPv4, IPv6 and dual-stack clusters, you can switch from the default of IPv4 by setting:
+KIND has support for IPv4, IPv6 and dual-stack clusters, it is possible to switch away from the default (IPv4).
```
BenTheElder, line 105:
```suggestion
KIND has support for IPv4, IPv6 and dual-stack clusters, with the default being IPv4. You can change this by setting` ipFamily` under `networking` to `ipv6` or `dual`, see below for more requirements.
```

site/content/docs/user/configuration.md
BenTheElder: suggested something below, I agree that this is awkward currently

site/content/docs/user/configuration.md
adrianmoisey: nice and clean, thanks!

BenTheElder (commented):
/lgtm
/approve
thanks!

Issue #3849, opened 2025-01-18
Broken Kubeadm Config Patches for kubeadmConfigPatches


rooty0:
**What happened**:

Start the KIND cluster like this
```bash
kind create cluster --config kind.yaml
```
My `kind.yaml` is 
```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
name: test
nodes:
  - role: control-plane
  - role: worker
    kubeadmConfigPatches:
      - |
        kind: KubeletConfiguration
        registerWithTaints:
          - effect: NoSchedule
            key: mri-agent
            value: presence
```

**What you expected to happen**:
The worker node should have the taint assigned 

**How to reproduce it (as _minimally_ and precisely as possible)**:
Take the config and run it, It is **easy** to reproduce 

**Anything else we need to know?**:
When I run an alternative following configuration everything works ok: 
```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
name: test
nodes:
  - role: control-plane
  - role: worker
    kubeadmConfigPatches:
      - |
        kind: JoinConfiguration
        nodeRegistration:
          kubeletExtraArgs:
            register-with-taints: "mri-agent=presence:NoSchedule"
``` 
Thought I'm getting the following: `Jan 18 00:28:49 test-worker2 kubelet[188]: Flag --register-with-taints has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.`

**Environment:**
- kind version: kind v0.26.0 go1.23.4 darwin/arm64

BenTheElder:
So kubeadm only respects kubelet configuration for the first node.

Note that these are _kubeadm_ config patches
/remove-kind bug
/kind support

There's an issue somewhere discussing having v1alpha5 have *kubernetes* config patches and splitting out the patches and applying them to kubelet config after kubeadm generates it, but that's not implemented

*kubeadm* config patches patch the config going to kubeadm, and in kubeadm the kubelet config is cluster-wide and shared by all nodes after the first `kubeadm init` node unfortunately.

It would be good to add a warning about this to the kind config docs.

BenTheElder:
You will have to use the argument version, and ignore the warning for now. Kubelet currently marks all* flags with config equivalents as deprecated, but that stance has been walked back for some flags already.

rooty0:
[...]

PR #3872, opened 2025-02-19
Document KubeletConfiguration limitation

rooty0:
Per discussion #3849

### DIFF
--- a/site/content/docs/user/configuration.md
+++ b/site/content/docs/user/configuration.md
@@ -442,7 +442,7 @@ nodes:
           enable-admission-plugins: NodeRestriction,MutatingAdmissionWebhook,ValidatingAdmissionWebhook
 {{< /codeFromInline >}}
 
-> **NOTE**: When using `KubeletConfiguration`, kubeadm only applies the kubelet configuration to the **first** node.
+> **NOTE**: When using `KubeletConfiguration`, kubeadm only reads the kubelet configuration from the **first** node, which will apply to all nodes.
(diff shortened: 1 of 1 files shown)


site/content/docs/user/configuration.md
```diff
@@ -442,6 +442,26 @@ nodes:
           enable-admission-plugins: NodeRestriction,MutatingAdmissionWebhook,ValidatingAdmissionWebhook
 {{< /codeFromInline >}}
 
+> **NOTE**: When using `KubeletConfiguration`, kubeadm only applies the kubelet configuration to the **first** node.
```
BenTheElder, line 445:
```suggestion
> **NOTE**: When using `KubeletConfiguration`, kubeadm only reads the kubelet configuration from the **first** node, which will apply to all nodes.
```

BenTheElder: We should probably also mention that you can instead use extraArgs for kubelet in most cases, in JoinConfiguration. We should also probably mention Init vs Join and how kubeadm works, but at some point we should just link out to kubeadm docs and that shouldn't block this PR

BenTheElder:
thanks! I was out, reviewed, I think as currently written this is a little misleading

rooty0:
@BenTheElder thanks, fixed. 
As for the 
>  mention that you can instead use extraArgs for kubelet

I already mentioned it below with an example. Is there something specific you'd like me to add?

BenTheElder (commented):
/lgtm
/approve
thank you!

PR #3881, opened 2025-03-01
fix: set global arp_ignore param for kindnet

### LINKED ISSUE #3880: Container networking failures on hosts with `net.ipv4.conf.all.arp_ignore=2`
**What happened**:

I use `kind` across a variety of modern Linux distributions. I deployed it today on a recent release of Ubuntu and noticed that the `local-path-provisioner` was failing right after cluster creation. The logs showed `dial tcp 10.96.0.1:443: connect: no route to host`. Digging in deeper, I created some pods and noticed that when I exec into them they are unable to send traffic to each other.

The `local-path-provisioner` should be able to start successfully, and pod networking should be functional.

**Environment:**

- kind version: `kind v0.27.0 go1.22.2 linux/amd64`
- Runtime info: `podman version 4.9.3`
- OS: `Ubuntu 24.04.2 LTS amd64`
- Kubernetes version: Client `v1.32.2`, Server `v1.32.2`
- Any proxies or other special environment settings?: `All default settings`

> **Note**: I was able to reproduce the problem with `docker` as well as `podman`.

> **Note**: Everything I did here was _very_ vanilla. `kindnet` and all default settings.

1. Create a VM with the desktop version of `Ubuntu 24.04.2 LTS`
2. `apt-get install podman -y`
3. `go install sigs.k8s.io/kind@v0.27.0`
4. `kind create cluster`

> **Note**: There are likely multiple distributions that could trigger the issue, but in this case I deployed a _desktop_ version of Ubuntu in a VM (I could not reproduce this on the server version).

At this point you will find the `local-path-provisioner` is failing, and any pods you start can't access the network.

Stepping through some standard diagnostics I found pretty quickly that I was unable to `arping` the default route IP in these containers, but `tcpdump` did show those ARP requests showing up on the node's veth. This led me to discover that `net.ipv4.conf.all.arp_ignore=2` was set, which causes the ARP requests to go unanswered because this mode ignores those ARP requests because the sender's address and the requested address are not on the same subnet (kindnet configures a /32). So it appears that `kind` has historically relied on `net.ipv4.conf.all.arp_ignore=0` being set, as this has been pretty common for a variety of distributions but some distributions set it higher.

I ran the following:

```console
sysctl -w net.ipv4.conf.all.arp_ignore=1
```

and the `local-path-provisioner` came up and pod networking started working.

> **Note**: I tested other systems like Fedora 41 and the issue was not present as the host system sets `net.ipv4.conf.all.arp_ignore=0`. This happens on systems where the host sets `net.ipv4.conf.all.arp_ignore=2` by default.

As such I have created a patch which has worked for me in my testing, for your considerations:

#3881

However I'm a little perplexed: I would be _kinda surprised if I was the first person to find this_ ? Did I do something weird here? I promise I searched around, if I missed another report staring me right in the face I'm sorry! 😂

shaneutt:
Resolves #3880

If we do decide to go this way, in my testing setting `1` appears to work properly, and since this is the first time we've been opinionated about it it seems like an opportunity to set it as restrictively as we can tolerate. However we went with a setting of `0` because that has been the historically anticipated setting.

### DIFF
--- a/images/base/files/etc/sysctl.d/10-network-magic.conf
+++ b/images/base/files/etc/sysctl.d/10-network-magic.conf
@@ -9,3 +9,17 @@
 # Kubernetes mitigates the possible security issue caused by enabling this option.
 # ref: https://nvd.nist.gov/vuln/detail/CVE-2020-8558
 net.ipv4.conf.all.route_localnet=1
+
+# The global kernel parameter net.ipv4.conf.all.arp_ignore governs the
+# conditions under which ARP requests will be accepted or ignored. This global
+# setting will override any individual interface settings. Some host systems
+# might set this global parameter to a more restrictive setting of 2 (or
+# greater). Specifically, in mode 2, the system ignores ARP requests directed to
+# /32 addresses (this is what kindnet assigns to its veth interfaces) because
+# the request originates from a different subnet. This behavior breaks routing
+# for kindnet (specifically the ptp plugin), so we opinionatedly set this to 0
+# when kindnet is in use to ensure proper networking.
+#
+# For more information see:
+# https://www.kernel.org/doc/html/latest/networking/ip-sysctl.html#arp_ignore
+net.ipv4.conf.all.arp_ignore=0


pkg/cluster/internal/providers/docker/provision.go
```diff
@@ -209,6 +209,21 @@ func commonArgs(cluster string, cfg *config.Cluster, networkName string, nodeNam
 		args = append(args, "-e", "KIND_DNS_SEARCH="+strings.Join(*cfg.Networking.DNSSearch, " "))
 	}
 
+	// The global kernel parameter net.ipv4.conf.all.arp_ignore governs the
+	// conditions under which ARP requests will be accepted or ignored. Some
+	// host systems might set this global parameter to a more restrictive
+	// setting of 2 (or greater). Specifically, in mode 2, the system ignores
+	// ARP requests directed to /32 addresses (this is what kindnet assigns to
+	// its veth interfaces) because the request originates from a different
+	// subnet. This behavior breaks routing for kindnet, so we opinionatedly set
```
aojea, line 218: it is not really kindnet the problem, is the ptp plugin that is the one that kindnet deploys

shaneutt:
```suggestion
	// subnet. This behavior breaks routing for kindnet (specifically, the ptp plugin it deploys), so we opinionatedly set
```

aojea:
just set it in https://github.com/kubernetes-sigs/kind/tree/main/images/base/files/etc/sysctl.d

images/base/files/etc/sysctl.d/10-network-magic.conf
```diff
@@ -9,3 +9,17 @@
...
+# might set this global parameter to a more restrictive setting of 2 (or
+# greater). Specifically, in mode 2, the system ignores ARP requests directed to
+# /32 addresses (this is what kindnet assigns to its veth interfaces) because
+# the request originates from a different subnet. This behavior breaks routing
+# for kindnet (specifically the ptp plugin), so we opinionatedly set this to 1
+# when kindnet is in use to ensure proper networking.
+#
+# For more information see:
+# https://www.kernel.org/doc/html/latest/networking/ip-sysctl.html#arp_ignore
+net.ipv4.conf.all.arp_ignore=1
```
aojea, line 25:
```suggestion
net.ipv4.conf.all.arp_ignore=0
```

default is 0 https://sysctl-explorer.net/net/ipv4/arp_ignore/ , 1 seems to break some scenarios https://serverfault.com/questions/935366/why-does-arp-ignore-1-break-arp-on-pointopoint-interfaces-kvm-guest, I think is better to use 0

shaneutt: I mentioned in the issue that in my testing 1 seemed sufficient, and all the CI seems to pass with this. I suspect that 1 may actually be reasonable given that I don't think we need to be able to respond to ARP requests for the addresses for other interfaces, but at the same time "0" being the historical setting I can see wanting to start there first. Changed in 5e66ec05a5fb40f41c7e466bd24d3e248ecc1efc

BenTheElder:
Thanks!

FYI: this will be exercised in CI on the next PR after the image is built (postsubmit), similar to agnhost changes in kubernetes/kubernetes.

I don't expect to need to revert, but just calling out that while CI ran, really all of the e2e jobs didn't test this yet, we'll get that in the next step.

/lgtm
/approve
/hold
@aojea any remaining concerns?

aojea:
/hold cancel

Thanks @shaneutt

PR #3882, opened 2025-03-02
[WIP] Cni kindnet

aojea:
Simplify the network stack by using the cni-kindnet standalone plugin, that contains all the functionality we were using in a single binary, so we don't have to chain different plugins.

https://kindnet.es/docs/design/cni/

Fixes: #3880


images/base/Dockerfile
```diff
@@ -165,26 +165,16 @@ RUN git clone --filter=tree:0 "${CRI_TOOLS_CLONE_URL}" /cri-tools \
     && make BUILD_BIN_PATH=./build crictl \
     && GOARCH=$TARGETARCH go-licenses save --save_path=/_LICENSES ./cmd/crictl
 
-# stage for building cni-plugins
+# stage for building cni-plugin
```
BenTheElder, line 168: hmm, this has a risk of skewing with kindnetd right?

we probably need to think about that, though I'm not sure what the best option is.
we could modify the node image build to handle this but I'm not sure that's better

images/kindnetd/cmd/kindnetd/cni.go
```diff
@@ -32,91 +30,34 @@ import (
...
-		}
-	}
-	]
+const cniConfigTemplate = `{
+  "cniVersion": "0.4.0",
+  "name": "kindnet",
+  "plugins": [
+    {
+      "type": "cni-kindnet",
+      "ranges": [
```
BenTheElder, line 52: I feel like the name here could be slightly more descriptive as to the purpose?

aojea:
/close

I thought we had a bug, but it turned out to be other thing, this is not solving any problem we have today in kind, current setup has not had any bug in years, so we should be fine with it

Issue #3884, opened 2025-03-05
Private Docker Registry using certificates not working anymore


paichinger:
**What happened**:

I updated from kind 0.26 to 0.27. After that kind could not pull images from my private registry anymore. I have it configured according to the docs:
https://kind.sigs.k8s.io/docs/user/private-registries#use-a-certificate

**What you expected to happen**:
That I can pull images from my private registry using certificates according to the offiicial documtentation.

**How to reproduce it (as _minimally_ and precisely as possible)**:
If you have a private registry at hand, just try the official docs. 

**Anything else we need to know?**:
I saw that kind 0.27 uses containerd 2.X. So it might also be a bug or breaking change in containerd.

**Environment:**

- kind version: kind v0.27.0 go1.23.4 linux/amd64
- Runtime info: Client: Docker Engine - Community
 Version:    28.0.1
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.21.1
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.33.1
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 1
  Running: 1
  Paused: 0
  Stopped: 0
 Images: 3
 Server Version: 28.0.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: bcc810d6b9066471b0b6fa75f557a15a1cbf31bb
 runc version: v1.2.4-0-g6c52b3f
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.8.0-54-generic
 Operating System: Linux Mint 22.1
 OSType: linux
 Architecture: x86_64
 CPUs: 22
 Total Memory: 62.26GiB
 Name: petera-Precision-3591
[...]

PR #3885, opened 2025-03-05
update registry docs for containerd 2.x

BenTheElder:
see: https://github.com/kubernetes-sigs/kind/issues/3884

note: private registry is untested, but consistent with the upstream docs

### DIFF
--- a/site/content/docs/user/private-registries.md
+++ b/site/content/docs/user/private-registries.md
@@ -104,9 +104,10 @@ nodes:
     extraMounts:
       - containerPath: /etc/docker/certs.d/registry.dev.example.com
         hostPath: /etc/docker/certs.d/registry.dev.example.com
+# NOTE: the following patch is not necessary with images from kind v0.27.0+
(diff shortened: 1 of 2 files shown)


stmcginnis (commented):
Do we need to, or should we, keep examples for both 1.x and 2.x so the user can choose which instructions to follow for their environment?

BenTheElder:
> Do we need to, or should we, keep examples for both 1.x and 2.x so the user can choose which instructions to follow for their environment?

This should also work with 1.x as long as it's new enough to have opt in certs_dir (which is pretty much any kind release likely to work...), we just didn't get migrated previously like we did with the local registry

In fact, if this were for 2.x only, we'd drop the config patch entirely.

That said, I don't have an easy way to test with either, but it's the same idea as the local registry (which I have tested).

BenTheElder:
confirmed in https://github.com/kubernetes-sigs/kind/issues/3884#issuecomment-2703325772

PR #3889, opened 2025-03-06
CI: simplify setting up Lima

AkihiroSuda:
https://github.com/lima-vm/lima-actions is used


aojea:
The job failed to write the logs, does not seem related at all, the VM workflow all passed

/lgtm
/approve

Thanks

aojea:
the error is the same as reported in https://github.com/kubernetes-sigs/kind/pull/3888

Issue #3642, opened 2024-06-03
Invalid docs on mouting docker configs


th0ger:
**What would you like to be documented**:

[Mount a Config File to Each Node ](https://kind.sigs.k8s.io/docs/user/private-registries/#mount-a-config-file-to-each-node) says:

> If you pre-create a docker config.json containing credential(s) on the host you can mount it to each kind node.
> 
> Assuming your file is at /path/to/my/secret.json, the kind config would be:
> ```yaml
> kind: Cluster
> apiVersion: kind.x-k8s.io/v1alpha4
> nodes:
> - role: control-plane
>   extraMounts:
>   - containerPath: /var/lib/kubelet/config.json
>     hostPath: /path/to/my/secret.json
> ```

But if I do this with my `hostPath: ~/.docker/config/secret.json`, it mounts an *empty directory* `/var/lib/kubelet/config.json/`.

The [extra mounts](https://kind.sigs.k8s.io/docs/user/configuration/#extra-mounts) docs only states how to mount directories.
I assume that extra mount are not able to mount single files. But which directory should I mount the `.docker/config.json` to then?

**Why is this needed**:

It's not clear how/where to mount docker secrets.

th0ger:
I found out that

```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
  extraMounts:
  - containerPath: /var/lib/kubelet/.docker/config.json
    hostPath: ~/.docker/config.json
```

creates a local folder structure in the pwd where I call `kind create cluster --config=config.yaml`: `/home/USER_NAME/PWD_DIR/~/.docker/config.json/`. Yes, a literal `\~` dir structure, super trippy side effect.

BenTheElder:
"~" is not a real path character. You should expand that value (your shell will expand it to the real path) or this is a feature request to specially handle that character.

BenTheElder:
I think you have the wrong path?

https://kind.sigs.k8s.io/docs/user/private-registries/

th0ger:
Thanks @BenTheElder it's working with:
```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
  extraMounts:
  - containerPath: /var/lib/kubelet/config.json
    hostPath: /home/[USER]/.docker/config.json  # Full path, don't attempt ~ or $HOME !
```
So the documentation is okay 👍🏼 

[...]

PR #3893, opened 2025-03-13
Handle tild (user home dir) when define a host path

ebuildy:
Using kind on macos, I want to share my Podman auth file, but tild ~ is not working.

This PR:

- add utils fs function "Abs" that handle tild
- use this function for podman provider only

I can write tests if needed!

Relates:

- https://github.com/kubernetes-sigs/kind/issues/3642

Thanks you

### DIFF
--- a/pkg/cluster/internal/providers/podman/provision.go
+++ b/pkg/cluster/internal/providers/podman/provision.go
@@ -28,6 +28,7 @@ import (
 	"sigs.k8s.io/kind/pkg/cluster/constants"
 	"sigs.k8s.io/kind/pkg/errors"
 	"sigs.k8s.io/kind/pkg/exec"
+	"sigs.k8s.io/kind/pkg/fs"
 
 	"sigs.k8s.io/kind/pkg/cluster/internal/loadbalancer"
 	"sigs.k8s.io/kind/pkg/cluster/internal/providers/common"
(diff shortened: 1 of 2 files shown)


aojea:
:/ sounds a bit like a stretch no?

aojea:
I'm not sure we want to do this, special since golang does not want to support it https://github.com/golang/go/issues/57569

stmcginnis:
I'm a bit mixed on this too. The tilde is a shell construct, so running outside of the shell is usually not expected to work. I do see the convenience of it though.

BenTheElder:
Our contributor guide asks that features be discussed in an issue first.

~ is a shell concept, and you could technically name a file system path "~" AFAIK.

We also nominally support windows, as long as you have docker CLI pointed at a Linux container environment.

I think users that want to mount home should wrap calling mind in a shell script or similar.

I generally agree with https://github.com/golang/go/issues/57569#issuecomment-1370751704

ebuildy:
ho thanks you for short replies, I agree if golang didnt merge it, this is not a good idea ^^

What about using env variable?

BenTheElder:
We recommend wrapping kind with your preferred scripting language to generate the config if you need templating, we don't want to wade into the templating debates and kind is very light weight.

EG you can use shell like:
```shell
# config here, can read $HOME if you're in a shell, since the shell should look it up and set it
EOF
```

PR #3896, opened 2025-03-19
site: update required Go version

pellared:
Per: https://github.com/kubernetes-sigs/kind/blob/main/go.mod#L11


BenTheElder:
Thanks!

PR #3898, opened 2025-03-19
pin new actions to commit.

BenTheElder:
all previous actions already pinned.

limits exposure to something like https://arstechnica.com/information-technology/2025/03/supply-chain-attack-exposing-credentials-affects-23k-users-of-tj-actions/

to explicit actions update PRs

we regressed in https://github.com/kubernetes-sigs/kind/pull/3889, though this is just "best practice" here, we have NONE with write permission, and actions CI results are already untrustworthy due to running with the config in the PR branch rather than the config of the target branch anyhow

### DIFF
--- a/.github/workflows/vm.yaml
+++ b/.github/workflows/vm.yaml
@@ -44,11 +44,11 @@ jobs:
           check-latest: true
 
       - name: "Install Lima"
-        uses: lima-vm/lima-actions/setup@v1
+        uses: lima-vm/lima-actions/setup@be564a1408f84557d067b099a475652288074b2e # v1
         id: lima-actions-setup
 
       - name: "Cache ~/.cache/lima"
-        uses: actions/cache@v4
+        uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
         with:
           path: ~/.cache/lima
           key: lima-${{ steps.lima-actions-setup.outputs.version }}


stmcginnis:
/lgtm

I wonder if it would be worth adding some comments at the top of the workflow files, or maybe a README in that folder, calling out this expectations so we don't have to rely solely on code reviews to make sure someone doesn't propose a tagged version. It doesn't add much beyond relying on the code review, but it would at least have a slight chance of someone discovering it and would give us a link to point to if they don't.

BenTheElder:
> I wonder if it would be worth adding some comments at the top of the workflow files, or maybe a README in that folder, calling out this expectations so we don't have to rely solely on code reviews to make sure someone doesn't propose a tagged version. It doesn't add much beyond relying on the code review, but it would at least have a slight chance of someone discovering it and would give us a link to point to if they don't.

Yeah, OTOH the number of reviewers is small, and I talked to @aojea about this problem since (as the current other approver)

I'm not sure if people would look at the README.

It's also not a big deal for presubmit testing, the vector there is like .... falsely pass? But that can already happen due to outdated actions in the branch.

However, we want to be consistent, so when the next project mimics us, they continue the best practice.

This is the use case I'm actually concerned about: https://github.com/kubernetes-sigs/cloud-provider-kind/pull/225#event-16878373367

BenTheElder:
I think we want to consider how we'd track this org-wide, but that's ... messy. There's a thread in #github-management

PR #3909, opened 2025-04-10
docs(website): Added content to highlight the usage of Kind with Podm…

shipsing:
…an Desktop
**Purpose**: To highlight the usage of Kind with Podman Desktop
- Help developers or community members to quickly get started with Kind using the Podman Desktop UI.


aojea:
Thanks for your contribution but podman is still experimental https://github.com/kubernetes-sigs/kind/issues/1778 and there were no contributions on trying to make it more stable, so we should avoid giving false expectations

/close

PR #3911, opened 2025-04-11
Update golangci-lint to v2.0.2

stmcginnis:
Golangci-lint has released a new major version that includes some changes to how it is configured, along with the usual linter updates that come with every release.

This updates the golangci-lint version we use to v2.0.2 (the current latest release) and updates the config file to conform to the new format. Also includes quite a few trivial updates to address the new warnings from the updated linters.


pkg/cluster/internal/create/actions/action.go
```diff
@@ -14,6 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// Package actions contains logic for steps to perform once an initial node
```
aojea, line 17: I don't feel this lint is adding value, should we remove this linter @BenTheElder ?

aojea:
github actions runners have bug in the kernel
```
2025-04-11T14:24:57.741538117Z stderr F 	ip6tables-restore v1.8.9 (nf_tables): unknown option "--xor-mark"
2025-04-11T14:24:57.741540752Z stderr F 	Error occurred at line: 22
2025-04-11T14:24:57.741543337Z stderr F 	Try `ip6tables-restore -h' or 'ip6tables-restore --help' for more information.
2025-04-11T14:24:57.741546342Z stderr F  > ipFamily="IPv6"
```

https://github.com/actions/runner-images/issues/11985

pkg/cluster/internal/create/actions/action.go
stmcginnis: True.

As someone that started using kind mostly as a library to integrate into other functionality, I do appreciate having docstrings on most things. But these package level docs are really mostly useful if there is actual details to provide about the package. The ones missing these docstrings really don't need it, so there's not really much value to it.

I will exclude this check from the list. I think two of the linters both detect it, so just need to add the config for both to ignore it. But since it doesn't hurt anything, I'll leave the package docstrings added rather than trying to clean them all out.

Will probably not get to it until tomorrow or Monday. But with the ubuntu job failures, probably no rush until that's resolved anyway.

aojea: we can switch the image , the newest ones does not have the bug

hack/tools/go.mod
```diff
@@ -1,50 +1,60 @@
 module sigs.k8s.io/kind/hack/tools
 
-go 1.23
+go 1.23.0
+
+toolchain go1.24.0
```
aojea, line 5: @BenTheElder where are we now on the use of `toolchain` ? I remember we discussed this in the past but can't remember what we have decided

stmcginnis: Does that apply for `tools/` too though? I thought we treated that a little differently than the root `go.mod`.

Mostly just curious. Updating to drop this part of the change as I don't think it's really related or necessary for the main goal.

BenTheElder: > Does that apply for tools/ too though? I thought we treated that a little differently than the root go.mod.

These tools are primarily applied to go code that we'll build at .go-version, I'm not sure if any of them will switch behavior if they assume N+1 go version.

BenTheElder: It's easier to reason about matching them, if possible.

BenTheElder:
something broke badly with a number of the actions, we're going to have to get that fixed

BenTheElder:
ok so it's the ones on 22.04, ip6tables is broken there

BenTheElder:
we can bypass the actions, they're not relevant to only bumping golangci-lint

I want to run it again now that .go-version is on 1.22 though
/test pull-kind-verify
/test pull-kind-test

pkg/cluster/internal/create/actions/action.go
BenTheElder: +1

hack/tools/go.mod
BenTheElder: apparently there is a go bug that will be fixed about inserting toolchain when it wasn't specified before

stmcginnis:
Sorry, this had dropped off my radar. Will get rebased and updated.

The Go 1.25 release reminded me of this. Saw this nice update in the release notes:

> When the go command updates the go line in a go.mod or go.work file, it [no longer](https://go.dev/ref/mod#go-mod-file-toolchain) adds a toolchain line specifying the command’s current version.

BenTheElder:
> When the go command updates the go line in a go.mod or go.work file, it [no longer](https://go.dev/ref/mod#go-mod-file-toolchain) adds a toolchain line specifying the command’s current version.

Yeah! 🎉 , been waiting on this bug.

I think, it might even be something to consider soonish as 1.21+ becomes available in stable / LTS distros to leverage this, but we probably don't want it writing in the current version every time we touch go mod ...

pkg/cluster/internal/create/actions/action.go
BenTheElder: it's less useful for internal packages but /shrug

BenTheElder:
we might have to hold a little on 1.25, runc for example can be slightly sensitive to go upgrades, I haven't checked into that yet

pkg/cluster/internal/create/actions/action.go
stmcginnis: I do agree for these. As I mentioned above, this check is now disabled in a later update. I just didn't think it was worth the effort to go back and undo what I had started. Even though it doesn't really add value, it doesn't hurt. I don't think any effort should be put in to adding more of it though.

hack/tools/go.mod
stmcginnis: Oops, I missed excluding this in the latest update. I'm not using 1.25 yet, and I agree we shouldn't move to it quite yet.

Will remove this line and push an update.

BenTheElder:
Kubernetes/Kubernetes is also still not updated, I do want to leverage 1.25 though 😅 

We're going to have go keep an eye out for our controllers attempting to become cgroup aware in the weird nested cgroup environment 🙃

stmcginnis:
> we might have to hold a little on 1.25

Is there a reason to hold on this until 1.25 is in use?

BenTheElder:
We're on 1.25 since that comment anyhow. We're not yet on 1.26

pkg/internal/runtime/runtime.go
```diff
@@ -1,3 +1,20 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
BenTheElder, line 2: Thanks. If you haven't seen, *new* files will exclude the year in the future.

We haven't put hard enforcement in this repo yet but we should probably consider it.

stmcginnis: That's right! I think I actually created this file before that became official, but I'll try to remember that now. Thanks!

BenTheElder: This file isn't new, but I just mentioned because we're not managing this policy already 🙈

PR #3920, opened 2025-04-24
bump go, containerd, runc

BenTheElder:
now that kubernetes 1.33 is out and the post-thaw merge queue is drained ...


BenTheElder:
hmmm the actions jobs are having issues, we might actually have a regression in here

/hold

BenTheElder:
ehhh they're also broken on the empty PR: https://github.com/kubernetes-sigs/kind/pull/648

I would like to see them passing though ...

BenTheElder:
will have to rebase with https://github.com/kubernetes-sigs/kind/pull/3921, which will annoyingly make the base image not match a commit anymore >.>

maybe just adding that as a new commit here ...

PR #3922, opened 2025-04-25
eliminate yaml.v3

BenTheElder:
not convinced these were necessary, users can patch arbitrary yaml in anyhow, I don't think there's a realistic threat model resolved by safetext/yamltemplate

further, the safetext repo doesn't seem to be responsively maintained, last commit 9 months ago, and the small number of issues are going unresponded (4 open, 0 closed, some multiple years old)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,6 @@ require (
 	al.essio.dev/pkg/shellescape v1.5.1
 	github.com/BurntSushi/toml v1.4.0
 	github.com/evanphx/json-patch/v5 v5.6.0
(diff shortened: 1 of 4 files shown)


stmcginnis:
Test failure doesn't look related. Do we drop 1.28 soon?

/lgtm

BenTheElder:
Yeah, we don't really have a clear approach to this, it's sort of useful to know if they broke but eventually we have to drop old releases and leave them more like "we accept bug fixes and don't actively break them"
/retest

BenTheElder:
ah, right, I wound up only having to change *the consuming side* of kindnetd, so no image build

PR #3926, opened 2025-05-05
docs: add note about vpn cidr clashes

mwdomino:
This PR is a quick note about clashing VPN CIDRs with the KIND docker network that is created by default.

I ran into this issue as the default 172.17.x.x network overlaps with some internal corporate VPN networks. I was able to resolve it by creating the network manually with a CIDR that did not conflict. This fix will persist as long as the network is not removed.

### DIFF
--- a/site/content/docs/user/known-issues.md
+++ b/site/content/docs/user/known-issues.md
@@ -41,7 +41,7 @@ description: |-
 * [Docker Desktop for macOS and Windows](#docker-desktop-for-macos-and-windows)
 * [Older Linux Distributions](#older-linux-distributions)
 * [Failure to Create Cluster on WSL2](#failure-to-create-cluster-on-wsl2)
-*  [VPN CIDR clashes](#vpn-cidr-clashes)
+* [VPN CIDR clashes](#vpn-cidr-clashes)
 
 ## Troubleshooting Kind
 
--- a/site/content/docs/user/known-issues.md
+++ b/site/content/docs/user/known-issues.md
@@ -41,7 +41,7 @@ description: |-
 * [Docker Desktop for macOS and Windows](#docker-desktop-for-macos-and-windows)
 * [Older Linux Distributions](#older-linux-distributions)
 * [Failure to Create Cluster on WSL2](#failure-to-create-cluster-on-wsl2)
-* [VPN CIDR clashes](#vpn-cidr-clashes)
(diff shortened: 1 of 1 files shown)


site/content/docs/user/known-issues.md
```diff
@@ -416,6 +417,25 @@ the project relies on community support and feedback. It has been noted that the
...
+
+```sh
+# stop your KIND cluster
+kind delete cluster
+
+# delete the docker network
+docker network rm kind
+
+# create a new network with a non-conflicting CIDR
+docker network create --subnet=10.253.0.0/22 kind
```
BenTheElder, line 434: We should recommend configuring docker IPAM instead?

This example is missing options that should be set on the network and some of them are programmatic (like mirroring MTU from the default network)

BenTheElder: https://docs.docker.com/engine/daemon/ + `default-address-pool`

https://forums.docker.com/t/docker-default-address-pool-customization-question/112969/6
https://github.com/moby/moby/issues/43128

mwdomino: Wow, thanks for this @BenTheElder - updating the daemon.json is a much cleaner fix. I just tested and it works for my issue as well. I'll update the docs

site/content/docs/user/known-issues.md
```diff
@@ -416,6 +417,26 @@ the project relies on community support and feedback. It has been noted that the
 steps detailed in [https://github.com/spurin/wsl-cgroupsv2](https://github.com/spurin/wsl-cgroupsv2)
 have been necessary to resolve this issue.
 
+## VPN CIDR Clashes
```
stmcginnis, line 420: Sorry, one more thing here. In your case it was a VPN, but I've seen this happen with someone who happened to have the same lab IP subnet. So it might be better to generalize this a little and just call it something like "Local Subnet Conflicts".

mwdomino: That's a great call, thanks

stmcginnis:
Thanks! I think it's sort of rare that we run in to this, but when someone does, having at least a breadcrumb to help them figure out what to do will be very useful.

/lgtm

BenTheElder (commented):
/lgtm
/approve
/hold
[holding for any more comments from @stmcginnis]

Thank you!

stmcginnis:
Good for me!

/unhold

PR #3933, opened 2025-05-15
push-node: default to staging registry

BenTheElder:
the newer script that uses pre-built binaries for 1.31+ already does this

/cc @stmcginnis


BenTheElder:
actually also cc @aojea just for the FYI

to get the old behavior `REGISTRY=kindest ...`

PR #3935, opened 2025-05-16
Update default image to use v1.33.1

stmcginnis:


BenTheElder (commented):
/lgtm
/approve

BenTheElder:
This PR only updates the default node image.
Kubernetes/Kubernetes e2e tests always build a new node image, and are not impacted by this change
Skipping kubernetes/kubernetes e2e test jobs under that narrow constraint:
/override pull-kind-e2e-kubernetes-1-28
/override pull-kind-e2e-kubernetes-1-29
/override pull-kind-e2e-kubernetes-1-30
/override pull-kind-e2e-kubernetes-1-31
/override pull-kind-e2e-kubernetes-1-32
/override pull-kind-e2e-kubernetes-1-33
/override pull-kind-conformance-parallel-ipv6
/override pull-kind-conformance-parallel-dual-stack-ipv4-ipv6
Ordinarily I'd just let them run for flake detection anyhow and erring on the side of cuation, but we're trying to ship a release, and I know this change will not impact those jobs.
Note: normally we need to be careful with changes that impact those jobs, because these are sentinels for Kubernetes/Kubernetes release blocking signal which uses kind at head.

We will still wait for all other test results.

PR #3951, opened 2025-06-04
fix builds with relative kubernetes-source paths

### LINKED ISSUE #3949: node-image build breaks on some relative source paths
**What happened**:

When building the node-image with a relative kubernetes-source path, this may result in an error stating the source folder can't be found.

```
$ kind build node-image ./kubernetes 
Detected build type: "source"
Building using source: "./kubernetes"
Starting to build Kubernetes
Failed to build Kubernetes: command "sh -c 'cd ./kubernetes && hack/print-workspace-status.sh'" failed with error: exit status 1
ERROR: error building node image: failed to build kubernetes: command "sh -c 'cd ./kubernetes && hack/print-workspace-status.sh'" failed with error: exit status 1
Command Output: sh: line 0: cd: ./kubernetes: No such file or directory
```

Given a directory structure like this:
```
.
└── open-source
    ├── kind
    └── kubernetes
```

If the current working directory is `open-source`, doing a build with paths `./kubernetes` or `kubernetes` won't work. What does work are absolute paths as well as:
- `kind build node-image .` from inside the `kubernetes` folder
- `kind build node-image ../kubernetes` from inside the `kind` folder

The build from source should start, as is the case if the command were to be run with an absolute path. E.g.:

```
$ kind build node-image ~/Development/open-source/kubernetes
Detected build type: "source"
Building using source: "/Users/<username>/Development/open-source/kubernetes"
Starting to build Kubernetes
+++ [0604 21:56:46] Verifying Prerequisites....
+++ [0604 21:56:46] Using docker on macOS
+++ [0604 21:56:47] Building Docker image kube-build:build-b12bb1ed9d-5-v1.33.0-go1.24.2-bullseye.0
```

cd into the directory containing your kubernetes source folder, then run `kind build node-image <k8s source dir>`

No

**Environment:**

- kind version: (use `kind version`): kind v0.27.0 go1.24.3 darwin/arm64
- Runtime info: (use `docker info`, `podman info` or `nerdctl info`): v28.1.1
- OS (e.g. from `/etc/os-release`): macOS Sequoia (v15.5)
- Kubernetes version: (use `kubectl version`): Client Version: v1.32.2

oduludo:
Fixes #3949 

When building the node-image from a relative path, `dockerBuilder.Build()` would cd into that relative path twice. Once in `Build()` itself and once inside the `sourceVersion()` function (to then call `hack/print-workspace-status.sh`). For some relative paths the second cd would fail, as that directory doesn't contain itself (effectively something like doing `cd kubernetes && cd kubernetes`).

This PR moves the `sourceVersion()` call up so it isn't affected by the cd that is performed by `Build()` itself.


pkg/build/nodeimage/internal/kube/builder_docker.go
```diff
@@ -65,12 +72,6 @@ func (b *dockerBuilder) Build() (Bits, error) {
 		return nil, err
 	}
 
-	// capture version info
-	sourceVersionRaw, err := sourceVersion(b.kubeRoot)
-	if err != nil {
-		return nil, err
-	}
-
 	kubeVersion, err := version.ParseSemantic(sourceVersionRaw)
```
stmcginnis: This makes sense to keep with the `sourceVersion` call. 

And actually, most of the next few things are not directory dependent either. Instead of moving this block, could you move the `os.Chdir()` bits to be right before the `make` call where it does the build since that is the piece that actually needs to be in the source directory?

oduludo: Nice catch, will do. Definitely makes sense to have that right where it is needed.

oduludo: @stmcginnis Done, please check if this is what you had in mind.

pkg/build/nodeimage/internal/kube/builder_docker.go
```diff
@@ -98,6 +90,15 @@ func (b *dockerBuilder) Build() (Bits, error) {
 		"cmd/kubelet",
 	}
 
+	// make sure we cd back when done
```
BenTheElder: EDIT: ~maybe we should do this right after source version, just in case the rest gets refactored?~

BenTheElder (commented):
thank you!

pkg/build/nodeimage/internal/kube/builder_docker.go
BenTheElder: we could also consider resolving kuberoot to an absolute path early in the process, or calling make with `-C` instead of Chdir ourselves

stmcginnis: You're call, but I recommended moving it here since this is in close proximity to where it is actually needed. So I think that might help reduce the risk of a refactor causing issues since it's more "visible" if someone is touching the make call.

`-C` seems reasonable too.

BenTheElder: > You're call, but I recommended moving it here since this is in close proximity to where it is actually needed. 

That makes sense, I was thinking it's "needed" by the rest of the code, but really it's not "needed",  I can't recall why we didn't just use `-C` ... but this code is an old amalgam at this point, and I don't see any reason we couldn't do that now and drop the chdir altogether.

BenTheElder: @oduludo I'd be open merging as-is, or if you want to take a stab at moving to `-C` we'll try that :-)
/hold
Will check back tomorrow

BenTheElder:
/lgtm
/approve
(held for https://github.com/kubernetes-sigs/kind/pull/3951#discussion_r2127524595)

pkg/build/nodeimage/internal/kube/builder_docker.go
oduludo: Sure, -C is fine to implement. Pushed the changes just now. Thanks for the input! :D

stmcginnis:
/lgtm
/hold
/approve

Just want to give Ben a chance to take a look.

BenTheElder (commented):
/lgtm
/approve
/hold cancel
thank you!

PR #3954, opened 2025-06-06
Add variable to local registry script to set cluster-name

mbwhite:
Adds in a variable to hold the cluster name; in the default case this doesn't add anything. However in cases (like mine) where I'd changed the cluster name to be that of the project - the standard script fails.  

It took quite a while to debug and find out why the local registry wasn't working. This change is quite simple but would have saved me quite some time! And hopefully others as well.

### DIFF
--- a/site/static/examples/kind-with-registry.sh
+++ b/site/static/examples/kind-with-registry.sh
@@ -1,6 +1,8 @@
 #!/bin/sh
 set -o errexit
 
+cluster_name='kind'
+
 # 1. Create registry container unless it already exists
 reg_name='kind-registry'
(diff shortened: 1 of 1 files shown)


stmcginnis:
/retitle Add variable to local registry script to set cluster-name

BenTheElder:
You could use `KIND_CLUSTER_NAME` with script as-is, otherwise this script is meant to be a simple example.

In any case that you use kind clusters across different commands you will need to pass `--name` consistently, or the env (I do usually recommend the flag, the env is easier to miss across terminals etc, but in the case of invoking the script it would be sufficient)

Issue #3941, opened 2025-05-25
[v0.29.0] Cannot connect to kube-apiserver after node restart


tanvp112:
**What happened**:

With cluster created with [v0.29.0](https://github.com/kubernetes-sigs/kind/releases/tag/v0.29.0); the kind-control-plane node becomes unreachable upon restart. Check kubelet log in the control plane container shows connection refused to https://172.18.0.2:6443, due to the IP is now occupied by worker container.

```
kind create cluster

Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.33.1) 🖼
 ✓ Preparing nodes 📦 📦 📦 📦  
 ✓ Writing configuration 📜 
 ✓ Starting control-plane 🕹️ 
 ✓ Installing StorageClass 💾 
 ✓ Joining worker nodes 🚜 
Set kubectl context to "kind-kind"
You can now use your cluster with:

kubectl cluster-info --context kind-kind

Thanks for using kind! 😊

docker inspect -f '{{.Name}} - {{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' $(docker ps -q)
/kind-worker - 172.18.0.5
/kind-control-plane - 172.18.0.2
/kind-worker3 - 172.18.0.3
/kind-worker2 - 172.18.0.4
```

Now **restart** the host.

```
docker inspect -f '{{.Name}} - {{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' $(docker ps -q)
/kind-control-plane - 172.18.0.3        <-- Noticed the IP has changed.
/kind-worker2 - 172.18.0.4
/kind-worker3 - 172.18.0.5
/kind-worker - 172.18.0.2
```

At this point do a `kubectl get nodes` the `STATUS` of `kind-control-plane` node will show `NOT READY`.

**What you expected to happen**:

The connection refused does not happen with previous `kind` release.

**How to reproduce it (as _minimally_ and precisely as possible)**:

Delete the cluster created with v0.29.0, download [v0.28.0](https://github.com/kubernetes-sigs/kind/releases/tag/v0.28.0) and redo the step,s there's no unreachable node issue.

**Anything else we need to know?**:

The node image pulled by `kind` is different for the same Kubernetes v1.33.1:

kindest/node                             <none>            050072256b9a   4 days ago    1.54GB  <-- This was pulled by v0.29.0
[...]

PR #3957, opened 2025-06-12
update kubelet server IP on reboot

BenTheElder:
/hold
draft fix for https://github.com/kubernetes-sigs/kind/issues/3941, untested

also haven't root caused the behavior change in 1.33 vs 1.32 yet


BenTheElder:
Note: This fix is potentially useful for single node-clusters on non-docker backends where we aren't able to use hostnames anyhow.

BenTheElder:
this change should be safe, even untested

stmcginnis:
Yeah, looks safe.

/lgtm
/approve

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/3964 will pick this up

PR #3962, opened 2025-06-27
loadbalancer/config: Parse resolv.conf for DNS

tchap:
Control plane in HA setup doesn't currently work using Podman. This is because the haproxy DNS server is hard-coded to use 127.0.0.1:53, which is not available on Podman.

To get the actual NS address, /etc/resolv.conf must be consulted. So this patch adds parse-resolv-conf to the haproxy resolvers block, which makes HA control plane work on Podman instantly.

127.0.0.1:53 is retained as a nameserver still.

### Context

Let's use the following config:

```yaml
kind: Cluster
name: multinode
nodes:
  - role: control-plane
  - role: control-plane
  - role: control-plane
  - role: worker
```

Then let's run the following:

```
$ ./bin/kind create cluster --config kindconfig.yaml --retain
```

The output with some truncated parts:

```
Creating cluster "multinode" ...
 ✓ Ensuring node image (kindest/node:v1.33.1) 🖼
 ✓ Preparing nodes 📦 📦 📦 📦  
 ✓ Configuring the external load balancer ⚖ 
 ✓ Writing configuration 📜 
 ✗ Starting control-plane 🕹 
ERROR: failed to create cluster: failed to init node with kubeadm: command "podman exec --privileged multinode-control-plane kubeadm init --config=/kind/kubeadm.conf --skip-token-print --v=6" failed with error: exit status 1

(...)

I0627 13:13:47.242628     171 round_trippers.go:632] "Response" verb="GET" url="https://multinode-external-load-balancer:6443/livez?timeout=10s" status="" milliseconds=1
I0627 13:13:48.243680     171 with_retry.go:234] "Got a Retry-After response" delay="1s" attempt=8 url="https://multinode-external-load-balancer:6443/livez?timeout=10s"
I0627 13:13:48.244978     171 round_trippers.go:632] "Response" verb="GET" url="https://multinode-external-load-balancer:6443/livez?timeout=10s" status="" milliseconds=1
I0627 13:13:49.245139     171 with_retry.go:234] "Got a Retry-After response" delay="1s" attempt=9 url="https://multinode-external-load-balancer:6443/livez?timeout=10s"
I0627 13:13:49.246500     171 round_trippers.go:632] "Response" verb="GET" url="https://multinode-external-load-balancer:6443/livez?timeout=10s" status="" milliseconds=1
[control-plane-check] kube-apiserver is not healthy after 4m0.000259768s
```

I then tried to `curl` the first control plane node to the the health status, everything was ok.
On the other hand, haproxy was showing

```
10.89.0.67:54516 [27/Jun/2025:13:10:07.229] control-plane kube-apiservers/<NOSRV> -1/-1/0 0 SC 1/1/0/0/0 0/0
10.89.0.67:58518 [27/Jun/2025:13:10:08.231] control-plane kube-apiservers/<NOSRV> -1/-1/0 0 SC 1/1/0/0/0 0/0
10.89.0.67:58528 [27/Jun/2025:13:10:09.232] control-plane kube-apiservers/<NOSRV> -1/-1/0 0 SC 1/1/0/0/0 0/0
10.89.0.67:58530 [27/Jun/2025:13:10:10.217] control-plane kube-apiservers/<NOSRV> -1/-1/0 0 SC 1/1/0/0/0 0/0
10.89.0.67:58538 [27/Jun/2025:13:10:11.218] control-plane kube-apiservers/<NOSRV> -1/-1/0 0 SC 1/1/0/0/0 0/0
10.89.0.67:58550 [27/Jun/2025:13:10:12.220] control-plane kube-apiservers/<NOSRV> -1/-1/0 0 SC 1/1/0/0/0 0/0
```

After much research I nailed it down to a DNS issue.
Applying this patch makes the cluster boot just fine.
haproxy was actually not logging any health check, which was confusing, but now:

```
[WARNING] 177/132752 (29) : Health check for server kube-apiservers/multinode-control-plane succeeded, reason: Layer7 check passed, code: 200, check duration: 5ms, status: 1/2 DOWN.
Server kube-apiservers/multinode-control-plane is UP. 1 active and 0 backup servers online. 0 sessions requeued, 0 total in queue.
```

### DIFF
--- a/pkg/cluster/internal/loadbalancer/config.go
+++ b/pkg/cluster/internal/loadbalancer/config.go
@@ -40,7 +40,7 @@ global
   maxconn 100000
 
 resolvers docker
-  nameserver dns 127.0.0.11:53
+  parse-resolv-conf
 
 defaults
   log global


stmcginnis:
cc @aojea

aojea:
> Control plane in HA setup doesn't currently work using Podman. This is because the haproxy DNS server is hard-coded to use 127.0.0.1:53, which is not available on Podman.

we do some magic to capture the dns traffic 

https://github.com/kubernetes-sigs/kind/blob/3df64e784cc0ea74125b2a2e9877817418afa3af/images/base/files/usr/local/bin/entrypoint#L458-L487

the problem that existed with podman is that didn't provide container discovery via dns like pod, and you had to use https://github.com/containers/dnsname but I see that got archived and replaced, if the host is not able to resolve the containernames then the loadbalancer fails ... 

what is the content of the resolv.conf on the failing nodes?

tchap:
@aojea 

```
search dns.podman .
nameserver fc00:f853:ccd:e793::1
nameserver 10.89.0.1
```

pkg/cluster/internal/loadbalancer/config.go
```diff
@@ -41,6 +41,7 @@ global
 
 resolvers docker
   nameserver dns 127.0.0.11:53
```
aojea: yep, 127.0.0.11 is a docker only thing , but this is this way to allow reboots to work, so we use the names as endpoints and deal with the ip renumbering ...  I do not know how both statements at the same time will work and if we'll regress ... maybe we are able to fill the template per provider?

tchap: Yeah, that was my other idea 🙂 I can look into that.

tchap: @aojea Would you find it too ugly to just add `Name()` to the `Provider` interface? It seems like the easiest way to get that information to where it is needed, and actually the provider selection is scattered across multiple packages.

tchap: Could be handled in this way, perhaps... https://github.com/kubernetes-sigs/kind/pull/3962/commits/b7d80c89197f65e0827a9dba83b0ea28d8a8c20a

I will add tests if you tell me this approach is ok, although I can understand if you don't like extending the provider interface. But as I said, passing the provider name around would IMO make the code more ugly.

pkg/cluster/internal/providers/docker/provider.go
```diff
@@ -39,6 +39,9 @@ import (
 	"sigs.k8s.io/kind/pkg/internal/sets"
 )
 
+// ProviderName is the constant provider name associated with the Docker provider.
+const ProviderName = "docker"
```
BenTheElder: We discussed this previously, we really don't want to litter the code with "if provider x do y, else do z", that is not sustainable with more providers.

Instead, we should expand the provider interface to surface the result we need.

BenTheElder:
> we do some magic to capture the dns traffic

There is a TODO that we should stop manipulating the iptables and run a minimal userspace proxy + swap it into resolv.conf instead.

pkg/cluster/internal/loadbalancer/config.go
```diff
@@ -40,7 +41,11 @@ global
   maxconn 100000
 
 resolvers docker
+  {{- if eq .ProviderName "podman" -}}
+  parse-resolv-conf
+  {{- else }}
   nameserver dns 127.0.0.11:53
```
BenTheElder: we should probably stop depending on this hardcoded value. do we not have a valid value in resolv.conf when using docker @aojea?

pkg/cluster/internal/providers/docker/provider.go
BenTheElder: But also, I'm wondering why we can't use resolv.conf with docker as well ...

pkg/cluster/internal/loadbalancer/config.go
aojea: yeah, that is what I was trying to verify

aojea: ok, it gets the 127.0.0.11 alreadt

```
more /usr/local/google/docker/containers/f24af881ca241756bcdc06eaace46bd6fc1aac2c703c301a182f4f2ab0f667dc/resolv.conf
[sudo] password for aojea:
search <omitted>
nameserver 127.0.0.11
nameserver 2001:4860:4860::8888
nameserver 2001:4860:4860::8844

```

@tchap can you just use `parse-resolv-conf` directly and remove the 127.0.0.11 hardcoded ip

aojea:
```suggestion
  parse-resolv-conf
```

tchap: I rolled back and amended the first commit to replace the NS line in haproxy.conf to only contain `parse-resolv-conf`.

aojea:
hmm, the loadbalancer works but I have a problem with etcd trying to connect  to a different IP

```
{"level":"warn","ts":"2025-07-01T10:10:29.755509Z","caller":"etcdserver/cluster_util.go:294","msg":"failed to reach the peer URL","address":"https://192.168.8.9:2380/version","remote-member-id":"a3a38fe92c9e1adf","error":"Get \"https://192.168.8.9:2380/version\": dial tcp 192.168.8.9:2380: connect: connection refused"}
{"level":"warn","ts":"2025-07-01T10:10:29.755549Z","caller":"etcdserver/cluster_util.go:158","msg":"failed to get version","remote-member-id":"a3a38fe92c9e1adf","error":"Get \"https://192.168.8.9:2380/version\": dial tcp 192.168.8.9:2380: connect: connection refused"}

```

aojea:
ok, it seems is something related to resources, with two nodes it works and the lb resolvs the containers correctly

```
 docker logs 307b0e6aaf52
[WARNING] 181/105817 (1) : config : missing timeouts for frontend 'controlPlane'.
   | While not properly invalid, you will certainly encounter various problems
   | with such a configuration. To fix this, please ensure that all following
   | timeouts are set to a non-zero value: 'client', 'connect', 'server'.
[NOTICE] 181/105817 (1) : New worker #1 (8) forked
[WARNING] 181/105820 (1) : Reexecuting Master process
[NOTICE] 181/105820 (1) : haproxy version is 2.2.9-2+deb11u5
[NOTICE] 181/105820 (1) : path to executable is /usr/sbin/haproxy
[ALERT] 181/105820 (1) : sendmsg()/writev() failed in logger #1: No such file or directory (errno=2)
[WARNING] 181/105820 (8) : Stopping frontend controlPlane in 0 ms.
[WARNING] 181/105820 (8) : Stopping backend kube-apiservers in 0 ms.
[WARNING] 181/105820 (8) : Stopping frontend GLOBAL in 0 ms.
[WARNING] 181/105820 (8) : Proxy controlPlane stopped (cumulated conns: FE: 0, BE: 0).
[NOTICE] 181/105820 (1) : New worker #1 (75) forked
[WARNING] 181/105820 (8) : Proxy kube-apiservers stopped (cumulated conns: FE: 0, BE: 0).
[WARNING] 181/105820 (8) : Proxy GLOBAL stopped (cumulated conns: FE: 0, BE: 0).
[WARNING] 181/105820 (1) : Former worker #1 (8) exited with code 0 (Exit)
[WARNING] 181/105820 (75) : kube-apiservers/mn-control-plane changed its IP from (none) to 192.168.8.8 by docker/127.0.0.11.
[WARNING] 181/105820 (75) : Server kube-apiservers/mn-control-plane ('mn-control-plane') is UP/READY (resolves again).
[WARNING] 181/105820 (75) : Server kube-apiservers/mn-control-plane administratively READY thanks to valid DNS answer.
[WARNING] 181/105820 (75) : kube-apiservers/mn-control-plane2 changed its IP from (none) to 192.168.8.7 by docker/127.0.0.11.
[WARNING] 181/105820 (75) : Server kube-apiservers/mn-control-plane2 ('mn-control-plane2') is UP/READY (resolves again).
[WARNING] 181/105820 (75) : Server kube-apiservers/mn-control-plane2 administratively READY thanks to valid DNS answer.
[WARNING] 181/105821 (75) : Server kube-apiservers/mn-control-plane2 is DOWN, reason: Layer4 connection problem, info: "SSL handshake failure", check duration: 0ms. 1 active and 0 backup servers left. 0 sessions active, 0 requeued, 0 remaining in queue.
[WARNING] 181/105822 (75) : Server kube-apiservers/mn-control-plane is DOWN, reason: Layer4 connection problem, info: "SSL handshake failure", check duration: 0ms. 0 active and 0 backup servers left. 0 sessions active, 0 requeued, 0 remaining in queue.
[ALERT] 181/105822 (75) : backend 'kube-apiservers' has no server available!
[WARNING] 181/105833 (75) : Server kube-apiservers/mn-control-plane is UP, reason: Layer7 check passed, code: 200, check duration: 3ms. 1 active and 0 backup servers online. 0 sessions requeued, 0 total in queue.
[WARNING] 181/105849 (75) : Server kube-apiservers/mn-control-plane2 is UP, reason: Layer7 check passed, code: 200, check duration: 4ms. 2 active and 0 backup servers online. 0 sessions requeued, 0 total in queue.

```

/lgtm
/approve

Thanks for fixing this

PR #3965, opened 2025-07-08
WIP: tag with git hash

BenTheElder:
Something changed and our automatic image builds now have fancy git describe output instead of just the date+hash.

This would ensure we keep using date+hash.

I think that's probably the best move for now.

### DIFF
--- a/images/Makefile.common.in
+++ b/images/Makefile.common.in
@@ -21,7 +21,7 @@ REGISTRY?=gcr.io/k8s-staging-kind
 # for appending build-meta like "_containerd-v1.7.1"
 TAG_SUFFIX?=
 # tag based on date-sha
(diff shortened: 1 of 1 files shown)


BenTheElder:
@aojea @stmcginnis WDYT?

aojea:
you are the expert on these things

BenTheElder:
I think we should, because the commit hash only changes if we rewrite history.

BenTheElder:
Also I think the git tags detected are wrong on recently pushed base etc images (v0.29.0....) ... better to just drop that, we only need date-sha for uniqueness and traceability, we're not advertising semver in these images

stmcginnis:
Yeah, makes sense to me. And good point about semver. I think that may have been misleading to some people.

/lgtm

PR #3969, opened 2025-07-10
Drop usages sigs.k8s.io/yaml/goyaml.v3

dims:
xref: https://github.com/kubernetes/kubernetes/issues/132056

Switch usages of `"sigs.k8s.io/yaml/goyaml.v2"` to `"go.yaml.in/yaml/v2"` to drop usages of forked code.


go.mod
```diff
@@ -19,6 +19,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/spf13/cobra v1.8.0
 	github.com/spf13/pflag v1.0.5
+	go.yaml.in/yaml/v3 v3.0.4
 	sigs.k8s.io/yaml v1.4.0
```
BenTheElder, line 23: should we also sigs.k8s.io/yaml ?

BenTheElder (commented):
/lgtm
/approve
/hold
one question

go.mod
dims: you mean remove it entirely? there's some other things in the root folder of that package we see to use, so just removing the yaml v3 parser here in this PR.

see other usages here:
https://cs.k8s.io/?q=%22sigs.k8s.io%5C%2Fyaml%22&i=nope&literal=nope&files=&excludeFiles=&repos=kubernetes-sigs/kind

dims:
> /hold 
one question

There are some useful things in the root folder of that package, so we can leave those as-is and just remove the duplicated parser usage.

stmcginnis:
Reading the linked issue and some of the background on this, this looks like good clean up to do. And makes sense to keep around the forked module for the things that are remaining there.

/lgtm

dims:
thanks @stmcginnis 

/hold cancel

PR #3974, opened 2025-07-18
feat: speed up ci

ls-2018:


stmcginnis:
`cache: true` is the action's default since v4. Is there any reason we would need to set it explicitly?

PR #3981, opened 2025-08-07
Add priority class to kindnet daemonset

tom1299:
This PR adds the priority class `system-node-critical` to the kindnet daemonset to avoid preemption.

### DIFF
--- a/pkg/build/nodeimage/const_cni.go
+++ b/pkg/build/nodeimage/const_cni.go
@@ -104,6 +104,7 @@ spec:
         kubernetes.io/os: linux
       tolerations:
       - operator: Exists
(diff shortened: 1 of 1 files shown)


aojea:
Did this happen or you just spotted it?

tom1299:
> Did this happen or you just spotted it?

It really happend. I could share the kind config and procedure how to reproduce it.

stmcginnis:
Seems reasonable to me. I could see how this could be an issue if you're not careful testing certain workloads. At least I can't think of any harm this should cause.

/ok-to-test
/lgtm

aojea:
good catch on a subtle problem

PR #3980, opened 2025-08-05
WIP: refactor log collection, collect metrics

BenTheElder:

PR #3986, opened 2025-08-12
refactor log collection

BenTheElder:
pulled out from https://github.com/kubernetes-sigs/kind/pull/3980/commits

I'm not sure if we should actually dump the metrics on each node or not, and currently I'm pursuing a KEP related to how we'd test metrics which is going to take another route ...

So in the meantime, splitting out the cleanup from the behavior changes.

### DIFF
--- a/pkg/cluster/internal/providers/common/logs.go
+++ b/pkg/cluster/internal/providers/common/logs.go
@@ -3,50 +3,7 @@ package common
 import (
 	"os"
 	"path/filepath"
-
-	"sigs.k8s.io/kind/pkg/cluster/nodes"
-	"sigs.k8s.io/kind/pkg/errors"
-	"sigs.k8s.io/kind/pkg/exec"
 )
-
-// CollectLogs provides the common functionality
-// to get various debug info from the node
-func CollectLogs(n nodes.Node, dir string) error {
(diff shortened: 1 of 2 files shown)


pkg/cluster/internal/providers/docker/provider.go
```diff
@@ -254,34 +254,20 @@ func (p *provider) CollectLogs(dir string, nodes []nodes.Node) error {
...
 	for _, n := range nodes {
 		node := n // https://golang.org/doc/faq#closures_and_goroutines
 		name := node.String()
 		path := filepath.Join(dir, name)
-		if err := internallogs.DumpDir(p.logger, node, "/var/log", path); err != nil {
-			errs = append(errs, err)
-		}
-
+		fns = append(fns, func() error {
+			return internallogs.DumpDir(p.logger, node, "/var/log", path)
```
stmcginnis, line 263: Isn't this being done for all providers at the top level?

BenTheElder: Looks like this got left behind in the docker provider but not the others, will follow up 😳

BenTheElder: https://github.com/kubernetes-sigs/kind/pull/3988

PR #3992, opened 2025-08-21
Be more clear on the expected format of --wait option

stmcginnis:
Follow up from #3418 and #3419.

`kind create cluster --help` does not explain what is the expected format of the `--wait` option. This is described in the quick start, but not in the command line help output. This change adds more a more verbose description to make it more clear.

Thanks for @radeksm for the original suggestion and work on this.


TerryHowe (approved):
/lgtm

PR #4000, opened 2025-09-01
Use YAMLToJSONStrict instead of YAMLToJSON when loading patches

tom1299:
Currently when yamls with duplicate key entries are used for config patches, neither a warning nor an error message is displayed. For example using this malformed patch (See duplicate key in the last line) :
```yaml
kind: Cluster
nodes:
- role: control-plane
kubeadmConfigPatches:
- |
  kind: InitConfiguration
  nodeRegistration:
    kubeletExtraArgs:
      anonymous-auth: "true"
      anonymous-auth: "false"
``` 
with `kind create cluster --config ./kind-config.yaml -v=9` will result in the following `kubeadm.config` (only relavant lines are shown):
```bash
  kubeletExtraArgs:
    anonymous-auth: "false"
``` 
This PR contains a change to use `YAMLToJSONStrict` instead of the more relaxed `YAMLToJSON`. I also added a test case for an example with a duplicate key.

### DIFF
--- a/pkg/internal/patch/kubeyaml.go
+++ b/pkg/internal/patch/kubeyaml.go
@@ -17,12 +17,11 @@ limitations under the License.
 package patch
 
 import (
-	"sigs.k8s.io/kind/pkg/log"
 	"strings"
 
 	"sigs.k8s.io/kind/pkg/errors"
-
 	"sigs.k8s.io/kind/pkg/internal/apis/config"
+	"sigs.k8s.io/kind/pkg/log"
 )
 
 // KubeYAML takes a Kubernetes object YAML document stream to patch,
--- a/pkg/internal/patch/kubeyaml.go
+++ b/pkg/internal/patch/kubeyaml.go
@@ -17,6 +17,7 @@ limitations under the License.
 package patch
 
 import (
+	"sigs.k8s.io/kind/pkg/log"
 	"strings"
 
 	"sigs.k8s.io/kind/pkg/errors"
@@ -34,13 +35,13 @@ import (
 //
 // Patches match if their kind and apiVersion match a document, with the exception
 // that if the patch does not set apiVersion it will be ignored.
-func KubeYAML(toPatch string, patches []string, patches6902 []config.PatchJSON6902) (string, error) {
+func KubeYAML(toPatch string, patches []string, patches6902 []config.PatchJSON6902, logger log.Logger) (string, error) {
 	// pre-process, including splitting up documents etc.
 	resources, err := parseResources(toPatch)
 	if err != nil {
 		return "", errors.Wrap(err, "failed to parse yaml to patch")
 	}
-	mergePatches, err := parseMergePatches(patches)
+	mergePatches, err := parseMergePatches(patches, logger)
 	if err != nil {
 		return "", errors.Wrap(err, "failed to parse patches")
 	}
--- a/pkg/internal/patch/mergepatch.go
+++ b/pkg/internal/patch/mergepatch.go
@@ -44,7 +44,7 @@ func parseMergePatches(rawPatches []string) ([]mergePatch, error) {
 		if err != nil {
 			return nil, errors.WithStack(err)
 		}
-		json, err := yaml.YAMLToJSON([]byte(raw))
+		json, err := yaml.YAMLToJSONStrict([]byte(raw))
 		if err != nil {
 			return nil, errors.WithStack(err)
 		}
--- a/pkg/internal/patch/mergepatch.go
+++ b/pkg/internal/patch/mergepatch.go
@@ -17,9 +17,10 @@ limitations under the License.
 package patch
 
 import (
+	"sigs.k8s.io/yaml"
+
 	"sigs.k8s.io/kind/pkg/errors"
 	"sigs.k8s.io/kind/pkg/log"
-	"sigs.k8s.io/yaml"
 )
 
 type mergePatch struct {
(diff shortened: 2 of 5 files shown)


stmcginnis:
Some may expect the more lenient handling at this point. Not sure how big of an impact this would be.

tom1299:
@stmcginnis I was thinking that as well. `kubeadm init` displays a warning but continues to setup the node when the `kubeadm.config` contains duplicate keys. I could alter the logic to first try strict, and if there was an error, display that and do it the lenient way. That would reflect the behavior of `kubeadm`.

tom1299:
@stmcginnis I changed the implementation to a more permissive approach which just writes a warning to stderr:
```bash
$ kind create cluster --config ./kind-config.yaml 
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.34.0) 🖼
 ✓ Preparing nodes 📦  
⠈⠑ Writing configuration 📜 
WARN: Error converting patch of kind InitConfiguration to json: "yaml: unmarshal errors:
  line 5: key "anonymous-auth" already set in map". Trying more permissive conversion.
 ✓ Writing configuration 📜 
 ✓ Starting control-plane 🕹️ 
 ✓ Installing CNI 🔌 
 ✓ Installing StorageClass 💾 

Thanks for using kind! 😊
```
The warning messes up the output a little bit though. Maybe there is a better way to log this error (?) This change would not impact existing malformed patches and would reflect the current behaviour of `kubeadm init` as this command also logs a warning and continues.

pkg/internal/patch/mergepatch.go
```diff
@@ -44,9 +46,19 @@ func parseMergePatches(rawPatches []string) ([]mergePatch, error) {
 		if err != nil {
 			return nil, errors.WithStack(err)
 		}
-		json, err := yaml.YAMLToJSON([]byte(raw))
+		json, err := yaml.YAMLToJSONStrict([]byte(raw))
 		if err != nil {
-			return nil, errors.WithStack(err)
+			fmt.Fprintf(
```
stmcginnis: I do like this approach overall, but a little concerned about writing to stderr. Especially using format encoding that might not be supported on the terminal or output device where that is being sent to.

It does muddy the output to have warnings inline with progress information, but I think that part is fine. Could you switch to using `sigs.k8s.io/kind/pkg/log` and make this a `logger.Warn` call instead so output handling remains somewhat centralized?

BenTheElder: the logging utils will handle the output coordination

tom1299: Thanks for the advise. I changed the code to use the log module and also adjusted the log message to sound less alarming:
```go
logger.Warnf("Failed to strictly convert patch of kind %s to json: \"%v\". "+
	"Trying more permissive conversion.",
	matchInfo.Kind,
	err,
)
```
And the log looks much better now:
```bash
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.34.0) 🖼
 ✓ Preparing nodes 📦  
Failed to strictly convert patch of kind InitConfiguration to json: "yaml: unmarshal errors:
  line 5: key "anonymous-auth" already set in map". Trying more permissive conversion.
 ✓ Writing configuration 📜 
 ✓ Starting control-plane 🕹️ 
 ✓ Installing CNI 🔌 
 ✓ Installing StorageClass 💾 
```
Although I had to add an additional parameter to some functions to pass the logger to the mergepatch.

pkg/internal/patch/kubeyaml.go
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 package patch
 
 import (
+	"sigs.k8s.io/kind/pkg/log"
```
stmcginnis: I believe this will cause linting failures. Need to have the separate stdlib, third party lib, local grouping.

pkg/internal/patch/mergepatch.go
```diff
@@ -17,9 +17,9 @@ limitations under the License.
 package patch
 
 import (
-	"sigs.k8s.io/yaml"
-
 	"sigs.k8s.io/kind/pkg/errors"
+	"sigs.k8s.io/kind/pkg/log"
+	"sigs.k8s.io/yaml"
```
stmcginnis: This too.

stmcginnis (commented):
A couple small things that will cause CI failure.

Not sure if we have a better way to get the logger instance. I know some people using `kind` as a library, so I don't love changing the method signatures. But if that's what we have to do...

pkg/internal/patch/mergepatch.go
tom1299: Thanks for spotting that. I corrected the imports

tom1299:
I didn't like changing the signatures of the functions either. Especially for such a small change. But I couldn't really find a better way.

stmcginnis:
/ok-to-test

@BenTheElder @aojea what do you think about the method signature change? I think it may be worth it to have this output. I actually have run into an instance where some templating errors were duplicating keys and it was not at all obvious why the resulting cluster wasn't configured right.

aojea:
+1 to strict parsing, I got bitted several times because of that and it takes a lot of time to figure out the problem

@BenTheElder is the best person for the plumbing of the logger

BenTheElder:
> +1 to strict parsing, I got bitted several times because of that and it takes a lot of time to figure out the problem

Yeah, I'm generally a fan of strict parse but ...  it has bitten during enablement in the past.

I think we should hold off temporarily since we have some other critical fixes that we wouldn't want to cause adoption issues for but I think this makes sense long term

PR #4003, opened 2025-09-09
Bump image from bullseye to bookworm for local-path-helper

shwetha-s-poojary:
### Changes:
1. Bumped base image from Debian Bullseye to Bookworm

2. Fixed dpkg-query failure due to path mismatch
   **Root cause:** The build script uses `which bash`, resolving to /usr/bin/bash. On systems with merged /usr layout, this path is valid, but dpkg-query only matches exact paths recorded in the package database—typically /bin/bash. (_mismatch occurs because dpkg-query doesn’t follow symlinks or filesystem structure; it relies solely on the original install path recorded by the package_)
   **Fix:** Used wildcard match — `dpkg-query -S '*/bin/bash'`

### **Testing:**
- Image built successfully using Podman
- Verification pending via Makefile-based tests

### DIFF
--- a/images/local-path-helper/stage-binary-and-deps.sh
+++ b/images/local-path-helper/stage-binary-and-deps.sh
@@ -30,7 +30,8 @@ file_to_package() {
     # `dpkg-query --search $file-pattern` outputs lines with the format: "$package: $file-path"
     # where $file-path belongs to $package
     # https://manpages.debian.org/jessie/dpkg/dpkg-query.1.en.html
-    dpkg-query --search "$(realpath "${1}")" | cut -d':' -f1
+    # Match bash path in dpkg regardless of /usr merge
+    dpkg-query --search "$(realpath "${1}" | sed 's|/usr|*|')" | cut -d':' -f1
 }
 
 # package_to_copyright gives the path to the copyright file for the package $1
--- a/images/local-path-helper/Dockerfile
+++ b/images/local-path-helper/Dockerfile
@@ -15,7 +15,7 @@
 # This image is contains the binaries needed for the local-path-provisioner
 # helper pod. Currently that means: sh, rm, mkdir
 
-ARG BASE="registry.k8s.io/build-image/debian-base:bullseye-v1.4.3"
+ARG BASE="registry.k8s.io/build-image/debian-base:bookworm-v1.0.5"
 FROM ${BASE} AS build
 
 # NOTE: copyrights.tar.gz is a quirk of Kubernetes's debian-base image
@@ -42,6 +42,6 @@ RUN mkdir -p "${STAGE_DIR}" && \
     find "${STAGE_DIR}"
 
 # copy staged binary + deps + copyright into distroless
-FROM "gcr.io/distroless/static-debian11"
+FROM "gcr.io/distroless/static-debian12"
 ARG STAGE_DIR="/opt/stage"
 COPY --from=build "${STAGE_DIR}/" /
--- a/images/local-path-provisioner/Dockerfile
+++ b/images/local-path-provisioner/Dockerfile
@@ -29,7 +29,7 @@ RUN eval "$(gimme "${GO_VERSION}")" \
     && GOBIN=/usr/local/bin go install github.com/google/go-licenses@latest \
     && GOARCH=$TARGETARCH go-licenses save --save_path=/_LICENSES .
 
-FROM gcr.io/distroless/base-debian11
+FROM gcr.io/distroless/base-debian12
 COPY --from=0 /usr/local/bin/local-path-provisioner /usr/local/bin/local-path-provisioner
 COPY --from=0 /_LICENSES/* /LICENSES/
 COPY --chmod=0644 files/LICENSES/* /LICENSES/*


shwetha-s-poojary:
/retitle Bump image from bullseye to bookworm for local-path-helper

aojea:
last time we had some issues with the bump so you should build the image and test it works

BenTheElder:
> last time we had some issues with the bump so you should build the image and test it works

yeah, this one may be more involved (but necessary at some point!)

recalling now https://github.com/kubernetes-sigs/kind/pull/3350 ...

images/local-path-helper/stage-binary-and-deps.sh
```diff
@@ -30,7 +30,8 @@ file_to_package() {
     # `dpkg-query --search $file-pattern` outputs lines with the format: "$package: $file-path"
     # where $file-path belongs to $package
     # https://manpages.debian.org/jessie/dpkg/dpkg-query.1.en.html
-    dpkg-query --search "$(realpath "${1}")" | cut -d':' -f1
+    # Match bash path in dpkg regardless of /usr merge
+    dpkg-query --search "$(realpath "${1}" | sed 's|/usr|*|')" | cut -d':' -f1
 }
```
shwetha-s-poojary, line 36: Replaced /usr/bin/bash with a wildcard match (*/bin/bash) in dpkg-query to handle systems with merged /usr layout. This ensures compatibility where bash is symlinked to /usr/bin/bash, but only /bin/bash is recorded in the package database

BenTheElder: Instead of sed, we can realpath the output path?

shwetha-s-poojary: realpath resolves to /usr/bin/bash on merged /usr systems, but dpkg-query only matches /bin/bash in the package database. The sed wildcard workaround does work for these cases, though I’m exploring whether there’s a more canonical or robust way to handle this mismatch. I’d appreciate any suggestions you might have.

shwetha-s-poojary:
Hey @bentheelder

I've built the local-path-provisioner and helper and the kindnetd image using the debian12 base image. I had used them in a node image that pointed to the same and things work as expected in my amd64 test environment.
Please let me know if we should be good to merge these changes or if you have any other concerns, will be happy to address.

Thanks,
Shwetha

stmcginnis:
Change makes sense to me. Built image locally and everything appears to be find.

/ok-to-test
/lgtm

images/local-path-helper/stage-binary-and-deps.sh
BenTheElder: Sorry I haven't had time to look into this problem myself, but I remain concerned about this being fragile. I hope we can find a generalized solution.

shwetha-s-poojary: @BenTheElder Agree it’s fragile. For now, this is a workable approach—let me take some time to explore a more generalized solution.

shwetha-s-poojary: @BenTheElder a simpler approach is to resolve the path, strip /usr if it exists, and then query dpkg. What are your thoughts on this ?

BenTheElder: https://wiki.debian.org/UsrMerge is down :(

I'll try again later to load this and get context for the best answer ...

BenTheElder:
the image build environment currently needs fixing (ongoing), otherwise just the thread about the usr merge situation to sort out, sorry!

images/local-path-helper/stage-binary-and-deps.sh
BenTheElder: It loads now:

> This page tracks Debian support for the merged /usr directories scheme, i.e. the /{bin,sbin,lib}/ directories becoming symbolic links to /usr/{bin,sbin,lib}/.

Ok so we should probably handle all of these paths. Dropping `/usr` seems like an OK workaround, can we add a comment pointing to this debian docs page for now? Thanks!

BenTheElder:
[the builds are fixed and the debian docs are loading now, commented]

shwetha-s-poojary:
@BenTheElder have added the docs pointer in comments.

BenTheElder:
Thanks!

I'm about to tag a new release, I want to include this in the next one (so we have more time to sort out any issues we haven't caught yet with updating)

Will circle back to make sure this merges before EOW, hopefully just later today.

BenTheElder (commented):
/lgtm
thank you!!
/hold cancel

BenTheElder:
the haproxy build had issues:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/post-kind-push-haproxy-image/2001487399288311808

local path went well:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/post-kind-push-local-path-helper-image/2001487399443501056

and local helper:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/post-kind-push-local-path-helper-image/2001487399443501056

those two are available at kindest/ now, but haproxy will need followup.

kindest/local-path-helper:v20251218-c8778675
kindest/local-path-provisioner:v20251218-c8778675

BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/4081 will fix the haproxy build issue I think, then for all three we'll want to pick them up in a subsequent PR updating the versions referenced in the kind code and double check HA clusters.

It may be that in the time this was open the available versions changed, and anyhow we intentionally decoupled changes to these images from picking them up in kind, so we can iterate :-)

thanks again for working on this!

BenTheElder:
Hmm, these are not working:
```
$ docker run --rm docker.io/kindest/local-path-helper:v20251218-c8778675 /usr/bin/sh -c 'echo "hello"'
/usr/bin/sh: error while loading shared libraries: libtinfo.so.6: cannot open shared object file: No such file or directory
```

images/haproxy/stage-binary-and-deps.sh
```diff
@@ -30,7 +30,9 @@ file_to_package() {
     # `dpkg-query --search $file-pattern` outputs lines with the format: "$package: $file-path"
     # where $file-path belongs to $package
     # https://manpages.debian.org/jessie/dpkg/dpkg-query.1.en.html
-    dpkg-query --search "$(realpath "${1}")" | cut -d':' -f1
+    # Handle Debian merged-/usr layout
+    # https://wiki.debian.org/UsrMerge
+    dpkg-query --search "$(realpath "${1}" | sed 's|/usr|*|')" | cut -d':' -f1
```
BenTheElder, line 35: not sure how but we didn't need to do this in https://github.com/kubernetes/release/blob/8dbdcf021e1e5745072b9c1e2951070771d78079/images/build/distroless-iptables/distroless-bookworm/package-utils.sh

BenTheElder: I see, there are other changes there, will update in kind

BenTheElder:
=> https://github.com/kubernetes-sigs/kind/pull/4103 for follow-up

shwetha-s-poojary:
> Hmm, these are not working:
> 
> ```
> $ docker run --rm docker.io/kindest/local-path-helper:v20251218-c8778675 /usr/bin/sh -c 'echo "hello"'
> /usr/bin/sh: error while loading shared libraries: libtinfo.so.6: cannot open shared object file: No such file or directory
> ```

Issue with base image ?

BenTheElder:
It's fixed now, this was tricky, see #4103 which also leapfrogged forward to trixie 😅 

thanks!

PR #4016, opened 2025-09-27
Update WSL2 Session Affinity instructions

shahar1:
I tried to build the kernel required for supporting session affinity with the latest WSL2 version, it failed because cpio was missing.


stmcginnis:
Thanks!

/ok-to-test
/lgtm
/approve

PR #4017, opened 2025-09-28
Add CONTROLLERS environment variable to e2e-k8s.sh

danwinship:
This allows overriding kube-controller-manager's `--controllers` flag, which will be used by the new periodic e2e test for [KEP-4974](https://github.com/kubernetes/enhancements/issues/4974), which doesn't exist yet.

I thought about naming the variable `KCM_CONTROLLERS` or something instead, but having it be _just_ the name of the CLI arg was consistent with `FEATURE_GATES` and `RUNTIME_CONFIG`...

### DIFF
--- a/hack/ci/e2e-k8s.sh
+++ b/hack/ci/e2e-k8s.sh
@@ -38,6 +38,9 @@ set -o errexit -o nounset -o xtrace
 #          JSON or YAML encoding of a string/string (!) map: {"apia.example.com/v1alpha1": "true", "apib.example.com/v1beta1": "false"}
 #          Enables API groups in the apiserver via --runtime-config.
 #          Cannot be used when GA_ONLY=true.
+# CONTROLLERS:
+#          Value for kube-controller-manager's "--controllers" argument.
+#          e.g. "-endpoints-controller,-endpointslice-mirroring-controller,*"
 
 # cleanup logic for cleanup on exit
 CLEANED_UP=false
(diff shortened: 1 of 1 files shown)


aojea:
for ad-hoc jobs we use to make a copy of this script and parametrize it accordenly , see https://github.com/kubernetes/test-infra/tree/master/experiment , specifically 
- https://github.com/kubernetes/test-infra/blob/master/experiment/kind-multizone-e2e.sh
- https://github.com/kubernetes/test-infra/blob/master/experiment/kind-logs-e2e-k8s.sh
- https://github.com/kubernetes/test-infra/blob/master/experiment/kind-detect-local-e2e.sh

then setting the job is something like

https://github.com/kubernetes/test-infra/blob/a190756db87de069764a104f1c024ecdd052257b/config/jobs/kubernetes/sig-network/sig-network-kind.yaml#L217-L263

This avoid growing the surface of this script and end in a situation like we had in-tree with the cluster-up scripts.

Let's see what @BenTheElder thinks about this too,

danwinship:
Ah, I should have looked more carefully at the various jobs...

Does e2e-k8s.sh ever get bugfixes? It seems like this approach would lead to divergence between the different jobs. At least, maybe there should only be _one_ fork of e2e-k8s.sh in test-infra, that handles all the variants?

Anyway, I'm happy to move this to test-infra. I doubt it would turn out to be that generically useful in the future.

aojea:
> At least, maybe there should only be _one_ fork of e2e-k8s.sh in test-infra, that handles all the variants?

this script is a wrapper to launch a kind cluster and run the ginkgo tests, the only changes it had was to add some new env variables to parametrize it.
If we use one to handle variants then it needs an owner, the risk of adding a bug or some change that impact other jobs exist.
If we fork and make it only per job, then the only risk is for that specific job that has a clear owner and once the job is configured it can not fail unless kind or kubernetes e2e test make a breaking change ... but then this specific job will be the less of our problems

Issue #3958, opened 2025-06-14
Apple Containerization framework


hown3d:
**What would you like to be added**:
Add support for [macOS containerization framework](https://github.com/apple/container) as a provider for kind clusters.

**Why is this needed**:
Apple announced there built-in containerization framework to run OCI compliant containers on macOS. This could allow users of kind on macOS to run kind clusters without the need for podman, docker etc.

aojea:
See some related job in https://www.linkedin.com/posts/jonesax_45-mins-in-kinda-got-kind-working-with-apple-activity-7338288332575039489-y28u

BenTheElder:
I have talked to folks at Apple about this, I think we should hold off for now while it matures and hopefully becomes easier to integrate. There should be more coming ...

We already have multiple under-supported (we need help from all of you!) backends other than docker.

We also definitely do not want to link directly to a library like this.

jglogan:
Hi, I'm one of the developers on the team that maintains the https://github.com/apple/containerization and https://github.com/apple/container projects. We've done some work on a backend provider for `kind` because many users rely on it for orchestrating containerized apps for local development. It helps users bridge the gap between their bare applications and deployment to production clouds.

The changes we currently have can be divided into three parts:

* A change to add move the APIs that support `kind load docker-image` into the provider, so that the command can `container image save` one or more images into a tar file for loading into cluster nodes.
* Since `container` maps containers 1:1 to lightweight VMs, there are additional config spec properties `nodes[].kernel`, `nodes[].resources.cpu` and `nodes[].resources.memory` to configure the VMs accordingly. These fields are nullable and should be ignored by a provider if they are not applicable for the backend.
* Finally, the `apple-container` provider itself, which wraps invocations of the `container` tool and marshals output into provider API return values.

@BenTheElder and I chatted recently about the approach and I'll summarize the issues here, and Ben and others can correct me if I misunderstood anything.

The principal issue isn't so much about the implementation details above, but relates to this part of Ben's previous comment:

> We already have multiple under-supported (we need help from all of you!) backends other than docker.

According to the [project scope](https://kind.sigs.k8s.io/docs/contributing/project-scope/), `kind` is first and foremost a tool for validating Kubernetes and parts of its ecosystem. That it's fantastic as a cloud native development tool is a happy side effect. Supporting different container backends isn't in scope (nor is it explicitly out of scope). As a result, every new provider adds support burden that can distract from the project's mission.

[...]

PR #4019, opened 2025-09-29
Detect ip6tables failure without full ipv6 support

afbjorklund:
Error when running with kernel from Kata Containers:
```
    can't initialize ip6tables table `filter':
    Table does not exist (do you need to insmod?)
    Perhaps ip6tables or your kernel needs to be upgraded.
```

This happens also in Apple Containers. Fallback to ipv4 only.
The alternative is building a custom kernel, with full support.

This fixes the `kind create cluster` error that was encountered in:
* https://github.com/kubernetes-sigs/kind/issues/3958#issuecomment-3339348358

----

Assuming that user has redirected from nft to legacy, that is.

```
update-alternatives --set iptables /usr/sbin/iptables-legacy
update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
```

Got some other error, when trying to create with default iptables.

```console
# docker network create -d=bridge -o com.docker.network.bridge.enable_ip_masquerade=true -o com.docker.network.driver.mtu=1500 kind
Error response from daemon: Failed to Setup IP tables: Unable to enable NAT rule:  (iptables failed: iptables --wait -t nat -I POSTROUTING -s 172.18.0.0/16 ! -o br-b505b0155599 -j MASQUERADE: Warning: Extension MASQUERADE revision 0 not supported, missing kernel module?
iptables v1.8.10 (nf_tables): Could not fetch rule set generation id: Invalid argument
 (exit status 4))
```

* Container 0.4.1
* Linux 6.12.28

https://github.com/apple/container/blob/main/docs/tutorial.md

https://github.com/kata-containers/kata-containers/releases/tag/3.17.0

* Ubuntu 24.04.3 LTS
* Docker 28.4.0 CE

----

Here is the regular Ubuntu kernel, for reference:

```
CONFIG_IP_NF_IPTABLES=m
CONFIG_IP6_NF_IPTABLES=m

CONFIG_NF_NAT_MASQUERADE=y
CONFIG_NETFILTER_XT_TARGET_MASQUERADE=m
CONFIG_IP_NF_TARGET_MASQUERADE=m
CONFIG_IP6_NF_TARGET_MASQUERADE=m
```

And here is the Kata Containers kernel, no IPv6:

```
# zgrep IPTABLES /proc/config.gz
CONFIG_IP_NF_IPTABLES_LEGACY=y
CONFIG_IP_NF_IPTABLES=y
# CONFIG_IP6_NF_IPTABLES is not set

# zgrep MASQUERADE /proc/config.gz
CONFIG_NF_NAT_MASQUERADE=y
CONFIG_NETFILTER_XT_TARGET_MASQUERADE=y
CONFIG_IP_NF_TARGET_MASQUERADE=y
```

### DIFF
--- a/pkg/cluster/internal/providers/docker/network.go
+++ b/pkg/cluster/internal/providers/docker/network.go
@@ -271,7 +271,10 @@ func isIPv6UnavailableError(err error) bool {
 	// even on hosts that lack ip6tables setup.
 	// Preferably users would either have ip6tables setup properly or else disable ipv6 in docker
 	const dockerIPV6TablesError = "Error response from daemon: Failed to Setup IP tables: Unable to enable NAT rule:  (iptables failed: ip6tables"
-	return strings.HasPrefix(errorMessage, dockerIPV6DisabledError) || strings.HasPrefix(errorMessage, dockerIPV6TablesError)
+	// we get this error when ipv6 is missing in kernel
+	const dockerIPV6PolicyError = "Error response from daemon: setting default policy to DROP in FORWARD chain failed:  (iptables failed: ip6tables"
+
+	return strings.HasPrefix(errorMessage, dockerIPV6DisabledError) || strings.HasPrefix(errorMessage, dockerIPV6TablesError) || strings.HasPrefix(errorMessage, dockerIPV6PolicyError)
 }
 
 func isPoolOverlapError(err error) bool {


BenTheElder:
Thanks for the PR!

> The alternative is building a custom kernel, with full support.

You can disable ipv6 in the docker daemon settings? I think that's the best fix, though kind should gracefully handle this, I do worry the error messages will keep shifting but ... 🤷‍♂️

afbjorklund:
> You can disable ipv6 in the docker daemon settings?

I tried disabling in daemon.json, but it doesn't change.

```json
{
    "ipv6": false
}
```

There seems to be some open Docker tickets, about this.

"I think there's a misunderstanding here due to misleading docs: the daemon's ipv6 flag doesn't disable ipv6 daemon-wide, but it only controls the availability of IPv6 on the default bridge network."

So I disabled it system-wide, with `sysctl` (procps).

```
net.ipv6.conf.all.disable_ipv6=1
net.ipv6.conf.default.disable_ipv6=1
```

But still got that error, when creating the "kind" network.

aojea:
it seems the daemon has an option to disable the ip6tables

https://docs.docker.com/engine/daemon/ipv6/#docker-in-docker

> Alternatively, use daemon option --ip6tables=false to disable ip6tables for the containerized Docker Engine

BenTheElder:
Interesting, thanks.

pkg/cluster/internal/providers/docker/network.go
```diff
@@ -271,7 +271,10 @@ func isIPv6UnavailableError(err error) bool {
 	// even on hosts that lack ip6tables setup.
 	// Preferably users would either have ip6tables setup properly or else disable ipv6 in docker
 	const dockerIPV6TablesError = "Error response from daemon: Failed to Setup IP tables: Unable to enable NAT rule:  (iptables failed: ip6tables"
```
BenTheElder, line 273: We should maybe rename and clarify this one versus the new error. That shouldn't block fixing the bug though

BenTheElder:
/hold cancel
IMO we can iterate on nits, let's land the bug fix

PR #4020, opened 2025-09-29
fix tar extraction add path validation to prevent escape from destination

odaysec:
### Description
This fixes addresses **path traversal vulnerabilities in tar archive extraction** within the `kind` codebase. Without proper validation, specially crafted tar entries could escape the intended extraction directory and overwrite arbitrary files on the host system. This PR introduces safeguards to ensure that extracted files remain confined to the target directory.

https://github.com/kubernetes-sigs/kind/blob/f20102c99eb6211fb125db7e6a9e244c107a9d9a/pkg/cluster/internal/logs/logs.go#L77-L97

* **Issue:** Tar archive entries are written to paths derived directly from the archive without validation. Attackers could craft entries with `..` or absolute paths to escape the destination directory.

  * Compute the absolute path of the intended extraction target using `filepath.Abs` and `filepath.Clean`.
  * Verify that the target path begins with the cleaned absolute destination directory path.
  * Skip extraction (or return an error) for any entry that falls outside of the allowed directory.
  * Ensure this check is performed before any file system operations (e.g., `os.Create`, `os.MkdirAll`).

https://github.com/kubernetes-sigs/kind/blob/f20102c99eb6211fb125db7e6a9e244c107a9d9a/pkg/build/nodeimage/internal/kube/tar.go#L44-L49

* **Issue:** Similar to `untar`, the function does not validate `hdr.Name`, allowing malicious tar entries to escape the intended destination.
  * After joining `hdr.Name` with the target directory, normalize the path using `filepath.Clean`.
  * Confirm that the cleaned path remains within the cleaned target directory.
  * Reject or skip entries containing absolute paths or `..` segments that lead outside the extraction directory.
  * Added a check using `strings.HasPrefix(cleanedPath, cleanedDestDir + string(os.PathSeparator))` to enforce confinement.

### Notes Refferences
* No new dependencies were added; the fix leverages existing `filepath`, `os`, and `strings` packages.
* The behavior of normal, well-formed tarballs remains unchanged.
* Malicious or malformed entries are safely rejected.

### DIFF
--- a/pkg/build/nodeimage/internal/kube/tar.go
+++ b/pkg/build/nodeimage/internal/kube/tar.go
@@ -7,7 +7,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
-
+	"strings"
 	"sigs.k8s.io/kind/pkg/log"
 )
 
@@ -40,13 +40,22 @@ func extractTarball(tarPath, destDirectory string, logger log.Logger) (err error
 			continue
 		}
 
+		// Sanitize and validate archive entry path to prevent Zip Slip (directory traversal)
+		cleanDestDirectory := filepath.Clean(destDirectory)
+		targetPath := filepath.Join(cleanDestDirectory, hdr.Name)
+		cleanTargetPath := filepath.Clean(targetPath)
+		// Ensure that the resulting path is within destDirectory
+		if !strings.HasPrefix(cleanTargetPath, cleanDestDirectory+string(os.PathSeparator)) && cleanTargetPath != cleanDestDirectory {
+			return fmt.Errorf("illegal file path in archive: %s", hdr.Name)
+		}
+
 		if err := os.MkdirAll(
-			filepath.Join(destDirectory, filepath.Dir(hdr.Name)), os.FileMode(0o755),
+			filepath.Dir(cleanTargetPath), os.FileMode(0o755),
 		); err != nil {
 			return fmt.Errorf("creating image directory structure: %w", err)
 		}
 
-		f, err := os.Create(filepath.Join(destDirectory, hdr.Name))
+		f, err := os.Create(cleanTargetPath)
 		if err != nil {
 			return fmt.Errorf("creating image layer file: %w", err)
 		}
--- a/pkg/cluster/internal/logs/logs.go
+++ b/pkg/cluster/internal/logs/logs.go
@@ -58,6 +58,11 @@ func DumpDir(logger log.Logger, node nodes.Node, nodeDir, hostDir string) (err e
 // untar reads the tar file from r and writes it into dir.
 func untar(logger log.Logger, r io.Reader, dir string) (err error) {
 	tr := tar.NewReader(r)
+	// Ensure target dir is absolute and cleaned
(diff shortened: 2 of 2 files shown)


stmcginnis:
Do you see a way where this is exposed in a way that could be exploited? Or was this just an AI recommendation?

BenTheElder:
/ok-to-test
> Do you see a way where this is exposed in a way that could be exploited? Or was this just an AI recommendation?

ummm if you have a malicious process running in the kind node which is bad enough on it's own, but perhaps sandboxed into say, a limactl vm, then this allow overwriting on the host as the host user IF the user calls `kind export logs` (which is pretty common)

for the other one, where you're downloading a kubernetes release, if that is malicious then you're probably in pretty bad shape already, but we should still mitigate it.

pkg/cluster/internal/logs/logs.go
```diff
@@ -74,11 +79,26 @@ func untar(logger log.Logger, r io.Reader, dir string) (err error) {
...
 		rel := filepath.FromSlash(f.Name)
-		abs := filepath.Join(dir, rel)
+		// Compute absolute extraction path
+		abs := filepath.Join(dirAbs, rel)
+		absClean, err := filepath.Abs(abs)
+		if err != nil {
+			return errors.Wrapf(err, "could not get absolute path for extraction file: %v", err)
+		}
+		// Path traversal check: absClean must be within dirAbs
+		// Ensure dirAbs ends with a separator to prevent prefix matching issues
```
BenTheElder: this should be hoisted  out of the loop?

BenTheElder (commented):
I think we should probably be looking at https://go.dev/blog/osroot in the future

unfortunately it would make go 1.24 the minimum version to build kind, which is a little aggressive at the moment

pkg/build/nodeimage/internal/kube/tar.go
```diff
@@ -7,7 +7,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
-
+	"strings"
```
BenTheElder, line 10: nit: this should be grouped with the stdlib imports to match the existing style

- stdlib

- same repo

- k8s.io

- arbitrary github deps

BenTheElder: this is non-blocking but widely used throughout kubernetes and preferred, it helps to quickly understand the dependencies

pkg/cluster/internal/logs/logs.go
BenTheElder: this comment remains, dirAbs isn't loop-dependent

either this should be checking `abs` instead, or it should be hoisted out of the loop.

stmcginnis:
This is waiting for updates and/or responses to some comments. Are you able to get back to this @odaysec?

odaysec:
@stmcginnis @BenTheElder 

Hi sorry for late respond, because im very busy in rl worked at office. I’ve added an additional patch to address the latest feedback and ensure the implementation aligns with the expected behavior. Please let me know if further changes are needed.

BenTheElder:
Hmm, this isn't building currently.

PR #4022, opened 2025-09-30
feat: Clarify Rootless Runtime Requirements

### LINKED ISSUE #3451: Flaky ingress behavior using ingress-nginx and rootless podman

adambkaplan:
Containers are a wonderful Rube Goldberg machine of Linux internals and configuration [1]. This is especially true for container runtimes that support a "rootless" mode, where spawned processes are subject to constraints and limits that are not present when the runtime executes as root.

This change clarifies the existing documentation for launching KinD with a rootless runtime. The guidance is split into logical sections, providing context and justification for each recommended host change. Callouts are made for changes that impact networking components, such as Ingress and Gateway controllers. These generally appear to push default performance guardrails for user containers/processes.

[1] https://en.wikipedia.org/wiki/Rube_Goldberg_machine

Fixes #3451

### DIFF
--- a/site/content/docs/user/ingress.md
+++ b/site/content/docs/user/ingress.md
@@ -21,6 +21,10 @@ Ingress exposes HTTP and HTTPS routes from outside the cluster to services withi
 > **NOTE**: You may also want to consider using [Gateway API](https://gateway-api.sigs.k8s.io/) instead of Ingress.
 > Gateway API has an [Ingress migration guide](https://gateway-api.sigs.k8s.io/guides/migrating-from-ingress/).
 
+> **WARNING**: If you are using a [rootless container runtime], ensure your host is
+> properly configured before creating the KinD cluster. Most Ingress and Gateway controllers will
+> not work if these steps are skipped.
+
 ### Create Cluster
 
 #### Option 1: LoadBalancer
@@ -139,3 +143,4 @@ curl localhost/bar
 
 [LoadBalancer]: /docs/user/loadbalancer/
 [Cloud Provider KIND]: /docs/user/loadbalancer/
+[rootless container runtime]: /docs/user/rootless/
--- a/site/content/docs/user/quick-start.md
+++ b/site/content/docs/user/quick-start.md
@@ -160,6 +160,10 @@ More usage can be discovered with `kind create cluster --help`.
 kind can auto-detect the [docker], [podman], or [nerdctl] installed and choose the available one. If you want to turn off the auto-detect, use the environment variable `KIND_EXPERIMENTAL_PROVIDER=docker`, `KIND_EXPERIMENTAL_PROVIDER=podman` or `KIND_EXPERIMENTAL_PROVIDER=nerdctl` to
 select the runtime.
 
+> **NOTE**: In some distributions (ex: Fedora), the container runtime operates in
+> [rootless mode](/docs/user/rootless) by default. Extra setup is needed for KinD clusters to be fully
+> functional.
+
 ## Interacting With Your Cluster
 
 After [creating a cluster](#creating-a-cluster), you can use [kubectl][kubectl]
--- a/site/content/docs/user/rootless.md
+++ b/site/content/docs/user/rootless.md
@@ -9,41 +9,58 @@ menu:
 Starting with kind 0.11.0, [Rootless Docker](https://docs.docker.com/go/rootless/), [Rootless Podman](https://github.com/containers/podman/blob/master/docs/tutorials/rootless_tutorial.md) and [Rootless nerdctl](https://github.com/containerd/nerdctl/blob/main/docs/rootless.md) can be used as the node provider of kind.
 
 ## Provider requirements
+
 - Docker: 20.10 or later
 - Podman: 3.0 or later
 - nerdctl: 1.7 or later
 
 ## Host requirements
+
+### cgroups v2
+
 The host needs to be running with cgroup v2.
 Make sure that the result of the `docker info` command contains `Cgroup Version: 2`.
 If it prints `Cgroup Version: 1`, try adding `GRUB_CMDLINE_LINUX="systemd.unified_cgroup_hierarchy=1"` to `/etc/default/grub` and
 running `sudo update-grub` to enable cgroup v2.
 
-Also, depending on the host configuration, the following steps might be needed:
+Your host may also need to enable [cgroup delegation](https://systemd.io/CGROUP_DELEGATION/) for daemon-based controller runtimes.
+This is not required for daemonless runtimes, such as podman. Note that this procedure may
+[negatively impact performance](https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/ZMKLS7SHMRJLJ57NZCYPBAQ3UOYULV65/).
 
-- Create `/etc/systemd/system/user@.service.d/delegate.conf` with the following content, and then run `sudo systemctl daemon-reload`:
+To enable cgroup delegation, perform the folowing actions:
 
-  ```ini
-  [Service]
-  Delegate=yes
-  ```
+1. As root, create the directory `/etc/systemd/system/user@.service.d/` if it does not already exist
 
-  (This is not enabled by default because ["the runtime impact of
-  [delegating the "cpu" controller] is still too
-  high"](https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/ZMKLS7SHMRJLJ57NZCYPBAQ3UOYULV65/).
-  Beware that changing this configuration may affect system
-  performance.)
+   ```sh
+   sudo mdkir -p /etc/systemd/system/user@.service.d/
+   ```
+2. As root, create the file `/etc/systemd/system/user@.service.d/delegate.conf` with the following content:
 
-  Please note that:
+   ```ini
+   [Service]
+   Delegate=yes
+   ```
 
-  - `/etc/systemd/system/user@.service.d/` directory needs to be created if not already present on your host
-  - If using Docker and it was already running when this step was done, a restart is needed for the changes to take
-    effect
-      {{< codeFromInline lang="bash" >}}
-      systemctl --user restart docker
-      {{< /codeFromInline >}}
+3. Reload the systemd daemon:
 
-- Create `/etc/modules-load.d/iptables.conf` with the following content:
+   ```sh
+   sudo systemctl daemon-reload
+   ```
+
+4. If using docker, reload the user docker daemon:
+
+   ```sh
+   systemctl --user restart docker
+   ```
+
+### Networking
+
+Containers running in rootless mode are not typically loaded with host-level iptable modules.
+This breaks the behavior of most Ingress and Gateway controllers.
+
+To load the iptable modules into the KinD containers, do the following:
+
+1. As root, create the file `/etc/modules-load.d/iptables.conf` with the following content:
 
   ```
   ip6_tables
@@ -52,13 +69,61 @@ Also, depending on the host configuration, the following steps might be needed:
   iptable_nat
   ```
 
-- If using podman, be aware that by default there is a [limit](https://docs.podman.io/en/v4.3/markdown/options/pids-limit.html#pids-limit-limit) to the number of pids that can be created. This can cause problems like nginx workers inside a container not spawning correctly.
-    - If you want to disable this limit, edit your `containers.conf` file (generally located in `/etc/containers/containers.conf`). Note that this could cause things like pid exhaustion to happen on the host machine. Alternatively, change `0` to your desired new limit:
+2. Restart your system to ensure these changes take effect.
+
+### Increase PID Limits
+
+KinD nodes are represented as individual containers on their hosts. Runtimes such as podman set default
+[process id limits](https://docs.podman.io/en/v4.3/markdown/options/pids-limit.html#pids-limit-limit)
+that may be too low for the node or for a pod running on the node. The NGINX ingress controller is
+[particularly susceptible](https://github.com/kubernetes-sigs/kind/issues/3451) to this issue.
+
+To increase the PID limit, do the following:
+
+1. If using podman, edit your `containers.conf` file (generally located in
+   `/etc/containers/containers.conf` or `~/.config/containers/containers.conf`) to increase the PIDs
+   limit to a desired value (default 4096 on most systems).
 
     ```ini
     [containers]
-    pids_limit = 0
+    pids_limit = 65536
     ```
+
+
+### Increase inotify Limits
+
+As documented in [known issues](/docs/user/known-issues/#pod-errors-due-to-too-many-open-files), pods may
+fail by reaching inotify watch and instance limits. Ingress controllers such as NGINX and Contour
+are particularly susceptible to this issue.
+
+To increase the inotify limits, do the following:
+
+1. As root, create a `.conf` file in `/etc/systctl.d` that increases the `fs.inotify` max user settings:
+
+   ```
+   fs.inotify.max_user_watches = 524288
+   fs.inotify.max_user_instances = 512
+   ```
+
+2. Restart your system for these changes to take effect.
+
+
+### Allow Unprivileged Binding to HTTP(S) Ports
+
+If you use the `extraPortMappings` method to provide ingress to your KinD cluster, you can allow
+the KinD container to bind to ports 80 and 443 on the host. User containers cannot bind to these
+ports by default as they are considered privileged.
+
+To allow a KinD node to bind to ports 80 and/or 443 on the host, do the following:
+
+1. As root, create a `.conf` file in `/etc/systctl.d` that lowers the privileged port start number:
+
+   ```
+   net.ipv4.ip_unprivileged_port_start=80
+   ```
+
+2. Restart your system for these changes to take effect.
+
 
 ## Restrictions
 


site/content/docs/user/ingress.md
```diff
@@ -21,6 +21,10 @@ Ingress exposes HTTP and HTTPS routes from outside the cluster to services withi
 > **NOTE**: You may also want to consider using [Gateway API](https://gateway-api.sigs.k8s.io/) instead of Ingress.
 > Gateway API has an [Ingress migration guide](https://gateway-api.sigs.k8s.io/guides/migrating-from-ingress/).
 
+> **WARNING**: If you are using a [rootless container runtime], ensure your host is
+> properly configured before creating the KinD cluster. Most Ingress and Gateway controllers will
```
AkihiroSuda: This paragraph does not seem meaningful; no program works unless it is properly configured.
This is not specific to rootless, nor to ingress/gateway controllers.

site/content/docs/user/quick-start.md
```diff
@@ -160,6 +160,10 @@ More usage can be discovered with `kind create cluster --help`.
 kind can auto-detect the [docker], [podman], or [nerdctl] installed and choose the available one. If you want to turn off the auto-detect, use the environment variable `KIND_EXPERIMENTAL_PROVIDER=docker`, `KIND_EXPERIMENTAL_PROVIDER=podman` or `KIND_EXPERIMENTAL_PROVIDER=nerdctl` to
 select the runtime.
 
+> **NOTE**: In some distributions (ex: Fedora), the container runtime operates in
```
AkihiroSuda: No, the default mode depends on the runtime implementation, not on the host distribution.
Regardless to the host distribution, Podman and nerdctl default to rootless, Docker defaults to rootful.

site/content/docs/user/rootless.md
```diff
@@ -9,41 +9,58 @@ menu:
...
+### cgroups v2
+
 The host needs to be running with cgroup v2.
 Make sure that the result of the `docker info` command contains `Cgroup Version: 2`.
 If it prints `Cgroup Version: 1`, try adding `GRUB_CMDLINE_LINUX="systemd.unified_cgroup_hierarchy=1"` to `/etc/default/grub` and
 running `sudo update-grub` to enable cgroup v2.
 
-Also, depending on the host configuration, the following steps might be needed:
+Your host may also need to enable [cgroup delegation](https://systemd.io/CGROUP_DELEGATION/) for daemon-based controller runtimes.
+This is not required for daemonless runtimes, such as podman. Note that this procedure may
```
AkihiroSuda: Untrue. What's the source of this misinformation? Is this from some hallucinating LLM?

site/content/docs/user/rootless.md
```diff
@@ -9,41 +9,58 @@ menu:
...
+To enable cgroup delegation, perform the folowing actions:
 
-  ```ini
-  [Service]
-  Delegate=yes
-  ```
+1. As root, create the directory `/etc/systemd/system/user@.service.d/` if it does not already exist
+
+   ```sh
+   sudo mdkir -p /etc/systemd/system/user@.service.d/
```
AkihiroSuda: typo: mdkir

AkihiroSuda: Isn't that obvious that mkdir is needed when the directory does not exist?

site/content/docs/user/rootless.md
```diff
@@ -9,41 +9,58 @@ menu:
...
-  - `/etc/systemd/system/user@.service.d/` directory needs to be created if not already present on your host
-  - If using Docker and it was already running when this step was done, a restart is needed for the changes to take
-    effect
-      {{< codeFromInline lang="bash" >}}
-      systemctl --user restart docker
-      {{< /codeFromInline >}}
+### Networking
 
-- Create `/etc/modules-load.d/iptables.conf` with the following content:
+Containers running in rootless mode are not typically loaded with host-level iptable modules.
```
AkihiroSuda: Disagree with "typically"

site/content/docs/user/rootless.md
```diff
@@ -9,41 +9,58 @@ menu:
...
+This breaks the behavior of most Ingress and Gateway controllers.
```
AkihiroSuda: Not specific to these controllers

site/content/docs/user/rootless.md
```diff
@@ -9,41 +9,58 @@ menu:
...
+
+To load the iptable modules into the KinD containers, do the following:
```
AkihiroSuda: No, kernel modules are loaded to the host kernel, not into "the KinD containers"

site/content/docs/user/rootless.md
```diff
@@ -52,14 +69,62 @@ Also, depending on the host configuration, the following steps might be needed:
...
+To increase the inotify limits, do the following:
+
+1. As root, create a `.conf` file in `/etc/systctl.d` that increases the `fs.inotify` max user settings:
+
+   ```
+   fs.inotify.max_user_watches = 524288
+   fs.inotify.max_user_instances = 512
+   ```
+
+2. Restart your system for these changes to take effect.
```
AkihiroSuda: Why not `sysctl --system` ?

site/content/docs/user/ingress.md
AkihiroSuda, line 1: s/KinD/kind/g

site/content/docs/user/rootless.md
```diff
@@ -52,14 +69,62 @@ Also, depending on the host configuration, the following steps might be needed:
   iptable_nat
   ```
 
-- If using podman, be aware that by default there is a [limit](https://docs.podman.io/en/v4.3/markdown/options/pids-limit.html#pids-limit-limit) to the number of pids that can be created. This can cause problems like nginx workers inside a container not spawning correctly.
-    - If you want to disable this limit, edit your `containers.conf` file (generally located in `/etc/containers/containers.conf`). Note that this could cause things like pid exhaustion to happen on the host machine. Alternatively, change `0` to your desired new limit:
+2. Restart your system to ensure these changes take effect.
```
AkihiroSuda: `sudo systemctl restart systemd-modules-load.service` should suffice

site/content/docs/user/rootless.md
```diff
@@ -9,41 +9,58 @@ menu:
...
 
 ## Provider requirements
+
 - Docker: 20.10 or later
 - Podman: 3.0 or later
 - nerdctl: 1.7 or later
 
 ## Host requirements
+
+### cgroups v2
```
AkihiroSuda:
```suggestion
### cgroup v2
```

For consistency with the other occurrences

site/content/docs/user/rootless.md
```diff
@@ -9,41 +9,58 @@ menu:
 Starting with kind 0.11.0, [Rootless Docker](https://docs.docker.com/go/rootless/), [Rootless Podman](https://github.com/containers/podman/blob/master/docs/tutorials/rootless_tutorial.md) and [Rootless nerdctl](https://github.com/containerd/nerdctl/blob/main/docs/rootless.md) can be used as the node provider of kind.
 
 ## Provider requirements
+
```
AkihiroSuda, line 12: ?

AkihiroSuda (changes requested):
Contains several misinformation

site/content/docs/user/rootless.md
```diff
@@ -52,14 +69,62 @@ Also, depending on the host configuration, the following steps might be needed:
...
+   ```
+   fs.inotify.max_user_watches = 524288
+   fs.inotify.max_user_instances = 512
+   ```
+
+2. Restart your system for these changes to take effect.
+
+
+### Allow Unprivileged Binding to HTTP(S) Ports
```
AkihiroSuda: Not specific to HTTP(S)

site/content/docs/user/ingress.md
adambkaplan: kind's installation instructions and quickstart assume a _rootful_ container runtime. There is no additional configuration required beyond installing said runtime and the `kind` CLI. See https://github.com/kubernetes-sigs/kind/blob/main/site/content/_index.md and https://github.com/kubernetes-sigs/kind/blob/main/site/content/docs/user/quick-start.md.

On Fedora (42) with podman (rootless), `kind create cluster` creates a control plane with no reported issues. You can even run most operators/helm charts and not have any problems. Most end users on Fedora would not think to consult the rootless guidance until problems arise. And in my experience, Ingress is where things fall apart in ways that are incredibly hard to debug.

site/content/docs/user/quick-start.md
adambkaplan: Thanks for this - I am not familiar with nerdctl's capabilities. My experience is Fedora + Podman, and wanted to hedge wrt Podman on other Linux distributions. Will correct in a follow-up commit.

site/content/docs/user/rootless.md
adambkaplan: Markdown style consistency - I like having an empty line after a heading. This doesn't appear to impact the site rendering by Hugo (see the deploy preview).

site/content/docs/user/rootless.md
adambkaplan: No, just a human trying to figure out how all this stuff works. I don't have this setting enabled, and my kind clusters (Fedora 42 + podman) seem to behave well once the other guidance in this article is adopted.

My key questions are:

1. When are systemd user services involved when running rootless kind clusters?
2. What are the consequences of not enabling cgroup delegation for user services here?

I don't think an LLM (Gemini in this case) is hallucinating when it claims "Podman itself doesn't hava a hard requirement on systemd (it can run without it)." Correct me if I'm wrong: if `kind create cluster` merely spawns containers by invoking Podman directly then there is no systemd user service involved. If true, then this recommendation a) does nothing for kind, and b) has potentially undesirable side effects for other user-scoped systemd services.

Invoking podman containers through `podman.socket` and the API service (**_a daemon_**) is a separate matter. There the user-scoped service does the right thing and enables cgroup delegation:

```sh
$ cat /usr/lib/systemd/user/podman.service 
[Unit]
Description=Podman API Service
Requires=podman.socket
After=podman.socket
Documentation=man:podman-system-service(1)
StartLimitIntervalSec=0

[Service]
Delegate=true
Type=exec
KillMode=process
Environment=LOGGING="--log-level=info"
ExecStart=/usr/bin/podman $LOGGING system service

[Install]
WantedBy=default.target
```

I would hope that rootless Docker and rootless nerdctl/containerd do similar things for their systemd services these days, in which case maybe the guidance here is obsolete?

site/content/docs/user/rootless.md
adambkaplan: Convenience for users - code blocks are easy to copy/paste.

site/content/docs/user/rootless.md
adambkaplan: My Linux experience is almost exclusively with Fedora. I am not familiar with what kernel mods are loaded by default in other distributions.

site/content/docs/user/rootless.md
adambkaplan: Better to phrase as "This breaks the behavior of many networking components, such as Ingress and Gateway controllers"?

site/content/docs/user/rootless.md
adambkaplan: My question - why isn't this procedure needed for rootful runtimes? How are these kernel modules getting loaded dynamically?

As with the other modifications here, these are system-level changes that may have undesirable side effects. I think it's helpful if end users understand the "deeper why" behind these changes. For all I know, this is a feature gap that could be fixed somewhere in the stack.

site/content/docs/user/rootless.md
adambkaplan: Will add in a follow-up commit.

site/content/docs/user/rootless.md
stmcginnis: Indeed, if we ever decided to enable a markdown linter (highly unlikely) it would complain about there not being a blank line between headers, code blocks, etc.

That said, unrelated changes to the file does make it slightly harder to review, but...

site/content/docs/user/rootless.md
AkihiroSuda: A clean installation of Ubuntu 25.04 seems to load `ip_tables` by default

site/content/docs/user/ingress.md
AkihiroSuda: > On Fedora (42) with podman (rootless), kind create cluster creates a control plane with no reported issues.

To me, `kind create cluster` with Podman failed on a clean installation of Fedora 42 (ARM).

- `/etc/subuid` and `/etc/subgid` were not configured
- `kind create cluster` fails with `ERROR: failed to create cluster: could not find a log line that matches "Reached target .*Multi-User System.*|detected cgroup v1"`

site/content/docs/user/rootless.md
AkihiroSuda: systemd seems to have begun to enable `cpu` delegation by default:
https://github.com/systemd/systemd/commit/b8df7f8629cb310beac982a4779b27eabe5362c6

This is the matter of the default configuration of systemd, not of Docker/Podman/nerdctl.

site/content/docs/user/rootless.md
AkihiroSuda: Because the modules are loaded on demand in the case of rootful

site/content/docs/user/rootless.md
AkihiroSuda: The delegation seems no longer needed to be configured manually, for systemd >= 252.
(Ubuntu >= 23.04 , Debian >= 12, RHEL >= 9, ...)

Confirmed with a clean installation of Ubuntu 25.04 (systemd 257) with Rootless Docker.

site/content/docs/user/ingress.md
AkihiroSuda: I had to run `sudo usermod -aG systemd-journal $(whoami)` and re-login to allow `podman logs` to be functional.

Didn't you need that?

site/content/docs/user/rootless.md
BenTheElder: Yeah, we will get a lot of novice users that are not familiar with any of this. A lot of Linux users will run less common configurations without much experience.

We occasionally get a bug from someone using a custom CPU scheduler that lacks cgroups .... etc.

site/content/docs/user/ingress.md
BenTheElder: We use `kind` or if avoiding codeblocks (perhaps start of a sentence) "KIND" but never "KinD"

I thought we'd added this to https://kind.sigs.k8s.io/docs/contributing/development/#documentation but we haven't, we should. I've been pretty short on time recently.

site/content/docs/user/quick-start.md
BenTheElder: This is technically a feature of the packaging, AIUI, but these runtimes are consistently packaged in this regard, at least currently.

BenTheElder: In any case, I think referring to the runtime default probably makes more sense than discussing the packaging.

site/content/docs/user/ingress.md
adambkaplan: Nope - I did not need to add myself explicitly to the systemd-journal group:

```
$ groups
<user-id> wheel video render libvirt
```

`libvirt` came from my install of OpenShift Local.

site/content/docs/user/ingress.md
adambkaplan: Done! https://github.com/kubernetes-sigs/kind/pull/4027

site/content/docs/user/rootless.md
```diff
@@ -9,57 +9,214 @@ menu:
...
-Make sure that the result of the `docker info` command contains `Cgroup Version: 2`.
-If it prints `Cgroup Version: 1`, try adding `GRUB_CMDLINE_LINUX="systemd.unified_cgroup_hierarchy=1"` to `/etc/default/grub` and
-running `sudo update-grub` to enable cgroup v2.
 
-Also, depending on the host configuration, the following steps might be needed:
+### cgroup v2
+
+The host needs to be running with cgroup v2, which is the default for many Linux disributions:
+
+- Ubuntu: 22.04 and later.
```
AkihiroSuda:
```suggestion
- Ubuntu: 21.10 and later.
```

Anyway this information seems no longer necessary, as practically every distro except EL8 has already switched to cgroup v2

site/content/docs/user/rootless.md
```diff
@@ -9,57 +9,214 @@ menu:
...
+- Fedora: 31 and later.
+- Arch: April 2021 release and later.
+
+You can verify the cgroup version used by your controller runtime with the following procedure:
+
+- `docker`: Run `docker info` and look for `Cgroup Version: 2` in the output.
+- `podman`: Run `podman info` and look for `cgroupVersion: v2` in the output.
+- `nerdctl`: Run `nerdctl info` and look for `Cgroup Version: 2` in the output.
+
+If the `info` output prints `Cgroup Version: 1` or equivalent, try the following to enable cgroup v2:
```
AkihiroSuda, line 33: Better to just upgrade the host

site/content/docs/user/rootless.md
```diff
@@ -9,57 +9,214 @@ menu:
...
+
+1. In `/etc/default/grub`, add the line `GRUB_CMDLINE_LINUX="systemd.unified_cgroup_hierarchy=1"`
+2. Run `sudo update-grub` to enable cgroup v2.
+
+Your host will also need to enable [cgroup delegation](https://systemd.io/CGROUP_DELEGATION/) for
```
AkihiroSuda:
```suggestion
Your host will also need to enable [cgroup delegation](https://systemd.io/CGROUP_DELEGATION/) of the `cpu` controller for
```

AkihiroSuda: `cpuset` is still not enabled by default.

`cpuset` is likely needed for `qosClass: Guaranteed` pods

site/content/docs/user/rootless.md
```diff
@@ -9,57 +9,214 @@ menu:
...
+user services. This is enabled by default for distributions running `systemd` version 252 and higher.
+
+To enable cgroup delegation, do the following:
```
AkihiroSuda:
```suggestion
To enable cgroup delegation for all the controllers, do the following:
```

site/content/docs/user/rootless.md
```diff
@@ -9,57 +9,214 @@ menu:
...
+   ```
+
+3. Alternatively, restart your system for these changes to take effect.
+
+
+### Allow Binding to Privileged Ports
+
+If you use the `extraPortMappings` method to provide ingress to your KIND cluster, you can allow
+the KIND node container to bind to ports 80 and 443 on the host. User containers cannot bind to
+these ports by default as they are considered privileged.
```
AkihiroSuda:
```suggestion
ports below 1024 by default as they are considered privileged.
```

site/content/docs/user/rootless.md
```diff
@@ -9,57 +9,214 @@ menu:
...
+   fs.inotify.max_user_instances = 512
+   ```
+
+2. Reload `sysctl` for these changes to take effect:
+
+   ```sh
+   sudo sysctl --system
+   ```
+
+3. Alternatively, restart your system for these changes to take effect.
```
AkihiroSuda:
```suggestion
Alternatively, restart your system for these changes to take effect.
```

site/content/docs/user/rootless.md
```diff
@@ -9,57 +9,214 @@ menu:
...
+   ```sh
+   systemctl --user restart docker
+   ```
+
+### Networking
+
+Containers running in rootless mode may not loaded with host-level iptable modules.
+This breaks the behavior of most networking components, such as Ingress and Gateway controllers.
+
+To load the iptable modules into the KIND containers, do the following:
```
AkihiroSuda:
```suggestion
To load the iptable modules, do the following:
```

adambkaplan:
Suggestions added and commits squashed. Thanks @AkihiroSuda for your thorough review!

site/content/docs/user/rootless.md
```diff
@@ -9,57 +9,214 @@ menu:
...
+   ip_tables
+   iptable_nat
+   ```
 
-  (This is not enabled by default because ["the runtime impact of
-  [delegating the "cpu" controller] is still too
-  high"](https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/ZMKLS7SHMRJLJ57NZCYPBAQ3UOYULV65/).
-  Beware that changing this configuration may affect system
-  performance.)
+4. Next, restart the `systemd-modules-load` service to make these changes effective immediately:
```
stmcginnis: `4` is used twice. Most markdown renderers handle that fine, but would be good to be consistent by either making them all `1` to be obvious that the numbering is left to the rendering step, or to have them be sequentially numbered.

site/content/docs/user/rootless.md
```diff
@@ -9,57 +9,214 @@ menu:
...
+   ```
+
+2. Reload `sysctl` for these changes to take effect:
+
+   ```sh
+   sudo sysctl --system
+   ```
+
+Alternatively, restart your system for these changes to take effect.
+
```
stmcginnis, line 173: nit: If we're going to be picky about having a blank line after a header, we should also be picky about not having multiple blank lines. ;)

site/content/docs/user/rootless.md
adambkaplan: fixed.

site/content/docs/user/rootless.md
adambkaplan: Fixed

site/content/docs/user/rootless.md
stmcginnis: Still need to address this typo.

stmcginnis:
Thanks, I think rootless is still an area with a lot of footguns, so having more details on how to do it properly will certainly help. And even if this isn't exhaustive, it gives us a good foundation to iterate on if/when others come along with other requirements for getting it to work right. So I think this is a great start. Thanks for putting it together.

/lgtm

site/content/docs/user/ingress.md
```diff
@@ -21,6 +21,10 @@ Ingress exposes HTTP and HTTPS routes from outside the cluster to services withi
 > **NOTE**: You may also want to consider using [Gateway API](https://gateway-api.sigs.k8s.io/) instead of Ingress.
 > Gateway API has an [Ingress migration guide](https://gateway-api.sigs.k8s.io/guides/migrating-from-ingress/).
 
+> **WARNING**: If you are using a [rootless container runtime], ensure your host is
+> properly configured before creating the KIND cluster. Most Ingress and Gateway controllers will
```
BenTheElder, line 27: maybe we should also cross link "properly configured" for people that know what rootless container runtime is but don't know what properly configured is.

BenTheElder:
thanks!

site/content/docs/user/ingress.md
adambkaplan: The link is to the Rootless page, which has the full guidance for host configuration. Perhaps we merge this and see if it indeed reduces the issues/questions around rootless that KIND can't solve?

BenTheElder (commented):
/approve

adambkaplan:
Needed to rebase with latest cloud provider kind updates. @BenTheElder @stmcginnis PTAL - thanks!

stmcginnis:
Rebase looks clean. This previously had approval, so to avoid the risk of hitting another merge conflict and delaying things any further:

/lgtm
/approve

PR #4023, opened 2025-10-01
CI: add AlmaLinux 8 VM for cgroup v1 testing

### LINKED ISSUE #3942: Ensure cgroups v1 CI coverage
We had been using github actions to get some easy to setup coverage of cgroup v2 VMs, and later some other things like podman and nerdctl

Today cgroup v2 is the default in Kubernetes CI, so we need some coverage for cgroup v1.

cc @aojea @stmcginnis 

We will need to re-evaluate the actions..

See also: #3915 for longer term

AkihiroSuda:
Fix #3942


.github/workflows/vm.yaml
```diff
@@ -13,20 +13,36 @@ permissions:
...
+            provider: podman
+            rootless: rootful
+          - template: fedora
+            provider: podman
+            rootless: rootless
+          # AlmaLinux 8 is used for testing kind with cgroup v1.
+          # Do not upgrade this to AlmaLinux 9 until dropping cgroup v1 support.
+          - template: almalinux-8
+            provider: docker
+            rootless: rootful
```
aojea, line 40: IIUIC this just adds one additional job, so we move from 4 to 5 that is ok

hack/ci/init-vm.sh
```diff
@@ -0,0 +1,52 @@
...
+	DNF="dnf5"
+	"$DNF" config-manager addrepo --from-repofile="${DNF_REPO}"
+else
+	# DNF 4
+	"$DNF" config-manager --add-repo="${DNF_REPO}"
+fi
+cat <<EOF | sudo tee /etc/yum.repos.d/kubernetes.repo
+[kubernetes]
+name=Kubernetes
+baseurl=https://pkgs.k8s.io/core:/stable:/v1.34/rpm/
```
aojea: is this just for kubectl?

.github/workflows/vm.yaml
AkihiroSuda: Yes, just adds one additional job

hack/ci/init-vm.sh
AkihiroSuda: Yes

aojea: should we use the `   curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"` to avoid hardcoding the version? 

it should be ok, since kubectl does not change much for this case, but is worth considering

AkihiroSuda: done

aojea:
/lgtm
/approve

Thanks

PR #4027, opened 2025-10-03
feat: KIND Style Guide

adambkaplan:
Adding a style guide for contributors. The following text branding for the project is now official:

- `kind` in backticks
- "KIND" in all uppercase.

Other prior forms, such as "KinD", are no longer allowed.

### DIFF
--- a/site/content/docs/contributing/development.md
+++ b/site/content/docs/contributing/development.md
@@ -140,6 +140,19 @@ inner markdown content. Usage: `{{</* securitygoose */>}} Notice markdown conten
 
 1. `readFile` -- is used to inline file contents. Usage: `{{%/* readFile "static/examples/ingress/contour/patch.json" */%}}`
 
+### Style Guide
+
+KIND documentation is primarily maintained by code contributors and project maintainers, and as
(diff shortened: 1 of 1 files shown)


site/content/docs/contributing/development.md
```diff
@@ -140,6 +140,19 @@ inner markdown content. Usage: `{{</* securitygoose */>}} Notice markdown conten
...
+such does not strictly obey the [Kubernetes style guide]. However, we ask that
+documentation improvements adhere to [content best practices] and avoid [documentation anti-patterns].
+
+Only use the following when referring to the project or command line:
+
+- `kind` in all lowercase with backticks.
+- "KIND" in all uppercase.
+
+Using "kind" without backticks, "Kind" with only one uppercase, or "KinD" with leading and trailing
+uppercase is not permitted.
```
stmcginnis, line 155: There are always exceptions, so "discouraged" might be a better choice than "not permitted". But not that critical.

stmcginnis:
Pointed out one minor nit, but I don't feel strongly about it.

/lgtm
/ok-to-test

PR #4031, opened 2025-10-15
feat: Add {add,delete} node commands

booxter:
On add, a new node of the chosen role is created and joins the k8s
cluster.

On delete, the node is removed from etcd cluster, drained and gracefully
removed from k8s cluster.

Note: only docker is implemented. Other backends can be added later as
needed.

Signed-off-by: Ihar Hrachyshka <ihrachyshka@nvidia.com>
Assisted-By: Claude Code; claude-sonnet-4-20250514

---

RFE: https://github.com/kubernetes-sigs/kind/issues/4034

### DIFF
--- a/hack/ci/e2e-node-ops.sh
+++ b/hack/ci/e2e-node-ops.sh
@@ -1,0 +1,567 @@
+#!/usr/bin/env bash
+# Copyright 2025 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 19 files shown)


BenTheElder:
Thanks for the PR. We generally ask that features be discussed in an issue first as per our contributing guide:

https://kind.sigs.k8s.io/docs/contributing/getting-started/#4-reaching-out

aojea:
yeah, also explain better what is the use case for adding this behavior

booxter:
Thanks folks. Let me put it in draft to make it clear it's not ready for review. Will follow up with a RFE issue later.

booxter:
This won't be allowed in kind. We will discuss in RFE ways forward; then close RFE too once we know where the code better belongs.

PR #4038, opened 2025-10-22
make setup-go.sh GOTOOLCHAIN aware

BenTheElder:
TODO: deal with gimme is broken

x-ref https://github.com/kubernetes/test-infra/pull/35756/files https://github.com/kubernetes/test-infra/issues/35755


BenTheElder:
> TODO: deal with gimme is broken

It's only broken in test-infra, which has:
> : "${GIMME_DOWNLOAD_BASE:=https://storage.googleapis.com/golang}"

in kind and kubernetes/kubernetes we have:
> : "${GIMME_DOWNLOAD_BASE:=https://dl.google.com/go}"

Technically downloads are now at https://go.dev/dl/ which redirect to https://dl.google.com/go

PR #4040, opened 2025-10-31
WIP: kubeadm v1beta4

### LINKED ISSUE #3847: Support kubeadm's v1beta4 API
Most noticeable change to me is v1beta4 uses list instead of map in `extraArgs`. The kubernetes' documentation now uses new format in all examples. (e.g. [here](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/control-plane-flags/#apiserver-flags)) 

ref: https://github.com/kubernetes/kubeadm/issues/2890

BenTheElder:
fixes #3847 

TODO: all of the mess of updating scripts and examples ...


BenTheElder:
https://github.com/kubernetes-sigs/kind/pull/4046

BenTheElder:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kind/4040/pull-kind-conformance-parallel-ipv6/2008284787059462144

> ERROR: failed to create cluster: failed to generate kubeadm config content: failed to parse yaml to patch: failed to parse type meta for "# config generated by kind\napiVersion: kubeadm.k8s.io/v1beta4\nkind: ClusterConfiguration\nmetadata:\n  name: config\nkubernetesVersion: v1.36.0-alpha.0.379+b2ac9e206fdd91\nclusterName: \"kind\"\n\ncontrolPlaneEndpoint: \"kind-control-plane:6443\"\n# on docker for mac we have to expose the api server via port forward,\n# so we need to ensure the cert is valid for localhost so we can talk\n# to the cluster after rewriting the kubeconfig to point to localhost\napiServer:\n  certSANs: [localhost, \"::1\"]\n  extraArgs:\n    - name: \"runtime-config\"\n      value: \"\"\n\ncontrollerManager:\n  extraArgs:\n\n    - name: \"enable-hostpath-provisioner\"\n      value: \"true\"\n# configure ipv6 default addresses for IPv6 clusters\n- name: \"bind-address\"\n      value: \"::\"\nscheduler:\n  extraArgs:\n\n# configure ipv6 default addresses for IPv6 clusters\n- name: \"bind-address\"\n      value: \"::\"\nnetworking:\n  podSubnet: \"fd00:10:244::/56\"\n  serviceSubnet: \"fd00:10:96::/112\"": error converting YAML to JSON: yaml: line 24: did not find expected key


Indentation issue on the `bind-address` which we only set for ipv6 clusters, issue with converting the template to the new extraArgs format. I think that's fixed now.

pkg/cluster/internal/kubeadm/config.go
```diff
@@ -502,8 +670,10 @@ func Config(data ConfigData) (config string, err error) {
 	}
 
 	// assume the latest API version, then fallback if the k8s version is too low
-	templateSource := ConfigTemplateBetaV3
-	if ver.LessThan(version.MustParseSemantic("v1.23.0")) {
+	templateSource := ConfigTemplateBetaV4
+	if ver.LessThan(version.MustParseSemantic("v1.35.0")) {
```
BenTheElder, line 674: this will need to be updated to something like v1.36.0-0

BenTheElder:
Ahhh ... that is going to be a problem ...

Because we went from a map to a list "map" for extra args, the patch runtime isn't smart enough to merge them without the user leveraging json6902 patches.

We need something like SMP but without depending on k/k directly ...

BenTheElder:
replaced by https://github.com/kubernetes-sigs/kind/pull/4173

PR #4041, opened 2025-10-31
drop GA-only special case

BenTheElder:
see https://github.com/kubernetes/test-infra/pull/35830


stmcginnis:
/lgtm
/approve
/hold

Holding just in case there's anything else to check.

stmcginnis:
Hmmm

> There are no nodes that your pod can schedule to

Issue #3946, opened 2025-06-04
containerd config patching should be version aware


BenTheElder:
We should make containerd config patching work similarly to kubeadm config patching:

- [ ] 1. Check `version` key, if present, apply only patches with matching `version` or no `version`
- [ ] 2. Start using `version` in our patches / examples / registry script / ...
- [ ] 3. Start specifying *both* version 2 and version 3 patches in our examples where relevant
- [ ] 4. Warn users in release notes that a future release will switch to version 3 (I assume eventually containerd will require this? cc @chrishenzie)
- [ ] 5. Switch to version 3 after some number of releases of warning and supplying patches for both versions

For context, containerd config versions have all the same _fields_ as far as I know, but the namespacing scheme has changed. Fields have been deprecated *across config versions* based on the *containerd version*, for e.g. #2875

chrishenzie:
> I assume eventually containerd will require this?

I think it should be required in that new features will only land in the latest config version

BenTheElder:
> I think it should be required in that new features will only land in the latest config version

ACK, that's much less of a problem than if it will be phased out. Thanks!

I still think we should migrate, but since we have split responsibility with users (e.g. https://kind.sigs.k8s.io/docs/user/local-registry/) we'll have to make it gradual and support version-aware patch targets.

Sounds like we can plan to do this long term but not urgently.

phillebaba:
I have been dealing with the small details of Containerd patching for a while across different cloud providers. It's been error prone as defaults are changes and people use different methods to merge toml. On top of that ordering matters sometimes but not all the time.

My suggestion is to just use a Containerd imports directory and then let Containerd deal with the merging of configuration. This was added as a default in Containerd https://github.com/containerd/containerd/pull/12323 to merge toml found in the directory `/etc/containerd/conf.d/`.

Overall moving to a solution where Containerd is responsible for merging configuration is good because it creates a standard behavior across different Kubernetes flavor. Happy to put in some work if you want help with this issue.

phillebaba:
To add to this as you mentioned in #4096 a way forward could be to just use extra mounts and document how users should write files to modify the existing Containerd configuration. This would remove some Containerd specific configuration and hopefully reduce maintenance burden.

PR #4042, opened 2025-10-31
add limited support for version aware containerd config patching

BenTheElder:
see: https://github.com/kubernetes-sigs/kind/issues/3946

### DIFF
--- a/pkg/internal/patch/toml.go
+++ b/pkg/internal/patch/toml.go
@@ -29,18 +29,33 @@ import (
 	"sigs.k8s.io/kind/pkg/errors"
 )
 
-// TOML patches toPatch with the patches (should be TOML merge patches) and patches6902 (should be JSON 6902 patches)
-func TOML(toPatch string, patches []string, patches6902 []string) (string, error) {
+// ContainerdTOML patches toPatch with the patches (should be TOML merge patches) and patches6902 (should be JSON 6902 patches)
+func ContainerdTOML(toPatch string, patches []string, patches6902 []string) (string, error) {
 	// convert to JSON for patching
 	j, err := tomlToJSON([]byte(toPatch))
 	if err != nil {
 		return "", err
 	}
+	version, err := containerdConfigVersion(toPatch)
+	if err != nil {
+		return "", errors.WithStack(err)
+	}
+	if version == 0 {
+		return "", errors.New("failed to detect containerd config version")
+	}
 	// apply merge patches
 	for _, patch := range patches {
 		pj, err := tomlToJSON([]byte(patch))
 		if err != nil {
-			return "", err
+			return "", errors.WithStack(err)
+		}
+		patchVersion, err := containerdConfigVersion(patch)
+		if err != nil {
+			return "", errors.WithStack(err)
+		}
+		// skip if patch sets version and version does not match
+		if patchVersion != 0 && patchVersion != version {
+			continue
 		}
 		patched, err := jsonpatch.MergePatch(j, pj)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/internal/patch/toml.go
```diff
@@ -64,6 +79,17 @@ func TOML(toPatch string, patches []string, patches6902 []string) (string, error
 	return jsonToTOMLString(j)
 }
 
+func containerdConfigVersion(configTOML string) (int, error) {
+	type version struct {
+		Version int `toml:"version,omitempty"`
+	}
+	v := version{}
+	if err := toml.Unmarshal([]byte(configTOML), &v); err != nil {
```
aojea, line 87: > The version field in the config file specifies the config’s version. If no version number is specified inside the config file then it is assumed to be a version 1 config and parsed as such. Please use version = 2 to enable version 2 config as version 1 has been deprecated.

should we default to version 1 instead of error?

BenTheElder: version 1 isn't the default in containerd as far as I know, but the default would depend on containerd version and that seems fragile.

This is the file to be patched, we are always specifying the version ourselves, so if it doesn't have the version or we can't parse it, something is wrong.

since this is in an internal package already, I didn't think it was worth designing for other use cases (hence the rename to Containerd and the assumption about the `version` key)

BenTheElder: Ah also, maybe this wasn't obvious: We won't error if there is no version, error would mean it fails to parse as toml or it has a version key with a different data type or something like that. 

(You can also confirm from the existing test cases without version specified continuing to work, in additon to the updated test cases)

We have version field as omitempty, so it defaults to 0 if not specified

pkg/internal/patch/toml.go
```diff
@@ -29,16 +29,31 @@ import (
...
-// TOML patches toPatch with the patches (should be TOML merge patches) and patches6902 (should be JSON 6902 patches)
-func TOML(toPatch string, patches []string, patches6902 []string) (string, error) {
+// ContainerdTOML patches toPatch with the patches (should be TOML merge patches) and patches6902 (should be JSON 6902 patches)
+func ContainerdTOML(toPatch string, patches []string, patches6902 []string) (string, error) {
 	// convert to JSON for patching
 	j, err := tomlToJSON([]byte(toPatch))
 	if err != nil {
 		return "", err
 	}
+	version, err := containerdConfigVersion(toPatch)
```
BenTheElder, line 39: there is an assumption that we always get a valid version from the base configuration here, we could error if this is not non-zero here, or below we could always match if it is zero similar to how we ignore match for patches with version = 0

BenTheElder: I think failing to read this one should be an error, it should only happen if we changed our base config and didn't update this properly (because version parsing changed in containerd config?) or because someone has forked kind with a custom containerd base config that doesn't set version, which I don't really care to guard against.

BenTheElder: updated to require detecting the version in the config to be patched

pohly:

Retesting to see whether the job works here (it fails in https://github.com/kubernetes-sigs/kind/pull/4046).

BenTheElder:
I think we're going to need to sort this out at some point to get ahead of containerd eventually dropping old config versions and make it possible to handle that gracefully by using versioned patches.

BenTheElder:
nudge @aojea @stmcginnis

stmcginnis:
Looks good to me, with the caveat that I haven't been closely following what has been going on with containerd v2.

/lgtm

PR #4046, opened 2025-11-03
e2e-k8s.sh drop CLUSTER_LOG_FORMAT and add versioned kubeadm config patches

BenTheElder:
x-ref https://github.com/kubernetes/test-infra/pull/35846 https://github.com/kubernetes-sigs/kind/issues/3847

### DIFF
--- a/hack/ci/e2e-k8s.sh
+++ b/hack/ci/e2e-k8s.sh
@@ -116,8 +116,6 @@ kubeadmConfigPatches:
 - |
   kind: ClusterConfiguration
   apiVersion: kubeadm.k8s.io/v1beta4
-  metadata:
-    name: config
   apiServer:
     extraArgs:
       - name: "v"
@@ -148,6 +146,10 @@ kubeadmConfigPatches:
     kubeletExtraArgs:
       - name: "v"
         value: "${KIND_CLUSTER_LOG_LEVEL}"
+      # Warning: these flags appear to be load bearing / impact performance
+      # See: https://github.com/kubernetes-sigs/kind/pull/4046
+      # Be careful when updating these.
+      # Most CI jobs should not need them, but some CI jobs might.
       - name: "container-log-max-files"
         value: "10"
       - name: "container-log-max-size"
@@ -156,8 +158,6 @@ kubeadmConfigPatches:
 - |
   kind: ClusterConfiguration
   apiVersion: kubeadm.k8s.io/v1beta3
-  metadata:
-    name: config
   apiServer:
     extraArgs:
       "v": "${KIND_CLUSTER_LOG_LEVEL}"
@@ -181,6 +181,10 @@ kubeadmConfigPatches:
   nodeRegistration:
     kubeletExtraArgs:
       "v": "${KIND_CLUSTER_LOG_LEVEL}"
+      # Warning: these flags appear to be load bearing / impact performance
+      # See: https://github.com/kubernetes-sigs/kind/pull/4046
+      # Be careful when updating these.
+      # Most CI jobs should not need them, but some CI jobs might.
       "container-log-max-files": "10"
       "container-log-max-size": "100Mi"
 EOF


BenTheElder:
Bit concerned that the non-default logging flags are loadbearing but that's not new to this PR ...

hack/ci/e2e-k8s.sh
```diff
@@ -146,29 +110,79 @@ nodes:
 featureGates: ${feature_gates}
 runtimeConfig: ${runtime_config}
 kubeadmConfigPatches:
+# v1beta4 for the future (v1.35.0+ ?)
+# https://github.com/kubernetes-sigs/kind/issues/3847
+# TODO: drop v1beta3 when we no longer need versions that use it
```
pohly, line 115: Can you add some comments about `container-log-max-files` and `container-log-max-size`? As we learned, they are required by at least one job. This is valuable information that we should preserve for the next person who needs to edit this file. Something like:

```suggestion
# TODO: drop v1beta3 when we no longer need versions that use it
#
# container-log-max-files and container-log-max-size affect overall
# performance and resource requirements of jobs. Extra care is
# necessary before changing them.
```

BenTheElder: Hmm, I suppose, but this script is frequently copied wholesale as the basis for new scripts / jobs where that may not be relevant at all. I'm currently working on cleaning up many of those scripts and some of them don't set these already. I think the real issue is this job is barely succeeding with insufficient resources and/or noisy neighbor issues.

This may or may not be true next time it is edited even for the one job, but we that is why we run an extensive set of critical jobs based on this script to any PR to this repo.

Either way, a follow up goal is to eliminate these non-default values, it appears they were desired in the context of testing structured logging but got added to all configurations.

BenTheElder: Perhaps a more detailed comment explaining that you should probably drop this from any derivative scripts, but I doubt on the other hand anyone will read it.

pohly: > it appears they were desired in the context of testing structured logging 

Now I am curious, because I don't think they were added for that.

The settings got added in https://github.com/kubernetes-sigs/kind/pull/3774 because of https://github.com/kubernetes-sigs/kind/issues/3772, on behalf of @aojea. ~~Then later it was updated in https://github.com/kubernetes-sigs/kind/pull/3774/files.~~ Different commit, same PR.

pohly: I've not been involved in any of this, so I lack context. You had some concerns about not testing defaults.

pohly: My takeaway from this is (once again) that explaining changes in GitHub PRs is not sufficient. It must be documented in the code for the next person who has to make changes to the file. It might even be the same person.

BenTheElder: I'll have more time to dig and update later

BenTheElder: I added a warning

hack/ci/e2e-k8s.sh
```diff
@@ -146,29 +110,87 @@ nodes:
...
+- |
+  kind: ClusterConfiguration
+  apiVersion: kubeadm.k8s.io/v1beta4
+  metadata:
+    name: config
```
neolit123: note that running strict validation will report this. there is no objectmeta on the kubeadm kinds.

neolit123 (commented):
extra args diff lgtm
/lgtm
/hold

hack/ci/e2e-k8s.sh
pohly: Warning looks good, thanks!

Now only https://github.com/kubernetes-sigs/kind/pull/4046/files#r2501870491 is open as far as I can tell.

hack/ci/e2e-k8s.sh
BenTheElder: IIRC this is optionally used to match the object in the patch runtime and stripped out before reaching kubeadm to avoid this

https://github.com/kubernetes-sigs/kind/blob/b2b72662ea4c0062a510e1ec5cf598352b739c40/pkg/cluster/internal/create/actions/config/config.go#L244-L255

BenTheElder: IIRC kustomize required these fields to match patches, so we have the code to post-process dropping them, but kind's own patch runtime doesn't require this IIRC (we should keep the post-process bit for compatibility)

BenTheElder: Yeah, our internal simplified patch runtime only matches on kind/version, it doesn't support having multiple named objects anyhow, so we can just drop this here but also all the examples ... I will leave the examples for another PR

BenTheElder: (will also leave strict validation to another PR)

hack/ci/e2e-k8s.sh
BenTheElder: I must be missing something. What's open?

pohly: I don't remember what this was and the link is dead now (thanks GitHub!). Let's assume it's resolved.

BenTheElder:
I think this is good to go.

pohly:
/lgtm

Based on https://github.com/kubernetes-sigs/kind/pull/4046#pullrequestreview-3431909912 and the changes since then.

pacoxu (approved):
/lgtm

BenTheElder:
nerdctl job is a flake uploading the logs afterwards and there doesn't seem to be an option to rerun it for some reason.

BenTheElder:
[merged because actions were stuck and blocking merge, but not due to any issues here]

hack/ci/e2e-k8s.sh
aojea: > > it appears they were desired in the context of testing structured logging
> 
> Now I am curious, because I don't think they were added for that.
> 
> The settings got added in #3774 because of #3772, on behalf of @aojea. ~Then later it was updated in https://github.com/kubernetes-sigs/kind/pull/3774/files.~ Different commit, same PR.

not having logs on CI jobs made impossible to debug anything

PR #4043, opened 2025-11-01
update ingress docs

rayowang:
fixes: #4039

PR #4052, opened 2025-11-21
fix: docs: update nginx controller image hash

JanGutter:
* In commit 918353f0b57a58ce4747266e1958d9f9c1c6192b the nginx controller image was updated to a new version, but the hash was just copied from `v1.12.0-beta.0`

* The new hash can be verified by doing: `skopeo inspect docker://registry.k8s.io/ingress-nginx/controller:v1.12.1 --override-os linux -f '{{.Digest}}'`

### DIFF
--- a/site/static/examples/ingress/deploy-ingress-nginx.yaml
+++ b/site/static/examples/ingress/deploy-ingress-nginx.yaml
@@ -442,7 +442,7 @@ spec:
               fieldPath: metadata.namespace
         - name: LD_PRELOAD
           value: /usr/local/lib/libmimalloc.so
-        image: registry.k8s.io/ingress-nginx/controller:v1.12.1@sha256:9724476b928967173d501040631b23ba07f47073999e80e34b120e8db5f234d5
+        image: registry.k8s.io/ingress-nginx/controller:v1.12.1@sha256:d2fbc4ec70d8aa2050dd91a91506e998765e86c96f32cffb56c503c9c34eed5b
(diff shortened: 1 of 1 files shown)


JanGutter:
Note to reviewers: I've verified that the other hashes are correct in that file, it seems to be just that one that's broken.

aojea:
we try to remove it now that it was announced it will be retired

JanGutter:
Does someone have a WIP doc available for gateway API?

One other reason to fix this (even though it will be taken away) is because the hash is referring to an image with vulnerabilities: https://securitylabs.datadoghq.com/articles/ingress-nightmare-vulnerabilities-overview-and-remediation/

aojea:
https://github.com/kubernetes-sigs/kind/pull/4043

aojea:
but I'm happy to approve another PR that just remove the ingress-nginx part and update to use cloud-provider-kind

JanGutter:
My only bit of sadness is that cloud-provider-kind requires elevated privilege, (and simple ingress was completely sufficient for my usecase). Closing this PR since it's unlikely to be merged.

aojea:
> My only bit of sadness is that cloud-provider-kind requires elevated privilege, (and simple ingress was completely sufficient for my usecase). Closing this PR since it's unlikely to be merged.

and you still can use it, is just that everything we publish in our docs turns into things we need to support in the project ... and we can not support a project that is going to be archived

PR #4053, opened 2025-11-23
replace `load docker-image` with `load container-image` to support multiple providers

Blaimi:
This POC abstracts `docker save` to the providers so that images from podman and nerdctl can also be used with the `load container-image` command can be used instead of restrict it to docker with `load docker-image`.

TBD
* With this implementation, you can only use the same provider for loading images as is used for running kind. I.e. it is not possible to load an image from podman and have the cluster running in docker. Changing this would either need to have multiple commands (like `load {docker,podman,nerdctl}-image` and/or have additional configuration options like `KIND_EXPERIMENTAL_PROVIDER`.

TODO
* ajust unit-tests
* run actual tests (didn't even test it on my machine)
* better abstraction of the backwards-compatibility code for `load docker-image` because it is actually copy&paste with very small ajustments.

Fix: #2038
Relates: https://github.com/GoogleContainerTools/skaffold/issues/9901

### DIFF
--- a/pkg/cmd/kind/load/docker-image/docker-image.go
+++ b/pkg/cmd/kind/load/docker-image/docker-image.go
@@ -1,0 +1,266 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package load
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"sigs.k8s.io/kind/pkg/cluster"
+	"sigs.k8s.io/kind/pkg/cluster/nodes"
+	"sigs.k8s.io/kind/pkg/cluster/nodeutils"
+	"sigs.k8s.io/kind/pkg/cmd"
+	"sigs.k8s.io/kind/pkg/errors"
+	"sigs.k8s.io/kind/pkg/fs"
+	"sigs.k8s.io/kind/pkg/log"
+
+	"sigs.k8s.io/kind/pkg/internal/cli"
+	"sigs.k8s.io/kind/pkg/internal/runtime"
+)
+
+type (
+	imageTagFetcher func(nodes.Node, string) (map[string]bool, error)
+)
+
+type flagpole struct {
+	Name  string
+	Nodes []string
+}
+
+// NewCommand returns a new cobra.Command for loading an image into a cluster
+func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
+	flags := &flagpole{}
+	cmd := &cobra.Command{
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) < 1 {
+				return errors.New("a list of image names is required")
+			}
+			return nil
+		},
+		Use:        "docker-image <IMAGE> [IMAGE...]",
+		Short:      "Loads docker images from host into nodes",
+		Long:       "Loads docker images from host into all or specified nodes by name",
+		Deprecated: "Please use 'kind load container-image' instead of 'kind load docker-image', 'docker-image' will be removed in future releases and is restricted to load images from docker.",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			cli.OverrideDefaultName(cmd.Flags())
+			return runE(logger, flags, args)
+		},
+	}
+	cmd.Flags().StringVarP(
+		&flags.Name,
+		"name",
+		"n",
+		cluster.DefaultName,
+		"the cluster context name",
+	)
+	cmd.Flags().StringSliceVar(
+		&flags.Nodes,
+		"nodes",
+		nil,
+		"comma separated list of nodes to load images into",
+	)
+	return cmd
+}
+
+func runE(logger log.Logger, flags *flagpole, args []string) error {
+	provider := cluster.NewProvider(
+		cluster.ProviderWithLogger(logger),
+		runtime.GetDefault(logger),
+	)
+
+	saveProvider := cluster.NewProvider(
+		cluster.ProviderWithDocker(),
+		runtime.GetDefault(logger),
+	)
+
+	// Check that the image exists locally and gets its ID, if not return error
+	imageNames := removeDuplicates(args)
+	var imageIDs []string
+	for _, imageName := range imageNames {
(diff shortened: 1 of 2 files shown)


pkg/cmd/kind/load/container-image/container-image.go
```diff
@@ -0,0 +1,261 @@
...
+// NewCommand returns a new cobra.Command for loading an image into a cluster
+func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
+	flags := &flagpole{}
+	cmd := &cobra.Command{
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) < 1 {
+				return errors.New("a list of image names is required")
+			}
+			return nil
+		},
```
stmcginnis, line 58: Would be great to set `Alias` to include just `image`.

pkg/cmd/kind/load/docker-image/docker-image.go
```diff
@@ -80,18 +79,22 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 	)
 	return cmd
 }
-
```
stmcginnis, line 83: This line shouldn't be removed.

pkg/cmd/kind/load/load.go
```diff
@@ -44,7 +45,8 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
 		},
 	}
 	// add subcommands
-	cmd.AddCommand(dockerimage.NewCommand(logger, streams))
+	cmd.AddCommand(containerimage.NewCommand(logger, streams))
 	cmd.AddCommand(imagearchive.NewCommand(logger, streams))
+	cmd.AddCommand(dockerimage.NewCommand(logger, streams))
```
stmcginnis, line 49: Is there a reason for moving this line?

pkg/cmd/kind/load/docker-image/docker-image.go
```diff
@@ -57,9 +55,10 @@ func NewCommand(logger log.Logger, streams cmd.IOStreams) *cobra.Command {
...
 			return nil
 		},
-		Use:   "docker-image <IMAGE> [IMAGE...]",
-		Short: "Loads docker images from host into nodes",
-		Long:  "Loads docker images from host into all or specified nodes by name",
+		Use:        "docker-image <IMAGE> [IMAGE...]",
+		Short:      "Loads docker images from host into nodes",
+		Long:       "Loads docker images from host into all or specified nodes by name",
+		Deprecated: "Please use 'kind load container-image' instead of 'kind load docker-image', 'docker-image' will be removed in future releases and is restricted to the docker-provider.",
 		RunE: func(cmd *cobra.Command, args []string) error {
```
stmcginnis, line 62: I haven't looked closely, but can we have this command just call the new code?

Blaimi: No. This would break existing installations, where the image is built/available via docker but the cluster runs on podman/nerdctl. I'm a huge fan of not breaking stuff because of new features ;-)

pkg/cluster/internal/providers/docker/provider.go
```diff
@@ -276,6 +276,28 @@ func (p *provider) Info() (*providers.ProviderInfo, error) {
 	return p.info, err
 }
 
+// ContainerSave saves images to dest, as in `docker save`
+func (p *provider) ContainerSave(images []string, dest string) error {
```
BenTheElder, line 280: Saving images from storage isn't really a property of implementing nodes, and this is already a pretty complicated object. I think we should probably split this out.

pkg/cmd/kind/load/container-image/container-image.go
```diff
@@ -0,0 +1,262 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
BenTheElder, line 2: new files should drop the year (previously they should've had the year of creation, so 2025 here)

dropping the year is ... new https://github.com/kubernetes/steering/issues/299

pkg/cluster/internal/providers/docker/provider.go
BenTheElder: We also lose the ability to import across providers cleanly, which while a bit weird we probably already have users doing.

pkg/cmd/kind/load/container-image/container-image.go
```diff
@@ -0,0 +1,262 @@
...
+
+// checkIfImageExists makes sure we only perform the reverse lookup of the ImageID to tag map
+func checkIfImageReTagRequired(node nodes.Node, imageID, imageName string, tagFetcher imageTagFetcher) (exists, reTagRequired bool, sanitizedImage string) {
+	tags, err := tagFetcher(node, imageID)
+	if len(tags) == 0 || err != nil {
+		exists = false
+		return
+	}
+	exists = true
+	sanitizedImage = sanitizeImage(imageName)
```
BenTheElder, line 229: this may be inaccurate on podman where the user may have configured a different default image host because :reasons:

for our own node images, forcing the default host we actually use makes sense, for user's images ... maybe not

BenTheElder (commented):
couple quick thoughts

I think this overall makes sense, but it's also not quite the UX discussed in the issue (can't override source, only supports locally stored images ...).

pkg/cluster/internal/providers/docker/provider.go
Blaimi: > Saving images from storage isn't really a property of implementing nodes, and this is already a pretty complicated object. I think we should probably split this out.

Any idea for locations? I would suggest `pkg/cmd/kind/load/providers/*/provider.go`

> We also lose the ability to import across providers cleanly, which while a bit weird we probably already have users doing.

Can you explain that? `load docker-image` will be backwards compatible.

Blaimi:
> renaming the existing command, modifying it, and then back-adding a supposedly compatible version seems ... far more confusing than just introducing a new command to begin with and leaving the existing one ~untouched

That's for the long run of the git history. Once the then deprecated `load docker-image` is removed, the history of the files has the correct origins. And I'm not happy with this implementation because it contains a lot of duplicate code. If this gets resolved with clean abstractions, it might make even more sense to "rename&add for bc" instead of "add new&deprecate"

pkg/cmd/kind/load/container-image/container-image.go
Blaimi: Any suggestion? Just not sanitizing will lead to an error then—or make the command interactive which might lead to new problems …

BenTheElder:
> That's for the long run of the git history. 

Moving files tends to obscure the history.

> Once the then deprecated load docker-image is removed, the history of the files has the correct origins.

I don't see us removing `load docker-image` for the forseeable future, it's baked into _way_ too many project's scripts that are working fine.

> And I'm not happy with this implementation because it contains a lot of duplicate code. If this gets resolved with clean abstractions, it might make even more sense to "rename&add for bc" instead of "add new&deprecate"

We prescribe to WET - Write Everything Thrice. We don't have a good abstraction yet when we've only needed to do something ~twice. It's more go-like to avoid abstractions when you don't yet have a really complex and heavily repeated pattern versus favoring local straightforward code.

PR #4056, opened 2025-11-25
update node image deps

BenTheElder:
there were a number of runc CVEs


images/base/Dockerfile
```diff
@@ -140,7 +140,7 @@ RUN git clone --filter=tree:0 "${CONTAINERD_CLONE_URL}" /containerd \
 # stage for building runc
 FROM go-build AS build-runc
 ARG TARGETARCH GO_VERSION
-ARG RUNC_VERSION="v1.3.0"
+ARG RUNC_VERSION="v1.3.3"
```
AkihiroSuda, line 143: Now v1.4.0

BenTheElder: We've been keeping this aligned with what containerd is testing at the containerd version we're using, generally.

(So I'm looking at https://github.com/containerd/containerd/blob/release/2.2/script/setup/runc-version)

BenTheElder:
I think we should let this build and iterate from here

BenTheElder:
The build failed because of test-infra changes to KRTE :/

https://github.com/kubernetes/test-infra/pull/35799

BenTheElder:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/post-kind-push-base-image/1998850601395425280

BenTheElder:
I think we should switch images, working on that. Separately, we're hitting a lot of issues since we started aggressively updating the CI images, so also sending a change to revert that for now.

BenTheElder:
containerd hitting https://github.com/kubernetes/kubernetes/issues/135214

images/base/Dockerfile
BenTheElder: Unfortunately we had to roll back even further to 1.2.9, due to https://github.com/kubernetes/kubernetes/issues/135214

We will roll forward once the tests are updated to handle the weighting change.

Issue #4044, opened 2025-11-02
Add environment variable information to CLI help and/or introduce kind env command


jaehanbyun:
**What would you like to be added**:
Add environment variable information to Kind CLI help output and/or introduce a new `kind env` command to display currently set environment variables (e.g., `KIND_EXPERIMENTAL_PROVIDER`, `KUBECONFIG`) and list available configurable environment variables.

**Why is this needed**:
When attempting to use Kind with Podman instead of Docker, users may encounter errors like:

```
ERROR: failed to list clusters: command "docker ps -a --filter label=io.x-k8s.kind.cluster --format '{{.Label "io.x-k8s.kind.cluster"}}'" failed with error: exit status 1
Command Output: Cannot connect to the Docker daemon at unix:///Users/byeonjaehan/.docker/run/docker.sock. Is the docker daemon running?
```

However, when checking `kind --help`, there is no information about the `KIND_EXPERIMENTAL_PROVIDER` environment variable or other configurable environment variables. This makes it difficult for users to discover how to configure Kind to use alternative container runtimes or other settings.

**Proposed Solution**:

1. Add a brief section to `kind --help` output that mentions key environment variables like `KIND_EXPERIMENTAL_PROVIDER`, `KIND_CLUSTER_NAME`, and `KUBECONFIG`.

2. Introduce a new `kind env` command that:
   - Displays currently set Kind-related environment variables and their values
   - Lists all available Kind environment variables with brief descriptions
   - Optionally allows setting/unsetting environment variables (similar to `go env`)

jaehanbyun:
If this proposal is acceptable, I'd be happy to work on implementing it.

2003Aditya:
Hey! 👋 I went through the issue and the proposal makes complete sense.

Currently, Kind doesn’t surface any information about configurable environment variables like `KIND_EXPERIMENTAL_PROVIDER`, `KIND_CLUSTER_NAME`, or `KUBECONFIG` in the CLI help, which can be confusing for users who run into runtime or provider-related errors (especially when using Podman instead of Docker).

Adding a short “Environment Variables” section in the `kind --help` output or introducing a new kind env command would make these options much more discoverable.
The kind env command could behave similarly to go env, showing currently set Kind-related variables along with their descriptions and default values.

I’d be happy to help draft or implement this feature if needed!

BenTheElder:
[...]

PR #4060, opened 2025-11-28
cli: add read-only kind env command

aliheydarabadii:
Adds a read-only `kind env` subcommand to surface key environment variables and their current values.
No behavior changes.

Refs #4044


BenTheElder:
Sorry but I'm -1 on this one. To be clear we have not agreed to this part of https://github.com/kubernetes-sigs/kind/issues/4044

PR #4064, opened 2025-12-08
Adopt Envoy in place of HAProxy for load balancing

shwetha-s-poojary:
Switching load balancing from HAProxy to Envoy.
Based on the discussion in the referenced Slack thread: [https://kubernetes.slack.com/archives/CEKK1KTN2/p1759769331878379](https://kubernetes.slack.com/archives/CEKK1KTN2/p1759769331878379?thread_ts=1759743363.654749&cid=CEKK1KTN2).

### DIFF
--- a/pkg/cluster/constants/constants.go
+++ b/pkg/cluster/constants/constants.go
@@ -20,7 +20,8 @@ package constants
 // DefaultClusterName is the default cluster Context name
 const DefaultClusterName = "kind"
 
-// start Envoy with dynamic configuration by using files that implement the xDS protocol.
+// DynamicFilesystemConfigTemplate holds the Envoy bootstrap configuration template
+// for file-based dynamic xDS (CDS and LDS).
 // https://www.envoyproxy.io/docs/envoy/latest/start/quick-start/configuration-dynamic-filesystem
 const DynamicFilesystemConfigTemplate = `
 node:
--- a/pkg/cluster/internal/loadbalancer/config.go
+++ b/pkg/cluster/internal/loadbalancer/config.go
@@ -32,7 +32,7 @@ type ConfigData struct {
 	IPv6             bool
 }
 
-// proxyLDSConfigTemplate is the loadbalancer config template for listeners
+// ProxyLDSConfigTemplate is the loadbalancer config template for listeners
 const ProxyLDSConfigTemplate = `
 resources:
 - "@type": type.googleapis.com/envoy.config.listener.v3.Listener
@@ -50,7 +50,7 @@ resources:
         cluster: kube_apiservers
 `
 
-// proxyCDSConfigTemplate is the loadbalancer config template for clusters
+// ProxyCDSConfigTemplate is the loadbalancer config template for clusters
 // https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/health_check.proto#envoy-v3-api-msg-config-core-v3-healthcheck-httphealthcheck
 const ProxyCDSConfigTemplate = `
 resources:


pkg/cluster/internal/create/actions/loadbalancer/loadbalancer.go
```diff
@@ -90,9 +91,9 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 		return errors.Wrap(err, "failed to copy loadbalancer config to node")
 	}
 
-	// reload the config. haproxy will reload on SIGHUP
-	if err := loadBalancerNode.Command("kill", "-s", "HUP", "1").Run(); err != nil {
-		return errors.Wrap(err, "failed to reload loadbalancer")
+	// restart loadbalancer to apply static configuration changes
+	if err := exec.Command("docker", "restart", "kind-external-load-balancer").Run(); err != nil {
```
stmcginnis: Others, please correct me if I am wrong, but I don't believe we want to explicitly call `docker` commands here. User may be using `podman` or `nerdctl` and not have `docker` present on their system. Is there an equivalent to sending HUP to envoy?

aojea: you do not need to restart , see how is done in cloud provider kind https://github.com/kubernetes-sigs/cloud-provider-kind/blob/4820e03979d5ff7fbcd8645794db2ff39a608b3f/pkg/loadbalancer/proxy.go#L276-L313

BenTheElder: Yeah, we should not exec directly to docker in particular here, but we also don't need to.

shwetha-s-poojary: @stmcginnis   envoy does support hot restarts, but for this scenario a simple container restart seems to be the most reliable and straightforward option. I’ll look into detecting the container runtime so we can use the appropriate tool instead of assuming docker.
@aojea  @BenTheElder  from my understanding, only dynamic (xDS) configuration avoids restarts; static configuration still requires one. For the kind use case, static config seems sufficient and avoids the complexity of a full dynamic config setup. The cloud-provider-kind example you shared above uses dynamic config, but I believe the earlier version relied on static config and required restarts (ref: [PR link](https://github.com/kubernetes-sigs/cloud-provider-kind/pull/61/files#diff-46c7fbbe166e4464ac2f78e38842da8be1e3b5280364f334c2035fa9a09ba49d)).
Please correct me if I’m mistaken anywhere — I appreciate the guidance.

aojea: @shwetha-s-poojary if we know all the configuration before hand you just need to pass the final config during the entrypoint and  you can just use existing libraries to create the container, see

https://github.com/kubernetes-sigs/cloud-provider-kind/blob/dc1a6e4c3b21716be9f87476d5cdb78389f05537/pkg/loadbalancer/server.go#L247-L263

the example I provided in my previous comment is to update the configuration dynamically without having to restart the container, you can write files on the filesystem of the container and envoy will pick the config and autoconfigure itself, no need to restarts

BenTheElder: >  I’ll look into detecting the container runtime so we can use the appropriate tool instead of assuming docker.

I'd really rather not expand the node container runtime abstraction just for this purpose, haproxy has been working ~fine.

But see antonio's comment above.

aojea: @shwetha-s-poojary so you need to follow the cloud-provider-kind implementation, a straw man approach will be

1. bootstrap the container with an initial config that allow to use dynamic configuration based on files

- example config
https://github.com/kubernetes-sigs/cloud-provider-kind/blob/4820e03979d5ff7fbcd8645794db2ff39a608b3f/pkg/loadbalancer/proxy.go#L33-L53

- override the config during container start  https://github.com/kubernetes-sigs/cloud-provider-kind/blob/37c196059f9c31de6f0734e22b608923f00eb195/pkg/loadbalancer/server.go#L247-L257

unfortunately I think this is per provider https://github.com/search?q=repo%3Akubernetes-sigs%2Fkind%20runArgsForLoadBalancer&type=code

2. at runtime, add the expected configuration

create the config as in cloudprovider kind

https://github.com/kubernetes-sigs/cloud-provider-kind/blob/4820e03979d5ff7fbcd8645794db2ff39a608b3f/pkg/loadbalancer/proxy.go#L82-L273

and write it to the corresponding node in the expected path (configured in step 1)

https://github.com/kubernetes-sigs/kind/blob/02382a0189a912c19c3239a4f6743f90d1856810/pkg/cluster/internal/create/actions/loadbalancer/loadbalancer.go#L77-L91

shwetha-s-poojary: @aojea @BenTheElder @stmcginnis Thanks for the guidance! Looks like switching to dynamic config is the way to go.I'll get on it.

go.mod
```diff
@@ -20,17 +20,30 @@ require (
 	github.com/spf13/cobra v1.8.0
 	github.com/spf13/pflag v1.0.5
 	go.yaml.in/yaml/v3 v3.0.4
+	k8s.io/api v0.23.17
+	k8s.io/utils v0.0.0-20211116205334-6203023598ed
```
shwetha-s-poojary: Pinned k8s.io/api v0.23.17 and k8s.io/utils v0.0.0-20211116205334-6203023598ed to stay compatible with Go 1.17, as newer versions require Go 1.25+.

aojea: we do not need this, do not worry, I will comment on the part that removes this dependency

pkg/cluster/internal/loadbalancer/config.go
```diff
@@ -81,5 +168,69 @@ func Config(data *ConfigData) (config string, err error) {
 	if err != nil {
 		return "", errors.Wrap(err, "error executing config template")
 	}
+
 	return buff.String(), nil
 }
+
+func GenerateConfig(service *v1.Service, nodes []*v1.Node) *proxyConfigData {
```
aojea: this is a config for a Service loadbalancer, we are not loadbalancing Services here

pkg/cluster/internal/loadbalancer/config.go
```diff
@@ -18,60 +18,147 @@ package loadbalancer
...
-  timeout connect 5000
-  timeout client 50000
-  timeout server 50000
-  # allow to boot despite dns don't resolve backends
-  default-server init-addr none
-
-frontend control-plane
-  bind *:{{ .ControlPlanePort }}
-  {{ if .IPv6 -}}
-  bind :::{{ .ControlPlanePort }};
```
aojea, line 59: you'' benefit of introducing this in a llm like gemini , chatgpt ,... and asking for the same template for envoy  ... the config from envoy you are getting needs to be adapted for this functionality

go.mod
BenTheElder: We definitely want to avoid these dependencies. I assume the comment is https://github.com/kubernetes-sigs/kind/pull/4064/changes#r2842905017 ?

Thanks

dims: 💯 let's please avoid circular dependency with k8s

pkg/cluster/constants/constants.go
```diff
@@ -20,6 +20,38 @@ package constants
 // DefaultClusterName is the default cluster Context name
 const DefaultClusterName = "kind"
 
+// start Envoy with dynamic configuration by using files that implement the xDS protocol.
+// https://www.envoyproxy.io/docs/envoy/latest/start/quick-start/configuration-dynamic-filesystem
+const DynamicFilesystemConfig = `node:
+  cluster: cloud-provider-kind
+  id: cloud-provider-kind-id
```
aojea: this will be the cluster name

pkg/cluster/internal/create/actions/loadbalancer/loadbalancer.go
```diff
@@ -70,29 +70,44 @@ func (a *Action) Execute(ctx *actions.ActionContext) error {
 	if err != nil {
 		return err
 	}
+
 	for _, n := range controlPlaneNodes {
-		backendServers[n.String()] = fmt.Sprintf("%s:%d", n.String(), common.APIServerInternalPort)
+		nodeIP, _, err := n.IP()
```
aojea: this will return the ipv4 and ipv6 addresses, we used dns with haproxy to avoid to have to figure this out ... IIUIC you can still use dns with envoy

shwetha-s-poojary: Right - will do that

aojea:
```yaml
kind: Cluster
nodes:
- role: control-plane
- role: control-plane
- role: control-plane
- role: worker
```

Use kind with the above config
```
bin/kind create cluster --name ha --config kind.config
```

It fails to start with

> [2026-02-25 11:30:06.524][12][warning][config] [source/extensions/config_subscription/filesystem/filesystem_subscription_impl.cc:103] Filesystem config update failure: in /home/envoy/lds.yaml, Unable to convert YAML as JSON:

In this case since we do not have dynamic joining of nodes, you can also avoid the cds and lds part and generate an static config like in haproxy , you can test it following above instructions

shwetha-s-poojary:
> ```yaml
> kind: Cluster
> apiVersion: kind.x-k8s.io/v1alpha4
> nodes:
> - role: control-plane
> - role: control-plane
> - role: control-plane
> - role: worker
> ```
> 
> Use kind with the above config
> 
> ```
> bin/kind create cluster --name ha --config kind.config
> ```
> 
> It fails to start with
> 
> > [2026-02-25 11:30:06.524][12][warning][config] [source/extensions/config_subscription/filesystem/filesystem_subscription_impl.cc:103] Filesystem config update failure: in /home/envoy/lds.yaml, Unable to convert YAML as JSON:
> 
> In this case since we do not have dynamic joining of nodes, you can also avoid the cds and lds part and generate an static config like in haproxy , you can test it following above instructions

@aojea do you mean static config like we previously had https://github.com/kubernetes-sigs/kind/commit/7b7d80bae6941b0790a6ec33ba971ed7d337611c ?

aojea:
> @aojea do you mean static config like we previously had [7b7d80b](https://github.com/kubernetes-sigs/kind/commit/7b7d80bae6941b0790a6ec33ba971ed7d337611c) ?

yeah, but now I realize. I think I was the one suggesting to move into this direction so I do not want to keep distracting you 😅  ... since it does not really matter for the final result, envoy will merge that config back ... the problem now is that one of the configs is not parsing correctly, maybe some template problem

shwetha-s-poojary:
> > @aojea do you mean static config like we previously had [7b7d80b](https://github.com/kubernetes-sigs/kind/commit/7b7d80bae6941b0790a6ec33ba971ed7d337611c) ?
> 
> yeah, but now I realize. I think I was the one suggesting to move into this direction so I do not want to keep distracting you 😅 ... since it does not really matter for the final result, envoy will merge that config back ... the problem now is that one of the configs is not parsing correctly, maybe some template problem

I'm not able to reproduce that error.Is it a warning or the cluster creation itself is failing?

```
$ bin/kind create cluster --name ha --config kind.config
Creating cluster "ha" ...
 ✓ Ensuring node image (gcr.io/k8s-staging-kind/node:v1.35.0) 🖼
 ✓ Preparing nodes 📦 📦 📦 📦
 ✓ Configuring the external load balancer ⚖️
 ✓ Writing configuration 📜
 ✓ Starting control-plane 🕹️
 ✓ Installing CNI 🔌
 ✓ Installing StorageClass 💾
 ✓ Joining more control-plane nodes 🎮
 ✓ Joining worker nodes 🚜
Set kubectl context to "kind-ha"

kubectl cluster-info --context kind-ha

Thanks for using kind! 😊
$ cat kind.config
kind: Cluster
nodes:
- role: control-plane
- role: control-plane
- role: control-plane
- role: worker

```

aojea:
do `docker logs ha-external-load-balancer`

pkg/cluster/internal/providers/docker/provision.go
```diff
@@ -282,7 +282,25 @@ func runArgsForLoadBalancer(cfg *config.Cluster, name string, args []string) ([]
...
+		constants.ProxyConfigPathCDS,
+		constants.ProxyConfigPathLDS,
+	)
+
+	// Create dynamic Envoy config files and start Envoy with retry,
+	// since it has an initialization phase before forwarding traffic.
+	cmd := []string{"bash", "-c",
+		fmt.Sprintf(`mkdir -p %s && echo -en '%s' > %s && touch %s && touch %s && while true; do envoy -c %s && break; sleep 1; done`, constants.ProxyConfigDir,
+			envoyConfig, constants.ProxyConfigPath, constants.ProxyConfigPathCDS, constants.ProxyConfigPathLDS, constants.ProxyConfigPath)}
+	args = append(args, cmd...)
```
shwetha-s-poojary, line 310: @aojea Fixed "Unable to convert YAML" startup errors by initializing cds.yaml and lds.yaml with resources: [] instead of empty files during container creation.

```
emptyConfig := "resources: []"
cmd := []string{"bash", "-c",
fmt.Sprintf(`mkdir -p %s && echo -en '%s' > %s && echo -en '%s' > %s && echo -en '%s' > %s && while true; do envoy -c %s && break; sleep 1; done`,
	constants.ProxyConfigDir,
	envoyConfig, constants.ProxyConfigPath,
	emptyConfig, constants.ProxyConfigPathCDS, // Initialize CDS
	emptyConfig, constants.ProxyConfigPathLDS, // Initialize LDS
	constants.ProxyConfigPath)}
```

Do you want to go ahead with this or fall back to static approach?

shwetha-s-poojary: @aojea just checking to see if you’ve had a chance to TAL.

aojea:
/close

thanks, this work was included in https://github.com/kubernetes-sigs/kind/pull/4127

shwetha-s-poojary:
@kishen-v appreciate the behind-the-scenes help on this one 😄

PR #4065, opened 2025-12-09
Bump go version to 1.25.5

stmcginnis:
Raised to match [current k/k version](https://github.com/kubernetes/kubernetes/blob/b4980b8d6e65602818ad27eab6c9c12398569420/.go-version).


BenTheElder:
> Raised to match [current k/k version](https://github.com/kubernetes/kubernetes/blob/b4980b8d6e65602818ad27eab6c9c12398569420/.go-version).

We actually do not need to match k/k at all, it will use it's own version.

We should probably land https://github.com/kubernetes-sigs/kind/pull/4056 (.go-version currently is used to build many things)

BenTheElder:
we should still bump to 1.25.5, but no rush to rebase

PR #4070, opened 2025-12-13
cli: align --name flag help and default

aliheydarabadii:
Standardize the --name/-n help text across commands and use cluster.DefaultName for create cluster so help output is consistent. Refs #4044

### DIFF
--- a/pkg/cmd/kind/create/cluster/createcluster.go
+++ b/pkg/cmd/kind/create/cluster/createcluster.go
@@ -19,6 +19,7 @@ package cluster
 
 import (
 	"io"
+	"os"
 	"time"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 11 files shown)


pkg/cmd/kind/create/cluster/createcluster.go
stmcginnis: I think with this change there wouldn't be any need for the `OverrideDefaultName` function anymore.

aliheydarabadii: I think OverrideDefaultName is still necessary for KIND_CLUSTER_NAME support. It only runs when --name isn’t explicitly set, and the default change is just for consistent help output.

aliheydarabadii:
Hi @BenTheElder (or @aojea) 👋
Would you mind taking a quick look and, if it seems reasonable, dropping a /ok-to-test so CI can run?
Happy to adjust anything needed. Thanks!

pkg/cmd/kind/create/cluster/createcluster.go
stmcginnis: That's misleading and possibly confusing now though. If there is a default name that is given in the help output, then that is what the user is going to expect to be set if they do not provide an explicit argument to override it.

If you want the help to accurately reflect the default, then you will need to determine that value at runtime and have the default value field set to something like:

```go
defaultName := os.GetEnv("KIND_CLUSTER_NAME")
if defaultName == "" {
	defaultName = cluster.DefaultName
}

cmd.Flags().StringVarP(
	&flags.Name,
	"name",
	"n",
	defaultName,
	cli.NameFlagHelp,
)
```

aliheydarabadii: Hi, sorry for the late response — I was busy with exams.
I’ve updated the code as suggested. Could you please take a look and confirm if this is correct? Thanks again! @stmcginnis

pkg/internal/cli/override.go
```diff
@@ -1,18 +1,4 @@
 package cli
 
-import (
-	"os"
-
-	"github.com/spf13/pflag"
-)
-
-// OverrideDefaultName conditionally allows overriding the default cluster name
```
BenTheElder, line 9: I don't understand why we inlined this?

aliheydarabadii: We inlined this because the default needs to be computed at runtime so that --help reflects the actual behavior (env var first, fallback otherwise), as discussed earlier in the review.
Keeping it inline avoids a helper that mutates defaults after flag definition, which was the source of the confusion.

PR #4075, opened 2025-12-16
e2e-k8s.sh: support e2e.test parameters

pohly:
This is useful as an escape hatch for special kind-based jobs which need to control aspect of the e2e.test which are not covered by dedicated env variables.

The specific needs is for kubernetes-e2e-kind-alpha-beta-features-race: it needs to enable race detection via command line flags (https://github.com/kubernetes/kubernetes/pull/133844).

That PR currently uses an env variable for that, but a generic env variable is harder to use correctly than this PR (word splitting is ambiguous) and adding a dedicated env variable doesn't scale - that way leads to replicating more and more command line flags as env variables.

### DIFF
--- a/hack/ci/e2e-k8s.sh
+++ b/hack/ci/e2e-k8s.sh
@@ -15,7 +15,11 @@
 
 # hack script for running a kind e2e
 # must be run with a kubernetes checkout in $PWD (IE from the checkout)
-# Usage: SKIP="ginkgo skip regex" FOCUS="ginkgo focus regex" kind-e2e.sh
+# Usage: SKIP="ginkgo skip regex" FOCUS="ginkgo focus regex" kind-e2e.sh [args]
+#
+# <args> are additional command line arguments for e2e.test. They get
+# passed through to k/k/hack/ginkgo-e2e.sh, which in turn passes
+# them on to e2e.test.
 
 set -o errexit -o nounset -o xtrace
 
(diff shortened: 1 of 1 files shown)


aojea:
are ci failures related?

hack/ci/e2e-k8s.sh
```diff
@@ -15,7 +15,11 @@
 
 # hack script for running a kind e2e
 # must be run with a kubernetes checkout in $PWD (IE from the checkout)
-# Usage: SKIP="ginkgo skip regex" FOCUS="ginkgo focus regex" kind-e2e.sh
+# Usage: SKIP="ginkgo skip regex" FOCUS="ginkgo focus regex" kind-e2e.sh [args]
```
aojea, line 18: we try to avoid to extend this file and encourage other jobs to bring their own script, I defer to @BenTheElder as he put way more thought on this script than me

/assign @BenTheElder

BenTheElder: This is hard to support, if we ever want to start setting a flag in this script, now it will conflict with callers that set the flag manually, and the bash will get _really_ complex.

This is the same reason we have refused support for arbitrary docker cli flags to the node containers.

We already have an env for setting the race detector?

BenTheElder:
> (word splitting is ambiguous) 

This is something shellcheck will catch and point to best practices, but I'm also not sure how this relates to env at all ... what're we running into word-splitting with, exactly?

hack/ci/e2e-k8s.sh
BenTheElder: This script is intended to be an example, rather than completely general purpose, but even in that capacity this sort of "just plumb through arbitrary arguments" is a bit of a nightmare for evolving tools.

What happens when we switch to some future attempt at https://github.com/kubernetes/kubernetes/pull/131058 and jobs are depending on passing arbitrary flags to this script ...?

pohly:
I'll use env variables instead of parameters.

/close

Issue #3915, opened 2025-04-18
Tracking Issue: Dropping cgroup v1 support


BenTheElder:
Kubernetes has supported cgroup v2 since v1.19

Kubernetes froze cgroup v1 to maintenance mode in v1.31 forward:
https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/4569-cgroup-v1-maintenance-mode/README.md

systemd (which we need for the node init) has recently begun to move forward on dropping v1 support:
https://github.com/systemd/systemd/pull/36622

Eventually this will reach distro packages and we'll be forced to either stay on ancient distros or pick this up.

cgroup v2, combined with with container runtimes defaulting to cgroupns=private in v2 is substantially cleaner for "nested containers", we'll be able to drop a bunch of half-working hacks / workarounds

however, when this happens, we're also going to have a lot of confused users

At some point I think we should check for cgroup v1 in `docker info` (and podman / nerdctl) and make v2 required in the `kind` command, so we can give a good error message and then just drop support from the node images from some time after that releases.

I'm not sure exactly when we should do this, but we need to start thinking about it.

cc @aojea @stmcginnis 

Semi related: https://github.com/kubernetes/k8s.io/issues/5276, we need to plan for kind CI migrating primarily to cgroup v2 as the Kubernetes distros migrate and begin to drop v1.

stmcginnis:
Thinking about a minimal first step, what do you think of [checking the provider info](https://github.com/kubernetes-sigs/kind/blob/main/pkg/cluster/internal/create/create.go#L248) and emitting a log message with a warning about the use of cgroupv1.

So something along the lines of:

```go
func validateProvider(logger log.Logger, p providers.Provider) error {
	info, err := p.Info()
	if err != nil {
		return err
	}
	if info.Rootless {
		if !info.Cgroup2 {
			return errors.New("running kind with rootless provider requires cgroup v2, see https://kind.sigs.k8s.io/docs/user/rootless/")
		}
		if !info.SupportsMemoryLimit || !info.SupportsPidsLimit || !info.SupportsCPUShares {
			return errors.New("running kind with rootless provider requires setting systemd property \"Delegate=yes\", see https://kind.sigs.k8s.io/docs/user/rootless/")
		}
	} else if !info.Cgroup2 {
		logger.V(0).Info("⚠️ Deprecated use of cgroupv1. Cgroupv2 will be required in v1.36 and later.")
	}

	return nil
}
```

[...]

PR #4077, opened 2025-12-18
bump node image to 1.35.0

BenTheElder:
ahead of v0.31.0 later today as discussed with co-maintainers

### DIFF
--- a/.github/workflows/vm.yaml
+++ b/.github/workflows/vm.yaml
@@ -33,11 +33,6 @@ jobs:
           - template: fedora
             provider: podman
             rootless: rootless
-          # AlmaLinux 8 is used for testing kind with cgroup v1.
-          # Do not upgrade this to AlmaLinux 9 until dropping cgroup v1 support.
-          - template: almalinux-8
-            provider: docker
-            rootless: rootful
     env:
       KIND_EXPERIMENTAL_PROVIDER: "${{ matrix.provider }}"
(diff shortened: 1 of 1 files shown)


stmcginnis:
/lgtm
/approve
/hold

Just holding in case we want to check anything out closer.

BenTheElder:
oh right, cgroupv1 CI is failing *because cgroupv1 isn't supported by default in Kubernetes* in 1.35+

we should add a prominent release note about this

.github/workflows/vm.yaml
```diff
@@ -33,11 +33,6 @@ jobs:
           - template: fedora
             provider: podman
             rootless: rootless
-          # AlmaLinux 8 is used for testing kind with cgroup v1.
```
BenTheElder, line 36: https://github.com/kubernetes-sigs/kind/issues/3915

BenTheElder:
OK, this passes all the github actions, which are what tests the node image.

We removed the cgroup v1 action, but we care more about providing 1.35 OOTB that cgroup v1 support, and I'd rather not employ special workarounds because frankly this is the opportunity we need to drop v1 support while that is coordinated ecosystem wide rather than just being us.

BenTheElder:
this is a blocker for v0.31, and I've discussed merging this for 0.31 and the problem with cgroup v1 with sean already but he had to head out

PR #4082, opened 2025-12-18
upgrade to runc 1.3.4

BenTheElder:
containerd 2.2.1 is not out yet but will use runc 1.3.4, we'll pickup 2.2.1 later

runc 1.4.0 is out yet, but I don't want to rush to the bleeding edge over the holidays after just getting 1.3.x unblocked


BenTheElder:
unblocked by https://github.com/kubernetes/kubernetes/pull/135789

BenTheElder:
=> https://github.com/kubernetes-sigs/kind/pull/4083

PR #4088, opened 2025-12-23
cmd: clarify stack trace logging behavior

loserfruit966:
This replaces a stale TODO with a comment that documents the intended
behavior. Stack traces are already gated by verbosity, so this clarifies
why they are only shown at higher log levels without changing behavior.


stmcginnis:
I don't think this is quite right. The TODO still applies. Right now it is shown at a `1` level, but stack traces are a very verbose and detailed thing that should probably be only emitted at a higher level than that.

loserfruit966:
Thanks for the clarification — that makes sense. I’ll close this for now.

PR #4089, opened 2025-12-24
add name for kind cluster

esmaeelE:
This multi node cluster need a name to override default in order to not conflict with previous simple node.

```
$ kind create cluster
```
create a Kubernetes cluster with name kind by default.
we need to change kind config file to prevent name conflict.

So beginner user can follow quick start guide without interruption.

### DIFF
--- a/site/content/docs/user/kind-example-config.yaml
+++ b/site/content/docs/user/kind-example-config.yaml
@@ -2,6 +2,7 @@
 # NOTE: this is not a particularly useful config file
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
+name: multinode
 # patch the generated kubeadm config with some extra settings
 kubeadmConfigPatches:
 - |


stmcginnis:
On the inverse side, now it may not be as obvious to a new user going through the quick start that they will need to pass `--name` when they delete the cluster. I'd personally rather keep this simple and less confusing, but we can see what the other maintainers think.

At a minimum, you will need to sign the CLA before we could accept any updates.

esmaeelE:
> On the inverse side, now it may not be as obvious to a new user going through the quick start that they will need to pass `--name` when they delete the cluster. I'd personally rather keep this simple and less confusing, but we can see what the other maintainers think.
> 
> At a minimum, you will need to sign the CLA before we could accept any updates.

I am unable to sign the Cloud Native Computing Foundation (CNCF) due to my current location in Iran. I kindly request that you or the other repository maintainers make this change if it is acceptable. Thank you.

BenTheElder:
Thanks for contributing, but I think we'll leave this example as-is.

Also sorry, the CLA policy is not up to us.

esmaeelE:
> Thanks for contributing, but I think we'll leave this example as-is.
> 
> Also sorry, the CLA policy is not up to us.

OK, thanks for your response, I understand.

PR #4093, opened 2026-01-09
docs: clarify resource requirement scope in quick start

### LINKED ISSUE #485: determine and document resource requirements

AkshatDudeja77:
This PR clarifies that the resource values mentioned in the quick start
are minimums for basic clusters and may vary depending on workload
(e.g., building node images or multi-node clusters).

Closes: #485.

### DIFF
--- a/site/content/docs/user/quick-start.md
+++ b/site/content/docs/user/quick-start.md
@@ -323,7 +323,7 @@ kind build node-image --type source $HOME/go/src/k8s.io/kubernetes/
 If you are building Kubernetes (for example - `kind build node-image`) on MacOS or Windows then you need a minimum of 6GB of RAM
 dedicated to the virtual machine (VM) running the Docker engine. 8GB is recommended.
 
-> Note: Resource requirements vary depending on usage.
+> **Note**: Resource requirements vary depending on usage.
 > The values listed here are minimums for basic clusters.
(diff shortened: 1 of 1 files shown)


site/content/docs/user/quick-start.md
```diff
@@ -323,6 +323,11 @@ kind build node-image --type source $HOME/go/src/k8s.io/kubernetes/
 If you are building Kubernetes (for example - `kind build node-image`) on MacOS or Windows then you need a minimum of 6GB of RAM
 dedicated to the virtual machine (VM) running the Docker engine. 8GB is recommended.
 
+> Note: Resource requirements vary depending on usage.
```
stmcginnis: It would be good to keep the formatting consistent with the other note call-outs in this doc.

```suggestion
> **Note**: Resource requirements vary depending on usage.
```

PR #4095, opened 2026-01-12
e2e: Stop force-skipping Serial tests in parallel mode

dims:
Ginkgo v2 natively handles Serial specs when running in parallel mode. According to the Ginkgo documentation [1], when parallel testing is enabled, Serial-decorated specs are automatically deferred until after parallel execution completes:

  "When it detects the presence of Serial specs, process #1 will wait
   for all other processes to exit before running the Serial specs."

The previous behavior of adding [Serial] to the SKIP regex was necessary for Ginkgo v1 but is no longer required. Removing this allows Serial tests to run automatically after parallel tests complete, which improves test coverage without requiring manual skip management.

This change enables CI jobs that use PARALLEL=true to get coverage of Serial tests (such as DaemonSet, PriorityClass, and Namespace tests) without needing separate test runs.

[1] https://onsi.github.io/ginkgo/#serial-specs

### DIFF
--- a/hack/ci/e2e-k8s.sh
+++ b/hack/ci/e2e-k8s.sh
@@ -25,9 +25,14 @@ set -o errexit -o nounset -o xtrace
 # LABEL_FILTER: ginkgo label query for selecting tests (see "Spec Labels" in https://onsi.github.io/ginkgo/#filtering-specs)
 #
 # The default is to focus on conformance tests when FOCUS is unset. To run all
-# tests (no focus filter), explicitly set FOCUS="". Serial tests get skipped when
-# parallel testing is enabled. Using LABEL_FILTER instead of combining SKIP and
-# FOCUS is recommended (more expressive, easier to read than regexp).
+# tests (no focus filter), explicitly set FOCUS="". Using LABEL_FILTER instead
+# of combining SKIP and FOCUS is recommended (more expressive, easier to read
+# than regexp).
+#
+# Note: Serial tests are handled automatically by Ginkgo v2. When running in
+# parallel mode, Ginkgo runs parallel specs first across all processes, then
+# runs Serial specs on process #1 after all other processes have exited.
+# See: https://onsi.github.io/ginkgo/#serial-specs
 #
 # FEATURE_GATES:
 #          JSON or YAML encoding of a string/bool map: {"FeatureGateA": true, "FeatureGateB": false}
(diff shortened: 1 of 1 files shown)


dims:
related: https://github.com/kubernetes/test-infra/pull/36206

dims:
ok, i gotta fix the CI job definitions.

hack/ci/e2e-k8s.sh
```diff
@@ -250,14 +255,11 @@ run_tests() {
...
+  # if we set PARALLEL=true, enable ginkgo parallel mode
+  # Note: Ginkgo v2 automatically handles Serial specs - they run on process #1
+  # after all parallel specs complete. No need to skip them explicitly.
   if [ "${PARALLEL:-false}" = "true" ]; then
     export GINKGO_PARALLEL=y
-    if [ -z "${SKIP}" ]; then
-      SKIP="\\[Serial\\]"
-    else
-      SKIP="\\[Serial\\]|${SKIP}"
-    fi
```
pohly, line 260: I have been bitten by this before (it's totally unexpected!), so I am all for removing this.

However, this is a breaking change for *all* consumers of this script, which includes also jobs forked for prior releases. It might not be possible to do this as proposed here.

My own prior attempt to improve the script is pending in https://github.com/kubernetes-sigs/kind/pull/4015.

The difference is that I avoid setting SKIP only if a job has opted into using LABEL_FILTER. Should be less disruptive.

But my PR stalled nonetheless. The [proposed job update](https://github.com/kubernetes/test-infra/pull/35594) never got feedback.

dims: @pohly i am happy for either https://github.com/kubernetes-sigs/kind/pull/4015 or this to land. you may be right in `LABEL_FILTER` is better, i just wanted to rip the whole thing out!

https://github.com/kubernetes/test-infra/pull/36210 just exposes what exactly the script did under the covers in the individual jobs in the job definition itself across all affected jobs, so that can land by itself independent of this one

I know @BenTheElder asked what is the need for both serial + parallel. i need that for `ci-kubernetes-e2e-kind-audit` which needs to capture audit logs for all tests for analysis.

pohly: I think the need is clear: have one job that really runs all tests that are available.

BenTheElder: This also needs to be safe across many kubernetes versions in addition to other jobs using it ...

I think it would be better to try this in a dedicated script first, this seems high risk for introducing flakes.

also re: kind CI jobs and risky changes at the moment, the kubeadm v1beta4 transition mess is going to consume a lot of energy for updating these and keeping CI healthy but has to be done. I need to work on some other things but that's blocking kubeadm so spare time kinda has to go there.

dims: @BenTheElder ack. will close this one.

PR #4129, opened 2026-04-07
kindnetd: validate noMasqueradeCIDRs before creating IPMasqAgent

rayowang:
To reject invalid or wrong-family noMasqueradeCIDRs early and add focused unit tests.

### DIFF
--- a/images/kindnetd/cmd/kindnetd/masq.go
+++ b/images/kindnetd/cmd/kindnetd/masq.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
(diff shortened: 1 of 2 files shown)


aojea:
This version of kindnet is tightly coupled to kind, you see this value comes from https://github.com/kubernetes-sigs/kind/blob/2a0ee8aca75b91296a217b6cf3076341785a7329/pkg/build/nodeimage/const_cni.go#L121-L122

so I can not see how this can get a wrong value there.

The full version of kindnet is in https://github.com/kubernetes-sigs/kindnet that is more general that this specific flavor

PR #4133, opened 2026-04-16
config: validate and default port mapping fields

ANISHSAJIKUMAR:
Improve node config validation by defaulting empty port mapping protocols to TCP and rejecting invalid listenAddress values early. Includes tests.


stmcginnis:
Is this necessary? What problem was this causing you?

PR #4135, opened 2026-04-16
docs: clarify rootless quick-start note

ANISHSAJIKUMAR:
Small docs-only wording cleanup in the quick-start guide. This keeps the note accurate while reading more naturally for users following the rootless provider paths.


stmcginnis:
Dupe of #4137

PR #4137, opened 2026-04-16
docs: clarify auditing configuration wording

ANISHSAJIKUMAR:
Small docs-only cleanup in the auditing guide. This improves the wording around configuration instructions without changing the example or the setup flow.

### DIFF
--- a/site/content/docs/user/using-wsl2.md
+++ b/site/content/docs/user/using-wsl2.md
@@ -44,7 +44,7 @@ If you're using a physical machine, you can mount the ISO, copy the files to a F
 
 If you want the full details, see the [Installation Instructions for WSL2](https://docs.microsoft.com/en-us/windows/wsl/wsl2-install). This is the TL;DR version.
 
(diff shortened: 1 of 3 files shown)


site/content/docs/user/using-wsl2.md
```diff
@@ -44,7 +44,7 @@ If you're using a physical machine, you can mount the ISO, copy the files to a F
 
 If you want the full details, see the [Installation Instructions for WSL2](https://docs.microsoft.com/en-us/windows/wsl/wsl2-install). This is the TL;DR version.
 
-Once your Windows machine is ready, you need to do a few more steps to set up WSL2
+Once your Windows machine is ready, you need to do a few more steps to set up WSL2.
```
stmcginnis, line 47: This leads into the bullet list, so I think it was fine as is. If anything it should be changed to `:`.

stmcginnis (commented):
This is more verbose, but I don't think it really adds anything or makes anything more clear. Is there a reason for making this change?

PR #4139, opened 2026-04-16
feat: add kind env command

ANISHSAJIKUMAR:
Add a new `kind env` command that prints the environment variables influencing cluster creation and the current value of each one. This makes it easier to discover provider, networking, and kubeconfig overrides from the CLI itself.\n\nTests cover unset, empty, and set values, and the root command now exposes the new subcommand.

### DIFF
--- a/pkg/cmd/kind/env/env.go
+++ b/pkg/cmd/kind/env/env.go
@@ -1,0 +1,99 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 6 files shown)


stmcginnis:
Please stop including all previous commits in new changes. You will need to clean this up.

A similar approach has already been discussed and I believe there is an open issue with some more context. Make sure to check for existing work before proposing new.

If there is something like this you want to add, [make sure you follow the process of filing an issue](https://kind.sigs.k8s.io/docs/contributing/getting-started/#4-reaching-out) so it can be discussed prior to submitting actual PRs.

If these stacked PRs continue and there is no response then I am going to assume these are being entirely LLM drive and I will start to just close new PRs without further discussion.

PR #4143, opened 2026-04-22
docs(rootless): fix `/etc/systctl.d` -> `/etc/sysctl.d` typo

### LINKED ISSUE #4132: Typo in documentation for kind in rootless mode
[This page](https://kind.sigs.k8s.io/docs/user/rootless/#increase-inotify-limits) suggests modifying `/etc/systctl.d/.conf` but I think it should be `/etc/sysctl.d/.conf`

SAY-5:
Closes #4132.

The [rootless guide](https://kind.sigs.k8s.io/docs/user/rootless/#increase-inotify-limits) told users to drop a \`.conf\` file in \`/etc/systctl.d\` in two places (the inotify-limits section and the unprivileged-port-range section). That path doesn't exist — the real one is \`/etc/sysctl.d\` (\`man sysctl.d\` / \`systemd-sysctl\`), which is also what the follow-up \`sudo sysctl --system\` lines right below each step already invoke. Anyone pasting the snippet verbatim ends up with a file systemd never reads.

One-character fix, same replacement in both places.

Signed-off-by: SAY-5 <SAY-5@users.noreply.github.com>

### DIFF
--- a/site/content/docs/user/rootless.md
+++ b/site/content/docs/user/rootless.md
@@ -156,7 +156,7 @@ are particularly susceptible to this issue.
 
 To increase the inotify limits, do the following:
 
-1. As root, create a `.conf` file in `/etc/systctl.d` that increases the `fs.inotify` max user settings:
+1. As root, create a `.conf` file in `/etc/sysctl.d` that increases the `fs.inotify` max user settings:
 
    ```
    fs.inotify.max_user_watches = 524288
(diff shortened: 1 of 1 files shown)


SAY-5:
Force-pushed to drop the "Closes #4132" auto-close keyword from the commit message (flagged by prow:invalid-commit-message). Now refers to the issue as "Refs kubernetes-sigs/kind issue 4132" instead.

stmcginnis:
> Now refers to the issue as "Refs kubernetes-sigs/kind issue 4132" instead.

Please remove that completely from the commit message.

PR #4146, opened 2026-05-01
Add release process documentation

stmcginnis:
This attempts to capture some of the tribal knowlege into a doc that new maintainers should be able to follow to do a new kind release. It also serves as a useful reference doc for non-maintainers to understand what is involved in a release, and for other projects to potentially learn and borrow from.

It's also hoped that by having all of the steps clearly written down we might be able to refine or automate more portions of the process.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,226 @@
+# kind Release Process
+
+This document describes the steps to cut a new kind release. It is intended for
+maintainers who have push access to the upstream repository and the staging
+image registry.
+
+## Prerequisites
+
+- GNU sed (macOS: `brew install gnu-sed`)
+- Docker with buildx support
+- [`crane`](https://github.com/google/go-containerregistry/blob/main/cmd/crane/README.md) installed (for image promotion to Docker Hub)
+- Push access to `github.com/kubernetes-sigs/kind`
+- Push access to `gcr.io/k8s-staging-kind`
+- Push access to `kindest` on Docker Hub
+
+---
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -0,0 +1,226 @@
+# kind Release Process
+
+This document describes the steps to cut a new kind release. It is intended for
+maintainers who have push access to the upstream repository and the staging
+image registry.
+
+## Prerequisites
+
+- GNU sed (macOS: `brew install gnu-sed`)
+- Docker with buildx support and experimental manifest commands enabled
```
BenTheElder: IIRC we enable manifest commands in the script, so you just need docker with buildx

RELEASE.md
```diff
@@ -0,0 +1,226 @@
...
+- `crane` installed (for image promotion to Docker Hub)
```
BenTheElder: maybe add a link to this project?

BenTheElder (commented):
/lgtm
/approve

stmcginnis:
Hitting that failure again:

```
There are no nodes that your pod can schedule to
```

/test pull-kind-verify

> maybe add a link to this project?

Yeah, good call. I'll make a note and follow up. My thought was we can use this as a sort of guide for the next few releases, and each time we should be looking at what we can add or refine to make sure all our assumptions make it into the doc so it just gets better over time.

BenTheElder (commented):
/lgtm

PR #4148, opened 2026-05-04
bump go to 1.26.2

BenTheElder:


BenTheElder:
`make -C images/base quick` works after this locally, will let CI check for some of the other bits.

Issue #3758, opened 2024-10-15
Cluster creation fails with `"Reached target .*Multi-User System.*"`


DerekTBrown:
I am unable to create a Kind cluster using `kind create`. I get an error:

```

ERROR: failed to create cluster: could not find a log line that matches "Reached target .*Multi-User System.*\|detected cgroup v1"
--
  | Stack Trace:
  | sigs.k8s.io/kind/pkg/errors.Errorf
  | sigs.k8s.io/kind/pkg/errors/errors.go:41
  | sigs.k8s.io/kind/pkg/cluster/internal/providers/common.WaitUntilLogRegexpMatches
  | sigs.k8s.io/kind/pkg/cluster/internal/providers/common/cgroups.go:84
  | sigs.k8s.io/kind/pkg/cluster/internal/providers/docker.createContainerWithWaitUntilSystemdReachesMultiUserSystem
  | sigs.k8s.io/kind/pkg/cluster/internal/providers/docker/provision.go:417
  | sigs.k8s.io/kind/pkg/cluster/internal/providers/docker.planCreation.func3
  | sigs.k8s.io/kind/pkg/cluster/internal/providers/docker/provision.go:123
  | sigs.k8s.io/kind/pkg/errors.UntilErrorConcurrent.func1
  | sigs.k8s.io/kind/pkg/errors/concurrent.go:30
  | runtime.goexit
  | runtime/asm_amd64.s:1695
  | creating kind cluster: exit status 1
```

When I run `kind export logs`, I plainly see this line in the `journal.log`:

```
Oct 15 00:54:00 <my-cluster-name> systemd[1]: Reached target multi-user.target - Multi-User System.
```

I don't see anything else in the log bundle to indicate an error:

[kind_logs.tar.gz](https://github.com/user-attachments/files/17370933/kind_logs.tar.gz)

--- 

## Appendix: things I eliminated as possible root causes from similar issues
- I am on the latest version of `kind`.
- It doesn't appear to be related to `max_user_instances` or `max_user_watches`. I have increased these to extremely large limits.
- It doesn't appear to be related to kernel version, I am on a relatively new kernel.
- It doesn't appear to be related to OS Architecture. I don't see anything in the logs to indicate a Docker image/machine arch mismatch, and the Docker image manifests seem to align with the underlying infrastructure (AMD64).

DerekTBrown:
[...]

PR #4151, opened 2026-05-08
Add log driver workaround to podman docs

stmcginnis:
This adds a note to the rootless podman docs for cases where the expected status line is not found in the logging output, but the node is actually ready. In some cases this has been found to be caused by delays in relaying the emitted logs. Changing the podman logging provider can work around this.

See discussion in #3758 for more details. Thanks @guimochila for finding this solution.


BenTheElder (approved):
/lgtm
/approve

PR #4153, opened 2026-05-10
fix: handle empty port mapping listen addresses

immanuwell:
Fixes a panic when `extraPortMappings` omits `listenAddress`. Docs say it is optional, and providers already default it to wildcard, so validation should treat it that way too. tiny footgun tbh.

Repro:
```yaml
kind: Cluster
nodes:
- role: control-plane
  extraPortMappings:
  - containerPort: 80
    hostPort: 8080
```

Run:
```bash
kind create cluster --config config.yaml
```

Before: validation can panic in `validatePortMappings` on `addr.String()`.
After: empty `listenAddress` is checked as `0.0.0.0`; bad values get a normal validation error, no boom.

Checked:
- `go test ./pkg/internal/apis/config`
- `make unit`
- `make verify` got through lints/generated, then shellcheck hit local Docker socket perms here.

Related: #4133

### DIFF
--- a/pkg/internal/apis/config/validate.go
+++ b/pkg/internal/apis/config/validate.go
@@ -163,6 +163,12 @@ func validatePortMappings(portMappings []PortMapping) error {
 		}
 
 		addr := net.ParseIP(portMapping.ListenAddress)
+		if portMapping.ListenAddress == "" {
+			addr = wildcardAddrIPv4
+		}
+		if addr == nil {
+			return fmt.Errorf("invalid listen address: %s", portMapping.ListenAddress)
+		}
 		addrString := addr.String()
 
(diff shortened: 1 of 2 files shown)


aojea:
/ok-to-test
/lgtm
/approve

Thanks

pkg/internal/apis/config/validate.go
```diff
@@ -163,6 +163,12 @@ func validatePortMappings(portMappings []PortMapping) error {
 		}
 
 		addr := net.ParseIP(portMapping.ListenAddress)
+		if portMapping.ListenAddress == "" {
+			addr = wildcardAddrIPv4
```
BenTheElder, line 167: Is this how docker behaves internally? Or will it sometimes use ipv6 / dualstack? I can't remember anymore.
I'm wondering if instead we should permit unset in validation?

PR #4160, opened 2026-05-13
docs: document resource requirements for running and building kind clusters

### LINKED ISSUE #485: determine and document resource requirements

fru1tworld:
Fixes #485

Reworked per maintainer feedback on this PR.

Instead of expanding `quick-start.md`, the resource notes now live in a
dedicated sub-page at `site/content/docs/user/resource-requirements.md`.
The page drops the prescriptive RAM/CPU figures from the original draft
and instead calls out that disk I/O for etcd is usually the limiting
factor on small hosts. It documents the
`io.x-k8s.kind.cluster=<name>` label-based measurement recipe with
variants for Docker, Podman, and nerdctl, and links to metrics-server
for per-pod numbers. The existing `Settings for Docker Desktop` section
in the quick start is left untouched and is referenced from the new
page for build-time VM sizing.

Supersedes #4093 in scope.

/kind documentation
/sig cluster-lifecycle

### DIFF
--- a/site/content/docs/user/quick-start.md
+++ b/site/content/docs/user/quick-start.md
@@ -330,6 +330,45 @@ kind build node-image --type source $HOME/go/src/k8s.io/kubernetes/
 > **NOTE**: modes other than source directory namely `url`, `file` and `release` are only
 > available in kind v0.24 and above.
 
+## Resource Requirements
+
+kind clusters run as Docker containers on the host, so the host's CPU and
+memory limits apply directly to the cluster.
+
+For typical single-node clusters with light workloads, Docker's default
+resource allocation (around 2 CPUs and 2 GB of memory on Docker Desktop) is
+usually enough to bring a cluster up. For more reliable operation with many
+pods or multi-node clusters, we recommend allocating at least **4 GB of memory
+and 2 CPUs** to Docker, and scaling up further for heavier workloads.
+
+Building Kubernetes node images with `kind build node-image` has a higher
+floor: see [Settings for Docker Desktop](#settings-for-docker-desktop) below.
+
+Exact numbers depend on the Kubernetes version, the CNI in use, and the
+workloads running on the cluster, so we recommend measuring against your own
+scenario rather than relying on fixed figures.
+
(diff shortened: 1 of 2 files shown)


stmcginnis:
We already have https://kind.sigs.k8s.io/docs/user/quick-start/#settings-for-docker-desktop that sort of covers this. There's also the earlier PR #4093 that's still pending.

That said, I think this is a little more prescriptive in what to use, so I think there's some value. But I think if we're going to add this then we should do something about the Docker Desktop section.

We also now support (sort of) other runtimes besides docker, so it might be nice if we're giving specific commands to run to also show variants for podman, nerdctl, etc. But not strictly required.

BenTheElder:
Thanks for the PR!

It's really hard to nail this down, it depends so much more on the workload than just running kind itself, I don't think you even need 2GB.

Usually what brings down running kind on really small environments is I/O for etcd.

I think if we're going to add a lot more text around e.g. docker desktop, we should break that out into a sub-page, "quick start" has become a monster.

fru1tworld:
@BenTheElder thanks — moved the material into a new `resource-requirements.md` sub-page so quick-start doesn't grow further.
Dropped the fixed RAM/CPU numbers and replaced them with the etcd-I/O point you raised, plus a runtime-agnostic measurement recipe.

fru1tworld:
@stmcginnis thanks — added Podman and nerdctl variants of the `stats` recipe alongside Docker in the new sub-page.
Left the existing `Settings for Docker Desktop` section untouched and link to it from the new page so the two don't overlap.

PR #4162, opened 2026-05-14
Add warning when using cgroupv1

stmcginnis:
To help prepare for the eventually removal of cgroup v1 support, this adds a check and warning if we are running on a system that does not support cgroup v2.

Related-to: #3915

### DIFF
--- a/pkg/cluster/internal/create/create.go
+++ b/pkg/cluster/internal/create/create.go
@@ -68,7 +68,7 @@ type ClusterOptions struct {
 // Cluster creates a cluster
 func Cluster(logger log.Logger, p providers.Provider, opts *ClusterOptions) error {
 	// validate provider first
-	if err := validateProvider(p); err != nil {
+	if err := validateProvider(logger, p); err != nil {
 		return err
 	}
 
@@ -241,7 +241,7 @@ func fixupOptions(opts *ClusterOptions) error {
 	return nil
 }
 
-func validateProvider(p providers.Provider) error {
+func validateProvider(logger log.Logger, p providers.Provider) error {
 	info, err := p.Info()
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


pkg/cluster/internal/create/create.go
```diff
@@ -253,6 +253,8 @@ func validateProvider(p providers.Provider) error {
 		if !info.SupportsMemoryLimit || !info.SupportsPidsLimit || !info.SupportsCPUShares {
 			return errors.New("running kind with rootless provider requires setting systemd property \"Delegate=yes\", see https://kind.sigs.k8s.io/docs/user/rootless/")
 		}
+	} else if !info.Cgroup2 {
+		logger.Warn("cgroup v1 is deprecated and will not be supported in a future kind release, please upgrade to cgroup v2")
```
BenTheElder:
```suggestion
		logger.Warn("cgroup v1 is deprecated in Kubernetes and will not be supported in a future kind release, please upgrade to cgroup v2")
```

Was going to add the announcement link but we don't really have one

BenTheElder:
+1, one suggestion

stmcginnis:
Looks like we do have a little more time for this, but good to start warning people now so it's less of a hassle when the time actually comes. https://github.com/kubernetes/website/pull/55809

BenTheElder:
> Looks like we do have a little more time for this, but good to start warning people now so it's less of a hassle when the time actually comes.

Agreed, I was just looking for what we can point users to.

> https://github.com/kubernetes/website/pull/55809

That's actually unrelated, in this case they mean systemd vs cgroupfs instead of v1 vs v2.

BenTheElder (commented):
/lgtm
/approve

stmcginnis:
/retest

Seems a lot of flakiness lately...

PR #4164, opened 2026-05-21
fix: handle registry ports in image normalization

immanuwell:
Tiny papercut in image ref normalization.

Registries with ports are valid refs, e.g. `localhost:5001/foo:tag`; kind local registry docs use this shape too. Some helpers treated the registry port colon like an image tag separator.

Repro, before this patch:

```sh
go test ./pkg/cluster/internal/providers/podman -run Test_sanitizeImage -count=1
go test ./pkg/cmd/kind/load/docker-image -run Test_sanitizeImage -count=1
```

The podman path turned `localhost:5000/kindest/node:v1.21.1@sha256:...` into `library/localhost@sha256:...`, yikes. `kind load docker-image localhost:5000/baz` also missed the implicit `:latest` tag.

This keeps colons before the last slash as part of the registry, and only strips/detects tags after the image path starts.

Verified:

```sh
go test ./pkg/... ./cmd/...
make unit
make verify
```

Related: #1213

### DIFF
--- a/pkg/cluster/internal/providers/podman/images_test.go
+++ b/pkg/cluster/internal/providers/podman/images_test.go
@@ -73,6 +73,11 @@ func Test_sanitizeImage(t *testing.T) {
 			pullImageName:     "foo.bar/baz@sha256:69860bda5563ac81e3c0057d654b5253219618a22ec3a346306239bba8cfa1a6",
 		},
 		{
+			image:             "localhost:5000/kindest/node:v1.21.1@sha256:69860bda5563ac81e3c0057d654b5253219618a22ec3a346306239bba8cfa1a6",
+			friendlyImageName: "localhost:5000/kindest/node:v1.21.1",
+			pullImageName:     "localhost:5000/kindest/node@sha256:69860bda5563ac81e3c0057d654b5253219618a22ec3a346306239bba8cfa1a6",
+		},
+		{
 			image:             "baz:quux@sha256:69860bda5563ac81e3c0057d654b5253219618a22ec3a346306239bba8cfa1a6",
 			friendlyImageName: "baz:quux",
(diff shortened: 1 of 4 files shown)


stmcginnis:
Reran the failing test, but failed again. That doesn't appear to be related to this change though.

/lgtm

stmcginnis:
I believe I've addressed the test failure in #4165.

BenTheElder:
> I believe I've addressed the test failure in https://github.com/kubernetes-sigs/kind/pull/4165.

unfortunately github actions will run from the PR branch, so this will keep failing without a rebase. we can bypass it though

pkg/cluster/internal/providers/podman/images_test.go
```diff
@@ -72,6 +72,11 @@ func Test_sanitizeImage(t *testing.T) {
 			friendlyImageName: "foo.bar/baz",
 			pullImageName:     "foo.bar/baz@sha256:69860bda5563ac81e3c0057d654b5253219618a22ec3a346306239bba8cfa1a6",
 		},
+		{
+			image:             "localhost:5000/kindest/node:v1.21.1@sha256:69860bda5563ac81e3c0057d654b5253219618a22ec3a346306239bba8cfa1a6",
+			friendlyImageName: "localhost:5000/kindest/node:v1.21.1",
```
BenTheElder, line 77: Not a blocker for this PR, but I wonder if we should revisit this concept entirely and just reformat the CLI output so the images names are not on the spinner line. It's misleading to exclude the authoritative digest in favor of the tag .. even if it looked better.

Something for follow-up discussion @aojea @stmcginnis

BenTheElder (commented):
/lgtm
/approve

pkg/cluster/internal/providers/podman/images_test.go
stmcginnis: That seems reasonable.

PR #4165, opened 2026-05-21
ci: fix rootless Docker permission issues

stmcginnis:
Something changed with containerd, the version of systemd on Fedora, or some other related package that has introduced a failure in setting up rootless docker, causing permission denied and other failures.

`loginctl enable-linger` was called with $USER from the host (runner), but the Lima guest user is runner.guest, so linger was never enabled for the user actually running rootless Docker. Fix by evaluating the username inside the VM with `id -un`.

Disables SELinux enforcement due to issues with the `--detach-ns` switch needed.

Also fix the Export logs step to not abort early when Docker is not running, and collect journalctl output for the docker user service so future failures are self-diagnosing.

### DIFF
--- a/.github/workflows/vm.yaml
+++ b/.github/workflows/vm.yaml
@@ -75,8 +75,10 @@ jobs:
           limactl cp -r . default:/tmp/kind
           # Install packages
           lima sudo /tmp/kind/hack/ci/init-vm.sh
-          # Enable systemd lingering for rootless
-          lima sudo loginctl enable-linger "$USER"
+          # Enable systemd lingering for rootless. $USER on the host may differ
+          # from the Lima guest user (e.g. runner vs runner.guest), so derive
+          # the username inside the VM.
+          lima bash -c 'sudo loginctl enable-linger "$(id -un)"'
           # Install kind
           lima sudo git config --global --add safe.directory /tmp/kind
           lima sudo make -C /tmp/kind install INSTALL_DIR=/usr/bin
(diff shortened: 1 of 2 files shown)


BenTheElder:
https://github.com/moby/moby/blob/721883f7329e0c555ec00126ecb3bbe5511d50f5/contrib/dockerd-rootless.sh#L25-L27

interesting

.github/workflows/vm.yaml
```diff
@@ -86,8 +88,11 @@ jobs:
         run: |
           # Disable the rootful daemon
           "$HELPER" sudo systemctl disable --now docker
-          # Install the systemd unit
-          "$HELPER" dockerd-rootless-setuptool.sh install
+          # Install the systemd unit.
+          # On some systems (Fedora 43) SELinux may cause --detach-netns to
```
BenTheElder: Tempted to disable selinux instead or find some other fix, it sounds like setting this to false has performance downsides

stmcginnis: True, it (SELinux) isn't really something we are testing, so we don't really need it in CI. I will update by disabling enforcement instead.

hack/ci/init-vm.sh
```diff
@@ -40,6 +40,10 @@ GOARCH="$(uname -m | sed -e 's/aarch64/arm64/' -e 's/x86_64/amd64/')"
 curl -L -o /usr/bin/kubectl "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/${GOARCH}/kubectl"
 chmod +x /usr/bin/kubectl
 
+# Disable SELinux enforcement in CI -- we're not testing SELinux policy and
+# enforcing mode causes permission denied errors with rootlesskit's detach-netns.
+setenforce 0 || true
```
BenTheElder, line 45: should we only do this for rootless? can follow-up

BenTheElder (commented):
/lgtm
/approve

PR #4149, opened 2026-05-04
bump images

BenTheElder:
~~I'm seeing issues locally with flaky ctr => containerd, checking if it's just my machine.~~

Resolved.

PR #4166, opened 2026-05-21
re-attempt upgrading containerd + runc after regression fixes

BenTheElder:


BenTheElder:
x-ref #4149

stmcginnis:
Looks like this is hitting what I'm addressing in #4165

stmcginnis:
So if that fixes the CI issue...

/lgtm
/hold

PR #4172, opened 2026-05-28
upgrade kindnetd

BenTheElder:
TODO: https://github.com/kubernetes-sigs/kube-network-policies/issues/352


BenTheElder:
I locally patched this into a kind image and ran some tests. It seems to be good.

stmcginnis:
Looks good and CI is happy!

/lgtm
/approve

aojea:
@BenTheElder without using NRI now kindnetd has this race https://github.com/kubernetes/kubernetes/issues/85966 to apply network policies ... not a big deal for kind clusters but just FYI, the network policy is not enforced until the PodIP is updated, that may lag a bit since the Pod has network connectivity

BenTheElder:
It wasn't using NRI before though, is that problem actually new or there is just a way to resolve it now? We can revisit later if we have issues in CI or something.

PR #4174, opened 2026-05-28
add support for building node images from Kubernetes CI artifacts

BenTheElder:
So you can do like `kind build node-image ci/latest`, `kind build node-image ci/latest-1.36` in addition to the release markers

FYI @kannon92 @dims @pohly 

/hold

### DIFF
--- a/pkg/build/nodeimage/build_test.go
+++ b/pkg/build/nodeimage/build_test.go
@@ -1,0 +1,98 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


pkg/build/nodeimage/build_test.go
```diff
@@ -0,0 +1,98 @@
+/*
+Copyright 2026 The Kubernetes Authors.
```
stmcginnis: nit: Year should be dropped from headers now.

BenTheElder: Still a force of habit 😅 
Will fix.

PR #4176, opened 2026-05-29
Update boilerplate template header to exclude year

stmcginnis:
Starting in 2026, the guidance is to not include the year in our copyright headers. This updates the boilerplate template file to drop the year so generated files no longer include it.


BenTheElder (approved):
Thanks!
/lgtm
/approve

We may also need some kind of boilerplate check I guess, there just hasn't been a great tool for it in any of the repos yet. They're all kind of buggy and most of them are python which we don't currently require to develop.

stmcginnis:
> We may also need some kind of boilerplate check I guess, there just hasn't been a great tool for it in any of the repos yet. They're all kind of buggy and most of them are python which we don't currently require to develop.

Prow has some validation checks that seem to work reasonably well. I can see what it would take to replicate something similar here.

PR #4186, opened 2026-06-08
fix: skip inaccessible kubeconfig entries when choosing merge target

immanuwell:
kind can panic here if `KUBECONFIG` starts with a path under a dir that can't be stat'd. Kinda nasty, and pretty easy to hit with a root owned dir or a busted mount.

This change treats any `os.Stat` error as not usable, then keeps going to the next path.

Repro
1. `tmp=$(mktemp -d)`
2. `mkdir "$tmp/locked" && touch "$tmp/fallback" && chmod 000 "$tmp/locked"`
3. `export KUBECONFIG="$tmp/locked/config:$tmp/fallback"`
4. Run a kubeconfig writing flow like `kind export kubeconfig` or `kind create cluster`
5. Before this fix, kind can blow up with a nil pointer panic
6. After this fix, it skips the bad entry and uses the fallback file

### DIFF
--- a/pkg/cluster/internal/kubeconfig/internal/kubeconfig/paths.go
+++ b/pkg/cluster/internal/kubeconfig/internal/kubeconfig/paths.go
@@ -74,7 +74,7 @@ func pathForMerge(explicitPath string, getEnv func(string) string) string {
 
 func fileExists(filename string) bool {
 	info, err := os.Stat(filename)
-	if os.IsNotExist(err) {
+	if err != nil {
 		return false
 	}
 	return !info.IsDir()
--- a/pkg/cluster/internal/kubeconfig/internal/kubeconfig/paths_test.go
+++ b/pkg/cluster/internal/kubeconfig/internal/kubeconfig/paths_test.go
@@ -129,7 +129,11 @@ func TestPathForMerge(t *testing.T) {
 		if err := os.Mkdir(inaccessibleDir, 0o700); err != nil {
 			t.Fatalf("failed to create inaccessible dir: %v", err)
 		}
-		defer os.Chmod(inaccessibleDir, 0o700)
+		defer func() {
+			if err := os.Chmod(inaccessibleDir, 0o700); err != nil {
+				t.Fatalf("failed to restore dir permissions: %v", err)
+			}
+		}()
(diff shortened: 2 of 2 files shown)


pkg/cluster/internal/kubeconfig/internal/kubeconfig/paths.go
```diff
@@ -74,7 +74,7 @@ func pathForMerge(explicitPath string, getEnv func(string) string) string {
 
 func fileExists(filename string) bool {
 	info, err := os.Stat(filename)
-	if os.IsNotExist(err) {
+	if err != nil {
 		return false
 	}
 	return !info.IsDir()
```
stmcginnis, line 80: I see, this would be nil if it's any error other than NotExist. Fix makes sense then. It might be better to rename the function call then since it's _slightly_ misleading. But I think it's fine - as far as our kubeconfig processing is concerned, the file does not exist. Even if there is a file with that name but can't be accessed for any reason.

stmcginnis (commented):
/ok-to-test
/lgtm

BenTheElder:
Thanks!

Has anyone checked how client-go handles this? We're basically aiming to match client-go behavior (E.g. kubectl etc) without directly depending on it.

immanuwell:
Checked this against client-go before pushing the follow-up fix.

`ClientConfigLoadingRules.GetDefaultFilename()` walks the precedence list and only returns a path when `os.Stat(filename)` succeeds; otherwise it keeps going. So an inaccessible entry is effectively skipped there the same way it would be here.

I also pushed `cd97c98a` to fix the verify failure in the new test (`errcheck` on the deferred `os.Chmod`). Local `hack/make-rules/verify/all.sh` passes now.

@BenTheElder

pkg/cluster/internal/kubeconfig/internal/kubeconfig/paths.go
immanuwell: Agreed, I left the name as-is since, for merge-target selection, an inaccessible path is effectively not usable
I also pushed cd97c98a to fix the verify failure in the new test

BenTheElder (commented):
/lgtm
/approve
thank you!!

PR #4188, opened 2026-06-10
docs: mention Docker-Mac-Net-Connect for LoadBalancer

### LINKED ISSUE #3448: Add reference to Docker-Mac-Net-Connect to LoadBalancer page
The LoadBalancer [page](https://kind.sigs.k8s.io/docs/user/loadbalancer/) correctly states that "On macOS and Windows, docker does not expose the docker network to the host." It also suggests using Ingress as a workaround. On MacOS, another workaround is to use [Docker-Mac-Net-Connect](https://github.com/chipmk/docker-mac-net-connect) to access the Docker network.

**Why is this needed**:
While Ingress is a suitable workaround for some cases, others may require direct access to a LB. Including this workaround can save time in that users would not need to discover it on their own.

popsiclelmlm:
## Summary
- Add a LoadBalancer documentation note for Docker Desktop on macOS users who cannot reach the assigned LoadBalancer IP directly.
- Link Docker-Mac-Net-Connect as a community routing helper for making Docker container networks reachable from macOS.

## Reproduction
- Follow the LoadBalancer guide on Docker Desktop for macOS.
- Cloud Provider KIND assigns a LoadBalancer IP from the Docker network.
- The existing verification step asks users to curl that IP directly from the host, which may fail if the Docker network is not routed from macOS.

## Root cause
Docker Desktop on macOS does not expose Docker container networks to the host by default, but the LoadBalancer guide did not mention this platform-specific networking limitation or a possible routing helper.

## Changes
- Add a concise note beside the LoadBalancer verification command.
- Keep the guidance scoped to macOS and describe Docker-Mac-Net-Connect as a community tool rather than an official requirement.

## Tests
- `git diff --check`
- `make -C site build` (passes; Hugo reports existing deprecation warnings from the site config)

## Screenshots/Logs
Not applicable; documentation-only change.

Fixes #3448.

### DIFF
--- a/site/content/docs/user/loadbalancer.md
+++ b/site/content/docs/user/loadbalancer.md
@@ -50,9 +50,16 @@ Now verify that the loadbalancer works by sending traffic to it's external IP an
 LB_IP=$(kubectl get svc/foo-service -o=jsonpath='{.status.loadBalancer.ingress[0].ip}')
 {{< /codeFromInline >}}
 
+> **NOTE**: If you are using Docker Desktop on macOS and cannot reach the
+> assigned LoadBalancer IP directly, the Docker network may not be routed from
+> the host. Community tools such as [Docker-Mac-Net-Connect] can make Docker
+> container networks reachable from macOS.
+
 ```bash
 # should output foo and bar on separate lines 
 for _ in {1..10}; do
   curl ${LB_IP}:5678
 done
 ```
+
+[Docker-Mac-Net-Connect]: https://github.com/chipmk/docker-mac-net-connect


site/content/docs/user/loadbalancer.md
```diff
@@ -50,9 +50,16 @@ Now verify that the loadbalancer works by sending traffic to it's external IP an
 LB_IP=$(kubectl get svc/foo-service -o=jsonpath='{.status.loadBalancer.ingress[0].ip}')
 {{< /codeFromInline >}}
 
+> **NOTE**: If you are using Docker Desktop on macOS and cannot reach the
```
BenTheElder, line 53: Is this still the case with cloud-provider-kind? I don't think it is anymore.

aojea: yeah, it should work with cloud provider kind

BenTheElder:
Thanks but this is obsolete.

PR #4189, opened 2026-06-11
docs: mention podman local registry push option

### LINKED ISSUE #3468: User guide for setting up kind with local registry does not work

Trying to follow the user guide on setting up [Kind cluster with local registry ](https://kind.sigs.k8s.io/docs/user/local-registry/) does not seem to work. After running the script as is, I get an error on step 3 `Then we’ll push it to the registry docker push localhost:5001/hello-app:1.0`. 
Error is `pinging container registry localhost:5001: Get "https://localhost:5001/v2/": http: server gave HTTP response to HTTPS client`. 

**Why is this needed**:
Following the documentation should simply work and common pitfalls should be addressed in the documentation.

popsiclelmlm:
## Summary
Add a Podman-specific note to the local registry guide for localhost HTTP registry pushes.

## Reproduction
Following the local registry guide with Podman can fail when pushing to `localhost:5001` with `http: server gave HTTP response to HTTPS client`.

## Root cause
The guide uses Docker commands. Docker allows HTTP on loopback registries by default, while Podman can require explicit insecure registry configuration or `--tls-verify=false` for this local registry push.

## Changes
- Add a short note after the Docker push example explaining the Podman-specific push option.
- Keep the main local registry example Docker-focused.

## Tests
- `git diff --check`
- `make -C site build`

## Screenshots/Logs
N/A, docs-only change.

Fixes #3468

### DIFF
--- a/site/content/docs/user/local-registry.md
+++ b/site/content/docs/user/local-registry.md
@@ -27,8 +27,13 @@ The registry can be used like this.
 3. Then we'll push it to the registry `docker push localhost:5001/hello-app:1.0`
 4. And now we can use the image `kubectl create deployment hello-server --image=localhost:5001/hello-app:1.0`
 
+> **NOTE**: The example commands above use Docker. If you are using Podman
+> and see `http: server gave HTTP response to HTTPS client` when pushing to
+> `localhost:5001`, configure Podman to treat the local registry as insecure,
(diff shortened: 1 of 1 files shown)


stmcginnis:
Not large enough for the dedicated Known Issues page, and that would make it harder to find. I think it makes sense as a note here.

Thanks!

/lgtm

BenTheElder (commented):
/approve
/lgtm

PR #4192, opened 2026-06-13
fix: respect Docker client proxy config for docker nodes

### LINKED ISSUE #1175: proxy info from docker config is not respected

immanuwell:
fixes #1175

kind was only looking at shell proxy envs. it missed Docker client proxy config from `~/.docker/config.json`, so docker node containers could miss the extra `NO_PROXY` entries kind adds for cluster traffic.

This keeps the old precedence rules too:
- if `HTTP_PROXY` or `HTTPS_PROXY` is set for kind, that still wins
- otherwise the docker provider reads client proxy config, using `DOCKER_HOST` specific entries first and `default` as fallback

How to repro:
1. write `~/.docker/config.json` with `{"proxies":{"default":{"httpProxy":"http://127.0.0.1:9999","httpsProxy":"http://127.0.0.1:9999","noProxy":"example.internal"}}}`
2. run `docker run --rm alpine:3.20 env | grep -i _proxy`
3. Docker injects those proxy vars for the container
4. run kind without `HTTP_PROXY` or `HTTPS_PROXY` in the shell
5. before this patch kind ignores that config path. after this patch it picks it up, and still appends kind `NO_PROXY` values for subnets, node names, and `.svc` names

Checks:
- `go test ./...`
- `go vet ./...`

### DIFF
--- a/pkg/cluster/internal/providers/docker/proxyconfig.go
+++ b/pkg/cluster/internal/providers/docker/proxyconfig.go
@@ -1,0 +1,155 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package docker
+
+import (
+	"encoding/json"
+	"path/filepath"
+	"strings"
+
+	"sigs.k8s.io/kind/pkg/cluster/internal/providers/common"
+	"sigs.k8s.io/kind/pkg/exec"
+)
+
+const (
+	dockerConfigEnv  = "DOCKER_CONFIG"
+	dockerContextEnv = "DOCKER_CONTEXT"
+	dockerHostEnv    = "DOCKER_HOST"
+)
+
+type dockerConfigFile struct {
+	CurrentContext string                       `json:"currentContext"`
+	Proxies        map[string]dockerProxyConfig `json:"proxies"`
+}
+
+type dockerProxyConfig struct {
+	HTTPProxy  string `json:"httpProxy"`
+	HTTPSProxy string `json:"httpsProxy"`
+	NOProxy    string `json:"noProxy"`
+}
+
+func dockerConfigProxyEnvs(
+	getEnv func(string) string,
+	readFile func(string) ([]byte, error),
+	inspectContextHost func(string) (string, error),
+) map[string]string {
+	configPath := dockerConfigPath(getEnv)
+	if configPath == "" {
+		return map[string]string{}
+	}
+
+	rawConfig, err := readFile(configPath)
+	if err != nil {
+		return map[string]string{}
+	}
+
+	cfg := dockerConfigFile{}
+	if err := json.Unmarshal(rawConfig, &cfg); err != nil {
+		return map[string]string{}
+	}
+
+	proxyCfg, ok := dockerProxySettings(cfg.Proxies, dockerHostForProxyLookup(getEnv, cfg, inspectContextHost))
+	if !ok {
+		return map[string]string{}
+	}
+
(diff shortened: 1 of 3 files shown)


pkg/cluster/internal/providers/common/proxy.go
```diff
@@ -35,7 +35,13 @@ const (
 // GetProxyEnvs returns a map of proxy environment variables to their values
 // If proxy settings are set, NO_PROXY is modified to include the cluster subnets
 func GetProxyEnvs(cfg *config.Cluster) map[string]string {
-	return getProxyEnvs(cfg, os.Getenv)
+	return GetProxyEnvsWith(cfg, os.Getenv)
+}
+
+// GetProxyEnvsWith returns proxy environment variables using the provided
+// environment lookup function.
+func GetProxyEnvsWith(cfg *config.Cluster, getEnv func(string) string) map[string]string {
```
stmcginnis: What value is this adding?

pkg/cluster/internal/providers/docker/proxyconfig.go
```diff
@@ -0,0 +1,108 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
stmcginnis: nit: Remove the year. It's incorrect, and current guidance is to not include it.

stmcginnis (commented):
This gets the client proxy settings, but there are also the [service settings](https://docs.docker.com/engine/daemon/proxy/). But I think those would be automatic, and probably be the more expected way to configure this. Not really sure.

Is there something similar for podman, nerdctl, etc?

pkg/cluster/internal/providers/common/proxy.go
immanuwell: you're right, GetProxyEnvsWith wasn't adding meaningful value beyond test injection for this change
I removed it

pkg/cluster/internal/providers/docker/proxyconfig.go
immanuwell: removed

immanuwell:
@stmcginnis I applied all suggestions, PTAL

pkg/cluster/internal/providers/docker/proxyconfig.go
stmcginnis: Forgot to push a commit?

immanuwell:
@stmcginnis I had missed pushing the header fix earlier
it is now included in the ef747abf 
`go test ./pkg/cluster/internal/providers/docker` passes
PTAL

pkg/cluster/internal/providers/docker/proxyconfig.go
immanuwell: sorry, now the year is definitely removed)

pkg/cluster/internal/providers/docker/proxyconfig.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"encoding/json"
+	"path/filepath"
+	"strings"
+
+	"sigs.k8s.io/kind/pkg/cluster/internal/providers/common"
+)
+
+const (
+	dockerConfigEnv = "DOCKER_CONFIG"
```
BenTheElder: is this standard? are there docs we can point to in a code comment? I'm familiar with DOCKER_HOST but not DOCKER_CONFIG, and I want to understand if we're handling remote hosts correctly ...

immanuwell: yes, DOCKER_CONFIG is standard Docker CLI config (https://docs.docker.com/reference/cli/docker/), and Docker's config.json proxies supports both default and per-daemon host entries
I added a code comment pointing at the Docker CLI docs
and I updated the implementation to resolve the active daemon host from DOCKER_CONTEXT / currentContext when DOCKER_HOST is unset, so remote-context cases are handled too

BenTheElder:
LGTM, @stmcginnis any other concerns / suggestions?

Thank you!

stmcginnis:
No concerns from me - looks like everything is covered.

/lgtm

immanuwell:
@BenTheElder @stmcginnis just a friendly reminder

PR #4198, opened 2026-06-22
Clean up some old commented-out code

stmcginnis:
While reviewing other changes in the loadbalancer config file (#4195), noticed we had some old commented-out code that is no longer needed. This is a minor update to clean that up.


BenTheElder (commented):
/lgtm
/approve

PR #4200, opened 2026-06-25
fix: default empty PortMapping protocol to TCP in SetDefaultsNode

carterpewpew:
When `extraPortMappings` omits `protocol` on one entry but another entry on the same `hostPort` sets it to `TCP` explicitly, `validatePortMappings` builds different map keys (`"8080/"` vs `"8080/TCP"`) and silently passes the duplicate through. At provision time all three providers (docker, podman, nerdctl) default empty protocol to TCP, so both mappings compete for the same host port and container creation fails with a bind error.

The fix defaults empty Protocol to PortMappingProtocolTCP in SetDefaultsNode, which runs before Validate(). This is the same pattern used for Image and Role in the same function, and matches what all three providers already do at provision time.

### DIFF
--- a/pkg/internal/apis/config/default.go
+++ b/pkg/internal/apis/config/default.go
@@ -101,4 +101,10 @@ func SetDefaultsNode(obj *Node) {
 	if obj.Role == "" {
 		obj.Role = ControlPlaneRole
 	}
+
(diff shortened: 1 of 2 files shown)


stmcginnis:
Looks reasonable - thanks!

/ok-to-test
/lgtm

PR #4209, opened 2026-07-02
Enable ControlPlaneKubeletLocalMode for kubeadm 1.31 and 1.32

### LINKED ISSUE #3932: HA clusters should use local apiserver for control plane nodes
https://github.com/kubernetes/kubernetes/issues/127316#issuecomment-2354195820

I'm pretty sure right now we route everything through haproxy, only the worker nodes should do that, then I think we're fine.

this is probably also relevant for `kinder` testing workflows

cc @pacoxu @carlory (some kubernetes/kubeadm owners for FYI)

/kind bug

larrasket:
In HA clusters every kubelet, including the control plane nodes', is pointed at the external load balancer. During an upgrade a control plane node's kubelet can then talk to an older apiserver on another node. this fix is the`ControlPlaneKubeletLocalMode` feature gate, which pins control plane kubelets to their local apiserver, but it's alpha and off by default in 1.31/1.32 (on by default since 1.33, removed in 1.36) and kind never enabled it.

Enable the gate for kubeadm 1.31 and 1.32 only.

Fixes #3932

### DIFF
--- a/pkg/cluster/internal/kubeadm/config.go
+++ b/pkg/cluster/internal/kubeadm/config.go
@@ -697,6 +697,19 @@ func Config(data ConfigData) (config string, err error) {
 		data.KubeadmFeatureGates["IPv6DualStack"] = true
 	}
 
+	// Have control plane kubelets talk to their local apiserver instead of
+	// the load balancer, so upgrades don't violate the kubelet/apiserver
+	// version skew policy. The kubeadm gate is alpha and off by default in
+	// 1.31/1.32, on by default since 1.33, and removed in 1.36.
+	// TODO: remove this when we no longer support Kubernetes 1.32.
+	if ver.AtLeast(version.MustParseSemantic("v1.31.0")) &&
+		ver.LessThan(version.MustParseSemantic("v1.33.0")) {
+		if data.KubeadmFeatureGates == nil {
+			data.KubeadmFeatureGates = make(map[string]bool)
+		}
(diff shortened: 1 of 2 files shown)


neolit123 (commented):
> Fixes https://github.com/kubernetes-sigs/kind/issues/3932

if this PR is merged, there should be a ticket to track the cleanup of this change once <= 1.32 is no longer supported in kind.

kubeadm does error on unknown FGs.

stmcginnis:
Just to note, users can build whatever version they want, but in the [latest release](https://github.com/kubernetes-sigs/kind/releases/tag/v0.32.0) we only provided official images back to 1.33.

BenTheElder:
> Just to note, users can build whatever version they want, but in the [latest release](https://github.com/kubernetes-sigs/kind/releases/tag/v0.32.0) we only provided official images back to 1.33.

We also don't currently offer any support for in-place upgrades, for that matter.

/ok-to-test

larrasket:
Opened #4230 to track removing this once kind drops support for Kubernetes <= 1.32.

On kubeadm erroring on unknown feature gates: the gate is only enabled for `>= 1.31, < 1.33`, where it exists and is off by default (on by default from 1.33), so other versions are never handed an unknown gate.

larrasket:
@neolit123 / @BenTheElder could you have a look and /lgtm so it can move to approval? Thanks

PR #4212, opened 2026-07-08
Update golangci-lint and switch to go tools

stmcginnis:
This updates the version of golangci-lint to the latest. It also switches our tools over from the old-style blank import to using the go tools method for better dependency tracking via go.mod.

Also added dependabot config to update the dependencies in hack/tools so we can be a little better about keeping our tooling up to date.


stmcginnis:
/retest

> There are no nodes that your pod can schedule to - check your requests, tolerations, and node selectors ((combined from similar events): 0/141 nodes are available

PR #4213, opened 2026-07-08
Update depedencies

stmcginnis:
This raises our minimum Go version to 1.18 from the previous 1.17. We already had dependency versions that were only claimed compatibility with 1.18, and 1.17 has now been end-of-life for just short of four years.

Raised the dependency versions to the most recent version still claiming support for 1.18 wherever newer versions were available.


BenTheElder:
any notable changes in the deps upstream?

stmcginnis:
This is a little stale now. I'll revisit this after the next release and update to the latest versions. While doing that I'll try to call out any interesting changes I see with them.

stmcginnis:
Superceded by #4254 

/close

PR #4214, opened 2026-07-09
docs: Add install options for Arch Linux

christian-heusel:
The package has now been imported into the official Arch Linux Package
Repositories so it can easily be installed via `pacman` now.


BenTheElder (commented):
/lgtm
/approve
thanks!

PR #4220, opened 2026-07-22
Remove memory limit from node image kindnet container

### LINKED ISSUE #4215: kindnet's 50Mi Guaranteed memory limit causes continuous executable-page refault thrashing (60+ MB/s disk reads) on rotational disks
**What happened**:

On a host where the kind node's storage is a rotational (5400 RPM) disk, the four `kindnetd` pods (one per node) collectively read **60–80 MB/s continuously** from disk, pinning it at ~90% utilization and making the whole host unusably slow (10–15s stalls opening/closing unrelated desktop apps). There is **no OOMKill** — kindnetd runs happily; it just thrashes the disk forever.

**Root cause** (traced end-to-end):

`kindnet-cni` is configured with `requests == limits == { cpu: 100m, memory: 50Mi }`, i.e. **Guaranteed QoS with a hard 50Mi cap** ([const_cni.go](https://github.com/kubernetes-sigs/kind/blob/main/pkg/build/nodeimage/const_cni.go)). The kindnetd binary is a ~40MB Go executable, mapped as **file-backed** (clean) code pages. The Go runtime heap + goroutine stacks push `memory.current` right up against the 50Mi cap. To stay under the limit the kernel does **direct reclaim**, and clean file-backed pages (the executable itself) are the cheapest thing to evict — they can always be re-read from disk. The very next instruction kindnetd runs then **refaults** that code page back in from the node's disk. Reclaim → refault → reclaim → refault, millions of times.

Measured on the affected node (cgroup v2, `memory.stat` for a single kindnetd container):

```
memory.max              = 52428800   (50 MiB)
memory.current          = 51974144   (pinned at the cap)
file                    = 40599552   (~the whole kindnetd binary, as page cache)
workingset_refault_file = 7,024,174
pgscan_direct           = 12,048,492
pgsteal_direct          = 11,421,139
```

Live correlation over 5s for one of the four pods:

```
workingset_refault_file: +16,458 refaults (64 MB of pages)
read_bytes:              +115 MB from disk
```

Signature that pointed here: kindnetd shows huge `read_bytes` in `/proc/<pid>/io` but **near-zero `read()` syscalls** under strace, **very low `majflt`** (readahead batches the refaults so they land as minor faults), RES pinned around 11–13 MB (far below the 40MB binary — its own code keeps getting evicted), and threads sitting in `D` state on the disk. It's pure executable thrash, not real I/O work.

On SSD/NVMe (kind's usual environment) each refault is a few microseconds, so this is invisible. On a rotational disk each refault costs milliseconds and saturates the spindle.

**How to reproduce**:

1. Put the kind node storage (Docker data-root) on a rotational HDD.
2. `kind create cluster` (multi-node makes it worse — the effect is per-node).
3. Observe with `iostat -dx 2` (device at ~90% util, tens of MB/s reads) and per-pod: `cat /sys/fs/cgroup/.../kindnet.../memory.stat` shows `workingset_refault_file` climbing by thousands/sec.

**Workaround** that fixes it completely:

```
kubectl -n kube-system patch ds kindnet --type=json \
  -p='[{"op":"replace","path":"/spec/template/spec/containers/0/resources/limits/memory","value":"256Mi"},
       {"op":"replace","path":"/spec/template/spec/containers/0/resources/requests/memory","value":"128Mi"}]'
```

Giving the cgroup enough headroom to keep the ~40MB of code pages resident stops the refault loop and the disk reads drop to ~zero.

**Suggested fix**:

Raise the default `kindnet-cni` memory limit (e.g. 50Mi → 128–256Mi), and/or set `limits.memory > requests.memory` so the executable's clean pages aren't forced out the instant the heap grows. A 50Mi hard cap is smaller than kindnetd's own on-disk binary, which makes some degree of code-page reclaim inevitable.

**Relationship to #3009**: #3009 is the same 50Mi limit being too tight, but manifests as an *OOMKill* on ARM64 with 64k pages (kernel memory). This is a distinct, non-OOM failure mode on 4k-page x86 with slow storage. Both point at the same undersized default.

**Environment**:
- kind: v0.31.0
- Docker: 29.6.0 (Fedora)
- kernel: 7.1.3-200.fc44.x86_64 (x86_64, 4k pages)
- node image / kindnet: `kindest/kindnetd:v20251212-v0.29.0-alpha-105-g20ccfc88`
- storage: WDC WD40EFRX (5400 RPM, rotational), Docker data-root on it

jboero:
Closes: #4215

### DIFF
--- a/pkg/build/nodeimage/const_cni.go
+++ b/pkg/build/nodeimage/const_cni.go
@@ -137,7 +137,6 @@ spec:
             memory: "50Mi"
           limits:
             cpu: "100m"
-            memory: "50Mi"
         securityContext:
           privileged: false
           capabilities:


jboero:
Half an hour of fork, index, sign up for LFX, sign CLA, wait for CI pass.

All to remove one line of YAML.

aojea:
/ok-to-test

should we just remove all the limits?

> Half an hour of fork, index, sign up for LFX, sign CLA, wait for CI pass.
> 
> All to remove one line of YAML.

and to become a kind contributor ❤️

jboero:
True. Thanks for the great project. It definitely simplifies life better than CRC. I think this is the only limit that was biting. Shouldn't be an issue going forward.

stmcginnis:
/lgtm
/retitle Remove memory limit from node image kindnet container

We probably don't need the CPU limit either, but the memory one is the one that was problematic.

aojea:
/lgtm
/approve

we can always follow up

BenTheElder:
thanks!

pkg/build/nodeimage/const_cni.go
```diff
@@ -137,7 +137,6 @@ spec:
             memory: "50Mi"
           limits:
             cpu: "100m"
-            memory: "50Mi"
```
BenTheElder, line 140: FYI @aojea @stmcginnis we won't see this in GHA until we build a new node image, since it's baked in. We may need to finally start having GHA build node images now that we don't need to compile k8s necessarily.

PR #4221, opened 2026-07-22
Build node images in CI if changed

stmcginnis:
This updates our provider testing to build the node images if it looks like source files have changed. This allows us to validate node image changes as part of normal CI workflows.

If there are changes in `pkg/build` or `images` we will build a node image using official binaries. That image will then be loaded and used for spinning up the test cluster used as part of the CI, for each provider being tested. If nothing has changed that looks like it will impact the node image, then it just pulls and uses the default - the way things have been operating so far.


stmcginnis:
cc @BenTheElder 

Based on https://github.com/kubernetes-sigs/kind/pull/4220#discussion_r3632717198

PR #4227, opened 2026-07-28
Fix index-out-of-range panic parsing docker info DriverStatus

onkar717:
`mountDevMapper()` in `pkg/cluster/internal/providers/docker/util.go` parses `docker info -f '{{json .DriverStatus }}'` output into a `[][]string` and indexed `item[0]`/`item[1]` with no length check. Docker's `DriverStatus` is a list of `[key, value]` pairs, but nothing guarantees every entry has exactly 2 elements across docker versions and storage drivers the podman
provider in this same repo already had to work around an analogous "output format changed between versions" issue on its own storage-info parsing (it now unmarshals into a typed struct instead of raw arrays).

This function runs unconditionally as part of building docker run args for  every node container on every `kind create cluster` call with the docker provider (the default provider), via `commonArgs` in `provision.go`. A single malformed `DriverStatus` entry would panic the whole CLI mid cluster-creation instead of just skipping that entry.

## What this PR does

Extracts the pair-walking loop into a small pure helper, `backingFilesystem`, that skips entries without exactly 2 elements instead of indexing them directly. Extracting it into its own function was necessary to get real unit test coverage without faking the `docker info` exec call `mountDevMapper` itself isn't practical to unit test directly since it shells out to `docker`.

Adds `util_test.go` the first tests this file has ever had covering the normal case, the no-match case, and the two malformed shapes that would have panicked before this fix (single-element entry, empty entry).

## Testing

- `go test ./pkg/cluster/internal/providers/docker/...` passes.
- Confirmed the malformed-entry test cases actually reproduce the original panic: temporarily removed the length guard locally and got the exact  `index out of range [1] with length 1` panic out of `mountDevMapper`'s real code path, then restored the guard and reran green.
- `golangci-lint run` (pinned v2.11.3, repo config at   `hack/tools/.golangci.yml`) reports 0 issues on the changed package.

No associated issue found while reading the docker-provider provisioning path for real, self-contained bugs. Happy to file one if a maintainer would prefer it tracked separately.

### DIFF
--- a/pkg/cluster/internal/providers/docker/util.go
+++ b/pkg/cluster/internal/providers/docker/util.go
@@ -82,14 +82,23 @@ func mountDevMapper() bool {
 	if err := json.Unmarshal([]byte(lines[0]), &dat); err != nil {
 		return false
 	}
-	for _, item := range dat {
-		if item[0] == "Backing Filesystem" {
-			storage = strings.ToLower(item[1])
-			break
-		}
-	}
+	storage = backingFilesystem(dat)
 
 	return storage == "btrfs" || storage == "zfs" || storage == "xfs"
+}
+
+// backingFilesystem returns the lowercased "Backing Filesystem" value from
+// docker's `DriverStatus` pairs, e.g. [["Backing Filesystem","extfs"], ...].
+// DriverStatus entries are not guaranteed to have exactly 2 elements across
+// docker versions/storage drivers, so malformed entries are skipped instead
+// of indexed directly.
+func backingFilesystem(driverStatus [][]string) string {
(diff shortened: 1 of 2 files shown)


BenTheElder:
> No associated issue found while reading the docker-provider provisioning path for real, self-contained bugs. Happy to file one if a maintainer would prefer it tracked separately.

Is this a real bug you experienced, or theoretical?

This PR repeatedly says "docker [...] is not guaranteed to [...]" but broadly docker's CLI data format is what it actually returns rather than a committed spec. Have we seen this in the wild?

https://www.kubernetes.dev/docs/guide/pull-requests/#ai-guidance

PR #4228, opened 2026-07-30
docs: label min kind versions for v1alpha4 config options

### LINKED ISSUE #2026: label required version for config options, and other compatibility details

https://kind.sigs.k8s.io/docs/user/configuration/ should surface what the minimum version of kind is for each config field, any differences in behavior, and the minimum apiVersion for the field.

The versions would probably look best as some sort of marker (table?) below each heading, but we should try some things to see what works best to effectively communicate all of this. Some of it should probably just be text within the details of the feature, renamed fields, etc.

**Why is this needed**:

To more clearly communicate support.

---

I think we should probably just start with v1alpha4, it's probably not worth it to go back and document older alphas in depth, we're not really going to support year old kind versions yet ...

amarkdotdev:
## Summary
Fixes #2026

Adds compatibility markers on the configuration guide for newer `kind.x-k8s.io/v1alpha4` fields where the minimum kind (and Kubernetes) versions matter for users scanning the page. Baseline fields that have been available since early `v1alpha4` are left unmarked to avoid noise. Also documents `networking.dnsSearch`, which was previously missing from the page.

This is intentionally incremental: call out useful newer requirements first, with a shortcode so later fields can follow the same pattern.

## Test plan
- [ ] Preview `site/content/docs/user/configuration.md` and confirm compatibility tables appear only for newer fields (dual-stack, kubeProxyMode additions, dnsSearch, labels)
- [ ] Confirm baseline sections no longer show v0.6.0-era compatibility tables
- [ ] Spot-check versions against release history (dual-stack/labels/`none` v0.11.0, `dnsSearch` v0.18.0, `nftables` v0.23.0)
- [ ] Confirm Hugo renders the new `configcompat` shortcode without errors

### DIFF
--- a/site/content/docs/user/configuration.md
+++ b/site/content/docs/user/configuration.md
@@ -39,6 +39,15 @@ You can also include a full file path like `kind create cluster --config=/foo/ba
 The structure of the `Cluster` type is defined by a Go struct, which is described
 [here](https://pkg.go.dev/sigs.k8s.io/kind/pkg/apis/config/v1alpha4#Cluster).
 
+### Version Compatibility
+
+This page documents the current config apiVersion, `kind.x-k8s.io/v1alpha4`
+(the only supported apiVersion in current kind releases).
+
+Where it helps, newer options include a short compatibility table with the
+minimum kind version (and Kubernetes version when one applies). Older baseline
+`v1alpha4` fields are left unannotated so this page stays easy to scan.
+
 ### A Note On CLI Parameters and Configuration Files
 
 Unless otherwise noted, parameters passed to the CLI take precedence over their
@@ -141,6 +150,7 @@ networking:
 {{< /codeFromInline >}}
 
 ##### Dual Stack clusters
+
 You can run dual stack clusters using `kind` 0.11+, on kubernetes versions 1.20+.
 
 {{< codeFromInline lang="yaml" >}}
@@ -215,11 +225,12 @@ networking:
   disableDefaultCNI: true
 {{< /codeFromInline >}}
 
-
 #### kube-proxy mode
 
-You can configure the kube-proxy mode that will be used, between iptables, nftables (Kubernetes v1.31+), and ipvs.
-By default iptables is used
+You can configure the kube-proxy mode that will be used, between iptables, nftables, and ipvs.
+By default iptables is used.
+
+{{< configcompat field="networking.kubeProxyMode (nftables)" kind="v0.23.0" kubernetes="v1.31" >}}{{< /configcompat >}}
 
 {{< codeFromInline lang="yaml" >}}
 kind: Cluster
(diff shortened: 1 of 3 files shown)


stmcginnis:
I think it may make sense to call out newer changes, but I don't see much value in making the configuration page longer to call out changes in very old versions like v0.6.0. Or maybe we need a separate compatibility reference page with complete details for everything. I know folks are already missing configuration details because this page is too long to quickly parse.

Also note the bot message. You will need to amend your commit and force push it to remove the "Fixes" line. Those issue management keywords are not allowed in commit messages, only in the PR description, as it can have unintended side effects for things like backports.

BenTheElder:
> [...]  but I don't see much value in making the configuration page longer to call out changes in very old versions like v0.6.0 [...]

Agreed. Those versions are unlikely to work at all on common distros currently. Not worth the noise. We don't have nearly that level of commitment to compatibility at this stage anyhow.

amarkdotdev:
Thanks @stmcginnis and @BenTheElder — updated based on your feedback:

- Dropped the old baseline callouts (`v0.9` / `v0.11` tables)
- Kept only newer annotations: `dnsSearch` (v0.18) and `nftables` kubeProxyMode (v0.23 / Kubernetes v1.31)
- Rewrote the commit message to remove the `Fixes` keyword (left the issue link in the PR description only)

Happy to trim further if you want even less on the page. Still needs `/ok-to-test` when convenient.

amarkdotdev:
Friendly bump on this one. CI is green on my side and it's been sitting a few weeks with no review yet. Happy to adjust if priorities shifted or if there's anything I can do to help move it forward. Thanks!

site/content/docs/user/quick-start.md
```diff
@@ -131,11 +131,6 @@ On Windows via Winget (https://github.com/microsoft/winget-pkgs/tree/master/mani
 winget install Kubernetes.kind
 {{< /codeFromInline >}}
 
-On Arch Linux via pacman (https://archlinux.org/packages/extra/x86_64/kind/)
```
stmcginnis, line 134: Is there a reason for removing this?

site/content/docs/user/configuration.md
```diff
@@ -215,11 +225,12 @@ networking:
...
 
-
 #### kube-proxy mode
 
-You can configure the kube-proxy mode that will be used, between iptables, nftables (Kubernetes v1.31+), and ipvs.
-By default iptables is used
+You can configure the kube-proxy mode that will be used, between iptables, nftables, and ipvs.
+By default iptables is used.
+
+{{< configcompat field="networking.kubeProxyMode (nftables)" kind="v0.23.0" kubernetes="v1.31" >}}{{< /configcompat >}}
```
stmcginnis, line 233: This rendered formatting doesn't look right. The fields are larger than the section headers.

Image

It also takes up a lot of screen real estate. Really for these things I think if it matters, a small text note of "this is available as of Kubernetes v1.31)" would be preferable to a big block of fields and values.

site/content/docs/user/configuration.md
```diff
@@ -39,6 +39,15 @@ You can also include a full file path like `kind create cluster --config=/foo/ba
 The structure of the `Cluster` type is defined by a Go struct, which is described
 [here](https://pkg.go.dev/sigs.k8s.io/kind/pkg/apis/config/v1alpha4#Cluster).
 
+### Version Compatibility
```
stmcginnis, line 42: Version compatibility mentioned inline should be self-explanatory. I don't think we need to add a section for this.

stmcginnis (commented):
Thanks for taking a look at this. I'm still not convinced this is needed though.

amarkdotdev:
Hey maintainers, a review when you can would be great.

amarkdotdev:
Closing this PR. Thanks for the review time.

PR #4232, opened 2026-08-05
fix: return batch deletion errors

immanuwell:
Return a nonzero exit code when any batch delete fails, while still trying every requested cluster

Repro
`KIND_EXPERIMENTAL_PROVIDER=docker DOCKER_HOST=unix:///tmp/kind-no-docker.sock kind delete clusters demo`

Before this exits 0 even though Docker is unreachable. Now it exits 1

### DIFF
--- a/pkg/cmd/kind/delete/clusters/deleteclusters.go
+++ b/pkg/cmd/kind/delete/clusters/deleteclusters.go
@@ -33,6 +33,11 @@ type flagpole struct {
 	All        bool
 }
 
+type clusterProvider interface {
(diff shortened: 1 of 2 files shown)


stmcginnis:
This is a change in behavior, but seems more correct to me.

We do have the "delete a non-existent cluster" question that comes up repeatedly, but this would not change that. That being returned as an error is up to the provider, so this would not see it as a problem and would end up returning `nil`. So really it only returns an error if there was a real legitimate error, which would be useful to know if this is being done in scripting.

/lgtm

PR #4233, opened 2026-08-07
Render framed_picture emoji like other emojis

antoineco:
Purely cosmetic change :nail_care: 

For reasons I ignore, a handful of emojis like `U+1F5BC` (FRAME WITH PICTURE) and `U+1F579` (JOYSTICK) render as a monochrome outline unless the variation selector `U+FE0F` (VS16) is appended to them.

The "JOYSTICK" emoji character printed by kind currently has this variation selector, but the "FRAME WITH PICTURE" emoji character doesn't, leading to some inconsistent visuals in certain combinations of terminal emulator + emoji font.

Below is an example in the Ghostty 1.3.1 terminal emulator + the "Noto Color Emoji" 2.051 font family:

kind emoje outline

### DIFF
--- a/pkg/cluster/internal/providers/docker/images.go
+++ b/pkg/cluster/internal/providers/docker/images.go
@@ -37,7 +37,7 @@ func ensureNodeImages(logger log.Logger, status *cli.Status, cfg *config.Cluster
 	for _, image := range common.RequiredNodeImages(cfg).List() {
 		// prints user friendly message
 		friendlyImageName, image := sanitizeImage(image)
-		status.Start(fmt.Sprintf("Ensuring node image (%s) 🖼", friendlyImageName))
+		status.Start(fmt.Sprintf("Ensuring node image (%s) 🖼️", friendlyImageName))
 		if _, err := pullIfNotPresent(logger, image, 4); err != nil {
 			status.End(false)
 			return err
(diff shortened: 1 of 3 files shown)


pkg/cluster/internal/providers/docker/images.go
```diff
@@ -37,7 +37,7 @@ func ensureNodeImages(logger log.Logger, status *cli.Status, cfg *config.Cluster
 	for _, image := range common.RequiredNodeImages(cfg).List() {
 		// prints user friendly message
 		friendlyImageName, image := sanitizeImage(image)
-		status.Start(fmt.Sprintf("Ensuring node image (%s) 🖼", friendlyImageName))
+		status.Start(fmt.Sprintf("Ensuring node image (%s) 🖼️", friendlyImageName))
```
antoineco, line 40: GitHub does not make the change obvious so here a hexadecimal diff:

```diff
-🖼   = F0 9F 96 BC           (U+1F5BC, no VS16)
+🖼️   = F0 9F 96 BC EF B8 8F  (U+1F5BC + U+FE0F)
```

stmcginnis:
Old one works, but this does look more visually consistent.

/ok-to-test

BenTheElder (commented):
/lgtm
/approve

I approve this unicode pendantry, thank you!

PR #4237, opened 2026-08-19
Bump to containerd v2.3.4 and runc v1.4.3

liggitt:
Picks up containerd v2.3.4 (and the runc version containerd [qualified with at that revision](https://github.com/containerd/containerd/blob/v2.3.4/script/setup/runc-version))

This includes https://github.com/containerd/containerd/pull/13454 which picks up the fix for https://github.com/kubernetes/kubernetes/issues/139132

/cc @BenTheElder

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -124,7 +124,7 @@ RUN eval "$(gimme "${GO_VERSION}")" \
 # stage for building containerd
 FROM go-build AS build-containerd
 ARG TARGETARCH GO_VERSION
-ARG CONTAINERD_VERSION="v2.3.1"
+ARG CONTAINERD_VERSION="v2.3.4"
 ARG CONTAINERD_CLONE_URL="https://github.com/containerd/containerd"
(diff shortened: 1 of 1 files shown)


liggitt:
I'm not sure if presubmits / postsubmits do all the things required to exercise this and build the resulting images, or if there's other work to do

BenTheElder:
> I'm not sure if presubmits / postsubmits do all the things required to exercise this and build the resulting images, or if there's other work to do

I'm actually forgetting if we merged all the required changes for that yet or not, but regardless we have follow-up steps to pickup the new image build (won't be pushed before merge) as the default. (So, this PR may test it, but definitely the next PR that bumps the default image reference will test it, and until that point it won't affect kubernetes CI jobs).

liggitt:
Hmm, not sure how to retry the GitHub job

BenTheElder:
that loos like a hard failure, i restarted it once earlier

> /hack/ci/lima-helper.sh: line 22: exec: lima: not found

I don't think it's related but ... that's not great.

PR #4238, opened 2026-08-19
Detect images already loaded with the containerd image store

### LINKED ISSUE #4236: kind load docker-image never detects an already-loaded image with Docker’s containerd image store
**What happened**:

When Docker uses the containerd image store, `kind load docker-image` does not recognize an image it has already loaded. Every invocation reports the image as absent and transfers it again:

```console
$ kind load docker-image arm64v8/alpine:3.22 --name kind
Image: "arm64v8/alpine:3.22" with ID "sha256:3304a409...0a9" not yet present on node "kind-worker", loading...

$ kind load docker-image arm64v8/alpine:3.22 --name kind
Image: "arm64v8/alpine:3.22" with ID "sha256:3304a409...0a9" not yet present on node "kind-worker", loading...
```

After the first load, the image is present, correctly tagged, and resolvable through CRI. Only the presence check fails, causing redundant transfers.

Switching Docker to the classic overlay2 image store fixes the problem on the same machine with the same image and kind version.

Disclaimer: text and research is AI assisted, but reviewed and submitted by a breathing human (hi!).

The first invocation should load the image. Subsequent invocations should detect it on the node and skip loading:

```text
Image ... found to be already present on all nodes.
```

Docker must use the containerd image store:

```console
$ docker info --format '{{json .DriverStatus}}'
[["driver-type","io.containerd.snapshotter.v1"]]
```

Create a cluster and use the architecture-specific Alpine image matching the host:

```bash
kind create cluster --name kind

# arm64:
IMAGE=arm64v8/alpine:3.22

# Use this instead on amd64:
# IMAGE=amd64/alpine:3.22

docker pull "$IMAGE"

kind load docker-image "$IMAGE" --name kind # loads
kind load docker-image "$IMAGE" --name kind # unexpectedly loads again
```

An architecture-specific image is important. A multi-platform tag such as `alpine:3.22` can fail earlier because kind imports the archive using `ctr images import --all-platforms` while content for other platforms is absent. That separate problem is reported in [#3795](https://github.com/kubernetes-sigs/kind/issues/3795).

A foreign-architecture image should also be avoided because CRI does not expose images incompatible with the node platform, producing the same symptom for a different reason.

The apparent root cause is that kind compares identifiers for different OCI objects:

| Side | Command | Returned value |
|---|---|---|
| Host | `docker image inspect -f '{{.Id}}'` | Index/manifest digest |
| Node | `crictl inspecti` -> `.status.id` | Image-config digest |

The comparison occurs at [`docker-image.go#L156`](https://github.com/kubernetes-sigs/kind/blob/v0.32.0/pkg/cmd/kind/load/docker-image/docker-image.go#L156). The values are obtained at [`docker-image.go#L216`](https://github.com/kubernetes-sigs/kind/blob/v0.32.0/pkg/cmd/kind/load/docker-image/docker-image.go#L216) and [`util.go#L131-L143`](https://github.com/kubernetes-sigs/kind/blob/v0.32.0/pkg/cluster/nodeutils/util.go#L131-L143).

With the containerd image store:

```console
$ docker image inspect -f '{{.Id}}' arm64v8/alpine:3.22
sha256:3304a409...0a9

$ docker exec kind-worker crictl inspecti docker.io/arm64v8/alpine:3.22
status.id = sha256:2c15e55d...189
repoTags  = ["docker.io/arm64v8/alpine:3.22"]
```

The node contains the correct image, but Docker reports its index digest while CRI reports the child config digest. These values cannot compare equal.

Control experiment:

| Docker image store | Docker `.Id` | CRI `.status.id` | Result |
|---|---|---|---|
| containerd | `3304a409...` (index) | `2c15e55d...` (config) | Reloads |
| overlay2 | `2c15e55d...` (config) | `2c15e55d...` (config) | Skips |

Our e2e tests load an approximately 1 GB image across four workers. Avoiding the redundant load saves roughly 25 seconds per run.

Disabling the containerd image store is a workaround:

```json
{
  "features": {
    "containerd-snapshotter": false
  }
}
```

However, switching stores hides containers and images created under the other backend and is unsuitable for users who need multi-platform support. Our current workaround is an external wrapper that compares manifest digests before invoking kind.

Possible fixes include comparing Docker's descriptor digest with the digest exposed by `ctr images ls`, selecting the comparison based on the Docker image-store type, or explicitly resolving Docker's config digest before comparing it with CRI.

**Environment:**

  ```text
  kind v0.32.0 go1.26.3 darwin/arm64
  ```

- Runtime info: (use `docker info`, `podman info` or `nerdctl info`):

  ```text
  Docker client: 29.6.2
  Docker server: 29.4.0

  Failing image store:
    Driver=overlayfs
    DriverStatus=[["driver-type","io.containerd.snapshotter.v1"]]

  Working control:
    Driver=overlay2
    Backing Filesystem=btrfs

  Node containerd: v2.2.0
  ```

  ```text
  macOS 26.6.1, Darwin arm64
  Docker provided by OrbStack
  Node architecture: aarch64
  ```

  ```text
  Server: v1.35.0
  Client version not recorded
  ```

  ```text
  None known.
  ```

ilyasslefhaili:
Fixes #4236

**What this fixes**

`docker image inspect -f '{{.Id}}'` does not always report the same kind of digest:

| Docker image store | `.Id` |
|---|---|
| classic (`overlay2`) | image **config** digest |
| containerd | digest of the **target descriptor** (manifest or index) |

`crictl inspecti` on the node only ever reports the config digest, so with the containerd image store the presence check compared two different objects and could never match. Every `kind load docker-image` re-transferred an image that was already on the node, correctly tagged and resolvable through CRI.

**Approach**

Containerd on the node records the target digest, and it matches what Docker reports with the containerd image store:

| image | host `docker .Id` | node `crictl status.id` | node `ctr images ls` DIGEST |
|---|---|---|---|
| single manifest | `b8d31d6e` | `5a13d892` | `b8d31d6e` |
| manifest list | `c510f0be` | `d54c4719` | `c510f0be` |

`imagePresentOnNode` now accepts a match on either the config digest (CRI) or the target digest (containerd). This needs no `docker info` probe or version branching, and cannot produce a false positive since both values are content-addressed digests that uniquely identify image content.

The CRI lookup is tried first, so with the classic image store the check returns on the first match exactly as before.

**Verified manually** on macOS/arm64, Docker 28.3.2 with the containerd image store:

| Case | Before | After |
|---|---|---|
| single-manifest image, repeated load | reloads every time | already present |
| manifest-list image, repeated load | reloads every time | already present |
| image rebuilt with new content, same tag | reloads | reloads |
| image not on node | loads | loads |

Unit tests cover both image stores, a stale image on the node, and absence.

**Out of scope**

`checkIfImageReTagRequired` looks up tags via `crictl inspecti <imageID>`, which also cannot resolve a target digest. Correctness is unaffected — it falls through to the check above — but the re-tag optimization is still unavailable with the containerd image store. Happy to address it here or in a follow-up, whichever you prefer.

**Note on testing**

kind's CI runs the classic image store, where this bug is not observable, so there is no e2e coverage for the fix. Let me know if adding a job with `containerd-snapshotter` enabled is worth doing.

```release-note
Fix `kind load docker-image` repeatedly reloading images that are already present on the node when Docker uses the containerd image store
```

### DIFF
--- a/pkg/cluster/nodeutils/util.go
+++ b/pkg/cluster/nodeutils/util.go
@@ -143,6 +143,33 @@ func ImageID(n nodes.Node, image string) (string, error) {
 	return crictlOut.Status.ID, nil
 }
 
+// ImageDigest returns the digest of the target descriptor (image manifest or
+// image index) recorded by containerd on the node for the given image
+// reference, if present.
+//
+// This is distinct from ImageID: CRI identifies an image by its config digest,
+// while containerd identifies it by the digest of the target descriptor.
+// Docker reports the former as the image ID with the classic image store and
+// the latter with the containerd image store, so both are needed to reliably
+// recognize an image that is already present on the node.
+func ImageDigest(n nodes.Node, image string) (string, error) {
+	lines, err := exec.OutputLines(n.Command(
+		"ctr", "--namespace=k8s.io", "images", "ls", "name=="+image,
+	))
+	if err != nil {
+		return "", err
+	}
+	// output is a table with a header row, and at most one row for an exact
+	// name match: REF TYPE DIGEST SIZE PLATFORMS LABELS
+	for _, line := range lines[1:] {
+		fields := strings.Fields(line)
+		if len(fields) >= 3 && fields[0] == image {
+			return fields[2], nil
+		}
+	}
+	return "", errors.Errorf("image %q not present on node %q", image, n.String())
+}
+
 // ImageTags is used to perform a reverse lookup of the ImageID to list set of available
 // RepoTags corresponding to the ImageID in question
 func ImageTags(n nodes.Node, imageID string) (map[string]bool, error) {
--- a/pkg/cmd/kind/load/docker-image/docker-image.go
+++ b/pkg/cmd/kind/load/docker-image/docker-image.go
@@ -39,7 +39,9 @@ import (
 )
 
 type (
-	imageTagFetcher func(nodes.Node, string) (map[string]bool, error)
+	imageTagFetcher    func(nodes.Node, string) (map[string]bool, error)
+	imageIDFetcher     func(nodes.Node, string) (string, error)
+	imageDigestFetcher func(nodes.Node, string) (string, error)
 )
 
 type flagpole struct {
@@ -153,8 +155,7 @@ func runE(logger log.Logger, flags *flagpole, args []string) error {
 				}
 				continue
 			}
-			id, err := nodeutils.ImageID(node, imageName)
-			if err != nil || id != imageID {
+			if !imagePresentOnNode(node, imageName, imageID, nodeutils.ImageID, nodeutils.ImageDigest) {
 				selectedNodes[node.String()] = node
 				logger.V(0).Infof("Image: %q with ID %q not yet present on node %q, loading...", imageName, imageID, node.String())
 			}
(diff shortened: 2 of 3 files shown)


stmcginnis:
Please note that the Kubernetes contribution policy requires adding a note to the PR description when AI tools are used in the creation of the PR. You must also understand and be able to support the reasons for the included change.

Thanks for taking a crack at this. Any contribution does require the CLA to be signed before we can accept it. Let us know if the bot-posted instructions for this are unclear.

Issue #4244, opened 2026-08-21
Newer kind-images increases `fs.nr_open`


ath88:
**What happened**:

From kind v0.32.0, all pre-built images increase `fs.nr_open` on startup, regardless of existing kernel settings. This causes issues with some applications running in kind, eg. `wodby/varnish:6`, and the referenced issue [here](https://github.com/kubernetes-sigs/kind/issues/4001), which results in very long startup times. 

**What you expected to happen**:
I expected kind to not mess with kernel settings, eg. `fs.nr_open`. 

**How to reproduce it (as _minimally_ and precisely as possible)**:
Initially, make sure your `fs.nr_open` is not already maxed - set it with `sudo sysctl -w fs.nr_open=1048576`.

Now, run `kind create cluster --image kindest/node:v1.35.0` - this image does not change `fs.nr_open`, as it was built for v0.31.0. Run `sysctl fs.nr_open` to verify that the setting is unchanged.

Now, delete the cluster with `kind delete cluster` and start another with `kind create cluster --image kindest/node:v1.35.1` - this image is built for v0.32.0. Run `sysctl fs.nr_open` and see that `fs.nr_open` has been maxed.

To test the resulting startup delay in varnish, run `kubectl create deployment varnish --image wodby/varnish:6` - this application will take a long time to get to a "VCL compiled" log statement if the `fs.nr_open` setting is too high.

**Anything else we need to know?**:
I originally came across this issue, as [systemd on some systems](https://github.com/systemd/systemd/commit/a8b627aaed409a15260c25988970c795bf963812) also maximizes `fs.nr_open`, which gave me some headaches trying to replicate the issue on other systems. Ubuntu 24.04 does not max `fs.nr_open` on startup, but Ubuntu 26.04 does. It is likely that for 24.04, systemd was compiled with `-Dbump-proc-sys-fs-nr-open=false`.

**Environment:**

- kind version: v0.31.0 and v0.32.0
- Runtime info: docker 28.4.0
- OS: Ubuntu 24.04 and Ubuntu 26.04
- Kubernetes version: 1.35.0 and 1.35.1
- Any proxies or other special environment settings?: Nope

BenTheElder:
To be clear: this is not being raised on the *host* right? Of course kind does configure settings within the kind node, it must do things like creating cgroups and setting appropriate sysctls for those cgroups to function at all.

There's no clear solution to this issue. I suggest the application deployments should be tuned.

[...]

PR #4245, opened 2026-08-21
base image: remove LimitNOFILE=infinity from containerd.service

pujitha24:
Motivation:
Since the base image moved from Debian bookworm to trixie, node images have shipped a systemd build with the newer default behavior of unconditionally bumping the fs.nr_open sysctl to its maximum value on boot. containerd.service set LimitNOFILE=infinity, which resolves against fs.nr_open at unit start, so containerd's own NOFILE hard limit is now an extremely large number on trixie-based nodes. Any container whose spec does not set its own rlimits inherits this huge NOFILE from containerd's process limits chain (containerd -> shim -> runc -> container init). Applications that iterate the open file descriptor range at startup (e.g. wodby/varnish:6, as reported in the issue) become very slow to start as a result.

Approach:
Remove the LimitNOFILE=infinity line from containerd.service, leaving LimitNOFILE unset so systemd falls back to its own implicit default, decoupling containerd from whatever value fs.nr_open happens to be bumped to at boot. A comment is added in its place explaining the removal, referencing upstream containerd's identical fix in commit 3ca39ef01608fdd44245c0173bf071682b3bfe3c ("fix: Remove LimitNOFILE from containerd.service") and linking the issue.

Report: https://github.com/kubernetes-sigs/kind/issues/4244
Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>
Assisted-by: claude-sonnet-5 (via Claude Code)

### DIFF
--- a/images/base/files/etc/systemd/system/containerd.service
+++ b/images/base/files/etc/systemd/system/containerd.service
@@ -20,7 +20,13 @@ RestartSec=1
 # in the kernel. We recommend using cgroups to do container-local accounting.
 LimitNPROC=infinity
 LimitCORE=infinity
-LimitNOFILE=infinity
+# LimitNOFILE is intentionally left unset (upstream removed LimitNOFILE=infinity
+# in containerd/containerd@3ca39ef01608). Resolving to "infinity" can inherit an
+# unexpectedly large fs.nr_open value from the host/OS on newer systemd releases,
+# which containerd's child processes (i.e. containers) then also inherit unless
+# they set their own limits, causing slow startup for apps that iterate over the
+# open file descriptor range.
+# https://github.com/kubernetes-sigs/kind/issues/4244
 TasksMax=infinity
 OOMScoreAdjust=-999
 


aojea:
IIRC we went back and forth with this value as it seems to fail in different places when it does not exist and we settled in this ... we may try to find the entire history to decide what path to follow

pujitha24:
Went digging through the history on this one. It's flip-flopped once before: #760 (2019) set it to a fixed `1048576` because `infinity` broke NFS/MySQL on hosts with a bumped `fs.nr_open` — basically the same class of problem we're hitting now. Then #1799 (2020) reverted to `infinity` specifically to match upstream containerd's own unit file at the time (containerd/containerd#4475), over an objection from @thomasjm raising this exact fd-iteration/slow-startup concern. Ben's reasoning there was that kind should track upstream containerd rather than carry a fixed value, so results stay representative of a standard cluster.

This PR is following that same "track upstream" precedent — upstream containerd has since removed the line itself (3ca39ef01608), which is the whole motivation here. I haven't found a prior case in kind's history where leaving `LimitNOFILE` fully unset (as opposed to pinning a fixed low value like `760` did) caused problems — if you're thinking of a specific incident, let me know and I'll look into it.

aojea:
This is the one I remember https://github.com/kubernetes-sigs/kind/pull/760 from the referenced issue ...

pujitha24:
Yes, that's the one from my earlier comment. #760 pinned LimitNOFILE to a fixed 1048576 because `infinity` was breaking NFS/MySQL on hosts with a bumped `fs.nr_open`, which is the same underlying failure mode we're seeing again on trixie. This PR doesn't go back to a fixed value though, it leaves LimitNOFILE unset entirely (systemd's own default), which is a different third option from either extreme and matches what upstream containerd itself settled on. I haven't found evidence of "unset" specifically causing issues in kind's history, but if you remember a case where that caused trouble (as opposed to a low fixed value), let me know and I'll dig into it.

aojea:
can't remember now, @BenTheElder is more knowledgable than me in this area, I prefer him to decide

BenTheElder:
I see https://github.com/containerd/containerd/pull/8924 received quite a bit of discussion.

https://github.com/kubernetes/kops/pull/16300 reverted this change and still sets LimitNOFILE. AFAICT many Kubernetes distros still set a large limit or infinity, including virtually all tools under the Kubernetes project:
https://cs.k8s.io/?q=LimitNOFILE&i=nope&literal=nope&files=&excludeFiles=&repos=
- minikube
- image-builder / cluster-api
- kops
- various test cluster tools

https://github.com/kubernetes/kops/blob/72ebf184c8ec7319e781235c60e3c727df12f436/nodeup/pkg/model/containerd.go#L209

> Applications that iterate the open file descriptor range at startup (e.g. wodby/varnish:6, as reported in the issue) become very slow to start as a result.

Naively ... this seems poorly behaved?

Surely this is a breaking change for applications expecting the opposite?

pujitha24:
Fair pushback, and digging further makes it more than "naively poorly behaved." kops made this exact change (unsetting LimitNOFILE, kubernetes/kops#16151) and had to revert it (kubernetes/kops#16300) because it broke real workloads: kubernetes/kops#16298 reports pod ulimits dropping from 1048576 to 1024 and causing production database issues. So "unset" isn't a safe middle ground here, it's a breaking change in practice for anything that relies on inheriting a large fd limit, same as pinning a low fixed value would be.

That undercuts this PR's approach — kind's containerd.service isn't the right place to fix a slow-starting container that iterates its fd range, since any value we pick here trades one class of app breakage for another. I don't have a confident answer for what kind should do instead (per-container rlimits on the container spec, docs, or just leave it as duplicating upstream containerd's current default), so I'd rather not guess. Happy to close this and let the fs.nr_open issue be tracked separately if that's your preference, or take a specific different approach if you have one in mind.

PR #4249, opened 2026-08-26
Mark V0.33.0 release

stmcginnis:


BenTheElder (commented):
/lgtm
/approve
I also trust you to use good judgement to force merge these when cutting a release, we already discussed cutting one and you have to push the tag directly anyhow, I do tend to split the readme from marking the release just so we know the release is published before the docs are updated but /shrug

PR #4254, opened 2026-09-01
Raise minimum Go version to 1.21

stmcginnis:
This raises the minimum. Along with that we are able to now raise a few dependencies that were blocked by them having a higher Go requirement than our minimum.

Go 1.21 was released in August 2023 and went end of life August 2024. We could arguably go up to a newer version, but in general we try to keep support for older versions to minimize the impact on downstream kind library users.

Includes some cleanup made possible by new language features that are now available with the more recent Go version.

Good to do this soon after an official release to make sure we have plenty of bake time until the next release. Though we've been testing with newer versions of Go for a long time, so I don't expect to run into any issues unless it's related to using a newer version of one of the third party dependencies.


BenTheElder:
Stable debian is on 1.24 these days, debian `go get` used to be one of the problems. Anything on 1.21+ also should have toolchain support, so there's less reason to keep it this low once we cross that threshold. Old stable is still on 1.19 though...

We should probably not push the language level to latest. We don't have a lot of use for it in this small codebase (e.g. I don't think we need generic methods) and pushing it to latest will be annoying for projects that embed it.

Issue #4256, opened 2026-09-02
Bug: pkg/internal/cli/env lacks Cygwin/MSYS2 support and violates NO_COLOR spec


mohitjoshi-hey:
**What happened**:

While auditing the CLI terminal detection logic in `pkg/internal/cli/env/env.go`, I noticed two edge-case issues in how terminal state is calculated:

1. **Missing Git Bash / MSYS2 / Cygwin support on Windows**

   `IsTerminal` currently only checks:

   ```go
   isatty.IsTerminal(v.Fd())
   ```

   Windows MSYS/Cygwin environments, including Git Bash, may use pipes rather than native Windows console APIs. Because of this, `isatty.IsTerminal()` can return `false`, causing KIND to unnecessarily fall back to dumb-terminal behavior.

   The underlying `go-isatty` dependency already supports detecting these environments through `isatty.IsCygwinTerminal()`.

2. **Incorrect handling of an empty `NO_COLOR` value**

   `IsSmartTerminal` currently disables ANSI output whenever `NO_COLOR` is present:

   ```go
   if _, set := lookupEnv("NO_COLOR"); set {
       return false
   }
   ```

   However, the official `NO_COLOR` standard linked in the source states that `NO_COLOR` should only disable color when it is present and **not an empty string**.

   As a result:

   ```bash
   NO_COLOR="" kind get clusters
   ```

   currently disables smart-terminal features, even though an empty `NO_COLOR` value should preserve normal color behavior.

   The existing table-driven test in `env_test.go` also currently enforces this incorrect behavior:

   ```text
   Name: "tty, NO_COLOR=",
   FakeEnv: map[string]string{"NO_COLOR": ""},
   IsSmart: false
   ```

**What you expected to happen**:

1. Users running KIND inside Git Bash, MSYS2, or Cygwin environments on Windows should receive normal smart-terminal functionality, including colors and spinners where supported.

2. Setting `NO_COLOR` to an empty string should not disable ANSI output or smart-terminal features:

   ```bash
   NO_COLOR="" kind <command>
   ```

   This should behave the same as `NO_COLOR` being unset, in accordance with the `NO_COLOR` standard.

**How to reproduce it (as *minimally* and precisely as possible)**:

**Bug 1 — Git Bash / MSYS terminal detection:**

1. Open Git Bash on Windows.
2. Run any KIND command.
3. Observe that KIND may detect the environment as a non-terminal and fall back to dumb-terminal behavior.

[...]

PR #4257, opened 2026-09-02
Fix terminal detection: Cygwin/MSYS support and empty NO_COLOR

pujitha24:
Motivation:

pkg/internal/env/term.go's isSmartTerminal disabled smart-terminal/color
output whenever NO_COLOR was set to any value, including an empty
string. Per the NO_COLOR spec (https://no-color.org/), color should
only be disabled when NO_COLOR is present with a non-empty value.

Approach:

- isSmartTerminal's NO_COLOR check now requires a non-empty value
  before disabling smart-terminal features.
- Updated the table-driven test in term_test.go: the existing
  "tty, NO_COLOR=" case now expects IsSmart=true, and a new
  "tty, NO_COLOR=1" case keeps the disable path covered.

Scope note: this PR originally also proposed using
isatty.IsCygwinTerminal to detect Cygwin/MSYS terminals (e.g. Git
Bash) in IsTerminal. @BenTheElder flagged that changing Windows
terminal-detection behavior without local interactive testing on
those environments is risky, since kind's "smart" output relies on
VT escape codes for things like spinner line-rewriting that may not
render well everywhere. That's a fair concern I can't verify without
a Windows/Cygwin/MSYS environment to test in, so I've dropped that
part from this PR and kept only the NO_COLOR fix, which is narrower
and was already confirmed reasonable. The Cygwin/MSYS detection
improvement is still worth pursuing separately, ideally by someone
who can test it interactively on Windows (the original issue
reporter had expressed interest in taking it on).

Validation:

- go build ./... succeeds.
- go vet ./pkg/internal/env/... is clean.
- golangci-lint --config hack/tools/.golangci.yml run
  ./pkg/internal/env/... reports 0 issues.
- go test ./pkg/internal/env/... -v -cover passes all tests
  (91.7% coverage).
- Confirmed fail-then-pass: with only the source fix reverted (test
  change kept), "tty, NO_COLOR=" failed (expected true, got false);
  restoring the fix makes it pass.

Report: https://github.com/kubernetes-sigs/kind/issues/4256

This PR now only addresses the NO_COLOR part of #4256; the
Cygwin/MSYS detection issue described there remains open.

```release-note
Fix terminal detection: NO_COLOR is now only treated as set when it has a non-empty value, per the NO_COLOR spec.
```

### DIFF
--- a/pkg/internal/env/term.go
+++ b/pkg/internal/env/term.go
@@ -64,7 +64,7 @@ func isSmartTerminal(w io.Writer, GOOS string, lookupEnv func(string) (string, b
 
 	// Explicit request for no ANSI escape codes
 	// https://no-color.org/
-	if _, set := lookupEnv("NO_COLOR"); set {
+	if val, set := lookupEnv("NO_COLOR"); set && val != "" {
 		return false
 	}
 
--- a/pkg/internal/env/term_test.go
+++ b/pkg/internal/env/term_test.go
@@ -99,9 +99,20 @@ func TestIsSmartTerminal(t *testing.T) {
 			Writer:  &testFakeTTY{},
 		},
 		{
+			// https://no-color.org/ only disables color when NO_COLOR is
+			// set to a non-empty value.
 			Name: "tty, NO_COLOR=",
 			FakeEnv: map[string]string{
 				"NO_COLOR": "",
+			},
+			GOOS:    "linux",
+			IsSmart: true,
+			Writer:  &testFakeTTY{},
+		},
+		{
+			Name: "tty, NO_COLOR=1",
+			FakeEnv: map[string]string{
+				"NO_COLOR": "1",
 			},
 			GOOS:    "linux",
 			IsSmart: false,


stmcginnis:
I don't have a way to test this, but seems safe enough.

/ok-to-test

BenTheElder:
I think the author of the issue was looking to work on this once we replied ...

> The Cygwin/MSYS fix (bug 1) cannot be exercised end-to-end on
this non-Windows dev machine, since IsCygwinTerminal only ever
returns true on Windows. It was validated by: confirming the
stub returns false unconditionally on all non-Windows build tags

I'm not comfortable changing the windows behavior without local interactive testing. This may behave poorly. See reply on the issue

BenTheElder:
can we squash these? there's no reason we need a history of iterative changes.

pujitha24:
Squashed into a single commit (774fc389) on top of main — just the NO_COLOR non-empty-value fix, no history of the dropped Cygwin/MSYS attempt.

pujitha24:
/assign @BenTheElder

Squashed down to just the NO_COLOR fix as requested, dropping the Cygwin/MSYS change you were unsure about — let me know if anything else would help review.

pkg/internal/env/term.go
```diff
@@ -64,7 +64,7 @@ func isSmartTerminal(w io.Writer, GOOS string, lookupEnv func(string) (string, b
 
 	// Explicit request for no ANSI escape codes
 	// https://no-color.org/
-	if _, set := lookupEnv("NO_COLOR"); set {
+	if val, set := lookupEnv("NO_COLOR"); set && val != "" {
```
retornam:
```
if Getenv("NO_COLOR") != "" {
    return false
}
```

I apologize for the drive-by review comment but as the spec collapses unset and empty into the same behavior, I think `Getenv` might be the appropriate choice here.

pujitha24:
Good catch, thanks. Switched to reusing the `getenv` helper already defined a few lines up (`getenv("NO_COLOR") != ""`) instead of the separate `lookupEnv` + `set` check — same semantics since unset and empty both come back as `""`, and now it's consistent with how TERM/WT_SESSION are checked in the same function.

PR #4264, opened 2026-09-17
rootless: distinguish missing cgroup controllers from missing Delegate=yes

### LINKED ISSUE #3868: kind incorrectly claims "Delegate=yes" is not configured
**What happened**:

Seeing this error message despite it being configured.
```
ERROR: failed to create cluster: running kind with rootless provider requires setting systemd property "Delegate=yes", see https://kind.sigs.k8s.io/docs/user/rootless/
```

At least a different error message as `Delegate=True` (appears to be) properly configured.

It's hard to provide minimalistic steps as I currently don't know what is going wrong and it very likely depends upon system specifics.

However what I can provide:
1. Edit `/etc/systemd/system/user@.service.d/delegate.conf` as in the docs
2. Run `systemctl daemon-reload`
3. Reboot (or relogin/restart user@1000.service)
4. Verify output of `systemctl --user status` does not contain a `cgroup1` tainted and in fact using cgroup2.
5. Run `kind create cluster` => Above error
6. For a quick test, also try the other suggestions from the documentation e.g. `systemd-run --scope --user -p "Delegate=yes" kind cluster create` => Same error.

I'm on Debian Trixie within the Linux Environment of a Chromebook and using podman.

**Environment:**

- kind version: (use `kind version`): kind v0.27.0 go1.23.6 linux/amd64
- Runtime info: (use `docker info`, `podman info` or `nerdctl info`):
```
host:
  arch: amd64
  buildahVersion: 1.38.1
  cgroupControllers: []
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon_2.1.12-3_amd64
    path: /usr/bin/conmon
    version: 'conmon version 2.1.12, commit: unknown'
  cpuUtilization:
    idlePercent: 99.98
    systemPercent: 0.01
    userPercent: 0.01
  cpus: 4
  databaseBackend: sqlite
  distribution:
    codename: trixie
    distribution: debian
    version: unknown
  eventLogger: journald
  freeLocks: 2048
  hostname: penguin
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 165536
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 165536
      size: 65536
  kernel: 6.6.56-05934-gf790fe104dba
  linkmode: dynamic
  logDriver: journald
  memFree: 2616881152
  memTotal: 2812977152
  networkBackend: netavark
  networkBackendInfo:
    backend: netavark
    dns:
      package: aardvark-dns_1.12.2-2_amd64
      path: /usr/lib/podman/aardvark-dns
      version: aardvark-dns 1.12.2
    package: netavark_1.12.1-9_amd64
    path: /usr/lib/podman/netavark
    version: netavark 1.12.1
  ociRuntime:
    name: crun
    package: crun_1.18.2-1_amd64
    path: /usr/bin/crun
    version: |-
      crun version 1.18.2
      commit: 00ab38af875ddd0d1a8226addda52e1de18339b5
      rundir: /run/user/1000/crun
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +WASM:wasmedge +YAJL
  os: linux
  pasta:
    executable: /usr/bin/pasta
    package: passt_0.0~git20250121.4f2c8e7-1_amd64
    version: |
      pasta 0.0~git20250121.4f2c8e7-1
      Copyright Red Hat
      GNU General Public License, version 2 or later
        <https://www.gnu.org/licenses/old-licenses/gpl-2.0.html>
      This is free software: you are free to change and redistribute it.
      There is NO WARRANTY, to the extent permitted by law.
  remoteSocket:
    exists: true
    path: /run/user/1000/podman/podman.sock
  rootlessNetworkCmd: pasta
  security:
    apparmorEnabled: false
    capabilities: CAP_CHOWN,CAP_DAC_OVERRIDE,CAP_FOWNER,CAP_FSETID,CAP_KILL,CAP_NET_BIND_SERVICE,CAP_SETFCAP,CAP_SETGID,CAP_SETPCAP,CAP_SETUID,CAP_SYS_CHROOT
    rootless: true
    seccompEnabled: true
    seccompProfilePath: /usr/share/containers/seccomp.json
    selinuxEnabled: false
  serviceIsRemote: false
  slirp4netns:
    executable: /usr/bin/slirp4netns
    package: slirp4netns_1.2.1-1+b1_amd64
    version: |-
      slirp4netns version 1.2.1
      commit: 09e31e92fa3d2a1d3ca261adaeb012c8d75a8194
      libslirp: 4.8.0
      SLIRP_CONFIG_VERSION_MAX: 4
      libseccomp: 2.5.5
  swapFree: 0
  swapTotal: 0
  uptime: 0h 43m 10.00s
  variant: ""
plugins:
  authorization: null
  log:
  - k8s-file
  - none
  - passthrough
  - journald
  network:
  - bridge
  - macvlan
  - ipvlan
  volume:
  - local
registries: {}
store:
  configFile: /home/user/.config/containers/storage.conf
  containerStore:
    number: 0
    paused: 0
    running: 0
    stopped: 0
  graphDriverName: overlay
  graphOptions: {}
  graphRoot: /home/user/.local/share/containers/storage
  graphRootAllocated: 10737418240
  graphRootUsed: 3600936960
  graphStatus:
    Backing Filesystem: btrfs
    Native Overlay Diff: "true"
    Supports d_type: "true"
    Supports shifting: "false"
    Supports volatile: "true"
    Using metacopy: "false"
  imageCopyTmpDir: /var/tmp
  imageStore:
    number: 0
  runRoot: /run/user/1000/containers
  transientStore: false
  volumePath: /home/user/.local/share/containers/storage/volumes
version:
  APIVersion: 5.3.2
  Built: 1737979078
  BuiltTime: Mon Jan 27 12:57:58 2025
  GitCommit: ""
  GoVersion: go1.23.5
  Os: linux
  OsArch: linux/amd64
  Version: 5.3.2
```
```
PRETTY_NAME="Debian GNU/Linux trixie/sid"
NAME="Debian GNU/Linux"
VERSION_CODENAME=trixie
ID=debian
```
- Kubernetes version: (use `kubectl version`): not present (yet)
* systemd is invoked by using `/sbin/init systemd.unified_cgroup_hierarchy=1` by LXC
* I also went into termina and reconfigured the penguin container (penguin is the "system" where systemd and kind are running in) to enable cgroup2 to begin with. I could still be missing some permissions, but kind doesn't provide a good enough error message for me to know. This is also why I'm reporting this as an invalid error message instead of a cluster creation issue.

harrisleesh:
Fixes #3868

**What this PR does / why we need it**:

When running kind with a rootless provider inside an LXC container (e.g. the Chrome OS Linux development environment from the issue), the root `/sys/fs/cgroup/cgroup.controllers` can be empty even though cgroup v2 is in use — the outer host does not delegate any controllers into the guest. In that situation kind unconditionally reported:

```
running kind with rootless provider requires setting systemd property "Delegate=yes", see https://kind.sigs.k8s.io/docs/user/rootless/
```

which is misleading: no amount of `Delegate=yes` configuration can help when there is nothing systemd could delegate (podman reports `cgroupControllers: []` in this case, so all three controller capability flags are false).

Following the analysis in #3868 and the maintainer suggestions there (better error message + known-issues entry), this PR:

- `pkg/cluster/internal/create/create.go`: splits provider info validation into a testable `validateProviderInfo` and distinguishes two failure modes:
  - **no controllers available at all** → a dedicated error explaining that if `Delegate=yes` is already configured, the host itself may not provide cgroup v2 controllers (LXC / nested containers), linking to the new known-issues entry;
  - **some controllers missing** → the existing `Delegate=yes` error, now listing exactly which controllers (`cpu`, `memory`, `pids`) are missing.
- `pkg/cluster/internal/create/create_test.go`: new unit tests covering both cases plus the existing rootful/rootless/cgroup-v1 paths.
- `site/content/docs/user/known-issues.md`: new "Missing cgroup Controllers inside LXC" section describing how to diagnose (`cat /sys/fs/cgroup/cgroup.controllers`) and why it cannot be fixed from inside the container, as suggested by @BenTheElder.
- `site/content/docs/user/rootless.md`: updated the troubleshooting paragraph to reference the new error wording and the known-issues entry.

Note: #4205 addressed the same issue but has been inactive since June; it reworded the error generically without distinguishing the "no controllers at all" case from the plain missing-`Delegate=yes` case.

**Testing**:

- `go build ./...`, `go vet`, and `go test ./pkg/cluster/internal/create/...` pass (new `TestValidateProviderInfo` covers all branches).
- I could not reproduce the actual LXC environment (development machine is macOS), so the empty-controllers case is covered at the unit level via `ProviderInfo` with all controller flags false, which is what the podman/docker providers produce when `cgroupControllers` is empty.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pkg/cluster/internal/create/create_test.go
```diff
@@ -0,0 +1,127 @@
+/*
+Copyright 2026 The Kubernetes Authors.
```
stmcginnis, line 2: Copyright year should be removed now based on current copyright header guidance.

stmcginnis (commented):
Claude line should be updated in the PR description to just note that AI tooling was used. Specific tool names are discouraged as they can be used to promote specific products or vendors.

/ok-to-test

PR #4265, opened 2026-09-21
Fix empty NO_COLOR disabling smart terminal output

### LINKED ISSUE #4256: Bug: pkg/internal/cli/env lacks Cygwin/MSYS2 support and violates NO_COLOR spec

vulragrag-star:
## Motivation

`isSmartTerminal` disabled smart-terminal / color output whenever `NO_COLOR` was set to any value, including an empty string. Per the [NO_COLOR](https://no-color.org/) spec, color should only be disabled when `NO_COLOR` is present with a **non-empty** value.

```bash
NO_COLOR= kind version   # should keep smart-terminal behavior
NO_COLOR=1 kind version  # should disable it
```

Fixes #4256 (NO_COLOR half only; Cygwin/MSYS terminal detection left alone per maintainer feedback on #4257).

## Changes

- Check `getenv("NO_COLOR") != ""` (same helper already used for `TERM` / `WT_SESSION`) instead of `lookupEnv` presence alone.
- Update the table-driven test: empty `NO_COLOR=` now expects smart; add `NO_COLOR=1` to cover the disable path.

## Test

```
go test ./pkg/internal/env/ -count=1
```


stmcginnis:
Thanks for the contribution. Please note that the CLA will need to be signed before we could accept this. Let us know if you have any questions that the bot instructions left unclear.

